PR #46, opened 2021-11-09
Rethrow intellisense exceptions in IntellisenseTestBase

bryancroteau-MSFT:
The Get function in StringResources.cs throws exceptions when resources are not found. Line 55 in Intellisense.cs silently ignores exceptions and returns an empty suggestion list.

This change rethrows any intellisense exceptions in the unit tests so that the error messages will be more useful in the case where those tests fail because of an exception.


lesaltzm (approved):
🚢 

Yeah, absolutely a good idea.

PR #47, opened 2021-11-09
Add error context declaration to math functions

CarlosFigueiraMSFT:
We want to add error context to math functions in Power Apps Canvas apps, and that require those functions to be tagged with requiring error context.


lesaltzm (approved):
As discussed offline, in the future the `RequiresErrorContext` flag should be removed from TexlFunction. It's only used by the JS code generator, and should be managed there if possible. 

In the short term, :ship:

PR #53, opened 2021-11-11
Add UTC functions and locale strings for SQL compiler

ian-legler:
Add UTCNow, UTCToday, IsUTCToday function implementations and resources
Add updated error strings for formula columns scenarios
Update unit test infra to be able to tag engine-independent compilation errors with "Compile Error"


MikeStall (approved):
:shipit:

lesaltzm:
Nit: update the PR title to something more meaningful

lesaltzm (approved):
:ship:

src/libraries/Microsoft.PowerFx.Core/Localization/Strings.cs
```diff
@@ -160,7 +160,9 @@ internal static class TexlStrings
         public static StringGetter RandBetweenArg2 = (b) => StringResources.Get("RandBetweenArg2", b);
 
         public static StringGetter AboutNow = (b) => StringResources.Get("AboutNow", b);
+        public static StringGetter AboutUTCNow = (b) => StringResources.Get("AboutNow", b);
```
bryancroteau-MSFT, line 163: Is this supposed to use the same resource string as the normal AboutNow function?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IsUTCToday.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    // IsUTCToday(date: d): b
+    internal sealed class IsUTCTodayFunction : BuiltinFunction
+    {
+        // Multiple invocations may result in different return values.
+        public override bool IsStateless => false;
+        public override bool IsSelfContained => true;
+        public override bool SupportsParamCoercion => true;
+        public IsUTCTodayFunction()
+            : base("IsUTCToday", TexlStrings.AboutIsUTCToday, FunctionCategories.Information, DType.Boolean, 0, 1, 1, DType.DateTime)
```
bryancroteau-MSFT, line 24: We should not be using the same DType for both DateTime local and DateTime utc

bryancroteau-MSFT (changes requested):

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IsUTCToday.cs
```diff
@@ -0,0 +1,51 @@
...
+            Contracts.AssertValue(args);
+            Contracts.AssertAllValues(args);
+            Contracts.AssertValue(argTypes);
+            Contracts.Assert(args.Length == argTypes.Length);
+            Contracts.AssertValue(errors);
+            Contracts.Assert(MinArity <= args.Length && args.Length <= MaxArity);
+            bool fValid = base.CheckInvocation(args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
+            DType type0 = argTypes[0];
+            // Arg0 should not be a Time
+            if (type0.Kind == DKind.Time)
```
bryancroteau-MSFT, line 42: Should we check for DKind.Time before or after we call the base CheckInvocation (which may in fact produce a coercion and modify the nodeToCoercedTypeMap)?

bryancroteau-MSFT (changes requested):

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IsUTCToday.cs
ian-legler: This is the logic used today for IsToday

bryancroteau-MSFT (approved):

PR #54, opened 2021-11-12
Initial Lexer changes to support String Interpolation

bryancroteau-MSFT:
Draft PR demonstrating the Lexer changes to add String Interpolation to PowerFx.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/FeatureFlags.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/FeatureFlags.cs
@@ -7,7 +7,7 @@ namespace Microsoft.PowerFx
     /// Hosts can enable these flags to try out early features. 
     /// Any flags will eventually default to true and then get removed once the feature is finalized. 
     /// </summary>
-    public static class FeatureFlags
+    internal static class FeatureFlags
     {
         /// <summary>
         /// Enable String Interpolation feature. 
--- a/src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
@@ -978,6 +978,14 @@ namespace Microsoft.PowerFx.Core.Lexer
 
         private sealed class LexerImpl
         {
+            /*
+             * The Mode of the lexer, required because the behavior of the lexer changes
+             * when lexing inside of a String Interpolation, for example $"Hello {"World"}"
+             * has special lexing behavior. In theory, you could do this with just 2 modes,
+             * but we are using a 3rd mode, Island, to help keep track of when we need
+             * to produce IslandStart and IslandEnd tokens, which will be used by the
+             * Parser to correctly organize the string interpolation into a function call.
+            */
             public enum LexerMode
             {
                 Normal,
@@ -1517,7 +1525,9 @@ namespace Microsoft.PowerFx.Core.Lexer
                         char nextCh;
                         if (Eof || CharacterUtils.IsLineTerm(nextCh = PeekChar(1)) || !IsCurlyClose(nextCh))
                         {
-                            return new ErrorToken(GetTextSpan());
+                            var res = new ErrorToken(GetTextSpan());
+                            NextChar();
+                            return res;
                         }
                         // If we are here, we are seeing a close curly followed immediately by another
                         // close curly. That is an escape sequence for close curly characters.
--- a/src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
@@ -978,14 +978,12 @@ namespace Microsoft.PowerFx.Core.Lexer
 
         private sealed class LexerImpl
         {
-            /*
-             * The Mode of the lexer, required because the behavior of the lexer changes
-             * when lexing inside of a String Interpolation, for example $"Hello {"World"}"
-             * has special lexing behavior. In theory, you could do this with just 2 modes,
-             * but we are using a 3rd mode, Island, to help keep track of when we need
-             * to produce IslandStart and IslandEnd tokens, which will be used by the
-             * Parser to correctly organize the string interpolation into a function call.
-            */
+            // The Mode of the lexer, required because the behavior of the lexer changes
+            // when lexing inside of a String Interpolation, for example $"Hello {"World"}"
+            // has special lexing behavior.In theory, you could do this with just 2 modes,
+            // but we are using a 3rd mode, Island, to help keep track of when we need
+            // to produce IslandStart and IslandEnd tokens, which will be used by the
+            // Parser to correctly organize the string interpolation into a function call.
             public enum LexerMode
             {
                 Normal,
--- a/src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
@@ -1512,6 +1512,18 @@ namespace Microsoft.PowerFx.Core.Lexer
                         _sb.Append(ch);
                         NextChar();
                     }
+                    else if (IsCurlyClose(ch))
+                    {
+                        char nextCh;
+                        if (Eof || CharacterUtils.IsLineTerm(nextCh = PeekChar(1)) || !IsCurlyClose(nextCh))
+                        {
+                            return new ErrorToken(GetTextSpan());
+                        }
+                        // If we are here, we are seeing a close curly followed immediately by another
+                        // close curly. That is an escape sequence for close curly characters.
+                        _sb.Append(ch);
+                        NextChar();
+                    }
                     else if (!CharacterUtils.IsFormatCh(ch))
                         _sb.Append(ch);
 
--- a/src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
@@ -1495,13 +1495,22 @@ namespace Microsoft.PowerFx.Core.Lexer
                         // If we are here, we are seeing a double quote followed immediately by another
                         // double quote. That is an escape sequence for double quote characters.
                         _sb.Append(ch);
+                        NextChar();
                     }
                     else if (IsCurlyOpen(ch))
                     {
-                        // Island start, do not call NextChar()
-                        if (Eof)
-                            return new ErrorToken(GetTextSpan());
-                        return new StrLitToken(_sb.ToString(), GetTextSpan());
+                        char nextCh;
+                        if (Eof || CharacterUtils.IsLineTerm(nextCh = PeekChar(1)) || !IsCurlyOpen(nextCh))
+                        {
+                            // Island start, do not call NextChar()
+                            if (Eof)
+                                return new ErrorToken(GetTextSpan());
+                            return new StrLitToken(_sb.ToString(), GetTextSpan());
+                        }
+                        // If we are here, we are seeing a open curly followed immediately by another
+                        // open curly. That is an escape sequence for open curly characters.
+                        _sb.Append(ch);
+                        NextChar();
                     }
                     else if (!CharacterUtils.IsFormatCh(ch))
                         _sb.Append(ch);
--- a/src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
@@ -648,6 +648,15 @@ namespace Microsoft.PowerFx.Core.Lexer
         // Returns true if the specified character is the start/end identifier delimiter.
         public static bool IsIdentDelimiter(char ch) => ch == IdentifierDelimiter;
 
+        // Returns true if the specified character starts an interpolated string.
+        public static bool IsInterpolatedStringStart(char ch, char nextCh) => ch == '$' && nextCh == '\"';
+
+        // Returns true if the specified character is an open curly bracket, used by interpolated strings.
+        public static bool IsCurlyOpen(char ch) => ch == '{';
+
+        // Returns true if the specified character is an open curly bracket, used by interpolated strings.
+        public static bool IsCurlyClose(char ch) => ch == '}';
+
         // Returns true if the specified character is a valid string delimiter.
         public static bool IsStringDelimiter(char ch) => ch == '\"';
 
@@ -969,11 +978,19 @@ namespace Microsoft.PowerFx.Core.Lexer
 
         private sealed class LexerImpl
         {
+            public enum LexerMode
+            {
+                Normal,
+                Island,
+                StringInterpolation
+            }
+
             private readonly TexlLexer _lex;
             private readonly string _text;
             private readonly int _charCount;
             private readonly StringBuilder _sb; // Used while building a token.
             private readonly bool _allowReplaceableTokens;
+            private readonly Stack<LexerMode> _modeStack;
 
             private int _currentPos; // Current position.
             private int _currentTokenPos; // The start of the current token.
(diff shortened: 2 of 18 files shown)


src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -1365,6 +1428,90 @@ private Token LexStringLit()
...
+                            // Interpolated string end, do not call NextChar()
+                            if (Eof)
+                                return new ErrorToken(GetTextSpan());
+                            return new StrLitToken(_sb.ToString(), GetTextSpan());
+                        }
+                        // If we are here, we are seeing a double quote followed immediately by another
+                        // double quote. That is an escape sequence for double quote characters.
+                        _sb.Append(ch);
+                    }
+                    else if (IsCurlyOpen(ch))
```
lesaltzm, line 1499: Unless I'm missing something, C# behavior treats `{` as the start of an island, but `{{` as a literal '{' char in the string, not starting an island. I think we probably want to replicate that behavior, so we need 1 char lookahead here.

bryancroteau-MSFT: That seems reasonable. Good catch.

bryancroteau-MSFT: Added that logic, how does it look?

lesaltzm: Looks good. Thanks.

MikeStall:
> using System;

nit: add headers

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Lexer/Tokens/IslandEndToken.cs:1 in 85c3ef3. [](commit_id = 85c3ef3c3e0d102c2d0df56d2dbb65c4f55e1d9f, deletion_comment = False)

MikeStall:
nit: can you add tests (LexerTests)? That also helps reviewers follow along.

src/tests/Microsoft.PowerFx.Core.Tests/LexerTests.cs
```diff
@@ -568,5 +568,47 @@ public void TestUnsupportedDecimalSeparatorCausesFallback()
 
             CultureInfo.CurrentCulture = oldCulture;
         }
+
+        [Fact]
+        public void TestStringInterpolation()
```
MikeStall, line 573: What about a recursive example:

$"One {$"Two {"Three"}"} Four"

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -666,6 +668,51 @@ private Identifier ParseIdentifier(Token at = null)
...
+                        ops.Push(new KeyToken(TokKind.Ampersand, token.Span));
+                        break;
+                    default:
+                        args.Push(ParseExpr(Precedence.None));
+                        break;
+                }
+            }
+
+            TexlNode binaryNode = null;
+            while (args.Count > 1)
```
lesaltzm, line 701: Should we be generating this as a bunch of binary ops, instead of a single callnode to Concatenate with the args as an ordered list? Either way works, just wondering why this choice?

src/libraries/Microsoft.PowerFx.Core/Public/ExpressionError.cs
```diff
@@ -13,6 +13,11 @@ public class ExpressionError
         public ErrorKind Kind { get; set; }
         public DocumentErrorSeverity Severity { get; set; }
 
+        public ExpressionError()
```
lesaltzm, line 16: intentional?

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
bryancroteau-MSFT: Honestly I don't know how to deal with all the parameters to the Concatenate function, for example the sources list and trivia list inputs, because we are generating a Call Node without one appearing in the source.

src/tests/Microsoft.PowerFx.Core.Tests/LexerTests.cs
bryancroteau-MSFT: Added

src/libraries/Microsoft.PowerFx.Core/Public/ExpressionError.cs
bryancroteau-MSFT: Removed

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/string.txt
```diff
@@ -280,4 +280,13 @@ true
...
+#Error
+
+>> $"Hello {"World!"}"
+"Hello World!"
+
+>> $"Hello {"World!"} 2"
+"Hello World! 2"
+
+>> $"One {$"Two {"Three"}"} Four"
+"One Two Three Four"
```
CarlosFigueiraMSFT, line 292: Can you also add tests for the `{{` "escape"? Something like

    >> $"This string {{{"has"}}} curly braces"
    "This string {has} curly braces"

src/tests/Microsoft.PowerFx.Core.Tests/LexerTests.cs
```diff
@@ -568,5 +568,71 @@ public void TestUnsupportedDecimalSeparatorCausesFallback()
...
+            Assert.Equal(7, tokens.Length);
+            Assert.Equal(TokKind.StrInterpStart, tokens[0].Kind);
+            Assert.Equal(TokKind.StrLit, tokens[1].Kind);
+            Assert.Equal(TokKind.IslandStart, tokens[2].Kind);
+            Assert.Equal(TokKind.Ident, tokens[3].Kind);
+            Assert.Equal(TokKind.IslandEnd, tokens[4].Kind);
+            Assert.Equal(TokKind.StrInterpEnd, tokens[5].Kind);
+            Assert.Equal(TokKind.Eof, tokens[6].Kind);
+
+            value = "$\"Hello {Table({a: 5})} World!\"";
```
MikeStall, line 590: >           value = "$\"Hello {Table({a: 5})} World!\""; [](http://example.com/codeflow?start=1&length=55)

nit: can you split these into separate [Fact] tests so that the run independently?

src/tests/Microsoft.PowerFx.Core.Tests/LexerTests.cs
```diff
@@ -568,5 +568,71 @@ public void TestUnsupportedDecimalSeparatorCausesFallback()
...
+            tokens = TexlLexer.LocalizedInstance.LexSource(value);
+            Assert.NotNull(tokens);
+            Assert.Equal(7, tokens.Length);
+            Assert.Equal(TokKind.StrInterpStart, tokens[0].Kind);
+            Assert.Equal(TokKind.StrLit, tokens[1].Kind);
+            Assert.Equal(TokKind.IslandStart, tokens[2].Kind);
+            Assert.Equal(TokKind.Ident, tokens[3].Kind);
+            Assert.Equal(TokKind.IslandEnd, tokens[4].Kind);
+            Assert.Equal(TokKind.StrInterpEnd, tokens[5].Kind);
+            Assert.Equal(TokKind.Eof, tokens[6].Kind);
```
MikeStall, line 588: nit: Could we use a test helper to make this more streamlined. ie:

AssertTokens("$\"Hello {name}\"", 
    TokKind.StrInterpStart, 
    TokKind.StrLit, 
 ....);  

AssertTokens(string text, params TokKind[] expected);

src/tests/Microsoft.PowerFx.Core.Tests/LexerTests.cs
CarlosFigueiraMSFT, line 590: Just to confirm, this line is a pure lexer test, since this is an invalid expression, correct?

lesaltzm: Yup. This test is just looking at how it's handling nested { within a {} section

CarlosFigueiraMSFT: Got it. We could use something like `"$\"Hello {Table({a: 5}).a} World!\""` that would be a valid expression while still validating that. But if this is a lexer-only test we don't need to worry too much.

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
MikeStall: Flat here could actually be important.  The nature of $".." usage could lead to abnormally long concatentations; and our binary op is inefficient. 

I think this would just be building up the arg list to Call().

bryancroteau-MSFT: Rewrote the parser changes to use Concatenate instead of a chain of & operators. One consequence of this is that I had to comment out an assert statement in the constructor of IdentToken.cs because the length of the span is incorrect when generating that token with the text "Concatenate".

We have 3 options:

1. Leave it the way it is in this PR
2. Make a new function named $" and map that function name to Concatenate at a later stage
3. Implement a new AST node to fully represent this feature, not as syntactic sugar but as an actual language construct

lesaltzm: I'm leaning towards 1 at the moment. For 2, I don't think we have a place to do that cleanly, and for 3, we'd basically be replicating a specialized version of all the various CallNode visitors later on. Removing that check does seem a little odd, but I think the benefits of keeping this feature limited to Lexer/Parser outweigh the weirdness of an IdentToken that doesn't map to a literal ident in the source.

bryancroteau-MSFT: Agreed

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/string.txt
bryancroteau-MSFT: Added

src/tests/Microsoft.PowerFx.Core.Tests/LexerTests.cs
bryancroteau-MSFT: Done

src/tests/Microsoft.PowerFx.Core.Tests/LexerTests.cs
bryancroteau-MSFT: Done

src/tests/Microsoft.PowerFx.Core.Tests/LexerTests.cs
bryancroteau-MSFT: Just a lexer test

src/libraries/Microsoft.PowerFx.Core/Lexer/Tokens/IdentToken.cs
```diff
@@ -21,7 +21,7 @@ public IdentToken(string val, Span span)
             : this(val, span, false, false)
         {
             Contracts.AssertValue(val);
-            Contracts.Assert(val.Length == span.Lim - span.Min);
+            Contracts.Assert(val.Length == span.Lim - span.Min || val == "Concatenate");
```
MikeStall, line 24: >|| val == "Concatenate" [](http://example.com/codeflow?start=63&length=23)

This deserves a comment!

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/string.txt
```diff
@@ -291,4 +291,25 @@ true
 "ab"
 
 >> Concatenate(Text(1/0, "000"), "b", Text(7))
-#Error
\ No newline at end of file
+#Error
+
+>> $"Hello {"World!"}"
```
MikeStall, line 296: >>> [](http://example.com/codeflow?start=0&length=2)

A few more ... 

$"Hello"
- this should test that new 1 arg overload of Concat.

$"{123}"
- testing cases with empty literals. Plus testing with values that need to get coerced to string (do we support that?). 

$"{123}{456}"
- testing immediately consecutive open and close.

MikeStall: also, can we test error conditions, like:

$"{" // missing close
$"}" // missing open

(maybe not in .txt , but somewhere)

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -969,11 +978,19 @@ internal static void ChoosePunctuators(ILanguageSettings loc, out string dec, ou
 
         private sealed class LexerImpl
         {
+            public enum LexerMode
+            {
+                Normal,
+                Island,
```
MikeStall, line 984: >Island [](http://example.com/codeflow?start=16&length=6)

Can you commend what these modes mean? Especially the difference between Island and StringInterpolation?

src/libraries/Microsoft.PowerFx.Core/Lexer/Tokens/IdentToken.cs
MikeStall: May deserve a const field so this value is shared with ParseStringInterpolation(). That also helps associate the code snippets together to self-document.

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
MikeStall: Agree with 1. Definitely not 2. We can always switch to 3 later - thay may be useful for some IDE features.

src/libraries/Microsoft.PowerFx.Core/Lexer/Tokens/IdentToken.cs
bryancroteau-MSFT: Added a comment.

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
bryancroteau-MSFT: Added a comment.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/string.txt
bryancroteau-MSFT: Added all of these tests.

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -969,11 +978,27 @@ internal static void ChoosePunctuators(ILanguageSettings loc, out string dec, ou
 
         private sealed class LexerImpl
         {
+            /*
```
MikeStall, line 981: >/* [](http://example.com/codeflow?start=12&length=2)

nit: can we use //
We generally avoid multiline.

MikeStall (dismissed):
:shipit:

MikeStall (dismissed):
:shipit:

lesaltzm (approved):
🚢

lesaltzm (changes requested):
Actually, needs to be updated with #78 I think

src/libraries/Microsoft.PowerFx.Core/FeatureFlags.cs
```diff
@@ -7,7 +7,7 @@ namespace Microsoft.PowerFx
     /// Hosts can enable these flags to try out early features. 
     /// Any flags will eventually default to true and then get removed once the feature is finalized. 
     /// </summary>
-    public static class FeatureFlags
+    internal static class FeatureFlags
```
MikeStall, line 10: >internal [](http://example.com/codeflow?start=4&length=8)

+1

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
bryancroteau-MSFT: Fixed

lesaltzm (approved):
👍

PR #59, opened 2021-11-17
Time Zone Offset

LeeParrishMSFT:
Added TimeZoneOffset function and extended the Text function formatting strings to handle more options for date/time formatting and to handle input locale. Also added a new unit test class test test TimeZoneOffset that can't be tested easily in a .txt file input as the result depends on the time zone the test is running within.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
@@ -388,15 +388,15 @@ namespace Microsoft.PowerFx.Functions
             if (args.Length == 0)
             {
                 var tzOffsetMinutes = tzInfo.GetUtcOffset(DateTime.Now).TotalMinutes;
-                return new NumberValue(irContext, tzOffsetMinutes * -1);
+                return new NumberValue(irContext, tzOffsetMinutes / 1440.0 * -1);
             }
 
             switch (args[0])
             {
                 case DateTimeValue dtv:
-                    return new NumberValue(irContext, tzInfo.GetUtcOffset(dtv.Value.ToUniversalTime()).TotalMinutes * -1);
+                    return new NumberValue(irContext, tzInfo.GetUtcOffset(dtv.Value.ToUniversalTime()).TotalMinutes / 1440.0 * -1);
                 case DateValue dv:
-                    return new NumberValue(irContext, tzInfo.GetUtcOffset(dv.Value.ToUniversalTime()).TotalMinutes * -1);
+                    return new NumberValue(irContext, tzInfo.GetUtcOffset(dv.Value.ToUniversalTime()).TotalMinutes / 1440.0 * -1);
                 default:
                     return CommonErrors.InvalidDateTimeError(irContext);
             }
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
@@ -381,5 +381,25 @@ namespace Microsoft.PowerFx.Functions
             else
                 return CommonErrors.InvalidDateTimeError(irContext);
         }
+
+        public static FormulaValue TimeZoneOffset(IRContext irContext, FormulaValue[] args)
+        {
+            TimeZoneInfo tzInfo = TimeZoneInfo.Local;
+            if (args.Length == 0)
+            {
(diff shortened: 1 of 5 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
```diff
@@ -381,5 +381,25 @@ public static FormulaValue TimeParse(EvalVisitor runner, SymbolContext symbolCon
             else
                 return CommonErrors.InvalidDateTimeError(irContext);
         }
+
+        public static FormulaValue TimeZoneOffset(IRContext irContext, FormulaValue[] args)
+        {
+            TimeZoneInfo tzInfo = TimeZoneInfo.Local;
+            if (args.Length == 0)
+            {
+                var tzOffsetMinutes = tzInfo.GetUtcOffset(DateTime.Now).TotalMinutes;
```
bryancroteau-MSFT, line 390: Is this correct? Why are we using minutes? Date/DateTime + Number operator assumes the Number has a unit of Days. Shouldn't we be returning the fraction of a Day, so that Date(...) + TimeZoneOffset() will be meaningful?

bryancroteau-MSFT (changes requested):

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
LeeParrishMSFT: Just following the documentation that states that the value returned from TimeZoneOffset is in Minutes: https://docs.microsoft.com/en-us/powerapps/maker/canvas-apps/functions/function-dateadd-datediff .

bryancroteau-MSFT: Talk to Greg about this. That documentation might be out of date.

LeeParrishMSFT: @gregli-msft Any comment on what direction we need to go on this? The documentation and the .ts implementation are in minutes, but @bryancroteau-MSFT thinks fractions of a day would be the right direction as you can then just use "+" against a Date/DateTime to add/subtract the offset. Documentation shows using DateAdd() and specifying "Minutes" as the unit when working with DateTimeOffset values.

IMHO having the convenience of being able to use the "+" operator at the expense of being able to actually understand the value that is being returned isn't worth it. Trying to explain to someone that the value returned by DateTimeOffset is a fraction of a day, especially to someone who is in a timezone such as UTC -9:30, or UTC +5:45 is going to be a lot more difficult than explaining it's a number of minutes as a whole number and not a fraction.

orcmid: > @LeeParrishMSFT IMHO having the convenience of being able to use the "+" operator at the expense of being able to actually understand the value that is being returned isn't worth it. Trying to explain to someone that the value returned by DateTimeOffset is a fraction of a day, especially to someone who is in a timezone such as UTC -9:30, or UTC +5:45 is going to be a lot more difficult than explaining it's a number of minutes as a whole number and not a fraction.

If the handling of DateTime here is based on the Excel (and OOXML) provisions, dealing with fractions of days seems inescapable.  And since, in this case, minutes are not exact (except with rounding), the question seems to be over when the abstraction leaks and who handles the support calls and/or forum/stack-overflow messages.  

We have the problem enough with non-decimal fractional arithmetic.    Unless you're proposing to *do* decimal arithmetic in these cases, the question seems to be when to bite the bullet.  

Consistency would be good.  On that basis, I would suppose fractions of days needs to be cleanly established, and if something is in minutes, how that is represented needs to be prominent, seems to me.

LeeParrishMSFT: @bryancroteau-MSFT I've updated to now return a fraction of a day as a NumberValue. All tests that were previously passing are passing now as well, once I made a few tweaks to let them understand the new return values.

bryancroteau-MSFT: The number 1440.0 appears in a number of places in this PR. Are you using TimeSpan objects? TimeSpan has a TotalDays field that will return a fractional number.

LeeParrishMSFT: Got it, thanks!

bryancroteau-MSFT (approved):

PR #61, opened 2021-11-17
Add ReferencedIdentifiers to CheckResult, all identifiers used in the…

ToddN:
… expression. ReferencedIdentifiers returns a HashSet<Span> instead of HashSet<string>, because creating all those strings would put a lot of load on the memory management and often code will not use the ReferencedIdentifiers. If people need the strings, it's easy enough to create from the spans, as the unit tests show.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
@@ -28,6 +28,12 @@ namespace Microsoft.PowerFx.Core.Public
         public HashSet<string> TopLevelIdentifiers { get; set; }
 
         /// <summary>
+        /// Spans of all fields that this formula uses. 
+        /// null if unavailable.  
+        /// </summary>
+        public HashSet<Localization.Span> ReferencedIdentifiers { get; set; }
+
+        /// <summary>
         /// null on success, else contains errors. 
         /// </summary>
         public ExpressionError[] Errors { get; set; }
--- a/src/libraries/Microsoft.PowerFx.Interpreter/AllDependencyFinder.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/AllDependencyFinder.cs
@@ -1,0 +1,43 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using Microsoft.PowerFx.Core.Binding;
+using Microsoft.PowerFx.Core.Localization;
+using Microsoft.PowerFx.Core.Syntax.Nodes;
+using Microsoft.PowerFx.Core.Syntax.Visitors;
+
+namespace Microsoft.PowerFx
+{
+    internal class AllDependencyFinder : IdentityTexlVisitor
+    {
+        public HashSet<Span> _spans = new();
+
+        public static HashSet<Span> FindAllDependencies(TexlNode node)
+        {
+            var v = new AllDependencyFinder();
+            node.Accept(v);
+
+            return v._spans;
+        }
+
+        public override void Visit(FirstNameNode node)
+        {
+            if (node.Parent?.Kind == Core.Syntax.NodeKind.DottedName)
+            {
+                TexlNode nodeT = node;
+                while (nodeT.Parent?.Kind == Core.Syntax.NodeKind.DottedName)
+                {
+                    nodeT = nodeT.Parent;
+                }
+                _spans.Add(nodeT.GetCompleteSpan());
+            }
+            else
+            {
+                _spans.Add(node.Token.Span);
+            }
+
+            base.Visit(node);
+        }
+    }
+}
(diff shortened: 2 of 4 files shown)


src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -274,6 +278,95 @@ public static NumberValue Execute(NumberValue x, NumberValue y)
...
+            var referencesIdentifierStrings = ReferencedIdentifersSet(result.ReferencedIdentifiers, expr);
+            Assert.Equal("user.age", referencesIdentifierStrings.First());
+        }
+
+        [Fact]
+        public void ManyVariables()
+        {
+            var engine = new RecalcEngine();
+            var expr = "(27/x * test(user.age) - global.shoeSize) ^ exponent";
+            var result = engine.Check(expr,
```
lesaltzm, line 360: What about something a bit more complex?

`With(user, age+1)`, where this is using the field "age" of the user record. That'll return {"user", "age"} with the current implementation, is that the behavior we want?

ToddN: We're currently requiring fully qualified names, so you'd have to say "user.age" to get the age field of the user record. So I think this matches the intended behavior.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -25,7 +25,13 @@ public class CheckResult
...
         /// This is only valid if there are no errors. 
         /// </summary>
-        public HashSet<string> TopLevelIdentifiers { get; set; }
+        public HashSet<string> TopLevelIdentifiers { get; set; }
+
+        /// <summary>
+        /// Spans of all fields that this formula uses. 
+        /// null if unavailable.  
+        /// </summary>
+        public HashSet<Localization.Span> ReferencedIdentifiers { get; set; }
```
MikeStall, line 34: >Localization.Span [](http://example.com/codeflow?start=23&length=17)

Can we use strings here? 
Strings are simpler to understand and we've already allocated the string object in the symbol table, so for perf concerns, just be sure to return the existing already-allocated object,

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
lesaltzm: That's not exactly how With and record scopes in general. User is a record, so it's fields are usable in lambdas of functions like With, ForAll, Filter, ...

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -25,7 +25,13 @@ public class CheckResult
         /// null if unavailable.  
         /// This is only valid if there are no errors. 
         /// </summary>
-        public HashSet<string> TopLevelIdentifiers { get; set; }
+        public HashSet<string> TopLevelIdentifiers { get; set; }
+
+        /// <summary>
+        /// Spans of all fields that this formula uses.
```
MikeStall, line 31: >Spans of all fields that this formula uses.  [](http://example.com/codeflow?start=12&length=44)

Can we clarify the difference between this and TopLevelIdentifiers?

Ie,  what does this do for:
 A.B.C? 
First(A.B).X.Y

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -169,6 +169,8 @@ private CheckResult CheckInternal(string expressionText, FormulaType parameterTy
                 if (binding.ResultType.Kind != DKind.Enum)
                     result.ReturnType = FormulaType.Build(binding.ResultType);
             }
+            // We find nexted identifiers even if there were errors in the expression.
```
lesaltzm, line 172: nit: nested

src/libraries/Microsoft.PowerFx.Interpreter/AllDependencyFinder.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.PowerFx.Core.Binding;
+using Microsoft.PowerFx.Core.Localization;
+using Microsoft.PowerFx.Core.Syntax.Nodes;
+using Microsoft.PowerFx.Core.Syntax.Visitors;
+
+namespace Microsoft.PowerFx
+{
+    internal class AllDependencyFinder : IdentityTexlVisitor
+    {
+        public HashSet<Span> _spans = new();
```
lesaltzm, line 14: nit: private?

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
lesaltzm: Or, let's say you have an object with two user children:

`If(PickFirst, Foo.User1, Foo.User2).Name`

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
ToddN: For the first line, it returns the span covering "A.B.C".  For the second, it returns the span covering "A.B".  Let's talk (outside the PR review?) about whether that is appropriate or not. I can see arguments on both sides.

PR #68, opened 2021-11-22
Initial work on two table functions

bryancroteau-MSFT:
Draft PR for the changes to support functions that can accept up to 2 tables, with various permutations:

F(Table, Scalar)
F(Scalar, Table)
F(Table, Table)


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ArgCoercion.txt
```diff
@@ -16,11 +16,9 @@
 "123456"
 
 >> Concatenate(-1, ["456", "abc"] )
-/*InterpreterRunner*/ #Skip
 [{Result:"-1456"},{Result:"-1abc"}]
 
 >> Concatenate([-1, 123], ["456", "abc"] )
-/*InterpreterRunner*/ #Skip
 [{Result:"-1456"},{Result:"123abc"}]
```
CarlosFigueiraMSFT, line 23: We should have many more tests for the "zipping" scenarios:
- Scalar+Table
  - Simple case (what we have in line 18)
  - Empty table
  - Null table (e.g., `If(1<0,["txt"])`)
  - Error table (e.g., `If(1/0<2,["txt"])`)
  - Table with error values (e.g., `["hello", Mid("great", -1), "world"]`)
  - Table with error records (e.g., `Table({a:"one"},If(1/0<2,{a:"two"}),{a:"three"})`)
  - Table with null values (e.g., `Table({a:"one"},If(1<0,{a:"two"}),{a:"three"))`)
  - Null scalar + valid table
  - Null scalar + table with errors
- Table+Scalar (not there)
  - Same cases listed above, opposite order
- Table+Table
  - Simple case (what we have in line 21)
  - Error/null tables
  - Tables with error / null values
  - Tables with different number of elements (e.g., `Concatenate(["one", "two"], [1, 2, 3, 4])`)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
```diff
@@ -173,6 +173,99 @@ private static bool IsInvalidDouble(double number)
...
+                KeyValuePair<IEnumerable<DValue<RecordValue>>, IEnumerable<DValue<RecordValue>>> allRows;
+                if (arg0 is TableValue both0 && arg1 is TableValue both1)
+                {
+                    if (both0.Rows.Count() == both1.Rows.Count())
+                    {
+                        allRows = new KeyValuePair<IEnumerable<DValue<RecordValue>>, IEnumerable<DValue<RecordValue>>>(both0.Rows, both1.Rows);
+                    }
+                    else
+                    {
+                        return CommonErrors.UnreachableCodeError(irContext);
```
CarlosFigueiraMSFT, line 193: We can have tables of different lengths being concatenated. The existing behavior is to complete the smaller table with blank / null values:
image

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
```diff
@@ -173,6 +173,99 @@ private static bool IsInvalidDouble(double number)
...
+                var resultType = tableType.ToRecord();
+                var itemType = resultType.GetFieldType(BuiltinFunction.OneColumnTableResultNameStr);
+
+                var zipped = allRows.Key.Zip(allRows.Value, (a, b) => new KeyValuePair<DValue<RecordValue>, DValue<RecordValue>>(a, b));
+                var resultRows = new List<DValue<RecordValue>>();
+                foreach (var pair in zipped)
+                {
+                    if (pair.Key.IsValue && pair.Value.IsValue)
+                    {
+                        var value0 = pair.Key.Value.GetField(BuiltinFunction.ColumnName_ValueStr);
```
CarlosFigueiraMSFT, line 229: The record won't necessarily have a 'Value' property
image

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
bryancroteau-MSFT: This is astonishing! Thanks for pointing this out, this is a major oversight in the current code.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
bryancroteau-MSFT: So what is the rule here? If the record only contains a single field, you just use the single field regardless of the name? If the record has more than 1 field, does it search for the "Value" field?

CarlosFigueiraMSFT: It needs to be a single-column table; if there is more than one column, the compilation will fail - see https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Concatenate.cs#L60

bryancroteau-MSFT: Added support for other single column names for the input table(s).

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
bryancroteau-MSFT: Added support for variable length input tables.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ArgCoercion.txt
bryancroteau-MSFT: Ok, I added all those tests and they are passing!

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
 using System.Diagnostics.Contracts;
 using System.Dynamic;
+using System.Linq;
```
CarlosFigueiraMSFT, line 8: Needed? Nothing else seems to change in this file.

src/libraries/Microsoft.PowerFx.Core/Public/Types/TableType.cs
```diff
@@ -35,9 +36,12 @@ public TableType Add(NamedFormulaType field)
             return new TableType(newType);
         }
 
+        public string SingleColumnFieldName => GetNames().First().Name;
+        public FormulaType SingleColumnFieldType => GetNames().First().Type;
+
```
CarlosFigueiraMSFT, line 41: We may want to assert (or throw an exception) if this is not a single-column table

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
```diff
@@ -173,6 +173,116 @@ private static bool IsInvalidDouble(double number)
...
+                var inputRecordType = new RecordType().Add(name, arg.Type);
+                var inputRecordNamedValue = new NamedValue(name, arg);
+                var inputRecord = new InMemoryRecordValue(IRContext.NotInSource(inputRecordType), new List<NamedValue>() { inputRecordNamedValue });
+                var inputDValue = DValue<RecordValue>.Of(inputRecord);
+                var rows = Enumerable.Repeat(inputDValue, size);
+                return new ExpandToSizeResult(name, rows);
+            }
+        }
+
+        private static KeyValuePair<List<string>, List<List<DValue<RecordValue>>>> Transpose(IEnumerable<ExpandToSizeResult> results, int size)
```
CarlosFigueiraMSFT, line 240: Can you add a comment that says what this function does? It's not clear by the name

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
```diff
@@ -173,6 +173,116 @@ private static bool IsInvalidDouble(double number)
...
+
+            for (int i = 0; i < size; i++)
+            {
+                second.Add(lists.Select(list => list[i]).ToList());
+            }
+
+            return new KeyValuePair<List<string>, List<List<DValue<RecordValue>>>>(first, second);
+        }
+
+        public static Func<EvalVisitor, SymbolContext, IRContext, FormulaValue[], FormulaValue> DoubleSingleColumnTable(FunctionPtr targetFunction)
```
CarlosFigueiraMSFT, line 255: What does a "DoubleSingleColumnTable" do? It's not clear by the name. Comments would help

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
```diff
@@ -263,6 +373,18 @@ private static FormulaValue ReplaceBlankWithEmptyString(IRContext irContext, int
             }
         }
 
+        private static FormulaValue ExactValueTypeOrTable<T>(IRContext irContext, int index, FormulaValue arg) where T : FormulaValue
```
CarlosFigueiraMSFT, line 376: `ExactValueTyoeOrTableOrBlank`? Blanks are also allowed in this method

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/string.txt
```diff
@@ -291,4 +291,88 @@ true
...
+[]
+
+>> Concatenate(If(1<0, ["txt"]), "a")
+[]
+
+>> Concatenate(If(1/0<2, ["txt"]), "a")
+#Error
+
+>> Concatenate(["hello", Mid("great", -1), "world"], "a")
+[{Result:"helloa"},{Result:Microsoft.PowerFx.Core.Public.Values.ErrorValue},{Result:"worlda"}]
```
CarlosFigueiraMSFT, line 341: We need something better than `Microsoft.PowerFx.Core.Public.Values.ErrorValue` in the expected result; not a blocker for this PR, but we should really get to it soon

src/libraries/Microsoft.PowerFx.Core/Public/Types/TableType.cs
bryancroteau-MSFT: Done

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/string.txt
bryancroteau-MSFT: Yes, we will implement the ability to specify specific error count and type in a future change.

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
bryancroteau-MSFT: Removed

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
bryancroteau-MSFT: The function signature has been simplified a lot. Do we still need a comment?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
bryancroteau-MSFT: This has been renamed as well.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
bryancroteau-MSFT: Done

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/string.txt
```diff
@@ -291,4 +291,88 @@ true
...
+
+>> Concatenate(Table({a:"one"},If(1<0,{a:"two"}),{a:"three"}), "a")
+[{Result:"onea"},{Result:"a"},{Result:"threea"}]
+
+>> Concatenate(Table({a:"one"},If(1/0<2,{a:"two"}),{a:"three"}), "a")
+Microsoft.PowerFx.Core.Public.Values.ErrorValue
```
CarlosFigueiraMSFT, line 347: This is not correct - it should return a table with three elements: "onea", <error>, "threea"

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
bryancroteau-MSFT: Done

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
bryancroteau-MSFT: Added comment

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/string.txt
bryancroteau-MSFT: This is correct, all functions return an error if any argument is an error.

CarlosFigueiraMSFT: The Table function should not return an error in this case. Not a blocker for this PR, but it should be addressed soon.

CarlosFigueiraMSFT (approved):

PR #74, opened 2021-12-02
New property rule category - "formulas"

emhomm4:
This PR creates a new PropertyRuleCategory for named formulas, called "Formulas".


src/libraries/Microsoft.PowerFx.Core/App/Controls/PropertyRuleCategory.cs
```diff
@@ -21,12 +21,13 @@ internal enum PropertyRuleCategory
...
         OnDemandData = 4,
         Scope = 5,
+        Formulas = 6,
         /// <summary>
         /// Represents a missing property category when deserializing
         /// Should be cleaned up by document converter, only occurs if 
         /// control template is invalid
         /// </summary>
-        Unknown = 6,
+        Unknown = 7,
```
lesaltzm, line 30: why are we renumbering this enum?

MikeStall: IE, should `Unknown` be the last item?  Or should Formulas be added as 7?

lesaltzm: Unknown is a valid value, it's not a catch-all, and it's serialized by PowerApps-Client. It should stay 6, the new one should be added as 7

lesaltzm (approved):
🚢

PR #75, opened 2021-12-02
Update Switch function type comparison to work with enum types

CarlosFigueiraMSFT:
Currently the Switch function doesn't work well if the value to be switched on is of an enum type. This means that expressions such as the ones below are considered invalid today:

    IfError(numericVar, Switch(FirstError.Kind, ErrorKind.Div0, -1, ErrorKind.Numeric, -2, -3))
    Switch(TextInput1.DisplayMode, DisplayMode.Edit, "editing", DisplayMode.View, "viewing", "other")

This changes the type comparison logic in that function to use the same as the one used by the equality operator.


src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1725,7 +1725,7 @@ public virtual bool Accepts(DType type, out KeyValuePair<string, DType> schemaDi
                     break;
 
                 case DKind.Enum:
-                    accepts = (Kind != type.Kind && type.Kind == DKind.Unknown) ||
+                    accepts = (Kind != type.Kind && (type.Kind == DKind.Unknown || EnumSuperkind == type.Kind)) ||
```
bryancroteau-MSFT, line 1728: The line below already covers the case where we want two enums to be compatible if the EnumSuperKind is the same for both. If the line below is failing, should we instead set the EnumSuperKind of an enum type to itself?

bryancroteau-MSFT (changes requested):

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
CarlosFigueiraMSFT: In the example listed in the PR description (`Switch(TextInput1.DisplayMode, DisplayMode.Edit, "editing", DisplayMode.View, "viewing", "other")`), when we use `DisplayMode.Edit` or `DisplayMode.View` we don't get an enum value; instead, we get the actual (string, in this case) value of the enum. So the comparison is "does DisplayMode enum accept a string"? Since "string" doesn't have an EnumSuperKind, the line below will return false.

bryancroteau-MSFT: Then why does your change fix the problem?

CarlosFigueiraMSFT: In that example (line 1728), [this.]Kind == DKind.Enum, type.Kind == DKind.String, and [this.]EnumSuperkind == DKind.String.

src/tests/Microsoft.PowerFx.Interpreter.Tests/FunctionCompilationTests.cs
```diff
@@ -0,0 +1,26 @@
...
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Interpreter;
+using Xunit;
+
+namespace Microsoft.PowerFx.Tests
+{
+    public class FunctionCompilationTests
+    {
+        [Theory]
+        [InlineData("Switch(A, 2, \"two\", \"other\")")]
```
MikeStall, line 14: .txt tests instead?

CarlosFigueiraMSFT: Cannot do it now - IfError doesn't expose the 'FirstError' scope variable yet

bryancroteau-MSFT: Indeed, that will be added in a future PR

bryancroteau-MSFT (dismissed):
Have Wyatt look at it as well

bryancroteau-MSFT (approved):

PR #76, opened 2021-12-03
Fix IfError return type checking

CarlosFigueiraMSFT:
If the IfError function is called with more than two arguments, some of them will never be returned by the function. Therefore, they don't need to be compatible with the possible argument types that are returned.

For example, in the expression `IfError(v1, f1, v2, f2, v3, f3)`, the possible return values are f1, f2, v3 and f3, so it's ok if the type of v1 and v2 are not compatible with the others.


bryancroteau-MSFT (approved):

MikeStall (approved):
:shipit:

PR #77, opened 2021-12-03
Add internal infrastructure to find state dependencies before Binding

babrekel:
## Problem statement and purpose
A system we are developing requires understanding of the external state that is leveraged by a particular expression, which could potentially trigger reordering of type context building operations. For example, pseudocode:
  
  `X` = `Y + Z`
  `Y` = `5`
  `Z` = `Y + 2`
  
The desired result from the above is knowing that `X`, `Y` and `Z` are numbers. Related: if there is a scenario `X = Y`, `Y = Z`, `Z = X`, we would like to understand that a circular dependency exists, and present an error to the user. Additionally, we would like the ability to show some visual cues or hints about the variables that are leveraged from expressions, which may not be obvious if the expressions are very large or complex. Variables are always multipart identifiers.
  
The Power FX engine is not aware of the order among these expressions or the assignments. In order to effectively leverage the public API RecalcEngine surface area, evaluating the contextual data needs to be done in dependency order. 
  
## Solution
  
This pull request includes a visitor that operates over the `Formula` syntax tree directly without requiring a `FormulaType` input, with a set of unit test cases that cover all the code in the visitor. The output is a set of `DPath` structures representing potentially accessed state. 
  
_If we wish to do so in the future, we could consider leveraging this infrastructure to extend RecalcEngine to support checking a list of expressions with associated assignments._
  
## Caveats
  
The output is potentially ambiguous. For example:
  
`With(User, Age + 1)` could either mean: `User.Age` or `Age` (top-level). 
  
This cannot be resolved until the binder runs. However, the information here is sufficient to establish dependency order and identify which known variables are referenced by an expression, by intersecting with known variable assignments maintained outside the scope of Power FX. 
  
A significant set of capabilities are not supported:
  
  - No data structure has been created to differentiate between ambiguous identifiers and known identifiers. For example, `User.Age + Age` returns the same result as `With(User, Age + 1)`. But in the first example, it is clear that both are required. In the second example, one of the two would be acceptable. We may need to fix this before merging, I would like some input as to how to structure this. 
  - Table operations such as `ForAll` and `Sum` return inline values as top-level identifiers, without processing the fact that these values may be properties of the table being enumerated. Example: evaluating `ForAll(Foo, { B: X })` does not convey that X could be a column of the `Foo` table. 
  - `Self`, `Parent` and replacement accessors are ignored by the processor. 
  - Only `If` and `With` are processed using special semantics. This may need to be expanded to other known cases. I would also love some input here to understand what is appropriate.


bryancroteau-MSFT:
We should not be looking for circular dependencies between identifiers. We should be only looking for dependencies between symbols in the symbol tables. We can generate (processFirst, processSecond) pairs during type checking. Then we can topological sort the pairs to find circular dependencies.

MikeStall:
> #nullable enable

Nit: add header:

// Copyright (c) Microsoft Corporation.
// Licensed under the MIT License.

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Syntax/Visitors/FindDependenciesVisitor.cs:1 in ea4d506. [](commit_id = ea4d506266f4fbaddb0704f16a1870e3f9c6a052, deletion_comment = False)

MikeStall (approved):
:shipit:

src/tests/Microsoft.PowerFx.Core.Tests/DependencyFinderTests.cs
```diff
@@ -0,0 +1,92 @@
...
+        [InlineData("Switch(A, B)", "")] // Error
+        [InlineData("Switch(Slider1.Value, 20, Result1, 10, Result2, 0, Result3)", "Result1,Result2,Result3,Slider1.Value")]
+        [InlineData("Switch(Slider1.Value, 20, Result1, 10, Result2, 0, Result3).Name", "Result1.Name,Result2.Name,Result3.Name,Slider1.Value")]
+        [InlineData("Switch(Slider1.Value, 20, Result1, 10, Result2, 0, Result3, DefaultResult)", "Result1,Result2,Result3,DefaultResult,Slider1.Value")]
+        [InlineData("Switch(Slider1.Value, 20, Result1, 10, Result2, 0, Result3, DefaultResult).Name", "Result1.Name,Result2.Name,Result3.Name,DefaultResult.Name,Slider1.Value")]
+        [InlineData("With(A).Name", "")] // error
+        [InlineData("With({ B: 1, C: B.Inner } As Y, Y.C)", "B.Inner")]
+        [InlineData("With(A, B).Name", "A.B.Name,B.Name")]
+        [InlineData("With(A As B, B).Name", "A.Name")]
+        [InlineData("With(User, Age + 1)", "User.Age,Age")]
```
bryancroteau-MSFT, line 48: Returning multiple paths isn't enough. You need a way of eventually converging on a coherent worldview.

For example, given a (potentially empty) list of input symbols with known types, you get back one set of dependencies for a given input formula. Some of the output symbols might be ambiguous or inferred symbols.

Given a list of formulas and a working set of input symbols (initially empty), in a loop, you:

1. Iterate through the list of formulas and find dependencies using the current working set as input/context
2. If there are no ambiguities and all types are known, you add the result type to the working set
3. You sort the failures by their degree of ambiguities, then go back to step 1

This process continues until no progress is made. If no progress is made in an iteration, then you return an error that the inference is intractable. Note the performance of such an algorithm might be very slow for large formula sets.

MikeStall (approved):

PR #78, opened 2021-12-03
Add feature flag concept.

MikeStall:


src/libraries/Microsoft.PowerFx.Core/FeatureFlags.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.PowerFx
+{
+    /// <summary>
+    /// Hosts can enable these flags to try out early features. 
+    /// Any flags will eventually default to true and then get removed once the feature is finalized. 
+    /// </summary>
+    public static class FeatureFlags
```
bryancroteau-MSFT, line 10: Should we use the [Flags] annotation?

src/libraries/Microsoft.PowerFx.Core/FeatureFlags.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        /// <summary>
+        /// Enable String Interpolation feature. 
+        /// Added 12/3/2021
+        /// </summary>
+        public static bool StringInterpolation = false;
```
bryancroteau-MSFT, line 16: Should these flags be 1, 2, 4, 8, etc instead of just booleans?

bryancroteau-MSFT (commented):
Why do we have a static class? Static classes are not thread safe. It people want to run the interpreter in multiple threads with different options, a static class would prevent that.

MikeStall:
Acknolwedged, but  
1. These flags should all be temporary and short lived.  Ie, a few weeks between when it's checked into Main and when we finish a test pass and commit.  So I wanted something very tactical. 
2. If it's an instance class, then we have to plumb it through the code (extra ctor parameters on parser, etc). 
3. Supporting multiple instances with different feature flags is a non-scenario.  If we really want that, then it should get promoted to a first class concept (like a real param passed into the various ctors, etc). 
4. Removing a flag should be easy - just replace all usages of the flag with 'true'. If we plumb thorugh as an instance, that's a larger footprint to touchup. 

---
In reply to: [823183807](https://github.com/microsoft/Power-Fx/pull/78#issuecomment-823183807) [](http://example.com/codeflow?ancestors=823183807,823183807)

src/libraries/Microsoft.PowerFx.Core/FeatureFlags.cs
MikeStall: Boolean is easier to check. We don't need to optimize storage with an enum.

Shpakh-MSFT (approved):
:shipit:

Shpakh-MSFT (approved):
:shipit:

PR #82, opened 2021-12-07
Example implementation of configurable Enums

bryancroteau-MSFT:
I am looking for feedback on this general direction and approach to customizable Enums.

- Make EnumStore an instance class instead of a static class
- Make a virtual property on EnumStore that allows users to override the return value
- Replace _enum with the new virtual property in EnumStore
- Remove the coupling between EnumStore and Error type schemas
- Pipe EnumStore instances from RecalcEngine all the way to IntellisenseData through constructor parameters

Note that this approach might cause a large number of compile errors when building Power-Apps Client.

Possible topics of discussion: Glue, INameResolver, TexlBinding, Localized Enums, Colors


src/libraries/Microsoft.PowerFx.Core/Types/Enums/EnumStore.cs
```diff
@@ -184,7 +192,7 @@ public static void RegisterTuple(Tuple<string, string, string> tuple, Dictionary
             }
         }
 
-        public static bool TryGetLocalizedEnumValue(string enumName, string enumValue, out string locValue)
+        public bool TryGetLocalizedEnumValue(string enumName, string enumValue, out string locValue)
```
lesaltzm, line 195: Thinking a bunch of these should probably be `internal` now that we're exposing this. Subclasses should be able to extend the dictionary but not modify the lookup functions, imo

bryancroteau-MSFT: Marked the functions as internal instead of public.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -32,15 +33,15 @@ public class RecalcEngine : IScope
 
         internal Dictionary<string, RecalcFormulaInfo> Formulas { get; } = new Dictionary<string, RecalcFormulaInfo>();
 
-        private readonly CultureInfo _cultureInfo;
+        PowerFxConfig _powerFxConfig;
```
MikeStall, line 36: >PowerFxConfig [](http://example.com/codeflow?start=8&length=13)

Design discussion: what should go on Config vs. on this class directly?
Ie, EnumStore can't be static anymore and needs to live somewhere.  We'll have a similar issue coming with BuiltinFunctions. We have a custom function lists here (line 32).

bryancroteau-MSFT: The extra functions can be added to the Config instead, either in this PR or in a future PR (when built-in functions are also made non-static)

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseProvider.cs
```diff
@@ -21,9 +23,9 @@ internal static class IntellisenseProvider
             new Intellisense.RecordNodeSuggestionHandler(),
         };
 
-        internal static IIntellisense GetIntellisense()
+        internal static IIntellisense GetIntellisense(PowerFxConfig powerFxConfig)
```
MikeStall, line 26: >GetIntellisense [](http://example.com/codeflow?start=38&length=15)

Something here feels like we should have symmetry with how we specify which functions are available.

bryancroteau-MSFT: We will. In the next PR I will do the same thing for built-in functions as I did for Enums. Make them non-static.

src/libraries/Microsoft.PowerFx.Core/Types/Enums/EnumStore.cs
```diff
@@ -12,19 +12,19 @@ namespace Microsoft.PowerFx.Core.Types.Enums
     /// <summary>
     /// Static class used to store built in Power Fx enums
     /// </summary>
-    internal static class EnumStore
+    public class EnumStore
```
MikeStall, line 15: >public [](http://example.com/codeflow?start=4&length=6)

Next step - How would hosts customize this in a reasonable way?

A lot of these look like control enums that don't belong outside of Canvas apps. 

In some cases, adding a function may want to pull in enums (ie, Date functions that consume DateTimeFormatEnumString enum may want to pull that in).

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -0,0 +1,35 @@
...
+using Microsoft.PowerFx.Core.Types.Enums;
+
+namespace Microsoft.PowerFx.Core.Public.Config
+{
+    /// <summary>
+    /// A container that allows for compiler customization.
+    /// </summary>
+    public sealed class PowerFxConfig
+    {
+        private readonly EnumStore _enumStore;
```
MikeStall, line 17: >EnumStore [](http://example.com/codeflow?start=25&length=9)

Another source of "intellisense noise" is colors: 
\Power-Fx\src\libraries\Microsoft.PowerFx.Core\Types\Enums\ColorTable.cs

Not sure what to do there. Maybe those are common enough we always include. Maybe hosts need more control.

src/libraries/Microsoft.PowerFx.Core/Types/Enums/EnumStore.cs
bryancroteau-MSFT: You can customize the enums by overriding the EnumDict property in a derived class.

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
lesaltzm: Most of the contents of the enum store here should be removed and moved back to PA-Client/other hosts as needed

bryancroteau-MSFT: I will be removing static data in future PRs. I am going to be splitting them up into smaller changes to make it easier to integrate with CanvasApps.

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Text;
+using Microsoft.PowerFx.Core.Types.Enums;
+
+namespace Microsoft.PowerFx.Core.Public.Config
```
MikeStall, line 10: >Microsoft.PowerFx.Core.Public.Config [](http://example.com/codeflow?start=10&length=36)

nit: For public namespaces, can we stick with just "Microsoft.PowerFx". 


(@lesaltzm ? )

bryancroteau-MSFT: All the types are located in Microsoft.PowerFx.Core.Public.Types, Values are in the Value namespace.

lesaltzm: Mike has a good point though, our file organization shouldn't be part of the public contract, I got that wrong when doing the refactor. I'd be ok if we moved Types/Values to just be Microsoft.PowerFx.Core (or .Types/.Values if we'd prefer putting more meaning into the namespace)

bryancroteau-MSFT: So do we want to rename all the namespaces in the Public folder in this PR?

lesaltzm: In a subsequent PR, seems outside the scope of this one. For this one, I'd just add this class as Microsoft.PowerFx.Core

bryancroteau-MSFT: Changed the namespace

src/libraries/Microsoft.PowerFx.Core/Types/Enums/EnumStore.cs
MikeStall: can you add a test case that demonstrates the customization? Just to show a e2e.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -32,15 +31,15 @@ public class RecalcEngine : IScope
...
```
MikeStall, line 34: >PowerFxConfig [](http://example.com/codeflow?start=8&length=13)

nit: private readonly

src/tests/Microsoft.PowerFx.Core.Tests/IntellisenseTests/SuggestTest.cs
```diff
@@ -22,14 +26,23 @@ public class SuggestTests : IntellisenseTestBase
...
-            var intellisense = Suggest(expression, contextTypeString);
+            var intellisense = Suggest(expression, powerFxConfig, contextTypeString);
             return intellisense.Suggestions.Select(suggestion => suggestion.DisplayText.Text).ToArray();
         }
 
+        private class EmptyEnumStore : EnumStore
+        {
+            private ImmutableDictionary<string, string> _enumDict = ImmutableDictionary<string, string>.Empty;
+            protected override ImmutableDictionary<string, string> EnumDict => _enumDict;
+        }
```
MikeStall, line 41: We should discuss this pattern more before making it public. 
If you keep EnumStore as internal, I'm good with the change.

MikeStall (approved):
:shipit:

PR #83, opened 2021-12-09
Implement IsNumeric function and add unit tests.

LeeParrishMSFT:
Implemented the IsNumeric function and added unit tests in math.txt file. Followed pattern used in IsBlank to also expose a public static function that returns a bool directly rather than a FormulaValue as a BooleanValue.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
@@ -996,14 +996,8 @@ namespace Microsoft.PowerFx.Functions
 
         public static FormulaValue IsNumeric(EvalVisitor runner, SymbolContext symbolContext, IRContext irContext, FormulaValue[] args)
         {
-            // Blank or empty. 
             var arg0 = args[0];
             return new BooleanValue(irContext, arg0 is NumberValue);
-        }
-
-        public static bool IsNumeric(FormulaValue arg)
-        {
-            return arg is NumberValue;
         }
 
         public static FormulaValue With(EvalVisitor runner, SymbolContext symbolContext, IRContext irContext, FormulaValue[] args)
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -983,6 +994,18 @@ public static bool IsBlank(FormulaValue arg)
...
         }
 
+        public static FormulaValue IsNumeric(EvalVisitor runner, SymbolContext symbolContext, IRContext irContext, FormulaValue[] args)
+        {
+            // Blank or empty. 
+            var arg0 = args[0];
+            return new BooleanValue(irContext, arg0 is NumberValue);
+        }
+
+        public static bool IsNumeric(FormulaValue arg)
```
lesaltzm, line 1004: The pattern IsBlank used was because the logic for IsBlank is a little more complex than `a is BlankValue`. Not sure if this IsNumeric function adds any value, especially since it's not actually called from the FormulaValue IsNumeric function.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/mathfuncs.txt
```diff
@@ -260,4 +260,22 @@ Blank()
...
+true
+
+>>IsNumeric(-123.1)
+true
+
+>>IsNumeric(123.1)
+true
+
+>>IsNumeric("1" + 2)
+true
```
lesaltzm, line 281: Also, `IsNumeric(1/0)` and `IsNumeric(Blank())` and maybe `IsNumeric({RecordValue: 123))`, `IsNu meric([1,2,3,4])`

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
LeeParrishMSFT: Agree, it's gone.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/mathfuncs.txt
LeeParrishMSFT: I've added those as well!

lesaltzm (approved):
🚢

bryancroteau-MSFT (approved):

PR #84, opened 2021-12-09
Language server expression type support

rsullivan00:
## Purpose

When a user modifies or adds a PowerFx expression in our system, we want to show them a preview of the return type of the expression they are authoring. This is important for us to communicate type errors between multiple components that each leverage their own PowerFx expressions.

The language server already supports type checking an expression on load and on change, but only errors (diagnostics) and suggestions are currently communicated through the LSP.

## Solution

I've added a new custom protocol event, `$/publishExpressionType`, which returns type information for an expression provided to the Language Server in the onLoad and onChange handling. This event serializes the check result's `FormulaType` for the client to display the type and shape of the expression result. The serialized shape is:

```json
// Primitive types only have type name
{ "$type": "Number" }
// Aggregate types have type name and nested structures
{ "$type": "Record", "MyBooleanKey": { "$type": "Boolean" }, "MyNumberKey": { "$type": "Number" }}
{ "$type": "Table", "Value": { "$type": "Number" }}
```

This feature is enabled with a `&getExpressionType=true` param in the document URI, so it should have no effect on existing implementations.

I've created a demo integration with the PowerFx demo app ([diff here](https://github.com/microsoft/power-fx-host-samples/compare/main...rsullivan00:ricksullivan/demo-formula-types?expand=1)) that adds expression types to the interface. Ideally this would be integrated with the `PowerFxFormulaEditor` directly using `lspConfig`, but that doesn't appear to be open sourced.

image


src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/LanguageServerTests.cs
```diff
@@ -647,6 +649,99 @@ public void TestPublishTokens()
...
+            Assert.Equal(2, _sendToClientData.Count);
+            var response = JsonSerializer.Deserialize<JsonRpcPublishExpressionTypeNotification>(_sendToClientData[1], _jsonSerializerOptions);
+            Assert.Equal("$/publishExpressionType", response.Method);
+            Assert.Equal(documentUri, response.Params.Uri);
+            Assert.Null(response.Params.Type);
+        }
+
+        [Theory]
+        [InlineData("{}", "{ A: 1 }", "{\"$type\":\"Record\",\"A\":{\"$type\":\"Number\"}}")]
+        [InlineData("{}", "[1, 2]", "{\"$type\":\"Table\",\"Value\":{\"$type\":\"Number\"}}")]
```
lesaltzm, line 717: What about `{'$type': 123}`?

lesaltzm: also, nested records, `{A: 1, B: { C: { D: "Qwerty" }, E: true } } `

ToddN (dismissed):

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/LanguageServerTests.cs
rsullivan00: > also, nested records, {A: 1, B: { C: { D: "Qwerty" }, E: true } }

Nested records serialize the whole shape. Your example would be
```json
"{\"$type\":\"Record\",\"A\":{\"$type\":\"Number\"},\"B\":{\"$type\":\"Record\",\"C\":{\"$type\":\"Record\",\"D\":{\"$type\":\"String\"}},\"E\":{\"$type\":\"Boolean\"}}}"
```
I can add another test for nested records.

rsullivan00: > What about {'$type': 123}?

This is currently undefined and produces duplicate `$type` keys in serialization, and that string then fails to deserialize. I considered adding another layer of nesting to avoid these conflicts, but I prefer the flattened interface.

I can update this to exclude `$type` from being serialized from the expression result, if you agree. Otherwise I can nest the type shape as well, something like

```json
{
  "type": "Record",
  "names": { "type": { /* Type from expression's `type` }}
}
```

lesaltzm: I'd lean towards nested, since `{'$type': 123}` is a valid expression. Especially if we ever want to add more keys to this, it'll be easier if we don't have to worry about name conflicts.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -475,6 +479,24 @@ private void PublishTokens(string documentUri, CheckResult result)
                 }));
         }
 
+        private void PublishExpressionType(string documentUri, CheckResult result)
```
MikeStall, line 482: >PublishExpressionType [](http://example.com/codeflow?start=21&length=21)

This is the key part of the change?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -475,6 +479,24 @@ private void PublishTokens(string documentUri, CheckResult result)
...
+        {
+            var uri = new Uri(documentUri);
+            var nameValueCollection = HttpUtility.ParseQueryString(uri.Query);
+            if (!bool.TryParse(nameValueCollection.Get("getExpressionType"), out var enabled) || !enabled)
+            {
+                return;
+            }
```
MikeStall, line 489: Do we need this check? This is only in response to a specific message request, so could "old" clients ever hit this?

MikeStall, line 489: This notification is fired on every change notification (along with other error notifications)? Hence the need for a switch (to avoid bombarding old clients).

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/JsonRpc/FormulaTypeJsonConverter.cs
```diff
@@ -0,0 +1,139 @@
+﻿
+using System;
+using System.Collections.Generic;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Microsoft.PowerFx.Core.Public.Types;
+
+namespace Microsoft.PowerFx.LanguageServerProtocol
+{
+    public class FormulaTypeJsonConverter : JsonConverter<FormulaType>
```
MikeStall, line 10: >FormulaTypeJsonConverter : JsonConverter<FormulaType> [](http://example.com/codeflow?start=17&length=53)

This is DType -->Json ... 
@lesaltzm - do we have similar code already to do this that we can leverage?

MikeStall:
We could probably also follow a similar pattern when we want to provide "Watch/Eval" functionality to help evaluate intermediate expressions.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/JsonRpc/FormulaTypeJsonConverter.cs
MikeStall: we need to be really clear on versioning expectations here. 
This is probably something that should live in Fx.Core.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
rsullivan00: Yep!

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
rsullivan00: Exactly my thinking. Was this your response to your own question above?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/JsonRpc/FormulaTypeJsonConverter.cs
lesaltzm: Right, yeah. I'm thinking maybe the core of this should be a virtual ToJson function on FormulaType, so it stays up to date if new formula types are added

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/JsonRpc/FormulaTypeJsonConverter.cs
```diff
@@ -0,0 +1,171 @@
...
+using System.Text.Json.Serialization;
+using Microsoft.PowerFx.Core.Public.Types;
+using Microsoft.PowerFx.Core.Utils;
+
+namespace Microsoft.PowerFx.LanguageServerProtocol
+{
+    public class FormulaTypeJsonConverter : JsonConverter<FormulaType>
+    {
+        private const string TypeProperty = "Type",
+                             NamesProperty = "Names";
```
lesaltzm, line 14: small nit: "Fields" might be a better property name here,

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/JsonRpc/FormulaTypeJsonConverter.cs
```diff
@@ -0,0 +1,171 @@
+﻿
+using System;
```
Shpakh-MSFT, line 2: >using [](http://example.com/codeflow?start=0&length=5)

Please add a header for this file.

MikeStall (dismissed):
:shipit:

MikeStall (approved):

lesaltzm (approved):
🚢

PR #85, opened 2021-12-09
Fix a bug in string interpolation where two open curly braces appear …

bryancroteau-MSFT:
…as the first characters after changing modes


CarlosFigueiraMSFT (approved):
🚢

PR #87, opened 2021-12-14
Added Rand and RandBetween functions, and new test class to test these.

LeeParrishMSFT:
Implemented Rand and RandBetween functions. Added a new unit test for these as it's difficult to test this type of function with the .txt file approach as the result can't be determined ahead of time.


src/tests/Microsoft.PowerFx.Interpreter.Tests/MiscInterpreterTests.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using Microsoft.PowerFx.Core.Tests;
+using System.Threading.Tasks;
+using Microsoft.PowerFx.Core.Public.Values;
+using Xunit;
+namespace Microsoft.PowerFx.Interpreter.Tests
+{
+    public class MiscInterpreterTests
```
MikeStall, line 10: >MiscInterpreterTests [](http://example.com/codeflow?start=17&length=20)

Can you add a .txt test case here too (or even instead of the C# ones)? 

[https://github.com/microsoft/Power-Fx/tree/main/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases](https://github.com/microsoft/Power-Fx/tree/main/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases)

They are portable and easy to  write (and hence hit lots of corner cases) :)

RandBetween(5,1) // error case, wrong 
RandBetween(100.5,100.7) ? non integer values

You could also do looping with ForAll+Sequence:

Sum(ForAll(Sequence(10), RandBetween(1,5)), Value)

LeeParrishMSFT: Sure, no problem, I understand the first test in that it's looking for an error case. For the second one because in PowerFx a NumberValue is not an integer, it will evaluate that as the same as RandomBetween(100,100) and return 100. I'm not sure how PowerFx deals with evaluating parameters to be integers if the NumberValue passed in is based on a double, unless the RandomBeween functions needs to do manual validation in the function itself, though I didn't see this anywhere else though I may have just missed it. I can do this, just wasn't sure the intent in PowerFx where parameters are provided as NumberValue and they need to be integers.

I'm also not sure how to evaluate the result of something like Sum(ForAll(Sequence(10), RandBetween(1,5)), Value) in a .txt file test case as the result will be different based on the sum of the random values. I'm also going to leave the C# tests as they are checking the positive result or at least a set of values to see if they are in an expected range which I'm not sure how you can do with a .txt file based test if you can't put a function into the expected result that can evaluate something other than exact result values.

lesaltzm: You could probably do something like 

```
>> Sum(ForAll(Sequence(10), With({RandomNumber: RandBetween(1,20)}, If(RandomNumber >1 && RandomNumber < 20, 1, 0))), Value)
10
```

LeeParrishMSFT: @lesaltzm That makes sense, evaluate each iteration and just count that we got to the final expected count.

lesaltzm: Actually, this can be simpler
```
>> Sum(Sequence(100), With({RandomNumber: RandBetween(1,20)}, If(RandomNumber >1 && RandomNumber < 20, 1, 0))))
100

Comment (LeeParrishMSFT) [reply src/tests/Microsoft.PowerFx.Interpreter.Tests/MiscInterpreterTests.cs]: With that assistance I was able to re-write all the tests I had in C# to be expressions that can run within a .txt file and they will test as much as the C# tests did before. So I'll remove the coded C# tests.

File: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/mathfuncs.txt [hunk 2]
```diff
@@ -290,4 +290,22 @@ false
...
 >>IsNumeric([1,2,3,4])
-false
\ No newline at end of file
+false
+
+>>RandBetween(1, 1)
+1
+
+>>RandBetween(5, 1)
+'minValue' cannot be greater than maxValue. (Parameter 'minValue')
```
lesaltzm, line 299: Instead of throwing, we should check this in RandBetween's impl and return an ErrorValue

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/mathfuncs.txt
```diff
@@ -290,4 +290,22 @@ false
 false
 
 >>IsNumeric([1,2,3,4])
-false
\ No newline at end of file
+false
+
+>>RandBetween(1, 1)
```
lesaltzm, line 295: also, tests for Blank() and 1/0

LeeParrishMSFT: Added additional tests.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -359,5 +364,40 @@ public static FormulaValue Power(IRContext irContext, NumberValue[] args)
...
+        {
+            var lower = (int)args[0].Value;
+            var upper = (int)args[1].Value;
+            lock (_randomizerLock)
+            {
+                if (_random == null)
+                {
+                    _random = new Random();
+                }
+                if (lower > upper)
```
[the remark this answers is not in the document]
LeeParrishMSFT: Done, and good catch, it's not a good practice to spend any more than needed inside a lock.

MikeStall (approved):
:shipit:

PR #88, opened 2021-12-14
Display Name support for Record/Table parameters

lesaltzm:
Adds support for Display Name/Logical Name mappings in the Record Type used to provide expression parameters, as well as adding two functions to convert between display and logical representations.


src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -127,10 +128,14 @@ public static IEnumerable<DType> GetPrimitiveTypes()
...
         protected readonly DType _attachmentType;
 
-        internal HashSet<IExternalTabularDataSource> AssociatedDataSources { get; private set; }
+        internal HashSet<IExternalTabularDataSource> AssociatedDataSources { get; }
         internal IExternalOptionSet<int> OptionSetInfo { get; }
         internal IExternalViewInfo ViewInfo { get; }
 
+        // Eventually, all display names should come from this centralized source
+        // We should not be using individual DataSource/OptionSet/View references
+        internal HashSet<BidirectionalDictionary<string, string>> DisplayNameProviders { get; }
```
wygoodin, line 137: I think this might be worth abstracting to a little interface, so that lookup handling can be isolated into a combo of e.g. `StaticDisplayNameProvider`s and `CompositeDisplayNameProvider`s.

wygoodin: This can also help to manage immutability/copying

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
```diff
@@ -34,9 +36,36 @@ public abstract class FormulaType
...
+                _type = type;
+            else 
+            {
+                ValidateDisplayNames(displayNameSet);
+                _type = DType.AttachDisplayNameProvider(type, new BidirectionalDictionary<string, string>(displayNameSet));
+            }
+
+        }
+
+        private void ValidateDisplayNames(Dictionary<string, string> displayNameSet)
```
wygoodin, line 50: This could also fit as a check on insert in the abstracted provider.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -265,6 +265,52 @@ public void CustomFunction()
...
+
+            var displayExpressions = engine.GetDisplayExpression("If(B, Num, 1234)", r1);
+
+            Assert.Equal("If(DisplayB, DisplayNum, 1234)", displayExpressions);
+        }
+
+        
+
+        [Fact]
+        public void ConvertToDisplayNameIsNoOp()
```
MikeStall, line 285: some other tests:

- can we have a mixed expression? (display + logical names) 

- calling GetDisplayExpression when WithDisplayNames() was never called?

- collision case (that throws the new exception)

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNameCollisionException.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.PowerFx.Core
+{
+    public class DisplayNameCollisionException : Exception
```
MikeStall, line 7: >DisplayNameCollisionException [](http://example.com/codeflow?start=17&length=29)

Do we need a dedicated exception here? Or just a generic Argument exception?

src/libraries/Microsoft.PowerFx.Core/Public/Types/AggregateType.cs
```diff
@@ -4,20 +4,21 @@
...
 using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.UtilityDataStructures;
 using Microsoft.PowerFx.Core.Utils;
 
 namespace Microsoft.PowerFx.Core.Public.Types
 {
     public abstract class AggregateType : FormulaType
     {
-        internal AggregateType(DType type) : base(type)
+        internal AggregateType(DType type, Dictionary<string, string> displayNameSet = null) : base(type, displayNameSet)
```
MikeStall, line 14: >Dictionary [](http://example.com/codeflow?start=43&length=10)

nit: IReadOnlyDictionary<> ?

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
```diff
@@ -33,9 +35,36 @@ public abstract class FormulaType
...
```
MikeStall, line 50: >ValidateDisplayNames [](http://example.com/codeflow?start=21&length=20)

static?

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
```diff
@@ -33,9 +35,36 @@ public abstract class FormulaType
...
+            HashSet<string> displayNames = new();
+            foreach (var kvp in displayNameSet)
+            {
+                // Display names can't collide with other display names
+                if (displayNames.Contains(kvp.Value))
+                {
+                    throw new DisplayNameCollisionException(kvp.Value);
+                }
+                // Display names can't collide with logical names other than their own
+                if (kvp.Key != kvp.Value && displayNameSet.ContainsKey(kvp.Value))
```
MikeStall, line 61: >!= [](http://example.com/codeflow?start=28&length=2)

What are the string casing rules?

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -163,13 +168,14 @@ public DType(DKind kind)
             Metadata = null;
             _attachmentType = null;
             AssociatedDataSources = new HashSet<IExternalTabularDataSource>();
+            DisplayNameProviders = new HashSet<BidirectionalDictionary<string, string>>();
```
MikeStall, line 171: >DisplayNameProviders = new HashSet<BidirectionalDictionary<string, string>>(); [](http://example.com/codeflow?start=12&length=78)

Is there a penalty to adding this to every dtype? (ie, should it be null most of the time)

MikeStall: Or share a single static readonly instance?

src/libraries/Microsoft.PowerFx.Core/Public/Types/AggregateType.cs
babrekel: Yes this should be updated -- I don't think we mutate the dictionary, correct?

lesaltzm: Yup. This should definitely be ReadOnly, good points

babrekel: An abstraction for `BidirectionalDictionary` could be good as well -- which avoids building and managing a hash set in the `FormulaType` constructor. You could use `ImmutableDictionary` which is a persistent form of dictionary, enabling things like adding a property including display name.

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
```diff
@@ -33,9 +35,36 @@ public abstract class FormulaType
...
+
+        private void ValidateDisplayNames(Dictionary<string, string> displayNameSet)
+        {
+            HashSet<string> displayNames = new();
+            foreach (var kvp in displayNameSet)
+            {
+                // Display names can't collide with other display names
+                if (displayNames.Contains(kvp.Value))
+                {
+                    throw new DisplayNameCollisionException(kvp.Value);
```
bryancroteau-MSFT, line 58: Should we be throwing an exception on the first display name conflict? What if there are multiple conflicts?

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
bryancroteau-MSFT: It's just another empty HashSet, it shouldn't be a problem

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -604,6 +624,17 @@ public static DType AttachDataSourceInfo(DType type, IExternalTabularDataSource
             return returnType;
         }
 
+        public static DType AttachDisplayNameProvider(DType type, BidirectionalDictionary<string, string> displayNames)
+        {
+            type.AssertValid();
+            Contracts.AssertValue(displayNames);
+
+            DType returnType = type.Clone();
+            returnType.DisplayNameProviders.Add(displayNames);
```
bryancroteau-MSFT, line 633: Does this need to be validated before being returned?

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -2355,6 +2411,20 @@ internal static bool TryGetConvertedDisplayNameAndLogicalNameForColumn(DType typ
                 }
             }
 
+            // The DisplayNameProvider path doesn't participate in Display -> Display remapping, just logical -> display and display -> logical.
+            // So in this case, we return the logical name and pretend we're upding display name to the same one we already have.
```
bryancroteau-MSFT, line 2415: upding (spelling)

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -2355,6 +2411,20 @@ internal static bool TryGetConvertedDisplayNameAndLogicalNameForColumn(DType typ
...
+            // This allows us to succeed even if we try to converting an expression that already is using display names to display format
```
bryancroteau-MSFT, line 2416: try to converting (grammar)

bryancroteau-MSFT (changes requested):

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNameCollisionException.cs
lesaltzm: I prefer a dedicated expression here, but I'm happy to hear why we shouldn't do that.

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
lesaltzm: Added comment calling it out, we're case-sensitive

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
lesaltzm: Changed API to make this clearer

src/libraries/Microsoft.PowerFx.Core/Public/NameCollisionException.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.PowerFx.Core
+{
+    public class NameCollisionException : Exception
+    {
+        public readonly string CollidingDisplayName;
```
MikeStall, line 9: >CollidingDisplayName [](http://example.com/codeflow?start=31&length=20)

nit - this should be a field, not property. 

CollidingDisplayName  {get;private set; }

MikeStall:
> using System;

nit: header

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/NameCollisionException.cs:1 in b167f31. [](commit_id = b167f31fcf8d0d78b2d8f25264af5ce209363d9a, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/DisplayNameProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.PowerFx.Core.UtilityDataStructures;
+
+namespace Microsoft.PowerFx.Core
+{
+    internal class DisplayNameProvider
+    {
+        // String comparison is case-sensitive by default, and that's the behavior we want
+        private BidirectionalDictionary<string, string> _displayNames;
```
MikeStall, line 14: >string [](http://example.com/codeflow?start=40&length=6)

nit:
- comment what the key & value are
- is the comparison case-sensitive or not?

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/DisplayNameProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        public override bool Equals(object obj)
+        {
+            return obj is DisplayNameProvider other && _displayNames.Equals(other._displayNames);
+        }
+
+        public override int GetHashCode()
+        {
+            return _displayNames.GetHashCode();
+        }
```
MikeStall, line 51: Do we need to add these?
These don't seem like operations we should be doing on this, particularly since this class is mutable.  (ie, dictionaries shouldn't have mutable keys; and GetHashCode() should be constant)

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -127,10 +128,14 @@ public static IEnumerable<DType> GetPrimitiveTypes()
...
         protected readonly DType _attachmentType;
 
-        internal HashSet<IExternalTabularDataSource> AssociatedDataSources { get; private set; }
+        internal HashSet<IExternalTabularDataSource> AssociatedDataSources { get; }
         internal IExternalOptionSet<int> OptionSetInfo { get; }
         internal IExternalViewInfo ViewInfo { get; }
 
+        // Eventually, all display names should come from this centralized source
+        // We should not be using individual DataSource/OptionSet/View references
+        internal HashSet<DisplayNameProvider> DisplayNameProviders { get; }
```
MikeStall, line 137: >HashSet [](http://example.com/codeflow?start=17&length=7)

Can this be a single item instead of a hashset?

lesaltzm: Updated to pair of DisplayNameProvider and flag whether to disable display names

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/DisplayNameProvider.cs
lesaltzm: added constructor with explicit comparator, and comments

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -127,10 +128,18 @@ public static IEnumerable<DType> GetPrimitiveTypes()
...
+        internal HashSet<IExternalTabularDataSource> AssociatedDataSources { get; }
         internal IExternalOptionSet<int> OptionSetInfo { get; }
         internal IExternalViewInfo ViewInfo { get; }
 
+        // Eventually, all display names should come from this centralized source
+        // We should not be using individual DataSource/OptionSet/View references
+        internal DisplayNameProvider DisplayNameProvider { get; private set; }
+
+        // If the type is derived from multiple DisplayNameProviders, disable display names.
+        private bool DisableDisplayNames = false;
```
MikeStall, line 140: >DisableDisplayNames [](http://example.com/codeflow?start=21&length=19)

Is this the same as setting DisplayNameProvider to null?

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/DisplayNameProvider.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            return _displayNames.TryGetFromSecond(displayName, out logicalName);
+        }
+        
+        public bool TryGetDisplayName(string logicalName, out string displayName)
+        {
+            return _displayNames.TryGetFromFirst(logicalName, out displayName);
+        }
+
+        public bool Matches(DisplayNameProvider other)
```
MikeStall, line 44: Is this O[N^2]?

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -199,6 +203,61 @@ public FormulaValue Eval(string expressionText, RecordValue parameters = null)
...
+                new Glue2DocumentBinderGlue(),
+                null,
+                new Core.Entities.QueryOptions.DataSourceToQueryOptionsMap(),
+                formula.ParseTree,
+                resolver,
+                ruleScope: parameters._type,
+                useThisRecordForRuleScope: false,
+                updateDisplayNames: toDisplayNames,
+                forceUpdateDisplayNames: toDisplayNames
+            );
```
MikeStall, line 250: Can this be shared with CheckInternal()?

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -199,6 +203,61 @@ public FormulaValue Eval(string expressionText, RecordValue parameters = null)
...
+
```
MikeStall, line 251: Can this be shared with CheckInternal()?

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/DisplayNameProvider.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Text;
+using System.Linq;
+using Microsoft.PowerFx.Core.UtilityDataStructures;
+
+namespace Microsoft.PowerFx.Core
+{
+    internal class DisplayNameProvider
+    {
+        // First is Logical Name, Second is Display Name
+        private BidirectionalDictionary<string, string> _displayNames;
```
MikeStall, line 15: >private [](http://example.com/codeflow?start=8&length=7)

nit: readonly

MikeStall, line 15: >string [](http://example.com/codeflow?start=40&length=6)

Should either of these be DName?

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
lesaltzm: No, let's say you had 3 different sources that get unioned into a single type. 
Then you'd wind up toggling it back to enabled if we were only going off of nullness of DNP.

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/DisplayNameProvider.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            return _displayNames.TryGetFromFirst(logicalName, out displayName);
+        }
+
+        public bool Matches(DisplayNameProvider other)
+        {
+            return other != null &&
+                _displayNames.Count() == other._displayNames.Count() &&
+                _displayNames.Keys.All(other._displayNames.ContainsFirstKey) &&
+                _displayNames.Values.Any(other._displayNames.ContainsSecondKey);
```
lesaltzm, line 49: all

src/libraries/Microsoft.PowerFx.Core/Public/Types/TableType.cs
```diff
@@ -31,10 +32,20 @@ public override void Visit(ITypeVistor vistor)
             vistor.Visit(this);
         }
 
-        public TableType Add(NamedFormulaType field)
+        public TableType Add(NamedFormulaType field, string optionalDisplayName = null)
```
MikeStall, line 35: >Add [](http://example.com/codeflow?start=25&length=3)

We're mixing mutable and immutable data types:

The fact that this returns a new TableType implies that it's immutable. 
But both TableTypes actually share the same displayNameProvider, and that's mutated.

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/DisplayNameProvider.cs
lesaltzm: They can both be, no harm in changing that.

src/libraries/Microsoft.PowerFx.Core/Public/Types/TableType.cs
MikeStall, line 35: >string optionalDisplayName = null [](http://example.com/codeflow?start=53&length=33)

to consider: should we add this param as part of NamedFormulaType?  They're conceptually locked, and that'd make the signatures cleaner.

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/DisplayNameProvider.cs
lesaltzm: No, it's O(N). Iterates each set of keys, and checks that it's a member of the other set, which is backed by a hash table.

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
```diff
@@ -33,9 +37,15 @@ public abstract class FormulaType
...
+        internal FormulaType(DType type, DisplayNameProvider displayNameProvider = null)
         {
-            _type = type;
+            if (displayNameProvider == null)
+                _type = type;
+            else 
+            {
+                _displayNameProvider = displayNameProvider;
+                _type = DType.AttachDisplayNameProvider(type, displayNameProvider);
+            }
```
MikeStall, line 48: to consider - this display name should already be associated with the incoming Dtype. 

Else we break this invariant:
DType t;
new FormulaType(t)._type == t.

MikeStall: ie, keep FormulaType as a lightweight wrapper over Dtype and avoid logic leaking into it.

src/tests/Microsoft.PowerFx.Interpreter.Tests/DisplayNameTests.cs
```diff
@@ -0,0 +1,172 @@
...
+            RecordType r1 = new RecordType()
+                .Add(new NamedFormulaType("Num", FormulaType.Number), "DisplayNum");
+
+            Assert.Throws<NameCollisionException>(() => r1.Add(new NamedFormulaType("DisplayNum", FormulaType.Date), "NoCollision"));
+            Assert.Throws<NameCollisionException>(() => r1.Add(new NamedFormulaType("NoCollision", FormulaType.Date), "DisplayNum"));
+            Assert.Throws<NameCollisionException>(() => r1.Add(new NamedFormulaType("NoCollision", FormulaType.Date), "Num"));
+        }
+
+        [Fact]
+        public void ConvertToDisplayNames()
```
MikeStall, line 32: another test case:
adding a comment / extra whitespace and verifying they're preserved during conversion. (ie, it's not just a pretty printer walking the AST).

src/tests/Microsoft.PowerFx.Interpreter.Tests/DisplayNameTests.cs
```diff
@@ -0,0 +1,172 @@
...
+            
+
+        [Fact]
+        public void ConvertToDisplayNameMixed()
+        {
+            var engine = new RecalcEngine();
+            RecordType r1 = new RecordType()
+                .Add(new NamedFormulaType("Num", FormulaType.Number), "DisplayNum")
+                .Add(new NamedFormulaType("B", FormulaType.Boolean), "DisplayB");
+
```
MikeStall, line 79: nit: a bunch of these could be Theories, and then they'd share the setup code.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -604,6 +616,29 @@ public static DType AttachDataSourceInfo(DType type, IExternalTabularDataSource
...
+            if (returnType.DisplayNameProvider == null)
+            {
+                returnType.DisplayNameProvider = displayNames;
+                returnType.DisableDisplayNames = false;
+            } 
+            else if (returnType.DisplayNameProvider.Matches(displayNames))
+            {
+                returnType.DisplayNameProvider = displayNames;
+                return returnType;
+            }
```
MikeStall, line 634: this is dangerous, given that a) the DNP is mutable, and b) from the caller's perspective, they don't know whether the DNP will end up getting shared or not.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
lesaltzm: only lines 235-239, which doesn't really give us that much benefit. Might be something to address later if we make the configuration part here more complex (functions, option sets, enums ...)

src/libraries/Microsoft.PowerFx.Core/Public/Types/TableType.cs
lesaltzm: Resolved. DisplayNameProvider now is more immutable w.r.t FormulaType constructors

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
lesaltzm: Great point, fixed

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
lesaltzm: Resolved, DNP is not mutateable within the internals of PowerFx.

src/libraries/Microsoft.PowerFx.Core/Public/Types/TableType.cs
```diff
@@ -31,10 +32,20 @@ public override void Visit(ITypeVistor vistor)
...
         {
+            var displayNameProvider = _displayNameProvider;
+            if (optionalDisplayName != null)
+            {
+                if (displayNameProvider == null)
+                    displayNameProvider = new DisplayNameProvider();
+
+                if (!displayNameProvider.TryAddField(field.Name, optionalDisplayName))
+                    throw new NameCollisionException(optionalDisplayName);
```
MikeStall: It would be more defensive to have this be AddField() and have the method throw. 

IE, when would somebody ever call TryAddField and *not* throw? That could open the door to some non-deterministic behavior (behavior depending on the order of calls to Try).

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/DisplayNameProvider.cs
MikeStall: per other conversations, we may just want to avoid calling Matches() altogether. It's not an efficient impl, and it's dubious how a caller would invoke it 100% correctly.

lesaltzm: Ah, good point. Reference equality should be fine here. thanks.

src/libraries/Microsoft.PowerFx.Core/Public/Types/TableType.cs
lesaltzm: yup, updated, and made more immutable.

src/libraries/Microsoft.PowerFx.Core/Public/Types/TableType.cs
```diff
@@ -33,7 +34,22 @@ public override void Visit(ITypeVistor vistor)
...
+
+                displayNameProvider = displayNameProvider.AddField(field.Name, field.DisplayName);
+            }
+
             var newType = _type.Add(field._typedName);
+
+            if (displayNameProvider != null)
+            {
+                newType = DType.ReplaceDisplayNameProvider(newType, displayNameProvider);
+            }
```
MikeStall, line 51: This looks the same as the new logic in RecordType.Add. 
Can this be shared and moved to AggregateType base class?

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/DisplayNameProvider.cs
```diff
@@ -0,0 +1,58 @@
...
+        
+        public bool TryGetDisplayName(DName logicalName, out DName displayName)
+        {
+            return _displayNames.TryGetFromFirst(logicalName, out displayName);
+        }
+
+        public DisplayNameProvider Clone()
+        {
+            return new DisplayNameProvider(_displayNames);
+        }
```
MikeStall, line 56: Cloning mutable dictionaries is expensive. Maybe ImmutableDictionary?  

If DNP is immutable, then the fields might need to be immutable too. 

[https://docs.microsoft.com/en-us/dotnet/api/system.collections.immutable.immutabledictionary-2?view=net-6.0](https://docs.microsoft.com/en-us/dotnet/api/system.collections.immutable.immutabledictionary-2?view=net-6.0)

src/libraries/Microsoft.PowerFx.Core/Public/Types/RecordType.cs
```diff
@@ -26,13 +26,14 @@ public override void Visit(ITypeVistor vistor)
...
+            return new RecordType(AddFieldToType(field));
         }
 
-        public RecordType Add(string name, FormulaType type)
+        public RecordType Add(string logicalName, FormulaType type, string optionalDisplayName = null)
         {
-            return Add(new NamedFormulaType(new TypedName(type._type, new DName(name))));
+            if (optionalDisplayName != null)
+                return Add(new NamedFormulaType(new TypedName(type._type, new DName(logicalName)), new DName(optionalDisplayName)));
+            return Add(new NamedFormulaType(new TypedName(type._type, new DName(logicalName))));
```
MikeStall, line 36: nit: Can this still change to the Add on line 27?

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
```diff
@@ -2,8 +2,10 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.UtilityDataStructures;
```
MikeStall, line 8: nit: can we remove this now?

src/tests/Microsoft.PowerFx.Interpreter.Tests/DisplayNameTests.cs
```diff
@@ -0,0 +1,90 @@
...
+using Microsoft.PowerFx.Core;
+using Microsoft.PowerFx.Core.Public.Types;
+using Microsoft.PowerFx.Core.Public.Values;
+using Microsoft.PowerFx.Core.Utils;
+using Xunit;
+using Xunit.Sdk;
+
+namespace Microsoft.PowerFx.Interpreter.Tests
+{    
+    public class DisplayNameTests
```
MikeStall, line 17: Is there a test for DisabledDisplayNameProvider?

MikeStall: Would also be good to add a test for the immutability issues we discussed. 

r = new RecordType() 

r1 = r.Add()  // "Logical" --> "Display1"
r2 = r.Add()  // "Logical" --> "Display2"

Ensure that r2 doesn't interfere with r1.

MikeStall (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -601,6 +609,40 @@ public static DType AttachDataSourceInfo(DType type, IExternalTabularDataSource
...
+            DType returnType = type.Clone();
+            returnType.DisplayNameProvider = displayNames;
+            return returnType;
+        }
+               
+
+        /// <summary>
+        /// This should be used by internal operations to update the set of display name providers associated with a type, i.e. during Union operations.
+        /// </summary>
+        public static DType AttachDisplayNameProvider(DType type, DisplayNameProvider displayNames)
```
bryancroteau-MSFT, line 632: This function actually disables display name providers in some cases, but this is not implied by the function name.

lesaltzm: updated name and doc comment

bryancroteau-MSFT (approved):

PR #89, opened 2021-12-16
Replace PowerFxConfig with EnumStore for Intellisense only

bryancroteau-MSFT:
The localization field was not being used by Intellisense, so there was no reason to pass the entire PowerFxConfig all the way down to that layer. DocumentServer calls the constructors in Intellisense, and treats localization differently than PowerFxConfig. More research is needed about how Intellisense actually handles localization internally.


MikeStall (approved):
:shipit:

PR #95, opened 2022-01-04
Toddne/expression

ToddN:
Added new class IExpression, with method Eval() to let us parse an expression once and evaluate it with different RecordValues.  The only change to RecalcEngine is really just adding ParseExpression, but I also sorting the using statements and maybe that's confused it?


src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -274,6 +275,40 @@ public static NumberValue Execute(NumberValue x, NumberValue y)
...
                 return FormulaValue.New(val);
             }
+        }
+
+        [Fact]
+        public void CheckBindErrorWithParseExpression()
+        {
+            var engine = new RecalcEngine();
+            IExpression parsedExpression; 
+            var result = engine.ParseExpression("3+foo+2", new RecordType(), out parsedExpression); // foo is undefined
```
MikeStall, line 285: >parsedExpression [](http://example.com/codeflow?start=81&length=16)

assert this is null?

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -1,348 +1,370 @@
...
-        /// <returns></returns>
-        public FormulaValue GetValue(string name)
-        {
-            var fi = Formulas[name];
-            return fi._value;
-        }
-    } // end class RecalcEngine
-}
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
```
MikeStall, line 2: nit: new files should have this header

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -1,348 +1,370 @@
...
+
+        /// <summary>
+        /// Returns the parse tree (AST) that Power Fx builds for this expression. 
+        /// </summary>
+        /// <param name="expressionText">The expression we plan to parse</param>
+        /// <param name="recordType">All types available at execution time</param>
+        /// <param name="expression">The parsed expression. Expression will be null if !CheckResult.IsSuccess</param>
+        /// <returns>CheckResult</returns> Returns any errors detected  
+
+        public CheckResult ParseExpression(string expressionText, RecordType recordType, out IExpression expression)
```
MikeStall, line 354: > out IExpression expression [](http://example.com/codeflow?start=88&length=27)

Should this just be another propery on CheckResult, and we use the existing Check method?

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
ToddN: Done.

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -30,7 +30,13 @@ public class CheckResult
...
         /// null on success, else contains errors. 
         /// </summary>
-        public ExpressionError[] Errors { get; set; }
+        public ExpressionError[] Errors { get; set; }
+
+
+        /// <summary>
+        /// Parsed expression, or null if IsSuccess is false
+        /// </summary>
+        public IExpression Expression {  get; set; }
```
MikeStall, line 39: >IExpression [](http://example.com/codeflow?start=15&length=11)

Some back ends (like SQL ) may still leave this null.

PR #96, opened 2022-01-05
Remove call to ToImmutableDictionary from EnumStore

bryancroteau-MSFT:
Enum storage is order sensitive. The call to ToImmutableDictionary was shuffling the order, leading to errors in some unit tests most of the time.


lesaltzm (approved):
Approved. We should discuss removing the ordering requirement from canvas (caused by the unqualified enum access feature), this is a fragile solution

MikeStall:
Some context from some offline discussion:
-  callers in PAClient are (wrongfully) depending on the order of the dictionary traversal.  They  should not be doing that.
-  The call to ToImmutableDictionary() hit this existing bug. 
We should make a fix in PAClient - but at least change fx to stay behavior preserving.

PR #99, opened 2022-01-10
Add an error message that appears for unmatched curly brackets

bryancroteau-MSFT:
The Lexer currently does not gracefully handle unmatched curly brackets, and has the potential to throw exceptions without useful error messages. This change removes the call to Pop on the lexer mode stack when the stack is empty, and provides error messages in the token stream when mismatched curly brackets are detected, including just before EoF is added to the token stream.


emhomm4 (approved):

MikeStall (approved):
:shipit:

PR #104, opened 2022-01-15
Initial working dot operator for CustomObjects

bryancroteau-MSFT:
Example of working dot operator for CustomObjects. Error messages and tests still need work.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
@@ -451,7 +451,9 @@ namespace Microsoft.PowerFx.Core.IR
                     // Field access within a custom object.
                     Contracts.Assert(typeLhs.IsCustomObject);
 
-                    result = new RecordFieldAccessNode(context.GetIRContext(node), left, nameRhs);
+                    var right = new TextLiteralNode(context.GetIRContext(node.RightNode), nameRhs);
+
+                    return new BinaryOpNode(context.GetIRContext(node), BinaryOpKind.DynamicGetField, left, right);
                 }
                 else
                 {
--- a/src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
@@ -446,6 +446,13 @@ namespace Microsoft.PowerFx.Core.IR
                         }
                     }
                 }
+                else if (typeLhs.IsCustomObject)
+                {
+                    // Field access within a custom object.
+                    Contracts.Assert(typeLhs.IsCustomObject);
+
+                    result = new RecordFieldAccessNode(context.GetIRContext(node), left, nameRhs);
+                }
                 else
                 {
                     Contracts.Assert(context.Binding.ErrorContainer.HasErrors(node.Left) || context.Binding.ErrorContainer.HasErrors(node));
--- a/src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
@@ -451,7 +451,7 @@ namespace Microsoft.PowerFx.Core.IR
                     // Field access within a custom object.
                     Contracts.Assert(typeLhs.IsCustomObject);
 
-                    var right = new TextLiteralNode(context.GetIRContext(node.RightNode), nameRhs);
+                    var right = new TextLiteralNode(IRContext.NotInSource(FormulaType.String), nameRhs);
 
                     return new BinaryOpNode(context.GetIRContext(node), BinaryOpKind.DynamicGetField, left, right);
                 }
--- a/src/libraries/Microsoft.PowerFx.Core/IR/Nodes/BinaryOpKind.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/IR/Nodes/BinaryOpKind.cs
@@ -75,6 +75,7 @@ namespace Microsoft.PowerFx.Core.IR.Nodes
         LeqTime,
         GtTime,
         GeqTime,
+        DynamicGetField,
 
         // And, Or, Pow, Concatenate get represented as FunctionNodes with lambdas to handle short-circuiting
     }
--- a/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
@@ -244,6 +244,34 @@ namespace Microsoft.PowerFx
                     return OperatorGtTime(this, context, node.IRContext, args);
                 case BinaryOpKind.GeqTime:
                     return OperatorGeqTime(this, context, node.IRContext, args);
+                case BinaryOpKind.DynamicGetField:
+                    if (arg1 is CustomObjectValue cov && arg1 is StringValue sv)
+                    {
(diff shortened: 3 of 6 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -341,6 +341,35 @@ public override FormulaValue Visit(RecordFieldAccessNode node, SymbolContext con
...
+            {
+                if (cov.Impl.IsObject)
+                {
+                    if (cov.Impl.TryGetProperty(node.Field.Value, out var res))
+                    {
+                        return new CustomObjectValue(node.IRContext, res);
+                    }
+                    else
+                    {
+                        return new ErrorValue(node.IRContext, new ExpressionError()
```
lesaltzm, line 354: We should return BlankValue for missing keys in an object, so we're consistent with First(Table({a: 3), {b: 4})) => Blank()

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -341,6 +341,35 @@ public override FormulaValue Visit(RecordFieldAccessNode node, SymbolContext con
...
+                            Span = node.IRContext.SourceContext,
+                            Kind = ErrorKind.BadLanguageCode
+                        });
+                    }
+                }
+                else
+                {
+                    return new ErrorValue(node.IRContext, new ExpressionError()
+                    {
+                        Message = "The CustomObject does not represent an object",
```
lesaltzm, line 366: This works, or maybe something like "Accessing a field is not valid on this value"

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -446,6 +446,13 @@ public override IntermediateNode Visit(DottedNameNode node, IRTranslatorContext
                         }
                     }
                 }
+                else if (typeLhs.IsCustomObject)
+                {
+                    // Field access within a custom object.
+                    Contracts.Assert(typeLhs.IsCustomObject);
+
+                    result = new RecordFieldAccessNode(context.GetIRContext(node), left, nameRhs);
```
lesaltzm, line 454: This makes sense. If we go and add a GetField function later, we'll want to unify it at the IR level.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
bryancroteau-MSFT: Done

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
bryancroteau-MSFT: Changed this message

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/CustomObject.txt
```diff
@@ -11,4 +11,7 @@
 "s"
 
 >> Sum(Table(ParseJson("[1, 2, 3, 4, 5]")), Value(Value))
-15
\ No newline at end of file
+15
+
+>> Value(ParseJson("{""a"": 5}").a)
+5
```
MikeStall, line 17: looking great!

can you add test for missing field too?
{a : 5} .b

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -341,6 +341,30 @@ public override FormulaValue Visit(RecordFieldAccessNode node, SymbolContext con
                 return left;
             }
 
+            if (left is CustomObjectValue cov)
```
MikeStall, line 344: > if (left is CustomObjectValue cov) [](http://example.com/codeflow?start=11&length=35)

Design question here....

We want to make an IR easy to implement, and easy to know what's *not* implemented - because multiple backends are handling. 
Binary operator has an enum, and so that's easy to know which kinds are nyi. So the IR nodes should be translated enough that we can avoid switching logic like this (how does a backend implementor know the various cases?)

Along that, what do you think of translating Co.field as a binary operator with a new BinaryOpKind.DynamicGetField?  
If we add GetField() function, that can also translate to this op. (I think And/Or have a similar issue) 

(@lesaltzm fyi)

src/libraries/Microsoft.PowerFx.Core/IR/Nodes/BinaryOpKind.cs
```diff
@@ -75,6 +75,7 @@ internal enum BinaryOpKind
         LeqTime,
         GtTime,
         GeqTime,
+        DynamicGetField,
```
MikeStall, line 78: >DynamicGetField, [](http://example.com/codeflow?start=8&length=16)

This is great. Now hosts that don't impl it will just get a NotImpl exception from their unhandled switch.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -244,6 +244,34 @@ public override FormulaValue Visit(BinaryOpNode node, SymbolContext context)
...
+                            {
+                                Message = "Accessing a field is not valid on this value",
+                                Span = node.IRContext.SourceContext,
+                                Kind = ErrorKind.BadLanguageCode
+                            });
+                        }
+                    }
+                    else
+                    {
+                        return new BlankValue(node.IRContext);
```
MikeStall, line 273: > return new BlankValue(node.IRContext); [](http://example.com/codeflow?start=23&length=39)

Can this happen? Or would this be a compiler error caught at an earlier binding stage (ie, should we throw here?)

MikeStall (dismissed):
:shipit:

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
lesaltzm: This should probably be split into cases. if Arg1 is blankvalue, we should absolutely return blankvalue, but other than that, we should probably thow InvalidOperation

lesaltzm: same for passing through error values from arg1, I think

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/CustomObject.txt
```diff
@@ -11,4 +11,13 @@
...
+
+>> Value(ParseJson("{""a"": 5}").b)
+Blank()
+
+>> Value(ParseJson("""s""").a)
+#Error
```
lesaltzm, line 23: What about something like 
`Value(ParseJson("This "" Is , "" Invalid ").a`?

bryancroteau-MSFT: This should be a compile error because the result of value is a number and cannot be used with the dot operator

lesaltzm: ah, whoops, meant 
Value(ParseJson("This "" Is , "" Invalid ").a)

PR #107, opened 2022-01-18
Added named formula class

emhomm4:
This PR creates a class for Named Formulas. It also adds a constructor for Formula that takes an already parsed parse tree. This is so that when named formulas are parsed, their right-hand sides can be Formula objects without re-parsing.


VamsiModem (approved):

lesaltzm:
Please add some tests parsing and constructing NamedFormulas

src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormula.cs
```diff
@@ -0,0 +1,49 @@
...
+using Microsoft.PowerFx.Core.Parser;
+using Microsoft.PowerFx.Core.Syntax.Nodes;
+using Microsoft.PowerFx.Core.Utils;
+
+namespace Microsoft.PowerFx.Core.Syntax
+{
+    /// <summary>
+    /// This encapsulates a named formula: its original script, the parsed result, and any parse errors.
+    /// </summary>
+    internal sealed class NamedFormula
```
lesaltzm, line 16: should this be internal or public? It seems like it's being used from the powerfx side.

src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormula.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        public readonly string Script;
+
+        // The language settings used for parsing this script.
+        // May be null if the script is to be parsed in the current locale.
+        public readonly ILanguageSettings Loc;
+
+        private List<TexlError> _errors;
+
+        internal Dictionary<DName, TexlNode> _formulasResult;
```
lesaltzm, line 26: consider something more complex, passing along the substring corresponding to the rule, such as Dict<DName, (TexlNode node, string script)>

lesaltzm, line 26: nit: `_` prefixes are for private members only.

lesaltzm: Or provide like a `TryGetScript(DName, out string)` method maybe

lesaltzm (changes requested):
Please add some tests parsing and constructing NamedFormulas

src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormula.cs
emhomm4: Added a method to get the substring.

src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormula.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            return _errors == null;
+        }
+
+        /// <summary>
+        /// Gets the part of the script associated with that node.
+        /// </summary>
+        /// <param name="node"></param>
+        /// <returns></returns>
+        public string GetSubscript(TexlNode node)
```
lesaltzm, line 66: How do we know this node is part of the script?

Instead, consider taking the DName

src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormula.cs
emhomm4: ILanguageSettings and TexlNode are both internal though.

lesaltzm: ah, fair enough.

emhomm4:
> Please add some tests parsing and constructing NamedFormulas

Added

src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormula.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// </summary>
+        public readonly string Script;
+
+        // The language settings used for parsing this script.
+        // May be null if the script is to be parsed in the current locale.
+        public readonly ILanguageSettings Loc;
+
+        public Dictionary<DName, TexlNode> FormulasResult;
+
+        private List<TexlError> _errors;
```
lesaltzm, line 29: Do we also need a HasErrors/GetErrors function?

lesaltzm (approved):

src/tests/Microsoft.PowerFx.Core.Tests/NamedFormulaTests.cs
```diff
@@ -0,0 +1,59 @@
...
+
+using System.Collections.Generic;
+using Microsoft.PowerFx.Core.Syntax;
+using Xunit;
+
+namespace Microsoft.PowerFx.Core.Tests
+{
+    public class NamedFormulaTests
+    {
+        private readonly string script = "x=1;y=2;";
```
lesaltzm, line 12: in your next PR for parser errors, we'll need a lot more tests, converting this to use [Theory] and [InlineData] would be good.

src/libraries/Microsoft.PowerFx.Core/Syntax/Formula.cs
```diff
@@ -40,6 +40,17 @@ public Formula(string script, ILanguageSettings loc = null)
             AssertValid();
         }
 
+        public Formula(string script, TexlNode tree, ILanguageSettings loc = null)
```
MikeStall, line 43: nit: We could use constructor chaining to share the impl between ctors:

 public Formula(string script, ILanguageSettings loc = null) 
 : this(script, null, loc) 
{ }

src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormula.cs
emhomm4: Added

src/tests/Microsoft.PowerFx.Core.Tests/NamedFormulaTests.cs
emhomm4: Changed

src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormula.cs
```diff
@@ -0,0 +1,74 @@
...
+using Microsoft.PowerFx.Core.Parser;
+using Microsoft.PowerFx.Core.Syntax.Nodes;
+using Microsoft.PowerFx.Core.Utils;
+
+namespace Microsoft.PowerFx.Core.Syntax
+{
+    /// <summary>
+    /// This encapsulates a named formula: its original script, the parsed result, and any parse errors.
+    /// </summary>
+    internal class NamedFormula
```
MikeStall, line 17: >NamedFormula [](http://example.com/codeflow?start=19&length=12)

nit: should this be plural?

src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormula.cs
```diff
@@ -0,0 +1,95 @@
...
+            return _errors == null;
+        }
+
+        /// <summary>
+        /// Returns the formula portion of the script for a given named formula identifier.
+        /// </summary>
+        /// <param name="name"></param>
+        /// <param name="subScript"></param>
+        /// <returns></returns>
+        public bool TryGetSubscript(DName name, out string subScript)
```
MikeStall, line 76: >string [](http://example.com/codeflow?start=52&length=6)

curious: who needs the original string (as opposed to the TexlNode)

lesaltzm: Most of PowerApps-Client expects a rule to have a string as well as a parse tree. It's mostly used for telemetry, but is in a large enough part of the codebase that it's easier to meet that assumption

MikeStall: Consider returning Formula (or even just enumerating formulas).

src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormula.cs
```diff
@@ -0,0 +1,95 @@
...
+        public readonly ILanguageSettings Loc;
+
+        public Dictionary<DName, TexlNode> FormulasResult;
+
+        public bool IsParsed => FormulasResult != null;
+
+        public bool HasParseErrors { get; private set; }
+
+        private List<TexlError> _errors;
+
```
MikeStall, line 35: Potentially future issue ... do the errors need to be scoped to the DName?

src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormula.cs
```diff
@@ -0,0 +1,95 @@
...
+using Microsoft.PowerFx.Core.Parser;
+using Microsoft.PowerFx.Core.Syntax.Nodes;
+using Microsoft.PowerFx.Core.Utils;
+
+namespace Microsoft.PowerFx.Core.Syntax
+{
+    /// <summary>
+    /// This encapsulates a named formula: its original script, the parsed result, and any parse errors.
+    /// </summary>
+    internal class NamedFormula
```
MikeStall, line 17: >NamedFormula [](http://example.com/codeflow?start=19&length=12)

Is this class mutable or not? 
See slide 57 in greg's XR - what happens when we have dedicated UI and show each formula as its own UI element (Rather than a single giant Script  value)

lesaltzm, line 17: Actually, how much are we getting from this class? Why do we need this as opposed to just 
`(Dict<Dname, Node> result, Errors errors) TexlParser.ParseNamedFormulas(Script, languagesettings)`

src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormulas.cs
```diff
@@ -0,0 +1,116 @@
...
+            Contracts.AssertValue(Script);
+            Contracts.Assert(IsParsed, "Should call EnsureParsed() first!");
+            return _errors ?? Enumerable.Empty<TexlError>();
+        }
+
+        /// <summary>
+        /// Returns a Formula object for each named formula.
+        /// </summary>
+        /// <returns></returns>
+        public List<Formula> GetFormulas()
```
MikeStall, line 87: >List [](http://example.com/codeflow?start=15&length=4)

IEnumerable?

src/tests/Microsoft.PowerFx.Core.Tests/NamedFormulasTests.cs
```diff
@@ -0,0 +1,87 @@
...
+            Assert.Empty(errors);
+        }
+
+        [Theory]
+        [InlineData("x=1;y=2;", "1", "2")]
+        public void GetFormulasTest(string script, string expectedX, string expectedY)
+        {
+            var namedFormula = new NamedFormulas(script);
+            namedFormula.EnsureParsed();
+            var formulas = namedFormula.GetFormulas();
```
MikeStall, line 67: >GetFormulas [](http://example.com/codeflow?start=40&length=11)

is Formulas ordered?

src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormulas.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            _formulas = new List<Formula>();
+            if (FormulasResult != null)
+            {
+                foreach (var kvp in FormulasResult)
+                {
+                    TryGetSubscript(kvp.Key, out var subScript);
```
lesaltzm, line 94: Seems odd to call a tryget that just looks up into the dict we're iterating over, and not check the result. Could just be 
`var subscript = kvp.Value.GetCompleteSpan.GetFragment(Script);`

src/tests/Microsoft.PowerFx.Core.Tests/NamedFormulasTests.cs
```diff
@@ -0,0 +1,87 @@
...
+
+            Assert.NotNull(formulas);
+            Assert.Equal(expectedX, formulas[0].Script);
+            Assert.Equal(expectedY, formulas[1].Script);
+        }
+
+        [Theory]
+        [InlineData("x=1;y=2;", "Not a correct script")]
```
lesaltzm, line 75: We should be expecting Null for these .Script values, right? Maybe just Assert.IsNull() on 83-84?

src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormulas.cs
```diff
@@ -0,0 +1,106 @@
...
+            return _errors ?? Enumerable.Empty<TexlError>();
+        }
+
+        /// <summary>
+        /// Returns a Formula object for each named formula.
+        /// </summary>
+        /// <returns></returns>
+        public IEnumerable<Formula> GetFormulas()
+        {
+            _formulas = new List<Formula>();
```
lesaltzm, line 89: Is there a reason to have this as a member variable, but then repopulate it each time?

src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormulas.cs
```diff
@@ -0,0 +1,106 @@
...
+        /// <summary>
+        /// A script containing one or more named formulas.
+        /// </summary>
+        public readonly string Script;
+
+        // The language settings used for parsing this script.
+        // May be null if the script is to be parsed in the current locale.
+        public readonly ILanguageSettings Loc;
+
+        public Dictionary<DName, TexlNode> FormulasResult;
```
lesaltzm, line 28: Should this be public, or do we expect it to be read only through GetFormulas()

lesaltzm (approved):

lesaltzm (approved):

PR #109, opened 2022-01-20
Update CustomObject semantics for Value, Text, Dot

bryancroteau-MSFT:
This change enables Value and Text to handle boolean values, json null, and strings/numbers respectively. It also enables json null support for the dot operator itself. Allows error values to pass through the dot operator, similar to functions.


src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryCustomObject.cs
```diff
@@ -124,12 +131,55 @@ public static FormulaValue Index_CO(IRContext irContext, FormulaValue[] args)
...
+            {
+                str = impl.GetString();
+            }
+            else if (impl.IsNull)
+            {
+                str = string.Empty;
+            }
+            else if (impl.IsBoolean)
+            {
+                str = impl.GetBoolean() ? "true" : "false";
```
MikeStall, line 175: > str = impl.GetBoolean() ? "true" : "false"; [](http://example.com/codeflow?start=15&length=44)

Can (should?) this chain to the existing Text function?
Ensure we get details right (like whether "true" is upper or lower; locales, etc. 
That provides a path for handling format string.

src/libraries/Microsoft.PowerFx.Core/Public/Values/CustomObjectValue.cs
```diff
@@ -23,6 +23,8 @@ public interface ICustomObject
 
         bool IsNumber { get; }
 
+        bool IsBoolean { get; }
```
MikeStall, line 26: are all these flags exclusive? What do you think of making this a single enum?

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -251,13 +251,22 @@ public override FormulaValue Visit(BinaryOpNode node, SymbolContext context)
...
+
                                 return new CustomObjectValue(node.IRContext, res);
                             }
                             else
                             {
                                 return new BlankValue(node.IRContext);
                             }
                         }
+                        else if (cov.Impl.IsNull)
+                        {
```
MikeStall, line 267: this check here is what prompted me wondering about being an enum instead of multiple IsXXX functions. 

Because if an impl did haev IsNull and IsObj both true, then the order of checks here would start to matter.

MikeStall:
>         ICustomObject this[int index] { get; }

nit: add comment whether this is 0 or 1-based index?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/Values/CustomObjectValue.cs:32 in be91902. [](commit_id = be919028a3f2a85531b7d445b4afebbcaea197f4, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Public/Values/CustomObjectValue.cs
```diff
@@ -23,6 +23,8 @@ public interface ICustomObject
...
+
         object ToObject();
```
MikeStall, line 28: >object ToObject(); [](http://example.com/codeflow?start=8&length=18)

can we avoid this on the interface? It's ambiguous semantics and can be hard for hosts to implement. 
The JSon obj just returns the json str, which is probably not what's expected. 

It looks like this is just used for FormulaValue.ToObject(), which is just used for marshalling back to C#. 

Can we have a default impl based on the other methods on the interface?
What about:
 object ToObj(ICustomObject x) {
    if (x.IsNumber()) return x.GetDouble();
    if (x.IsString()) return x.GetString(); 
    ...
    throw new NotImplemented("can't marshal to ob");
}

MikeStall:
>         bool TryGetProperty(string value, out ICustomObject result);

Another doc nit: Is the assumption that IsObject is true before calling this?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/Values/CustomObjectValue.cs:34 in be91902. [](commit_id = be919028a3f2a85531b7d445b4afebbcaea197f4, deletion_comment = False)

bryancroteau-MSFT:
> > ```
> >     bool TryGetProperty(string value, out ICustomObject result);
> > ```
> 
> Another doc nit: Is the assumption that IsObject is true before calling this?
> 
> Refers to: src/libraries/Microsoft.PowerFx.Core/Public/Values/CustomObjectValue.cs:34 in [be91902](https://github.com/microsoft/Power-Fx/commit/be919028a3f2a85531b7d445b4afebbcaea197f4). [](commit_id = [be91902](https://github.com/microsoft/Power-Fx/commit/be919028a3f2a85531b7d445b4afebbcaea197f4), deletion_comment = False)

The error handling pipeline does do a runtime value check.

src/libraries/Microsoft.PowerFx.Core/Public/Values/CustomObjectValue.cs
```diff
@@ -11,17 +11,19 @@
 
 namespace Microsoft.PowerFx.Core.Public.Values
 {
-    public interface ICustomObject
+    public enum CustomObjectKind
```
MikeStall, line 14: >CustomObjectKind [](http://example.com/codeflow?start=16&length=16)

Some notes for design discussion:

1. How will it handle future ICustomObject impls beyond json. Like Xml, which has more date,time and other types.

2. The thunk idea where we create a ICustomObj impl around a FormulaValue. Or wrap around a C# object (ala a dynamic version of FormulaValue.New)

3. Potential overlap with other FormulasValues. Eg,  ICustomObject.TryGetProperty() vs RecordValue.GetField(). Maybe there are some common interfaces?

The sole purpose of an ICustomObject is to produce FormulaValues, consider if ICustomObject.Kind should return a FormulaType rather than a new enum (which then needs to just be converted to a formula type).

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/CustomObject.txt
```diff
@@ -1,17 +1,24 @@
-﻿>> Value(Index(ParseJson("[5]"), 1))
+﻿// Value tests
+>> Value(Index(ParseJson("[5]"), 1))
```
MikeStall, line 2: >Value [](http://example.com/codeflow?start=3&length=5)

add test : 
equality ops on CO is an error. 
x = ParseJson("5"); x= x  is an error.

src/libraries/Microsoft.PowerFx.Core/Public/Values/CustomObjectValue.cs
bryancroteau-MSFT: Changed to an enum

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
bryancroteau-MSFT: Changed to an enum

src/libraries/Microsoft.PowerFx.Core/Public/Values/CustomObjectValue.cs
MikeStall: We discussed offline - 
this needs to describe foriegn types (like json arrays), so we'll need to keep the enum - we can't change it to FormulaType.

src/libraries/Microsoft.PowerFx.Core/Public/Values/CustomObjectValue.cs
bryancroteau-MSFT: This will not work because we have no way of returning arrays/objects.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/CustomObject.txt
bryancroteau-MSFT: Do we want to have negative tests everywhere in the language for things we do not support (and plan to never support)?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryCustomObject.cs
bryancroteau-MSFT: Are you saying we need to construct a new BooleanValue to pass into the existing coercion?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/CustomObject.txt
MikeStall: Not "everywhere"  - that could be a never ending chore. 

But things that we explicitly had design discussions on and agreed should fail are good candidates for testing.  It gives an opportunity to capture the reasoning, ensure it doesn't accidentally work (perhaps some other bug, or coercion that wrongfully kicks in)  and prevent regressions. 
Especially protects us against a future dev that wants to "helpfully" enable it.

src/libraries/Microsoft.PowerFx.Core/Public/Types/ExternalType.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Microsoft.PowerFx.Core.Public.Types
+{
+    /// <summary>
+    /// This enum represents types which are not supported by PowerFx
+    /// but which may be supported by external data formats.
+    /// </summary>
+    public enum ExternalTypeKind
+    {
+        Array,
```
MikeStall, line 15: >Array [](http://example.com/codeflow?start=8&length=5)

Can you add some more comments, especially to help clarify why these aren't just FormulaTypes. :

Arrays - PowerFx only supports Single Column Tables. Json supports arrays. 
Object - maybe rename to UntypedObject? (to help distinguish from fx's record).

MikeStall: You've got it on the ExternalType class, but it would be good to comment the enum members.

src/libraries/Microsoft.PowerFx.Core/Public/Values/CustomObjectValue.cs
```diff
@@ -13,15 +13,7 @@ namespace Microsoft.PowerFx.Core.Public.Values
 {
     public interface ICustomObject
```
MikeStall, line 14: >ICustomObject [](http://example.com/codeflow?start=21&length=13)

This interface needs some more comments.

MikeStall:
>         protected readonly ICustomObject _impl;

can this be private?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/Values/CustomObjectValue.cs:35 in 7d7f7a5. [](commit_id = 7d7f7a521b3767468c8e50c8aadd472526211505, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryCustomObject.cs
```diff
@@ -15,13 +15,39 @@ internal static partial class Library
     {
         internal class JsonCustomObject : ICustomObject
         {
+            private static readonly FormulaType _objectType = new ExternalType(ExternalTypeKind.Object);
+            private static readonly FormulaType _arrayType = new ExternalType(ExternalTypeKind.Array);
```
MikeStall, line 19: Should this be static fields on the ExternalType class itself? They're generally useful.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryCustomObject.cs
```diff
@@ -34,6 +60,8 @@ public JsonCustomObject(JsonElement element)
 
             public bool IsNumber => _element.ValueKind == JsonValueKind.Number;
 
+            public bool IsBoolean => _element.ValueKind == JsonValueKind.True || _element.ValueKind == JsonValueKind.False;
```
MikeStall, line 63: Do we still need this?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryCustomObject.cs
```diff
@@ -49,9 +77,14 @@ public string GetString()
...
 
+            public bool GetBoolean()
+            {
+                return _element.GetBoolean();
+            }
+
             public object ToObject()
             {
-                return _element.GetRawText();
+                return _element;
```
MikeStall, line 87: >return _element; [](http://example.com/codeflow?start=16&length=16)

I don't think there's any correct way to implement the ToObject() method. 
Can we just remove it entirely from the interface? 
In theory, if we really had to create an .net Object for it, an adatper could provide it leveraging the existing type support.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryCustomObject.cs
```diff
@@ -15,13 +15,39 @@ internal static partial class Library
     {
         internal class JsonCustomObject : ICustomObject
```
MikeStall, line 16: >JsonCustomObject [](http://example.com/codeflow?start=23&length=16)

nit: can you move this to its own file? It's very specific to json. (in the future, we might even consider moving it to its own assembly along with ParseJson and our STJ dependency)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryCustomObject.cs
```diff
@@ -124,12 +157,55 @@ public static FormulaValue Index_CO(IRContext irContext, FormulaValue[] args)
 
         public static FormulaValue Value_CO(EvalVisitor runner, SymbolContext symbolContext, IRContext irContext, CustomObjectValue[] args)
```
MikeStall, line 158: >Value_CO [](http://example.com/codeflow?start=35&length=8)

This should chain to the existing Value() func
- because there are a lot of corner cases (like % handling)
- to handle format args. 

What about:

FormulaValue fv = GetAsFv(impl);
return Value(runner, symbolContext, irContext,  
   new FV[] { fv, arg1 })


And then GetAsFv() has this switch logic and can be reused by the other converter functions.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/CustomObject.txt
```diff
@@ -1,23 +1,91 @@
...
+#Error
+
+>> Value(ParseJson("null"))
+Blank()
+
+>> Value(ParseJson("{""a"": 5}").a)
+5
+
+>> Value(ParseJson("{""a"": 5}").b)
+Blank()
```
MikeStall, line 27: a few more cases (to catch the various conditions in EvalVisitor)

- ParseJson("{}").a // empty record
- ParseJson("{a:null}").a.b //  coalesce, via TryGetProperty check
- ParseJson("null").a.b // getting a BlankType custom object  directly (bypassing the TryGetProperty check)

MikeStall:
> // Value tests

Can we rename this ParseJson.txt?  It's excercising custom objects, but it's really testing the ParseJson impl. 

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/CustomObject.txt:1 in 7d7f7a5. [](commit_id = 7d7f7a521b3767468c8e50c8aadd472526211505, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/CustomObject.txt
```diff
@@ -1,23 +1,91 @@
...
+>> Value(ParseJson("""5"""))
+5
+
+>> Value(ParseJson("""a"""))
+#Error
+
+>> Value(ParseJson("null"))
+Blank()
+
+>> Value(ParseJson("{""a"": 5}").a)
```
MikeStall, line 23: >.a) [](http://example.com/codeflow?start=32&length=3)

Can you add Case sensitivity tests?
Json is case sensitive  :| ...

ParseJson("{a :15, A : 20}").a // 15
ParseJson("{a :15, A : 20}").A // 20

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryCustomObject.cs
bryancroteau-MSFT: This should be a topic of discussion for a future PR.

src/libraries/Microsoft.PowerFx.Core/Public/Types/ExternalType.cs
bryancroteau-MSFT: Done

src/libraries/Microsoft.PowerFx.Core/Public/Values/CustomObjectValue.cs
bryancroteau-MSFT: Added a few comments

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryCustomObject.cs
bryancroteau-MSFT: Done

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryCustomObject.cs
bryancroteau-MSFT: Removed

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryCustomObject.cs
bryancroteau-MSFT: Removed, we just return the Impl now

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/CustomObject.txt
bryancroteau-MSFT: Done

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/CustomObject.txt
bryancroteau-MSFT: Done

src/libraries/Microsoft.PowerFx.Core/Public/Values/CustomObjectValue.cs
```diff
@@ -11,47 +11,41 @@
...
-        bool IsNull { get; }
-
-        bool IsObject { get; }
-
-        bool IsString { get; }
-
-        bool IsNumber { get; }
-
-        object ToObject();
+        FormulaType Type { get; } // Use ExternalType if the type is incompatible with PowerFx
```
MikeStall, line 20: >// [](http://example.com/codeflow?start=34&length=2)

nit: these should be xml comments ///  (since they're on a public interface) rather than just //. The C# IDE will then show xml comments as tooltips when called.

src/libraries/Microsoft.PowerFx.Core/Public/Values/CustomObjectValue.cs
```diff
@@ -11,47 +11,41 @@
...
         {
             Contract.Assert(IRContext.ResultType == FormulaType.CustomObject);
-            _impl = impl;
+            Impl = impl;
         }
 
         public override object ToObject()
         {
-            return _impl.ToObject();
+            return Impl; // Hosts will need to be able to interpret the backing value
```
MikeStall, line 48: >return Impl; // Hosts will need to be able to interpret the backing value [](http://example.com/codeflow?start=12&length=73)

Let's just throw NotImpl. - that's better than returning something surprising. 

The only right way to impl this is probably as C# dynamic object that's an adapter over ICustomObject. 

We can impl later if somebody actually needs this.

bryancroteau-MSFT: Done

src/libraries/Microsoft.PowerFx.Core/Public/Values/CustomObjectValue.cs
bryancroteau-MSFT: Done

MikeStall (approved):
:shipit:

PR #110, opened 2022-01-20
Revert "Add new error kind enum value (canceled)"

CarlosFigueiraMSFT:
Reverts microsoft/Power-Fx#108


src/libraries/Microsoft.PowerFx.Core/Public/ErrorKind.cs
```diff
@@ -33,8 +33,6 @@ public enum ErrorKind
         InsufficientMemory = 21,
         QuotaExceeded = 22,
         Network = 23,
-        Numeric = 24,
-        InvalidArgument = 25,
```
lesaltzm, line 37: Maybe keep InvalidArgument? Looks like that was just fixing something that was missing

CarlosFigueiraMSFT: Good point - added it back

lesaltzm (approved):

PR #115, opened 2022-01-26
Changed GetFormulas to GetNamedFormulas

emhomm4:
Changed GetFormulas to GetNamedFormulas with a Tuple return type and returning a DName as well as a formula.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormulas.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormulas.cs
@@ -83,14 +83,14 @@ namespace Microsoft.PowerFx.Core.Syntax
         /// Returns a Tuple of a DName and Formula object for each named formula.
         /// </summary>
         /// <returns></returns>
-        public IEnumerable<Tuple<DName, Formula>> GetNamedFormulas()
+        public IEnumerable<(DName name, Formula formula)> GetNamedFormulas()
         {
-            var formulas = new List<Tuple<DName, Formula>>();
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormulas.cs
```diff
@@ -79,17 +80,17 @@ public IEnumerable<TexlError> GetParseErrors()
         }
 
         /// <summary>
-        /// Returns a Formula object for each named formula.
+        /// Returns a Tuple of a DName and Formula object for each named formula.
         /// </summary>
         /// <returns></returns>
-        public IEnumerable<Formula> GetFormulas()
+        public IEnumerable<Tuple<DName, Formula>> GetNamedFormulas()
```
lesaltzm, line 86: C# has some nice things now like named tuples. This type could be 
`IEnumerable<(DName name, Formula formula)>`, and then instead of Tuple.Create, just do `formulas.Add((kvp.key, GetFormula(kvp.Value)))`

lesaltzm: see https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/value-tuples

emhomm4: Updated

VamsiModem (approved):

lesaltzm (approved):

PR #117, opened 2022-01-26
Fix whitespace bug in formulas parser

emhomm4:
Currently, the named formulas parser has a bug where it is not resilient to spaces in a formula. This PR fixes that. Unit tests updated to include random whitespace throughout formulas.


src/tests/Microsoft.PowerFx.Core.Tests/ParseTests.cs
```diff
@@ -655,25 +655,26 @@ public void TestParseTables_Negative(string script)
...
-        [InlineData("a=10;b")]
-        [InlineData("A=If(true,1;);")]
-        [InlineData("A=If(true,1;2);")]
-        [InlineData("1+2);")]
-        [InlineData("1=10")]
+        [InlineData("a = 10;;")]
+        [InlineData("a = 10; b")]
+        [InlineData("A = If(true,1;);")]
+        [InlineData("A = If(true,1;2);")]
+        [InlineData("1 + 2);")]
```
lesaltzm, line 676: what about multi character names?
Multi character names with special characters?
Ones with whitespace in the name, surrounded by `'`?

emhomm4: Is there a way for me to know which special characters are allowed in identifiers?

emhomm4: I updated the tests. I don't think any special characters are allowed in identifiers, but correct me if I'm wrong.

emhomm4: Added tests with special characters inside of `'`

lesaltzm (approved):

VamsiModem (approved):

PR #118, opened 2022-01-26
Add Environment Symbols registration to Power Fx Interpreter

lesaltzm:
We want a path to add symbols specific to a host's context. This adds an immutable store, and methods on RecalcEngine/PowerFxContext for updating that. 
A Context must be created before being passed to the interpreter, and the interpreter then locks the context, preventing further mutation.


src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -45,14 +49,43 @@ public RecalcEngine(PowerFxConfig powerFxConfig = null)
...
+        {
+            _powerFxConfig = powerFxConfig;
+            Formulas = formulas;
+            _enviromentSymbolTable = environmentSymbolTable;
+            _extraFunctions = extraFunctions;
+            
+            // Run rebind for all formulas here
+        }
+
+        public RecalcEngine WithEnvironmentSymbol(IExternalEntity entity)
```
lesaltzm, line 70: Main change: immutable pattern for adding symbols used here.

lesaltzm:
Example snippet of use:
```
var optionSet = new OptionSet("OptionSet", new Dict<string, string>(){
{ "Option1", "option_1" }
{ "Option2", "option_2")
});

var recalcEngine = new RecalcEngine(settings);
recalcEngine = recalcEngine.WithEnvironmentSymbol(optionSet);


var context = new RecordType.AddField("OptionSetField", new OptionSetType(optionSet))

result = recalcEngine.Eval("If(OptionSetField = OptionSet.Option1, \"matches 1\", \"doesn't match 1\")", context,...)
```

src/libraries/Microsoft.PowerFx.Interpreter/Environment/OptionSet.cs
```diff
@@ -0,0 +1,66 @@
...
+            return new SingleSourceDisplayNameProvider(Options);
+        }
+
+        
+        // Refactor the interface to separate out all this Canvas stuff from Power Fx. 
+        public string Name => EntityName;
+
+        public bool IsBooleanValued => false;
+
+        public string RelatedEntityName => throw new NotImplementedException();
```
MikeStall, line 47: >public [](http://example.com/codeflow?start=8&length=6)

nit: These might be able to be explicit to hide them. 
[https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/explicit-interface-implementation](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/explicit-interface-implementation)

src/libraries/Microsoft.PowerFx.Interpreter/Environment/PowerFxConfig.cs
```diff
@@ -0,0 +1,80 @@
...
+using Microsoft.PowerFx.Core;
+using Microsoft.PowerFx.Core.Functions;
+using Microsoft.PowerFx.Core.Types.Enums;
+
+namespace Microsoft.PowerFx
+{
+    /// <summary>
+    /// A container that allows for compiler customization.
+    /// </summary>
+    public sealed class PowerFxConfig
```
MikeStall, line 16: >PowerFxConfig [](http://example.com/codeflow?start=24&length=13)

I like the direction here.

src/libraries/Microsoft.PowerFx.Interpreter/Environment/PowerFxConfig.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        private bool _isLocked;
+        private readonly Dictionary<string, TexlFunction> _extraFunctions;
+
+        internal ImmutableEnvironmentSymbolTable ImmutableEnvironmentSymbolTable;
```
MikeStall, line 21: >ImmutableEnvironmentSymbolTable ImmutableEnvironmentSymbolTable; [](http://example.com/codeflow?start=17&length=64)

The Immutable here is out of place next to all the mutable things. 

Could this just be: 
Dictionary<DName, IExternalEntity> ?

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -43,39 +42,30 @@ public class RecalcEngine : IScope, IPowerFxEngine
...
-            AddFunction(BuiltinFunctionsCore.Boolean_UO);
+            powerFxConfig = powerFxConfig ?? new PowerFxConfig(null, null);
+            AddInterpreterFunctions(powerFxConfig);
+            powerFxConfig.Lock();
+            _powerFxConfig = powerFxConfig;
         }
 
         // Add Builtin functions that aren't yet in the shared library. 
-        private void AddFunction(TexlFunction function)
+        private void AddInterpreterFunctions(PowerFxConfig powerFxConfig)
```
MikeStall, line 52: >AddInterpreterFunctions [](http://example.com/codeflow?start=21&length=23)

Nit: 
Now that this is successfully decoupled from the interpreter...  could this be moved to the bottom of BuiltinFunctionsCore?  (next to where these functions are already defined)

src/libraries/Microsoft.PowerFx.Interpreter/Environment/PowerFxConfig.cs
MikeStall, line 16: >PowerFxConfig [](http://example.com/codeflow?start=24&length=13)

Going forward - would we create one of these around an MsApp? 
We get lots of requests for tools doing intellisense over formulas in an msapp, but they can't create the context today.

MikeStall: Would Document (eventually) use this? Or other backends?
Should we move to Core?

`AddFunction(ReflectionFunction function)` may be an issue (that's interpreter specific), but that could easily be an extension method in interpreter that calls  ` AddFunction(TexlFunction function)`

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
lesaltzm: What I'd like is to  have BuiltinFunctionsCore contain a bunch of smaller sets of functions, and then each host would call like: AddUOFunctions, AddMathFunctions, AddStringFunctions,...that also take into account the Enums required for those functions. I think we'll want to address that as a separate PR, so I'll leave this here for now.

src/libraries/Microsoft.PowerFx.Interpreter/Environment/PowerFxConfig.cs
lesaltzm: Not sure on the future with document specifically, but other back ends would most likely want this, I've moved it Core with some extension methods.

MikeStall (approved):
:shipit:

MikeStall (dismissed):

VamsiModem (approved):

PR #119, opened 2022-01-27
Do arg expansion for FirstN/LastN

lesaltzm:
FirstN and LastN are variadic and should expand args as needed. This fixes an IndexOutOfBounds exception that occurs if they're called with only one arg.


almineev (approved):
Looks good!

MikeStall (approved):
:shipit:

PR #120, opened 2022-01-27
Ensure that new functions don't show up in Canvas apps:

MikeStall:
Notably:
  IndexFunction_CO
  ParseJsonFunction

Just setting IsHidden=true is insufficient to hide from canvas. We can't add them to _library at all.

Still include in the interpreter by default. 

This is blocking integrating into PAClient.


lesaltzm:
To be safe, we should probably pull out the other _CO functions that were added to that list. (Table_CO, Value_CO, and Text_CO)

lesaltzm (approved):

PR #123, opened 2022-01-28
Changed NamedFormulas to IEnumerable

VamsiModem:
Changed NamedFormulas to Enumerable so the consumer can enumerate on the instance, instead of calling GetNamedFormulas().

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormulas.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormulas.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


lesaltzm:
This change feels odd, especially because there's actually two things you could enumerate in that class, the formulas, and the errors. What benefit does this have other than making the caller less clear that it's enumerating over formulas instead of errors?

lesaltzm (changes requested):
This change feels odd, especially because there's actually two things you could enumerate in that class, the formulas, and the errors. What benefit does this have other than making the caller less clear that it's enumerating over formulas instead of errors?

VamsiModem:
> This change feels odd, especially because there's actually two things you could enumerate in that class, the formulas, and the errors. What benefit does this have other than making the caller less clear that it's enumerating over formulas instead of errors?

It makes sense, hence closing this PR.

PR #124, opened 2022-01-28
Rename CustomObject to UntypedObject

bryancroteau-MSFT:
Greg requested that we rename CustomObject to UntypedObject.


MikeStall (approved):
:shipit:

PR #126, opened 2022-01-29
Add new error kind enum value (Internal)

CarlosFigueiraMSFT:
New error kind to be used for errors generated due to internal issues in the Power Fx functions.


MikeStall (approved):
:shipit:

PR #128, opened 2022-01-31
Rename _CO functions to use the _UO suffix

bryancroteau-MSFT:
Cleans up the _CO prefix to use _UO instead.


MikeStall:
>     internal sealed class ParseJsonFunction : BuiltinFunction

nit: I don't think UntypedObject is the right file grouping here.  (for the same reason all the table functions aren't in a single table.cs)
We should group by function. Eg: 
ParseJson.cs and Index.cs 


---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/UntypedObject.cs:13 in 5b80aac. [](commit_id = 5b80aaccb49074628799a35ef1edc333fa012d29, deletion_comment = False)

MikeStall (approved):
:shipit:

PR #129, opened 2022-01-31
Remove runtime type coercions from Value, Text functions

bryancroteau-MSFT:
This change moves the Value and Text functions for UntypedObject instances towards a "type casting" style and away from parsing. Now the Text_UO and Value_UO functions only allow String and Number types on the UO impl. Tests have been updated to reflect the new RuntimeTypeMismatch errors.

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ParseJson.txt
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ParseJson.txt
@@ -6,13 +6,13 @@
 5
 
 >> Value(ParseJson("true"))
-1
+#Error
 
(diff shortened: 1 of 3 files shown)


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ParseJson.txt
```diff
@@ -6,13 +6,13 @@
 5
 
 >> Value(ParseJson("true"))
```
MikeStall, line 8: >Value [](http://example.com/codeflow?start=3&length=5)

can we add a comment here to explain why it's an error, eg:

// Value() doesn't parse and must be a json number.

MikeStall (approved):
:shipit:

PR #130, opened 2022-01-31
[Option Set Cleanup 3/4] Remove more properties from IExternalEntity and IExternalOptionSet

lesaltzm:
This change is to remove more properties from Entity/OptionSet interfaces, to allow creating clean public OptionSet implementation for Interpreter clients. As part of that,  some fields specific to Dataverse have been pulled off of the general option set definition, and checks for Canvas specific logic have been moved out of binder and into Canvas.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -4212,8 +4212,8 @@ namespace Microsoft.PowerFx.Core.Binding
                 if (!(typeLeft.IsPrimitive && typeRight.IsPrimitive) && !(typeLeft.IsPolymorphic && typeRight.IsPolymorphic) && !(typeLeft.IsControl && typeRight.IsControl)
                     && !(typeLeft.IsPolymorphic && typeRight.IsRecord) && !(typeLeft.IsRecord && typeRight.IsPolymorphic))
                 {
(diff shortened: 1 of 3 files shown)


src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4212,8 +4195,8 @@ private void CheckEqualArgTypes(TexlNode left, TexlNode right)
                 if (!(typeLeft.IsPrimitive && typeRight.IsPrimitive) && !(typeLeft.IsPolymorphic && typeRight.IsPolymorphic) && !(typeLeft.IsControl && typeRight.IsControl)
                     && !(typeLeft.IsPolymorphic && typeRight.IsRecord) && !(typeLeft.IsRecord && typeRight.IsPolymorphic))
                 {
-                    var leftTypeDisambiguation = typeLeft.IsOptionSet && typeLeft.OptionSetInfo != null ? $"({typeLeft.OptionSetInfo.Name})" : string.Empty;
-                    var rightTypeDisambiguation = typeRight.IsOptionSet && typeRight.OptionSetInfo != null ? $"({typeRight.OptionSetInfo.Name})" : string.Empty;
+                    var leftTypeDisambiguation = typeLeft.IsOptionSet && typeLeft.OptionSetInfo != null ? $"({typeLeft.OptionSetInfo.EntityName})" : string.Empty;
```
lesaltzm, line 4198: This is for displaying in an error message. Using the entity name which is what Authors would use to reference this in formulas instead of an internal CDS-specific `Name` field that may not show up to the user will result in a better error message, an accidental benefit of this change.

MikeStall (approved):
:shipit:

PR #131, opened 2022-02-01
Add Boolean function

bryancroteau-MSFT:
Adds 2 new functions, Boolean and Boolean_UO, for conversion to boolean.

Valid

Boolean("true")
Boolean("True")
Boolean(" true")

Invalid

Boolean(0)
Boolean("0")
Boolean(true)

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
@@ -36,7 +36,6 @@ namespace Microsoft.PowerFx.Core.Texl
         public static readonly TexlFunction AverageT = _library.Append(new AverageTableFunction());
         public static readonly TexlFunction Blank = _library.Append(new BlankFunction());
         public static readonly TexlFunction Boolean = _library.Append(new BooleanFunction());
-        public static readonly TexlFunction Boolean_UO = _library.Append(new BooleanFunction_UO());
         public static readonly TexlFunction Clock24 = _library.Append(new IsClock24Function());
         public static readonly TexlFunction Char = _library.Append(new CharFunction());
         public static readonly TexlFunction CharT = _library.Append(new CharTFunction());
@@ -187,6 +186,7 @@ namespace Microsoft.PowerFx.Core.Texl
         public static readonly TexlFunction Table_UO = new TableFunction_UO();
         public static readonly TexlFunction Text_UO = new TextFunction_UO();
         public static readonly TexlFunction Value_UO = new ValueFunction_UO();
+        public static readonly TexlFunction Boolean_UO = new BooleanFunction_UO();
 
         public static readonly TexlFunction IsUTCToday = new IsUTCTodayFunction();
         public static readonly TexlFunction UTCNow = new UTCNowFunction();
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
@@ -35,6 +35,8 @@ namespace Microsoft.PowerFx.Core.Texl
         public static readonly TexlFunction Average = _library.Append(new AverageFunction());
         public static readonly TexlFunction AverageT = _library.Append(new AverageTableFunction());
         public static readonly TexlFunction Blank = _library.Append(new BlankFunction());
+        public static readonly TexlFunction Boolean = _library.Append(new BooleanFunction());
+        public static readonly TexlFunction Boolean_UO = _library.Append(new BooleanFunction_UO());
         public static readonly TexlFunction Clock24 = _library.Append(new IsClock24Function());
         public static readonly TexlFunction Char = _library.Append(new CharFunction());
         public static readonly TexlFunction CharT = _library.Append(new CharTFunction());
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
@@ -35,7 +35,6 @@ namespace Microsoft.PowerFx.Core.Texl
         public static readonly TexlFunction Average = _library.Append(new AverageFunction());
         public static readonly TexlFunction AverageT = _library.Append(new AverageTableFunction());
         public static readonly TexlFunction Blank = _library.Append(new BlankFunction());
-        public static readonly TexlFunction Boolean = _library.Append(new BooleanFunction());
         public static readonly TexlFunction Clock24 = _library.Append(new IsClock24Function());
         public static readonly TexlFunction Char = _library.Append(new CharFunction());
         public static readonly TexlFunction CharT = _library.Append(new CharTFunction());
@@ -186,6 +185,7 @@ namespace Microsoft.PowerFx.Core.Texl
         public static readonly TexlFunction Table_UO = new TableFunction_UO();
         public static readonly TexlFunction Text_UO = new TextFunction_UO();
         public static readonly TexlFunction Value_UO = new ValueFunction_UO();
+        public static readonly TexlFunction Boolean = new BooleanFunction();
         public static readonly TexlFunction Boolean_UO = new BooleanFunction_UO();
 
         public static readonly TexlFunction IsUTCToday = new IsUTCTodayFunction();
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Boolean.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Boolean.cs
@@ -8,7 +8,7 @@ using Microsoft.PowerFx.Core.Types;
 
 namespace Microsoft.PowerFx.Core.Texl.Builtins
 {
-    // Boolean(arg:s|n)
+    // Boolean(arg:s)
     // Corresponding Excel and DAX function: Boolean
     internal sealed class BooleanFunction : BuiltinFunction
     {
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Boolean.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Boolean.cs
@@ -1,0 +1,58 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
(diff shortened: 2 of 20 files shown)


src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
```diff
@@ -35,6 +35,8 @@ internal class BuiltinFunctionsCore
         public static readonly TexlFunction Average = _library.Append(new AverageFunction());
         public static readonly TexlFunction AverageT = _library.Append(new AverageTableFunction());
         public static readonly TexlFunction Blank = _library.Append(new BlankFunction());
+        public static readonly TexlFunction Boolean = _library.Append(new BooleanFunction());
+        public static readonly TexlFunction Boolean_UO = _library.Append(new BooleanFunction_UO());
```
MikeStall, line 39: Just confirming if we want these to show up in Canvas yet? 
(see [https://msazure.visualstudio.com/OneAgile/_wiki/wikis/OneAgile.wiki/216333/Merging-into-PAClient](https://msazure.visualstudio.com/OneAgile/_wiki/wikis/OneAgile.wiki/216333/Merging-into-PAClient))

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Value.txt
```diff
@@ -91,4 +91,25 @@ true
 
 >> Value(Date(1900, 3, 1)) - Value(Date(1900, 2, 28))
 /*SqlRunner*/ Compile Error
-1
\ No newline at end of file
+1
+
+>> Boolean("true")
```
MikeStall, line 96: >>> [](http://example.com/codeflow?start=0&length=2)

Boolean(Blank())? 

Boolean(error)?

MikeStall, line 96: >>> [](http://example.com/codeflow?start=0&length=2)

nit: these should be in their own file, boolean.txt

lesaltzm: Also, Boolean(0), Boolean(1), Boolean([ ])


Boolean(true)? Coercion should round trip this, right?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Boolean.cs
```diff
@@ -0,0 +1,58 @@
...
+    // Corresponding Excel and DAX function: Boolean
+    internal sealed class BooleanFunction : BuiltinFunction
+    {
+        public const string BooleanInvariantFunctionName = "Boolean";
+
+        public override bool RequiresErrorContext => true;
+
+        public override bool IsSelfContained => true;
+
+        public override bool SupportsParamCoercion => false;
```
lesaltzm, line 21: Why not allow param coercion?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Boolean.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        public BooleanFunction()
+            : base(BooleanInvariantFunctionName, TexlStrings.AboutBoolean, FunctionCategories.Text, DType.Boolean, 0, 1, 1, DType.String)
```
lesaltzm, line 24: arg type here doesn't match the signature on line 11

lesaltzm: Maybe a checkinvocation overload is needed?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs
```diff
@@ -59,6 +59,16 @@ public static ErrorValue InvalidNumberFormatError(IRContext irContext)
             });
         }
 
+        public static ErrorValue InvalidBooleanFormatError(IRContext irContext)
+        {
+            return new ErrorValue(irContext, new ExpressionError()
+            {
+                Message = "The Boolean could not be parsed",
```
lesaltzm, line 66: Maybe: The value passed to the Boolean function could not be interpreted as a boolean?

Not sure if "parsed" is a verb we expect our customers to be familiar with

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Value.txt
MikeStall: I don't think we need to support the other Boolean overloads  - it can start to get ambiguous with coercions (Boolean("0") ... false or error?) And they have reasonable unambiguous workarounds, like  (x != 0) for number.

bryancroteau-MSFT: Coercion is not enabled for Boolean.

What would Boolean("1") return?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Boolean.cs
bryancroteau-MSFT: See the above question: What would Boolean("1") return?

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
bryancroteau-MSFT: For Boolean, yes, we can add the TS implementation when we upgrade the Nuget package version. I moved the _UO function down to the bottom with the rest. It is no longer being added to the library by default.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Value.txt
bryancroteau-MSFT: Done

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Boolean.cs
bryancroteau-MSFT: Removed |n from the comment. If we want more overloads then we need to design the behavior.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs
bryancroteau-MSFT: Changed

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
MikeStall: " we can add the TS implementation when we upgrade the Nuget package version" 
We should either have that PR ready to go, or keep Boolean off the list until it is.  Else whoever upgrades the version number is going to be hit with this.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Value.txt
MikeStall: Boolean("0") should be an explicit test case here (and include a comment and have it fail) - so to be clear this was a design choice and protect against somebody helpfully flipping the coercion flag in the future.

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
```diff
@@ -53,6 +53,10 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
   </ItemGroup>
+
+  <ItemGroup>
+    <None Remove="ExpressionTestCases\Boolean.txt" />
```
lesaltzm, line 58: Intended?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Value.txt
bryancroteau-MSFT: Added negative tests

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
bryancroteau-MSFT: Removed

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Boolean.cs
bryancroteau-MSFT: Added negative tests to prevent this flag from changing.

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
```diff
@@ -35,6 +35,7 @@ internal class BuiltinFunctionsCore
         public static readonly TexlFunction Average = _library.Append(new AverageFunction());
         public static readonly TexlFunction AverageT = _library.Append(new AverageTableFunction());
         public static readonly TexlFunction Blank = _library.Append(new BlankFunction());
+        public static readonly TexlFunction Boolean = _library.Append(new BooleanFunction());
```
MikeStall, line 38: > Boolean = _library.Append(new BooleanFunction()); [](http://example.com/codeflow?start=43&length=50)

This will require a corresponding change to PAClient.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -50,12 +50,14 @@ public RecalcEngine(PowerFxConfig powerFxConfig = null)
...
             AddFunction(BuiltinFunctionsCore.Value_UO);
+            AddFunction(BuiltinFunctionsCore.Boolean);
+            AddFunction(BuiltinFunctionsCore.Boolean_UO);
         }
 
         // Add Builtin functions that aren't yet in the shared library. 
         private void AddFunction(TexlFunction function)
         {
-            _extraFunctions.Add(function.Name, function);
+            _extraFunctions.Add(function.GetUniqueTexlRuntimeName(), function);
```
MikeStall, line 60: >function.GetUniqueTexlRuntimeName() [](http://example.com/codeflow?start=32&length=35)

This change lets us add overloads for a function.

MikeStall (approved):
:shipit:

PR #132, opened 2022-02-01
Added checks for duplicate variables 

VamsiModem:
Checked for duplicate variables and added them errors.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Localization/Strings.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Localization/Strings.cs
@@ -553,6 +553,7 @@ namespace Microsoft.PowerFx.Core.Localization
         public static ErrorResourceKey OpNotSupportedByClientSuggestionMessage_OpNotSupportedByClient = new ErrorResourceKey("SuggestRemoteExecutionHint_OpNotSupportedByClient");
 
         public static ErrorResourceKey ErrNamedFormula_MissingSemicolon = new ErrorResourceKey("ErrNamedFormula_MissingSemicolon");
+        public static ErrorResourceKey ErrNamedFormula_DuplicateVariable = new ErrorResourceKey("ErrNamedFormula_DuplicateVariable");
 
         // ErrorResourceKey for creating an error from an arbitrary string message. The key resolves to "{0}", meaning
         // that a single string arg can be supplied representing the entire text of the error.
--- a/src/strings/en-US/PowerFxResources.resw
+++ b/src/strings/en-US/PowerFxResources.resw
@@ -1,45 +1,104 @@
-<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<?xml version="1.0" encoding="utf-8"?>
 <root>
+  <!-- 
+    Microsoft ResX Schema 
+    
+    Version 2.0
+    
+    The primary goals of this format is to allow a simple XML format 
+    that is mostly human readable. The generation and parsing of the 
+    various data types are done through the TypeConverter classes 
+    associated with the data types.
+    
+    Example:
+    
+    ... ado.net/XML headers & schema ...
+    <resheader name="resmimetype">text/microsoft-resx</resheader>
+    <resheader name="version">2.0</resheader>
+    <resheader name="reader">System.Resources.ResXResourceReader, System.Windows.Forms, ...</resheader>
+    <resheader name="writer">System.Resources.ResXResourceWriter, System.Windows.Forms, ...</resheader>
+    <data name="Name1"><value>this is my long string</value><comment>this is a comment</comment></data>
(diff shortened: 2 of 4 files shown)


src/tests/Microsoft.PowerFx.Core.Tests/NamedFormulasTests.cs
```diff
@@ -69,5 +70,28 @@ public void GetNamedFormulasTest(string script, string expectedX, string expecte
...
         }
+
+        [Theory]
+        [InlineData("x = 1;x = 2;", "1", 1, "x")]
+        [InlineData("y = 1;y = 2;y = 2;", "1", 2, "y")]
+        [InlineData("y = 1;y = 2;y=3", "1", 3, "y")]
+        public void NamedFormulasWithDuplicateVariablesTest(string script, string expectedX, int expectedErrorCount, string errorToken)
+        {
+            var namedFormula = new NamedFormulas(script);
+            Assert.False(namedFormula.EnsureParsed());
```
MikeStall, line 81: >Assert.False(namedFormula.EnsureParsed()); [](http://example.com/codeflow?start=12&length=42)

Duplicate assigns are syntactically correct, but semantic errors. 

The "pure" way to do this would be:
- the parser succeeds, and returns a list (not dictionary)
- the binder catches the error. 

I don't think it matters now, but it might matter as we get to more IDE scenarios (getting the right error squiggles, correction scenarios)

src/tests/Microsoft.PowerFx.Core.Tests/NamedFormulasTests.cs
```diff
@@ -69,5 +70,28 @@ public void GetNamedFormulasTest(string script, string expectedX, string expecte
...
+            Assert.Equal(errors.Count(), expectedErrorCount);
+
+            var identifierDName = errors.ElementAt(0)?.Tok.As<IdentToken>()?.Name;
+            Assert.NotNull(identifierDName);
+            Assert.Equal(errorToken, identifierDName.Value);
+
+            var formulas = namedFormula.GetNamedFormulas();
+            Assert.NotNull(formulas);
+
+            Assert.Equal(expectedX, formulas.ElementAt(0).formula.Script);
```
MikeStall, line 94: >Assert.Equal(expectedX, formulas.ElementAt(0).formula.Script); [](http://example.com/codeflow?start=12&length=62)

what is this testing?

src/strings/en-US/PowerFxResources.resw
```diff
@@ -1,45 +1,104 @@
-<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
lesaltzm, line 1: What happened here?

src/tests/Microsoft.PowerFx.Core.Tests/NamedFormulasTests.cs
lesaltzm: That's a very good point. The return here should actually be something like List<KeyValuePair<IdentToken , Formula>>, and validate this on the other side, post the correct errors on that ident token

src/tests/Microsoft.PowerFx.Core.Tests/NamedFormulasTests.cs
VamsiModem: I am checking with the first assigned value. 
for example: x=1;x=2; 
in this case the value of x should be 1.

src/strings/en-US/PowerFxResources.resw
VamsiModem: weird, not sure how this got changed, will revert this change.

lesaltzm: still an issue?

src/libraries/Microsoft.PowerFx.Core/Localization/Strings.cs
```diff
@@ -553,6 +553,7 @@ internal static class TexlStrings
         public static ErrorResourceKey OpNotSupportedByClientSuggestionMessage_OpNotSupportedByClient = new ErrorResourceKey("SuggestRemoteExecutionHint_OpNotSupportedByClient");
 
         public static ErrorResourceKey ErrNamedFormula_MissingSemicolon = new ErrorResourceKey("ErrNamedFormula_MissingSemicolon");
+        public static ErrorResourceKey ErrNamedFormula_DuplicateVariable = new ErrorResourceKey("ErrNamedFormula_DuplicateVariable");
```
lesaltzm, line 556: not needed anymore

lesaltzm (approved):

src/strings/en-US/PowerFxResources.resw
VamsiModem: yeah, it's changing that on every commit.

lesaltzm (approved):

PR #133, opened 2022-02-01
Adding support for Find/FindT

almineev:
- Adding support for Find & FindT functions
- Ported all tests from /src/AppMagic/unittest/AppMagic.Functions.Test/FindTests.ts, all except 5 are passing. The 5 remaining are marked as Skip and related to error handling inconsistency with JS over table/row args which will need a wider change to reconsile.


src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Find.cs
```diff
@@ -75,7 +75,7 @@ public override bool CheckInvocation(TexlBinding binding, TexlNode[] args, DType
             var type1 = argTypes[1];
 
             // Arg0 should be either a string or a column of strings.
-            if (type0.IsTable)
+            if (type0.IsTable && type0.Kind != DKind.ObjNull)
```
bryancroteau-MSFT, line 78: Why are we making changes to the front-end for this function?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -243,62 +244,59 @@ internal static partial class Library
...
+                    expandArguments: InsertDefaultValues(outputArgsCount: 3, fillWith: new NumberValue(IRContext.NotInSource(FormulaType.Number), 1)),
+                    replaceBlankValues: ReplaceBlankWith(
+                        new StringValue(IRContext.NotInSource(FormulaType.String), string.Empty),
+                        new StringValue(IRContext.NotInSource(FormulaType.String), string.Empty),
+                        new BlankValue(IRContext.NotInSource(FormulaType.Blank))),
+                    checkRuntimeTypes: ExactSequence(
+                        ExactValueType<StringValue>,
+                        ExactValueType<StringValue>,
+                        ExactValueTypeOrBlank<NumberValue>),
                     checkRuntimeValues: DeferRuntimeValueChecking,
```
bryancroteau-MSFT, line 270: This function accepts a numeric value as input, so we need to check that it is not infinite. There is a runtime value check that does this already.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -243,62 +244,59 @@ internal static partial class Library
...
                     returnBehavior: ReturnBehavior.AlwaysEvaluateAndReturnResult,
```
bryancroteau-MSFT, line 271: This should be "ReturnBlankIfAnyArgIsBlank"

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -352,6 +352,31 @@ public static FormulaValue Right(IRContext irContext, FormulaValue[] args)
             return new StringValue(irContext, source.Value.Substring(source.Value.Length - (int)count.Value, (int)count.Value));
         }
 
+        private static FormulaValue Find(IRContext irContext, FormulaValue[] args)
+        {
+            var findText = (StringValue)args[0];
+            var withinText = (StringValue)args[1];
+
+            NumberValue startIndex;
+            if (args[2] is BlankValue)
```
bryancroteau-MSFT, line 361: This should be taken care of in the wrapper. See the above comment. We avoid doing checks for blank in the body of functions.

bryancroteau-MSFT (changes requested):

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Find.cs
almineev: yep, bad idea. ignored the fact that JS tests for Find operate post-frontend.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
almineev: This would break
>> Find(Blank(), Blank())
1

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
almineev: return blank if arg is blank only works for arg[2] (startIndex)...

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
almineev: There is additional validation that needs to be done on startIndex that can't be done using runtime-value-check infra (logic needs to look at twi args at once). I added infinities check and I think keeping all validation in one place (either runtime-value-check xor func impl) is better than splitting. Open for discussion

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
```diff
@@ -53,6 +53,12 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
   </ItemGroup>
+
+  <ItemGroup>
+    <None Remove="ExpressionTestCases\Concatenate.txt" />
+    <None Remove="ExpressionTestCases\Find.txt" />
```
lesaltzm, line 59: Unintended change to .csproj?

almineev: fixed

lesaltzm (approved):

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
bryancroteau-MSFT: No, because the 2 blanks would be replaced with empty strings.

almineev: I see now, ye. Fixed. 
A bit convoluted though

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
almineev: fixed

bryancroteau-MSFT (approved):

bryancroteau-MSFT (approved):

PR #135, opened 2022-02-03
Make FeatureFlags accessible to hosts

gregli-msft:
At present, a host can't modify the feature flags.  They should be able to.  For example, I'd like to be able to `FeatureFlags.StringInterpolation = true;` when instancing a new engine in the ConeoleREPL sample, to test out this feature without needing to build the libraries.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/FeatureFlags.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/FeatureFlags.cs
@@ -1,18 +1,37 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+using System;
+
 namespace Microsoft.PowerFx
 {
     /// <summary>
     /// Hosts can enable these flags to try out early features. 
     /// Any flags will eventually default to true and then get removed once the feature is finalized. 
+    /// Removing a flag is a breaking change and requires a semver update. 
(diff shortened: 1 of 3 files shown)


src/libraries/Microsoft.PowerFx.Core/FeatureFlags.cs
```diff
@@ -7,7 +7,7 @@ namespace Microsoft.PowerFx
     /// Hosts can enable these flags to try out early features. 
     /// Any flags will eventually default to true and then get removed once the feature is finalized. 
     /// </summary>
-    internal static class FeatureFlags
+    public static class FeatureFlags
```
lesaltzm, line 10: Strongly disagree with this. The existence of FeatureFlags is a hack to enable Canvas not to pick up some changes immediately. It should not be considered part of our public contract, because we want to be able to remove and clean up these flags easily.

Shpakh-MSFT: I agree with McCall. Also this needs to be immutable once it is initialized.

Shpakh-MSFT: We should add some checks so that it can't be changed after initializing once.

lesaltzm (approved):

MikeStall (approved):

PR #136, opened 2022-02-03
Adding string interpolation tests

gregli-msft:
The last five tests currently do not pass with various failures.  The first three should not work but do, the date example returns "! shortdate !" instead of the date as a string, and the last one produces a compile time error.


bryancroteau-MSFT:
This test cases were added already in a previous commit.

PR #137, opened 2022-02-03
Adding index function support for tables

rick-nguyen:
Index function takes (table, row_index) and returns the corresponding row from the given table


rick-nguyen:
Should I rename IndexT to Index? I wasn't sure I saw multiple functions that are named as FunctionT if it's a table function.

See - https://support.microsoft.com/en-us/office/index-function-a5dcf0dd-996d-40a4-a822-b56b061328bd

I wasn't sure how much adding the second form, with a third column argument made sense, since you have to select the tables columns in PA Studio. Any thoughts?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Index.cs
```diff
@@ -0,0 +1,65 @@
...
+    // IndexT(source:*, [n])
+    internal sealed class IndexTFunction : FunctionWithTableInput
+    {
+        public override bool IsSelfContained => true;
+
+        public override bool SupportsParamCoercion => false;
+
+        public IndexTFunction()
+            : base(
+                  "IndexT",
```
lesaltzm, line 23: The T suffix is used when we have functions that have both a scalar and table overload, like Log. 
In those cases, the C# class name, and the JS runtime function get the _T suffix. The string name of the function does not, and we do overload resolution based on the argument types to pick which function is actually called. 

In this case, it's a function that applies to tables, and only tables, so there's no need for a T suffix anywhere.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Index.cs
```diff
@@ -0,0 +1,65 @@
...
+using Microsoft.PowerFx.Core.App.ErrorContainers;
+using Microsoft.PowerFx.Core.Binding;
+using Microsoft.PowerFx.Core.Localization;
+using Microsoft.PowerFx.Core.Syntax.Nodes;
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.Utils;
+
+namespace Microsoft.PowerFx.Core.Texl.Builtins
+{
+    // IndexT(source:*, [n])
```
lesaltzm, line 14: why is the second param optional

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -164,6 +164,13 @@ public static FormulaValue FilterTable(EvalVisitor runner, SymbolContext symbolC
             return new InMemoryTableValue(irContext, rows);
         }
 
+        public static FormulaValue IndexTable(IRContext irContext, FormulaValue[] args)
+        {
+            var arg0 = (TableValue)args[0];
+            var arg1 = (NumberValue)args[1];
+            return arg0.Rows.ElementAt((int)arg1.Value).ToFormulaValue();
```
lesaltzm, line 171: Check range first and return an error if the index is out of bounds

src/strings/en-US/PowerFxResources.resw
```diff
@@ -2354,11 +2354,23 @@
     <value>index</value>
     <comment>function_parameter - Second argument of the Index function - Number type.</comment>
   </data>
+  <data name="AboutIndexT" xml:space="preserve">
```
lesaltzm, line 2357: you can reuse the arg and about strings that Bryan added for Index

src/libraries/Microsoft.PowerFx.Core/Localization/StringResources.cs
```diff
@@ -325,4 +325,4 @@ private static void UpdateErrorResource(string resourceName, string resourceValu
             return errorResources.ToDictionary(kvp => kvp.Key, kvp => ErrorResource.Reassemble(kvp.Value));
         }
     }
-}
+}
```
lesaltzm, line 328: nit: unnecessary edit here

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
```diff
@@ -76,6 +76,7 @@ internal class BuiltinFunctionsCore
         public static readonly TexlFunction Hour = _library.Append(new HourFunction());
         public static readonly TexlFunction If = _library.Append(new IfFunction());
         public static readonly TexlFunction IfError = _library.Append(new IfErrorFunction());
+        public static readonly TexlFunction IndexT = _library.Append(new IndexTFunction());
```
lesaltzm, line 79: If you're adding this to the library, make sure you create and get signoff on a pr in Canvas before checking this in. Otherwise, just add it at the bottom here, not as part of the library

src/libraries/Microsoft.PowerFx.Core/Localization/Strings.cs
```diff
@@ -437,6 +437,9 @@ internal static class TexlStrings
         public static StringGetter AboutIndex = (b) => StringResources.Get("AboutIndex", b);
         public static StringGetter IndexArg1 = (b) => StringResources.Get("IndexArg1", b);
         public static StringGetter IndexArg2 = (b) => StringResources.Get("IndexArg2", b);
+        public static StringGetter AboutIndexT = (b) => StringResources.Get("AboutIndexT", b);
```
lesaltzm, line 440: nit: unused, clean up

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Index.cs
```diff
@@ -0,0 +1,65 @@
...
+                  1,
+                  2,
+                  DType.EmptyTable,
+                  DType.Number)
+        {
+        }
+
+        public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
+        {
+            yield return new[] { TexlStrings.IndexArg1 };
```
lesaltzm, line 37: just the one signature, with 2 args

rick-nguyen: done

src/libraries/Microsoft.PowerFx.Core/Localization/Strings.cs
rick-nguyen: done

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Index.cs
rick-nguyen: removed T suffix

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
rick-nguyen: added range check

src/strings/en-US/PowerFxResources.resw
rick-nguyen: done

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
rick-nguyen: removed from library

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Index.cs
rick-nguyen: fixed

src/libraries/Microsoft.PowerFx.Core/Localization/StringResources.cs
rick-nguyen: fixed

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Index.cs
```diff
@@ -0,0 +1,64 @@
...
+        public override bool SupportsParamCoercion => false;
+
+        public IndexFunction()
+            : base(
+                  "Index",
+                  TexlStrings.AboutIndex,
+                  FunctionCategories.Table,
+                  DType.EmptyRecord,
+                  0,
+                  1,
```
lesaltzm, line 28: This should be minArity:2, maxArity:2.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -164,6 +164,24 @@ public static FormulaValue FilterTable(EvalVisitor runner, SymbolContext symbolC
             return new InMemoryTableValue(irContext, rows);
         }
 
+        public static FormulaValue IndexTable(IRContext irContext, FormulaValue[] args)
+        {
+            var arg0 = (TableValue)args[0];
+            var arg1 = (NumberValue)args[1];
+
+            if ((int)arg1.Value < 1 || (int)arg1.Value >= arg0.Rows.Count())
```
lesaltzm, line 172: do this cast in one place, it's getting reused a few places.

rick-nguyen: done

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Index.cs
rick-nguyen: done

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Index.txt
```diff
@@ -0,0 +1,20 @@
+>> Index([1, 2, 3, 4, 5], 3)
```
MikeStall, line 1: >[1, 2, 3, 4, 5] [](http://example.com/codeflow?start=9&length=15)

nit: 

With this case, it's ambiguous whether the result of 3 was from the array or just taking the 2nd arg. 

Consider a test case like this to be more assertive:
>> Index([10, 20, 30, 40, 50], 3)
{ Value :30}

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Index.txt
```diff
@@ -0,0 +1,20 @@
...
+{Value:3}
```
MikeStall, line 2: The excel function ([https://support.microsoft.com/en-us/office/index-function-a5dcf0dd-996d-40a4-a822-b56b061328bd](https://support.microsoft.com/en-us/office/index-function-a5dcf0dd-996d-40a4-a822-b56b061328bd))  returns  a scalar. 

Should this just return 3, not {value:3} ?

lesaltzm: `[1,2,3]` is just syntactic sugar for `Table({Value: 1}, {Value: 2}, {Value: 3})`. We should return a Record and not have special cases for single-column tables.

MikeStall: Both statements are true, so there's a contradiction to resolve: excel consistency vs. sct semantics ... can we confirm with Greg?

MikeStall: Discussed offline ... this is also consistent with `First(x)` = `Index(x,1)`. 
So this is probably inevitable. But we should not it's different than excel.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -164,6 +164,25 @@ public static FormulaValue FilterTable(EvalVisitor runner, SymbolContext symbolC
...
 
+        public static FormulaValue IndexTable(IRContext irContext, FormulaValue[] args)
+        {
+            var arg0 = (TableValue)args[0];
+            var arg1 = (NumberValue)args[1];
+            int rowIndex = (int)arg1.Value;
+
+            if (rowIndex < 1 || rowIndex >= arg0.Rows.Count())
+            {
+                return new ErrorValue(irContext, new ExpressionError()
```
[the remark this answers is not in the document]
rick-nguyen: Done

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Index.txt
rick-nguyen: Done

lesaltzm (approved):

src/strings/en-US/PowerFxResources.resw
```diff
@@ -2359,6 +2359,7 @@
   </data>
   <data name="AboutIndex_index" xml:space="preserve">
     <value>An integer index.</value>
```
CarlosFigueiraMSFT, line 2361: We can have a better description, something along the lines of `The index of the record to be retrieved`.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Index.cs
```diff
@@ -0,0 +1,64 @@
...
+
+namespace Microsoft.PowerFx.Core.Texl.Builtins
+{
+    // Index(source:*, n)
+    internal sealed class IndexFunction : FunctionWithTableInput
+    {
+        public override bool IsSelfContained => true;
+
+        public override bool SupportsParamCoercion => false;
+
```
CarlosFigueiraMSFT, line 20: Since we will return specific errors from this function in JS, we must declare it as requiring an error context: `public override bool RequiresErrorContext => true;`.

rick-nguyen: done

emhomm4:
Has it been discussed why are we indexing at 1 instead of 0?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -164,6 +164,20 @@ public static FormulaValue FilterTable(EvalVisitor runner, SymbolContext symbolC
...
+            var arg0 = (TableValue)args[0];
+            var arg1 = (NumberValue)args[1];
+            int rowIndex = (int)arg1.Value;
+
+            if (rowIndex < 1 || rowIndex > arg0.Rows.Count())
+            {
+                return CommonErrors.ArgumentOutOfRange(irContext);
+            }
+
+            return arg0.Rows.ElementAtOrDefault(rowIndex - 1).ToFormulaValue();
```
emhomm4, line 178: Why not just start indexing at 0?

rick-nguyen: Hi Emily, it's because to get the first row in Excel's INDEX function we provide an index of 1. An index of 0 would return the entire table as is.

MikeStall:
It's because of Excel compat. :| Excel is 1-based. 

---
In reply to: [1062145288](https://github.com/microsoft/Power-Fx/pull/137#issuecomment-1062145288) [](http://example.com/codeflow?ancestors=1062145288,1062145288)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
CarlosFigueiraMSFT: Power Fx (because of its Excel) starts all indices at 1. Mid("Hello", 2) = "ello", not "llo"; our current way to implement Index is by using `Last(FirstN(<table>, <index>))`, and that is also 1-based

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Index.txt
```diff
@@ -0,0 +1,41 @@
...
+>> Index([1, 2, 3, 4, 5], Blank())
+Blank()
+
+>> Index(Blank(), 2)
+Blank()
+
+>> Index(Sort([1,2,7,3,4,2],1/Value), 1)
+{Value:7}
+
+>> Index([1, 2, 3, 4, 5], 1/0)
```
MikeStall, line 19: Other cases:

- 1st argument is an error, returns error.

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
MikeStall: it looks like it's still in _library

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Index.txt
```diff
@@ -0,0 +1,41 @@
...
+#Error
```
CarlosFigueiraMSFT, line 20: `#Error(Kind=Div0)` - we can now validate the error kinds

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Index.txt
```diff
@@ -0,0 +1,41 @@
...
+
+>> Index(Table({a:"1", b:"101"}, {a:"2"}, {b:"3"}), 1)
+{a:"1",b:"101"}
+
+>> Index(Filter([1,2,3],Value=4),2)
+#Error
```
CarlosFigueiraMSFT, line 26: `#Error(Kind=<the error kind returned by this function>)` - we can now validate the error kinds, and it is good to have it in the tests as the "specification"

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Index.txt
```diff
@@ -0,0 +1,41 @@
+>> Index([10, 20, 30, 40, 50], 3)
+{Value:30}
+
+>> Index([1, 2, 3, 4, 5], 6)
+#Error
```
CarlosFigueiraMSFT, line 5: `#Error(Kind=<the error kind returned by this function>)` - we can now validate the error kinds, and it is good to have it in the tests as the "specification". Same for the next test

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
emhomm4: Got it, thanks!

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Index.txt
rick-nguyen: added

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Index.txt
rick-nguyen: I specified the error kinds but it doesn't appear to make a difference. I can put anything as the kind and the test cases with an expected error will still pass.

CarlosFigueiraMSFT: The test logic is ignoring the kinds, good catch. PR to fix this: https://github.com/microsoft/Power-Fx/pull/211

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
rick-nguyen: fixed

src/strings/en-US/PowerFxResources.resw
MikeStall: nit:   (call out 1-based) "The index of the record to be retrieved, starting at 1"

MikeStall (approved):
:shipit:

PR #141, opened 2022-02-04
Expose functionality for custom functions/types

jcoc611-microsoft:
Changes:

- Expose extensibility of `ValidFormulaValue` for creating custom typed values.
- Expose `recordValue.GetField(string)` for custom functions to access record field values.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
@@ -29,7 +29,7 @@ namespace Microsoft.PowerFx.Core.Public.Values
             return new InMemoryRecordValue(IRContext.NotInSource(type), new List<NamedValue>());
         }
 
-        internal virtual FormulaValue GetField(string name)
+        public virtual FormulaValue GetField(string name)
         {
             return GetField(IRContext.NotInSource(FormulaType.Blank), name);
         }
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/ValidFormulaValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/ValidFormulaValue.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Core.Public.Types;
 
 namespace Microsoft.PowerFx.Core.Public.Values
 {
@@ -15,6 +16,12 @@ namespace Microsoft.PowerFx.Core.Public.Values
     {
         internal ValidFormulaValue(IRContext irContext)
             : base(irContext)
+        {
+        }
+
+        // for custom types
+        protected ValidFormulaValue(FormulaType type)
+            : this(IRContext.NotInSource(type))
         {
         }
     }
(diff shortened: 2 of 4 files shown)


src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
```diff
@@ -29,7 +29,7 @@ public static RecordValue Empty()
             return new InMemoryRecordValue(IRContext.NotInSource(type), new List<NamedValue>());
         }
 
-        internal virtual FormulaValue GetField(string name)
+        public virtual FormulaValue GetField(string name)
```
lesaltzm, line 32: This seems reasonable.

src/libraries/Microsoft.PowerFx.Core/Public/Values/ValidFormulaValue.cs
```diff
@@ -17,5 +18,11 @@ internal ValidFormulaValue(IRContext irContext)
             : base(irContext)
         {
         }
+
+        // for custom types
+        protected ValidFormulaValue(FormulaType type)
```
lesaltzm, line 23: What's the goal of this? Can you provide some examples of why you'd need this added?

jcoc611-microsoft: One simple example would be `PictureValue`, but you can imagine any type that only custom functions or consumers would be able to process. I'm guessing it would be paired with `ExternalType`.

Otherwise, how would one feed PowerFX values that don't match any of the built-in types?

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -275,6 +276,65 @@ public static NumberValue Execute(NumberValue x, NumberValue y)
             }
         }
 
+        [Fact]
+        public void CustomValueFunction()
```
MikeStall, line 280: >CustomValueFunction [](http://example.com/codeflow?start=20&length=19)

This example is very close to the "Host object" scenario in: [https://msazure.visualstudio.com/OneAgile/_queries/edit/12955629/?triage=true](https://msazure.visualstudio.com/OneAgile/_queries/edit/12955629/?triage=true)

lesaltzm: Right, we actually have something already to support these tagged opaque values. If we create fields of type DType.NamedValue, they'll have a tag associated with them, and don't interact with the rest of the system. Functions that expect that type can validate against the tag, so it can be restricted cleanly as well. 

This isn't currently exposed in PowerFx Interpreter, but it's supported in our type system and was implemented a while back for custom pages.

MikeStall: We may be able to solve this scenario already with the UnytpedObject support:
[https://github.com/microsoft/Power-Fx/commit/d53e22b5816e8c2d0722d0f38fd2d42688c62619#diff-3425f96c0ed11735bffc552c807588e38b305797cdbd28459430381533aa8705R78](https://github.com/microsoft/Power-Fx/commit/d53e22b5816e8c2d0722d0f38fd2d42688c62619#diff-3425f96c0ed11735bffc552c807588e38b305797cdbd28459430381533aa8705R78)

MikeStall:
Superseded by: [https://github.com/microsoft/Power-Fx/pull/146](https://github.com/microsoft/Power-Fx/pull/146)

PR #142, opened 2022-02-04
Update Filter function to not ignore predicate errors

CarlosFigueiraMSFT:
The Filter function should not ignore errors that happen when evaluating its predicate. Those errors should be returned in the result as "error records", as announced at https://powerusers.microsoft.com/t5/Error-Handling/3-21122-Filter-behavior-change-with-errors/td-p/1385512. This change updates the Filter implementation, and also the test runner, to use the proper syntax when defining expected values for tables that don't have a *[Value:&lt;type&gt;] schema.


MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -266,6 +266,10 @@ private static bool IsValueTypeErrorOrBlank<T>(FormulaValue val)
                     {
                         include = booleanValue.Value;
                     }
+                    else if (result is ErrorValue errorValue)
```
MikeStall, line 269: This is the key change: if the filter predicate returns an error, then the whole row is a single error object.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/FilterFunctions.txt
```diff
@@ -33,7 +33,7 @@
 4
 
 >> Filter([1,2,0,3,4], 1/ThisRecord.Value >= 0)
-[1,2,3,4]
+Table({Value:1},{Value:2},Microsoft.PowerFx.Core.Public.Values.ErrorValue,{Value:3},{Value:4})
```
MikeStall, line 36: >Microsoft.PowerFx.Core.Public.Values.ErrorValue [](http://example.com/codeflow?start=26&length=47)

this is the change

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
CarlosFigueiraMSFT: Correct. All the other changes are to support the concept of an "error record" in the test expected result.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/FilterFunctions.txt
CarlosFigueiraMSFT: Yes - all the other changes in the tests are related to the change in representation of tables that don't have a single column called 'Value'

MikeStall (approved):
:shipit:

PR #144, opened 2022-02-08
Add hidden StringInterpolation function that does not have a Table ov…

bryancroteau-MSFT:
Add hidden StringInterpolation function that does not have a Table override.

The problem with using Concatenate is that it has a Table override which returns a table. This means that:

$"{Table(...)}"

would have a return type of Table, instead of String. This violates the user's expectation that interpolated strings would behave like string literals and return a value of String type.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/StringInterpolation.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/StringInterpolation.cs
@@ -1,0 +1,80 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+using Microsoft.PowerFx.Core.App.ErrorContainers;
+using Microsoft.PowerFx.Core.Binding;
+using Microsoft.PowerFx.Core.Functions;
+using Microsoft.PowerFx.Core.Localization;
+using Microsoft.PowerFx.Core.Syntax.Nodes;
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.Utils;
+
+namespace Microsoft.PowerFx.Core.Texl.Builtins
+{
+    // StringInterpolation(source1:s, source2:s, ...)
+    // No DAX function, compiler-only, not available for end users, no table support
+    // String interpolations such as $"Hello {"World"}" translate into a call to this function
+    internal sealed class StringInterpolationFunction : BuiltinFunction
+    {
+        public override bool IsSelfContained => true;
+
+        public override bool SupportsParamCoercion => true;
+
+        public StringInterpolationFunction()
+            : base("StringInterpolation", TexlStrings.AboutStringInterpolation, FunctionCategories.Text, DType.String, 0, 1, int.MaxValue)
+        {
+        }
+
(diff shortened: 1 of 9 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -754,6 +754,16 @@ internal static partial class Library
...
             },
+            {
+                BuiltinFunctionsCore.StringInterpolation,
+                StandardErrorHandling<StringValue>(
+                    expandArguments: NoArgExpansion,
+                    replaceBlankValues: ReplaceBlankWithEmptyString,
+                    checkRuntimeTypes: ExactValueType<StringValue>,
+                    checkRuntimeValues: DeferRuntimeValueChecking,
+                    returnBehavior: ReturnBehavior.AlwaysEvaluateAndReturnResult,
+                    targetFunction: Concatenate)
```
MikeStall, line 765: >Concatenate [](http://example.com/codeflow?start=36&length=11)

This is subtle: we introduce a new function for type-checking purposes, but it executes the same way as concatenate.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/StringInterpolation.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            if (arity > 2)
+            {
+                return GetGenericSignatures(arity, TexlStrings.StringInterpolationArg1, TexlStrings.StringInterpolationArg1);
+            }
+
+            return base.GetSignatures(arity);
+        }
+
+        public override bool CheckInvocation(TexlBinding binding, TexlNode[] args, DType[] argTypes, IErrorContainer errors, out DType returnType, out Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
```
MikeStall, line 47: >CheckInvocation [](http://example.com/codeflow?start=29&length=15)

This looks the same as ConcatenateFunction's CheckInvocation. 
Where are we rejecting tables?

lesaltzm: However, this is basically the same as ConcatenateFunction, could you refactor to share an implementation?

bryancroteau-MSFT: Done

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/StringInterpolation.cs
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the MIT license.
+
+namespace Microsoft.PowerFx.Core.Texl.Builtins
+{
+    // StringInterpolation(source1:s, source2:s, ...)
+    // No DAX function, compiler-only, not available for end users, no table support
+    // String interpolations such as $"Hello {"World"}" translate into a call to this function
+    internal sealed class StringInterpolationFunction : ConcatenateFunctionBase
+    {
+        public StringInterpolationFunction()
```
lesaltzm, line 11: To support the JS impl, consider overriding GetUniqueTexlRuntimeName to codegen this as the concatenate function

lesaltzm (approved):

PR #146, opened 2022-02-08
Allow hosts to create custom objects.

MikeStall:


src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValue.cs
```diff
@@ -150,6 +150,13 @@ public static RecordValue NewRecord(object obj, Type type)
             return RecordFromProperties(obj, type);
         }
 
+        public static UntypedObjectValue New(IUntypedObject @object)
```
lesaltzm, line 153: nit, no reason to have this collide with a keyword, rename to untypedobject?

lesaltzm (approved):

PR #149, opened 2022-02-09
Enforce that FormulaTypeJsonConverter is one-way

lesaltzm:
We shouldn’t be getting types into powerfx via JSON generally, we have a cleaner C# api for constructing types. This is only used for notifying clients of the types. Move to throwing an exception instead of allowing reads using this serializer.


Shpakh-MSFT (approved):
:shipit:

PR #150, opened 2022-02-09
Add new TexlNode for string interpolation

bryancroteau-MSFT:
Initial work on a new TexlNode for string interpolation


src/libraries/Microsoft.PowerFx.Core/Logging/StructuralPrint.cs
```diff
@@ -259,6 +259,11 @@ public override LazyList<string> Visit(VariadicOpNode node, Precedence parentPre
             }
         }
 
+        public override LazyList<string> Visit(StrInterpNode node, Precedence context)
+        {
+            throw new NotImplementedException();
```
lesaltzm, line 264: couldn't this be basically, write "$"", then for each child, if it's a string literal write the string literal, if it's not, write `{` node.Accept() `}`, then write the final "?

lesaltzm: basicall, this'll need to be implemented before enabling, it's called from canvas.

src/libraries/Microsoft.PowerFx.Core/Syntax/TexlPretty.cs
```diff
@@ -267,6 +267,11 @@ public override LazyList<string> Visit(VariadicOpNode node, Precedence parentPre
             }
         }
 
+        public override LazyList<string> Visit(StrInterpNode node, Precedence context)
+        {
+            throw new NotImplementedException();
```
lesaltzm, line 272: These I think also need implementations before enabling in canvas

src/libraries/Microsoft.PowerFx.Core/Syntax/Visitors/FindNodeVisitor.cs
```diff
@@ -73,6 +74,37 @@ public override bool PreVisit(VariadicOpNode node)
...
         }
 
+        public override bool PreVisit(StrInterpNode node)
+        {
+            Contracts.AssertValue(node);
+            Contracts.Assert(node.Token.Kind == TokKind.StrInterpStart);
+
+            if (_cursorPosition <= node.Token.Span.Min // Cursor position is before the $"
+                || (node.StrInterpEnd != null && node.StrInterpEnd.Span.Lim <= _cursorPosition) // Cursor is after the close quote.
+                || node.Children.Count() == 0) //// Cursor is inside empty string interpolation.
```
lesaltzm, line 84: nit `////`?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/string.txt
```diff
@@ -350,7 +350,7 @@ Errors: Error 2-2: Unexpected characters. Characters are used in the formula in
 
 >> $"{
 /*SqlRunner*/ Compile Error
-Errors: Error 0-2: Invalid number of arguments: received 0, expected 1 or more.
+""
```
lesaltzm, line 353: Why is this case not erroneous? There's no matching `"`, shouldn't it be a parser error?

bryancroteau-MSFT: Actually no, and functions work the same way. Intellisense still works with partial expressions like this, which is why it's not actually a parse error.

src/libraries/Microsoft.PowerFx.Core/Logging/StructuralPrint.cs
bryancroteau-MSFT: Added an implementation. Do we have tests for this in PowerFx or are they only in PA Client?

src/libraries/Microsoft.PowerFx.Core/FeatureFlags.cs
```diff
@@ -13,6 +13,6 @@ internal static class FeatureFlags
         /// Enable String Interpolation feature. 
         /// Added 12/3/2021.
         /// </summary>
-        public static bool StringInterpolation = false;
+        public static bool StringInterpolation = true;
```
lesaltzm, line 16: Consider taking this flag toggle as a separate change, after updating on the Canvas side.

bryancroteau-MSFT: Removed the feature flag change

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/SuggestionHandlers/StrInterpSuggestionHandler.cs
```diff
@@ -0,0 +1,70 @@
...
+                    IntellisenseHelper.AddSuggestionsForValuePossibilities(intellisenseData, strInterpNode);
+                }
+                else if (strInterpNode.Token.Span.Lim > cursorPos || strInterpNode.StrInterpEnd == null)
+                {
+                    // Handling the erroneous case when user enters a space after $ and cursor is after space.
+                    // Cursor is before the open quote.
+                    // Eg: $ | "
+                    return false;
+                }
+                else
```
lesaltzm, line 44: Does this need to handle `$"{ | ` or would that be an Error node instead of StringInterp?

src/tests/Microsoft.PowerFx.Core.Tests/IntellisenseTests/SuggestTest.cs
```diff
@@ -70,7 +70,7 @@ private class EmptyEnumStore : EnumStore
 
         // DottedNameNodeSuggestionHandler
         [InlineData("{a:{},b:{},c:{}}.|", "a", "b", "c")]
-        [InlineData("$\"Hello { {a:{},b:{},c:{}}.| } World\"", "a", "b", "c")]
+        [InlineData("$\"Hello { First(Table({a:{},b:{},c:{}})).| } World\"", "a", "b", "c")]
```
lesaltzm, line 73: Can we get a few more suggest tests of the error cases in StrInterpSuggestionHandler?

i.e. `$ |`, `$"foo { |`, ...

bryancroteau-MSFT: Added these tests

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/SuggestionHandlers/StrInterpSuggestionHandler.cs
bryancroteau-MSFT: I added tests for this

lesaltzm (approved):
🚢

PR #152, opened 2022-02-10
Update ForAll to return errors if any of the expressions return error

CarlosFigueiraMSFT:
An expression such as `ForAll(<table>, <expression>)` should return an error if one of the invocations of the &lt;expression&gt; return an error (instead of adding the error to the resulting table).

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
@@ -1213,10 +1213,16 @@ namespace Microsoft.PowerFx.Functions
             var arg0 = (TableValue)args[0];
             var arg1 = (LambdaFormulaValue)args[1];
 
-            var rows = LazyForAll(runner, symbolContext, arg0.Rows, arg1);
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1213,10 +1213,16 @@ public static FormulaValue ForAll(EvalVisitor runner, SymbolContext symbolContex
             var arg0 = (TableValue)args[0];
             var arg1 = (LambdaFormulaValue)args[1];
 
-            var rows = LazyForAll(runner, symbolContext, arg0.Rows, arg1);
+            var rows = LazyForAll(runner, symbolContext, arg0.Rows, arg1).ToArray();
```
MikeStall, line 1216: >ToArray [](http://example.com/codeflow?start=74&length=7)

I thought we did not want to do that, specifically because it prevents optimizations. It requires reading the entire table to scan for an error (ie, this ToArray() call). 
 
For example, we should have semantics where First()only needs to look at the top row. So 
First(ForAll(...)) ought to succeed even if item 999 has an error.

PR #153, opened 2022-02-10
Migrate FormatterTests to PowerFx

bryancroteau-MSFT:
Required prerequisite commit before String Interpolation pretty/structural print changes.


lesaltzm (approved):
🚢

PR #155, opened 2022-02-11
Set diagnostic severity on language server publish diagnostics notification

Kcola:
I believe this was omitted by accident. Was causing all diagnostic responses from the language server to have severity 0.


MikeStall (approved):

lesaltzm (approved):
Thanks!

PR #158, opened 2022-02-11
Add flag to remove Single Column Table support

lesaltzm:
The Single Column Table syntax is a common source of difficult-to-resolve user errors. This is a proposal to remove it.

For the same behavior, users could explicitly use ShowColumns(Table, "<ColName>")


MikeStall (changes requested):
discussed offline; just putting a flag to note this is experimental and should not be accidentally checked in.

PR #160, opened 2022-02-14
Remove EnumStore from PowerFx Config constructor

lesaltzm:
The EnumStore is on our list to update  to allow hosts to configure the Enums and Functions available in their scenarios. As part of that, it shouldn't be part of the PowerFxConfig constructor, and instead should be updated using the builder pattern shared by functions and environment symbols.


bryancroteau-MSFT (approved):

MikeStall (approved):
:shipit:

PR #161, opened 2022-02-15
Add Guid and Color FormulaTypes/Values

lesaltzm:


src/libraries/Microsoft.PowerFx.Core/Public/Types/GuidType.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+using Microsoft.PowerFx.Core.Types;
+
+namespace Microsoft.PowerFx.Core.Public.Types
+{
+    public class GuidType : FormulaType
```
MikeStall, line 9: >GuidType [](http://example.com/codeflow?start=17&length=8)

Sql engine has a SqlGuidType - it could be moved onto this!

MikeStall (approved):
:shipit:

Issue #106, opened 2022-01-15
LookUp support


TonyTroeff:
Hello, I am experimenting with this library and encountered the following problem:
This is my code:
```
var expression = "LookUp(Settings, Name = \"Client\")";
var engine = new RecalcEngine();
var parameters = FormulaValue.RecordFromFields(
    new NamedValue("Settings",
        FormulaValue.TableFromRecords(
            FormulaValue.RecordFromFields(new NamedValue("Name", FormulaValue.New("Client"))),
            FormulaValue.RecordFromFields(new NamedValue("Name", FormulaValue.New("Test"))))));
var result = engine.Eval(expression, parameters);
```
I expect that this expression will be evaluated correctly and will return only the first table record (the name of which is exactly "Client"). But instead an exception is thrown. This is its message:
```
Errors: Error 0-33: Invocation of unknown or unsupported function.
Error 17-21: Name isn't valid. This identifier isn't recognized.
Error 22-23: Incompatible types for comparison. These types can't be compared: Error, Text.
```

I found that in your `BuiltInFunctionsCore` file there are `public static readonly` fields for each built-in function but the `LookUp` and `Search` are missing. Maybe this is one reason for this behavior?

lesaltzm:
Right. A bunch of the functions present in Canvas' PowerFx aren't yet implemented in this runtime. We're working to close that gap over time.

MikeStall:
We should give a better error than "Name isn't valid. This identifier isn't recognized.". We should call out the specific identifier. 

@TonyTroeff - First and Filter are implemented, so you could use those to do the Lookup. 
Note that we don't have delegation implemented in the interpreter, so this would be a linear scan.

timnudecision:
related:  ShowColumns is one of those not yet implemented.  it is a priority for my project, should that affect the roadmap.

is there a work-around?

MikeStall:
@timnudecision -  That's correct - it's not implemented yet. can you file a separate issue just for that?    

You may be able to work around it via a ForAll. 
`ShowColumns(_table, "a")`

is the same as: 
`ForAll(_table, { a : a })`

timnudecision:
that workaround works on the REPL and CanvasApp
thanks

MikeStall:
Function is added.

PR #163, opened 2022-02-15
Add Identifier to Error Messages

rick-nguyen:
On using an undefined symbol, we error: 
"Name isn't valid. This identifier isn't recognized."

We should give the actual symbol name in the error message. Like "Name isn't valid. '{identifer}' isn't recognized."

We hit this a lot, especially in other hosts. Here's a customer example: [LookUp support · Issue #106 · microsoft/Power-Fx (github.com)](https://github.com/microsoft/Power-Fx/issues/106) 

Note that it sometimes shows in the formula bar, but not not always:
squiggles may overlap so it can get confusing. 
Or the error is showing outside the formula bar (like in the side panel or solution checker).

C# show in the error message as well as a squiggle. 
1b553fa6-cbf5-4171-a57f-e357072f3b1b


src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -237,7 +237,7 @@ public void CheckBindError()
 
             Assert.False(result.IsSuccess);
             Assert.Single(result.Errors);
-            Assert.StartsWith("Error 2-5: Name isn't valid. This identifier isn't recognized", result.Errors[0].ToString());
+            Assert.StartsWith("Error 2-5: Name isn't valid. 'foo' isn't recognized", result.Errors[0].ToString());
```
MikeStall, line 240: >'foo' isn't recognized [](http://example.com/codeflow?start=60&length=22)

Yes!! This is great

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -3201,7 +3201,7 @@ public override void Visit(ParentNode node)
 
                 if (_nameResolver == null || _nameResolver.CurrentEntity == null)
                 {
-                    _txb.ErrorContainer.Error(node, TexlStrings.ErrInvalidName);
+                    _txb.ErrorContainer.Error(node, TexlStrings.ErrInvalidName, node.Id);
```
MikeStall, line 3204: >node.Id [](http://example.com/codeflow?start=80&length=7)

node.Id is an integer id for indexing into binder info; not a string to show the user. 
The tests added don't cover this case (I'm not sure how we'd actually hit this). 

We may need 2 separate errors, and to leave these cases unchanged.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -3337,6 +3337,7 @@ public override void PostVisit(DottedNameNode node)
                 var nameRhs = node.Right.Name;
 
                 nameRhs = GetLogicalNodeNameAndUpdateDisplayNames(leftType, node.Right);
+                var nodeNameStr = nameRhs.ToString();
```
MikeStall, line 3340: >nodeNameStr [](http://example.com/codeflow?start=20&length=11)

nit: can we rename this var to 'nameRhsString' - just to make it very clear in the cases below that it's tightly associated to nameRhs?

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -3376,22 +3377,22 @@ public override void PostVisit(DottedNameNode node)
...
-                        SetDottedNameError(node, TexlStrings.ErrInvalidName);
+                        SetDottedNameError(node, TexlStrings.ErrInvalidName, nodeNameStr);
                         return;
                     }
                 }
                 else if (leftType.IsAttachment)
                 {
                     // Error: Attachment Type should never be the left hand side of dotted name node
-                    SetDottedNameError(node, TexlStrings.ErrInvalidName);
+                    SetDottedNameError(node, TexlStrings.ErrInvalidName, nodeNameStr);
```
MikeStall, line 3395: >nodeNameStr [](http://example.com/codeflow?start=73&length=11)

Not all of these errors are for looking up the RHS. 

The basic rule is that if we lookup string x and can't find it, then issue an error for string x.  Line 3380 is a good case of that. 

In the other cases (like this one) - we should probably leave unchanged.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
rick-nguyen: I put in the old error as "ErrInvalidIdentifier", and changed these to use the old error again and leave them basically unchanged. Thoughts?

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
rick-nguyen: Done

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -3337,6 +3337,7 @@ public override void PostVisit(DottedNameNode node)
                 var nameRhs = node.Right.Name;
 
                 nameRhs = GetLogicalNodeNameAndUpdateDisplayNames(leftType, node.Right);
+                var nameRhsString = nameRhs.ToString();
```
lesaltzm, line 3340: same here, nameRhs.Value

lesaltzm: Also, consider using node.Right.Name.Value instead of the logicalName that you get back from this function. The error message will be better if it corresponds to what's in the expression rather than the names under the hood

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -3490,7 +3491,7 @@ public override void PostVisit(DottedNameNode node)
 
                         if (!leftType.ToRecord().TryGetType(property.InvariantName, out typeRhs))
                         {
-                            SetDottedNameError(node, TexlStrings.ErrInvalidName);
+                            SetDottedNameError(node, TexlStrings.ErrInvalidName, firstNodeLhs.Ident.Name.ToString());
```
lesaltzm, line 3494: the invalid name here is Property.InvariantName

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -5306,7 +5307,7 @@ private void PreVisitBottomUp(CallNode node, int argCountVisited, Scope scopeNew
                     }
                     else
                     {
-                        _txb.ErrorContainer.Error(node, TexlStrings.ErrInvalidName);
+                        _txb.ErrorContainer.Error(node, TexlStrings.ErrInvalidName, node.Head.Name.ToString());
```
lesaltzm, line 5310: .Value here as well

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -290,7 +290,7 @@ public void CheckBindErrorWithParseExpression()
             Assert.False(result.IsSuccess);
             Assert.Null(result.Expression);
             Assert.Single(result.Errors);
-            Assert.StartsWith("Error 2-5: Name isn't valid. This identifier isn't recognized", result.Errors[0].ToString());
+            Assert.StartsWith("Error 2-5: Name isn't valid. 'foo' isn't recognized", result.Errors[0].ToString());
```
lesaltzm, line 293: Let's get more tests covering Dotted Names, lambdas, etc..

lesaltzm (changes requested):

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -2897,7 +2897,7 @@ public override void Visit(FirstNameNode node)
 
                 if (!haveNameResolver || !_nameResolver.Lookup(node.Ident.Name, out lookupInfo, preferences: lookupPrefs))
                 {
-                    _txb.ErrorContainer.Error(node, TexlStrings.ErrInvalidName);
+                    _txb.ErrorContainer.Error(node, TexlStrings.ErrInvalidName, node.Ident.Name.ToString());
```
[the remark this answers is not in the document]
rick-nguyen: Done

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
rick-nguyen: Done

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
rick-nguyen: Fixed

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
rick-nguyen: Done

lesaltzm (approved):

lesaltzm (approved):

PR #169, opened 2022-02-17
Improve Suggest api performance

Robulane:
This change caches the enum objects and re-uses them for Suggestion api.

The following test takes 320 seconds in master and only 12s with the proposed change

        [Fact]
        public void TestSuggestdf()
        {
            var eng = new RecalcEngine();
            for (int i = 0; i < 1000; i++)
            {
                eng.Suggest("Abs", null, 3);
            }
        }

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Types/Enums/EnumStore.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Types/Enums/EnumStore.cs
@@ -435,7 +435,7 @@ namespace Microsoft.PowerFx.Core.Types.Enums
 
             return CollectionUtils.EnsureInstanceCreated(ref _enumsUsedPowerApps, () =>
             {
-                var list = ImmutableList.Create<Tuple<DName, DName, DType>>();
+                var list = ImmutableList.CreateBuilder<Tuple<DName, DName, DType>>();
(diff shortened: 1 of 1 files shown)


src/libraries/Microsoft.PowerFx.Core/Types/Enums/EnumStore.cs
```diff
@@ -430,21 +433,27 @@ internal void ResetCustomEnums()
                 return RegenerateEnumTypes();
             });
 
-            foreach (var enumSpec in EnumDict)
+            return CollectionUtils.EnsureInstanceCreated(ref _enumsUsedPowerApps, () =>
             {
-                Contracts.Assert(DName.IsValidDName(enumSpec.Key));
+                var list = ImmutableList.CreateBuilder<Tuple<DName, DName, DType>>();
```
MikeStall, line 438: >ImmutableList [](http://example.com/codeflow?start=27&length=13)

curious - why Immutable list instead of just List? 
Either way, we're returning an IEnumerable, so it will be readonly.

Robulane: Just to make it 100% readonly. I've seen this: ((List<string>)result).Mutate()

PR #174, opened 2022-02-23
Fix thread safety issue when engine was giving results for a wrong expression

Robulane:
There is a thread safety bug in PowerFX. This bug is exposed by added EvalInMultipleThreads test.

This PR fixes this thread safety bug.


MikeStall:
Fx is not thread safe, so there may be other issues.

PR #179, opened 2022-02-24
Split ExpressionTestCases into folders for standard and sql diff tests

bryancroteau-MSFT:
Splits the test files in the ExpressionTestCases folder into 2 subfolders, one which contains files that contain Sql test differences.


MikeStall:
Is the idea that /Standard is common to all engines , and then Sql, JS, etc can have specific dirs?
What does it mean that /string.txt is in both dirs?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Concat.txt
```diff
@@ -45,13 +45,13 @@
     ),
     Value)
 
-"ACADBCBD"
```
MikeStall, line 48: ?

PR #180, opened 2022-02-25
Add date/time support for min/max functions

emhomm4:
Addresses a bug where previously min/max on datetimes would return the result in seconds. This fix ensures the result preserves the Date/Time/DateTime type provided. If a Date and DateTime is compared, Date will be coerced to DateTime. For all other mixing of types, the result will still be coerced to numeric.

minmax1

minmax2

minmax5

minmax4

Mixed types:
image
image
image
image
image


src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MinMax.cs
```diff
@@ -1,21 +1,74 @@
...
+        public override bool CheckInvocation(TexlNode[] args, DType[] argTypes, IErrorContainer errors, out DType returnType, out Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
+        {
+            Contracts.AssertValue(args);
+            Contracts.AssertValue(argTypes);
+            Contracts.Assert(args.Length == argTypes.Length);
+            Contracts.AssertValue(errors);
+            Contracts.Assert(MinArity <= args.Length && args.Length <= MaxArity);
+
+            var fArgsValid = true;
+            returnType = argTypes[0];
```
lesaltzm, line 38: how does this handle mixed time/date/number?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/TestRunner.cs
```diff
@@ -311,6 +311,18 @@ internal static void TestToString(FormulaValue result, StringBuilder sb)
 
                 sb.Append('}');
             }
+            else if (result is TimeValue tv)
+            {
+                sb.Append(tv.Value.ToString());
+            }
+            else if (result is DateValue d)
+            {
+                sb.Append(d.Value.ToString(@"MM/dd/yyyy hh:mm:ss tt", System.Globalization.CultureInfo.InvariantCulture));
```
CarlosFigueiraMSFT, line 320: [nit] can we use 24h time for the tests (format specifier `HH`)? That's more "global" than using am/pm.

emhomm4: The build system formats it in AM/PM, so it was done this way to pass the unit tests.

CarlosFigueiraMSFT: IIRC this is used to compare with the expected value defined in the .txt test files, so if we update the value there (e.g., mathfuncs.txt, line 149), it should work.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MinMax.cs
emhomm4: Hmm that's a good question. Should it maybe be an error to compare numbers with datetimes?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/TestRunner.cs
CarlosFigueiraMSFT: That's ok, we can look at fixing it later...

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MinMax.cs
emhomm4: I've updated my PR to resolve this. Any time there are mixed types, we will coerce to numeric (see final screenshot).

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/mathfuncs.txt
```diff
@@ -119,9 +119,30 @@ Blank()
 >> 2^-2
 0.25
 
+>> Max(1,2)
```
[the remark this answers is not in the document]
emhomm4: Moved them to their own file.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MinMax.cs
```diff
@@ -1,21 +1,81 @@
...
+            Contracts.Assert(args.Length == argTypes.Length);
+            Contracts.AssertValue(errors);
+            Contracts.Assert(MinArity <= args.Length && args.Length <= MaxArity);
+
+            nodeToCoercedTypeMap = null;
+            var fArgsValid = true;
+            returnType = argTypes[0];
+
+            // If there are elements of different types OR if the elements are not a Date/Time/DateTime, coerce to numeric.
+            if (!Array.TrueForAll(argTypes, element => element == argTypes[0]) || !Array.Exists(argTypes, element => element == DType.Date || element == DType.DateTime || element == DType.Time))
```
CarlosFigueiraMSFT, line 43: It would be good to add some tests for expressions that don't compile. Things like

    Min({a:1}, {a:2}, {a:3})
    Max(1, 2, {a:3})
    Min([1,2,3], [1])

This is very similar to the previous one, but it should work, so we should add as a test:

    >> Min([1, 2, 3], 4) // This uses the `Min(<table>, <expression>)` overload, and the expression is a constant
    4

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/MinMax.txt
```diff
@@ -0,0 +1,50 @@
...
+#Error
+
+>> Max(1, 2, 3, Blank(), 2, 3)
+3
+
+>> Max(Table({Name:"John Doe", DOB:Date(1985,3,5)},{Name:"Jane Roe", DOB:Date(1986,10,19)}),DOB)
+Date(1986,10,19)
+
+>> Max(Table({Name:"John Doe", DOB:Date(1985,3,5), Thing: "1"},{Name:"Jane Roe", DOB:Date(1986,10,19), Thing: "2"}),Thing)
+2
```
CarlosFigueiraMSFT, line 50: More tests:

- Single arguments (this is valid): `Max(1)`, `Min(Date(2022,3,8))`, ...
- Min/max of very large tables: `Max(Sequence(33333, 10) => 33342`, `Min(Sequence(10000), DateAdd(Date(2000,1,1), -Value, TimeUnit.Days) => Date(1992,8,5)`, ...
- Errors in multi-parameter overloads: `Min(1, 3, Sqrt(-1), -3) => #Error(Kind=Numeric)`, `Max(Time(3,4,5), Time(2,1/0,8)) => #Error(Kind=Div0)`, ...

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/MinMax.txt
```diff
@@ -0,0 +1,50 @@
...
+Date(1985,3,5)
+
+>> Min(-100, If(true, -1000, 200))
+-1000
+
+>> Min([1, 2, 3, 4, 5], Value)
+1
+
+>> Min([1, 2, 1/0, 4, 5], Value)
+#Error
```
CarlosFigueiraMSFT, line 32: `#Error(Kind=Div0)` - we can test for the error kind now

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -58,6 +58,8 @@ public virtual FormulaValue GetResult(IRContext irContext)
         private class MinAgg : IAggregator
         {
             protected double _minValue = double.MaxValue;
+            protected DateTime _minValueDT = DateTime.MaxValue;
+            protected TimeSpan _minValueT = TimeSpan.MaxValue;
             protected int _count;
 
             public void Apply(FormulaValue value)
```
MikeStall, line 65: >FormulaValue [](http://example.com/codeflow?start=30&length=12)

Are we guaranteed that all values are of the same type?

Ie, if somebody does: Min(number, datetime, timespan), it would either be a binding error or the IR would have insert coercions?

Ie, is it possible to call this same aggregator instance with a different types?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/MinMax.txt
emhomm4: Added

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/MinMax.txt
emhomm4: Added these. Thanks!

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MinMax.cs
```diff
@@ -1,21 +1,81 @@
...
+            Contracts.AssertValue(argTypes);
+            Contracts.Assert(args.Length == argTypes.Length);
+            Contracts.AssertValue(errors);
+            Contracts.Assert(MinArity <= args.Length && args.Length <= MaxArity);
+
+            nodeToCoercedTypeMap = null;
+            var fArgsValid = true;
+            returnType = argTypes[0];
+
+            // If there are elements of different types OR if the elements are not a Date/Time/DateTime, coerce to numeric.
```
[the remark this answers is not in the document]
emhomm4: This is handled now, and I added tests for this case.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MinMax.cs
emhomm4: Thanks, tests added!

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
emhomm4: The only mixed type that is allowed is Date and DateTime, and in that case dates are coerced to DateTime.

Whenever a number is mixed with another type, the other type is coerced to numeric (where possible).

MikeStall: In that case, I think we should consider having multiple derived classes per type,  MinDoubleAgg, MinDateTimeAgg, MinTimeSpanAgg.

It's more classes, but each class is trivial and only has a single field and no switching in Apply/GetResult.  Instead, the switching is done once when we create the class instance, and the rest is polymorphism.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MinMaxTable.cs
```diff
@@ -23,5 +28,46 @@ public MinMaxTableFunction(bool isMin)
...
+
+        public override bool CheckInvocation(TexlNode[] args, DType[] argTypes, IErrorContainer errors, out DType returnType, out Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
+        {
+            Contracts.AssertValue(args);
+            Contracts.AssertValue(argTypes);
+            Contracts.Assert(args.Length == argTypes.Length);
+            Contracts.AssertValue(errors);
+            Contracts.Assert(MinArity <= args.Length && args.Length <= MaxArity);
+
+            //var fValid = CheckInvocation(args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
```
MikeStall, line 40: >//var fValid [](http://example.com/codeflow?start=12&length=12)

nit: dead code

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
```diff
@@ -54,6 +54,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <None Remove="ExpressionTestCases\MinMax.txt" />
     <None Remove="ExpressionTestCases\StringInterpolate.txt" />
   </ItemGroup>
```
MikeStall: nit - we can remove this. (covered by line 47).

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
emhomm4: I created the new Agg classes

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -193,25 +331,101 @@ public static FormulaValue SumTable(EvalVisitor runner, SymbolContext symbolCont
...
+                return RunAggregator(new MaxDateAgg(), irContext, args);
+            }
+            else if (irContext.ResultType == FormulaType.Time)
+            {
+                return RunAggregator(new MaxTimeAgg(), irContext, args);
+            }
+            else
+            {
+                return new BlankValue(irContext);
+            }
```
MikeStall, line 353: Is it possible to hit this? (will binder have prevented).

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -193,25 +331,101 @@ public static FormulaValue SumTable(EvalVisitor runner, SymbolContext symbolCont
...
+                return RunAggregator(new MaxDateTimeAgg(), runner, symbolContext, irContext, args);
+            }
+            else if (irContext.ResultType == FormulaType.Date)
+            {
+                return RunAggregator(new MaxDateAgg(), runner, symbolContext, irContext, args);
+            }
+            else if (irContext.ResultType == FormulaType.Time)
+            {
+                return RunAggregator(new MaxTimeAgg(), runner, symbolContext, irContext, args);
+            }
```
MikeStall, line 374: Can we do it like this (shares the call to RunAggregator)):

            IAggregator agg;
            if (irContext.ResultType == FormulaType.Number)
            {
                agg = new MaxNumberAgg();
            }
            else if (irContext.ResultType == FormulaType.DateTime)
            {
                agg = new MaxDateTimeAgg(), irContext, args);
            }
            else if (irContext.ResultType == FormulaType.Date)
            {
                agg = new MaxDateAgg();
            }
            else if (irContext.ResultType == FormulaType.Time)
            {
                agg = new MaxTimeAgg();
            }
   else
            {
                return new BlankValue(irContext);
            }
   return RunAggregator(agg, irContext, args);

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
emhomm4: removed

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MinMaxTable.cs
```diff
@@ -23,5 +28,44 @@ public MinMaxTableFunction(bool isMin)
...
+            Contracts.AssertValue(errors);
+            Contracts.Assert(MinArity <= args.Length && args.Length <= MaxArity);
+
+            // The return type will be the result type of the expression in the 2nd argument
+            var fArgsValid = true;
+            returnType = argTypes[1];
+            nodeToCoercedTypeMap = null;
+
+            // Coerce everything except date/times to numeric.
+            for (var i = 1; i < argTypes.Length; i++)
```
lesaltzm, line 46: there should be only 2 args to this function, I don't think the loop is needed

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -193,25 +331,88 @@ public static FormulaValue SumTable(EvalVisitor runner, SymbolContext symbolCont
...
-            return RunAggregator(new MinAgg(), irContext, args);
+            var agg = GetMinMaxAggType(irContext, true);
+
+            if (agg != null)
+            {
+                return RunAggregator(agg, irContext, args);
+            }
+            else
+            {
+                return new BlankValue(irContext);
```
lesaltzm, line 372: When can this happen? Should it be an unreachable error?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MinMaxTable.cs
emhomm4: removed

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MinMax.cs
```diff
@@ -1,21 +1,92 @@
...
+
+            if (!fArgsValid)
+            {
+                nodeToCoercedTypeMap = null;
+            }
+
+            return fArgsValid;
+        }
+
+        public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
```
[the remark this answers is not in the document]
emhomm4: added

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
emhomm4: This would happen if irContext.ResultType wasn't a type we expected. But I don't think it should ever actually happen.

lesaltzm: so, in those cases, why return blank instead of an unreachablecode error?

lesaltzm: return CommonErrors.UnreachableCodeError(irContext);

emhomm4: Added

lesaltzm (approved):

PR #182, opened 2022-02-25
Add back a customizable EnumStore as a stopgap awaiting refactor

lesaltzm:
We intend to refactor the enum store, so that hosts shouldn’t be configuring an enum stores directly, rather, they should be picking which sets of functions they want to use, and we’ll bring in the appropriate enums as well. 

For hosts using enum store currently, this adds back support for customizing it until that work is completed. This constructor should be removed when that is finished.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
@@ -40,6 +40,20 @@ namespace Microsoft.PowerFx.Core
             EnumStore = new EnumStore();
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PowerFxConfig"/> class.
+        /// Stopgap constructor until Enum Store is refactored. Do not rely on, this will be removed. 
+        /// </summary>
+        internal PowerFxConfig(CultureInfo cultureInfo = null, EnumStore enumStore = null)
+        {
+            CultureInfo = cultureInfo ?? CultureInfo.CurrentCulture;
+            _isLocked = false;
+            _extraFunctions = new Dictionary<string, TexlFunction>();
+            _environmentSymbols = new Dictionary<DName, IExternalEntity>();
+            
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -40,6 +40,20 @@ public PowerFxConfig(CultureInfo cultureInfo = null)
             EnumStore = new EnumStore();
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PowerFxConfig"/> class.
+        /// Stopgap constructor until Enum Store is refactored. Do not rely on, this will be removed. 
+        /// </summary>
+        internal PowerFxConfig(CultureInfo cultureInfo = null, EnumStore enumStore = null)
```
MikeStall, line 47: >PowerFxConfig [](http://example.com/codeflow?start=17&length=13)

nit: this can lead to ambiguous ctor overloads. 
The challenge is we want a public ctor and EnumStore is internal. 

If this is stopgap, then we could we use a a static builder method to avoid the ambiguity?

lesaltzm: Yeah, good point.

src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/ScenarioMutation.cs
```diff
@@ -19,7 +19,7 @@ public class ScenarioMutation
         [Fact]
         public void MutabilityTest()
         {
-            var config = new PowerFxConfig();
+            var config = new PowerFxConfig(null);
```
MikeStall, line 22: >null [](http://example.com/codeflow?start=43&length=4)

nit, do we need this change? 
(ie, is this a breaking change)

MikeStall (dismissed):
:shipit:

MikeStall (approved):
:shipit:

PR #184, opened 2022-02-28
[Cleanup] Fix inconsistent .txt file test cases

bryancroteau-MSFT:
Fixes a few tests that have inconsistent formatting.

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/string.txt
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/string.txt
@@ -83,7 +83,7 @@ true
 >> Mid("foo", 0)
 #Error
 
-Mid(Text(1/0, "#.000"), 1)
+>> Mid(Text(1/0, "#.000"), 1)
 #Error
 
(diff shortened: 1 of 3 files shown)


lesaltzm (approved):

MikeStall:
I've rolled these fixes into https://github.com/microsoft/Power-Fx/pull/190 
That also adds some enforcement in TestRunner to catch it automatically.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/string.txt
```diff
@@ -83,7 +83,7 @@ true
 >> Mid("foo", 0)
 #Error
 
-Mid(Text(1/0, "#.000"), 1)
+>> Mid(Text(1/0, "#.000"), 1)
```
MikeStall, line 86: It turns out this test case actually hangs the interpreter!
Filed https://github.com/microsoft/Power-Fx/issues/195

MikeStall: That probably also explains why this PR is failing the CI

PR #187, opened 2022-03-02
Allow CountIf and LookUp to coerce yes/no to Boolean

emhomm4:
Filter coerces yes/no columns in Dataverse to Boolean, but CountIf and LookUp do not do the same thing. This PR remedies that.

Test App:
image

Formulas used:
image

image

image

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Lookup.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Lookup.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using Microsoft.PowerFx.Core.App.ErrorContainers;
 using Microsoft.PowerFx.Core.Binding;
 using Microsoft.PowerFx.Core.Entities;
+using Microsoft.PowerFx.Core.Errors;
 using Microsoft.PowerFx.Core.Functions;
 using Microsoft.PowerFx.Core.Functions.Delegation;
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Lookup.cs
```diff
@@ -47,6 +48,20 @@ public override bool CheckInvocation(TexlBinding binding, TexlNode[] args, DType
             // The return type is dictated by the last argument (projection) if one exists. Otherwise it's based on first argument (source).
             returnType = args.Length == 2 ? argTypes[0].ToRecord() : argTypes[2];
 
+            // Ensure that the arg at index 1 is boolean or can be coersed.
+            if (CheckType(args[1], argTypes[1], DType.Boolean, DefaultErrorContainer, out var matchedWithCoercion) || DType.Boolean.Accepts(argTypes[1]))
```
lesaltzm, line 52: Doesn't CheckType cover the Dtype.Boolean.Accepts case?

emhomm4: You are correct, I removed it.

emhomm4:
> When integrating in PA-Client, double check that this works with CDS delegation. Additionally, not as part of this, but we need to bring binding-only tests like TexlTests.cs over into Power-Fx

Confirmed.

PR #190, opened 2022-03-03
Update .txt runner for maintainability 

MikeStall:
Move away from a single .txt having all the engine results in it. This is problematic because other engines live in other repos and they need an easy way (in their repo) to mark what they do and don't support. 

Allow #Override and #Disable directives. Files in the engine's specific repo can then disable specific tests or entire files. 
Apply this to the interpreter. 

If we go this direction, we should probably completely remove the /* engine */ annotations in favor of the #Override directive.


MikeStall:
> #OVERRIDE: Blank.txt

The files in this directory merge with the files specified in the new Override directive. 

In practice, that's using it to skip individual tests. So effectively, this new directory captures all the interpreter bugs / missing features relative to the baseline. 



---
Refers to: src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterExpressionTestCases/BlankOverride.txt:1 in eaef280. [](commit_id = eaef2805229db7192e4cffd75fb5cc8c1a64a2aa, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Blank.txt
```diff
@@ -91,7 +91,6 @@ true
 0
 
 >> Blank() = Blank()
-/*InterpreterRunner*/ #Skip
```
MikeStall, line 94: Engine-specific issues are not called out via a dedicate engine-specific .txt file in that engine's repo.  See BlankOverride.txt for example here.

src/tests/Microsoft.PowerFx.Core.Tests/TestRunnerTests/File2.txt
```diff
@@ -0,0 +1,11 @@
+﻿
+>> MultiInput
+  secondline
+Result
+
+>> Engines2
+/* Engine1*/ ER1
+/* Engine2*/ ER2
```
MikeStall, line 8: If we fully embrace this shift, we could get rid of the /*engine*/ support altogether.

MikeStall:
>             if (testCase.GetExpected(nameof(InterpreterRunner)) == "#Skip")

I think some of the logic in here, like #SKIP,  should also be in the base runner and shared with all backends. 

---
Refers to: src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs:57 in eaef280. [](commit_id = eaef2805229db7192e4cffd75fb5cc8c1a64a2aa, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/FindT.txt
```diff
@@ -128,29 +128,24 @@ Table({Result:3},{Result:Microsoft.PowerFx.Core.Public.Values.ErrorValue},{Resul
 Table({Result:3},{Result:Microsoft.PowerFx.Core.Public.Values.ErrorValue},{Result:1})
 
 >> Find("c", ["abc", Error({Kind: "Validation"}), "cde"])
```
CarlosFigueiraMSFT, line 130: I know this is not part of your changes, but since you're changing this file, can you replace all instances of `Error({Kind: "Validation"})` with `Error({Kind:ErrorKind.Validation})`? They are not useful today: they don't work in the JS interpreter, and it's only "working" in the C# interpreter because the Error function is not implemented (so it's currently returning an error). Similar for the new 'FindT.txt' file in the new folder.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/TestCase.cs
```diff
@@ -8,7 +8,10 @@
...
 namespace Microsoft.PowerFx.Core.Tests
 {
-    internal class TestCase
+    /// <summary>
+    ///  Describe a test case from the .txt file.
+    /// </summary>
+    public class TestCase
     {
         // Formula string to run 
         public string Input;
```
lesaltzm, line 17: Should we be maintaining this Dict indexed by runner name? It seems like the goal of this PR is to have just one expected result loaded at a time, based on what's been overriden.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/TestRunner.cs
```diff
@@ -52,6 +64,24 @@ public void AddFile(IEnumerable<string> files)
             }
         }
 
+        // Directive should start with #, end in : like "#SETUP:"
+        // Returns true if matched; false if not. Throws on error.
+        private static bool ParseDirective(string line, string directive, ref string param)
```
CarlosFigueiraMSFT, line 69: Suggestion: rename to `TryParseDirective`

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/TestCase.cs
MikeStall: exactly. I mentioned that in the other comments. I wanted to get validation of the direction before ripping it out wholesale.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/TestRunner.cs
```diff
@@ -67,12 +97,45 @@ public void AddFile(string thisFile)
...
+                        DisabledFiles.Add(fileDisable);
+
+                        // Will remove all cases in this file.
+                        // Can apply to multiple files. 
+                        var countRemoved = _tests.RemoveAll(test => string.Equals(Path.GetFileName(test.SourceFile), fileDisable, StringComparison.OrdinalIgnoreCase));                        
+                    }
+                    else if (ParseDirective(line, "#SETUP:", ref fileSetup) ||
+                      ParseDirective(line, "#OVERRIDE:", ref fileOveride))
+                    {
+                    }
```
ian-legler, line 127: Nit: add a no-op comment inside the braces to clarify nothing is needed

src/tests/Microsoft.PowerFx.Core.Tests/TestRunnerTests/FileOverride.txt
```diff
@@ -0,0 +1,5 @@
+﻿#OVERRIDE: File1.txt
+
+// Override
+>> input1
+override_input1
```
ian-legler, line 5: Nit: override_result1

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/TestCase.cs
MikeStall: We first need to migrate other backends away from /*engine*/ and then we can remove this.

MikeStall:
> // Test String interpolation feature. 

splitting into separate file to make it easy for backends that don't support it yet to disable the whole file. 

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/StringInterpolate.txt:1 in 84f4091. [](commit_id = 84f4091cd18c785c5f9c85118f0ed923516fbe12, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/AndOrCases.txt
```diff
@@ -60,10 +60,10 @@ false
 >> Or(false,false, true)
 true
 
--- Test short circuiting. Don't hit args in unevaled params.
+// -- Test short circuiting. Don't hit args in unevaled params.
```
MikeStall, line 63: >// [](http://example.com/codeflow?start=0&length=2)

test parser now throws an error to catch these syntax errors in the .txt.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/arithmetic.txt
```diff
@@ -100,15 +100,9 @@ true
 >> 5 < 5
 false
 
->> 5 < 7
```
MikeStall, line 103: .txt parser now throws an error on duplicate cases within a single file (this is repeated on line 94)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/BaseRunner.cs
```diff
@@ -4,21 +4,133 @@
...
+        /// Returns (Pass,Fail,Skip) and a status message.
+        /// </summary>
+        /// <param name="test">test case to run.</param>
+        /// <returns>status from running.</returns>
+        public async Task<(TestResult, string)> RunAsync(TestCase testCase)
+        {
+            var result = TestResult.Fail;
+            string message = null;
+
+            var t = new Thread(() =>
```
MikeStall, line 40: >var t = new Thread(() => [](http://example.com/codeflow?start=12&length=24)

XUnit doesn't support a timeout attribute. 
[https://github.com/xunit/xunit/issues/217](https://github.com/xunit/xunit/issues/217)

So offload to a thread on a timer.

ian-legler (approved):

CarlosFigueiraMSFT (approved):
:shipit:

PR #194, opened 2022-03-04
Remove restriction for output behavior properties in components

CarlosFigueiraMSFT:
Currently the binder forbids output behavior properties in components from being called from outside the component; this is a scenario that we want to enable, so we are removing this restriction. Notice that there is no way to create such property today, so this won't have any immediate impact.


src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4362,12 +4362,6 @@ private static bool IsValidAccessToScopedProperty(IExternalControl lhsControl, I
                    (currentControl.IsComponentControl ||
                    (currentControl.TopParentOrSelf is IExternalControl { IsComponentControl: false })))
                 {
-                    // Behavior property is blocked from outside the component.
-                    if (isBehaviorOnly)
-                    {
-                        return false;
```
jack-work, line 4368: I this something we should (or even can) write a test for?

jack-work: In this codebase, anyway

jack-work (approved):
Pretty straightforward, just wondering about tests.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
CarlosFigueiraMSFT: Not at this point.

Shpakh-MSFT (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4361,12 +4361,6 @@ private static bool IsValidAccessToScopedProperty(IExternalControl lhsControl, I
                    (currentControl.IsComponentControl ||
```
lesaltzm, line 4361: the `isBehaviorOnly` argument to this function is now unused, consider removing it?

lesaltzm (approved):

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
CarlosFigueiraMSFT: Good idea, done

PR #196, opened 2022-03-04
Added offset to named formulas

VamsiModem:
1. Created a new class for named formulas
2. Every named formula will have an offset, and min & lim in span will reset to 0.


src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -123,10 +124,11 @@ private ParseFormulasResult ParseFormulas(string script)
                             // Parse expression
                             var result = ParseExpr(Precedence.None);
 
-                            namedFormulas.Add(new KeyValuePair<IdentToken, TexlNode>(thisIdentifier.As<IdentToken>(), result));
+                            namedFormulas.Add(new NamedFormula(thisIdentifier.As<IdentToken>(), result, offset));
                         }
 
                         _curs.TokMove();
+                        offset = _curs.CurrentCharIndex;
```
lesaltzm, line 131: consider setting the offset at the call to ParseExpr, that way the token starting at 0 is the initial token of the actual rule.

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -1387,7 +1389,7 @@ private bool EatTid(TokKind tid)
 
         // Returns the current token if it's of the given kind and moves to the next token.
         // If the token is not the right kind, reports an error, leaves the token, and returns null.
-        private Token TokEat(TokKind tid)
+        private Token TokEat(TokKind tid, int offset = 0)
```
lesaltzm, line 1392: unused param

src/libraries/Microsoft.PowerFx.Core/Parser/TokenCursor.cs
```diff
@@ -103,11 +115,21 @@ public Token TokMove()
...
+            
             if (_currentTokenId != TokKind.Eof)
             {
                 MoveTo(_currentTokenIndex + 1);
             }
 
+            var tokenLength = tok.Span.Lim - tok.Span.Min;
+            _currentCharIndex += tokenLength;
+
+            if (_flags.HasFlag(TexlParser.Flags.NamedFormulas))
```
lesaltzm, line 127: I don't think you need to base this off of the flag. You could add a function like `ResetOffset()` that does this and only call it when you move the offset in the outer parser.

src/tests/Microsoft.PowerFx.Core.Tests/NamedFormulasTests.cs
```diff
@@ -19,7 +19,9 @@ public void NamedFormulaTest(string script)
         }
 
         [Theory]
-        [InlineData("x=1;y=2;", 2)]
+        [InlineData("x=    \n\n\t2;colStyles=Table({\n\tkey:\"colorPrimary\", \n\tvalue:\"Red\"\n})", 2)]
+
+        //[InlineData("x=158558 +\t                  \n\n289885;y=\n\n     \t2;", 2)]
```
lesaltzm, line 24: tests with the offset please

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
lesaltzm: that way it'll match non-namedformula tokens within a single expression

src/libraries/Microsoft.PowerFx.Core/Parser/TokenCursor.cs
VamsiModem: we still have to clone the spans for named formulas with new min & lim, I have tried creating them, if there is a difference in offset and min, but tests for Format() are failing as whitespaces created by format are ignored, will check with Bryan. please suggest if you have a different approach in mind.

src/tests/Microsoft.PowerFx.Core.Tests/NamedFormulasTests.cs
MikeStall, line 24: >// [](http://example.com/codeflow?start=8&length=2)

remove comment?

src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormula.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Microsoft.PowerFx.Core.Syntax
+{
+    internal class NamedFormula
+    {
+        public IdentToken Name { get; }
+
+        public TexlNode Node { get; }
+
+        public int Offset { get; }
```
MikeStall, line 15: >public int Offset { get; } [](http://example.com/codeflow?start=8&length=26)

Can we comment what this field means?

MikeStall (changes requested):
Need more discussion about adding Offset and how that fits in e2e

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -87,13 +87,14 @@ public static ParseFormulasResult ParseFormulasScript(string script, ILanguageSe
 
         private ParseFormulasResult ParseFormulas(string script)
         {
-            var namedFormulas = new List<KeyValuePair<IdentToken, TexlNode>>();
+            var namedFormulas = new List<NamedFormula>();
             ParseTrivia();
-
+            var offset = _curs.CurrentCharIndex;
```
MikeStall, line 92: >var offset = _curs.CurrentCharIndex; [](http://example.com/codeflow?start=12&length=36)

can we delete this?
offset is only used in the nested scope.

VamsiModem: yeah, it can  be deleted.

src/libraries/Microsoft.PowerFx.Core/Parser/TokenCursor.cs
```diff
@@ -103,11 +121,26 @@ public Token TokMove()
...
+            var tokenLength = tok.Span.Lim - tok.Span.Min;
+            _currentCharIndex += tokenLength;
+
+            if (_shouldUseOffset)
+            {
+                tok = tok.Clone(new Localization.Span(_offset, _offset + tokenLength));
+                if (tok.Kind == TokKind.Semicolon)
+                {
+                    _shouldUseOffset = false;
+                }
```
MikeStall, line 139: We shouldn't do this check.  Semi could appear within a formula, and not necessarily indicate the end of a formula.

PR #202, opened 2022-03-07
Implement Error function, pass scope parameter to IfError, validate error kind

CarlosFigueiraMSFT:
There are many tests that are using the Error function, but it is not currently implemented. This change implements that function, and updates IfError to pass the error information to the handler lambdas. To test this we're also allowing the error kind to be passed in .txt test files in the `#error` directive (if present the test will validate that they match the kind of the error result).


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/BaseRunner.cs
```diff
@@ -17,7 +20,7 @@ public abstract class BaseRunner
         /// Maximum time to run test - this catches potential hangs in the engine. 
         /// Any test should easily run in under 1s. 
         /// </summary>
-        public static TimeSpan Timeout = TimeSpan.FromSeconds(20);
+        public static TimeSpan Timeout = TimeSpan.FromSeconds(20000);
```
MikeStall, line 23: >20000 [](http://example.com/codeflow?start=62&length=5)

?
This will blow the timeout. 
Is the goal here to help under a debugger?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Error.cs
```diff
@@ -83,7 +83,7 @@ public override bool CheckInvocation(TexlNode[] args, DType[] argTypes, IErrorCo
             }
 
             var argumentKindType = names.First(tn => tn.Name == requiredKindField.Name).Type;
-            if (!argumentKindType.CoercesTo(requiredKindField.Type))
+            if (argumentKindType.Kind != requiredKindField.Type.Kind)
```
MikeStall, line 86: >if (argumentKindType.Kind != requiredKindField.Type.Kind) [](http://example.com/codeflow?start=12&length=57)

what's the significance of this change?

CarlosFigueiraMSFT: Currently we allow people to write something like `Error({Kind: "a string"})`. This is wrong - the value of Kind should only be a member of the EnumKind enumeration (which is numeric).

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -834,16 +847,7 @@ internal static partial class Library
...
-                StandardErrorHandling<FormulaValue>(
-                    expandArguments: NoArgExpansion,
-                    replaceBlankValues: DoNotReplaceBlank,
-                    checkRuntimeTypes: DeferRuntimeTypeChecking,
-                    checkRuntimeValues: DeferRuntimeValueChecking,
-                    returnBehavior: ReturnBehavior.AlwaysEvaluateAndReturnResult,
-                    targetFunction: Table)
-            },
+            { BuiltinFunctionsCore.Table, Table },
             {
```
CarlosFigueiraMSFT, line 851: The Table function can receive error parameters (and existing tests were expecting this). This change fixes it.

MikeStall:
>                 // Timeout!!!

For better "under the debugger support", we could add this here:
if (Debugger.IsAttached) retry


---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/BaseRunner.cs:58 in b6a36de. [](commit_id = b6a36debed331a67186ef90ea3e52ff14042ea98, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/BaseRunner.cs
```diff
@@ -42,7 +47,7 @@ public async Task<(TestResult, string)> RunAsync(TestCase testCase)
                 (result, message) = RunAsync2(testCase).Result;
             });
             t.Start();
-            var success = t.Join(Timeout);
+            var success = t.Join(System.Diagnostics.Debugger.IsAttached ? TimeSpan.FromSeconds(20000) : Timeout);
```
MikeStall, line 50: >System.Diagnostics.Debugger.IsAttached ? TimeSpan.FromSeconds(20000) [](http://example.com/codeflow?start=33&length=68)

The looping is really the proper way to do this. 
With this approach, this always disabled the timeout if they run under debugger (even if not hitting breakpoints).

CarlosFigueiraMSFT: Pushed new changes

MikeStall (approved):
:shipit:

PR #203, opened 2022-03-07
Add new method to TexlBinding for generating transient Concatenate

bryancroteau-MSFT:
This change makes the binder aware of generated call nodes. This will allow querying of call nodes in the future, for example in JsTranslator.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -68,6 +68,10 @@ namespace Microsoft.PowerFx.Core.Binding
         // Maps Ids to Info, where the Id is an index in the array.
         private readonly object[] _infoMap;
 
+        // Call nodes which do not appear in the source code
+        // For example, $"" (interpolated string) maps to a call to Concatenate
+        private readonly CallNode[] _compilerGeneratedCallNodes;
+
         private readonly IDictionary<int, IList<FirstNameInfo>> _lambdaParams;
 
         // Whether a node is stateful or has side effects or is contextual or is constant.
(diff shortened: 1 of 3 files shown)


src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -68,6 +68,10 @@ internal sealed partial class TexlBinding
         // Maps Ids to Info, where the Id is an index in the array.
         private readonly object[] _infoMap;
 
+        // Call nodes which do not appear in the source code
+        // For example, $"" (interpolated string) maps to a call to Concatenate
+        private readonly CallNode[] _compilerGeneratedCallNodes;
```
MikeStall, line 73: >_compilerGeneratedCallNodes [](http://example.com/codeflow?start=36&length=27)

How will we use this?

bryancroteau-MSFT: JsTranslator often queries the call info.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -2097,6 +2110,27 @@ private void SetInfo(CallNode node, CallInfo info, bool markIfAsync = true)
...
+            var callNode = new CallNode(
+                ref id,
+                primaryToken: ident,
+                sourceList: node.SourceList,
+                head: new Identifier(ident),
+                headNode: null,
+                new ListNode(ref listNodeId, tok: node.Token, args: node.CloneChildren(ref minChildId, node.GetCompleteSpan()), delimiters: null, sourceList: node.SourceList),
+                node.StrInterpEnd);
+            _compilerGeneratedCallNodes[node.Id] = callNode;
+            SetInfo(callNode, new CallInfo(callNode));
```
MikeStall, line 2130: >SetInfo(callNode, new CallInfo(callNode)); [](http://example.com/codeflow?start=12&length=42)

This line is the critical line that was missing before

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
MikeStall: This is effectively additional information similar to _infoMap, but infoMap's length is fixed + locked after the parser. whereas these new nodes are created in the binder after the parse, so we need a separate place to store them.

StrInterpolation needs both a CallInfo and CallNode.

MikeStall (approved):
:shipit:

PR #205, opened 2022-03-08
Add more error testing.

MikeStall:
Fix .txt test parser hole

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/BaseRunner.cs
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/BaseRunner.cs
@@ -65,6 +65,26 @@ namespace Microsoft.PowerFx.Core.Tests
             }
         }
 
+        // If we override Error values, then for a test case:
+        //   >> X
+        //   #error
+        // also check that:
+        //   >> IsError(x)
+        //   true
+        private async Task<(TestResult, string)> RunErrorCaseAsync(TestCase testCase)
+        {
+            var case2 = new TestCase
+            {
+                SetupHandlerName = testCase.SetupHandlerName,
+                SourceLine = testCase.SourceLine,
+                SourceFile = testCase.SourceFile,
+                Input = $"IsError({testCase.Input})",
+                Expected = "true"
+            };
+
+            return await RunAsync2(case2);
+        }
+
         private async Task<(TestResult, string)> RunAsync2(TestCase testCase)
         {
             string actualStr;
@@ -167,7 +187,8 @@ namespace Microsoft.PowerFx.Core.Tests
                 } 
                 else if (IsError(result))
                 {
-                    return (TestResult.Pass, null);
+                    // If they override IsError, then do additional checks. 
+                    return await RunErrorCaseAsync(testCase);                    
                 }
 
                 // If the actual result is not an error, we'll fail with a mismatch below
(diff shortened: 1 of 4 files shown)


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/BaseRunner.cs
```diff
@@ -167,7 +187,8 @@ private async Task<(TestResult, string)> RunAsync2(TestCase testCase)
                 } 
                 else if (IsError(result))
                 {
-                    return (TestResult.Pass, null);
+                    // If they override IsError, then do additional checks. 
+                    return await RunErrorCaseAsync(testCase);
```
CarlosFigueiraMSFT, line 191: This is nice; can we also call the RunErrorCaseAsync for the cases above where the tests passed (lines 155, 164, 176)?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/TestRunner.cs
```diff
@@ -17,13 +17,14 @@ namespace Microsoft.PowerFx.Core.Tests
...
 
-        // Mapping of a Test's key to the test case in _test list.
+        // Mapping of a Test's key to the test case in Test list.
         // Used for when we need to update the test. 
         private readonly Dictionary<string, TestCase> _keyToTests = new Dictionary<string, TestCase>(StringComparer.Ordinal);
 
-        public IEnumerable<TestCase> Tests => _tests;
+        // Expose tests so that host can manipulate list directly. 
+        // Also populate this by calling various Add*() functions to parse. 
+        public List<TestCase> Tests { get; set; } = new List<TestCase>();
```
CarlosFigueiraMSFT, line 27: [minor] Expose as `IList<TestCase>`

src/tests/Microsoft.PowerFx.Core.Tests/TestRunnerTests/TestRunnerTests.cs
```diff
@@ -257,15 +266,22 @@ public async Task TestErrorOverride()
             // Test override BaseRunner.IsError
             var runner = new MockErrorRunner
             {
-                _hook = (expr, setup) => FormulaValue.New(1),
-                _isError = (value) => true
+                _hook = (expr, setup) => 
+                    expr switch {
+                        "1" => FormulaValue.New(1),
+                        "IsError(1)" => FormulaValue.New(true),
```
CarlosFigueiraMSFT, line 272: It would be good to also assert that `IsError(1)` was called in this test (i.e., initialize a variable as false, set it to true in this case, then assert that it's true at the end)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/BaseRunner.cs
MikeStall: We could, but it's less valuable there because we're already returning an ErrorValue. 

It's so special in this case because Sql is returning a blank instead of error; and we want to make sure that it's still generating errors within the expression.

src/tests/Microsoft.PowerFx.Core.Tests/TestRunnerTests/TestRunnerTests.cs
MikeStall: I'll add a 2nd test that returns IsError() false, and verifies a failure. That should cover this.

PR #209, opened 2022-03-09
Update local package version number

CarlosFigueiraMSFT:
Update local package version number to a version higher than the public one, so that it can be used with dependencies that require a version higher than one of the public releases.


MikeStall (approved):
:shipit:

PR #211, opened 2022-03-11
Fix error kind matching in test cases

CarlosFigueiraMSFT:
The matching for the kind property of errors is not working properly; this fixes the regex that is used to validate it (and adds test cases to guard against regressions)


MikeStall (approved):
:shipit:

PR #210, opened 2022-03-10
Coercing only OptionSetValues to boolean for CountIf and LookUp

emhomm4:
Re-do of this PR: 
https://github.com/microsoft/Power-Fx/pull/187

But instead only coercing OptionSetValue.

PR #212, opened 2022-03-11
LookUp function

jorisdg:
Implement LookUp function in C# interpreter


src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
```diff
@@ -54,6 +54,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <None Remove="ExpressionTestCases\Lookup.txt" />
     <None Remove="ExpressionTestCases\StringInterpolate.txt" />
   </ItemGroup>
```
MikeStall, line 59: nit: we should be able to remove all of this, covered by line 48

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -13,6 +13,53 @@ namespace Microsoft.PowerFx.Functions
...
+                {
+                    var childContext = symbolContext.WithScopeValues(row.Value);
+                    var value = arg2.Eval(runner, childContext);
+
+                    if (value is NumberValue number)
+                    {
+                        value = FiniteChecker(irContext, 0, number);
+                    }
+
+                    if (value is ErrorValue error)
```
lesaltzm, line 51: nit, this condition doesn't actually do anything. You can just `return value`

lesaltzm (approved):

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -13,6 +13,48 @@ namespace Microsoft.PowerFx.Functions
...
+
+            if (args.Length > 3)
+            {
+                return new ErrorValue(irContext, new ExpressionError()
+                {
+                    Message = "LookUp() only supports two predicates",
+                    Span = irContext.SourceContext,
+                    Kind = ErrorKind.Validation
+                });
+            }
```
CarlosFigueiraMSFT, line 31: Can this codepath ever be hit? The definition of the LookUp function specifies that the max arity is 3

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -13,6 +13,48 @@ namespace Microsoft.PowerFx.Functions
...
+
+            var row = LazyFilter(runner, symbolContext, arg0.Rows, arg1).FirstOrDefault();
+
+            if (row != null)
+            {
+                if (args.Length == 2)
+                {
+                    return row?.ToFormulaValue() ?? new BlankValue(irContext);
```
CarlosFigueiraMSFT, line 39: >row? [](http://example.com/codeflow?start=27&length=4)

No need for the `?` here since you already know that `row` is not null (4 lines above)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -13,6 +13,48 @@ namespace Microsoft.PowerFx.Functions
...
+                }
+                else
+                {
+                    var childContext = symbolContext.WithScopeValues(row.Value);
+                    var value = arg2.Eval(runner, childContext);
+
+                    if (value is NumberValue number)
+                    {
+                        value = FiniteChecker(irContext, 0, number);
```
CarlosFigueiraMSFT, line 48: >value = FiniteChecker(irContext, 0, number); [](http://example.com/codeflow?start=24&length=44)

It would be good to add a test that exercises this code path (something like `LookUp([100, 200, 300], Value > 250, Power(100, Value))`

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Lookup.txt
```diff
@@ -0,0 +1,70 @@
...
+// $$$ Corner cases like  negative indices, N > length, doubles instead of int, empty. 
+
+>> LookUp([1,2,3], ThisRecord.Value > 3)
+Blank()
+
+>> LookUp([1,0,3,4], 1/ThisRecord.Value >= 0)
+{Value:1}
+
+>> LookUp([0,3,4], 1/ThisRecord.Value >= 0)
+#Error
```
CarlosFigueiraMSFT, line 64: `#Error(Kind=Div0)` - we should validate the error kinds

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Lookup.txt
```diff
@@ -0,0 +1,70 @@
...
+
+>> LookUp([1,2,3], ThisRecord.Value >= 2, Value)
+2
+
+>> LookUp([1,2,3], ThisRecord.Value > 2).Value
+3
+
+// $$$ Corner cases like  negative indices, N > length, doubles instead of int, empty. 
+
+>> LookUp([1,2,3], ThisRecord.Value > 3)
```
CarlosFigueiraMSFT, line 57: More interesting tests:
- Lookups in very long lists: `LookUp(Sequence(30000), Value > 25000, Value * 2) ==> 50002` or `LookUp(Sequence(30000), Value > 35000, Value * 2) ==> Blank()`

MikeStall: Re long lists - 30000 is a big number ... would 300 suffice? I don't want to just burn cpu cycles.

CarlosFigueiraMSFT: We've had bugs in the past that start showing up with tables around 10000 elements... I would hope that we could deal with ~30k+ items in memory fairly quickly

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
jorisdg: Fair question... I guess it's defensive code, honestly I just mimicked what was done in the FilterTable() method.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Lookup.txt
jorisdg: @CarlosFigueiraMSFT these are interesting tests, but they are not specific to the lookup function? They are tests to check if errors fall through correctly, which the implementation of the function doesn't deal with directly. The ones returning an error at runtime can be properly caught (like Div0), but those that throw an exception during the formula check (for example, where the types are incompatible like Len("string" > 3) cannot be "caught" with IfError or checking the tests result for Error (with or without a specific Kind).
Happy to take your examples (and ensure to use runtime errors we can check for) but they aren't tests for LookUp code specifically? Let me know what you'd like me to do.

jorisdg: Correction - it does return the errorvalue so that should be checked, but any one error is sufficient as a test for lookup()

CarlosFigueiraMSFT: @jorisdg I see those tests not only as "tests" per se (i.e., preventing regressions), but also as a de facto specification for the function. Also, in this C# implementation we have a wrapper that deals with errors, but it's possible that in another language / platform they won't have such a thing, so having those tests that clearly define what the behavior of the function wrt errors is useful to help the language be consistent across different platforms.

jorisdg: The long list tests seem to average between 250 and 300 ms each when running on my machine (Xeon with 3.6Ghz cores).

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Lookup.txt
```diff
@@ -0,0 +1,104 @@
...
+Blank()
+
+>> LookUp(Blank(), Blank())
+Blank()
+
+>> LookUp(Blank(), Blank(), "constant")
+Blank()
+
+>> LookUp([100, 200, 300], Value > 250, Power(100, Value))
+#Error(Kind=ArgumentOutOfRange)
```
CarlosFigueiraMSFT, line 89: >ArgumentOutOfRange [](http://example.com/codeflow?start=12&length=18)

We don't have ErrorKind.ArgumentOutOfRange defined... Ì would expect this test to be failing. Is it passing?

jorisdg: Yes, it passes.

jorisdg: My repo here is out of date with the changes Carlos made for the error kind checking. That is also why the verification of my last PR update failed on the test framework.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Lookup.txt
emhomm4: Invalid argument tests for each argument would be good as well.
`LookUp("123", true)`
`LookUp(["test"], 123)`

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
```diff
@@ -53,10 +53,6 @@
     </Content>
   </ItemGroup>
 
-  <ItemGroup>
```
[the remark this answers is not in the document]
lesaltzm: yes

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Lookup.txt
jorisdg: @emhomm4 I would agree, but those throw exceptions and don't return Error() values so I don't see a way to define a test like that?

lesaltzm: Exceptions? Or compile time errors?
For compile time errors, see the last few tests in 
https://github.com/microsoft/Power-Fx/blob/9938c06b7d2c4d01732d5f535d4314fbb0f98d11/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/CountFunctions.txt

lesaltzm: If it's exceptions, than that's concerning, and might need emily's #210 to go in first

jorisdg: Compile time errors. Thanks for the PR link, I will try that out!!

jorisdg: Errors added.

emhomm4 (approved):

lesaltzm (approved):

MikeStall (approved):
:shipit:

PR #214, opened 2022-03-14
Improved marshalling from .net to fx

MikeStall:
Index() supports direct lookup rather than linear scan. 
Count() support direct count. 

Convert some linear scans to direct lookups, like InMemoryRecordValue.GetField, RecordScope.Resolve

Improve marshalling from .Net objects to Power Fx:
1. lazy field marshaling 
2. strong typing 
3. control over propertyInfo marshaling (renames, skips) 
4. basic recursion (truncate at depth N rather than crash)

IUntypedObject gives us 1 or 2, but not both. 

This uses a "provider" pattern (ASP uses this a lot): 
- MarshalerProvider: goes from System.Type --> Marshaler. This is static, so it ensures the marshaling is based purely on the static type and runtime values.
- Marshaler: goes from System.Object --> FormulaValue. We have many marshallers (primitive, identity, object, array, custom ,etc).

See new test case files. 

Updated FormulaValue.New* updates.  (Breaking changes here due to renames) 
- use TypeMarshallerCache
- rationalize table/record creation


src/tests/Microsoft.PowerFx.Interpreter.Tests/MarshalTests.cs
```diff
@@ -0,0 +1,285 @@
...
+using System.Linq;
+using System.Reflection;
+using Microsoft.PowerFx.Core.Public.Types;
+using Microsoft.PowerFx.Core.Public.Values;
+using Xunit;
+
+namespace Microsoft.PowerFx.Tests
+{
+    // Test marshaling between C# objectrs and Power Fx values. 
+    public class MarshalTests
```
MikeStall, line 14: >public class MarshalTests [](http://example.com/codeflow?start=4&length=25)

Key thing are these new test cases

src/libraries/Microsoft.PowerFx.Core/Public/Marshal/IdentityMarshalerProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// </summary>
+    public class IdentityMarshalerProvider : ITypeMashalerProvider
+    {
+        public ITypeMarshaler New(Type type, TypeMarshallerCache cache, int maxDepth)
+        {
+            if (typeof(FormulaValue).IsAssignableFrom(type))
+            {
+                return new IdentityMarshaler
+                {
+                    Type = null // $$$ Fix this
```
MikeStall, line 21: >Type = null // $$$ Fix this  [](http://example.com/codeflow?start=20&length=28)

this may be another switch case.  Similar to what we do in ForumulaValue.New

src/libraries/Microsoft.PowerFx.Core/Public/Marshal/ObjectMarshalerProvider.cs
```diff
@@ -0,0 +1,112 @@
...
+    /// Marshal .Net classes (with fields). This supports strong typing and lazy marshalling. 
+    /// </summary>
+    public class ObjectMarshalerProvider : ITypeMashalerProvider
+    {
+        // $$$ To marshal an object, we need:
+        //  - fxName,
+        //  - .Net type (for getting the marhsaler)
+        //  - getter for runtime value
+        // Customization point 
+        public Func<PropertyInfo, string> PropertyMapperFunc = (propInfo) => propInfo.Name;
```
MikeStall, line 27: >PropertyMapperFunc [](http://example.com/codeflow?start=42&length=18)

This gives some customization ability. 
If they need radically more control (say allow fields instead of property infos), they can always write their own provider and return a ObjectMarshaler  directly.

src/libraries/Microsoft.PowerFx.Core/Microsoft.PowerFx.Core.csproj
```diff
@@ -21,6 +21,8 @@
 		<Description>The core library for all Microsoft Power Fx Compiler targets</Description>
 		<Summary>The core library for all Microsoft Power Fx Compiler targets</Summary>
 		<GeneratePackageOnBuild>True</GeneratePackageOnBuild>
+		<RunAnalyzersDuringBuild>true</RunAnalyzersDuringBuild>
+		<EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
```
lesaltzm, line 25: This property is in Directory.Build.props, covers every project. I don't think it needs to be repeated here.

src/libraries/Microsoft.PowerFx.Core/Public/Values/ObjectRecordValue.cs
```diff
@@ -0,0 +1,45 @@
...
+using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Core.Public.Types;
+using static Microsoft.PowerFx.ObjectMarshalerProvider;
+
+namespace Microsoft.PowerFx.Core.Public.Values
+{
+    // $$$ Move to interpreter? Or move Source property to RecordValue?
+    public class ObjectRecordValue : RecordValue
+    {
+        public object Source { get; private set; }
```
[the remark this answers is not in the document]
lesaltzm: Not all recordvalues will have an `object` typed source, so I'd say no. Keep this as a member of ObjectRecordValue.

src/libraries/Microsoft.PowerFx.Core/Public/Marshal/ITypeMashalerProvider.cs
```diff
@@ -0,0 +1,32 @@
...
+        // Pass in a cache for aggregate types that need to marshal sub types. 
+        public ITypeMarshaler New(Type type, TypeMarshallerCache cache, int maxDepth);
+    }
+
+    /// <summary>
+    /// A mrashaller for a given System.Type. 
+    /// </summary>
+    public interface ITypeMarshaler
+    {
+        public FormulaType Type { get; }
```
MikeStall, line 26: >public FormulaType Type { get; } [](http://example.com/codeflow?start=8&length=32)

subtle win here - the Type is known regardless of the runtime object (which is only passed to Marshal).

src/libraries/Microsoft.PowerFx.Core/Public/Marshal/ITypeMashalerProvider.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    /// <summary>
+    /// Handles marshaling a given type. Invoked by the <see cref="TypeMarshallerCache"/>.
+    /// </summary>
+    public interface ITypeMashalerProvider
+    {
+        // Return null if it doesn't handle it. 
+        // A single ITypeMarshaler can be created once per type and then reused for each instance.
+        // Pass in a cache for aggregate types that need to marshal sub types. 
+        public ITypeMarshaler New(Type type, TypeMarshallerCache cache, int maxDepth);
```
MikeStall, line 18: >New [](http://example.com/codeflow?start=30&length=3)

TryGetMarshaler()

src/libraries/Microsoft.PowerFx.Core/Public/Marshal/TableMarshalerProvider.cs
```diff
@@ -0,0 +1,119 @@
...
+namespace Microsoft.PowerFx
+{
+    /// <summary>
+    /// Marshal .Net classes (with fields). This supports strong typing and lazy marshalling. 
+    /// </summary>
+    public class TableMarshalerProvider : ITypeMashalerProvider
+    {
+        public ITypeMarshaler New(Type type, TypeMarshallerCache cache, int maxDepth)
+        {
+            if (!type.IsArray)
```
MikeStall, line 24: >if (!type.IsArray) [](http://example.com/codeflow?start=12&length=18)

IEnumerable too?

src/tests/Microsoft.PowerFx.Interpreter.Tests/MarshalTests.cs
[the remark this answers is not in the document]
MikeStall: good idea - this triggers a call to RecordValue.Fields, which I was hoping to avoid.

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValue.cs
```diff
@@ -157,6 +157,22 @@ public static UntypedObjectValue New(IUntypedObject untypedObject)
                 untypedObject);
         }
 
+/*
+        // $$$
```
MikeStall, line 161: >// $$$ [](http://example.com/codeflow?start=8&length=6)

We should be able to cleanup a lot of New() code in this file and leverage the providers.

src/tests/Microsoft.PowerFx.Interpreter.Tests/MarshalTests.cs
```diff
@@ -0,0 +1,347 @@
...
+            var result1 = engine.Eval("x.Data");
+            Assert.Equal(10.0, ((NumberValue)result1).Value);
+
+            var result2 = engine.Eval("x.Next.Data");
+            Assert.Equal(20.0, ((NumberValue)result2).Value);
+
+            var result3 = engine.Eval("x.Next.Next.Data");
+            Assert.Equal(30.0, ((NumberValue)result3).Value);
+
+            // $$$ test that it was Truncated...
```
CarlosFigueiraMSFT, line 122: >// $$$ test that it was Truncated... [](http://example.com/codeflow?start=12&length=36)

What should be the behavior here? Set any object reference which we previously saw to null?

src/libraries/Microsoft.PowerFx.Core/Public/Values/UntypedObjectValue.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.Contracts;
+using System.Reflection;
 using System.Text;
```
CarlosFigueiraMSFT, line 9: Only change in this file. Needed?

src/tests/Microsoft.PowerFx.Interpreter.Tests/MarshalTests.cs
```diff
@@ -0,0 +1,347 @@
...
+        {
+            var cache = new TypeMarshallerCache();
+            var tm = cache.GetMarshaler(typeof(int));
+
+            var value = tm.Marshal(5);
+
+            Assert.Equal(5.0, ((NumberValue)value).Value);
+        }
+
+        [Fact]
```
CarlosFigueiraMSFT, line 29: What would be the expected result of marshalling an object that we don't support in Power Fx, such as `new int[] { 1, 2, 3 }`? Do we wrap the objects in an object with a Value property (which is what is done in custom connectors that need to talk to APIs that have arrays of primitive values)? Fail?

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValue.cs
```diff
@@ -130,109 +140,23 @@ public static ErrorValue NewError(ExpressionError error)
...
-        public static FormulaValue New(object obj, Type type)
+        // Marshal an arbitray object (scalar, record, table, etc) into a FormulaValue. 
+        public static FormulaValue New(object obj, Type type, TypeMarshallerCache cache = null)
         {
-            if (obj == null)
-            {
-                return NewBlank();
-            }
-
-            if (obj is FormulaValue value)
```
MikeStall, line 168: > if (obj is FormulaValue value) [](http://example.com/codeflow?start=11&length=31)

This gets replaced by the provider model.

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValue.cs
```diff
@@ -20,6 +20,11 @@ namespace Microsoft.PowerFx.Core.Public.Values
     [DebuggerDisplay("{ToObject().ToString()} ({Type})")]
     public abstract class FormulaValue
     {
+        // We place the .New*() methods on FormulaValue for discoverability. 
+        // If we're "marshalling" a T, we need a TypeMarshallerCache
+        // Else, if we're "constructing" a Table/Record from existing FormulaValues, we don't need a marshaller.
+        // We can use C# overloading to resolve.
```
MikeStall, line 26: These rules capture the principles that drive the rest of the changes in this file.

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
```diff
@@ -28,10 +28,18 @@ public static RecordValue Empty()
...
+        /// </summary>
+        /// <param name="name"></param>
+        /// <returns></returns>
+        public FormulaValue GetField(string name)
+        {            
+            var fieldType = ((RecordType)Type).MaybeGetFieldType(name) ?? FormulaType.Blank;
 
-        public virtual FormulaValue GetField(string name)
-        {
-            return GetField(IRContext.NotInSource(FormulaType.Blank), name);
```
MikeStall, line 34: >FormulaType.Blank) [](http://example.com/codeflow?start=50&length=18)

existing bug caught by new tests. Blank is wrong when missing a field in a heterogenous table.

src/libraries/Microsoft.PowerFx.Core/Public/Values/InMemoryTableValue.cs
```diff
@@ -10,31 +11,54 @@
...
-                {
-                    return DValue<RecordValue>.Of(new InMemoryRecordValue(IRContext.NotInSource(recordType), r.Value.Fields));
-                }
-                else
-                {
-                    return r;
-                }
-            });
+                return DValue<RecordValue>.Of(
+                    new InMemoryRecordValue(IRContext.NotInSource(_recordType), record.Value.Fields));
```
MikeStall, line 34: Would be great to get rid of this in the next PR. 
We make table lazy, but still do a full copy for each record.

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
```diff
@@ -24,6 +24,49 @@ internal TableValue(IRContext irContext)
...
+            {
+                record = null;
+                return false;
+            }
+
+            record = Rows.ElementAtOrDefault(index0);
+            return record != null;
+        }
+                
+        private static ErrorValue ArgumentOutOfRange(IRContext irContext)
```
MikeStall, line 60: >ArgumentOutOfRange [](http://example.com/codeflow?start=34&length=18)

this class is in Core; but the common error definitions are in the interpreter. 
Having these value classes kind of blurs the line.

src/tests/Microsoft.PowerFx.Core.Tests/ValueTests/ValueTests.cs
```diff
@@ -0,0 +1,325 @@
...
+        }
+
+        // Heterogenous table.
+        [Fact]
+        public void TableFromMixedRecords()
+        {
+            var cache = new TypeMarshallerCache();
+            RecordValue r1 = FormulaValue.NewRecord(new { a = 10, b = 20, c = 30 }, cache);
+            RecordValue r2 = FormulaValue.NewRecord(new { a = 11,         c = 31 }, cache);
+            TableValue val = FormulaValue.NewTable(r1, r2);
```
CarlosFigueiraMSFT, line 217: Would it have the same result if the order was switched (i.e., first record containing only 'a' and 'c', and second record adding 'b')?

src/tests/Microsoft.PowerFx.Core.Tests/ValueTests/ValueTests.cs
```diff
@@ -0,0 +1,325 @@
...
+        // Single Column Table
+        [Fact]
+        public void TableFromPrimitive()
+        {
+            NumberValue r1 = FormulaValue.New(10);
+            NumberValue r2 = FormulaValue.New(20);
+            TableValue val = FormulaValue.NewSingleColumnTable(r1, r2);
+
+            dynamic d = val.ToObject();
+            Assert.Equal(20.0, d[1]); // SCT returned as arrays
```
CarlosFigueiraMSFT, line 258: >// SCT returned as arrays [](http://example.com/codeflow?start=38&length=25)

Is it the case for all single-column tables? Even if the schema is not ![Value:?]? For example, `FormulaValue.FromJson("[{\"a\":1},{\"a\":2}]")`

src/libraries/Microsoft.PowerFx.Core/Public/Marshal/ITypeMashallerProvider.cs
```diff
@@ -0,0 +1,36 @@
...
+using Microsoft.PowerFx.Core.Public.Values;
+
+namespace Microsoft.PowerFx.Core
+{
+    /// <summary>
+    /// Handles marshalling a given type. Invoked by the <see cref="TypeMarshallerCache"/>.
+    /// </summary>
+    public interface ITypeMashallerProvider
+    {
+        // Return null if it doesn't handle it.
```
CarlosFigueiraMSFT, line 15: >null [](http://example.com/codeflow?start=18&length=4)

Return *false* if it doesn't handle it

src/tests/Microsoft.PowerFx.Core.Tests/ValueTests/ValueTests.cs
MikeStall: No, it's pulling the first .... which could get very confusing ... I think we should require NewTable() takes in the recordType just to be explicit.

src/tests/Microsoft.PowerFx.Core.Tests/ValueTests/ValueTests.cs
MikeStall: Yes.  See TableValue.ToObject()
It's keyed off DType.IsColumn, and then returns an array of the *first* field.

CarlosFigueiraMSFT: That can get confusing - there are reasons to have a single-column record, and we would lose it (including its schema) if we "objectify" it. I don't have a good solution, but I don't like that we change the user's data "silently"

src/libraries/Microsoft.PowerFx.Core/Public/Marshal/ObjectMarshalerProvider.cs
```diff
@@ -0,0 +1,139 @@
...
+        /// <summary>
+        /// Provides a customization point to control how properties are marshalled. 
+        /// This returns null to skip the property, else return the name it should have on the power fx record.
+        /// If this is insufficient, a caller can always implement their own marshaller and return a 
+        /// a <see cref="ObjectMarshaler"/> directly. 
+        /// </summary>
+        public Func<PropertyInfo, string> PropertyMapperFunc = (propInfo) => propInfo.Name;
+
+        /// <inheritdoc/>
+        public bool TryGetMarshaller(Type type, TypeMarshallerCache cache, int maxDepth, out ITypeMarshaller marshaler)
```
CarlosFigueiraMSFT, line 34: >TryGetMarshaller [](http://example.com/codeflow?start=20&length=16)

We should have tests for derived classes, including those that override property definitions. I don't think it will have any problems, but it will be good to verity.

src/libraries/Microsoft.PowerFx.Core/Public/Marshal/PrimitiveMarshallerProvider.cs
```diff
@@ -0,0 +1,137 @@
...
+using Microsoft.PowerFx.Core.Public.Values;
+
+namespace Microsoft.PowerFx.Core
+{
+    public class PrimitiveMarshallerProvider : ITypeMashallerProvider
+    {
+        // Map from .net types to formulaTypes
+        private static readonly Dictionary<Type, FormulaType> _map = new Dictionary<Type, FormulaType>()
+        {
+            // Fx needs more number typeS:
```
CarlosFigueiraMSFT, line 17: >typeS [](http://example.com/codeflow?start=36&length=5)

[minor] `types` (no capital 'S')

src/tests/Microsoft.PowerFx.Core.Tests/ValueTests/ValueTests.cs
MikeStall: Ack. Maybe ToObject() is always returning a table. And we can have a separate ToArray()  that returns as an SCT.

src/libraries/Microsoft.PowerFx.Core/Public/Marshal/TableMarshallerProvider.cs
```diff
@@ -0,0 +1,122 @@
...
+
+            var tm = cache.GetMarshaller(et, maxDepth);
+             
+            if (tm.Type is RecordType recordType)
+            {
+                // Array of records 
+            }       
+            else
+            {
+                // Single Column table. Wrap the scalar in a record.
```
CarlosFigueiraMSFT, line 42: >scalar [](http://example.com/codeflow?start=49&length=6)

It may not be a scalar - it can be a table instead (e.g., [[1,2],[3,4,5]]). wrapping it in {Value:*} is still correct, it's just the comment that needs to be updated.

src/libraries/Microsoft.PowerFx.Core/Public/Marshal/TableMarshallerProvider.cs
```diff
@@ -0,0 +1,122 @@
...
+                    }
+                }
+            }
+
+            elementType = null;
+            return false;
+        }
+
+        // Convert a single value into a Record for a SCT,  { value : x }
+        [DebuggerDisplay("SCT(_inner)")]
```
CarlosFigueiraMSFT, line 83: >SCT(_inner) [](http://example.com/codeflow?start=26&length=11)

`SCT{_inner}` - curly braces

src/libraries/Microsoft.PowerFx.Core/Public/Marshal/TableMarshallerProvider.cs
```diff
@@ -0,0 +1,122 @@
...
+        internal class SCTMarshaler : ITypeMarshaller
+        {
+            public FormulaType Type { get; set; }
+
+            public ITypeMarshaller _inner;
```
CarlosFigueiraMSFT, line 88: >_inner [](http://example.com/codeflow?start=35&length=6)

Since you require those values to be present, we should create a .ctor that takes them as arguments

src/libraries/Microsoft.PowerFx.Core/Public/Marshal/TableMarshallerProvider.cs
```diff
@@ -0,0 +1,122 @@
...
+                var record = FormulaValue.NewRecordFromFields(defaultField);
+                return record;
+            }
+        }
+
+        internal abstract class TableMarshaller : ITypeMarshaller
+        {
+            public FormulaType Type { get; set; }
+
+            public ITypeMarshaller _rowMarshaler;
```
CarlosFigueiraMSFT, line 104: >public [](http://example.com/codeflow?start=12&length=6)

protected

src/libraries/Microsoft.PowerFx.Core/Public/Marshal/TableMarshallerProvider.cs
```diff
@@ -0,0 +1,122 @@
...
+
+            public abstract FormulaValue Marshal(object value);
+        }
+
+        // T is record type of the table. 
+        [DebuggerDisplay("TableMarshal({_rowMarshaler})")]
+        internal class TableMarshaller<T> : TableMarshaller
```
CarlosFigueiraMSFT, line 111: >TableMarshaller [](http://example.com/codeflow?start=23&length=15)

Use .ctor that takes required fields

src/libraries/Microsoft.PowerFx.Core/Public/Marshal/ITypeMashallerProvider.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Microsoft.PowerFx.Core
+{
+    /// <summary>
+    /// Handles marshalling a given type. Invoked by the <see cref="TypeMarshallerCache"/>.
+    /// </summary>
+    public interface ITypeMashallerProvider
+    {
+        // Return null if it doesn't handle it. 
+        // A single ITypeMarshaler can be created once per type and then reused for each instance.
```
lesaltzm, line 16: nit: for these comments on public interfaces, use 
/// <summary> style doc comments so it gets picked up by intellisense in hosts.

src/libraries/Microsoft.PowerFx.Core/Public/Marshal/ITypeMashallerProvider.cs
```diff
@@ -0,0 +1,36 @@
...
+using System;
+using Microsoft.PowerFx.Core.Public.Types;
+using Microsoft.PowerFx.Core.Public.Values;
+
+namespace Microsoft.PowerFx.Core
+{
+    /// <summary>
+    /// Handles marshalling a given type. Invoked by the <see cref="TypeMarshallerCache"/>.
+    /// </summary>
+    public interface ITypeMashallerProvider
```
lesaltzm, line 13: >ITypeMashallerProvider [](http://example.com/codeflow?start=21&length=22)

spelling

lesaltzm: check file name spellings as well.

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValue.cs
```diff
@@ -366,57 +291,74 @@ internal static RecordValue RecordFromJsonObject(JsonElement element)
...
-            var recordType = (RecordType)records.First().IRContext.ResultType;
-            return TableFromRecords(records, recordType.ToTable());
+            // Statically marshal the T to a FormulaType.             
+            var value = cache.Marshal(rows);
+            return (TableValue)value;
         }
 
-        public static TableValue TableFromRecords(params RecordValue[] values)
+        /// <summary>
+        /// Construct a table from records. Assumed that Records must be the same type.
```
CarlosFigueiraMSFT, line 326: >Assumed that Records must be the same type. [](http://example.com/codeflow?start=44&length=43)

Should we union the types? For example, Table({a:1}, {b:2}, {a:3}) creates a result of type *[a:n,b:n]

src/libraries/Microsoft.PowerFx.Core/Public/Marshal/PrimitiveMarshallerProvider.cs
```diff
@@ -0,0 +1,137 @@
...
+            { typeof(double), FormulaType.Number },
```
lesaltzm, line 18: If we do go add more number types, changing this could be a breaking change in how existing uses marshal types. Just a note, not actionable.

CarlosFigueiraMSFT:
>                     return TableFromJsonArray(element);

We need to have some validation to ensure that we don't lose data. For example, we should throw if we have something like `[{"a":1},{"a":2},"oops",{"a":4}]`.

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValue.cs:207 in fbf7f02. [](commit_id = fbf7f024ffc79e1a0312bb2c31d0dbdaf6b5a6d4, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Public/Marshal/TypeMarshallerCache.cs
```diff
@@ -0,0 +1,152 @@
...
+        /// <param name="maxDepth">maximum depth to marshal.</param>
+        /// <returns>A marshaller instance that can marshal objects of the given type.</returns>
+        public ITypeMarshaller GetMarshaller(Type type, int maxDepth = DefaultDepth)
+        {
+            if (maxDepth < 0)
+            {
+                return new Empty();
+            }
+
+            // The ccahe requires an exact type match and doesn't handle base types.
```
lesaltzm, line 53: >ccahe [](http://example.com/codeflow?start=19&length=5)

sp

src/tests/Microsoft.PowerFx.Interpreter.Tests/MarshalTests.cs
```diff
@@ -0,0 +1,569 @@
...
+        // Test a custom marshaler. 
+        [Fact]
+        public void CustomMarshalerType()
+        {
+            var cache = new TypeMarshallerCache();
+
+            // Insert at 0 to get precedence over generic object marshaller
+            var marshaler = new WidgetMarshalerProvider();
+            cache.Marshallers.Insert(0, marshaler);
+
```
lesaltzm, line 440: This seems like a method we should provide. Exposing the list and requiring hosts to know about the order and the meaning of the order isn't ideal.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -125,7 +125,7 @@ public static FormulaValue CountRows(IRContext irContext, TableValue[] args)
             var arg0 = args[0];
 
             // Streaming 
-            var count = arg0.Rows.Count();
+            var count = arg0.Count();
```
CarlosFigueiraMSFT, line 128: CountRows should return an error if there are error records in the table. I know that this makes the operation O(n) instead of O(1), but we need that for consistency.

src/libraries/Microsoft.PowerFx.Core/Public/Marshal/ITypeMashallerProvider.cs
MikeStall: What's the correct spelling?
I think it's: "Marshaller" / "Marshalling", "Marshal"

src/libraries/Microsoft.PowerFx.Core/Public/Marshal/PrimitiveMarshallerProvider.cs
MikeStall: Agreed. Although this is just capturing the existing semantics we had in FormulaValue.New.

src/libraries/Microsoft.PowerFx.Core/Public/Marshal/ITypeMashallerProvider.cs
lesaltzm: That sounds right, yeah.

src/libraries/Microsoft.PowerFx.Core/Public/Marshal/TableMarshallerProvider.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// <inheritdoc/>
+        public bool TryGetMarshaller(Type type, TypeMarshallerCache cache, int maxDepth, out ITypeMarshaller marshaler)
+        {
+            if (!IsIEnumerableT(type, out var et))
+            {
+                marshaler = null;
+                return false;
+            }
+
+            var tm = cache.GetMarshaller(et, maxDepth);
```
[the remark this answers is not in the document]
MikeStall: I'm not sure ... 
Throwing is "safer" in that it more aggressively fails and we could loosen later. 
Returning false means we're allowing another marshaller to handle it - in theory, that marshaller could have just been added first to get higher priortiy.

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValue.cs
MikeStall: I think we should just require the type to be passed in and be explicit.  This is getting into the kind of analysis the binder does. 
We can add a future overload that tries to be smarter.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
MikeStall: ack - but this is existing code. So that means we're missing a test...

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
```diff
@@ -24,6 +24,49 @@ internal TableValue(IRContext irContext)
...
 
+        public virtual int Count()
+        {
+            return Rows.Count();
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="index0">0-based index.</param>
```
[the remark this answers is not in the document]
MikeStall: ok. will switch to 1-based.

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNewJson.cs
```diff
@@ -0,0 +1,140 @@
...
+using System.Linq;
+using System.Reflection;
+using System.Text.Json;
+using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Core.Public.Types;
+
+namespace Microsoft.PowerFx.Core.Public.Values
+{
+    // Marshalling from Json. These are convenience operations. 
+    public partial class FormulaValue
```
MikeStall, line 18: >FormulaValue [](http://example.com/codeflow?start=25&length=12)

next - planning to move these separate files to extension methods.

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/PrimitiveTypeMarshaler.cs
```diff
@@ -0,0 +1,110 @@
...
+using System.Diagnostics;
+using Microsoft.PowerFx.Core.Public.Types;
+using Microsoft.PowerFx.Core.Public.Values;
+
+namespace Microsoft.PowerFx
+{
+    /// <summary>
+    /// Marshaller for builtin primitives. 
+    /// </summary>
+    [DebuggerDisplay("ObjMarshal({Type})")]
```
CarlosFigueiraMSFT, line 15: >ObjMarshal({Type}) [](http://example.com/codeflow?start=22&length=18)

`PrimitiveMarshal({Type})`

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        /// <summary>
+        /// This table can be directly indexed (without a linear scan).
+        /// </summary>
+        public bool CanIndex => _sourceIndex != null;
+
+        /// <summary>
+        /// This table can return a count (without a linear scan). 
+        /// </summary>
+        public bool CanCount => _sourceCount != null;
```
CarlosFigueiraMSFT, line 37: >CanCount [](http://example.com/codeflow?start=20&length=8)

Typically if types have a CanXXX member and a XXX member, if CanXXX returns false then XXX will fail / throw (example: Stream.CanRead/CanWrite and Read/Write), which is not the case here...

CarlosFigueiraMSFT: Are these properties (CanIndex / CanCount) needed in the first place?

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNew.cs
```diff
@@ -0,0 +1,124 @@
...
+
+namespace Microsoft.PowerFx.Core.Public.Values
+{
+    // Basc new operators. 
+    public partial class FormulaValue
+    {
+        #region Host Utility API
+
+        // Host utility creation methods, listed here for discoverability.
+        // NOT FOR USE IN THE INTERPRETER! When creating new instances in
```
CarlosFigueiraMSFT, line 16: If those methods should not be called outside this assembly, can we make them internal?

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNew.cs
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        public static GuidValue New(Guid guid)
+        {
+            return new GuidValue(IRContext.NotInSource(FormulaType.Guid), guid);
+        }
+
+        public static StringValue New(string value)
+        {
+            return new StringValue(IRContext.NotInSource(FormulaType.String), value);
```
CarlosFigueiraMSFT, line 63: >StringValue [](http://example.com/codeflow?start=23&length=11)

Check if value is null, and return BlankValue if that is the case

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNewJson.cs
```diff
@@ -0,0 +1,140 @@
...
+        /// </summary>
+        /// <param name="jsonString"></param>
+        public static FormulaValue FromJson(string jsonString)
+        {
+            try
+            {
+                using var document = JsonDocument.Parse(jsonString);
+                using var jsonMemStream = new MemoryStream();
+                using var paJsonWriter = new Utf8JsonWriter(jsonMemStream);
+                var propBag = document.RootElement;
```
CarlosFigueiraMSFT, line 31: Unused

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNewJson.cs
```diff
@@ -0,0 +1,140 @@
...
+                return FromJson(propBag);
+            }
+            catch
+            {
+                // $$$ better error handling here?
+                throw;
+            }
+        }
+
+        public static FormulaValue FromJson(JsonElement element)
```
CarlosFigueiraMSFT, line 42: >public [](http://example.com/codeflow?start=8&length=6)

Private? Or add comments

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
MikeStall: We could remove.  It's telling if the ability is O[1] or O[N]. 
This is similar to delegation capabilities. You "can" do an operation that 's not delegated, it's just linear & on client.

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNew.cs
MikeStall: They are public for hosts. Ie, PAD, PVA would call them when creating initial values to pass into fx. 

But the interpreter should *not* call them from within the core of the interpreter.

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNewJson.cs
```diff
@@ -0,0 +1,140 @@
...
+                case JsonValueKind.Array:
+                    return TableFromJsonArray(element);
+
+                default:
+                    throw new NotImplementedException($"Unrecognized JsonElement {element.ValueKind}");
+            }
+        }
+
+        // Json objects parse to records. 
+        internal static RecordValue RecordFromJsonObject(JsonElement element)
```
CarlosFigueiraMSFT, line 73: >internal [](http://example.com/codeflow?start=8&length=8)

`private`? This is already exposed via `FromJson(JsonElement)`. Same for TableFromJsonArray

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNewJson.cs
```diff
@@ -0,0 +1,140 @@
...
+        internal static TableValue TableFromJsonArray(JsonElement array)
+        {
+            Contract.Assert(array.ValueKind == JsonValueKind.Array);
+
+            var records = new List<RecordValue>();
+
+            for (var i = 0; i < array.GetArrayLength(); ++i)
+            {
+                var element = array[i];
+                var val = GuaranteeRecord(FromJson(element));
```
CarlosFigueiraMSFT, line 106: >GuaranteeRecord(FromJson(element)) [](http://example.com/codeflow?start=26&length=34)

This would work with things like `[1, { Value: 2 }, 3]`, which we probably want to block

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNewJson.cs
```diff
@@ -0,0 +1,140 @@
...
+
+            // Constructor will handle both single-column table 
+            TableType type;
+            if (records.Count == 0)
+            {
+                type = new TableType();
+            }
+            else
+            {
+                type = TableType.FromRecord((RecordType)GuaranteeRecord(records[0]).IRContext.ResultType);
```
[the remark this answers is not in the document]
CarlosFigueiraMSFT: As a minimum we should fail if records have different schemas, so that we can add it in the future without being a breaking change.

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNewRecord.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Linq;
+using System.Reflection;
+using System.Text.Json;
+using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Core.Public.Types;
+
+namespace Microsoft.PowerFx.Core.Public.Values
+{
+    // Marshalling a tables. 
+    // Tables need to know their record type.
```
CarlosFigueiraMSFT, line 18: The comment is stale

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNewTable.cs
```diff
@@ -0,0 +1,77 @@
...
+        public static TableValue NewSingleColumnTable<T>(IEnumerable<PrimitiveValue<T>> values)
+        {
+            // This is a convenience method. For anything requiring more control,
+            // use NewTable and be explicit. 
+            if (!BuiltinFormulaTypeConversions.TryGetFormulaType(typeof(T), out var fxType))
+            {
+                throw new InvalidOperationException($"Use NewTable() instead");
+            }
+
+            const string valueField = "Value";
```
CarlosFigueiraMSFT, line 65: >"Value" [](http://example.com/codeflow?start=38&length=7)

[minor] this constant is used everywhere, we should define it as a proper `const` and reuse it. Can be done in another PR, though

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
CarlosFigueiraMSFT: I'd rather remove them to reduce the surface area, and add them in the future if we need them

src/libraries/Microsoft.PowerFx.Core/Public/Values/InMemoryTableValue.cs
CarlosFigueiraMSFT: Why do we need to copy each record?

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNewJson.cs
MikeStall, line 18: >FormulaValue [](http://example.com/codeflow?start=25&length=12)

This file is all existing code, moved out from FormulaValue.cs

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNew.cs
```diff
@@ -0,0 +1,124 @@
...
+        // Host utility creation methods, listed here for discoverability.
+        // NOT FOR USE IN THE INTERPRETER! When creating new instances in
+        // the interpreter, call the constructor directly and pass in the
+        // IR context from the IR node.
+        public static NumberValue New(double number)
+        {
+            return new NumberValue(IRContext.NotInSource(FormulaType.Number), number);
+        }
+
+        public static FormulaValue New(double? number)
```
lesaltzm, line 24: Why is this one nullable, but not the others? Should we be consistent and accept nullable for all these formula values?

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNewTable.cs
```diff
@@ -0,0 +1,77 @@
...
+
+namespace Microsoft.PowerFx.Core.Public.Values
+{
+    // Marshalling a tables. 
+    // Tables need to know their record type.
+    // - a ITypeMarshaller  (which can be obtained from a TypeMarshallerCache and a T)
+    // - an explicit RecordType
+    public partial class FormulaValue
+    {
+        #region Host Tables API
```
lesaltzm, line 24: If it's split into files, the regions aren't particularly helpful

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/ObjectMarshaler.cs
```diff
@@ -0,0 +1,73 @@
...
+using Microsoft.PowerFx.Core.Public.Types;
+using Microsoft.PowerFx.Core.Public.Values;
+
+namespace Microsoft.PowerFx
+{
+    /// <summary>
+    /// Marshal a specific type of object to a record. 
+    /// </summary>
+    [DebuggerDisplay("ObjMarshal({Type})")]
+    public class ObjectMarshaler : ITypeMarshaller
```
lesaltzm, line 17: nit: Marshaller

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/ObjectMarshaler.cs
```diff
@@ -0,0 +1,73 @@
...
+        public FormulaType Type { get; private set; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ObjectMarshaler"/> class.
+        /// </summary>
+        /// <param name="type">The FormulaType that these objects product.</param>
+        /// <param name="fieldMap">A mapping of fx field names to functions that produce that field. </param>
+        public ObjectMarshaler(FormulaType type, IReadOnlyDictionary<string, Func<object, FormulaValue>> fieldMap)
+        {
+            if (!(type is RecordType))
```
lesaltzm, line 32: Not even a nit, no need to change this, but you can do `type is not RecordType` now.

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/ObjectMarshallerProvider.cs
```diff
@@ -0,0 +1,88 @@
...
+                    throw new NameCollisionException(fxName);
+                }
+
+                mapping[fxName] = (object objSource) =>
+                {
+                    var propValue = prop.GetValue(objSource);
+                    return tm.Marshal(propValue);
+                };
+
+                fxType = fxType.Add(fxName, fxFieldType);
```
lesaltzm, line 81: As discussed offline, for future direction here, we could consider supporting display/localized names in this as well.

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TableMarshallerProvider.cs
```diff
@@ -0,0 +1,140 @@
...
+
+            var rowMarshaller = cache.GetMarshaller(et, maxDepth);
+             
+            if (rowMarshaller.Type is RecordType recordType)
+            {
+                // Array of records 
+            }       
+            else
+            {
+                // Single Column table. Wrap the in a record.
```
lesaltzm, line 43: nit: comment grammar

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TableMarshallerProvider.cs
```diff
@@ -0,0 +1,140 @@
...
+                return false;
+            }
+
+            var rowMarshaller = cache.GetMarshaller(et, maxDepth);
+             
+            if (rowMarshaller.Type is RecordType recordType)
+            {
+                // Array of records 
+            }       
+            else
```
lesaltzm, line 41: consider structuring this as `if (rowMarshaller.Type is not RecordType recordType)` and avoid the empty if block

src/libraries/Microsoft.PowerFx.Core/Public/Values/BuiltinFormulaTypeConversions.cs
```diff
@@ -0,0 +1,45 @@
...
+using System;
+using System.Collections.Generic;
+using Microsoft.PowerFx.Core.Public.Types;
+
+namespace Microsoft.PowerFx.Core
+{
+    /// <summary>
+    /// Mapping between builtin dotnet Types and power fx types.
+    /// </summary>
+    public static class BuiltinFormulaTypeConversions
```
lesaltzm, line 13: Do we expect this to be useful publicly?

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/ObjectMarshaler.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// <inheritdoc/>
+        public FormulaValue Marshal(object source)
+        {
+            var value = new ObjectRecordValue(IRContext.NotInSource(Type), source, this);
+            return value;
+        }
+
+        // Get the value of the field. 
+        // Return null on missing
+        internal FormulaValue TryGetField(object source, string name)
```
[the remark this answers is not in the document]
CarlosFigueiraMSFT: `GetFieldOrDefault`?

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNew.cs
```diff
@@ -0,0 +1,124 @@
...
+// Licensed under the MIT license.
+
+using System;
+using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Core.Public.Types;
+
+namespace Microsoft.PowerFx.Core.Public.Values
+{
+    // Basc new operators. 
+    public partial class FormulaValue
```
MikeStall, line 11: >FormulaValue [](http://example.com/codeflow?start=25&length=12)

This whole file is existing code., move from FormulaValue.cs

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNew.cs
MikeStall: Not sure - this was existing code moved from FormulaValue.cs

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNew.cs
MikeStall: This returns a StringValue, so it doesn't allow Blank. 
I'll just add the explicit null check.

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNewJson.cs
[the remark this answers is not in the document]
MikeStall: Existing code, will handle in future PR. Filed to track: [https://github.com/microsoft/Power-Fx/issues/239](https://github.com/microsoft/Power-Fx/issues/239)

src/libraries/Microsoft.PowerFx.Core/Public/Values/BuiltinFormulaTypeConversions.cs
MikeStall: Yes.  I wanted it for the SqlRunner, help in marshalling Sql results to Formula Value.

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/ObjectMarshaler.cs
jorisdg: ObjectMarshaler? Or ObjectMarshaller (double L)?

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNew.cs
lesaltzm: Cool, added #241  to look into this

lesaltzm (approved):
🚢 !!

PR #217, opened 2022-03-15
Added Tests for LazyEvalNode

omprakaash:
Added tests to document Source-span errors for LazyEvalNodes in IR.

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests/IRTranslatorTests.cs
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/IRTranslatorTests.cs
@@ -1,0 +1,60 @@
+﻿using System;
+using Microsoft.PowerFx.Core.Binding;
+using Microsoft.PowerFx.Core.Glue;
+using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Core.IR.Nodes;
+using Microsoft.PowerFx.Core.Parser;
+using Microsoft.PowerFx.Core.Public.Types;
+using Microsoft.PowerFx.Core.Public.Values;
+using Microsoft.PowerFx.Core.Syntax;
+using Microsoft.PowerFx.Core.Types.Enums;
+using Xunit;
+using CallNode = Microsoft.PowerFx.Core.IR.Nodes.CallNode;
+
+namespace Microsoft.PowerFx.Core.Tests
+{
+    public class IRTranslatorTests
+    {
+        private readonly EnumStore _enumStore = new EnumStore();
+
+        [Theory]
+        [InlineData("CountIf(numtable, val > 5)")]
+        public void TestLazyEvalNodeSourceSpan(string expression)
+        {
+            var parameterType = new RecordType();
+
+            var tableType = new TableType();
+            tableType = tableType.Add(new NamedFormulaType("val", FormulaType.Number));
+
+            parameterType = parameterType.Add(new NamedFormulaType("numtable", tableType));
+
+            var formula = new Formula(expression);
+            formula.EnsureParsed(TexlParser.Flags.None);
+
+            var binding = TexlBinding.Run(
+                new Glue2DocumentBinderGlue(),
+                formula.ParseTree,
+                new SimpleResolver(_enumStore.EnumSymbols),
+                ruleScope: parameterType._type,
+                useThisRecordForRuleScope: false);
+
+            Assert.False(formula.HasParseErrors);
+            Assert.False(binding.ErrorContainer.HasErrors());
+
+            (var irNode, var ruleScopeSymbol) = IRTranslator.Translate(binding);
+
+            Assert.True(irNode.GetType() == typeof(CallNode));
+            var callNode = irNode as CallNode;
+
+            Assert.True(callNode.IsLambdaArg(1));
+
+            callNode.TryGetArgument(1, out var lazyEvalNode);
+
+            Assert.True(lazyEvalNode.GetType() == typeof(LazyEvalNode));
+
+            // SourceSpan Check
+            Assert.Equal(18, lazyEvalNode.IRContext.SourceContext.Min);
+            Assert.Equal(25, lazyEvalNode.IRContext.SourceContext.Lim);
+        }
+    }
+}


omprakaash:
Wrote a test to document the error. Will go ahead and fix it if everything so far is okay.

omprakaash:
Not sure if I added the test in an appropriate location(Added a new file). Will appreciate any help on this. Thank You !

src/tests/Microsoft.PowerFx.Core.Tests/IRTranslatorTests.cs
```diff
@@ -0,0 +1,63 @@
...
+                ruleScope: parameterType._type,
+                useThisRecordForRuleScope: false);
+
+            Assert.False(formula.HasParseErrors);
+            Assert.False(binding.ErrorContainer.HasErrors());
+
+            (var irNode, var ruleScopeSymbol) = IRTranslator.Translate(binding);
+
+            Assert.True(irNode.GetType() == typeof(CallNode));
+            var callNode = irNode as CallNode;
```
MikeStall, line 50: nit:  
Or you could use:
   Assert.IsType<CallNode>(irNode)

Or you could just cast and drop the assert. It will throw (and still cause test failure) if wrong. Unhandled exception is a fine way to fail a test: 

var callNode = (CallNode) irNode;

src/tests/Microsoft.PowerFx.Core.Tests/IRTranslatorTests.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            Assert.True(callNode.IsLambdaArg(1));
+
+            callNode.TryGetArgument(1, out var lazyEvalNode);
+
+            Assert.True(lazyEvalNode.GetType() == typeof(LazyEvalNode));
+
+            // SourceSpan Check
+            Assert.Equal(18, lazyEvalNode.IRContext.SourceContext.Min);
```
MikeStall, line 59: >Assert [](http://example.com/codeflow?start=12&length=6)

For maintenance, 18 and 25 are magic numbers (it's not obvious to the reader what they mean). 

You could say this:

var fragment = lazyEvalNode.IRContext.SourceContext.GetFragment(expression); 
Assert.Equal("val > 5", fragment)

MikeStall:
Can you place in a subfolder,  \IRTests
We'll add more in the future. 



---
In reply to: [1067519954](https://github.com/microsoft/Power-Fx/pull/217#issuecomment-1067519954) [](http://example.com/codeflow?ancestors=1067519954,1067519954)

MikeStall:
Looking good!

---
In reply to: [1067502402](https://github.com/microsoft/Power-Fx/pull/217#issuecomment-1067502402) [](http://example.com/codeflow?ancestors=1067502402,1067502402)

src/tests/Microsoft.PowerFx.Core.Tests/IRTranslatorTests.cs
omprakaash: Got it ! Thank you for the review. Will edit and push the changes.

omprakaash:
I implemented the code review suggestions and included the fix ( changing node to current arg), but hit a snag. It seems that cases like `Sum(numtable, Sum(val,1))` works fine, but BinOpNode.GetTextSpan() only returns the span of its token(In this case '<'). Thus always resulting in a SourceSpan of length 1.

lesaltzm (dismissed):
Thanks! 🚢

lesaltzm:
Yes, if you're actually interested in the original text span, use GetCompleteSpan. GetTextSpan is used for determining where to put error messages, so it's generally more narrow than the whole text of the node + children

lesaltzm:
For these tests, GetCompleteSpan is what you'd want.

lesaltzm:
Ah, yup! The span should just cover `>` for the IRContext (+ @bryancroteau-MSFT to keep me honest)

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -268,7 +268,7 @@ public override IntermediateNode Visit(TexlCallNode node, IRTranslatorContext co
                     if (func.IsLazyEvalParam(i))
                     {
                         var child = arg.Accept(this, scope != null ? context.With(scope) : context);
-                        args.Add(new LazyEvalNode(context.GetIRContext(node), child));
```
bryancroteau-MSFT, line 271: Why are we making this change? It means that we are treating each argument as returning the same type as the parent node.

src/tests/Microsoft.PowerFx.Core.Tests/IRTests/IRTranslatorTests.cs
```diff
@@ -0,0 +1,63 @@
...
+            var callNode = (CallNode)irNode;
+
+            Assert.True(callNode.IsLambdaArg(1));
+
+            callNode.TryGetArgument(1, out var lazyEvalNode);
+
+            Assert.True(lazyEvalNode.GetType() == typeof(LazyEvalNode));
+
+            // SourceSpan Check
+            var fragment = lazyEvalNode.IRContext.SourceContext.GetFragment(expression);
```
bryancroteau-MSFT, line 59: Why are we only checking the source span?

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
lesaltzm: I think you have it reversed, that's what it was doing previously, each child had the IR Context of the call node, instead of it's own.

src/tests/Microsoft.PowerFx.Core.Tests/IRTests/IRTranslatorTests.cs
lesaltzm: Good point! The IR Context also has the type of the node, we should validate that as well to make this test better.

omprakaash: Got it. Will include type checks as well.

omprakaash:
Added and pushed code that includes tests for ResultType in IRContext.

lesaltzm (approved):

src/tests/Microsoft.PowerFx.Core.Tests/IRTests/IRTranslatorTests.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            (var irNode, var ruleScopeSymbol) = IRTranslator.Translate(binding);
+           
+            var callNode = (CallNode)irNode;
+
+            Assert.True(callNode.IsLambdaArg(1));
+
+            callNode.TryGetArgument(1, out var lazyEvalNode);
+
+            Assert.True(lazyEvalNode.GetType() == typeof(LazyEvalNode));
```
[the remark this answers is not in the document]
omprakaash: Thank you ! Was having trouble with type checking and forgot that is a cleaner way to do it. Should I update the code and push again since the PR has already been approved.

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
MikeStall: Could this lead to wrong execution? Ie, Is there a .txt case that we could have added which would have failed?

src/tests/Microsoft.PowerFx.Core.Tests/IRTests/IRTranslatorTests.cs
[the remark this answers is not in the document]
omprakaash: Got it. Will push the changes

bryancroteau-MSFT (approved):

PR #218, opened 2022-03-15
Add trigonometric functions to C# interpreter

CarlosFigueiraMSFT:
Implementation and tests for trigonometric functions. Also updated the test runner to allow for differences smaller than 1e-5 for expected numeric values.


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/TrigFunctions.txt
```diff
@@ -0,0 +1,351 @@
...
+
+>> Sin(Blank())
+0
+
+>> Sin(Table({a:0},{a:Pi()/6},{a:Pi()/2}))
+#skip
+// Tabular version not implemented yet: expected: Table({a:0},{a:0.5},{a:1})
+
+>> Sin(Table({a:0},{a:1/0},{a:Pi()/2}))
+#skip
```
MikeStall, line 33: Can you move tabular to a separate file?
(easier to exclude for other backends that don't support tabular)

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
```diff
@@ -28,6 +28,8 @@ internal class BuiltinFunctionsCore
         public static readonly TexlFunction AcotT = _library.Append(new AcotTableFunction());
         public static readonly TexlFunction AddColumns = _library.Append(new AddColumnsFunction());
         public static readonly TexlFunction And = _library.Append(new VariadicLogicalFunction(isAnd: true));
+        public static readonly TexlFunction Asin = _library.Append(new AsinFunction());
```
MikeStall, line 31: >_library [](http://example.com/codeflow?start=51&length=8)

These may require a corresponding move from PAClient.

MikeStall (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
CarlosFigueiraMSFT: Yes, we'll remove it when bringing Power Fx to PAClient; those functions should have been moved before with the rest of the trig functions, it was probably an oversight. It's good that we have a test on PAClient that will fail if there are duplicates, so we won't "forget" to remove by mistake.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/TrigFunctions.txt
CarlosFigueiraMSFT: Done

MikeStall (approved):
:shipit:

PR #220, opened 2022-03-15
Add support for Concat's third parameter (separator)

CarlosFigueiraMSFT:
Currently the Concat function declares that it accepts three parameters, but the C# interpreter can only handle two (and will throw an exception if the third one is passed). This adds support for it.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
@@ -438,19 +438,6 @@ namespace Microsoft.PowerFx.Functions
             };
         }
 
-        private static Func<IRContext, int, FormulaValue, FormulaValue> ExactSequenceWithOptionalParameters(Func<IRContext, int, FormulaValue, FormulaValue>[] requiredParamRuntimeChecks, Func<IRContext, int, FormulaValue, FormulaValue>[] optionalParamRuntimeChecks)
-        {
-            return (irContext, index, arg) =>
-            {
-                if (index < requiredParamRuntimeChecks.Length)
-                {
-                    return requiredParamRuntimeChecks[index](irContext, index, arg);
-                }
-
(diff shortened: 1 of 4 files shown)


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Concat.txt
```diff
@@ -12,6 +15,10 @@
 >> Concat(["A", "B", "C"], ThisRecord.Value)
 "ABC"
 
+// Explicit 'ThisRecord', with separator
```
lesaltzm, line 18: Some more tests to consider:

As with separator?
Interpolated string as separator?
Nested concats with separator?
Lambda reference is invalid in separator arg

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
```diff
@@ -438,6 +438,19 @@ private static FormulaValue ExactValueTypeOrTableOrBlank<T>(IRContext irContext,
             };
         }
 
+        private static Func<IRContext, int, FormulaValue, FormulaValue> ExactSequenceWithOptionalParameters(Func<IRContext, int, FormulaValue, FormulaValue>[] requiredParamRuntimeChecks, Func<IRContext, int, FormulaValue, FormulaValue>[] optionalParamRuntimeChecks)
```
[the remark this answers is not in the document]
CarlosFigueiraMSFT: Good point, I didn't realize that. Will update.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Concat.txt
CarlosFigueiraMSFT: - As with separator: added
- Interpolated as separator: added
- Nested concat with separator: already there (Concat(ForAll(["A","B"] As X,Concat(ForAll(["C","D"] As Y, X.Value & Y.Value),Value, "-")),Value, ", "))
- Lambda reference invalid in separator: added

PR #222, opened 2022-03-15
Fix RandBetween to be equivalent to Excel

CarlosFigueiraMSFT:
The function RandBetween takes two parameters, and it should return _an integer_ greater than or equal to the first argument, and smaller than _or equal to_ the second argument (i.e., inclusive). The current implementation was not returning the second argument, nor was it working properly if the first argument was not a whole number. This change brings it to parity with Excel.


MikeStall (approved):
:shipit:

MikeStall (approved):

PR #228, opened 2022-03-18
Expose OptionSet name, and correctly marshal in FormulaType.Build

### LINKED ISSUE #226: Internal access needed to use option sets, issue converting `DType` -> `FormulaType`
1. Internal access needs to be used to access the `IExternalOptionSet` used in a particular formula type value through `DType`.
2. When converting `DType` to `FormulaType` in `FormulaType.Build`, the code below is dropping the `IExternalOptionSet` information, which effectively renders it unusable.

```
  // FormulaType.cs -> FormulaType.Build 
  case DKind.OptionSetValue:
      {
          bool? flag = type.OptionSetInfo?.IsBooleanValued;
          if (!flag.HasValue || !flag.Value)
          {
              return OptionSetValue; // <---- IExternalOptionSet is not passed in here.
          }

          return Boolean;
      }
```

The request here is to expose public `IExternalOptionSet` information on the `OptionSetValueType`. The minimum necessary to avoid using `DType` in PVA is to expose the entity name as a string and propagate it in the above code.

lesaltzm:
Fix #226.

When adding proper option set support, FormulaType.Build wasn't updated to ensure that the internal option set type gets properly converted back to the public FormulaType. This also exposes the name of the option set publicly on the OptionSetValueType, for use in hosts.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
@@ -81,7 +81,16 @@ namespace Microsoft.PowerFx.Core.Public.Types
 
                 case DKind.OptionSetValue:
                     var isBoolean = type.OptionSetInfo?.IsBooleanValued;
-                    return isBoolean.HasValue && isBoolean.Value ? Boolean : OptionSetValue;
+                    if (isBoolean.HasValue && isBoolean.Value)
+                    {
+                        return Boolean;
+                    }
+                    else
+                    {
+                        // In all non-test cases, this option set info must be present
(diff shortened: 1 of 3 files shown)


MikeStall (approved):
:shipit:

Shpakh-MSFT:
> 

Why is this needed? 

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs:52 in ce92819. [](commit_id = ce92819bcf3a6d26dc44b5a4945ff74a10bb2cc4, deletion_comment = False)

Shpakh-MSFT (dismissed):
:shipit:

MikeStall (approved):

Shpakh-MSFT (approved):
:shipit:

PR #229, opened 2022-03-18
Rename ParseJson to ParseJSON

bryancroteau-MSFT:
Rename ParseJson to ParseJSON to be consistent with the existing JSON function.


Ananya-Gupta24 (dismissed):

lesaltzm:
For resource files, to avoid noisy commits, we only change the /en-us/ file. All the others are updated by the loc pipeline

lesaltzm (approved):

PR #230, opened 2022-03-18
Add static analysis (via reflection) tests to look for mutable static…

MikeStall:
… fields - these could cause issues for multi-threading scenarios.


src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryOperators.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.PowerFx.Functions
     internal static partial class Library
     {
         #region Operator Standard Error Handling Wrappers
-        public static FunctionPtr OperatorBinaryAdd = StandardErrorHandling<NumberValue>(
+        public static readonly FunctionPtr OperatorBinaryAdd = StandardErrorHandling<NumberValue>(
```
MikeStall, line 15: >readonly [](http://example.com/codeflow?start=22&length=8)

A lot of the failures in the test are just missing readonly like this...

src/tests/Microsoft.PowerFx.Core.Tests/IntellisenseTests/SignatureHelpTest.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.PowerFx.Tests.IntellisenseTests
 {
     public class SignatureHelpTest : IntellisenseTestBase
     {
-        private static readonly bool RegenerateSignatureHelp = false;
+        private const bool RegenerateSignatureHelp = false;
```
MikeStall, line 16: >const [](http://example.com/codeflow?start=16&length=5)

`const` is stronger than `static readonly`
Const is just literals (like numbers). 
A static readonly could still point to something (a collection) that could get modified.

src/libraries/Microsoft.PowerFx.Core/Types/Enums/ColorTable.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.PowerFx.Core.Types.Enums
 {
     internal static class ColorTable
     {
-        public static Dictionary<string, uint> InvariantNameToHexMap { get; } = new Dictionary<string, uint>
+        public static IReadOnlyDictionary<string, uint> InvariantNameToHexMap { get; } = new Dictionary<string, uint>
```
MikeStall, line 12: >IReadOnlyDictionary [](http://example.com/codeflow?start=22&length=19)

We have a *a lot* of static collection classes. We mark them as safe by using a IReadOnly* interface. 

This isn't 100% bullet proof - somebody could still cast to the mutable object and mutate.

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
```diff
@@ -14,6 +15,7 @@ namespace Microsoft.PowerFx.Core.Public.Types
     /// Formula Types are a class hiearchy.
     /// </summary>
     [DebuggerDisplay("{_type}")]
+    [ImmutableObject(true)]
```
MikeStall, line 18: >[ImmutableObject(true)] [](http://example.com/codeflow?start=4&length=23)

per the test, we could mark "safe" classes as immutable (either via the Sys.ComponentModel attribute or creating our own). 
This just tells the test it's safe; it doesn't actually enforce it (although maybe there's a followup way to enforce classes marked as immutable are actually immutable)

src/tests/Microsoft.PowerFx.Interpreter.Tests/ThreadingTest.cs
```diff
@@ -0,0 +1,214 @@
...
+                        typeof(Core.Localization.TexlStrings.StringGetter),
+                        typeof(Core.Localization.ErrorResourceKey),
+                    };
+                    if (bugs.Contains(field.FieldType))
+                    {
+                        continue;
+                    }
+
+                    // TODO - Drive this list to zero. 
+                    var bugNames = new HashSet<string>
```
MikeStall, line 83: >bugNames [](http://example.com/codeflow?start=24&length=8)

These 2 bug lists give us a denominator. 
Most of these look like just collections / flags to mark readonly; 

There's actually a few interesting cases to review in here...

src/tests/Microsoft.PowerFx.Core.Tests/IntellisenseTests/SignatureHelpTest.cs
lesaltzm: This is in a test assembly, it shouldn't have been picked up by your new coverage (although, yes, good change still)

src/tests/Microsoft.PowerFx.Interpreter.Tests/ThreadingTest.cs
```diff
@@ -0,0 +1,214 @@
...
+    {
+        [Fact]
+        public void CheckIntepreter()
+        {
+            var asm = typeof(RecalcEngine).Assembly;            
+            CheckStatics(asm);
+        }
+
+        [Fact]
+        public void CheckFxCore()
```
lesaltzm, line 26: very minor organizational nit: Tests that cover Core should generally live in Core.Tests (the Interpreter.Tests depends on Core.Tests and has InternalsVisibleTo, so it could still have it's own test that calls the same CheckStatics method)

src/tests/Microsoft.PowerFx.Interpreter.Tests/ThreadingTest.cs
```diff
@@ -0,0 +1,214 @@
...
+                    }
+
+                    if (type.Name.Contains("<>") ||
+                        type.Name.Contains("<PrivateImplementationDetails>"))
+                    {
+                        continue; // exclude compiler generated closures. 
+                    }
+
+                    // Whitelist - these are reviewed and ok
+                    var whitelist = new HashSet<string>
```
lesaltzm, line 58: nit: I don't think we're running policheck on this repo, but it'll require this be allowlist

MikeStall:
>         private FunctionScopeInfo _scopeInfo;

For types marked as Immutable, we still need to verify they actually are. Ie, this field is mutable. 


---
Refers to: src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs:38 in 0fd8dd4. [](commit_id = 0fd8dd4053617291fb24601a475e62991f34d350, deletion_comment = False)

src/tests/Microsoft.PowerFx.Interpreter.Tests/ThreadingTests.cs
```diff
@@ -0,0 +1,29 @@
...
+using Microsoft.PowerFx.Core.Tests;
+using Xunit;
+
+namespace Microsoft.PowerFx.Interpreter.Tests
+{
+    // Do static analysis to look for potential threading issues. 
+    public class ThreadingTests
+    {
+        [Fact]
+        public void CheckIntepreter()
```
MikeStall, line 15: >CheckIntepreter [](http://example.com/codeflow?start=20&length=15)

Interpreter is clean!

src/tests/Microsoft.PowerFx.Core.Tests/ThreadingTests.cs
```diff
@@ -0,0 +1,71 @@
...
+{
+    // Do static analysis to look for potential threading issues. 
+    public class ThreadingTests
+    {
+        [Fact]
+        public void CheckFxCore()
+        {
+            var asm = typeof(Core.Syntax.Nodes.TexlNode).Assembly;
+
+            var bugsFieldType = new HashSet<Type>
```
MikeStall, line 18: >bugsFieldType [](http://example.com/codeflow?start=16&length=13)

Bug list here captures work on Fx Core to get it "clean". 
Will handle these in followup PRs.

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
MikeStall: We also have a potential hole to ensure that the *derived* classes are immutable.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -19,10 +19,10 @@ internal static partial class Library
     {
         public delegate FormulaValue FunctionPtr(EvalVisitor runner, SymbolContext symbolContext, IRContext irContext, FormulaValue[] args);
 
-        public static IEnumerable<TexlFunction> FunctionList => _funcsByName.Keys;
+        public static IEnumerable<TexlFunction> FunctionList => FuncsByName.Keys;
 
         // Some TexlFunctions are overloaded
-        private static readonly Dictionary<TexlFunction, FunctionPtr> _funcsByName = new Dictionary<TexlFunction, FunctionPtr>
```
MikeStall, line 25: >Dictionary [](http://example.com/codeflow?start=32&length=10)

Switch Dict to IReadOnlyDict,.

src/tests/Microsoft.PowerFx.Core.Tests/AnalyzeThreadSafety.cs
```diff
@@ -0,0 +1,177 @@
...
+                    {
+                        errors.Add($"{name} readonly, but still a mutable type {field.FieldType}");
+                    }
+
+                    // Safe! The static field is readonly and set to an immutable object. 
+                }
+            }
+
+            // Sanity check that we actually ran the test. 
+            Assert.True(total > 10, "failed to find fields");
```
Shpakh-MSFT, line 103: >10 [](http://example.com/codeflow?start=32&length=2)

What is this 10? What does this represent?

MikeStall: I just wanted to make sure we're actually scanning something and the test is doing meaningful work. 
I didn't want a bug where we failed to scan fields (eg, wrong reflection flags came up empty) and then passed because we didn't do any work.

Shpakh-MSFT (approved):
:shipit:

PR #233, opened 2022-03-21
Mark rule as async, if async NamedFormulas are used.

VamsiModem:
This PR will help with the deferred execution for the NamedFormulas, if async namedFormulas are used in the rule, we would like to mark the  rule as async.

1. Added IsAsync property in NameLookupInfo


src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalNamedFormulaRule.cs
```diff
@@ -0,0 +1,24 @@
...
+using Microsoft.PowerFx.Core.Lexer.Tokens;
+using Microsoft.PowerFx.Core.Syntax;
+using Microsoft.PowerFx.Core.Types;
+
+namespace Microsoft.PowerFx.Core.App
+{
+    /// <summary>
+    /// Wraps a named formula's Formula, Rule, Token, and ResultType.
+    /// </summary>
+    internal interface IExternalNamedFormulaRule
```
MikeStall, line 14: >IExternalNamedFormulaRule [](http://example.com/codeflow?start=23&length=25)

We should be removing the Externals to reduce coupling. 

Do we need the whole structure here? Or just Async?
If we do need it, is there a first class way to do it in power fx.

VamsiModem: we just need to know if the rule is async.

CarlosFigueiraMSFT, line 14: >IExternalNamedFormulaRule [](http://example.com/codeflow?start=23&length=25)

Can we use the IExternalRule interface for named formulas, instead of having a separate interface?

MikeStall: we chatted offline, and we'll avoid this new interface and instead have a flag on the NameLookupInfo.

src/libraries/Microsoft.PowerFx.Core/Binding/BindInfo/NameLookupInfo.cs
```diff
@@ -21,11 +21,12 @@ internal struct NameLookupInfo
         /// it has the display name of the object in Data. 
         /// </summary>
         public readonly DName DisplayName;
+        public readonly bool IsAsync;
 
         // Optional data associated with a name. May be null.
         public readonly object Data;
 
-        public NameLookupInfo(BindKind kind, DType type, DPath path, int upCount, object data = null, DName displayName = default)
+        public NameLookupInfo(BindKind kind, DType type, DPath path, int upCount, object data = null, DName displayName = default, bool isAsync = default)
```
MikeStall, line 29: >default [](http://example.com/codeflow?start=146&length=7)

nit - can we use the literal. false or true?
It's a bit clearer to the reader.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -3058,7 +3058,7 @@ public override void Visit(FirstNameNode node)
                 }
 
                 // Any connectedDataSourceInfo or option set or view needs to be accessed asynchronously to allow data to be loaded.
-                if (lookupInfo.Data is IExternalTabularDataSource || lookupInfo.Kind == BindKind.OptionSet || lookupInfo.Kind == BindKind.View)
+                if (lookupInfo.Data is IExternalTabularDataSource || lookupInfo.Kind == BindKind.OptionSet || lookupInfo.Kind == BindKind.View || lookupInfo.IsAsync)
```
MikeStall, line 3061: >lookupInfo.Data is IExternalTabularDataSource || lookupInfo.Kind == BindKind.OptionSet || lookupInfo.Kind == BindKind.View  [](http://example.com/codeflow?start=20&length=123)

Eventually, these checks should get removed from the binder and we should just use the IsAsync flag. 

Idea to consider: 
As an intermediate step, we could move this check into the NameLookupInfo ctor and set IsAync == true if these other flags are set. 
That lets us simplify the binder.cs now; and then remove that check altogether when PAclient is eventually updated.

src/libraries/Microsoft.PowerFx.Core/App/Controls/IExternalRule.cs
```diff
@@ -20,8 +20,10 @@ internal interface IExternalRule
 
         bool HasErrors { get; }
 
+        bool IsAsync { get; }
```
MikeStall, line 23: >bool IsAsync { get; } [](http://example.com/codeflow?start=8&length=21)

Is this still needed?

VamsiModem: Removed it.

MikeStall (approved):
:shipit:

PR #236, opened 2022-03-21
Add PowerFxConfig support for top level display names

lesaltzm:


src/libraries/Microsoft.PowerFx.Core/Binding/BindInfo/NameLookupInfo.cs
```diff
@@ -15,12 +15,17 @@ internal struct NameLookupInfo
         public readonly DPath Path;
         public readonly int UpCount;
         public readonly DType Type;
-        public readonly DName LogicalName;
```
lesaltzm, line 18: This was entirely unused

MikeStall:
>     internal struct NameLookupInfo

As we add more fields here, we may want to convert from struct to class. 

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Binding/BindInfo/NameLookupInfo.cs:12 in 562f191. [](commit_id = 562f1918f7af40dfed3e939500485210c98f27e6, deletion_comment = False)

src/tests/Microsoft.PowerFx.Interpreter.Tests/DisplayNameTests.cs
```diff
@@ -141,9 +146,9 @@ public void OptionSetDisplayNames(string inputExpression, string outputExpressio
             }
             else
             {
-                var outInvariantExpression = engine.GetInvariantExpression(outputExpression, new RecordType());
+                var outInvariantExpression = engine.GetInvariantExpression(inputExpression, new RecordType());
```
MikeStall, line 149: >inputExpression [](http://example.com/codeflow?start=75&length=15)

double checking , outputExpression --> inputExpression?

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -20,6 +20,7 @@ public sealed class PowerFxConfig
         private bool _isLocked;
         private readonly Dictionary<string, TexlFunction> _extraFunctions;
         private readonly Dictionary<DName, IExternalEntity> _environmentSymbols;
+        private SingleSourceDisplayNameProvider _environmentSymbolDisplayNameProvider;
```
MikeStall, line 23: >SingleSourceDisplayNameProvider [](http://example.com/codeflow?start=16&length=31)

nit: readonly?

src/tests/Microsoft.PowerFx.Interpreter.Tests/DisplayNameTests.cs
lesaltzm: Yeah, this test was accidentally passing previously, it wasn't testing the right thing. It was actually testing that passing an invariant expression to GetInvariantExpression returns the same invariant expression.

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -64,6 +73,22 @@ internal void AddFunction(TexlFunction function)
...
+                
+        internal bool TryGetSymbol(DName name, out IExternalEntity symbol, out DName displayName)
+        {
+            var lookupName = name;
+            if (_environmentSymbolDisplayNameProvider.TryGetLogicalName(name, out var logicalName))
+            {
+                lookupName = logicalName;
+                displayName = name;
+            }
+            else if (_environmentSymbolDisplayNameProvider.TryGetDisplayName(name, out displayName))
```
MikeStall, line 85: >TryGetDisplayName [](http://example.com/codeflow?start=59&length=17)

should lookup by logical name be done first to have higher precedence?

lesaltzm: Good point. I'll also update it to include entities without display names in the collision checks, so lookups by one vs the other won't have any shadowing/precedence behavior.

PR #237, opened 2022-03-22
Utility method to create Color from host app

jorisdg:
When working on a personal project, ran into the fact that FormulaValue doesn't have a utility method to create from Color


lesaltzm:
Can you add any tests that exercise this?

jorisdg:
> Can you add any tests that exercise this?

There don't seem to be any explicit tests for these utility methods. Some (but not all) are sporadically exercised in other tests (in interpreter tests, not core). I'm not a fan of implicit tests as it sort of violates the precision principle for unit tests. Should I add a new utility tests class to the core tests? Happy to add tests for all of those methods, as simple as they would be.

jorisdg:
Yeah looks like he has a primitive tests file in there that would be a good place for it.

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValue.cs
```diff
@@ -115,6 +116,11 @@ public static TimeValue New(TimeSpan value)
             return new TimeValue(IRContext.NotInSource(FormulaType.Time), value);
         }
 
+        public static ColorValue New(Color value)
```
MikeStall, line 119: >Color [](http://example.com/codeflow?start=37&length=5)

One concern - is this taking a dependency on a new assembly? Color is defined in System.Drawing namespace; although looks like in netstandard.dll, so maybe we're ok.

MikeStall: And I guess we already hit this for ColorValue anyways, so not a new concern.

jorisdg: Yes, correct. This is also a supported primitive with an existing dependency on System.Drawing.

MikeStall (dismissed):
:shipit:

MikeStall:
>         #region Host Utility API

fyi, these New helpers are now moved to a separate file. You'll get this on rebase. 

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValue.cs:36 in bd7d82c. [](commit_id = bd7d82c24b1824def380c8dcbdf54654a0a85717, deletion_comment = False)

lesaltzm (approved):
🚢

raghavthind2005 (approved):

PR #238, opened 2022-03-22
Add GetRequiredEnumNames method to TexlFunction

bryancroteau-MSFT:
Part 1 of a series of changes that will allow required enums to be automatically configured for a given function library. This change adds a new method, GetRequiredEnumNames, to the TexlFunction base class. It overrides this method for functions within our library which require enums.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
@@ -344,6 +344,13 @@ namespace Microsoft.PowerFx.Core.Functions
         // Functions with optional parameters have more than one signature.
         public abstract IEnumerable<TexlStrings.StringGetter[]> GetSignatures();
 
+        // Return all enums that are required by this function.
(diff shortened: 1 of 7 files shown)


lesaltzm (approved):
:ship:

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -344,6 +344,13 @@ protected set
         // Functions with optional parameters have more than one signature.
         public abstract IEnumerable<TexlStrings.StringGetter[]> GetSignatures();
 
+        // Return all enums that are required by this function.
+        // This can be used to generate a list of enums required for a function library.
+        public virtual IEnumerable<string> GetRequiredEnumNames()
+        {
+            return new List<string>();
```
MikeStall, line 351: >List<string>(); [](http://example.com/codeflow?start=23&length=15)

nit: 
This allocates a new List object on each time.  
0-length array would be better:  `new string[0]`
Or even  `Enumerable.Empty<string>();`

PR #240, opened 2022-03-22
Add GUID support to C# Interpreter

lesaltzm:
This moves the GUID functions into the C# interpreter, and brings up an recurring. The DType.Accepts() logic is based heavily around JS's ducktyping behavior. We should consider removing much of that, and switching to a model where we always inject coercion between types. This would require improving function signatures and making the function type checking more metadata-driven, since some functions have custom behavior around Coercion logic.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/IR/CoercionMatrix.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/IR/CoercionMatrix.cs
@@ -346,6 +346,10 @@ namespace Microsoft.PowerFx.Core.IR
             {
                 return CoercionKind.ViewToText;
             }
+            else if (DType.Guid.Accepts(fromType))
+            {
+                return CoercionKind.GuidToText;
+            }
             else
             {
                 return CoercionKind.None; // Implicit coercion?
--- a/src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
@@ -1881,7 +1881,6 @@ namespace Microsoft.PowerFx.Core.Types
                         type.Kind == DKind.Media ||
                         type.Kind == DKind.Blob ||
                         type.Kind == DKind.Unknown ||
-                        type.Kind == DKind.Guid ||
                         (type.Kind == DKind.Enum && Accepts(type.GetEnumSupertype()));
                     break;
 
--- a/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Guid.txt
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Guid.txt
@@ -1,0 +1,40 @@
+﻿// Guid function 
+// https://docs.microsoft.com/en-us/powerapps/maker/canvas-apps/functions/function-guid
+
+// Guid with no args is volatile, so we can't directly test the output
+
+>> IsBlank(GUID())
+false
+
+>> IsError(GUID())
+false
+
+// Multiple invocations are unique
+>> GUID() <> GUID()
+true
+
+// Coercion to string is 32 chars + 4 hyphens
+>> Len(GUID())
+36
+
+>> GUID("c3987aa4-52ff-462f-9a7c-a13c0dedb16d")
+c3987aa4-52ff-462f-9a7c-a13c0dedb16d
+
+>> GUID("8A9E8006-3EDE-4584-863F-4226BAA73D08")
+8a9e8006-3ede-4584-863f-4226baa73d08
+
+>> GUID("0A6D72E27FDE47C5BB9513BDB51BD7C0")
+0a6d72e2-7fde-47c5-bb95-13bdb51bd7c0
+
+>> GUID("0c99c9a834ed4247861dc9680d72dfd9")
+0c99c9a8-34ed-4247-861d-c9680d72dfd9
+
+>> Text(GUID("41aacfaa-8630-4d9b-8cdd-f62ed8ee8581"))
+"41aacfaa-8630-4d9b-8cdd-f62ed8ee8581"
+
+>> GUID("76353e29-1d57-4315-8af1-884ce7e6c75c") = GUID("76353e29-1d57-4315-8af1-884ce7e6c75c")
+true
+
+// Not comparable to string
+>> GUID("76353e29-1d57-4315-8af1-884ce7e6c75c") = "76353e29-1d57-4315-8af1-884ce7e6c75c"
+Errors: Error 45-46: GUID values can only be compared to other GUID values.
--- a/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/TestRunner.cs
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/TestRunner.cs
@@ -309,6 +309,10 @@ namespace Microsoft.PowerFx.Core.Tests
                 // $$$ proper escaping?
                 sb.Append('"' + s.Value + '"');
             }
+            else if (result is GuidValue guid)
+            {
+                sb.Append(guid.Value);
+            }
             else if (result is TableValue t)
             {
                 var tableType = (TableType)t.Type;
--- a/src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
@@ -54,6 +54,10 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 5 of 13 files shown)


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Guid.txt
```diff
@@ -0,0 +1,40 @@
...
+>> GUID("0A6D72E27FDE47C5BB9513BDB51BD7C0")
+0a6d72e2-7fde-47c5-bb95-13bdb51bd7c0
+
+>> GUID("0c99c9a834ed4247861dc9680d72dfd9")
+0c99c9a8-34ed-4247-861d-c9680d72dfd9
+
+>> Text(GUID("41aacfaa-8630-4d9b-8cdd-f62ed8ee8581"))
+"41aacfaa-8630-4d9b-8cdd-f62ed8ee8581"
+
+>> GUID("76353e29-1d57-4315-8af1-884ce7e6c75c") = GUID("76353e29-1d57-4315-8af1-884ce7e6c75c")
```
CarlosFigueiraMSFT, line 35: >6353e29-1d57-4315-8af1-884ce7e6c75 [](http://example.com/codeflow?start=57&length=34)

Can you change it to upper case to show that it doesn't matter?

CarlosFigueiraMSFT, line 35: >6353e29-1d57-4315-8af1-884ce7e6c75 [](http://example.com/codeflow?start=57&length=34)

Some negative tests (expecting error):

- GUID("not a guid")
- GUID("12345678-1234-1234-1234-12345678901") // missing one char
- GUID("12345678-1234-1234-1234-1234567890123") // one extra char
- GUID("12345678-gggg-1234-1234-123456789012") // characters other than 0-9a-fA-F

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Guid.txt
```diff
@@ -0,0 +1,40 @@
...
+true
+
```
CarlosFigueiraMSFT, line 37: More positive tests:

- GUID(Blank()) ---> expect Blank()
- GUID("11111111222233334444555555555555") // dashes are not necessary

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
```diff
@@ -52,7 +52,7 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
   </ItemGroup>
-
+
```
CarlosFigueiraMSFT, line 55: Change not needed

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1881,7 +1881,6 @@ public virtual bool Accepts(DType type, out KeyValuePair<string, DType> schemaDi
                         type.Kind == DKind.Media ||
                         type.Kind == DKind.Blob ||
                         type.Kind == DKind.Unknown ||
-                        type.Kind == DKind.Guid ||
```
CarlosFigueiraMSFT, line 1884: > type.Kind == DKind.Guid [](http://example.com/codeflow?start=23&length=24)

Will this cause any issues on PAClient?

lesaltzm: Yes, possibly. I'm double checking if there's any impact there. Should be safe, as long as we implement the coercion operator in JS as well.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/TestRunner.cs
```diff
@@ -309,6 +309,10 @@ internal static void TestToString(FormulaValue result, StringBuilder sb)
                 // $$$ proper escaping?
                 sb.Append('"' + s.Value + '"');
             }
+            else if (result is GuidValue guid)
+            {
+                sb.Append(guid.Value);
+            }
```
MikeStall, line 315: Nit: can we have this emit a fx expression, 
Guid("xxxx") 

like we do for date on line 392. 
Helps keep everything unambiguous.

src/libraries/Microsoft.PowerFx.Core/IR/CoercionMatrix.cs
```diff
@@ -346,6 +346,10 @@ private static CoercionKind GetToStringCoercion(DType fromType)
             {
                 return CoercionKind.ViewToText;
             }
+            else if (DType.Guid.Accepts(fromType))
+            {
+                return CoercionKind.GuidToText;
+            }
```
MikeStall, line 352: This is shared by PAClient - can it impact Canvas apps?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/GUID.cs
```diff
@@ -40,7 +40,7 @@ internal sealed class GUIDPureFunction : BuiltinFunction
 
         public override bool IsSelfContained => true;
 
-        public override bool SupportsParamCoercion => false;
+        public override bool SupportsParamCoercion => true;
```
MikeStall, line 43: >true [](http://example.com/codeflow?start=54&length=4)

Is this a potential breaking change? (last time we did something similar we broke tests in paclient)

CarlosFigueiraMSFT, line 43: >true [](http://example.com/codeflow?start=54&length=4)

Why do we want to allow coercion here?

lesaltzm: This is a conversation we should have more broadly, pulled out of this PR.

src/libraries/Microsoft.PowerFx.Core/IR/CoercionMatrix.cs
lesaltzm: See PR 5860239 in PA Client. Yes, there's a potential impact on canvas apps, if an author was using a GUID in a place that expected a string and didn't permit coercion. I don't think there's anything matching that description, but I could have overlooked something. This is a similar change to what we did with DateTime->Number coercion a few months back.

CarlosFigueiraMSFT (dismissed):
🚢

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Guid.txt
```diff
@@ -0,0 +1,79 @@
...
+// Not a valid guid
+>> GUID("560fd2fc-cef6-xxxx-ad40-628a9d2cb164")
+#Error(Kind=InvalidArgument)
+
+
+
+// Not currently supported formats
+>> GUID("{275238ab-f73d-4167-862b-7d6c4e0e288c}")
+#Error(Kind=InvalidArgument
```
bryancroteau-MSFT, line 73: Missing paren?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Guid.txt
```diff
@@ -0,0 +1,79 @@
...
+
+// Coercion to string is 32 chars + 4 hyphens
+>> Len(GUID())
+36
+
+>> GUID(Blank())
+Blank()
+
+>> GUID(Text(1/0))
+#Error(Kind=Div0)
```
bryancroteau-MSFT, line 24: Have we decided on a format for how to specify kind and count for errors?

lesaltzm: @CarlosFigueiraMSFT did a little while back, yes.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Guid.txt
lesaltzm: hm. This should have been a failure with that mismatch, looking into why it passed.

CarlosFigueiraMSFT: The regex looks for it is "lazy" - it will only enforce the error kind if it is defined properly, and ignore the kind otherwise. We should be more strict here - either allow `#Error` only, or force it to be properly formed.

bryancroteau-MSFT: We could fix the regex issue in a future PR. Just fix the missing paren here and then ship.

CarlosFigueiraMSFT (approved):

PR #244, opened 2022-03-23
Add index function to library

rick-nguyen:


src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/UntypedObject.cs
```diff
@@ -54,10 +54,5 @@ public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
         {
             yield return new[] { TexlStrings.IndexArg1, TexlStrings.IndexArg2 };
         }
-
-        public override string GetUniqueTexlRuntimeName(bool isPrefetching = false)
```
MikeStall, line 58: >GetUniqueTexlRuntimeName [](http://example.com/codeflow?start=31&length=24)

what does this change do?

rick-nguyen: When I had put the Index function in the RecalcEngine before, I added that function to separate it from `Index_UO` - which the unit tests were defaulting to and failing.

CarlosFigueiraMSFT (approved):
:shipit:

Shpakh-MSFT (approved):
:shipit:

PR #246, opened 2022-03-23
RecalcEngineResolver will now auto-include required enums

bryancroteau-MSFT:
Adds a new utility method for auto-including required enums into an EnumStore. It follows the "With" function pattern where the instance is returned, only so that it can be used in the base class constructor of RecalcEngineResolver.


src/libraries/Microsoft.PowerFx.Core/Types/Enums/EnumStore.cs
```diff
@@ -298,6 +299,29 @@ internal class EnumStore
...
+        {
+            var missingEnums = new Dictionary<string, string>();
+            foreach (var function in functions)
+            {
+                var requiredEnumNames = function.GetRequiredEnumNames();
+                foreach (var name in requiredEnumNames)
+                {
+                    if (!EnumDict.ContainsKey(name) && !missingEnums.ContainsKey(name))
+                    {
+                        missingEnums.Add(name, _enums[name]);
```
MikeStall, line 312: >name [](http://example.com/codeflow?start=41&length=4)

is this possible to have collisions?

bryancroteau-MSFT: If the user added their own enum with the same name as one of the enums that are required by the functions, then the required enum will not get added.

lesaltzm (approved):

PR #247, opened 2022-03-24
Async support for interpreter.

MikeStall:
Add EvalAsync() to Interpreter and plumb async through EvalVisitor and functions.
Add tests demonstrating true async. 

Still a few .Results to chase down.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/IExpression.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/IExpression.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.PowerFx.Core.Public.Values;
 
 namespace Microsoft.PowerFx.Core.Public
@@ -11,9 +13,25 @@ namespace Microsoft.PowerFx.Core.Public
     /// <returns></returns>
     public interface IExpression
     {
+        public Task<FormulaValue> EvalAsync(RecordValue parameters, CancellationToken cancel);
+    }
+
+    /// <summary>
+    /// Extensions for <see cref="IExpression"/>.
+    /// </summary>
+    public static class IExpressionExtensions
+    {
         /// <summary>
         /// Evaluate the expression with a given set of record values.
         /// </summary>
-        public FormulaValue Eval(RecordValue parameters);
+        public static FormulaValue Eval(this IExpression expr, RecordValue parameters, CancellationToken cancel)
+        {
+            return expr.EvalAsync(parameters, cancel).Result;
+        }
+
+        public static FormulaValue Eval(this IExpression expr, RecordValue parameters)
+        {
+            return expr.Eval(parameters, CancellationToken.None);
+        }
     }
 }
--- a/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
@@ -32,7 +32,7 @@ namespace Microsoft.PowerFx
             CultureInfo = cultureInfo;
             _cancel = cancel;
         }
-
+                
         // Check this cooperatively - especially in any loop. 
         public void CheckCancel()
         {
--- a/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
@@ -2,8 +2,11 @@
 // Licensed under the MIT license.
 
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.Contracts;
 using System.Globalization;
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.PowerFx.Core.IR;
 using Microsoft.PowerFx.Core.IR.Nodes;
 using Microsoft.PowerFx.Core.IR.Symbols;
@@ -15,22 +18,35 @@ using static Microsoft.PowerFx.Functions.Library;
 
 namespace Microsoft.PowerFx
 {
-    internal class EvalVisitor : IRNodeVisitor<FormulaValue, SymbolContext>
+    // This used ValueTask for async, https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/ 
+    // Perf comparison of Task vs. ValueTask: https://ladeak.wordpress.com/2019/03/09/valuetask-vs-task 
+    // Use Task for public methods, but ValueTask for internal methods that we expect to be mostly sync. 
+    internal class EvalVisitor : IRNodeVisitor<ValueTask<FormulaValue>, SymbolContext>
     {
         public CultureInfo CultureInfo { get; }
 
-        public EvalVisitor(CultureInfo cultureInfo)
+        private readonly CancellationToken _cancel;
+
+        public EvalVisitor(CultureInfo cultureInfo, CancellationToken cancel)
         {
             CultureInfo = cultureInfo;
+            _cancel = cancel;
+        }
+
+        // Check this cooperatively - especially in any loop. 
+        public void CheckCancel()
+        {
+            // Throws OperationCanceledException exception
+            _cancel.ThrowIfCancellationRequested();
         }
 
         // Helper to eval an arg that might be a lambda.
-        internal DValue<T> EvalArg<T>(FormulaValue arg, SymbolContext context, IRContext irContext)
+        internal async ValueTask<DValue<T>> EvalArgAsync<T>(FormulaValue arg, SymbolContext context, IRContext irContext)
             where T : ValidFormulaValue
         {
             if (arg is LambdaFormulaValue lambda)
             {
-                var val = lambda.Eval(this, context);
+                var val = await lambda.EvalAsync(this, context);
                 return val switch
                 {
                     T t => DValue<T>.Of(t),
@@ -49,22 +65,22 @@ namespace Microsoft.PowerFx
             };
         }
 
-        public override FormulaValue Visit(TextLiteralNode node, SymbolContext context)
+        public override async ValueTask<FormulaValue> Visit(TextLiteralNode node, SymbolContext context)
         {
             return new StringValue(node.IRContext, node.LiteralValue);
         }
 
-        public override FormulaValue Visit(NumberLiteralNode node, SymbolContext context)
+        public override async ValueTask<FormulaValue> Visit(NumberLiteralNode node, SymbolContext context)
         {
             return new NumberValue(node.IRContext, node.LiteralValue);
         }
 
-        public override FormulaValue Visit(BooleanLiteralNode node, SymbolContext context)
+        public override async ValueTask<FormulaValue> Visit(BooleanLiteralNode node, SymbolContext context)
         {
             return new BooleanValue(node.IRContext, node.LiteralValue);
         }
 
-        public override FormulaValue Visit(TableNode node, SymbolContext context)
+        public override async ValueTask<FormulaValue> Visit(TableNode node, SymbolContext context)
         {
             // single-column table.
 
@@ -74,8 +90,10 @@ namespace Microsoft.PowerFx
             var args = new FormulaValue[len];
             for (var i = 0; i < len; i++)
             {
+                CheckCancel();
+
                 var child = node.Values[i];
-                var arg = child.Accept(this, context);
+                var arg = await child.Accept(this, context);
                 args[i] = arg;
             }
 
@@ -85,30 +103,34 @@ namespace Microsoft.PowerFx
             return tableValue;
         }
 
-        public override FormulaValue Visit(RecordNode node, SymbolContext context)
+        public override async ValueTask<FormulaValue> Visit(RecordNode node, SymbolContext context)
         {
             var fields = new List<NamedValue>();
 
             foreach (var field in node.Fields)
             {
+                CheckCancel();
+
                 var name = field.Key;
                 var value = field.Value;
 
-                var rhsValue = value.Accept(this, context);
+                var rhsValue = await value.Accept(this, context);
                 fields.Add(new NamedValue(name.Value, rhsValue));
             }
 
             return new InMemoryRecordValue(node.IRContext, fields);
         }
 
-        public override FormulaValue Visit(LazyEvalNode node, SymbolContext context)
+        public override async ValueTask<FormulaValue> Visit(LazyEvalNode node, SymbolContext context)
         {
-            var val = node.Child.Accept(this, context);
+            var val = await node.Child.Accept(this, context);
             return val;
         }
 
-        public override FormulaValue Visit(CallNode node, SymbolContext context)
+        public override async ValueTask<FormulaValue> Visit(CallNode node, SymbolContext context)
         {
+            CheckCancel();
+
             // Sum(  [1,2,3], Value * Value)
             // return base.PreVisit(node);
 
@@ -120,12 +142,14 @@ namespace Microsoft.PowerFx
 
             for (var i = 0; i < carg; i++)
             {
+                CheckCancel();
+
                 var child = node.Args[i];
                 var isLambda = node.IsLambdaArg(i);
 
                 if (!isLambda)
                 {
-                    args[i] = child.Accept(this, context);
+                    args[i] = await child.Accept(this, context);
                 }
                 else
                 {
@@ -135,7 +159,12 @@ namespace Microsoft.PowerFx
 
             var childContext = context.WithScope(node.Scope);
 
-            if (func is CustomTexlFunction customFunc)
+            if (func is IAsyncTexlFunction asyncFunc)
+            {
+                var result = await asyncFunc.InvokeAsync(args, _cancel);
+                return result;
+            }
+            else if (func is CustomTexlFunction customFunc)
             {
                 var result = customFunc.Invoke(args);
                 return result;
@@ -144,7 +173,7 @@ namespace Microsoft.PowerFx
             {
                 if (FuncsByName.TryGetValue(func, out var ptr))
                 {
-                    var result = ptr(this, childContext, node.IRContext, args);
+                    var result = await ptr(this, childContext, node.IRContext, args);
 
                     Contract.Assert(result.IRContext.ResultType == node.IRContext.ResultType || result is ErrorValue || result.IRContext.ResultType is BlankType);
 
@@ -155,12 +184,16 @@ namespace Microsoft.PowerFx
             }
         }
 
-        public override FormulaValue Visit(BinaryOpNode node, SymbolContext context)
+        public override async ValueTask<FormulaValue> Visit(BinaryOpNode node, SymbolContext context)
         {
-            var arg1 = node.Left.Accept(this, context);
-            var arg2 = node.Right.Accept(this, context);
+            var arg1 = await node.Left.Accept(this, context);
+            var arg2 = await node.Right.Accept(this, context);
             var args = new FormulaValue[] { arg1, arg2 };
+            return await VisitBinaryOpNode(node, context, args);
+        }
 
+        private ValueTask<FormulaValue> VisitBinaryOpNode(BinaryOpNode node, SymbolContext context, FormulaValue[] args)
+        { 
             switch (node.Op)
             {
                 case BinaryOpKind.AddNumbers:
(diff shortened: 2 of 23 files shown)


src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/LanguageServerTests.cs
```diff
@@ -24,9 +24,9 @@ public class LanguageServerTests
             Converters = { new FormulaTypeJsonConverter() }
         };
 
-        protected static List<string> _sendToClientData;
```
MikeStall, line 27: >static [](http://example.com/codeflow?start=18&length=6)

Pre-existing bug - causes the tests to randomly fail.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -135,7 +159,12 @@ public override FormulaValue Visit(CallNode node, SymbolContext context)
 
             var childContext = context.WithScope(node.Scope);
 
-            if (func is CustomTexlFunction customFunc)
+            if (func is IAsyncTexlFunction asyncFunc)
+            {
+                var result = await asyncFunc.InvokeAsync(args, _cancel);
+                return result;
+            }
```
MikeStall, line 166: This is the critical change - ability to invoke an async function. 
everything else is viral from this.

src/libraries/Microsoft.PowerFx.Core/Public/IExpression.cs
```diff
@@ -10,10 +12,26 @@ namespace Microsoft.PowerFx.Core.Public
     /// </summary>
     /// <returns></returns>
     public interface IExpression
+    {
+        public Task<FormulaValue> EvalAsync(RecordValue parameters, CancellationToken cancel);
```
MikeStall, line 16: >EvalAsync [](http://example.com/codeflow?start=34&length=9)

This is the key public facing change. add EvalAsync(), takes a Task and cancellation token.

src/libraries/Microsoft.PowerFx.Interpreter/ParsedExpression.cs
```diff
@@ -22,12 +24,13 @@ internal ParsedExpression(IntermediateNode irnode, ScopeSymbol topScope, Culture
...
-        public FormulaValue Eval(RecordValue parameters)
+        public Task<FormulaValue> EvalAsync(RecordValue parameters, CancellationToken cancel)
         {
-            var ev2 = new EvalVisitor(_cultureInfo);
+            var ev2 = new EvalVisitor(_cultureInfo, cancel);            
+
             var newValue = _irnode.Accept(ev2, SymbolContext.NewTopScope(_topScopeSymbol, parameters));
 
-            return newValue;
+            return newValue.AsTask();
```
MikeStall, line 33: >.AsTask() [](http://example.com/codeflow?start=27&length=9)

ValueTask<> is has restrictions, so ok internally. Use for when the task is most commonly synchronous - to avoid heap allocations. 
 But public surface should use Task, so we call AsTask to convert.

src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
```diff
@@ -153,4 +155,10 @@ public FormulaValue Invoke(FormulaValue[] args)
             return (FormulaValue)result;
         }
     }
+
+    // A function capable of async invokes. 
+    internal interface IAsyncTexlFunction
```
MikeStall, line 160: >IAsyncTexlFunction [](http://example.com/codeflow?start=23&length=18)

Currently, just tests implement this. 
Can followup on the proper extension story in a future PR.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -17,12 +18,16 @@ namespace Microsoft.PowerFx.Functions
 {
     internal static partial class Library
     {
-        public delegate FormulaValue FunctionPtr(EvalVisitor runner, SymbolContext symbolContext, IRContext irContext, FormulaValue[] args);
+        // Sync FunctionPtr - all args are evaluated before invoking this function.  
+        public delegate FormulaValue FunctionPtr(SymbolContext symbolContext, IRContext irContext, FormulaValue[] args);
```
MikeStall, line 22: >SymbolContext [](http://example.com/codeflow?start=49&length=13)

If a function invokes lambdas, then it's could be async, and so needs to be a AsyncFunctionPtr. 
If it doesn't invoke lambas, then it doesn't need a EvalVisitor parameter.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineAsyncTests.cs
```diff
@@ -0,0 +1,221 @@
...
+            Assert.False(task.IsCompleted);
+
+            cts.Cancel();
+
+            await Assert.ThrowsAsync<TaskCanceledException>(async () => { await task; });
+        }
+
+        // Verify cancellation 
+        [Fact]
+        public async Task InfiniteLoop()
```
MikeStall, line 166: >InfiniteLoop [](http://example.com/codeflow?start=26&length=12)

Even for purely sync functions, we still need a way to abort a long running function.

MikeStall:
>                     return OperatorBinaryAdd(this, context, node.IRContext, args);

These delegates are still async :( 
This is because they go through StandardErrorHandling.cs, which produces async delegates.  


---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs:200 in 068dcf4. [](commit_id = 068dcf4ffc7da9a7db8e5386324ed4449461c099, deletion_comment = False)

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineAsyncTests.cs
```diff
@@ -0,0 +1,221 @@
...
+
+            // Eval may never return.             
+            await Assert.ThrowsAsync<OperationCanceledException>(async () =>
+            {
+                await engine.EvalAsync(expr, cts.Token);
+            });
+        }
+
+        // Helper for making async functions. 
+        internal class CustomAsyncTexlFunction : TexlFunction, IAsyncTexlFunction
```
MikeStall, line 187: >IAsyncTexlFunction [](http://example.com/codeflow?start=63&length=18)

Implementing this is key.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineAsyncTests.cs
```diff
@@ -0,0 +1,221 @@
...
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.Utils;
+using Xunit;
+using Xunit.Sdk;
+using static Microsoft.PowerFx.Core.Localization.TexlStrings;
+
+namespace Microsoft.PowerFx.Tests
+{
+    // Test async eval features. 
+    public class RecalcEngineAsyncTests
```
MikeStall, line 26: >RecalcEngineAsyncTests [](http://example.com/codeflow?start=17&length=22)

The new test cases are key here.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
```diff
@@ -41,16 +42,16 @@ private static bool IsInvalidDouble(double number)
         /// <param name="returnBehavior">A flag that can be used to activate pre-defined early return behavior, such as returning Blank() if any argument is Blank().</param>
         /// <param name="targetFunction">The implementation of the builtin function.</param>
         /// <returns></returns>
-        private static FunctionPtr StandardErrorHandling<T>(
+        private static AsyncFunctionPtr StandardErrorHandling<T>(
```
MikeStall, line 45: >AsyncFunctionPtr [](http://example.com/codeflow?start=23&length=16)

This is subtle but very significant. 

If I didn't produce async, then I'd need 2 copies of this file (one producing sync and another producing async). So to have 1 copy, everything becomes async.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -15,22 +18,35 @@
 
 namespace Microsoft.PowerFx
 {
-    internal class EvalVisitor : IRNodeVisitor<FormulaValue, SymbolContext>
+    // This used ValueTask for async, https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/ 
+    // Perf comparison of Task vs. ValueTask: https://ladeak.wordpress.com/2019/03/09/valuetask-vs-task 
+    // Use Task for public methods, but ValueTask for internal methods that we expect to be mostly sync. 
+    internal class EvalVisitor : IRNodeVisitor<ValueTask<FormulaValue>, SymbolContext>
```
MikeStall, line 24: >ValueTask [](http://example.com/codeflow?start=47&length=9)

Visitor must become Task<>

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -74,8 +90,10 @@ public override FormulaValue Visit(TableNode node, SymbolContext context)
             var args = new FormulaValue[len];
             for (var i = 0; i < len; i++)
             {
+                CheckCancel();
```
MikeStall, line 93: >CheckCancel [](http://example.com/codeflow?start=16&length=11)

We have a cooperative-abort model here, checking for CancellationToken in any loop. (We should have a Code Analyzer enforce this).  For safety, just do all loops - don't bother trying to rationalize between large vs. small loops.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
```diff
@@ -354,12 +355,12 @@ public static FormulaValue Time(IRContext irContext, NumberValue[] args)
...
+        private static async ValueTask<FormulaValue> Now(EvalVisitor runner, SymbolContext symbolContext, IRContext irContext, FormulaValue[] args)
         {
             return new DateTimeValue(irContext, DateTime.Now);
         }
 
-        public static FormulaValue DateParse(EvalVisitor runner, SymbolContext symbolContext, IRContext irContext, StringValue[] args)
+        private static async ValueTask<FormulaValue> DateParse(EvalVisitor runner, SymbolContext symbolContext, IRContext irContext, StringValue[] args)
         {
             var str = args[0].Value;
             if (DateTime.TryParse(str, runner.CultureInfo, DateTimeStyles.None, out var result))
```
MikeStall, line 366: >runner.CultureInfo [](http://example.com/codeflow?start=39&length=18)

If we didn't need runner here, then sig could have been synchronous:
FormulaValue DateParse(IRContext irContext, StringValue[] args)

Felx-B:
Hi 👋
This feature will be very usefull for us !
(async execution + cancellable task to avoid overloading server by heavy formula)

I don't know if it is the right place to leave this comment, but there is another place we would need Async interface : in the `IPowerFxScopeFactory`
We create the engine in async way, so we need to sync over async the `GetOrCreateInstance` method to respect interface 😞

Should I declare and issue about this ?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
```diff
@@ -354,12 +355,12 @@ public static FormulaValue Time(IRContext irContext, NumberValue[] args)
...
         }
 
-        private static FormulaValue Now(EvalVisitor runner, SymbolContext symbolContext, IRContext irContext, FormulaValue[] args)
+        private static async ValueTask<FormulaValue> Now(EvalVisitor runner, SymbolContext symbolContext, IRContext irContext, FormulaValue[] args)
         {
             return new DateTimeValue(irContext, DateTime.Now);
         }
 
-        public static FormulaValue DateParse(EvalVisitor runner, SymbolContext symbolContext, IRContext irContext, StringValue[] args)
+        private static async ValueTask<FormulaValue> DateParse(EvalVisitor runner, SymbolContext symbolContext, IRContext irContext, StringValue[] args)
```
rick-nguyen, line 363: Hi Mike, is this supposed to be private now? Isn't this function targeted by the DateValue function?

lesaltzm: DateValue is in this class, has access to this private static function.

lesaltzm: Actually, pretty much all these functions in Library could be private.

MikeStall: This instance is just a copy&paste error (I had *a lot* of mechanical changes to make ;) )

The class is non-public, so these methods all effectively private anyways.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -15,22 +18,35 @@
...
-        public EvalVisitor(CultureInfo cultureInfo)
+        private readonly CancellationToken _cancel;
+
+        public EvalVisitor(CultureInfo cultureInfo, CancellationToken cancel)
         {
             CultureInfo = cultureInfo;
+            _cancel = cancel;
+        }
+
+        // Check this cooperatively - especially in any loop.
```
bryancroteau-MSFT, line 36: Cooperatively?

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -74,8 +90,10 @@ public override FormulaValue Visit(TableNode node, SymbolContext context)
             var args = new FormulaValue[len];
             for (var i = 0; i < len; i++)
```
bryancroteau-MSFT, line 91: 1) We should configure our code analysis to make sure that CheckCancel() appears in every loop, OR
2) We should have a loop-helper class or function that guarantees that CheckCancel() is called during the loop. Then configure code analysis to not allow loops anywhere else in the file.

@lesaltzm Could we use our code analysis tool for either of these?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
bryancroteau-MSFT: This causes some pain in the standard error handling wrappers, because the library functions sometimes omit 2/4 parameters.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
```diff
@@ -130,13 +131,14 @@ private static bool IsInvalidDouble(double number)
             where T : FormulaValue
         {
             return StandardErrorHandling<T>(expandArguments, replaceBlankValues, checkRuntimeTypes, checkRuntimeValues, returnBehavior, (runner, symbolContext, irContext, args) =>
-            {
-                return targetFunction(irContext, args);
+            {                
+                var result = targetFunction(irContext, args);
```
bryancroteau-MSFT, line 135: So we are enforcing the fact that functions which do not accept the EvalVisitor/Runner must always be sync.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
MikeStall: We're employing a cooperative multitasking approach here - where code is instrumented with checks (ie, the code "cooperates" with us)
[https://en.wikipedia.org/wiki/Cooperative_multitasking](https://en.wikipedia.org/wiki/Cooperative_multitasking)

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
MikeStall: I'd vote for the first.  

But we need to write a custom analyzer for it - so that's a separate task. (Even the 2nd item will still require a custom analyzer).

Felx-B:
> @Felx-B - can you file a new issue about IPowerFxScopeFactory? We can tackle that separately - those changes would be in separatee files from these. Do you have any feedback on additional test cases to add here?

Ok, I'll do a separate issue for this.

I think it would be interesting to add a test case with concurrent formula evaluations.
To illustrate, with the use case I showed with the gif : 
I aim to evaluate every values in parallel and then wait for all task to be completed. 
I don't know yet if the RecalcEngine would support this, and how thread safe it is.

MikeStall:
good idea - recalc engine is not thread safe - but multiple recalc engines should be able to interleave on a single thread. 

---
In reply to: [1078386562](https://github.com/microsoft/Power-Fx/pull/247#issuecomment-1078386562) [](http://example.com/codeflow?ancestors=1078386562,1078386562)

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineAsyncTests.cs
```diff
@@ -76,9 +76,10 @@ public async Task MultipleAsync()
             Assert.Equal(12.0, result.ToObject());
         }
 
-        private class Helper
+        // Helper for creating a function that waits, and then returns 2x the resu;t
```
bryancroteau-MSFT, line 79: super nid: resu;t

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineAsyncTests.cs
```diff
@@ -0,0 +1,340 @@
...
+            await Assert.ThrowsAsync<OperationCanceledException>(async () =>
+            {
+                await engine.EvalAsync(expr, cts.Token);
+            });
+        }
+
+        // Test interleaved concurrent runs. 
+        // RecalcEngine is single threaded, but the same engine should be able to do multiple evals.
+        [Fact]
+        public async Task ConcurrentInterleave()
```
MikeStall, line 200: >ConcurrentInterleave [](http://example.com/codeflow?start=26&length=20)

@Felx-B - is this what you had in mind?

bryancroteau-MSFT (approved):

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineAsyncTests.cs
Felx-B: Yes, it looks like the way we will use it
Thank you

src/libraries/Microsoft.PowerFx.Interpreter/ParsedExpression.cs
almineev: ^^ should be in the summary of the ValueTask<> class?

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
almineev: We can probably come up with "CancellableLoop" or "CancellableForeach" primitive for this purpose

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -85,30 +103,34 @@ public override FormulaValue Visit(TableNode node, SymbolContext context)
...
         {
-            var val = node.Child.Accept(this, context);
+            var val = await node.Child.Accept(this, context);
             return val;
         }
 
-        public override FormulaValue Visit(CallNode node, SymbolContext context)
+        public override async ValueTask<FormulaValue> Visit(CallNode node, SymbolContext context)
         {
+            CheckCancel();
```
almineev, line 132: It feels like pattern for CheckCancel is more of an art rather than a pattern. Here we are calling it at the beginning of the method. In other places we are not calling it at all, and then we also call it in the loop.

Maybe we can agree that every visitor method needs to call it at the beginning. Then most loops are also covered (as the call Accept inside). ?

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -419,17 +464,18 @@ public override FormulaValue Visit(ErrorNode node, SymbolContext context)
...
-        public override FormulaValue Visit(ColorLiteralNode node, SymbolContext context)
+        public override async ValueTask<FormulaValue> Visit(ColorLiteralNode node, SymbolContext context)
         {
             return CommonErrors.NotYetImplementedError(node.IRContext, "Color literal");
         }
 
-        public override FormulaValue Visit(ChainingNode node, SymbolContext context)
+        public override async ValueTask<FormulaValue> Visit(ChainingNode node, SymbolContext context)
         {
+            CheckCancel();
```
almineev, line 474: Referring to my previous comment with sporadic appearance of CheckCancel - why do we check here and not in the method above? If I'm a new dev adding a new method, how do I decide to call CheckCancel or not?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -17,12 +18,16 @@ namespace Microsoft.PowerFx.Functions
 {
     internal static partial class Library
     {
-        public delegate FormulaValue FunctionPtr(EvalVisitor runner, SymbolContext symbolContext, IRContext irContext, FormulaValue[] args);
+        // Sync FunctionPtr - all args are evaluated before invoking this function.
```
almineev, line 21: Add this comment to ///summary :)
costs nothing but may save of F12s

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1381,7 +1391,7 @@ public static FormulaValue Switch(EvalVisitor runner, SymbolContext symbolContex
             for (var i = 1; i < args.Length - 1; i += 2)
```
almineev, line 1391: Can't find CheckCancel in this loop

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1436,18 +1446,20 @@ public static FormulaValue Switch(EvalVisitor runner, SymbolContext symbolContex
...
         {// Streaming 
             var arg0 = (TableValue)args[0];
             var arg1 = (LambdaFormulaValue)args[1];
 
-            var rows = LazyForAll(runner, symbolContext, arg0.Rows, arg1);
+            var rowsAsync = LazyForAll(runner, symbolContext, arg0.Rows, arg1);
 
             // TODO: verify semantics in the case of heterogeneous record lists
-            return new InMemoryTableValue(irContext, StandardTableNodeRecords(irContext, rows.ToArray()));
+            var rows = await Task.WhenAll(rowsAsync);
```
almineev, line 1457: I wonder if it messes with natural order of rows (likely not, but would be good to check)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
almineev: Still private/public reads odd - can we normalize to one or another? Think of the "next guy" trying to infer hidden meaning from the difference :)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryLogical.cs
```diff
@@ -38,11 +39,13 @@ public static FormulaValue And(EvalVisitor runner, SymbolContext symbolContext,
         }
 
         // Lazy evaluation 
-        public static FormulaValue Or(EvalVisitor runner, SymbolContext symbolContext, IRContext irContext, FormulaValue[] args)
+        public static async ValueTask<FormulaValue> Or(EvalVisitor runner, SymbolContext symbolContext, IRContext irContext, FormulaValue[] args)
         {
             foreach (var arg in args)
             {
-                var res = runner.EvalArg<BooleanValue>(arg, symbolContext, arg.IRContext);
+                runner.CheckCancel();
```
almineev, line 46: here but not above (And) drives me nuts :), will shut up though :))

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -151,7 +152,7 @@ private static FormulaValue RunAggregator(IAggregator agg, EvalVisitor runner, S
                 if (row.IsValue)
                 {
                     var childContext = context.WithScopeValues(row.Value);
-                    var value = arg1.Eval(runner, childContext);
+                    var value = arg1.EvalAsync(runner, childContext).Result;
```
almineev, line 155: Why is RunAggregator not async?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -216,7 +222,7 @@ public static FormulaValue SortTable(EvalVisitor runner, SymbolContext symbolCon
                 if (row.IsValue)
                 {
                     var childContext = symbolContext.WithScopeValues(row.Value);
-                    return new KeyValuePair<DValue<RecordValue>, FormulaValue>(row, arg1.Eval(runner, childContext));
+                    return new KeyValuePair<DValue<RecordValue>, FormulaValue>(row, arg1.EvalAsync(runner, childContext).Result);
```
almineev, line 225: await?

src/libraries/Microsoft.PowerFx.Interpreter/ParsedExpression.cs
almineev: Please please put this into a comment. Always think of the "next guy" wondering what's going on, save their brain cycles :)

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngineWorker.cs
```diff
@@ -69,9 +70,9 @@ private void RecalcWorker2(string name)
                 (var irnode, var ruleScopeSymbol) = IRTranslator.Translate(binding);
 
                 var scope = this;
-                var v = new EvalVisitor(_cultureInfo);
+                var v = new EvalVisitor(_cultureInfo, CancellationToken.None);
```
almineev, line 73: When is RecalcAsync coming?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
MikeStall: good point - will double check

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
MikeStall: next PR.. this is part 1

PR #248, opened 2022-03-24
Make TexlNode public

ivanradicek:
Convert `TexlNode` class from internal to public, in order to remove some of `InternalsVisibleTo` projects. This also includes making public some of dependent types.


src/libraries/Microsoft.PowerFx.Core/AssemblyProperties.cs
```diff
@@ -30,8 +30,6 @@
...
 [assembly: InternalsVisibleTo("PowerFx.Dataverse.Tests, PublicKey=002400000c800000140100000602000000240000525341310008000001000100613399aff18ef1a2c2514a273a42d9042b72321f1757102df9ebada69923e2738406c21e5b801552ab8d200a65a235e001ac9adc25f2d811eb09496a4c6a59d4619589c69f5baf0c4179a47311d92555cd006acc8b5959f2bd6e10e360c34537a1d266da8085856583c85d81da7f3ec01ed9564c58d93d713cd0172c8e23a10f0239b80c96b07736f5d8b022542a4e74251a5f432824318b3539a5a087f8e53d2f135f9ca47f3bb2e10aff0af0849504fb7cea3ff192dc8de0edad64c68efde34c56d302ad55fd6e80f302d5efcdeae953658d3452561b5f36c542efdbdd9f888538d374cef106acf7d93a4445c3c73cd911f0571aaf3d54da12b11ddec375b3")]
-[assembly: InternalsVisibleTo("Microsoft.ProgramSynthesis.MultiModality.PowerApps, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
```
ivanradicek, line 33: This needs to be reverted before merging.

src/libraries/Microsoft.PowerFx.Core/Lexer/BinaryOp.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Microsoft.PowerFx.Core.Lexer
 {
-    internal enum BinaryOp
+    public enum BinaryOp
```
MikeStall, line 6: >public [](http://example.com/codeflow?start=4&length=6)

nit - can you add the /// comments for these new public types (it's probably giving a warning for them)

src/libraries/Microsoft.PowerFx.Core/Syntax/Identifier.cs
```diff
@@ -7,14 +7,18 @@
...
-        public readonly DPath Namespace;
+        internal readonly Token AtToken; // The "@" token, if any. May be null.
+        internal readonly IdentToken Token;
 
-        public Identifier(DPath theNamespace, Token atToken, IdentToken tok)
+        public DName Name { get; }
+
+        public DPath Namespace { get; }
+
+        public bool HasAtToken => AtToken != null;
```
MikeStall, line 19: >HasAtToken [](http://example.com/codeflow?start=20&length=10)

Can this stay private (ie, who uses it)

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/NumLitNode.cs
```diff
@@ -10,26 +10,30 @@
...
+    public sealed class NumLitNode : TexlNode
     {
         // If Value is non-null, then the token represents its value.
         // Otherwise, the value is in NumValue.
-        public readonly double NumValue;
+        internal readonly double NumValue;
 
-        public NumLitNode(ref int idNext, NumLitToken tok)
+        // This seams like a reasonable public interface
+        // TODO: Should we do this?
```
MikeStall, line 20: Yes, this is good.

MikeStall:
Could you add a test that uses reflection to verify some of the rules here for the TextlNode* derived classes?

- no fields
- ctors are internal
- classes are either abstract or sealed. 

We do something similar in this test for checking suspicious threading behavior:
[https://github.com/microsoft/Power-Fx/blob/main/src/tests/Microsoft.PowerFx.Core.Tests/AnalyzeThreadSafety.cs](https://github.com/microsoft/Power-Fx/blob/main/src/tests/Microsoft.PowerFx.Core.Tests/AnalyzeThreadSafety.cs)

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/VariadicBase.cs
```diff
@@ -11,12 +11,13 @@
 namespace Microsoft.PowerFx.Core.Syntax.Nodes
 {
     // Base for all variadic nodes.
-    internal abstract class VariadicBase : TexlNode
+    public abstract class VariadicBase : TexlNode
     {
-        public readonly TexlNode[] Children;
+        // TODO: Would IReadOnlyList<TexlNode> work better here?
+        public TexlNode[] Children { get; }
```
MikeStall, line 17: Yes, Array is bad because it's mutable. 

IReadOnlyList is the right one (provides both Count and Index)

That reflection test could catch this!

MikeStall:
>     [ThreadSafeImmutable]

here's an example of the immutable attr that we can put on texl nodes.

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Utils/DName.cs:11 in 2e1230d. [](commit_id = 2e1230da98bf0e4337a8f53157840fff47ceab0f, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/DottedNameNode.cs
```diff
@@ -12,28 +12,38 @@
...
+        /// <summary>
+        /// True if the name uses bangs, e.g. A!B!C.
+        /// </summary>
         public bool UsesBang => Token.Kind == TokKind.Bang;
 
-        // True if the name uses brackets, e.g. A[B]
+        /// <summary>
+        /// True if the name uses brackets, e.g. A[B].
+        /// </summary>
         public bool UsesBracket => Token.Kind == TokKind.BracketOpen;
```
lesaltzm, line 44: What is this feature? Maybe keep internal until we understand how it's used or if it's removable?

lesaltzm: Same for HasOnlyIdentifiers/HasPossibleNamespaceQualifier

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/ReplaceableNode.cs
```diff
@@ -9,9 +9,9 @@
 
 namespace Microsoft.PowerFx.Core.Syntax.Nodes
 {
-    internal sealed class ReplaceableNode : TexlNode
+    public sealed class ReplaceableNode : TexlNode
     {
-        public ReplaceableNode(ref int idNext, ReplaceableToken tok)
```
lesaltzm, line 14: >ReplaceableNode [](http://example.com/codeflow?start=15&length=15)

This node type was for a feature that never shipped and should be cleaned up.

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/SelfNode.cs
```diff
@@ -9,34 +9,37 @@
 
 namespace Microsoft.PowerFx.Core.Syntax.Nodes
 {
-    internal sealed class SelfNode : NameNode
+    public sealed class SelfNode : NameNode
```
lesaltzm, line 12: >SelfNode [](http://example.com/codeflow?start=24&length=8)

Note: Self and Parent should not be special node types. Before GA, we should remove these nodes, and just make them special reserved identifiers in Canvas.

lesaltzm: added #257

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/ReplaceableNode.cs
MikeStall: Are you saying we should we remove this class altogether?

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/SelfNode.cs
lesaltzm: Independent followup, needed before GA after this is checked in.

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/ReplaceableNode.cs
lesaltzm: Yes, although that can be done independently before GA if this is checked in.

ivanradicek:
> 



> Could you add a test that uses reflection to verify some of the rules here for the TextlNode* derived classes?
> 
> * no fields
> * ctors are internal
> * classes are either abstract or sealed.
> 

This part should be done.

> We do something similar in this test for checking suspicious threading behavior: https://github.com/microsoft/Power-Fx/blob/main/src/tests/Microsoft.PowerFx.Core.Tests/AnalyzeThreadSafety.cs
> 
> In fact, that threading test also introduces an [Immutable] attribute annotating. We could require all text nodes have that. And then we can update the threading test to verify that all properties of an immutable type are immutable (which would catch the case of returning arrays).

For this one, I'm not 100% sure. If I understand correctly, the file you mention above checks only static fields. Perhaps we should do this:
- Update the tests that I have added above to also require `[ThreadSafeImmutable]` on all `Texl*` classes
- Add new test that would check that all public fields/properties return only immutable types (for all types with `[ThreadSafeImmutable]`).

Does that make sense?

src/tests/Microsoft.PowerFx.Core.Tests/PublicSurfaceTests.cs
```diff
@@ -116,5 +148,72 @@ public void Test()
             // Types we expect to be in the assembly are all there. 
             Assert.Empty(allowed);
         }
+
+        [Fact]
+        public void TestTexlNodeTypes()
```
MikeStall, line 153: >TestTexlNodeTypes [](http://example.com/codeflow?start=20&length=17)

Yes!!!

MikeStall (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Utils/DPath.cs
```diff
@@ -11,7 +11,10 @@
 
 namespace Microsoft.PowerFx.Core.Utils
 {
-    // A path is essentially a list of simple names, starting at "root".
+    /// <summary>
+    /// A list of simple names (<see cref="DName" />), starting at "root" (<see cref="Root" />).
+    /// </summary>
+    /// // TASK: 67008 - Make this public, or expose a public shim in Document.
```
lesaltzm, line 17: nit: remove this, you're making it public so the ancient todo is no longer needed

src/libraries/Microsoft.PowerFx.Core/Utils/DPath.cs
```diff
@@ -107,16 +113,36 @@ private void AssertValid()
...
 
-        internal DPath Parent => _node == null ? this : new DPath(_node.Parent);
-
-        internal DName Name => _node == null ? default : _node.Name;
-
-        internal int Length => _node == null ? 0 : _node.Length;
-
-        internal bool IsRoot => _node == null;
-
+        /// <summary>
```
lesaltzm, line 116: All of these should be internal, no need to expose them unless you've got a use for them. In general, access should be through ToDotted, operator[] or the Segments function.

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/ReplaceableNode.cs
lesaltzm: Resolved by #273

src/tests/Microsoft.PowerFx.Core.Tests/ImmutabilityTests.cs
```diff
@@ -0,0 +1,71 @@
...
+                        errors.AppendLine($"{fieldName} is not readonly");
+                    }
+
+                    if (!AnalyzeThreadSafety.IsTypeImmutable(field.FieldType))
+                    {
+                        errors.AppendLine($"{fieldName} type ({field.FieldType}) is mutable");
+                    }
+                }
+
+                // TODO: Probably does not make sense to check methods? They could return mutable stuff
```
MikeStall, line 62: >// TODO: Probably does not make sense to ch [](http://example.com/codeflow?start=16&length=43)

If a method is "mutable", there must be a mutable field behind it.  In other words, if there are no mutable fields, there can't be any mutable methods.

MikeStall (approved):
:shipit:

Issue #250, opened 2022-03-24
"Invocation of unknown or unsupported function." should have function name 


MikeStall:
Error message is coming from here: 

https://github.com/microsoft/Power-Fx/blob/bfb78ead6c096a8fdf4c8121ba93acf650b2135e/src/strings/en-US/PowerFxResources.resw#L1632

This is similar to https://github.com/microsoft/Power-Fx/commit/a6b42afc11935c481813ff634a50154889d228f1

PR #252, opened 2022-03-24
Add name of invoked unknown function to error message

rick-nguyen:
https://github.com/microsoft/Power-Fx/issues/250


src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4631,7 +4631,7 @@ public override bool PreVisit(CallNode node)
                 var overloads = LookupFunctions(funcNamespace, node.Head.Name.Value);
                 if (!overloads.Any())
                 {
-                    _txb.ErrorContainer.Error(node, TexlStrings.ErrUnknownFunction);
+                    _txb.ErrorContainer.Error(node, TexlStrings.ErrUnknownFunction, node.Head.Name.Value);
```
Shpakh-MSFT, line 4634: >node.Head.Name.Value [](http://example.com/codeflow?start=84&length=20)

If the funcNamespace is not global or empty then you may want to include it here.

MikeStall:
Is there a test update? IF not, we should add a test. 
(PAClient will likely require a test update)

MikeStall (changes requested):
:clock1:

rick-nguyen:
Added a couple tests

MikeStall (approved):
:shipit:

PR #253, opened 2022-03-24
Add utility to ensure names meet PowerFx display name requirements

lesaltzm:


MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Interpreter/Environment/OptionSet.cs
```diff
@@ -28,11 +28,14 @@ public class OptionSet : IExternalOptionSet
...
-        public OptionSet(string name, IDictionary<string, string> options)
+        /// <param name="options">The members of the option set. Enumerable of pairs of logical name to display name.
+        /// Consider using <see cref="DisplayNameUtility.MakeUnique(IEnumerable{KeyValuePair{string, string}})"/> 
+        /// to ensure that display and logical names for options are unique.
+        /// </param>
+        public OptionSet(string name, IEnumerable<KeyValuePair<DName, DName>> options)
         {
             EntityName = new DName(name);
-            Options = ImmutableDictionary.CreateRange(options.Select(kvp => new KeyValuePair<DName, DName>(new DName(kvp.Key), new DName(kvp.Value))));
+            Options = ImmutableDictionary.CreateRange(options);
```
ian-legler, line 38: Is there a valid scenario where we would want to allow non-unique option sets at this level?  Should MakeUnique be called internally and any issues handled?

lesaltzm: No, and we will throw if that happens. However, hosts should be the one choosing how they handle display names. MakeUnique is a utility we're providing, but if a host wants to disambiguate display names differently, they can do that.

ian-legler (approved):

PR #255, opened 2022-03-25
Better numeric comparison.

MikeStall:
Needed for allowing large number expressions like  Exp(430)

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests/TestRunnerTests/TestRunnerTests.cs
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/TestRunnerTests/TestRunnerTests.cs
@@ -98,59 +98,49 @@ namespace Microsoft.PowerFx.Core.Tests
             Assert.Equal(2, passed);
         }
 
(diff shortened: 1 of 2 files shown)


src/tests/Microsoft.PowerFx.Core.Tests/TestRunnerTests/TestRunnerTests.cs
```diff
@@ -98,6 +98,60 @@ public void TestList()
...
+            runner.Tests.Add(new TestCase
+            {
+                Input = "1.56",
+                Expected = "1.57"
+            });
+            runner.Tests.Add(new TestCase
+            {
+                Input = "1.57",
+                Expected = "1.56"
+            });
```
CarlosFigueiraMSFT, line 149: Let's add tests with very large numbers as well. 5.5e186 vs 5.6e186 (the example that you have in BaseRunner.cs) should fail (it's a significant difference). Other interesting cases: positive + negative (5.5e186 vs 5.5e-186; 5.5e186 vs. -5.5e186)

PR #258, opened 2022-03-25
Make DPath public and add tests

lesaltzm:
This cleans up DPath's interface, adds tests, and enforces that DPath is always printed escaped and with `.` separators (removing deprecated ! separator support from dpath)


src/libraries/Microsoft.PowerFx.Core/Utils/DPath.cs
```diff
@@ -137,15 +139,22 @@ private void AssertValid()
...
         }
 
         public DPath Append(DPath path)
         {
             AssertValid();
-            path.AssertValid();
+            if (!path.IsValid)
+            {
+                throw new ArgumentException("Invalid DPath", nameof(path));
+            }
```
Shpakh-MSFT, line 157: Can you add like CheckValid which does this instead of duplicating this check?

Shpakh-MSFT, line 157: You can add CheckValid(path.IsValid, $"Invalid DPath {nameof(path)}");

src/libraries/Microsoft.PowerFx.Core/Utils/DPath.cs
```diff
@@ -137,15 +139,22 @@ private void AssertValid()
         }
 
         public readonly DPath Append(DName name)
-        {
-            Contracts.Assert(name.IsValid);
+        {            
+            if (!name.IsValid)
+            {
+                throw new ArgumentException("Invalid DName", nameof(name));
+            }
```
Shpakh-MSFT, line 146: CheckValid like below.

src/libraries/Microsoft.PowerFx.Core/Utils/DPath.cs
```diff
@@ -12,12 +12,10 @@
...
 {
     // A path is essentially a list of simple names, starting at "root".
-    // TASK: 67008 - Make this public, or expose a public shim in Document.
     [ThreadSafeImmutable]
-    internal struct DPath : IEquatable<DPath>, ICheckable
+    public struct DPath : IEquatable<DPath>, ICheckable
     {
-        public const char RootChar = '\u2202';
-        private const string RootString = "\u2202";
+        private const char RootChar = '\u2202';
```
Shpakh-MSFT, line 18: >private const char RootChar = '\u2202'; [](http://example.com/codeflow?start=8&length=39)

Not used anywhere

Shpakh-MSFT (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Utils/DPath.cs
```diff
@@ -258,39 +214,23 @@ public string ToDottedSyntax(string punctuator = ".", bool escapeInnerName = fal
...
                 sb.Append(escapedName);
-                sep = punctuator;
+                sep = TexlLexer.PunctuatorDot;
             }
 
             return sb.ToString();
         }
 
-        // Convert a path specified as a string to a DPath.
-        // Does not support individual path segments that contain '.' and '!' characters.
```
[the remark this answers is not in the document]
lesaltzm: Sadly no, that's going to stick around, I think. Just not on this interface

Shpakh-MSFT (approved):
:shipit:

src/tests/Microsoft.PowerFx.Core.Tests/DPathTests.cs
```diff
@@ -0,0 +1,55 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.PowerFx.Core.Utils;
+using Xunit;
+
+namespace Microsoft.PowerFx.Core.Tests
+{
+    public class DPathTest
```
MikeStall, line 12: >DPathTest [](http://example.com/codeflow?start=17&length=9)

Another case (From DPathTEst.cs in Paclient)
        [TestMethod, Owner("hekum")]
        public void TryParseEmptyString()
        {
            DPath result;
            Assert.IsTrue(DPath.TryParse("", out result));
            Assert.AreEqual(DPath.Root, result);
        }

src/libraries/Microsoft.PowerFx.Core/Utils/DPath.cs
```diff
@@ -258,39 +206,23 @@ public string ToDottedSyntax(string punctuator = ".", bool escapeInnerName = fal
...
-        public static bool TryParse(string dotted, out DPath path)
```
MikeStall, line 271: >TryParse [](http://example.com/codeflow?start=27&length=8)

Are we removing TryParse completely?

src/libraries/Microsoft.PowerFx.Core/Utils/DPath.cs
```diff
@@ -179,67 +180,14 @@ public DPath Append(DPath path)
             return new DPath(node);
         }
 
-        public DPath GoUp(int count)
```
MikeStall, line 182: > public DPath GoUp(int count) [](http://example.com/codeflow?start=7&length=29)

Are we removing this? It looks like it's still used.

MikeStall (changes requested):
:clock1:

src/tests/Microsoft.PowerFx.Core.Tests/DPathTests.cs
MikeStall: Although I see we're removing TryParse, so maybe this test doesn't matter.

lesaltzm: Yeah, deliberately removed. See corresponding PR in Client Repo.

MikeStall: ok - there's still that other question about GoUp

src/libraries/Microsoft.PowerFx.Core/Utils/DPath.cs
lesaltzm: yes, unused except by tests, and not a safe implementation.

src/libraries/Microsoft.PowerFx.Core/Utils/DPath.cs
lesaltzm: Yup, removed. Thanks, I missed that during testing, the PA Client build hid the errors.

MikeStall (approved):
:shipit:

PR #259, opened 2022-03-26
Fix inline table+record bug

lesaltzm:
The [] syntax always creates a single column table, even if it contains records. Adds capability to force single column tables to the helper functions, and adds test coverage for this use.


MikeStall (approved):
:shipit:

PR #260, opened 2022-03-27
Proposal for allowing extending of marshaller and provider

jorisdg:
This is what I needed to be able to do what I wanted.


src/libraries/Microsoft.PowerFx.Interpreter/Marshal/ObjectMarshaller.cs
```diff
@@ -59,7 +59,7 @@ internal bool TryGetField(object source, string name, out FormulaValue fieldValu
             return false;
         }
 
-        internal IEnumerable<NamedValue> GetFields(object source)
+        protected internal virtual IEnumerable<NamedValue> GetFields(object source)
```
MikeStall, line 62: >source [](http://example.com/codeflow?start=76&length=6)

I think we could actually make this non-virtual by just relying on the property enumeration from the FormulaType, and then calling Marshal() for each one.

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/ObjectMarshallerProvider.cs
```diff
@@ -50,6 +60,30 @@ public bool TryGetMarshaller(Type type, TypeMarshallerCache cache, int maxDepth,
 
             var fxType = new RecordType();
 
+            fxType = GetProperties(type, cache, maxDepth, mapping, fxType);
```
MikeStall, line 63: >mapping [](http://example.com/codeflow?start=58&length=7)

so this populates the mapping (as a side-effect) and then returns the fxType? effectively 2 out params.

MikeStall:
>     public class ObjectMarshallerProvider : ITypeMarshallerProvider

note: We want to dervie form OMP rather than create a new ITMP because the cache builder methods take a OMP. 

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Marshal/ObjectMarshallerProvider.cs:23 in eca2c50. [](commit_id = eca2c50952749fae1ef14975944c203fd728c8f6, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/ObjectMarshaller.cs
jorisdg: Having control over this list of fields is important. It allows you to expose fields that aren't properties, or expose properties from sub-classes as if they were properties on this object itself.

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/ObjectMarshallerProvider.cs
jorisdg: Yeah I just extracted the existing logic out of the TryGetMarshaller into its own (virtual) method so it can be extended.

jorisdg:
So to give the exact examples I'm using this for.
1) I have an object that has a few structs as properties. I handle the structs explicitly with a marshaller/provider. It has to expose the public fields (not properties).
2) I have a generic that wraps objects to allow formula binding to their properties. That logic of "binding" is handled entirely in my own code. However, the generic<T> is exposed to PowerFx, but the properties it's supposed to expose are in the T property, not on the object itself.

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/ObjectMarshaller.cs
MikeStall: ack - but that's already captured in Type information.

jorisdg:
Example of how this could be used: [Marshaller.cs.txt](https://github.com/microsoft/Power-Fx/files/8477107/Marshaller.cs.txt)

The difference with https://github.com/microsoft/Power-Fx/pull/277 is that now the recalcengine has the Wrapper<T> object and not just the T object. So from the host, you can do this:

Engine.GetValue("x").Source which will give you the Wrapper<T> and not just the T. This allows the host to wrap existing types T with extra functionality to deal with Power Fx.

MikeStall:
@Jorisdg - #277 is updated to deal with Wrapper<T> and now merged.

MikeStall:
Samples added, closing out old issues.

PR #261, opened 2022-03-28
Create async .txt tests.

MikeStall:
Use .txt for testing async - particularly to focus on parallelization.


MikeStall:
>     public class RecalcEngineAsyncTests

probably quite a few of these could be converted to the next .txt format. 

---
Refers to: src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineAsyncTests.cs:26 in 95619df. [](commit_id = 95619dfea04b6732f02cd53754ae112b7ac792e0, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -294,37 +291,62 @@ private static bool IsValueTypeErrorOrBlank<T>(FormulaValue val)
...
-                }
+                var task = LazyFilterRowAsync(runner, context, row, filter);
+                tasks.Add(task);
             }
+
+            // WhenAll will allow running tasks in parallel. 
+            var results = await Task.WhenAll(tasks);
+
+            // Remove all nulls. 
+            var final = results.Where(x => x != null);
```
CarlosFigueiraMSFT, line 347: >var final = results.Where(x => x != null); [](http://example.com/codeflow?start=12&length=42)

This should not be the condition - an expression such as `Filter(Table({a:1},{a:2},If(1<0,{a:3}),{a:4},{a:5}), true)` should return a table with 5 rows (one of them empty).

CarlosFigueiraMSFT: Just noticed that it is a current bug, not related to your changes... Created issue #263 to track it

MikeStall: thanks - I believe this is preserving the existing logic (right or wrong) and just making it async.

src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterExpressionTestCases/ParallelAsync.txt
```diff
@@ -0,0 +1,29 @@
...
+>> If(Async(0) = 0, Async(1), 1/0)
+1
+
+>> ForAll([1,0], If(Async(Value) = 0, Async(2), Async(3)))
+[3,2]
+
+//-------------------------
+// Hangs - any await in a for loop is suspect.
+
+// With({x:Async(1), y: Async(0)}, x & y)
```
CarlosFigueiraMSFT, line 26: >With({x:Async(1), y: Async(0)}, x & y) [](http://example.com/codeflow?start=3&length=38)

It it a problem with the With function (which doesn't evaluate the creation of the record fields in parallel), or the concatenation function? Would something like `With({x:Async(1), y:Async(0)}, x)` work? What about `Async(1) & Async(0)`?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -294,37 +291,62 @@ private static bool IsValueTypeErrorOrBlank<T>(FormulaValue val)
...
+                if (include)
+                {
+                    return row;
+                }
+            }
+
+            return null;
+        }
+
+        private static async Task<DValue<RecordValue>[]> LazyFilterAsync(
```
CarlosFigueiraMSFT, line 325: >Task<DValue<RecordValue>[]> [](http://example.com/codeflow?start=29&length=27)

Instead of returning a (task of) array, can we return a task of an enumerable? This way something like `LookUp([0, 10], Async(Value) >= 0)` would complete (today I'm guessing it would hang)

src/tests/Microsoft.PowerFx.Interpreter.Tests/AsyncFunctionsHelper.cs
```diff
@@ -0,0 +1,79 @@
...
+using Microsoft.PowerFx.Core.Localization;
+using Microsoft.PowerFx.Core.Public;
+using Microsoft.PowerFx.Core.Public.Types;
+using Microsoft.PowerFx.Core.Public.Values;
+using Microsoft.PowerFx.Core.Tests;
+using Microsoft.PowerFx.Core.Texl;
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.Utils;
+using Xunit;
+using static Microsoft.PowerFx.Core.Localization.TexlStrings;
```
CarlosFigueiraMSFT, line 21: Can you sort/remove usings? This seems too much

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
```diff
@@ -47,6 +47,19 @@ public void InterpreterTestCase(ExpressionTestCase testCase)
             }
         }
 
+        // Helpful for debugging. Can run a single test.
```
CarlosFigueiraMSFT, line 50: >// Helpful for debugging. Can run a single test.  [](http://example.com/codeflow?start=8&length=49)

We used to have a `RunSingleTestCase` method in this file, which is referenced in the file Testing.txt. Can you either revive this method, or remove Testing.txt?

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -51,6 +53,16 @@ private static (RecalcEngine engine, RecordValue parameters) OptionSetTestSetup(
             return (new RecalcEngine(config), parameters);
         }
 
+        private static (RecalcEngine engine, RecordValue parameters) AsyncSetTestSetup()
```
CarlosFigueiraMSFT, line 56: >AsyncSetTestSetup [](http://example.com/codeflow?start=69&length=17)

Each "setup" is removing the configuration from other setups... We probably need to find a way to "chain" those - for example, if we want to have async tests with option sets. Your call whether to do it in this PR or in a different one, though.

src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterExpressionTestCases/ParallelAsync.txt
MikeStall: Record, Table, CallNode args, Binary op nodes - none of these support full parallelism and require evaluating in order. 

Need to discuss the semantics.

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
MikeStall: I found it in dataverse tests (different erpo): \PowerApps-Client\src\Language\Tests\PowerFx.Dataverse.Tests\PowerFxEvaluationTests.cs

src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterExpressionTestCases/ParallelAsync.txt
```diff
@@ -0,0 +1,99 @@
...
+3
+
+
+// Record fields can be async. 
+// Order of multiple fields is unspecified!
+>> { x : Async(0), y: 1 }.x
+0
+
+// Call args are serial, left-to-right
+>> Sum(WaitFor(0), WaitFor(1), WaitFor(2))
```
MikeStall, line 75: >WaitFor [](http://example.com/codeflow?start=7&length=7)

Many of these WaitFor() are things we'd *like* to be parallel, but we don't need them to be yet. This is still capturing their current semantics.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -340,8 +366,8 @@ public static NamedLambda[] Parse(FormulaValue[] args)
 
                 for (var i = 1; i < args.Length; i += 2)
                 {
-                    var columnName = ((StringValue)args[1]).Value;
-                    var arg1 = (LambdaFormulaValue)args[2];
+                    var columnName = ((StringValue)args[i]).Value;
```
MikeStall, line 369: >i [](http://example.com/codeflow?start=56&length=1)

This was an unrelated bug I fixed that I hit with Async +AddColumn tests.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -290,7 +290,7 @@ private static FormulaValue RunAggregator(IAggregator agg, EvalVisitor runner, S
                 if (row.IsValue)
                 {
                     var childContext = context.WithScopeValues(row.Value);
-                    var value = arg1.EvalAsync(runner, childContext).Result;
```
MikeStall, line 293: >Result [](http://example.com/codeflow?start=69&length=6)

This was a real bug, caught by the new AsyncVerify tests.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -110,17 +111,19 @@ private static IEnumerable<DValue<RecordValue>> LazyAddColumns(EvalVisitor runne
 
                     foreach (var column in newColumns)
                     {
-                        var value = column.Lambda.EvalAsync(runner, childContext).Result;
```
MikeStall, line 113: >Result [](http://example.com/codeflow?start=82&length=6)

fixed - remove .Result.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
almineev: awesome!

lesaltzm (approved):

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -110,17 +111,19 @@ private static IEnumerable<DValue<RecordValue>> LazyAddColumns(EvalVisitor runne
...
+                        var value = await column.Lambda.EvalAsync(runner, childContext);
                         fields.Add(new NamedValue(column.Name, value));
```
almineev, line 115: var fields = new NamedValue[row.Value.Fields.Length + newColumns.Length];
.. copy row.Value.Fields by position to fields
here fields[i] = new NamedValue(...)
below remove .ToArray() below

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -294,37 +295,62 @@ private static bool IsValueTypeErrorOrBlank<T>(FormulaValue val)
...
+            return null;
+        }
+
+        private static async Task<DValue<RecordValue>[]> LazyFilterAsync(
             EvalVisitor runner,
             SymbolContext context,
             IEnumerable<DValue<RecordValue>> sources,
-            LambdaFormulaValue filter)
+            LambdaFormulaValue filter, 
+            int topN = int.MaxValue)
```
almineev, line 334: 1. Don't see topN used anywhere in the body.
2. "topN" assumes order, filter doesn't have order spec, is the meaning here to use natural order or semantics are "takeN" (take _any_ N rows)?
3. Nit: I'd go with topN as int? to be more specific when condition is not present
4. Need to be careful with async implementation. On a large rowset we don't want to evaluate filter for all rows. We want to implement strategy when we fill N requested matches with minimal filter evals. One way is to define M as batch window size and keep hopping its evaluation until N rows are filled up

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
MikeStall: @LucGenetier  - here's an example of a hang.

Issue #60, opened 2021-11-17
Help() have redundant function names 


fadyanwar:
Several function names are repeated twice on the help message output of Help(), such as Max, Sum and Average

MikeStall:
This is referring to https://github.com/microsoft/power-fx-host-samples/tree/main/Samples/ConsoleREPL  ?
This is probably showing all the overloads - ie both the table and scalar versions. 
For Help(), we could consider grouping by overload.

fadyanwar:
Yes, I've found it via the REPL console sample, but if I'm not mistaken the Help() is implemented in the core PowerFx library.

MikeStall:
Help()  is a helper function in the Console repl sample: 
https://github.com/microsoft/power-fx-host-samples/blob/417a3100b96420167004e259c06ed8c4d81d7420/Samples/ConsoleREPL/ConsoleREPL.cs#L182

The problem is engine.GetAllFunctionNames() is including overloads.

fadyanwar:
Interesting, I was under the impression that all functions are self contained in the Nugets.

MikeStall:
The built-in ones are in the core nuget. But hosts can extend with additional helpers.

PR #266, opened 2022-03-29
Split RecalcEngine into a base class that can live in FxCore and just…

MikeStall:
… has binding support. Tooling can use this to do binding.

Able to directly create a SimpleResolver from a PowerFxConfig.

Stronger type safety - a bunch of places that took FormulaType for Parameters really need a RecordType.


src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs
```diff
@@ -48,18 +50,30 @@ internal class SimpleResolver : INameResolver
 
         public IExternalEntity CurrentEntity => null;
 
-        public SimpleResolver(IEnumerable<EnumSymbol> enumSymbols, params TexlFunction[] extraFunctions)
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SimpleResolver"/> class.
+        /// </summary>
+        /// <param name="config"></param>
+        public SimpleResolver(PowerFxConfig config)
```
MikeStall, line 57: >public SimpleResolver(PowerFxConfig config) [](http://example.com/codeflow?start=8&length=43)

This is the critical ctor. We should just need a config.  If there's anything else, then that we would have passed to the ctor, then it ought to be in the config. 

Remove the other ctors.

MikeStall:
>             }

This check can be moved to SimpleResolver, and have it take the Parameters. 

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngineResolver.cs:62 in b4a7958. [](commit_id = b4a79585df0c543729b9327a026816cff4147263, deletion_comment = False)

MikeStall:
>             }

Also move this check to SimpleResolver. Anything based on _config can be in simple resolver. 

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngineResolver.cs:98 in b4a7958. [](commit_id = b4a79585df0c543729b9327a026816cff4147263, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/BindingEngineTests.cs
```diff
@@ -0,0 +1,78 @@
...
+using Microsoft.PowerFx.Core;
+using Microsoft.PowerFx.Core.Public.Types;
+using Xunit;
+
+namespace Microsoft.PowerFx.Tests
+{
+    public class BindingEngineTests
+    {
+        [Fact]
+        public void CheckSuccess()
```
MikeStall, line 14: >CheckSuccess [](http://example.com/codeflow?start=20&length=12)

Allow tools to bind expressions with just FxCore (and not have to pull in interpreter).

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -0,0 +1,144 @@
...
+using Microsoft.PowerFx.Core.Texl.Intellisense;
+using Microsoft.PowerFx.Core.Types;
+
+namespace Microsoft.PowerFx
+{
+    /// <summary>
+    /// Expose binding logic for Power Fx. 
+    /// Derive from this to provide evaluation abilities. 
+    /// </summary>
+    public class Engine : IPowerFxEngine
```
MikeStall, line 22: >Engine [](http://example.com/codeflow?start=17&length=6)

beter name? "BindingEngine"?

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -125,70 +126,6 @@ public void UpdateVariable(string name, FormulaValue value)
...
-            // Ok to continue with binding even if there are parse errors. 
-            // We can still use that for intellisense. 
-
-            var resolver = new RecalcEngineResolver(this, _powerFxConfig, (RecordType)parameterType);
-
-            var binding = TexlBinding.Run(
-                new Glue2DocumentBinderGlue(),
-                formula.ParseTree,
-                resolver,
-                ruleScope: parameterType._type,
```
MikeStall, line 159: >uleScope: parameterType._type, [](http://example.com/codeflow?start=17&length=30)

'intellisense' parameter is no longer used. It used to be used on this line but was removed. So just remove the parameter altogether.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -125,70 +126,6 @@ public void UpdateVariable(string name, FormulaValue value)
             Recalc(name);
         }
 
-        /// <summary>
-        /// Type check a formula without executing it. 
-        /// </summary>
-        /// <param name="expressionText"></param>
-        /// <param name="parameters"></param>
-        /// <returns></returns>
-        public CheckResult Check(string expressionText, FormulaType parameterType = null)
```
MikeStall, line 134: >Check [](http://example.com/codeflow?start=27&length=5)

This moves to the base class where it can be invoked from just FxCore.

MikeStall:
>         public string GetInvariantExpression(string expressionText, RecordType parameters)

Looks like more non-eval methods in here ... Can any of these other methods be moved to the base class?


---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs:163 in b4a7958. [](commit_id = b4a79585df0c543729b9327a026816cff4147263, deletion_comment = False)

MikeStall:
>             }

Actually - it looks like this can be removed altogether?  Already handled by Binder's Visit(FirstNameNode node) as a RowScope. 

---
In reply to: [1081386066](https://github.com/microsoft/Power-Fx/pull/266#issuecomment-1081386066) [](http://example.com/codeflow?ancestors=1081386066)

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngineResolver.cs:62 in b4a7958. [](commit_id = b4a79585df0c543729b9327a026816cff4147263, deletion_comment = False)

MikeStall:
>             }

@lesaltzm  - it looks like I can just delete this (tests still pass and it binds as rowscope). Any concerns?

---
In reply to: [1081498030](https://github.com/microsoft/Power-Fx/pull/266#issuecomment-1081498030) [](http://example.com/codeflow?ancestors=1081498030,1081386066)

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngineResolver.cs:62 in b4a7958. [](commit_id = b4a79585df0c543729b9327a026816cff4147263, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs
```diff
@@ -24,6 +24,8 @@ internal class SimpleResolver : INameResolver
     {
         protected TexlFunction[] _library;
         protected EnumSymbol[] _enums = new EnumSymbol[] { };
+
+        // $$$ This isn't used anymore, but still required by INameResolver ...
```
lesaltzm, line 28: I mean, this can be cleaned up, and change the Document prop to return null at least

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
lesaltzm: CheckerEngine maybe?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -0,0 +1,144 @@
...
+            return resolver.Functions.Select(func => func.Name);
+        }
+
+        /// <summary>
+        /// Create a resolver for use in binding. This is called from <see cref="Check(string, FormulaType)"/>.
+        /// Base classes can override this is there are additional symbols not in the config.
+        /// </summary>
+        /// <param name="parameterType"></param>
+        /// <returns></returns>
+        private protected virtual SimpleResolver CreateResolver(RecordType parameterType)
```
lesaltzm, line 54: ParamterType arg left over?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -0,0 +1,144 @@
...
+            Config = powerFxConfig;
+        }
+
+        /// <summary>
+        /// List all functions (both builtin and custom) registered with this evaluator. 
+        /// </summary>
+        public IEnumerable<string> GetAllFunctionNames()
+        {
+            var resolver = CreateResolver(null);
+            return resolver.Functions.Select(func => func.Name);
```
lesaltzm, line 45: Running .Distinct() on this will resolve #60

lesaltzm (commented):
GetInvariantExpression/GetDisplayExpression and ConvertExpression should also be movable to Core using this change.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngineResolver.cs
```diff
@@ -45,6 +45,7 @@ public override bool Lookup(DName name, out NameLookupInfo nameInfo, NameLookupP
 
             var str = name.Value;
 
+            // $$$ MOve to base class?
```
lesaltzm, line 48: eliminate, I think. Should be unused.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngineResolver.cs
```diff
@@ -29,7 +29,7 @@ internal class RecalcEngineResolver : SimpleResolver
             RecalcEngine parent,
             PowerFxConfig powerFxConfig,
             RecordType parameters)
-            : base(powerFxConfig.EnumStore.WithRequiredEnums(GetFullFunctionLibrary(powerFxConfig.ExtraFunctions.Values.ToArray())).EnumSymbols, powerFxConfig.ExtraFunctions.Values.ToArray())
+            : base(powerFxConfig)
```
lesaltzm, line 32: no need for parameters in constructor? Should be handled by row scope.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngineResolver.cs
```diff
@@ -74,6 +75,7 @@ public override bool Lookup(DName name, out NameLookupInfo nameInfo, NameLookupP
                 return true;
             }
 
+            // $$$ Should this be in base class?
```
lesaltzm, line 78: Yes.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
MikeStall: The derived classes where using this; but per our other discussion, looks like this can be removed now.

src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs
MikeStall: upon further review, derived classes are setting ...

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngineResolver.cs
```diff
@@ -29,7 +29,7 @@ internal class RecalcEngineResolver : SimpleResolver
             RecalcEngine parent,
             PowerFxConfig powerFxConfig,
             RecordType parameters)
-            : base(powerFxConfig.EnumStore.WithRequiredEnums(GetFullFunctionLibrary(powerFxConfig.ExtraFunctions.Values.ToArray())).EnumSymbols, powerFxConfig.ExtraFunctions.Values.ToArray())
```
MikeStall, line 32: >GetFullFunctionLibrary [](http://example.com/codeflow?start=61&length=22)

Calling GetFullFunctionLibrary() is problematic. It's pulling in functions that the interpreter doesn't use.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -60,22 +60,23 @@ private void AddInterpreterFunctions(PowerFxConfig powerFxConfig)
...
             powerFxConfig.AddFunction(BuiltinFunctionsCore.Boolean);
             powerFxConfig.AddFunction(BuiltinFunctionsCore.Boolean_UO);
+
+            return powerFxConfig;
         }
 
-        /// <summary>
-        /// List all functions (both builtin and custom) registered with this evaluator. 
-        /// </summary>
-        public IEnumerable<string> GetAllFunctionNames()
```
MikeStall, line 68: >GetAllFunctionNames [](http://example.com/codeflow?start=35&length=19)

subtle: this is just returning the functions *implemented* in the interpreter (which is good), not necessarily all the builtins.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -237,6 +238,22 @@ public void ChangeRecord()
             AssertUpdate("A-->12;");
         }
 
+        [Fact]
+        public void CheckFunctionCounts()
```
MikeStall, line 242: >CheckFunctionCounts [](http://example.com/codeflow?start=20&length=19)

@lesaltzm / @bryancroteau-MSFT  - this is a bit wacky so I wanted to be explicit about it in a test. 

The problem is that we have defaults, and too many different things want different defaults and for it to happen automatically.

src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs
bryancroteau-MSFT: Is this used at all in the base class?

src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs
```diff
@@ -48,27 +51,44 @@ internal class SimpleResolver : INameResolver
...
-        }
-
-        public static TexlFunction[] GetFullFunctionLibrary(TexlFunction[] extraFunctions)
-        {
-            var list = new List<TexlFunction>();
-            list.AddRange(BuiltinFunctionsCore.BuiltinFunctionsLibrary);
-            list.AddRange(extraFunctions);
-            return list.ToArray();
+            _library = config.Functions.ToArray();
+            _enums = config.EnumStore.WithRequiredEnums(_library).EnumSymbols.ToArray();
```
bryancroteau-MSFT, line 63: Actually we could make a new function here that returns the list of enums from the EnumStore as an array, augmented with the required enums, without actually modifying the EnumStore instance.

_enums = config.EnumStore.ToArrayWithRequiredEnums(_library);

src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs
```diff
@@ -48,27 +51,44 @@ internal class SimpleResolver : INameResolver
...
-            : this(extraFunctions)
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SimpleResolver"/> class.
+        /// </summary>
+        /// <param name="config"></param>
+        public SimpleResolver(PowerFxConfig config)            
         {
-            _enums = enumSymbols.ToArray();
-        }
+            _config = config ?? throw new ArgumentNullException(nameof(config));
```
bryancroteau-MSFT, line 60: Why are we keeping both the config as well as the _library and _enums (which are also stored on the config?)

src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs
```diff
@@ -48,27 +51,44 @@ internal class SimpleResolver : INameResolver
...
+            _library = config.Functions.ToArray();
+            _enums = config.EnumStore.WithRequiredEnums(_library).EnumSymbols.ToArray();
         }
-
+        
         public virtual bool Lookup(DName name, out NameLookupInfo nameInfo, NameLookupPreferences preferences = NameLookupPreferences.None)
         {
+            if (_config != null)
+            {
+                if (_config.TryGetSymbol(name, out var symbol, out var displayName))
```
bryancroteau-MSFT, line 70: Why does _config have a field called TryGetSymbol? Would it be better if PowerFx config implemented INameResolver directly?

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -51,6 +56,22 @@ internal static PowerFxConfig BuildWithEnumStore(CultureInfo cultureInfo, EnumSt
             return new PowerFxConfig(cultureInfo, enumStore);
         }
 
+        internal static PowerFxConfig BuildWithEnumStore(CultureInfo cultureInfo, EnumStore enumStore, IEnumerable<TexlFunction> coreFunctions)
```
bryancroteau-MSFT, line 59: Is there a reason we can't have the core functions and extra functions included at the same time in this constructor? Why do we ever allow people to add extra functions after the object has been constructed? Don't we want to know the entire function library up-front so that we can add the required enums to the enum store?

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -63,6 +84,15 @@ internal void AddEntity(IExternalEntity entity, DName displayName = default)
             _environmentSymbols.Add(entity.EntityName, entity);
         }
 
+        // Sets the "core" builtin functions. This can vary from host to host. 
+        // Overwrite list and set. 
+        internal void SetCoreFunctions(IEnumerable<TexlFunction> functions)
+        {
+            CheckUnlocked();
+
+            _coreFunctions = functions ?? throw new ArgumentNullException(nameof(functions));
```
bryancroteau-MSFT, line 93: As a general pattern shouldn't we be checking for null at the start of the function for each parameter and throwing exceptions there, instead of in expressions?

src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs
MikeStall: yes, sql uses it.

src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs
MikeStall: INameResolve has many different resolve hooks. 

"Would it be better if PowerFx config implemented INameResolver directly" - this PR moves us *much* closer to that possibility.  
However, there are still derived cases that need to change resolution in a way that the config doesn't handle yte.

src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs
MikeStall: They could be stored in different formats. the _library and _enums are efficient array caches. (might even be something to build as a dictionary).  But at least they've gone from protected to private state now.

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
MikeStall: "Why do we ever allow people to add extra functions after the object has been constructed" this is the common pattern.  
Config is a builder object, and then passed to Engine's ctor.  Essentially, config has enough knobs on it that passing everything at once into the ctor is problematic (and would just encourage another builder pattern to initialize the config bject). 

The pattern is:

config = new Config()
config.Add(....) // other mutations here. 
engine = new Engine(config);

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
MikeStall: Style cop told me to use the throw expression instead of statement. ;)

[https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0016](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0016)

src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs
bryancroteau-MSFT: Sql is a derived class. Is it needed in the base class or can this be moved to the derived classes?

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
bryancroteau-MSFT: But the user should know everything that they want up-front before they construct an interpreter, right? For example, let's say that you mutate a config on another thread while a binding is running. It should not affect the behavior of the binding right? For example, it should not make a "symbol not found" error stop happening halfway through type checking.

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
bryancroteau-MSFT: Don't we usually pre-check arguments to functions? For example with Contracts.Assert

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
lesaltzm: You can't mutate the function after it's passed to an Engine() constructor, see the Check guard on all the mutating functions, so that shouldn't be possible.

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
lesaltzm: Yeah, that style rule snuck in from the PVA ruleset, and we actually don't have anywhere in our code that's doing this pattern 
if (a == null) throw
var b = a;

since we've mostly used Contracts.Assert/Check for that.

lesaltzm: I like the coalescing approach, but if you don't feel free to disable that rule in editorconfig

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngineResolver.cs
MikeStall: removed the call.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
MikeStall: added test for this

MikeStall:
moved!

---
In reply to: [925307837](https://github.com/microsoft/Power-Fx/pull/266#issuecomment-925307837) [](http://example.com/codeflow?ancestors=925307837,925307837)

MikeStall:
>         }

Surprised this is nyi.  We resolve enums in the main Lookup(). 

Although looks like the only place this is called is SortOrderValidator. 



---
Refers to: src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs:197 in e781261. [](commit_id = e78126177e026f09031a55f8637041bd2629de0a, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
bryancroteau-MSFT: I don't like this pattern of locked/unlocked instance. It's not type safe. Why did we do it this way?

src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs
MikeStall: It's used by the INameResolver interface that this type implements, so it can't be moved to a derived class - we'd need to remove it from the itnerface.  (interface is like a base class here)

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
bryancroteau-MSFT: In object-oriented land, the pattern is to use some sort of "builder" and then call .Build() to finalize and build an instance that is locked. Then only the final type is used, so that it is a type-safe operation to use an instance that is locked.

MikeStall: Is that just  this:

```
config = new Config()
config.Add(....) // other mutations here.
engine = config.Build(); 
```

instead of:
```
config = new Config()
config.Add(....) // other mutations here.
engine = new Engine(config);
```

bryancroteau-MSFT: More like this:

config = new Config()
config.Add(....) // other mutations here.
engine = new Engine(config.Build());

bryancroteau-MSFT:
We can address mutability problems in a future PR.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -0,0 +1,206 @@
...
+            if (parameterType == null)
+            {
+                parameterType = new RecordType();
+            }
+
+            var formula = new Formula(expressionText);
+
+            formula.EnsureParsed(TexlParser.Flags.None);
+
+            // Ok to continue with binding even if there are parse errors.
```
ivanradicek, line 77: It might be useful to add a flag that would stop here and just report errors. In some context that might be useful for performance purposes - for example, when binding is not important if there are already syntax errors. However, this does not seem super useful right now.

PR #268, opened 2022-03-29
Change TableNode FindNodeVisitor to use the commas

bryancroteau-MSFT:
Change TableNode FindNodeVisitor to use the commas instead of the children. The previous implementation should have never worked.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Syntax/Visitors/FindNodeVisitor.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Syntax/Visitors/FindNodeVisitor.cs
@@ -270,18 +270,18 @@ namespace Microsoft.PowerFx.Core.Syntax.Visitors
             }
 
             // Cursor is between the open and closed bracket.
-            for (var i = 0; i < node.Children.Length; i++)
+            for (var i = 0; i < node.Commas.Length; i++)
             {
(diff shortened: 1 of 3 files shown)


lesaltzm (approved):

src/libraries/Microsoft.PowerFx.Core/Syntax/Visitors/FindNodeVisitor.cs
```diff
@@ -270,18 +270,18 @@ public override bool PreVisit(TableNode node)
...
                 {
                     node.Children[i].Accept(this);
                     return false;
                 }
             }
 
-            // If we got here the cursor is not within the brackets. return tableNode.
-            _result = node;
+            // If we got here the cursor should be in the last child.
+            node.Children[node.Children.Length - 1].Accept(this);
```
MikeStall, line 284: >node.Children.Length  [](http://example.com/codeflow?start=26&length=21)

Is this possible to be 0 length?

bryancroteau-MSFT: No, the previous code block checks for 3 edge cases:

Cursor before the [
Empty (0 children)
Cursor after the ]

PR #269, opened 2022-03-29
Move IExternalViewInfo to DisplayNameProvider pattern

lesaltzm:
DisplayNameProvider is the preferred approach to handling display names. This PR continues the work of several previous ones, moving away from the approach of` IDisplayMapped<T>` for another IExternal type.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalViewInfo.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalViewInfo.cs
@@ -2,13 +2,24 @@
 // Licensed under the MIT license.
 
 using System;
+using System.Collections.Generic;
+using Microsoft.PowerFx.Core.Utils;
 
 namespace Microsoft.PowerFx.Core.Entities
 {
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalViewInfo.cs
```diff
@@ -2,13 +2,24 @@
...
+        /// <summary>
+        /// Logical names for the members in this View.
+        /// </summary>
+        IEnumerable<DName> ViewNames { get; }
+
         string Name { get; }
 
         string RelatedEntityName { get; }
+
+        bool IsConvertingDisplayNameMapping { get; }
```
bryancroteau-MSFT, line 23: How is this being used?

bryancroteau-MSFT (approved):

MikeStall:
is there some corresponding test for this?

lesaltzm:
No, views aren't useable or creatable in standalone powerfx in any way at the moment. We could build mocks for it on this side if we wanted, but the real test at this point is integration in PAClient

PR #270, opened 2022-03-30
Add missing resources from DateTime function

CarlosFigueiraMSFT:
Added descriptions for the DateTime arguments.


lesaltzm (commented):
I vaguely remember having a test that validated the existence of these resources. Can we move that over or create it if I’m misremembering?

CarlosFigueiraMSFT:
> I vaguely remember having a test that validated the existence of these resources. Can we move that over or create it if I’m misremembering?

Yes, we do have a test in the other repository. I'll bring it over, this is a quick fix to unblock updating it.

lesaltzm (approved):

PR #272, opened 2022-04-04
Add a Check method overload to take Formula as input

Robulane:
This PR adds a simple method overload that takes a Formula instead of a string in RecalcEngine.

Motivation: Thanks to this change, we can re-use already parsed formula that will lead to significant performance improvements.


MikeStall:
fyi, you'll want to rebase on latest to get:
[https://github.com/microsoft/Power-Fx/commit/c38806c836960153c23076ad8561121d9ee8e388](https://github.com/microsoft/Power-Fx/commit/c38806c836960153c23076ad8561121d9ee8e388)

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -136,15 +136,30 @@ public CheckResult Check(string expressionText, FormulaType parameterType = null
...
         }
 
+        /// <summary>
+        /// Type check a formula without executing it. 
+        /// </summary>
+        /// <param name="formula">The formula to check.</param>
+        /// <param name="parameterType">The type context.</param>
+        /// <remarks>Used by PVA.</remarks>
+        /// <returns></returns>
+        internal CheckResult Check(Formula formula, FormulaType parameterType = null)
```
MikeStall, line 146: >Formula [](http://example.com/codeflow?start=35&length=7)

Note that we're tyring to move off internals, so Formula would need to be public too.

Robulane: I tried, it changes a lot of things from internal to public and starts to require the summaries...

lesaltzm: Should we go for a 2 stage approach and keep the Formula object internal?

public ParseResult {
     public TexlNode ParseTree => Formula.ParseTree;
     public ExpressionError[] Errors => Formula.GetParseErrors();
     public bool IsSuccess => Errors == null;


     private Formula _formula;
}

ParseResult Parse(string expression)

CheckResult Check(IParseResult parseResult, FormulaType parameterType = null)

lesaltzm: obviously also still keeping 
CheckResult Check(string expression, FormulaType parameterType)

Robulane: TexlNode is also internal that brings a lot of other internal stuff. Let me see if I can give this PR to the team to follow up change internal to public

ivanradicek: FYI, `TexlNode` should be public soon. See: #248.

Robulane: Oh, nice, then I'll wait for this PR to be merged :)

MikeStall: We do need to huddle about the right shape for the public surface here.  Ability to parse-only makes sense, but Formula class may not be the right public surface.

MikeStall:
Superceded by #319, which has been merged.

PR #275, opened 2022-04-05
Add parameter rename capability

lesaltzm:
Adds RenameDriver and entry point in Engine. This allows hosts to rename arbitrary fields in the parameter record, and apply that update to as many expressions as needed.


src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs
```diff
@@ -55,10 +55,10 @@ internal class SimpleResolver : INameResolver
         /// Initializes a new instance of the <see cref="SimpleResolver"/> class.
         /// </summary>
         /// <param name="config"></param>
+        /// <param name="skipGlobalDisplayNames">For some scenarios, we force lookups to only be by logical names.</param>
```
lesaltzm, line 58: remove

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -116,6 +135,16 @@ internal bool TryGetSymbol(DName name, out IExternalEntity symbol, out DName dis
             return _environmentSymbols.TryGetValue(lookupName, out symbol);
         }
 
+        /// <summary>
+        /// Some scenarios require that lookups be done with logical names only.
+        /// This returns the same PowerFxConfig with the display name provider disabled.
+        /// </summary>
+        /// <returns></returns>
+        internal PowerFxConfig WithoutDisplayNames()
```
lesaltzm, line 143: For this, see #276. It would make this cleaner.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -3,6 +3,7 @@
 
 using System.Collections.Generic;
 using System.Linq;
+using System.Security.Cryptography;
```
MikeStall, line 6: >Cryptography [](http://example.com/codeflow?start=22&length=12)

this is strange to see in a rename pr?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -47,14 +49,14 @@ public IEnumerable<string> GetAllFunctionNames()
         }
 
         /// <summary>
-        /// Create a resolver for use in binding. This is called from <see cref="Check(string, FormulaType)"/>.
+        /// Create a resolver for use in binding. This is called from <see cref="Check(string, RecordType)"/>.
         /// Base classes can override this is there are additional symbols not in the config.
         /// </summary>
-        /// <param name="parameterType"></param>
+        /// <param name="alternateConfig">An alternate config that can be provided. Should default to engine's config if null.</param>
```
MikeStall, line 55: >An alternate config  [](http://example.com/codeflow?start=42&length=20)

What is this for?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
lesaltzm: It is indeed. That shouldn't have been added.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -151,6 +153,19 @@ public IIntellisenseResult Suggest(string expression, RecordType parameterType,
...
 
+        /// <summary>
+        /// Creates a renamer instance for updating a field reference from <paramref name="parameters"/> in expressions.
+        /// </summary>
+        /// <param name="parameters">Type of parameters for formula. The fields in the parameter record can 
+        /// be acecssed as top-level identifiers in the formula. Must be the names from before any rename operation is applied.</param>
+        /// <param name="pathToRename">Path to the field to rename.</param>
+        /// <param name="updatedName">New name. Replaces the last segment of <paramref name="pathToRename"/>.</param>
+        /// <returns></returns>
+        public RenameDriver CreateFieldRenamer(RecordType parameters, DPath pathToRename, DName updatedName)
```
MikeStall, line 164: >CreateFieldRenamer [](http://example.com/codeflow?start=28&length=18)

Great that this is on the base class now! (and not just interpreter)

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -48,6 +48,22 @@ public PowerFxConfig(CultureInfo cultureInfo = null)
...
+        /// </summary>
+        /// <param name="other">Config to clone from.</param>
+        private PowerFxConfig(PowerFxConfig other)
+        {
+            _isLocked = other._isLocked;
+            _extraFunctions = other._extraFunctions;
+            _environmentSymbols = other._environmentSymbols;
+            _environmentSymbolDisplayNameProvider = other._environmentSymbolDisplayNameProvider;
+            _coreFunctions = other._coreFunctions;
+            EnumStoreBuilder = other.EnumStoreBuilder;
```
bryancroteau-MSFT, line 63: This does present a problem. There is special logic in PowerFxConfig that attempts to enforce the following axiom: Every time you add a function to the config, it updates the EnumStoreBuilder by calling WithRequiredEnums. This means that if two configs point to the same builder, they could each add functions. Collisions are impossible, but it could lead to unwanted enums. I am not sure if this is a problem in practice.

lesaltzm: Right, totally agree. For now, not an immediate issue since this is only being called from a path where it's already locked, but #276 is the right direction here, as you brought up a week ago.

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/RenameDriver.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// <returns>Expression with rename applied.</returns>
+        public string ApplyRename(string expressionText)
+        {
+            return Engine.ConvertExpression(expressionText, _parameters, _resolver, true);
+        }
+
+        private static FormulaType RenameFormulaTypeHelper(AggregateType nestedType, Queue<DName> segments, DName updatedName)
+        {
+            var field = segments.Dequeue();
+            if (segments.Count == 0)
```
bryancroteau-MSFT, line 41: Are we assuming that the queue length will be either 0 or 1, shouldn't we have a parameter check for that assumption?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
lesaltzm: Used for providing a config that doesn't have display names for globals. We want to be sure we're only changing the name of the field provided, nothing else.

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/RenameDriver.cs
bryancroteau-MSFT: Wait, I see this is recursive

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -645,6 +645,27 @@ public static DType AttachDataSourceInfo(DType type, IExternalTabularDataSource
...
+            type.AssertValid();
+            var returnType = type.Clone();
+            returnType.DisplayNameProvider = DisabledDisplayNameProvider.Instance;
+
+            if (!type.IsAggregate)
+            {
+                return returnType;
+            }
+
+            var fError = false;
```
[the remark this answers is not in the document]
lesaltzm: fError is a pretty standard usage here. It's a ref param, and used to indicate if there was an error in the operation. Given that all it's used for is an assert here, I'll leave it as is

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/RenameDriver.cs
```diff
@@ -0,0 +1,70 @@
...
+            }
+
+            var updatedType = RenameFormulaTypeHelper(aggregateType, segments, updatedName);
+            var fError = false;
+
+            // Use DType internals to swap one field type for the updated one and disable all other display names
+            var dropped = DType.DisableDisplayNameProviders(nestedType._type.Drop(ref fError, DPath.Root, field));
+            Contracts.Assert(!fError);
+
+            return FormulaType.Build(dropped.Add(field, updatedType._type));
```
bryancroteau-MSFT, line 67: At this point the queue should be empty right?

lesaltzm: yes, after the recursive call, the queue is empty.

lesaltzm: or we hit the case where the path doesn't fully exist in the type, in which case the queue may still have values in it but we  stopped recursing early.

bryancroteau-MSFT (dismissed):

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -151,6 +152,23 @@ public IIntellisenseResult Suggest(string expression, RecordType parameterType,
...
+        {
+            Contracts.AssertValue(parameters);
```
MikeStall, line 165: > Contracts.AssertValue [](http://example.com/codeflow?start=11&length=22)

Are these debug-only checks? Since this is a public method, we want to check them in release.

src/tests/Microsoft.PowerFx.Core.Tests/DisplayNameTests.cs
```diff
@@ -107,5 +107,48 @@ public void ConvertToInvariantNamesNoNames()
 
             Assert.Equal("If(B, Num, 1234)", displayExpressions);
         }
+
+        [Theory]
```
MikeStall, line 111: >Theory [](http://example.com/codeflow?start=9&length=6)

A few more tests: 
Add a  case that shows we're not just doing string.replace
"If(B, Text(B), "B")",  B-->"A",  ==> "If(A, Text(A), "B")" 

"B", B-->'B b'  ==> "'B b'"  // add spaces in identifier

src/tests/Microsoft.PowerFx.Core.Tests/DisplayNameTests.cs
```diff
@@ -107,5 +107,48 @@ public void ConvertToInvariantNamesNoNames()
...
+        [InlineData("With({RecordNest: {nest2: {datetest: 123}}}, RecordNest.nest2.datetest)", "With({RecordNest: {nest2: {datetest: 123}}}, RecordNest.nest2.datetest)", "RecordNest.nest2.datetest", "Foo")]
+        [InlineData("With(RecordNest, SomeString + nest2.datetest)", "With(Foo, SomeString + nest2.datetest)", "RecordNest", "Foo")]
+        [InlineData("With(RecordNest, SomeString + nest2.datetest)", "With(RecordNest, Foo + nest2.datetest)", "RecordNest.SomeString", "Foo")]
+        [InlineData("With(RecordNest, SomeString + nest2.datetest)", "With(RecordNest, SomeString + Foo.datetest)", "RecordNest.nest2", "Foo")]
+        [InlineData("With(RecordNest, SomeString + nest2.datetest)", "With(RecordNest, SomeString + nest2.Foo)", "RecordNest.nest2.datetest", "Foo")]
+        [InlineData("With({value: RecordNest.SomeString}, value & B)", "With({value: RecordNest.'abcd efg'}, value & B)", "RecordNest.SomeString", "abcd efg")]
+        public void RenameParameter(string expressionBase, string expectedExpression, string path, string newName)
```
MikeStall, line 127: >path [](http://example.com/codeflow?start=93&length=4)

we take a path, which is based on instance (not type) , there may be some interesting tests here...

RecordNest.SomeString --> S2

With({ x : RecordNest, y : RecordNest}, 
   x.SomeString & y.SomeString)

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/RenameDriver.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.Collections.Generic;
+using Microsoft.PowerFx.Core;
+using Microsoft.PowerFx.Core.Glue;
+using Microsoft.PowerFx.Core.Public.Types;
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.Utils;
+
+namespace Microsoft.PowerFx.Core
+{
+    public class RenameDriver
```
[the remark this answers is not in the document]
lesaltzm: good catch, thanks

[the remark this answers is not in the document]
bryancroteau-MSFT: Should probably be internal as well right?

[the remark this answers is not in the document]
bryancroteau-MSFT: Wait no this is in the public namespace.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -151,6 +152,23 @@ public IIntellisenseResult Suggest(string expression, RecordType parameterType,
...
+        /// <param name="pathToRename">Path to the field to rename.</param>
+        /// <param name="updatedName">New name. Replaces the last segment of <paramref name="pathToRename"/>.</param>
+        /// <returns></returns>
+        public RenameDriver CreateFieldRenamer(RecordType parameters, DPath pathToRename, DName updatedName)
+        {
+            Contracts.CheckValue(parameters, nameof(parameters));
+            Contracts.CheckValid(pathToRename, nameof(pathToRename));
+            Contracts.CheckValid(updatedName, nameof(updatedName));
+
+            return new RenameDriver(parameters, pathToRename, updatedName, CreateResolver(Config.WithoutDisplayNames()));
```
bryancroteau-MSFT, line 169: Why is .WithoutDisplayNames() being used here?

lesaltzm: added comment explaining the logic. We basically just need an environment where the only logical => display pair is the one that we're actively renaming.

bryancroteau-MSFT (approved):

PR #277, opened 2022-04-06
add examples of more Marshallers.

MikeStall:
test-only, no product change.


jorisdg (commented):
Looks good and covers the scenario we discussed. Just wondering if maybe it needs a bit more comment in the code to explain the point of marshalling the T type, ignoring the wrapper type itself?

src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/ScenarioCustomMarshaler.cs
```diff
@@ -69,5 +69,164 @@ public void Test()
...
+            ((TestObj)d["test"]).Field1 = 12;
+            var result2 = engine.Eval("x.test.Field1");
+            Assert.Equal(12.0, ((NumberValue)result2).Value);
+
+            // It's strongly typed, so we can give design-time errors for missing fields. 
+            var result3 = engine.Check("x.missing");
+            Assert.False(result3.IsSuccess);
+
+            // runtime casting error.             
+            Assert.Throws<AggregateException>(() => engine.Eval("x.bool"));
```
lesaltzm, line 229: Curious, why AggregateException here? Should we be providing a more meaningful exception for runtime marshalling failures, or is that on the implementer of the marshaller?

lesaltzm (dismissed):
🚢  but see comment about exception

src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/ScenarioCustomMarshaler.cs
MikeStall: It's a .net thing unrelated to this change - Task generally causes AggregateException to be thrown; `await` will unwrap that for you to the inner exception.  But that means things that call `.Result` may still get an aggregate whereas `await` gets the inner.

raghavthind2005 (approved):

src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/ScenarioCustomMarshaler.cs
```diff
@@ -62,5 +62,164 @@ public void Test()
...
+
+            var obj = new Wrapper<TestObj> { _value = inner };
+
+            var x = cache.Marshal(obj);
+
+            var engine = new RecalcEngine();
+            engine.UpdateVariable("x", x);
+
+            var result1 = engine.Eval("x.Field1");
+            Assert.Equal(10.0, ((NumberValue)result1).Value);
```
jorisdg, line 144: Assert that eval("x").toobject() should be obj

LucGenetier (approved):

PR #278, opened 2022-04-06
Make Token public

ivanradicek:
Convert `Token` class from internal to public, in order to remove some of `InternalsVisibleTo` projects. This also includes making public some of dependent types.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Lexer/WhitespaceToken.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Lexer/WhitespaceToken.cs
@@ -21,7 +21,7 @@ namespace Microsoft.PowerFx.Core.Lexer
             Value = value;
         }
 
-        public override Token Clone(Span ts)
+        internal override Token Clone(Span ts)
         {
             return new WhitespaceToken(Value, ts);
         }
--- a/src/libraries/Microsoft.PowerFx.Core/Localization/Span.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Localization/Span.cs
@@ -28,12 +28,7 @@ namespace Microsoft.PowerFx.Core.Localization
         /// </summary>
         public int Lim { get; }
 
-        /// <summary>
-        /// Initializes a new instance of the <see cref="Span"/> class.
-        /// </summary>
-        /// <param name="min"></param>
-        /// <param name="lim"></param>
-        public Span(int min, int lim)
+        internal Span(int min, int lim)
         {
             Contracts.CheckParam(min >= 0, "min");
             Contracts.CheckParam(lim >= min, "lim");
@@ -87,15 +82,8 @@ namespace Microsoft.PowerFx.Core.Localization
             return Min + match.Length <= script.Length && string.CompareOrdinal(script, Min, match, 0, match.Length) == 0;
         }
 
-        /// <summary>
-        /// Generic span replacer. Given a set of unordered spans and replacement strings for 
-        /// each, this produces a new string with all the specified spans replaced accordingly.
-        /// </summary>
-        /// <param name="script"></param>
-        /// <param name="worklist"></param>
-        /// <returns></returns>
         [TransportDisabled]
-        public static string ReplaceSpans(string script, IEnumerable<KeyValuePair<Span, string>> worklist)
+        internal static string ReplaceSpans(string script, IEnumerable<KeyValuePair<Span, string>> worklist)
         {
             Contracts.AssertValue(script);
             Contracts.AssertValue(worklist);
--- a/src/libraries/Microsoft.PowerFx.Core/Localization/Span.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Localization/Span.cs
@@ -66,14 +66,8 @@ namespace Microsoft.PowerFx.Core.Localization
             return string.Format(CultureInfo.InvariantCulture, "({0},{1})", Min, Lim);
         }
 
-        /// <summary>
-        /// Checks whether this span starts with some text.
-        /// </summary>
-        /// <param name="script"></param>
-        /// <param name="match"></param>
-        /// <returns></returns>
         [TransportDisabled]
-        public bool StartsWith(string script, string match)
+        internal bool StartsWith(string script, string match)
         {
             Contracts.AssertValue(script);
             Contracts.Assert(Min <= script.Length);
--- a/src/libraries/Microsoft.PowerFx.Core/Localization/Span.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Localization/Span.cs
@@ -11,14 +11,28 @@ using StringBuilderCache = Microsoft.PowerFx.Core.Utils.StringBuilderCache<Micro
 
 namespace Microsoft.PowerFx.Core.Localization
 {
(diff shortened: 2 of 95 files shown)


src/libraries/Microsoft.PowerFx.Core/Localization/Span.cs
```diff
@@ -60,8 +87,13 @@ public bool StartsWith(string script, string match)
...
-        // each, this produces a new string with all the specified spans replaced accordingly.
+        /// <summary>
+        /// Generic span replacer. Given a set of unordered spans and replacement strings for 
+        /// each, this produces a new string with all the specified spans replaced accordingly.
+        /// </summary>
+        /// <param name="script"></param>
+        /// <param name="worklist"></param>
+        /// <returns></returns>
         [TransportDisabled]
         public static string ReplaceSpans(string script, IEnumerable<KeyValuePair<Span, string>> worklist)
```
lesaltzm, line 98: Not sure ReplaceSpans or StartsWith need to be public. They're internal utilities, but I don't think they need to be exposed.

src/libraries/Microsoft.PowerFx.Core/Lexer/WhitespaceToken.cs
```diff
@@ -21,9 +21,9 @@ public WhitespaceToken(string value, Span span)
             Value = value;
         }
```
lesaltzm, line 22: Not your change, but why isn't this in Lexer/Tokens?

src/libraries/Microsoft.PowerFx.Core/Localization/Span.cs
```diff
@@ -11,13 +11,28 @@
...
+        /// End index of this span.
+        /// </summary>
         public int Lim { get; }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Span"/> class.
+        /// </summary>
+        /// <param name="min"></param>
+        /// <param name="lim"></param>
         public Span(int min, int lim)
```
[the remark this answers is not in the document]
ivanradicek: Note that this was all already public, I mostly added comments. But for sure, I can change relevant stuff to internal.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -70,8 +83,9 @@ public CheckResult Check(string expressionText, RecordType parameterType = null)
                 parameterType = new RecordType();
             }
 
-            var formula = new Formula(expressionText);
+            var formula = new Formula(expressionText, loc: Config.LanguageSettings);
```
lesaltzm, line 86: We don't want to enable locale-specific parsing yet, without writing a ton of tests. I think leaving out this for now is the right approach.

src/libraries/Microsoft.PowerFx.Core/Localization/Span.cs
ivanradicek: Same as above - this was already public. Will make internal.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -70,8 +83,9 @@ public CheckResult Check(string expressionText, RecordType parameterType = null)
...
 
+            // TODO: Can we enable expression chaining via some config option?
```
lesaltzm, line 88: Chaining is for side-effectful expressions, and we haven't yet added support for that in the interpreter. For now, Flags.None is the only supported mode

src/libraries/Microsoft.PowerFx.Core/Localization/Span.cs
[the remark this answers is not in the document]
lesaltzm: Ah, good point. Sorry I missed that.

src/libraries/Microsoft.PowerFx.Core/Lexer/WhitespaceToken.cs
ivanradicek: Good point. Note that the namespace is also different (matches the location). Should I move it?

src/libraries/Microsoft.PowerFx.Core/Localization/Span.cs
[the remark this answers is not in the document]
lesaltzm: In general all these constructors should be internal. We want these to come from Parse/Tokenize apis, not directly built by hosts.

src/libraries/Microsoft.PowerFx.Core/Lexer/WhitespaceToken.cs
lesaltzm: Yeah, I think so.

src/libraries/Microsoft.PowerFx.Core/Localization/Span.cs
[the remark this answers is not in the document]
ivanradicek: Makes sense.

src/libraries/Microsoft.PowerFx.Core/Localization/Span.cs
```diff
@@ -11,14 +11,24 @@
 
 namespace Microsoft.PowerFx.Core.Localization
 {
+    /// <summary>
+    /// Span in the text formula.
+    /// </summary>
     [TransportType(TransportKind.ByValue)]
```
MikeStall, line 17: >TransportType [](http://example.com/codeflow?start=5&length=13)

TransportType classes are brittle ... there's a custom codegen step that runs against them, reflects over it, and may do weird things based on public vs. internal.  
Definitely double check the changes with PAClient.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
ivanradicek: Removed.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
ivanradicek: OK. Removed.

src/libraries/Microsoft.PowerFx.Core/Localization/Span.cs
ivanradicek: I only changed public constructors to internal as McCall suggested. Other changes include adding only comments. I did try to compile with PAClient and there seem to be no problem. Should I do any further tests?

MikeStall: TransportType uses reflection, so it may actually care about the public vs. internal ctor. Not sure.. It's brittle.

lesaltzm: I changed the generator there to look at internals as well when we were doing this migration initially, so it should be safe. In general for transport, if it built, you should be fine. But it's something we'll be aware of when this is checked in.

lesaltzm (approved):
🚢

src/libraries/Microsoft.PowerFx.Core/Localization/Span.cs
ivanradicek: Great, thanks!

PR #280, opened 2022-04-06
Show OptionSets in interpreter intellisense

lesaltzm:
This pipes the PowerFxConfig object into intellisense, so that it can be used to show suggestions for globals in the interpreter.


src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Intellisense.cs
```diff
@@ -22,13 +22,13 @@ namespace Microsoft.PowerFx.Core.Texl.Intellisense
     internal partial class Intellisense : IIntellisense
     {
         protected readonly IReadOnlyList<ISuggestionHandler> _suggestionHandlers;
-        protected readonly EnumStore _enumStore;
+        protected readonly PowerFxConfig _config;
 
-        public Intellisense(EnumStore enumStore, IReadOnlyList<ISuggestionHandler> suggestionHandlers)
+        public Intellisense(PowerFxConfig config, IReadOnlyList<ISuggestionHandler> suggestionHandlers)
```
MikeStall, line 27: >Intellisense(PowerFxConfig config, [](http://example.com/codeflow?start=15&length=34)

Will this be an inssue for PAClient? 
CanvasIntellisense., DataverseIntellisense both impl it.

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseData/IntellisenseData.cs
```diff
@@ -308,6 +311,17 @@ internal virtual void AddCustomSuggestionsForValuePossibilities()
...
         internal virtual void AddCustomSuggestionsForGlobals()
         {
+            foreach (var global in _powerFxConfig.GetSymbols())
+            {
+                var type = DType.Unknown;
+
+                if (global is IExternalOptionSet optionSet)
+                {
+                    type = optionSet.Type;
+                }
```
MikeStall, line 321: Can symbols know their type? 
Seems like every symbol would need that, and having to cast to specific symbol types is brittle.

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseData/IntellisenseData.cs
lesaltzm: #287

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Intellisense.cs
lesaltzm: Yes, updated

VamsiModem (approved):

PR #281, opened 2022-04-06
Remove unused methods and fields from EnumStore/Builder

bryancroteau-MSFT:
Continuing the story of cleaning up EnumStore, this change introduces the IEnumStore interface and changes Intellisense classes to use the new interface. The old EnumStore will be added to PA Client with a new class named LegacyEnumStore, which will implement IEnumStore. This old code was fragile and cannot be changed.


lesaltzm (approved):
:shipit:

PR #282, opened 2022-04-07
Test updates -

MikeStall:
Allow harnesses to handle known unsupported features without require 100s of overrides.

This brings the sql test failures from 300 down to 100 by eliminating false positives due to known limitations.


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/BaseRunner.cs
```diff
@@ -14,6 +14,65 @@
...
 {
+    /// <summary>
+    /// Result of a call to <see cref="BaseRunner.RunAsyncInternal(string, string)"/>.
+    /// - Compilation error - set Errors
+    /// - Runtime error - set Value
+    ///
+    /// If the eval used unsupported behavior, set UnsupportedReason to a message.
+    /// This may cause th etest to get skipped rather than fail. 
+    /// </summary>
+    public class RunResult
```
MikeStall, line 25: >RunResult [](http://example.com/codeflow?start=17&length=9)

Before, we would throw in some cases, return values in others.  This makes it more consistent - which is important since we want to allow a UnsupportedReason in any scenario.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/BaseRunner.cs
```diff
@@ -250,6 +322,29 @@ public virtual bool IsError(FormulaValue value)
         {
             return value is ErrorValue;
         }
+
+        // Derived harness may need to override if they have a different precision level. 
+        public virtual bool NumberCompare(double a, double b)
```
MikeStall, line 327: >NumberCompare [](http://example.com/codeflow?start=28&length=13)

Different backends have different numerical precision and needed a hook to compare.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/BaseRunner.cs
```diff
@@ -118,47 +177,56 @@ private async Task<(TestResult, string)> RunAsync2(TestCase testCase)
...
-                    result = await RunAsyncInternal(testCase.Input, testCase.SetupHandlerName);
-                }
-                catch (SetupHandlerNotFoundException)
-                {
-                    return (TestResult.Skip, $"was skipped due to missing setup handler {testCase.SetupHandlerName}");
-                }
-
-                actualStr = TestRunner.TestToString(result);
-            }
-            catch (Exception e)
```
MikeStall, line 130: >catch (Exception e) [](http://example.com/codeflow?start=12&length=19)

Key difference here is that before we would expect an exception for compile errors (thrown via CheckResult.ThrowOnErrors). 

Now we expect an explicit return result.   And exceptions are truly exceptional.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/BaseRunner.cs
```diff
@@ -14,6 +14,65 @@
...
+        public FormulaValue Value;
+
+        // Test may have run and failed (so Value is set) due to
+        // known unsupported behavior in an engine. Let engine mark that this is a known case.
+        //
+        // This will often mean skipping the test, but may still be a failure if:
+        // - the test is already from an override (so it should have marked #skip)
+        // - the test already expected to fail (so we should have gotten the failure)
+        //
+        // This allows tests to over lots of overrides for known behavior that isn't implemented.
```
lesaltzm, line 41: >over lots of overrides [](http://example.com/codeflow?start=32&length=22)

grammar?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/BaseRunner.cs
```diff
@@ -118,47 +177,56 @@ private async Task<(TestResult, string)> RunAsync2(TestCase testCase)
             {
                 try
                 {
```
lesaltzm, line 179: nit: nested try could just be 

try {
}
catch (SetupHandlerException e)
{}
catch (Exception e)
{}

lesaltzm (approved):
:shipit:

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/BaseRunner.cs
MikeStall: Good catch - I had removed some of the code that was preventing it from being flattened.

lesaltzm (approved):
🚢

PR #285, opened 2022-04-07
Convert non-space white space characters to spaces in DNames

jelopezf:
DNames shouldn't contain tabs, new line, carriage feeds, etc.


src/libraries/Microsoft.PowerFx.Core/Utils/DName.cs
```diff
@@ -1,179 +1,205 @@
...
+            Contracts.AssertValueOrNull(str);
+            return name.Value != str;
+        }
+
+        /// <summary>
+        /// Returns whether the given name is a valid <see cref="DName" />. 
+        /// </summary>
+        /// <param name="strName"></param>
+        /// <returns></returns>
+        public static bool IsValidDName(string strName)
```
MikeStall, line 125: >IsValidDName [](http://example.com/codeflow?start=27&length=12)

Is there a potential regression here where a field was invalid but unused - so it used to be ignored but now is causing a failure?

src/libraries/Microsoft.PowerFx.Core/Utils/DName.cs
```diff
@@ -1,179 +1,205 @@
...
+            var fHasSpecialWhiteSpaceCharacters = false;
+            
+            fModified = false;
+
+            for (var i = 0; i < strName.Length; i++)
+            {
+                bool fIsSpace = strName[i] == ChSpace;
+                bool fIsWhiteSpace = char.IsWhiteSpace(strName[i]);
+                fAllSpaces = fAllSpaces && fIsWhiteSpace;
+                fHasSpecialWhiteSpaceCharacters = fHasSpecialWhiteSpaceCharacters || (fIsWhiteSpace && !fIsSpace);
```
MikeStall, line 172: >fHasSpecialWhiteSpaceCharacters = fHasSpecialWhiteSpaceCharacters || [](http://example.com/codeflow?start=16&length=68)

nit:
fHasSpecialWhiteSpaceCharacters |= (fIsWhiteSpace && !fIsSpace);

src/libraries/Microsoft.PowerFx.Core/Utils/DName.cs
```diff
@@ -1,179 +1,205 @@
...
+            }
+
+            if (fHasSpecialWhiteSpaceCharacters)
```
MikeStall, line 175: >if (fHasSpecialWhiteSpaceCharacters) [](http://example.com/codeflow?start=12&length=36)

is this possible to encode an invalid dname differently than before?

src/tests/Microsoft.PowerFx.Core.Tests/DNameTests.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.PowerFx.Core.Utils;
+using Xunit;
+
+namespace Microsoft.PowerFx.Core.Tests
+{
+    public sealed class DNameTests
+    {
+        [Fact]
```
lesaltzm, line 14: Can this use [Theory] and [InlineData]? Separates out each test case individually, makes it easier to debug in VS

src/libraries/Microsoft.PowerFx.Core/Utils/DName.cs
jelopezf: no, this only replaces some characters with spaces which are valid.

src/libraries/Microsoft.PowerFx.Core/Utils/DName.cs
jelopezf: reverted since the previous check also checked for a lot of the whitespace characters already and it's safer to keep previous validation.

MikeStall (approved):
:shipit:

lesaltzm (approved):

PR #287, opened 2022-04-08
All symbols know their type without casting

lesaltzm:
This is from a comment in #280, deserves to be it's own PR for clarity.


bryancroteau-MSFT (approved):

MikeStall:
Per #280, Can we apply this to AddCustomSuggestionsForGlobals?

lesaltzm:
Yup, that’s the idea.

PR #288, opened 2022-04-08
Draft: Fix DType.MaxDepth no elements exception

rick-nguyen:
GetName.Max(...) code throws exception when GetNames() returns no elements.
This could happen if the Child Types contains only 1 element whose key is MetaFieldName ("meta-6de62757-ecb6-4be6-bb85-349b3c7938a9")
GetNames() exclude MetaFieldName from the returned IEnumerable, so this function returns Empty although ChildCount is 1


rick-nguyen:
Collect(d, {'meta-6de62757-ecb6-4be6-bb85-349b3c7938a9': 100}) will still throw an error because it's not a valid expression, but will no longer cause Studio to crash

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -601,7 +601,7 @@ public bool IsNamedValueOfKind(string kind)
                     ? 1 +
                       (ChildCount == 0
                           ? 0
-                          : GetNames(DPath.Root)
+                          : GetAllNames(DPath.Root)
```
MikeStall, line 604: >GetAllNames [](http://example.com/codeflow?start=28&length=11)

This is definitely subtle - is there a test case we can add for it in core (preferably a .txt one if possible)?

MikeStall: fwiw, I agree with this fix - we just need a test for it. 

Max() is throwing an exception because GetNames() is returning empty despite ChildCount being > 0. That's happening because GetNames() is GetallNames() with a filter. So remove the filter and just use GetAllNames().

MikeStall (approved):
:shipit:

Issue #283, opened 2022-04-07
Add more .txt tests


MikeStall:
Add more tests to the .txt

https://github.com/microsoft/Power-Fx/tree/main/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases

These are run in interpreter here:
https://github.com/microsoft/Power-Fx/blob/main/src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs

Verify expected behavior in Power Apps.

PowerAppsDarren:
May I contribute on this issue?

---

The two links in the original post don't seem to exist. ⬇️

---

<img width="1276" height="543" alt="Image" src="https://github.com/user-attachments/assets/89a69037-323e-443e-b65a-08ce3658c507" />

---

<img width="1133" height="492" alt="Image" src="https://github.com/user-attachments/assets/b09c7253-08aa-4d48-aae2-80fb4864bd13" />

MikeStall:
@PowerAppsDarren  - you are very welcome to contribute! 

The files have moved: 
https://github.com/microsoft/Power-Fx/tree/main/src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases

https://github.com/microsoft/Power-Fx/blob/main/src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/PowerFxEvaluationTests.cs

PR #290, opened 2022-04-11
added some tests to isToday()

felipeiagocruz:
Hi! I'm interested in contributing to testing if no one has taken up this issue.
However, it is not clear to me what would be the correct way to submit these tests. For example in line 9 of the PR, PowerApps returns as false:
`>> IsToday(Date(-1, 2, 3))`
But the previous test said error ("#Error").
Can you check if I'm doing it right?

issue #283

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/IsToday.txt
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/IsToday.txt
@@ -6,7 +6,19 @@ true
 true
 
 >> IsToday(Date(-1, 2, 3))
-#Error
+false
+
+>> IsToday(DateDiff(Now(),Today()))
+false
+
+>> IsToday("30,12,4")
+false
+
+>> IsToday(Date(-1995,1,30))
+false
+
+>> IsToday(Blank())
+false
 
 >> IsToday(Date(Blank(), 2, 3))
 false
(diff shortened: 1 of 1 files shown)


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/IsToday.txt
```diff
@@ -6,7 +6,19 @@ true
...
 
 >> IsToday(Date(-1, 2, 3))
-#Error
+false
+
+>> IsToday(DateDiff(Now(),Today()))
+false
+
+>> IsToday("30,12,4")
+false
```
MikeStall, line 15: This will be an error, passing the string to IsToday(). 

#Error(Kind=25)

MikeStall:
An easy way to test and check for expected results would be to try the expressions inside of a Canvas app and see what the results are. There are corner cases around passing blank, passing "bad" inputs (like a string when a date is expected), etc.

lesaltzm:
Make sure your canvas app has the error management experimental feature enabled, that should give you the correct results here.
image
image

felipeiagocruz:
> Make sure your canvas app has the error management experimental feature enabled, that should give you the correct results here. image image

Thanks @lesaltzm! Now I'm getting the results back correctly, I'll start working on the issue.

PR #294, opened 2022-04-12
Expose a couple more internals

ivanradicek:
This PR exposes a bit more internals, but then removes a project from `InternalsVisibleTo`.


src/libraries/Microsoft.PowerFx.Core/Lexer/Tokens/IdentToken.cs
```diff
@@ -70,7 +71,18 @@ internal override Token Clone(Span ts)
         // REVIEW ragru: having a property for every possible error isn't scalable.
         internal bool HasDelimiters => HasDelimiterStart;
 
-        internal bool HasErrors => IsModified || (HasDelimiterStart && !HasDelimiterEnd);
+        /// <summary>
+        /// Whether an identifier has errors.
+        /// </summary>
+        public bool HasErrors => IsModified || (HasDelimiterStart && !HasDelimiterEnd);
+
+        // TODO: How to do this properly?
```
ivanradicek, line 79: This is probably not good. Looking forward for refactor suggestions.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -67,6 +67,31 @@ private protected virtual SimpleResolver CreateResolver(PowerFxConfig alternateC
         public IReadOnlyList<Token> Tokenize(string expressionText)
             => TexlLexer.LocalizedInstance.GetTokens(expressionText);
 
+        /// <summary>
+        /// Check only syntax of the expression.
+        /// </summary>
+        /// <param name="expressionText"></param>
+        /// <returns></returns>
+        // TODO: This should be done differently, temporary hack!
```
ivanradicek, line 75: As discussed earlier we should have a separate parse-only result. This is a temporary hack that should be done properly.

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
```diff
@@ -57,6 +61,25 @@ internal FormulaType(DType type)
             _type = type;
         }
 
+        /// <summary>
```
ivanradicek, line 64: Perhaps we could do without this (it would be nice :)). I'll need to check whether this can be avoided at our side.

src/libraries/Microsoft.PowerFx.Core/Syntax/Visitors/IdentityTexlVisitor.cs
```diff
@@ -5,77 +5,97 @@
 
 namespace Microsoft.PowerFx.Core.Syntax.Visitors
 {
-    // Identity visitor
-    internal abstract class IdentityTexlVisitor : TexlVisitor
+    /// <summary>
+    /// Visitor that by default does nothing. <seealso cref="TexlVisitor" />.
+    /// </summary>
+    public abstract class IdentityTexlVisitor : TexlVisitor
```
ivanradicek, line 11: Did not make this public earlier, but it seems it is useful, and it doesn't expose any new things really.

src/libraries/Microsoft.PowerFx.Core/Public/FunctionInfo.cs
```diff
@@ -0,0 +1,57 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.PowerFx.Core.Functions;
+using Microsoft.PowerFx.Core.Public.Types;
+
+namespace Microsoft.PowerFx.Core.Public
```
ivanradicek, line 10: Not sure about namespace of this.

src/libraries/Microsoft.PowerFx.Core/Lexer/Tokens/IdentToken.cs
```diff
@@ -70,7 +71,18 @@ internal override Token Clone(Span ts)
...
+        public static string MakeValidIdentifier(string value)
```
MikeStall, line 80: >MakeValidIdentifier [](http://example.com/codeflow?start=29&length=19)

add a unit test.  (Theory + InlineData)

src/libraries/Microsoft.PowerFx.Core/Lexer/Tokens/IdentToken.cs
MikeStall: This seems reasonable place.

src/libraries/Microsoft.PowerFx.Core/Lexer/Tokens/IdentToken.cs
LucGenetier: Consider null, empty string and a set of whitespaces

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -62,6 +68,46 @@ public void ThrowOnErrors()
...
+
+        /// <summary>
+        /// Check whether function invocation in type-correct.
+        /// </summary>
+        /// <param name="fnc"></param>
+        /// <param name="nodes"></param>
+        /// <param name="types"></param>
+        /// <param name="retType"></param>
+        /// <returns></returns>
+        public bool CheckInvocation(FunctionInfo fnc, IReadOnlyList<TexlNode> nodes, IReadOnlyList<FormulaType> types, out FormulaType retType)
```
MikeStall, line 93: > IReadOnlyList<FormulaType> types [](http://example.com/codeflow?start=84&length=33)

Could we remove these and get from the nodes?

Rename nodes to args?

src/libraries/Microsoft.PowerFx.Core/Public/FunctionInfo.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// Whether this function has lambdas as parameters.
+        /// </summary>
+        public bool HasLambdas => _fnc.HasLambdas;
+
+        /// <summary>
+        /// Whether the <paramref name="i" />th parameter is a lambda parameter.
+        /// </summary>
+        /// <param name="i"></param>
+        /// <returns></returns>
+        public bool IsLambdaParam(int i) => _fnc.IsLambdaParam(i);
```
MikeStall, line 50: >IsLambdaParam [](http://example.com/codeflow?start=20&length=13)

need to double check - there are cases where this is not sufficient.

src/libraries/Microsoft.PowerFx.Core/Public/Types/ErrorType.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    /// <summary>
+    /// An error type, attached to syntax nodes whose type is incorrect.
+    /// </summary>
+    public class ErrorType : FormulaType
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ErrorType"/> class.
+        /// </summary>
+        public ErrorType()
```
MikeStall, line 17: >public [](http://example.com/codeflow?start=8&length=6)

Ivan is ok with internal ctor

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
```diff
@@ -44,6 +44,10 @@ public abstract class FormulaType
         public static FormulaType Color { get; } = new ColorType();
 
         public static FormulaType Guid { get; } = new GuidType();
+
+        public static FormulaType Unknown { get; } = new UnknownType();
+
+        public static FormulaType Error { get; } = new ErrorType();
```
MikeStall, line 50: People may abuse this - can we make safer.

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
MikeStall: yes, let's remove. 
Exposing the string format for types as part of public surface area is large surface - and we'll want to to replace with a proper maker-facing syntax.

src/libraries/Microsoft.PowerFx.Core/Public/Types/Serialization/FormulaTypeSchema.cs
```diff
@@ -28,6 +28,8 @@ public enum ParamType
             UntypedObject,
             EntityRecord,
             EntityTable,
+            Unknown,
+            Error
```
MikeStall, line 32: To check - should we just fail serialization out-right if there are errors here?

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
lesaltzm, line 93: Why do we want to expose CheckInvocation?

lesaltzm: Ideally, we'd really like to move away from using CheckInvocation as our method of validating params. Exposing it (and it's oddities) publicly seems like the wrong direction.

MikeStall: @ivanradicek - can we just pass the candidates as strings into a Check() call?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
MikeStall: See https://github.com/microsoft/Power-Fx/pull/319 for making ParseResult public.

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
ivanradicek: Remove these completely perhaps? Then there is no public way anymore to create those two types (ctors are internal now).

src/libraries/Microsoft.PowerFx.Core/Lexer/Tokens/IdentToken.cs
ivanradicek: Added some tests, let me know if we need to be more thorough.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
ivanradicek: Awesome, this works perfectly.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
ivanradicek: Removed types and renamed to args.

src/libraries/Microsoft.PowerFx.Core/Public/Types/ErrorType.cs
```diff
@@ -0,0 +1,25 @@
...
+
+using System;
+using Microsoft.PowerFx.Core.Types;
+
+namespace Microsoft.PowerFx.Core.Public.Types
+{
+    /// <summary>
+    /// An error type, attached to syntax nodes whose type is incorrect.
+    /// </summary>
+    public class ErrorType : FormulaType
```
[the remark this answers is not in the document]
ivanradicek: Renamed to BindingErrorType.

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
```diff
@@ -139,6 +168,13 @@ public override int GetHashCode()
 
         #endregion // Equality
 
+        /// <summary>
+        /// Whether <paramref name="other"/> can be accepted where this type is accepted.
+        /// </summary>
+        /// <param name="other"></param>
+        /// <returns></returns>
+        public bool Accepts(FormulaType other) => _type.Accepts(other._type);
```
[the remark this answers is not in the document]
ivanradicek: Removed.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
ivanradicek: This is quite important for our use cases. String is problematic because we don't have mapping from names to functions at this point. What can we do about this?

src/libraries/Microsoft.PowerFx.Core/Public/Types/BindingErrorType.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using System;
+using Microsoft.PowerFx.Core.Types;
+
+namespace Microsoft.PowerFx.Core.Public.Types
+{
+    /// <summary>
+    /// A binding error type, attached to syntax nodes whose type is incorrect.
+    /// </summary>
+    public class BindingErrorType : FormulaType
```
lesaltzm, line 12: sealed

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
ivanradicek: Or at least do this for error type and leave for unknown?

src/libraries/Microsoft.PowerFx.Core/Public/FunctionInfo.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        /// <summary>
+        /// Function parameter types.
+        /// </summary>
+        public IReadOnlyList<FormulaType> ParamTypes => _fnc.ParamTypes.Select(FormulaType.Build).ToList();
```
MikeStall, line 55: This may need more discussion. 
The challenge is that this information is often not correct (see If() statement). 

How are you using this information?

src/libraries/Microsoft.PowerFx.Core/Lexer/Tokens/IdentToken.cs
```diff
@@ -70,7 +72,28 @@ internal override Token Clone(Span ts)
...
+        /// </summary>
+        public bool HasErrors => IsModified || (HasDelimiterStart && !HasDelimiterEnd);
+
+        /// <summary>
+        /// Converts a string value of an identifier to a valid identifier (e.g., "a b" -> "'a b'").
+        /// </summary>
+        /// <param name="value"></param>
+        /// <returns></returns>
+        /// <exception cref="ArgumentNullException">Throw if value is <c>null</c>.</exception>
+        public static string MakeValidIdentifier(string value)
```
MikeStall, line 86: >MakeValidIdentifier [](http://example.com/codeflow?start=29&length=19)

It seems like this should have some failure path. 
 Ie, if we pass bad chars ("***") 

Related to: 
public static DName MakeValid(string strName, out bool fModified)

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
ivanradicek: Removed for now altogether.

src/libraries/Microsoft.PowerFx.Core/Public/FunctionInfo.cs
ivanradicek: Removed.

src/libraries/Microsoft.PowerFx.Core/Public/FunctionInfo.cs
ivanradicek: Removed.

src/libraries/Microsoft.PowerFx.Core/Lexer/Tokens/IdentToken.cs
ivanradicek: It seems that the lexer is accepting anything really (e.g., in the web demo `'***'` can be used as a legal identifier). The only failure path is `null`. Or should we test this in some other way?

MikeStall (approved):
:shipit:

Issue #284, opened 2022-04-07
Implement missing functions in Interpreter 


MikeStall:
The interpreter doesn't implement all Power Apps functions yet. 

See list of power apps functions:  https://docs.microsoft.com/en-us/power-apps/maker/canvas-apps/formula-reference

List of functions implemented in Interpreter is at: https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs 

Looking at history for that file is an easy way to spot when new functions were added. 

Here's are some good previous examples:

Function | PR
--- | ---
Lookup() |  https://github.com/microsoft/Power-Fx/pull/212
Index() | https://github.com/microsoft/Power-Fx/commit/2b16f6faf9ceaf22e7ca268b3c82d60ea2077ef6
Find() | https://github.com/microsoft/Power-Fx/commit/b175c1fea3c276fbcb8f32270def09e531e42f4b
Boolean() | https://github.com/microsoft/Power-Fx/commit/27bd908ec78408704cb2634dd34ebf1919da1e17

MikeStall:
Some ideas for good candidates here:

Proper
Color
Count
CountA
ISOWeekNum
Shuffle
SortByColumns

Statistical ones:  StdevP, StdevPT, VarP, VarPT

anderson-joyle:
There is a **CountRow** function already implemented. Is there any difference between **Count** and **CountRow**?

jorisdg:
> There is a **CountRow** function already implemented. Is there any difference between **Count** and **CountRow**?

This has tripped me up before as well, but it's an Excel thing :-) https://docs.microsoft.com/en-us/power-apps/maker/canvas-apps/functions/function-table-counts
The Count function counts the number of **records that contain a number in a single-column table**.
The CountRows function counts **the number of records** in a table.

DavidLuong98:
Currently working on the shuffle function

anderson-joyle:
I was working on the **Proper** function when I came across to the following:
According to the [documentation](https://docs.microsoft.com/en-us/power-apps/maker/canvas-apps/functions/function-lower-upper-proper#syntax) these are possible syntaxes:

```
Lower( SingleColumnTable )
Upper( SingleColumnTable )
Proper( SingleColumnTable )
```

[...]

PR #302, opened 2022-04-13
Implement SortByColumns

jcoc611-microsoft:
Implements SortByColumns with varargs and support for Number, String, Boolean, Date and DateTime.

Also adds unit tests. #284

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
@@ -265,6 +265,109 @@ namespace Microsoft.PowerFx.Functions
                 return SortValueType<BooleanValue, bool>(pairs, irContext, compareToResultModifier);
             }
         }
+        
+        public static async ValueTask<FormulaValue> SortByColumns(EvalVisitor runner, SymbolContext symbolContext, IRContext irContext, FormulaValue[] args)
+        {
+            var table = (TableValue)args[0];
+            var tableType = (TableType)table.Type;
+
+            var records = table.Rows.ToList();
+            var errors = new List<ErrorValue>();
+            for (var argIndex = 1; argIndex < args.Length; argIndex += 2)
+            {
+                var fieldName = ((StringValue)args[argIndex]).Value;
+                if (tableType.MaybeGetFieldType(fieldName) is not NumberType and not StringType
+                    and not BooleanType and not DateTimeType and not DateType)
+                {
+                    // column not found error
+                    errors.Add(CommonErrors.RuntimeTypeMismatch(irContext));
+                    break;
+                }
+            }
+
+            if (errors.Count != 0)
+            {
+                return ErrorValue.Combine(irContext, errors);
+            }
+
+            foreach (var record in records)
+            {
+                if (record.IsError)
+                {
+                    errors.Add((ErrorValue)record.ToFormulaValue());
+                    continue;
+                }
+
+                for (var argIndex = 1; argIndex < args.Length; argIndex += 2)
+                {
+                    var fieldName = ((StringValue)args[argIndex]).Value;
+                    var fieldValue = record.Value.GetField(fieldName);
+                    if (fieldValue is ErrorValue fieldErrorValue)
+                    {
+                        errors.Add(fieldErrorValue);
+                    }
+                }
+            }
+
+            if (errors.Count != 0)
+            {
+                return ErrorValue.Combine(irContext, errors);
+            }
+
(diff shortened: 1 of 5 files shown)


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Sort.txt
```diff
@@ -38,4 +38,40 @@
 [true,true,false,false,Blank()]
 
 >> First(LastN(Sort([-2, -1, 0, 1, 2], 1 / Value), 3)).Value
-#Error
\ No newline at end of file
+#Error
+
+>> SortByColumns([-2, -1, 5, 1, 2], "Value")
```
MikeStall, line 43: >SortByColumns [](http://example.com/codeflow?start=3&length=13)

nit - can you place in its own SortByColumns.txt file?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -295,6 +398,15 @@ private static bool IsValueTypeErrorOrBlank<T>(FormulaValue val)
             return new InMemoryTableValue(irContext, pairs.Select(pair => pair.Key));
         }
 
+        private static int CompareColumns<TPFxPrimitive, TDotNetPrimitive>(FormulaValue a, FormulaValue b)
```
MikeStall, line 401: >CompareColumns [](http://example.com/codeflow?start=27&length=14)

This is elegant - but semantically is it consistent with the existing binary comparer?

See OperatorBinaryLeq etc in LibraryOperators.

This could be an issue for blanks, errors. and possibly string casing.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -265,6 +265,109 @@ public static async ValueTask<FormulaValue> SortTable(EvalVisitor runner, Symbol
...
+            if (errors.Count != 0)
+            {
+                return ErrorValue.Combine(irContext, errors);
+            }
+
+            foreach (var record in records)
+            {
+                if (record.IsError)
+                {
+                    errors.Add((ErrorValue)record.ToFormulaValue());
```
lesaltzm, line 297: Is this the correct error behavior? @CarlosFigueiraMSFT

lesaltzm (changes requested):
@gregli-msft @jorisdg We need to chat about the design for these functions that take string literal column names. If we want to change them, we should do so before they're added to the C# interpreter to avoid making the back compat problem worse.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
CarlosFigueiraMSFT: Yes - this is correct. When sorting, if we find any errors then the result will be an error. This implementation is returning all errors as it should.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Sort.txt
```diff
@@ -38,4 +38,40 @@
...
+
+>> SortByColumns(["a", "b", Blank(), "x", "J", "C"], "Value", SortOrder.Descending)
+["x","J","C","b","a",Blank()]
+
+>> SortByColumns([Date(2020, 01, 05), Date(2020, 01, 01), Date(1995, 01, 01)], "Value")
+[Date(1995,1,1),Date(2020,1,1),Date(2020,1,5)]
+
+>> SortByColumns([Date(2020, 01, 05), Date(2020, 01, 01), Date(1995, 01, 01)], "Value", SortOrder.Descending)
+[Date(2020,1,5),Date(2020,1,1),Date(1995,1,1)]
+
```
CarlosFigueiraMSFT, line 72: Can you please add tests for DateTime values (in addition to those ones for dates)?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Sort.txt
jcoc611-microsoft: Moved!

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Sort.txt
jcoc611-microsoft: Added!

PR #305, opened 2022-04-13
Compile string resources into satellite assemblies to reduce assembly size

lesaltzm:
All our resources were compiled into PowerFx.Core prevously. This splits them out into satellite assemblies, and uses ResourceManager to load the appropriate resource assembly on demand. See https://docs.microsoft.com/en-us/dotnet/api/system.resources.resourcemanager for more details

Before:
image

After:
image


src/libraries/Microsoft.PowerFx.Core/Microsoft.PowerFx.Core.csproj
```diff
@@ -30,11 +31,11 @@
...
 
 	<ItemGroup>
-		<EmbeddedResource Include="..\..\strings\**\PowerFxResources.resw" Condition="'$(Configuration)' == 'Debug'">
-			<Link>Strings\%(RecursiveDir)%(Filename)%(Extension)</Link>
+		<EmbeddedResource Include="..\..\strings\PowerFxResources.*.resx" Condition="'$(Configuration)' == 'Debug'">
+			<Link>strings\%(Filename)%(Extension)</Link>
 		</EmbeddedResource>
-		<EmbeddedResource Include="..\..\strings\**\PowerFxResources.resw" Exclude="$(SolutionDir)\strings\qps-ploc\**" Condition="'$(Configuration)' == 'Release'">
-			<Link>Strings\%(RecursiveDir)%(Filename)%(Extension)</Link>
+		<EmbeddedResource Include="..\..\strings\PowerFxResources.*.resx" Exclude="..\..\strings\PowerFxResources.qps-ploc.resx" Condition="'$(Configuration)' == 'Release'">
```
lesaltzm, line 37: Resx files with this naming pattern get converted to satellite assemblies by default

src/libraries/Microsoft.PowerFx.Core/Microsoft.PowerFx.Core.csproj
```diff
@@ -21,6 +21,7 @@
 		<Description>The core library for all Microsoft Power Fx Compiler targets</Description>
 		<Summary>The core library for all Microsoft Power Fx Compiler targets</Summary>
 		<GeneratePackageOnBuild>True</GeneratePackageOnBuild>
+		<GenerateNeutralResourcesLanguageAttribute>false</GenerateNeutralResourcesLanguageAttribute>
```
lesaltzm, line 24: This needs to be disabled so that the en-us resources can also be loaded from a satellite.

src/tests/Microsoft.PowerFx.Core.Tests/ResourceValidationTests.cs
```diff
@@ -14,6 +18,48 @@ namespace Microsoft.PowerFx.Tests
...
+            var loadedCount = 0;
+
+            void ResourceAssemblyLoadHandler(object sender, AssemblyLoadEventArgs args)
+            {
+                loaded = args.LoadedAssembly.FullName;
+                loadedCount++;
+            }
+
+            var currentDomain = AppDomain.CurrentDomain;
+            currentDomain.AssemblyLoad += ResourceAssemblyLoadHandler;
```
MikeStall, line 37: nit: put in try finally and unregister this so that other tests don't call it.

src/tests/Microsoft.PowerFx.Core.Tests/ThreadingTests.cs
```diff
@@ -34,8 +34,8 @@ public void CheckFxCore()
                 "DateAddFunction.SubDayStringList",
                 "PrettyPrintVisitor.BinaryPrecedence",
                 "ErrorResource.ErrorResourceTagToReswSuffix",
-                "StringResources.Strings",
-                "StringResources.ErrorResources",
+                "StringResources._resourceManager",
```
MikeStall, line 37: >_resourceManager [](http://example.com/codeflow?start=33&length=16)

If you add a [ThreadSafeProtectedByLockAttribute] around these fields, they'll be considered safe and you don't have to add here.

src/libraries/Microsoft.PowerFx.Core/Localization/StringResources.cs
```diff
@@ -31,13 +27,16 @@ internal enum ResourceFormat
 
         private const string FallbackLocale = "en-US";
 
+        private static readonly ResourceManager _resourceManager = new ResourceManager("Microsoft.PowerFx.Core.strings.PowerFxResources", typeof(StringResources).Assembly);
+        private static readonly Dictionary<string, Dictionary<string, ErrorResource>> _errorResources = new Dictionary<string, Dictionary<string, ErrorResource>>(StringComparer.OrdinalIgnoreCase);
```
MikeStall, line 31: >_errorResources [](http://example.com/codeflow?start=86&length=15)

[ThreadSafeProtectedByLockAttribute("_errorResources")]

src/libraries/Microsoft.PowerFx.Core/Localization/StringResources.cs
```diff
@@ -31,13 +27,16 @@ internal enum ResourceFormat
 
         private const string FallbackLocale = "en-US";
 
+        private static readonly ResourceManager _resourceManager = new ResourceManager("Microsoft.PowerFx.Core.strings.PowerFxResources", typeof(StringResources).Assembly);
```
MikeStall, line 30: >_resourceManager [](http://example.com/codeflow?start=48&length=16)

This is static, not protected by a lock. 

If ResourceManager is semantically immutable (and thread safe),  then add it to _knownImmutableTypes in AnalyzeThreadSafety.cs. 

Else, we should add a lock.

MikeStall:
How much does this shrink by?

lesaltzm:
> How much does this shrink by?

From the description: 8984 MB -> 785 KB

CarlosFigueiraMSFT:
> <root>

We should check with the localization team - not sure if their pipeline supports resources with different suffixes (instead of different folders)

---
Refers to: src/strings/PowerFxResources.ca-ES.resx:2 in 111e746. [](commit_id = 111e7467ebeeb91db3ece3b26ac23437ebb58ccb, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Localization/StringResources.cs
lesaltzm: Docs claim it's thread safe
https://docs.microsoft.com/en-us/dotnet/api/system.resources.resourcemanager?view=net-6.0#thread-safety

MikeStall: Docs are lying :) 
[https://docs.microsoft.com/en-us/dotnet/api/system.resources.resourcemanager.releaseallresources?view=net-6.0](https://docs.microsoft.com/en-us/dotnet/api/system.resources.resourcemanager.releaseallresources?view=net-6.0)

GetString() is thread safe w.r.t other calls to GetString()
But not safe w.r.t. calls to ReleaseAll()

We don't call ReleaseAll(), so not an issue in practice. 

Not required, but If we really wanted to by "defensive", here, could create a wrapper around ResourceManager that only allows GetString() and declare that wrapper as ThreadSafe.

MikeStall:
That's like a 90% reduction in size! Excellent!

---
In reply to: [1098588196](https://github.com/microsoft/Power-Fx/pull/305#issuecomment-1098588196) [](http://example.com/codeflow?ancestors=1098588196,1098588196)

MikeStall (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Microsoft.PowerFx.Core.csproj
CarlosFigueiraMSFT, line 37: >Exclude="..\..\strings\PowerFxResources.qps-ploc.resx"  [](http://example.com/codeflow?start=68&length=55)

We can probably remove this exclusion - and remove the ploc file (it is not used and hasn't been updated since it was forked from the other repository)

CarlosFigueiraMSFT: It can be done in a separate PR, though

CarlosFigueiraMSFT (approved):
:shipit:

Shpakh-MSFT (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Microsoft.PowerFx.Core.csproj
lesaltzm: #309

MikeStall (approved):

PR #311, opened 2022-04-15
Fix ErrorResource locale bug

lesaltzm:
Basic strings looked up using the correct locale, but error resources were using the fallback. This fixes the issue and adds a test to catch that.


CarlosFigueiraMSFT (approved):
:shipit:

CarlosFigueiraMSFT (approved):
:shipit:

Issue #301, opened 2022-04-13
Support for Behavior properties


MikeStall:
Allow the semicolon chaining operator and behavior properties, for example: `A;B;C` 
This is just meaningful for behavior function (Side-effecting). 

- [ ] This requires a special flag to parser (AllowExpressionChaining ) to accept semicolons. 
- [ ] Clean public surface area, (may be related to https://github.com/microsoft/Power-Fx/pull/272 and https://github.com/microsoft/Power-Fx/pull/294 ) 
- [ ] Flow through binding 
- [ ] Support in the IR and Interpreter so it can be evaluated. 
- [ ] How do user defined functions (ie, reflectionFunction) specify they're a behavior function. 
- [ ]  ensure RecalcEngine can't set a formula to a behavior property

LucGenetier:
#322

PR #313, opened 2022-04-15
Move config option on INameResolver to a new BindingConfig type

lesaltzm:
These flags shouldn't be driven from the Name Resolver. 
Ideally, we should keep per-rule configuration separate from a name resolver, which gets shared more globally. This starts moving in that direction. 

contributes to #301


src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -315,12 +324,23 @@ private TexlBinding(IBinderGlue glue, IExternalRuleScopeResolver scopeResolver,
...
+            IExternalRuleScopeResolver scopeResolver,
+            DataSourceToQueryOptionsMap queryOptionsMap,
+            TexlNode node,
+            INameResolver resolver,
+            BindingConfig bindingConfig,
+            bool updateDisplayNames = false,
+            DType ruleScope = null,
+            bool forceUpdateDisplayNames = false,
+            IExternalRule rule = null,
+            bool useThisRecordForRuleScope = false)
```
MikeStall, line 338: Can some of these other parameters also get moved to  BindingConfig? (maybe in a future PR).

MikeStall:
I assume this will require a corresponding change to PAClient?

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
lesaltzm: Yeah, that's the idea. This is an initial direction, but we should get most of these "How should this expression bind" switches moved over to the config object.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -197,11 +197,7 @@ public IExternalControl Control
 
         public bool HasSelfReference { get; private set; }
 
-        public bool IsBehavior => NameResolver != null && NameResolver.CurrentPropertyIsBehavior;
-
-        public bool IsConstantData => NameResolver != null && NameResolver.CurrentPropertyIsConstantData;
-
-        public bool IsNavigationAllowed => NameResolver != null && NameResolver.CurrentPropertyAllowsNavigation;
```
CarlosFigueiraMSFT, line 204: >IsNavigationAllowed [](http://example.com/codeflow?start=20&length=19)

This is used in the Navigate function; are we going to use the same 'AllowSideEffects' flag to deal with it?

lesaltzm: No, that logic will stay in PA Client. Navigation isn't a concept used in PowerFx, the implementation in Canvas can cast to it's own name resolver impl to retrieve that flag.

LucGenetier (approved):

src/libraries/Microsoft.PowerFx.Core/Binding/BindingConfig.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// Configuration for an invocation of the binder.
+    /// </summary>
+    [ThreadSafeImmutable]
+    internal class BindingConfig
+    {
+        public static readonly BindingConfig Default = new BindingConfig(allowsSideEffects: false);
+
+        public bool AllowsSideEffects { get; }
+
+        public BindingConfig(bool allowsSideEffects)
```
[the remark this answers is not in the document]
MikeStall: Line 16 initializes an entire BindingConfig object (not just 1 flag). 
This will get more interesting as we add more fields to this class.

raghavthind2005 (approved):

src/libraries/Microsoft.PowerFx.Core/Binding/BindingConfig.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            AllowsSideEffects = allowsSideEffects;
```
rick-nguyen, line 22: Is ConstantData not necessary either here?

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
MikeStall: We have a lot of these "viral" flags... 
where my node has the flag if any of my children have the flag.  Some of these are general and some of them are canvas specific. 
Should we have something generic generic like a bit vector that we just OR together?

- constant, async, navigation,

src/libraries/Microsoft.PowerFx.Core/Binding/BindingConfig.cs
lesaltzm: It is not, same logic as navigation. Those are going to live in PA Client's name resolver and be cast if needed.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
lesaltzm: Agreed in general for binder. These (IsConstantData/IsNavigation/IsBehavior) ones are for what the top level property permits and are a separate but similarly named problem.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
lesaltzm: 2: Yes, that's the goal of this. 

1: Maybe? Still thinking about that. Would definitely make more sense. At least some of them can be grouped meaningfully into one set of flags.

PR #315, opened 2022-04-17
Added RecalcEngine Tests - Issue#308

omprakaash:
Added RecalEngine tests to document behavior of a SetFormula() call followed by an UpdateVariable() call on the same formula. I am not clear on the intended behaviour when UpdateVariable() is called to override a formula's value. Currently it seems to silently fail in overriding the formula's value.

### DIFF
--- a/src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
+++ b/src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
@@ -210,6 +210,22 @@ namespace Microsoft.PowerFx.Tests
         }
 
         [Fact]
+        public void VariableVersusFormula()
+        {
+            var engine = new RecalcEngine();
+            engine.SetFormula("A", "4", OnUpdate);
+            AssertUpdate("A-->4;");
+            engine.SetFormula("B", "A*10", OnUpdate);
+            AssertUpdate("B-->40;");
+
+            engine.UpdateVariable("A", FormulaValue.New(40));
+            AssertUpdate("A-->40;");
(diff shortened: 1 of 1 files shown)


src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -209,6 +209,22 @@ public void FormulaCantRedefine()
...
+        {
+            var engine = new RecalcEngine();
+            engine.SetFormula("A", "4", OnUpdate);
+            AssertUpdate("A-->4;");
+            engine.SetFormula("B", "A*10", OnUpdate);
+            AssertUpdate("B-->40;");
+
+            engine.UpdateVariable("A", FormulaValue.New(40));
+            AssertUpdate("A-->40;");
+            AssertUpdate("B-->400;");
```
MikeStall, line 223: This looks like it will fail given current implementation of AssertUpdate/OnUpdate.  AssertUpdate clears the buffer, so it can only be called once. You may need to combine these into 1 line, `AssertUpdate("A-->40;B-->400;");`, although that may beg sorting issues.

omprakaash: Yep it does fail, but with an actual update of (A-->4). The test fails even when I change the line(222) to AssertUpdate("A-->40;B-->400;");. A's value does not seem to change from its initial formula value.

omprakaash:
There seems to a calculation bug that occurs when multiple dependencies (diamond shaped graph) are present. Although I believe this could be easily avoided when topological sort is used to order the execution of variables.

Issue #308, opened 2022-04-14
Improve recalc engine algorithm 


MikeStall:
(this is a bit more advanced for a first issue - requires knowledge of [topological sorts](https://en.wikipedia.org/wiki/Topological_sorting?msclkid=57b2ed53bc0f11ec8f5e9d869b1106eb). But it's marked as a "first issue" since the code is very accessible with few dependencies) 

RecalcEngine's lets you define Formulas (via SetFormula) and then it will automatically recalc the formula if the dependencies change. 

But the implementation here is extremely hack:
https://github.com/microsoft/Power-Fx/blob/e2ffacc79cd57a2f5b9a30cc05c41460d3478fb0/src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngineWorker.cs#L18 

This is the ability to call UpdateVariable() and cause other Formulas (created from SetFormula) to be recalculated. 

It's good  enough to pass the tests, but a very poor implementation. 
See tests here:
https://github.com/microsoft/Power-Fx/blob/e2ffacc79cd57a2f5b9a30cc05c41460d3478fb0/src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs#L124

A good implementation:
- would statically understand what each formula depends on
- efficient recalc - does not unnecessary work. 
- robust handling for circular references
- include aggressive tests. 
- allow deleting a formula. 

The current implementation is also very inefficient. We could construct a perf benchmark to show before and after.

omprakaash:
Can I take up this issue? I am interested in working on this. My current understanding is that the implementation of RecalcEngine avoids cycles by only allowing new formulas to refer to previously defined formulas and by not allowing formulas to be redefined. Would an improved implementation allow for formula redefinition by detecting cycles in the top-sort algorithm ?

orcmid:
This is an off-hand comment.  I have not examined RecalcEngine, just musing from somewhere close to first principles.  I recall that there was once an iteration cap that was used to clamp off recalcs with cycles.  Is that still around?

> @omprakaash asks: Would an improved implementation allow for formula redefinition by detecting cycles in the top-sort algorithm ?

Technically, there is no topological sort of a directed graph having cycles longer than 1. I just dug this out of "The Art of Computer Programming", volume 1 Exercise 2.3.4.2-4.

So, the ideal conditions given by @MikeStall require some sort of hacks for which there is assured termination and minimal update (addition, change, deletion) consequences.   That's a large order.

[...]

PR #318, opened 2022-04-18
DeleteFormula implementation

anderson-joyle:
Adding the ability to delete formulas from the formula dict.
I'm releasing small pieces of code while still working on the main goal for #308.

### DIFF
--- a/src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
+++ b/src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
@@ -144,6 +144,26 @@ namespace Microsoft.PowerFx.Tests
             // Batched up (we don't double fire)            
             AssertUpdate("B-->20;C-->25;D-->22;");
         }
+        
+        [Fact]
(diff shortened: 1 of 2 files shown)


src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -144,6 +144,26 @@ public void Recalc2()
...
+
+            Assert.Throws<InvalidOperationException>(() =>
+                engine.DeleteFormula("X"));
+
+            Assert.Throws<InvalidOperationException>(() =>
+                engine.DeleteFormula("B"));
+
+            engine.DeleteFormula("D");
+            Assert.False(engine.Formulas.TryGetValue("D", out var ret));
+        }
```
MikeStall, line 166: Some other test cases:
- B fails, but if I delete C,D , can I then successfully delete B. 

- Can I readd B (perahps w/ different type) after deleting?
Ie, ensure that B really is fully deleted.

MikeStall (approved):
:shipit:

PR #320, opened 2022-04-18
TopSort for future multi-formula sort feature

bryancroteau-MSFT:
Draft PR for an experimental feature for sorting a set of formulas to produce an order in which they should be evaluated.


src/tests/Microsoft.PowerFx.Core.Tests/TopologicalSortTests.cs
```diff
@@ -0,0 +1,118 @@
...
+
+            var success = TopologicalSort.TrySort(nodes, edges, out var result, out var cycles);
+            Assert.True(success);
+            Assert.NotNull(result);
+            Assert.Null(cycles);
+            Assert.True(result.Count() == 0);
+        }
+
+        [Fact]
+        public void MissingNodeTest1()
```
MikeStall, line 89: >MissingNodeTest1 [](http://example.com/codeflow?start=20&length=16)

Test idea: 
node with no edges, returns success. (ie, empty edge list)

src/libraries/Microsoft.PowerFx.Core/Utils/FormulaSort/IdentifierVisitorScope.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+
+namespace Microsoft.PowerFx.Core.Utils.FormulaSort
+{
+    internal sealed class IdentifierVisitorScope
+    {
+        private readonly Dictionary<string, IdentifierVisitorMode> _idents = new Dictionary<string, IdentifierVisitorMode>();
```
MikeStall, line 10: >string [](http://example.com/codeflow?start=36&length=6)

nit: can this be DName? avoids the casing questions.

src/tests/Microsoft.PowerFx.Core.Tests/TopologicalSortTests.cs
bryancroteau-MSFT: Done

src/libraries/Microsoft.PowerFx.Core/Utils/FormulaSort/IdentifierVisitorScope.cs
bryancroteau-MSFT: Removed, this PR can focus just on Topsort

MikeStall (approved):
:shipit:

jcoc611-microsoft:
just wondering, wouldn't we want some data structure that maintains a topological sort and has cheap insert and remove operations?

bryancroteau-MSFT:
> 

We could add that as well, but we still need to do an initial sort before we have even constructed such a data structure.

jcoc611-microsoft:
I was thinking it would be a data structure with similar semantics to `SortedList` i.e. you don't have to insert items in a sorted order...as you insert items it keeps them sorted such that you don't have to sort beforehand

orcmid:
> @jcoc611-microsoft  asks: just wondering, wouldn't we want some data structure that maintains a topological sort and has cheap insert and remove operations?
> and @jcoc611-microsoft elaborates: I was thinking it would be a data structure with similar semantics to `SortedList`

Some thoughts from first principles (not examination of the current algorithms).

I think what you'd like is a linearization of nodes that preserves the topological-sort partial ordering but where one can rapidly determine what is impacted by a change on a dependency (an incoming directed edge).  Even forbidding cycles, there needs to be some heuristic work on how to deal with the ripple of a change to all that is impacted, and etc.

Intuitively, a deletion does not require change of the linearization (though one might carry a hole to be expedient).  It might require change to (other) formulas and so there is a reform/recalc potential there.

Insertion is more difficult since it adds (at least) one node and possibly some edges.  This impacts the partial ordering but there should be insertion points in the linearization that will preserve partial ordering.  (This suggests a symmetric list, and that works for deletion also.)

I think a replacement (as opposed to update of a scalar) might be handled best by a deletion followed by an insertion if there are different edges in the result.  It would be useful to not do two recalc, suggesting some mechanism for squashing however recalc needs are queued.

I also think there might be some confusion on type changing.  Inserting/updating a scalar is very different from inserting/updating a formula.  And changing type, if actually allowed, looks at least like the delete/insert case.

Does this fit the situation?

orcmid:
> I said, "Some thoughts from first principles (not examination of the current algorithms)."

The proposal of a symmetric list structure simplifies the cases I mentioned.  But it does mean one cannot find things by index; the locations are not ordered, the list is.  I thought there was an occasion for needing to walk the list, but I may be mistaken.  

Oh, one does need to determine a correct place to do an insertion without redoing the topological sort but transforming it in a deterministic manner.  I.e., can we assume the list is always sorted, in the sense that if _a_ ≺ _b_ topographically, then _a_ is earlier on the linear list than _b_ in sense that "≺" is used to reflect dependency. 

There needs to be some assistance in the existing dependencies to aid in the determination of where an insertion belongs.  One case is if a leaf can be turned into a node or a node can be updated.  If something has to be moved on the linear list to maintain the partial orderings, there needs to be a deterministic heuristic.  That is, when list order is incorrect as the result of an added topological constraint.  Fortunately, moving a cell on a symmetrical list does not change the location of the cell, it is done by adjusting links.  That's some help.

PS: I don't know if working-set proximity and such are an issue here and whether we care very much about cache and page faults while chasing a list whose cells are not always nearby from one to the next.  I am presuming that there are higher-order performance matters than that.

PR #322, opened 2022-04-19
Enable EnableExpressionChaining in PowerFx Engine

LucGenetier:


src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -82,7 +81,7 @@ public CheckResult Check(string expressionText, RecordType parameterType = null)
 
             var formula = new Formula(expressionText);
 
-            formula.EnsureParsed(TexlParser.Flags.None);
+            formula.EnsureParsed(GetTexlParserFlags());
```
MikeStall, line 84: >GetTexlParserFlags [](http://example.com/codeflow?start=33&length=18)

fyi, This will get superseded by this PR for exposing parseresult: [https://github.com/microsoft/Power-Fx/pull/319](https://github.com/microsoft/Power-Fx/pull/319)

src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/ScenarioMutation.cs
```diff
@@ -43,6 +43,33 @@ public void MutabilityTest()
...
+            {
+                ["prop"] = FormulaValue.New(123)
+            };
+
+            var obj = MutableObject.New(d);
+            engine.UpdateVariable("obj", obj);
+
+            var exprs = new string[]
+            {
+                "Assert2(obj.prop, 123); Set2(obj, \"prop\", 456); Assert2(obj.prop, 456)"
```
MikeStall, line 64: >"Assert2(obj.prop, 123); Set2(obj, \"prop\", 456); Assert2(obj.prop, 456)" [](http://example.com/codeflow?start=16&length=74)

Good test. 
Note expr can now be a string, not a string[].

src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/ScenarioMutation.cs
```diff
@@ -43,6 +43,33 @@ public void MutabilityTest()
...
+            };
+
+            foreach (var expr in exprs)
+            {
+                var x = engine.Eval(expr); // Assert failures will throw.
```
MikeStall, line 69: > // Assert failures will throw. [](http://example.com/codeflow?start=42&length=31)

I'm guessing this is from EvalVisitor? next step is to implement evaluation :)

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -33,18 +33,26 @@ public sealed class PowerFxConfig
 
         internal CultureInfo CultureInfo { get; }
 
-        private PowerFxConfig(CultureInfo cultureInfo, EnumStoreBuilder enumStoreBuilder) 
+        internal PowerFxFlags PowerFxFlags { get;  }
```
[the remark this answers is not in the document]
lesaltzm: Also see https://github.com/microsoft/Power-Fx/pull/313

src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/ScenarioMutation.cs
MikeStall: If this is passing, this test should have stronger asserts

src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/ScenarioMutation.cs
LucGenetier: Added one more test in the array

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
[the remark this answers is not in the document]
LucGenetier: Fixed

src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/ScenarioMutation.cs
LucGenetier: Fixed

src/tests/Microsoft.PowerFx.Core.Tests/IntellisenseTests/SuggestTest.cs
```diff
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-using System.Globalization;
```
MikeStall, line 4: >Globalization [](http://example.com/codeflow?start=13&length=13)

Can we revert these files? (leftover from a previous iteration; adds a using we don't need)

src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/ScenarioMutation.cs
MikeStall: Just make this a [Theory] and avoid string[] / for-loop.

src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/ScenarioMutation.cs
```diff
@@ -43,6 +46,42 @@ public void MutabilityTest()
...
+
+                var obj = MutableObject.New(d);
+                engine.UpdateVariable("obj", obj);
+
+                var x = engine.Eval(expr, options: new ParserOptions() { AllowsSideEffects = true }); // Assert failures will throw.
+
+                if (x is ErrorValue ev)
+                {
+                    throw new XunitException($"FormulaValue is ErrorValue: {string.Join("\r\n", ev.Errors)}");
+                }
```
MikeStall, line 79: this is asserting we're not a failure. 
The stronger assert is to assert we're returning the expected value. 

Although Assert2() is not an ideal function for this since it returns blank ;)

Maybe have that return something (even just 'true') and then we can verify.

MikeStall:
We should also be able to add this test to [https://github.com/microsoft/Power-Fx/blob/main/src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterExpressionTestCases/ParallelAsync.txt](https://github.com/microsoft/Power-Fx/blob/main/src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterExpressionTestCases/ParallelAsync.txt) 

>> WaitFor(0);WaitFor(1);WaitFor(2)
2

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
LucGenetier: Good catch

src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/ScenarioMutation.cs
LucGenetier: Done

MikeStall: This is good. Now you can remove the (x is ErrorValue ev) check - the new NumberValue check is a stronger assert that supersedes it.

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -83,13 +83,14 @@ protected override async Task<RunResult> RunAsyncInternal(string expr, string se
...
+                var iSetup = InternalSetup.Parse(setupHandlerName);
 
-                if (setupHandlerName == "AsyncTestSetup")
+                if (string.Equals(iSetup.HandlerName, "AsyncTestSetup", StringComparison.OrdinalIgnoreCase))
                 {
                     return new RunResult(await RunVerifyAsync(expr));
                 }
 
-                if (setupHandlerName != null) 
+                if (!string.IsNullOrEmpty(iSetup.HandlerName))
```
MikeStall, line 93: >IsNullOrEmpty [](http://example.com/codeflow?start=28&length=13)

nit: null check is a stronger enforcement than null or empty. So this is loosening a check. 
We should keep it strong (just null check) to simplify downstream callers,  and have test parser should fail if it's empty.

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -83,13 +83,14 @@ protected override async Task<RunResult> RunAsyncInternal(string expr, string se
                 FeatureFlags.StringInterpolation = true;
                 RecalcEngine engine;
                 RecordValue parameters;
+                var iSetup = InternalSetup.Parse(setupHandlerName);
```
MikeStall, line 86: >InternalSetup.Parse [](http://example.com/codeflow?start=29&length=19)

This is test parsing logic - all backends would need this, but the check here is only in interpreter. 

We should discuss the pattern here.

src/tests/Microsoft.PowerFx.Interpreter.Tests/InternalSetupTests.cs
```diff
@@ -0,0 +1,100 @@
...
+
+using System;
+using Microsoft.PowerFx.Core.Parser;
+using Microsoft.PowerFx.Core.Tests;
+using Microsoft.PowerFx.Interpreter.Tests.Helpers;
+using Xunit;
+
+namespace Microsoft.PowerFx.Interpreter.Tests
+{
+    public class InternalSetupTests : PowerFxTest
```
MikeStall, line 12: >InternalSetupTests [](http://example.com/codeflow?start=17&length=18)

This should be in FxCore\TestRunnerTests.

src/tests/Microsoft.PowerFx.Interpreter.Tests/InternalSetupTests.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            Assert.NotNull(iSetup);
+            Assert.Equal("SomeHandler", iSetup.HandlerName);
+            Assert.Equal(TexlParser.Flags.None, iSetup.Flags);
+        }
+
+        [Fact]
+        public void InternalSetup_Parse_OneFlag()
+        {
+            var iSetup = InternalSetup.Parse(TexlParser.Flags.NamedFormulas.ToString());
```
MikeStall, line 57: >TexlParser.Flags.NamedFormulas.ToString() [](http://example.com/codeflow?start=45&length=41)

from a test perspective - at least one of these cases should be a string literal. 

This change now makes the C# identifier name part of our "public" test contract. Ie, if somebody renamed the C# enum value, it could now break tests.  
So adding as a literal would explicitly catch that.

MikeStall: (as an example, see the LSP tests like TestLogCallbackExceptions - they pass in literals for "method" even though we have symbols).

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
LucGenetier: If the handler name would be an empty, it would ignore it and execute the code normally like if there was no handler
With the null check, it will throw an exception
Either way, I'm fine but with a little preference for the first one (accepting more)

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
LucGenetier: I moved InternalSetup (and related UTs) to Core.Texts\TestRunnerTests

src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterExpressionTestCases/ParallelAsync_Chain.txt
```diff
@@ -0,0 +1,4 @@
+﻿#SETUP: AsyncTestSetup, EnableExpressionChaining
+
+>> WaitFor(0); WaitFor(1); WaitFor(2)
```
MikeStall, line 3: >>> [](http://example.com/codeflow?start=0&length=2)

Can you move into existing one: 
[https://github.com/microsoft/Power-Fx/blob/main/src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterExpressionTestCases/ParallelAsync.txt](https://github.com/microsoft/Power-Fx/blob/main/src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterExpressionTestCases/ParallelAsync.txt) 

And then modify the #Setup tag there to this new one:
#SETUP: AsyncTestSetup, EnableExpressionChaining

src/tests/Microsoft.PowerFx.Interpreter.Tests/InternalSetupTests.cs
LucGenetier: Added a test with a string

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -469,7 +469,22 @@ public override async ValueTask<FormulaValue> Visit(ColorLiteralNode node, Symbo
         public override async ValueTask<FormulaValue> Visit(ChainingNode node, SymbolContext context)
```
MikeStall, line 469: >ValueTask [](http://example.com/codeflow?start=30&length=9)

This is the key part of the change to actually enable evaluation:

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -140,19 +140,19 @@ public void UpdateVariable(string name, FormulaValue value)
...
-        public async Task<FormulaValue> EvalAsync(string expressionText, CancellationToken cancel, RecordValue parameters = null)
+        public async Task<FormulaValue> EvalAsync(string expressionText, CancellationToken cancel, RecordValue parameters = null, ParserOptions options = null)
         {
             if (parameters == null)
             {
                 parameters = RecordValue.Empty();
             }
 
-            var check = Check(expressionText, (RecordType)parameters.IRContext.ResultType);
+            var check = Check(expressionText, (RecordType)parameters.IRContext.ResultType, options);
```
MikeStall, line 155: >Check [](http://example.com/codeflow?start=24&length=5)

I think there's still a followup issue about enforcing / rejecting behavior functions. But we don't have behavior funcs in fx yet (and custom Funcs don't allow specifying behavior properties yet).

MikeStall (approved):
:shipit:

PR #323, opened 2022-04-19
CountA function implementation

anderson-joyle:
Issue #284 
Sending this draft to ensure I'm on the right path.

Out of the 3 tests added to the CountA.txt file, just one pass (line 1).
Blank values in the array are being assigned as RecordValue.IsValue = true. Am I missing something?

Thanks in advance.


lesaltzm:
I think this is an issue with SingleColumnTables. CountA considers both null and {Column: null} to be blank for the purposes of exclusion, but IsValue that you're using in the implementation only checks the first. 

[a,b,c] is syntactic sugar for Table({Value: a}, {Value: b}, {Value: c}), so those nulls would need the 2nd case. 

For more context, see the JS implementation of CountA:
image

anderson-joyle:
Thanks for taking the time to clarify @lesaltzm.
This latest PR contains the fix and also the Count implementation.
One thing that I want to ask though is why the test **CountA(Blank())** says "did not return a value"?

From my understanding, this should result in Blank() value since the CountA return behavior was declared as **ReturnBehavior.ReturnBlankIfAnyArgIsBlank**.

```
{
              BuiltinFunctionsCore.CountA,
              StandardErrorHandling<TableValue>(
                  replaceBlankValues: DoNotReplaceBlank,
                  checkRuntimeTypes: ExactValueTypeOrBlank<TableValue>,
                  returnBehavior: ReturnBehavior.ReturnBlankIfAnyArgIsBlank,
                  targetFunction: CountA)
}, 
```

Thanks in advance.

lesaltzm:
For CountA(Blank()), the first arg to countA needs to be a table, so you're running into a typechecking error. We should probably surface that better in the test harness.

PR #324, opened 2022-04-19
Add 6000+ new .txt tests written by Mitu.

MikeStall:
Many of these don't pass yet, so putting in a "NotYetReady" directory and we'll migrate to ExpressionTestCase dir as we verify they pass.

Add unit test to verify they at least all parse, and fix some parser issues.


src/tests/Microsoft.PowerFx.Core.Tests/TestRunnerTests/Bad3.txt
```diff
@@ -0,0 +1,3 @@
+﻿>>
+>> Abs(2)
+2
```
Shpakh-MSFT, line 3: What is this file?

Shpakh-MSFT (approved):
:shipit:

MikeStall:
> >>

Bad*.txt files let us test that we catch parser errors. 
(common to get parse errors when writing so many tests...)



---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/TestRunnerTests/Bad3.txt:1 in 1f9e6a2. [](commit_id = 1f9e6a2451a2191679adab52ab2382020ebe9527, deletion_comment = False)

PR #326, opened 2022-04-19
Make Lexer thread safe

lesaltzm:
We only actually support two sets of punctuators in practice : `. , ;` and `, ; ;;`. This cleans up a bunch of statics in the Lexer and moves to use CurrentCulture's decimal separator for picking which lexer to use.


src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -76,316 +68,67 @@ public enum Flags
...
-            PopulateKeywordArrays();
-        }
+        public string LocalizedPunctuatorChainingSeparator { get; }
 
-        private void AddKeyword(string str, TokKind tid)
-        {
-            Contracts.AssertNonEmpty(str);
-            _keywords.Add(str, tid);
-        }
+        private Tuple<string, Flags, Token[]> _cache;
```
lesaltzm, line 126: This Cache is the last thing I need to clean up to mark TexlLexer  as ThreadSafeImmutable

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -76,316 +68,67 @@ public enum Flags
...
-            AddKeyword(KeywordSelf, TokKind.Self);
-            AddKeyword(KeywordParent, TokKind.Parent);
-            AddKeyword(KeywordAnd, TokKind.KeyAnd);
-            AddKeyword(KeywordOr, TokKind.KeyOr);
-            AddKeyword(KeywordNot, TokKind.KeyNot);
-            AddKeyword(KeywordAs, TokKind.As);
-
-            PopulateKeywordArrays();
-        }
+        private static readonly Lazy<TexlLexer> _invariantDecimalSeparatorLexer = new Lazy<TexlLexer>(() => new TexlLexer(PunctuatorDecimalSeparatorInvariant));
```
MikeStall, line 103: >Lazy [](http://example.com/codeflow?start=32&length=4)

can we avoid using Lazy<T>?  It's an extra level of complexity and We already inevitably create these, so lazy<t> doesn't buy us much. 
And C#'s static ctor fields will already accomplish something very similar.

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
MikeStall: Do we even need the _cache here? it only handles a single item- how often do we actually get a cache hit?

lesaltzm: Yeah, agreed. Removed the cache, seems like it's not a huge benefit to maintain.

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -454,34 +195,124 @@ private void PopulateKeywordArrays()
...
+            return strList;
+        }
+
+        private TexlLexer(string preferredDecimalSeparator)
+        {
+            // List and decimal separators.
+            // These are the default global settings. If there is a collision between the two,
+            // the list separator automatically becomes ;.
+            LocalizedPunctuatorDecimalSeparator = ChooseDecimalSeparator(preferredDecimalSeparator);
+            LocalizedPunctuatorListSeparator = ChooseListSeparatorPunctuator();
```
CarlosFigueiraMSFT, line 224: >ChooseListSeparatorPunctuator [](http://example.com/codeflow?start=47&length=29)

There is an expected ordering of execution of this method - it needs to run after `LocalizedPunctuatorDecimalSeparator` is initialized - but it's not clear. You can pass the decimal separator as a parameter, and that would make the ordering explicit. Same for ChooseChainingPunctuator.

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -510,7 +341,6 @@ public Token[] LexSource(string text, Flags flags = Flags.None)
 
             // Update the cache and return the result
```
CarlosFigueiraMSFT, line 342: >// Update the cache and return the result [](http://example.com/codeflow?start=12&length=41)

stale comment

src/tests/Microsoft.PowerFx.Core.Tests/LexerTests.cs
```diff
@@ -260,24 +254,11 @@ public void TestLexGetFragment(string script, int min, int lim, string fragment)
             Assert.Equal(fragment, span.GetFragment(script));
         }
 
-        private ILanguageSettings SomeFrenchLikeSettings()
+        private CultureInfo SomeFrenchLikeSettings()
```
CarlosFigueiraMSFT, line 257: >SomeFrenchLikeSettings [](http://example.com/codeflow?start=28&length=22)

[minor] Can you rename it to `GetFrenchSettings` now, since it's not just "French-like"? :-)

CarlosFigueiraMSFT (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -599,39 +427,39 @@ public string RemoveWhiteSpace(string text)
         }
 
         // Enumerate all supported unary operator keywords.
-        public string[] GetUnaryOperatorKeywords() => _unaryOperatorKeywords;
+        public static string[] GetUnaryOperatorKeywords() => _unaryOperatorKeywords.ToArray();
 
         // Enumerate all supported binary operator keywords.
-        public string[] GetBinaryOperatorKeywords() => _binaryOperatorKeywords;
+        public static string[] GetBinaryOperatorKeywords() => _binaryOperatorKeywords.ToArray();
```
MikeStall, line 433: >.ToArray() [](http://example.com/codeflow?start=85&length=10)

Does this need to be an array? or can we return the ReadOnlyList?

Else, this is allocating an array on each call.

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -453,34 +193,124 @@ private void PopulateKeywordArrays()
...
             };
+        }
+
+        public static TexlLexer GetLocalizedInstance(CultureInfo culture)
+        {
+            culture ??= CultureInfo.InvariantCulture;
+            return culture.NumberFormat.NumberDecimalSeparator == PunctuatorDecimalSeparatorInvariant ? InvariantLexer : CommaDecimalSeparatorLexer;
+        }
+
+        public static IList<string> GetKeywordDictionary()
```
MikeStall, line 204: >GetKeywordDictionary [](http://example.com/codeflow?start=36&length=20)

Is this called? (I don't see a caller) 

If so, can it just return _keywords.Values?  (avoid the array creation)

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -75,316 +67,64 @@ public enum Flags
...
-        // This creates a new Lexer with the same locale as the original, which is a lot less expensive than spinning up a whole new one
-        private TexlLexer(TexlLexer original)
-        {
-            Contracts.AssertValue(original);
+        public static TexlLexer InvariantLexer { get; } = new TexlLexer(PunctuatorDecimalSeparatorInvariant);
 
-            LocalizedPunctuatorDecimalSeparator = original.LocalizedPunctuatorDecimalSeparator;
-            LocalizedPunctuatorListSeparator = original.LocalizedPunctuatorListSeparator;
-            LocalizedPunctuatorChainingSeparator = original.LocalizedPunctuatorChainingSeparator;
+        public static TexlLexer CommaDecimalSeparatorLexer { get; } = new TexlLexer(PunctuatorCommaInvariant);
```
MikeStall, line 104: IT looks like this is one of the key simplifications? That we really only have 2 lexers, and so we don't need a CultureInfo-->Lexer cache.

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -75,316 +67,64 @@ public enum Flags
...
-        }
+        // Limits the StringBuilderCache TLS memory usage for LexerImpl.
+        // Usually our tokens are less than 128 characters long, unless it's a large string.
+        private const int DesiredStringBuilderSize = 128;
 
-        // This creates a new Lexer with the same locale as the original, which is a lot less expensive than spinning up a whole new one
-        private TexlLexer(TexlLexer original)
-        {
-            Contracts.AssertValue(original);
+        public static TexlLexer InvariantLexer { get; } = new TexlLexer(PunctuatorDecimalSeparatorInvariant);
```
MikeStall, line 102: >{ get; } = new  [](http://example.com/codeflow?start=47&length=15)

As a property, this looks like it's returning an existing lexer; but each call to the property is actually creating a new lexer ....

Ie, it breaks this expectation: 
obj.ReferenceEquals(a.Prop1, a.Prop1);  


But also, creating a new lexer is expensive ... we still recreate the _punctuators dictionaries... that's state that could be shared across invokes.

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
lesaltzm: Yes, this is one of the key pieces. We only actually ever have two lexers

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
MikeStall: used in PAClient... LanguageHelper

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
lesaltzm: Discussed offline, confusion between `obj A { get; } = new ...` and `obj A {get { return new ... } }`
Will add a test covering the identity discussion.

MikeStall (approved):

PR #327, opened 2022-04-20
Statistical Functions(StdevP, StdevPT, VarP, VarPT) implementation.

omprakaash:
Implemented some statistical functions in the interpreter and included tests.
Edit: Used Welford's algorithm to implement VarianceAggreator. Is that okay or is there any other specific alg/implementation that is generally used for variance : https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm.

Seems to avoid numerical issues and avoids doing two passes through the data to find the mean of the set first.


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/TableMathfuncs.txt
```diff
@@ -34,4 +34,49 @@ Blank()
 2
 
 >> Average([1, 2, 1/0, 4, 5], Value)
-#Error
\ No newline at end of file
+#Error
+
+>> VarP(1, 3, 5)
```
lesaltzm, line 39: We're moving towards putting function-specific tests in their own .txt file. Makes searching easier, and it becomes kind of a spec for the function.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -57,6 +57,58 @@ public virtual FormulaValue GetResult(IRContext irContext)
...
+
+            public virtual FormulaValue GetResult(IRContext irContext)
+            {
+                if (_count == 0)
+                {
+                    return CommonErrors.DivByZeroError(irContext);
+                }
+                else
+                {
+                    return new NumberValue(irContext, _m2Acc / _count);
```
lesaltzm, line 92: Looks good!
I think NaN and +/- Inf as the result of the division should also return an error.

omprakaash: Got it. Added the necessary checks.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/TableMathfuncs.txt
omprakaash: Moved the VarP and StdevP tests into their own files, but left sum and average alone as I noticed they have unused tests in the NotYetReady test-case folder.

lesaltzm: Thanks. The existing ones will get cleaned up later, no need to worry about that for now. Just trying to make sure new ones are sensible.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
lesaltzm: We have a function for this, it's a pretty common case. Consider writing the return statement as 
`return FiniteChecker(irContext, 0, new NumberValue(irContext, _m2Acc/_count))`

omprakaash: Modified the function to use FiniteChecker !

lesaltzm (approved):

MikeStall (approved):
:shipit:

MikeStall:
> >> StdevP(1,2,3)

fwiw - I double checked that all these new tests do pass on the JS implementation. 

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/StdevP.txt:1 in f194fff. [](commit_id = f194fffcfde8f2b1ca3648a60c2766849da8b781, deletion_comment = False)

PR #331, opened 2022-04-20
Add the FormulaSet class

bryancroteau-MSFT:
A class that contains a set of formulas and maintains a topological sort of those formulas internally. Adds a new interface, IDependencyFinder, that implementers can use to return a set of dependencies given a formula.

Also modifies topological sort so that it can return both a list of results and a list of cycles. This is to allow for heuristics that don't fail if cycles exist.


src/libraries/Microsoft.PowerFx.Core/Public/FormulaSet.cs
```diff
@@ -0,0 +1,110 @@
...
+using System.Linq;
+using Microsoft.PowerFx.Core.Utils;
+
+namespace Microsoft.PowerFx.Core.Public
+{
+    /// <summary>
+    /// A set of name/formula paris, which creates and updates a topologically-sorted list
+    /// of name/formula pairs.
+    /// </summary>
+    public class FormulaSet
```
lesaltzm, line 15: Why public?

bryancroteau-MSFT: This could be a very useful class in cases where you have multiple formulas and you want a topological sort for those formulas.

src/libraries/Microsoft.PowerFx.Core/Public/FormulaSet.cs
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        private void SortFormulas()
+        {
+            var nodes = _formulas.Keys.ToList();
+            var edges = new List<TopologicalSortEdge<string>>();
+
+            foreach (var kvp in _formulas)
+            {
+                var checkResult = _engine.Check(kvp.Value._expression, kvp.Value._schema);
```
lesaltzm, line 43: Do we need to rerun _check on every formula to sort, or can we cache bindings we know haven't changed?

src/libraries/Microsoft.PowerFx.Core/Public/FormulaSet.cs
lesaltzm: Wouldn't we expect the RecalcEngine APIs to manage that, or is this more for other back ends eventually?

src/libraries/Microsoft.PowerFx.Core/Public/FormulaSet.cs
bryancroteau-MSFT: We could and we probably should

bryancroteau-MSFT: So the call to .Check is now abstracted behind an interface, so if the implementation wants to do caching, it can.

src/libraries/Microsoft.PowerFx.Core/Public/IDependencyFinder.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using System.Collections.Generic;
+
+namespace Microsoft.PowerFx.Core.Public
+{
+    /// <summary>
+    /// Exposes the ability parse a formula, walk the tree and return a list of dependencies
+    /// or variable names as strings.
+    /// </summary>
+    public interface IDependencyFinder
```
lesaltzm, line 12: This doesn't seem all that useful as a public interface. I see it's used for FormulaSet to avoid exposing DependencyFinder directly, but I'm not sure that makes sense.

lesaltzm: DependencyFinder's method is internal and static, couldn't you just use it directly without the reference?

bryancroteau-MSFT: Made it internal

lesaltzm (approved):

PR #335, opened 2022-04-22
Implement Shuffle

DavidLuong98:
Implements Suffle. #284

Adds unit tests, but I'm having trouble making this more deterministic - opened for some feedback / suggestions.

I assured the same numbers of records are being returned and I sorted the records to assure that each values in the table exist after the shuffle.


rick-nguyen:
Looks good - I'd maybe add a test to shuffle a table with one record, and tests to shuffle blank and error.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Shuffle.txt
```diff
@@ -0,0 +1,5 @@
+>> CountRows(Shuffle([-2, -1, 5, 1, 2]))
+5
+
+>> Sort(Shuffle([-2, -1, 5, 1, 2]), Value)
+[-2,-1,1,2,5]
```
lesaltzm, line 5: Hard to write a test that captures random behavior here. 
Maybe something like

`With({SortedFirstItems: Sort(ForAll(Sequence(2000), First(Shuffle([5,6,7,8,9]).Value))}, First(SortedFirstItems).Value <> Last(SortedFirstItems).Value)`
true

lesaltzm: if there was a clean way to set the randomizer seed we use, you could also use a `#SETUP: RandomizerTestSetup`
at the top of the file, like OptionSet.txt and make the randomizer seed fixed in a setup handler you write in src\tests\Microsoft.PowerFx.Interpreter.Tests\PowerFxEvaluationTests.cs

MikeStall: My concern here is that seeding the random number locks us into a very specific implementation and platform. It'd would for .net (and possibly only a specific build version), but how would a JS test runner use the seed?

MikeStall: fwiw, we had a similar challenge for Rand().  Tough to do deterministically especially in a .txt.

lesaltzm: Yeah, true. I think we have to go with indirectly testing randomness like above then.

MikeStall, line 5: Can you add a few other test cases:
(I looked at our existing tests for the JS implementation for inpsiration)

Arrays that contain:
- duplicate values 
- errors, blanks,
- non numbers

I think these will all just work with the current implementation. 


Also: 
Shuffle(1/0)  // pass in an error instead of array
Shuffle(blank()) // pass in blank instead of array 

Shuffle([]) //  empty array

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -212,6 +212,20 @@ public static FormulaValue IndexTable(IRContext irContext, FormulaValue[] args)
             return arg0.Index(rowIndex).ToFormulaValue();
         }
 
+        public static async ValueTask<FormulaValue> Shuffle(EvalVisitor runner, SymbolContext symbolContext, IRContext irContext, FormulaValue[] args)
```
MikeStall, line 215: >EvalVisitor runner, SymbolContext symbolContext,  [](http://example.com/codeflow?start=60&length=49)

The signature here could be simpler (see IndexTable above). 
It doesn't need to be async; and doesn't need runner object (since it doesn't invoke lambdas). So it could just be:

FormulaValue Shuffle(IRContext irContext, FormulaValue[] args)

There are some overloads in the library setup that allow various signatures.

DavidLuong98: Thanks, i addressed this change

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Shuffle.txt
DavidLuong98: I added the unit tests

MikeStall (approved):
:shipit:

lesaltzm (approved):

PR #339, opened 2022-04-22
Allow public derivations from RecordValue

MikeStall:
Enable public hosts to derive from RecordValue.
RecordValue doesn't need an IRContext (which is private), just a Type.

RecordValue.Fields shouldn't be virtual - it can already be inferred form the RecordType. 
 
This moves the type propagation from occurring in the InMemoryRecordValue ctor to now being down lazily via a CompileTimeTypeWrapperRecordValue.  This is necessary so that derived RecordValues don't get eagerly flattened into a InMemoryRecordValue.


bryancroteau-MSFT (approved):

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
```diff
@@ -15,48 +15,109 @@ namespace Microsoft.PowerFx.Core.Public.Values
...
+        /// <summary>
+        /// Derived classes can override. 
+        /// </summary>
+        /// <param name="fieldType"></param>
+        /// <param name="fieldName"></param>
+        /// <returns></returns>
+        protected virtual bool TryGetField(FormulaType fieldType, string fieldName, out FormulaValue result)
         {
             // Derived class can have more optimized lookup.
             foreach (var field in Fields)
```
MikeStall, line 110: >    foreach (var field in Fields) [](http://example.com/codeflow?start=8&length=33)

I have a problem - 
there's a circular reference between TryGetField --> Fields --> TryGetFields. 

TryGetFields should be abstract.

src/libraries/Microsoft.PowerFx.Core/Public/Values/CompileTimeTypeWrapperRecordValue.cs
```diff
@@ -0,0 +1,46 @@
...
+    // Project the correct compile-time type onto the runtime value. 
+    // Important for union / intersection types, such as Table() or If(). For example:
+    //    First(Table({a:1},{b:2})), result is a record with both fields a and b. 
+    internal class CompileTimeTypeWrapperRecordValue : RecordValue
+    {
+        private readonly RecordValue _inner;
+
+        public static RecordValue AdjustType(RecordType expectedType, RecordValue inner)
+        {
+            // TODO - Could this check be in the IR?  And so have IR add calls to TypeOverrideRecordValue?
```
lesaltzm, line 17: I don't think so, record values with a subset of the expected fields could also be passed in from the host, or from connectors. We wouldn't know that at compile time.

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
MikeStall: fixed in latest.

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
```diff
@@ -78,14 +78,20 @@ public FormulaValue GetField(string fieldName)
             return GetField(fieldType, fieldName);
         }
 
+        // Create an exception object for when the host violates the TryGetField() contract. 
+        private Exception HostException(string fieldName, string message)
+        {
+            return new InvalidOperationException($"{GetType().Name}.TryGetField({fieldName}): {message}");
```
bryancroteau-MSFT, line 84: This will format the error message with a field value instead of a name, right?

src/libraries/Microsoft.PowerFx.Core/Public/Values/CompileTimeTypeWrapperRecordValue.cs
```diff
@@ -0,0 +1,45 @@
...
+        private CompileTimeTypeWrapperRecordValue(RecordType type, RecordValue inner)
+            : base(type)
+        {
+            _inner = inner;
+        }
+
+        protected override bool TryGetField(FormulaType fieldType, string fieldName, out FormulaValue result)
+        {
+            // If the runtime value is missing a field of the given type, it will be Blank().
+            result = _inner.GetField(fieldType, fieldName);
```
bryancroteau-MSFT, line 34: Why GetField and not TryGetField?

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
MikeStall: The error message tells the fieldname that was being accessed. Most importantly, it tells the derived type (which is likely where the bug is).

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
```diff
@@ -15,51 +15,135 @@ namespace Microsoft.PowerFx.Core.Public.Values
...
+        /// <param name="fieldName">Name of field on this record.</param>
+        /// <returns>Field value or blank if missing. </returns>
+        public FormulaValue GetField(string fieldName)
+        {
+            if (fieldName == null)
+            {
+                throw new ArgumentNullException(nameof(fieldName));
+            }
+
+            var fieldType = Type.MaybeGetFieldType(fieldName) ?? FormulaType.Blank;
```
bryancroteau-MSFT, line 76: why would this ever fail?

src/libraries/Microsoft.PowerFx.Core/Public/Values/CompileTimeTypeWrapperRecordValue.cs
MikeStall: visibility checks. Not allowed to call a protected method here on a different derived type (even though they have the same base class).

src/libraries/Microsoft.PowerFx.Core/Public/Values/InMemoryRecordValue.cs
```diff
@@ -1,57 +1,47 @@
...
 namespace Microsoft.PowerFx.Core.Public.Values
 {
+    // Represent record backed by known list of values. 
     internal class InMemoryRecordValue : RecordValue
     {
-        private readonly Dictionary<string, FormulaValue> _fields = new Dictionary<string, FormulaValue>();
-
-        public override IEnumerable<NamedValue> Fields =>
-            from field in _fields select new NamedValue(field);
+        private readonly IReadOnlyDictionary<string, FormulaValue> _fields;
```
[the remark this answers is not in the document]
MikeStall: field access should be unordered. (and we use dict elsewhere too)

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
MikeStall: I did a test run and here are 2 cases where it fails:
>> Index([1, 2, 3, 4, 5], Error({Kind: 11}))

>> Filter(Table({a:1},{a:2},If(1<0,{a:3}),{a:4},{a:5}), IsBlank(a) Or a > 0)

bryancroteau-MSFT (approved):

PR #342, opened 2022-04-25
Add 1275 tests and fix several math and text issues

LucGenetier:


src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
```diff
@@ -41,7 +46,113 @@ public void InterpreterTestCase(ExpressionTestCase testCase)
                     Skip.If(true, prefix + msg);
                     break;
             }
-        } 
+        }        
+
+        [InterpreterTheory] //(Skip="NotReadyYet")]
```
MikeStall, line 51: >InterpreterTheory [](http://example.com/codeflow?start=9&length=17)

Is the intention to check this in? Or just keep for private testing?

MikeStall:
>                 Kind = ErrorKind.BadLanguageCode

I wonder if this one is wrong too? 
(and the one in InvalidDateTimeError)

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs:68 in 39804d2. [](commit_id = 39804d26701fade9ce4dc830bc93b5a278e37b2a, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
```diff
@@ -53,6 +53,11 @@
     </Content>
   </ItemGroup>
 
+  <ItemGroup>
+    <None Remove="ExpressionTestCases\Abs_SkipDate.txt" />
+    <None Remove="ExpressionTestCases\Average_SkipDate.txt" />
+  </ItemGroup>
```
MikeStall, line 59: Do we need this in the project file?

For interpreter, you can add files to [https://github.com/microsoft/Power-Fx/tree/main/src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterExpressionTestCases](https://github.com/microsoft/Power-Fx/tree/main/src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterExpressionTestCases)

and they'll get overlaid and write over the core ones.

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
LucGenetier: I was thinking at private testing but we can check that in with Skip="NotReadyYet" while we fix all functions

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -33,7 +33,7 @@ internal static partial class Library
                 BuiltinFunctionsCore.Abs,
                 StandardErrorHandling<NumberValue>(
                     expandArguments: NoArgExpansion,
-                    replaceBlankValues: ReplaceBlankWithZero,
```
lesaltzm, line 36: As far as I know, this is intended for excel compat and the JS impl is wrong. We have a backlog item to bring the JS impl in line with this. 
@CarlosFigueiraMSFT 
image

CarlosFigueiraMSFT: That's correct. Moving forward, we will treat blank values as 0 *in numeric context*, which is the case for all math functions. We will introduce a new flag soon in Canvas apps to move to this new "Excel-compat" behavior.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Abs_SkipDate.txt
```diff
@@ -0,0 +1,15 @@
+﻿
+// ******** FUNCTION VALUE AS PARAMETERS ********
```
MikeStall, line 2: >// [](http://example.com/codeflow?start=0&length=2)

The *Skip and override tests should be in the /interpreter override folder. 
They're interpreter specific issues; and the test harness will merge them together and apply the overrides.

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
```diff
@@ -41,7 +46,113 @@ public void InterpreterTestCase(ExpressionTestCase testCase)
...
+            // make sure they build is actually copying them over. 
+            Assert.True(testCase.FailMessage == null, testCase.FailMessage);
+
+            _runner = new InterpreterRunner();
+
+            var (result, msg) = _runner.RunAsync(testCase).Result;
+
+            var prefix = $"Test {Path.GetFileName(testCase.SourceFile)}:{testCase.SourceLine}: ";            
+
+            switch (result)
```
MikeStall, line 65: >switch [](http://example.com/codeflow?start=12&length=6)

If you want to hack something to skip all tests containing blank, you could do something like:

if (testCase.Input.Contains("Blank()")) { result = TestResult.Skip; }

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
LucGenetier: Thanks, I'll fix the code and corresponding tests

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
LucGenetier: I've a better idea on how to handle Blank() for all functions

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Abs_SkipDate.txt
LucGenetier: Looks strange (tests under one project, overrides in another) but I'll follow the existing pattern

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Abs.txt
```diff
@@ -74,15 +76,18 @@ true
...
 
+>> Date(1976, 12,23) - Date(2000, 01, 01)
+-8409
+
 >> Abs(Date(1976, 12,23) - Date(2000, 01, 01))
 8409
 
 //Excel returns false, as the value of Abs(Date(2022, 10, 1)) in Excel is 44835 in General format, while in Date format its 10/1/2022
 >> Abs(Date(2022, 10, 1)) = 1664607600000
 true
```
CarlosFigueiraMSFT, line 90: Not related to your changes, but this should return the same as Excel in C#...

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Abs.txt
```diff
@@ -74,15 +76,18 @@ true
...
 
-
 //Abs(Infinity)
 //Abs(NaN)
```
CarlosFigueiraMSFT, line 93: Those are not valid PowerFx expressions, can you remove those?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Abs.txt
LucGenetier: I need to move all these tests to exclude them
I'll update Date() code and tests in a separated PR

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Abs.txt
LucGenetier: Ok

src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterExpressionTestCases/AbsOverride.txt
```diff
@@ -0,0 +1,15 @@
+﻿
+// ******** FUNCTION VALUE AS PARAMETERS ********
+
+// Should be -8409
```
MikeStall, line 4: can we comment why we're skipping these. Is it just not-yet-impl functionality?

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
```diff
@@ -41,7 +46,111 @@ public void InterpreterTestCase(ExpressionTestCase testCase)
...
+
+            return TestCategory.Unknown;            
+        }
+
+        private static readonly ConcurrentDictionary<string, (TestCase TestCase, string Message)> Stats = new ();
+
+        [Fact]
+        public void GetStats()
+        {
+            Thread.Sleep(1000); // Let's things start
```
MikeStall, line 120: With Thread.Sleep, Let's not check this in :)

src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterExpressionTestCases/AbsOverride.txt
LucGenetier: All Date calculations are incorrect

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
LucGenetier: WIP code, just temporary to extract some statistics
I won't check that in.

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
LucGenetier: I'll remove that code before checking this in

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
LucGenetier: Removed

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Abs.txt
LucGenetier: Done

src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterExpressionTestCases/AbsOverride.txt
LucGenetier: I've removed these files & recreated new ones under "NotReadyYet" for what needs to be fixed

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/NotYetReady/MinMax.txt
```diff
@@ -0,0 +1,6 @@
+﻿>> Min(Date(2022,3,8),DateTimeValue("March 08, 2022 1:50 PM"))
+DateTime(2022,3,8,0,0,0,0)
+
+>> Max(If(Left("hello", -1), [Date(2020,1,1), Date(2021,2,2)]), Value)
+#Error(Kind=Argument)
```
CarlosFigueiraMSFT, line 5: >Argument [](http://example.com/codeflow?start=12&length=8)

`InvalidArgument`?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -46,11 +51,26 @@ public void Apply(FormulaValue value)
...
+            {
+                return new NumberValue(context, _defaultN);
+            }
+
             public virtual FormulaValue GetResult(IRContext irContext)
             {
                 if (_count == 0)
                 {
-                    return new BlankValue(irContext);
+                    return GetDefault(irContext);
```
CarlosFigueiraMSFT, line 68: >return GetDefault(irContext); [](http://example.com/codeflow?start=20&length=29)

This is a change from the current behavior; while I think this is correct (`0` is the "neutral / identity" value for sum), it would be good to run through @gregli-msft and @jorisdg before checking this in

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -110,23 +140,30 @@ public override FormulaValue GetResult(IRContext irContext)
...
             }
 
+            public FormulaValue NoElementValue(IRContext context)
+            {
+                return GetDefault(context);
+            }
+
+            public FormulaValue GetDefault(IRContext context)
+            {
+                return new NumberValue(context, 0);
```
CarlosFigueiraMSFT, line 164: >return new NumberValue(context, 0) [](http://example.com/codeflow?start=16&length=34)

Similar comment than the one for Sum; should we return 0? 0 isn't really a "neutral" value for min.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -136,21 +173,28 @@ public FormulaValue GetResult(IRContext irContext)
...
             }
 
+            public FormulaValue NoElementValue(IRContext context)
+            {
+                return GetDefault(context);
+            }
+
+            public FormulaValue GetDefault(IRContext context)
+            {
+                return new DateValue(context, new DateTime(1900, 1, 1));
```
CarlosFigueiraMSFT, line 195: >new DateTime(1900, 1, 1) [](http://example.com/codeflow?start=46&length=24)

Use `_defaultDT`

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -136,21 +173,28 @@ public FormulaValue GetResult(IRContext irContext)
         private class MinDateTimeAgg : IAggregator
         {
             protected DateTime _minValueDT = DateTime.MaxValue;
+            protected static readonly DateTime _defaultDT = new DateTime(1900, 1, 1);
```
CarlosFigueiraMSFT, line 176: >new DateTime(1900, 1, 1) [](http://example.com/codeflow?start=60&length=24)

If we want to use '0', then the value should be `new DateTime(1899, 12, 30)`

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -510,9 +612,14 @@ public static FormulaValue Mod(IRContext irContext, NumberValue[] args)
...
 
             // r = a – N × floor(a/b)
-            var q = (int)Math.Floor(arg0 / arg1);
+            var q = (long)Math.Floor(arg0 / arg1);
             var result = arg0 - (arg1 * q);
 
+            if (result < -Math.Abs(arg1) || result > Math.Abs(arg1))
+            {
+                result = 0;
+            }
```
CarlosFigueiraMSFT, line 621: Why this condition here? Can you add a comment?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -541,6 +648,12 @@ private static IEnumerable<double> LazySequence(double records, double start, do
         public static FormulaValue Abs(IRContext irContext, NumberValue[] args)
         {
             var arg0 = args[0];
+            
+            if (arg0 == null) 
+            {
+                return new BlankValue(irContext);
+            }
+
```
CarlosFigueiraMSFT, line 656: `Abs(Blank)` should return 0 (blank values as input to functions in numeric context should be coerced to 0)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -552,92 +665,67 @@ public static FormulaValue Round(IRContext irContext, NumberValue[] args)
...
+            if (x == double.NaN)
+            {
+                return CommonErrors.NumericOutOfRange(irContext);
+            }
+
             return new NumberValue(irContext, x);
         }
 
-        public static double Round(double number, double digits)
+        public static double Round(double number, double digits, RoundType rt = RoundType.Default)
```
CarlosFigueiraMSFT, line 676: >public [](http://example.com/codeflow?start=8&length=6)

I know this was already there... but does it need to be public?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
LucGenetier: When we have extremely large number (1e308) the calculated reminder could be completely wrong,hence this extra validation

I can add a comment if you want to explain this

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -552,92 +665,67 @@ public static FormulaValue Round(IRContext irContext, NumberValue[] args)
             var digitsArg = args[1].Value;
 
             var x = Round(numberArg, digitsArg);
+            if (x == double.NaN)
+            {
+                return CommonErrors.NumericOutOfRange(irContext);
+            }
+
```
CarlosFigueiraMSFT, line 672: We should be able to have the numeric boundary checking for all functions in the 'StandardErrorHandling' wrapper, so it doesn't need to be done for each individual function

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
LucGenetier: Correct

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -355,34 +430,56 @@ public static FormulaValue Mid(IRContext irContext, FormulaValue[] args)
...
-            if (count.Value == 0)
+        private static FormulaValue LeftOrRight(IRContext irContext, FormulaValue[] args, Func<string, int, string> leftOrRight)
+        {
+            if (args[0] is BlankValue || args[1] is BlankValue)
             {
                 return new StringValue(irContext, string.Empty);
             }
 
-            if (count.Value >= source.Value.Length)
+            if (args[1] is not NumberValue)
```
CarlosFigueiraMSFT, line 468: >args[1] is not NumberValue [](http://example.com/codeflow?start=16&length=26)

`if (args[1] is not NumberValue count)` - and you won't need the declaration/cast on line 474

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -25,13 +25,18 @@ private interface IAggregator
         {
             void Apply(FormulaValue value);
 
+            FormulaValue GetDefault(IRContext context);
+
+            FormulaValue NoElementValue(IRContext context);
```
MikeStall, line 30: Do these need to be on the interface? Does somebody ever call these directly or is it always used via GetResult()

We want to keep interfaces as small as possible. (easy to implement).

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -160,91 +204,118 @@ public FormulaValue GetResult(IRContext irContext)
         private class MinDateAgg : IAggregator
         {
             protected DateTime _minValueDT = DateTime.MaxValue;
+            protected static readonly DateTime _defaultDT = new DateTime(1900, 1, 1);
 
             public void Apply(FormulaValue value)
-            {
-                if (value is BlankValue)
-                {
```
MikeStall, line 167: >new DateTime(1900, 1, 1) [](http://example.com/codeflow?start=60&length=24)

This literal should only appear once. It could be a static at the top of this class,  or even a internal static on DateType

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -303,31 +388,48 @@ public FormulaValue GetResult(IRContext irContext)
...
+
+            foreach (var value in vals)
             {
                 agg.Apply(value);
             }
 
             return agg.GetResult(irContext);
         }
 
-        private static async Task<FormulaValue> RunAggregatorAsync(IAggregator agg, EvalVisitor runner, SymbolContext context, IRContext irContext, FormulaValue[] args)
```
MikeStall, line 327: Do we need this? 
Or do we get this naturally from just calling getResult() with 0 items?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
LucGenetier: GetDefault isn't called, I'll remove this.
NoElementValue is called in private static FormulaValue RunAggregator(IAggregator agg, IRContext irContext, FormulaValue[] values)

Ideally, these aggregators should be rewritten using generics to avoid so much duplicated code

public class Aggregator<F, T> : IAggregator<T>
    where F : IAggregatorFunction<T>, new()
    where T : new()

public interface IAggregator<T>
{
    void AddValue(T element);
    T Aggregate();
}

public interface IAggregatorFunction<T>
{
    T Aggregate(List<T> elements);
}

public class Sum<T> : IAggregatorFunction<T>
    where T : ISum<T>, new()

...

LucGenetier:
> > ```
> >             Kind = ErrorKind.BadLanguageCode
> > ```
> 
> Was wondering the same and will check later when I'll have made more progress in fixing all functions
> 
> In reply to: [1108729432](https://github.com/microsoft/Power-Fx/pull/342#issuecomment-1108729432)
> 
> Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs:68 in [39804d2](https://github.com/microsoft/Power-Fx/commit/39804d26701fade9ce4dc830bc93b5a278e37b2a). [](commit_id = [39804d2](https://github.com/microsoft/Power-Fx/commit/39804d26701fade9ce4dc830bc93b5a278e37b2a), deletion_comment = False)

This error is only generated when we call bool.TryParse, so we're good with this one I think.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
LucGenetier: You're right we don't need this.
I've simplified the code here

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
LucGenetier: Correct, fixed

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
LucGenetier: Good catch!

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
LucGenetier: Fixed

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
LucGenetier: I moved this to FiniteChecker method

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -324,8 +424,8 @@ private static FormulaValue RunAggregator(IAggregator agg, IRContext irContext,
 
         private static async Task<FormulaValue> RunAggregatorAsync(IAggregator agg, EvalVisitor runner, SymbolContext context, IRContext irContext, FormulaValue[] args)
         {
-            var arg0 = (TableValue)args[0];
-            var arg1 = (LambdaFormulaValue)args[1];
+            var arg0 = (TableValue)args.Where(a => a is not BlankValue).First();
```
MikeStall, line 427: >(a => a is not BlankValue) [](http://example.com/codeflow?start=45&length=26)

?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -92,47 +98,116 @@ public static async ValueTask<FormulaValue> Value(EvalVisitor runner, SymbolCont
...
+
+                case ConvertionStatus.InvalidNumber:
+                    return CommonErrors.ArgumentOutOfRange(irContext);
+
+                case ConvertionStatus.Ok:
+                    return new NumberValue(irContext, val);
             }
 
-            if (IsInvalidDouble(val))
+            throw new NotImplementedException("Unknown ConvertionStatus value");
```
MikeStall, line 131: >throw new NotImplementedException("Unknown ConvertionStatus value"); [](http://example.com/codeflow?start=12&length=68)

nit:  good to include the value, like
$"Unknown ConvertionStatus value {err}"

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -92,47 +98,116 @@ public static async ValueTask<FormulaValue> Value(EvalVisitor runner, SymbolCont
...
+        }
+
+        private static (double, ConvertionStatus) ConvertToNumber(string str, CultureInfo culture)
+        {
+            double div = 1;
+            var loop = true;
+            var signFound = false;
+            var percentFound = false;
+
+            while (loop)
```
MikeStall, line 141: >while (loop) [](http://example.com/codeflow?start=12&length=12)

can we  avoid the while loop here?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using Microsoft.PowerFx.Core.Functions.Delegation.DelegationStrategies;
```
MikeStall, line 8: >DelegationStrategies [](http://example.com/codeflow?start=50&length=20)

curious why we need this namespace?

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
```diff
@@ -62,8 +62,4 @@
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
   </ItemGroup>
-
-  <ItemGroup>
-    <Folder Include="ExpressionTestCases\NotYetReady\" />
-  </ItemGroup>
```
MikeStall, line 68: revert?

MikeStall (dismissed):
:shipit:

MikeStall (approved):

MikeStall (approved):

PR #353, opened 2022-04-27
Allow interpreter to work on any IExternalOptionSet instead of OptionSet

MikeStall:
Allow interpreter to work on any IExternalOptionSet  implementation (not just OptionSet) so that it can execute on optionsets coming from dataverse. 

Also add more helpers on  OptionSetValueType and OptionSetValue (which live in Core) to make them more usable. This lets them work on dataverse option sets too.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalOptionSet.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalOptionSet.cs
@@ -2,11 +2,15 @@
 // Licensed under the MIT license.
 
 using System.Collections.Generic;
+using Microsoft.PowerFx.Core.Public.Values;
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
 
 namespace Microsoft.PowerFx.Core.Entities
 {
+    /// <summary>
+    /// Describe an option set - maybe implemented by each back end over their existing enum-like symbols. 
+    /// </summary>
     internal interface IExternalOptionSet : IExternalEntity
     {
         DisplayNameProvider DisplayNameProvider { get; }
@@ -18,6 +22,8 @@ namespace Microsoft.PowerFx.Core.Entities
                 
         bool IsBooleanValued { get; }
 
-        bool IsConvertingDisplayNameMapping { get; } 
+        bool IsConvertingDisplayNameMapping { get; }
+
+        bool TryGetValue(DName fieldName, out OptionSetValue optionSetValue);
     }
 }
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using Microsoft.PowerFx.Core.Entities;
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
 
@@ -57,9 +58,47 @@ namespace Microsoft.PowerFx.Core.Public.Types
             _type = type;
         }
 
+        // Entites may be recursive and their Dytype is tagged with additional schema metadata. 
+        // Expand that metadata into a proper Dtype. 
+        private static DType GetExpandedEntityType(DType expandEntityType, string relatedEntityPath)
+        {
+            Contracts.AssertValid(expandEntityType);
+            Contracts.Assert(expandEntityType.HasExpandInfo);
+            Contracts.AssertValue(relatedEntityPath);
+
+            var expandEntityInfo = expandEntityType.ExpandInfo;
+
+            if (expandEntityInfo.ParentDataSource is not IExternalTabularDataSource dsInfo)
+            {
+                return expandEntityType;
+            }
+
+            DType type;
+
+            if (!expandEntityType.TryGetEntityDelegationMetadata(out var metadata))
+            {
+                // We need more metadata to bind this fully
+                return DType.Error;
+            }
+
+            type = expandEntityType.ExpandEntityType(metadata.Schema, metadata.Schema.AssociatedDataSources);
+            Contracts.Assert(type.HasExpandInfo);
+
+            // Update the datasource and relatedEntity path.
+            type.ExpandInfo.UpdateEntityInfo(expandEntityInfo.ParentDataSource, relatedEntityPath);
+
+            return type;
+        }
+
         // Get the correct derived type
         internal static FormulaType Build(DType type)
         {
+            if (type.IsExpandEntity)
+            {
+                var expandedType = GetExpandedEntityType(type, string.Empty);
+                return Build(expandedType);
+            }
+
             switch (type.Kind)
             {
                 case DKind.ObjNull: return Blank;
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Types/OptionSetValueType.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Types/OptionSetValueType.cs
@@ -1,12 +1,19 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
 using Microsoft.PowerFx.Core.Entities;
+using Microsoft.PowerFx.Core.Public.Values;
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
 
 namespace Microsoft.PowerFx.Core.Public.Types
 {
+    /// <summary>
+    /// Power Fx type for an enum-like things such as OptionSets. 
+    /// </summary>
     public class OptionSetValueType : FormulaType
     {
         /// <summary>
@@ -33,7 +40,30 @@ namespace Microsoft.PowerFx.Core.Public.Types
             vistor.Visit(this);
         }
 
-        // $$$ Add this:
-        // public static bool TryGetValue(this OptionSetValueType type, string logicalName, out OptionSetValue osValue)
+        /// <summary>
+        /// List of logical names within this option set. 
+        /// </summary>
+        public IEnumerable<DName> LogicalNames => _type.OptionSetInfo.OptionNames; 
+
+        /// <summary>
+        /// Try to get a value given the logical name. 
+        /// </summary>
+        /// <param name="logicalName"></param>
+        /// <param name="osValue"></param>
+        /// <returns>False if logical name is not in the option set. </returns>
+        public bool TryGetValue(string logicalName, out OptionSetValue osValue)
+        {
+            var info = _type.OptionSetInfo;
+
+            // Verify this value exists in the option set. 
+            if (info.DisplayNameProvider.TryGetDisplayName(new DName(logicalName), out var displayName))
+            {
+                osValue = new OptionSetValue(logicalName, this);
+                return true;
+            }
+
+            osValue = null;
+            return false;
+        }
     }
 }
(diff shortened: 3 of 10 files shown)


src/libraries/Microsoft.PowerFx.Core/Public/Types/OptionSetValueType.cs
```diff
@@ -1,12 +1,19 @@
...
+using Microsoft.PowerFx.Core.Public.Values;
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
 
 namespace Microsoft.PowerFx.Core.Public.Types
 {
+    /// <summary>
+    /// Power Fx type for an enum-like things such as OptionSets. 
+    /// </summary>
     public class OptionSetValueType : FormulaType
```
MikeStall, line 17: >OptionSetValueType [](http://example.com/codeflow?start=17&length=18)

This (lives in Core) is very similar to OptionSet (lives in interpreter). 


OptionSet is just an impl of IExternalOptionSet which is used to instantiate a OSVT.  OS even gives you a OSVT - but OSVT can't give you back the OS (at best, it just has the IExternalOptionSet interface, which is buried in the DType). 


So it's more valuable to have useful methods (like TryGetValue) here than in OS.  Here, they're available in Core and work with dataverse and other impls.

src/libraries/Microsoft.PowerFx.Core/Public/Values/OptionSetValue.cs
```diff
@@ -6,36 +6,65 @@
...
         {
             Option = option;
         }
 
+        public new OptionSetValueType Type => (OptionSetValueType)base.Type;
+
         public override object ToObject()
         {
-            return Option;
+            return DisplayName;
```
MikeStall, line 35: >DisplayName [](http://example.com/codeflow?start=19&length=11)

This is a change. I think should be ok?

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -492,7 +493,7 @@ public override async ValueTask<FormulaValue> Visit(ResolvedObjectNode node, Sym
             return node.Value switch
             {
                 RecalcFormulaInfo fi => ResolvedObjectHelpers.RecalcFormulaInfo(fi),
-                OptionSet optionSet => ResolvedObjectHelpers.OptionSet(optionSet, node.IRContext),
+                IExternalOptionSet optionSet => ResolvedObjectHelpers.OptionSet(optionSet, node.IRContext),
```
MikeStall, line 496: >IExternalOptionSet [](http://example.com/codeflow?start=16&length=18)

Key change to enable it for interpreters!

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -85,6 +85,14 @@ protected override IExpression CreateEvaluator(CheckResult result)
             return new ParsedExpression(irnode, ruleScopeSymbol);
         }
 
+        public static IExpression CreateEvaluator2(CheckResult result)
```
MikeStall, line 88: >CreateEvaluator2 [](http://example.com/codeflow?start=34&length=16)

Ignore... didn't mean to check in. 
However , this does demonstrate that we can statically create an interpreter over a CheckResult. We don't actually need the recalc engine instance - all the state is captured in the IR by now.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
```diff
@@ -403,21 +403,8 @@ public static DateTimeValue TimeToDateTime(IRContext irContext, TimeValue[] args
 
         public static FormulaValue OptionSetValueToString(IRContext irContext, OptionSetValue[] args)
         {
-            // The type checker and IR have already validated that this is a valid OptionSet and that it contains a member matching this Option.
-            // These are just defensive error checks, and should be unreachable.
-            var os = args[0].Type?._type?.OptionSetInfo;
-            if (os is not OptionSet optionSet) 
-            {
-                return CommonErrors.UnreachableCodeError(irContext);
-            }
```
MikeStall, line 412: this case goes away now that we're strong typing. 
We didn't always have a OptionSet, but we *do* always have a IExternalOptionSet

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
```diff
@@ -403,21 +403,8 @@ public static DateTimeValue TimeToDateTime(IRContext irContext, TimeValue[] args
...
-
-            var option = args[0].Option;
-
-            if (!optionSet.Options.TryGetValue(new DName(option), out var displayName))
-            { 
-                return CommonErrors.UnreachableCodeError(irContext);
```
MikeStall, line 418: >CommonErrors.UnreachableCodeError(irContext); [](http://example.com/codeflow?start=23&length=45)

Just made this return logical name when display name is missing.  rather than error.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/ResolvedObjectHelpers.cs
```diff
@@ -20,19 +21,24 @@ public static FormulaValue RecalcFormulaInfo(RecalcFormulaInfo fi)
             return fi._value;
         }
 
-        public static FormulaValue OptionSet(OptionSet optionSet, IRContext irContext)
+        // $$$ - this allows behavior we really don't want ... can we block in the IR?
+        // OptionSets can act like Records: 
+        //   TimeUnit.Hours
+        //   If(true, TimeUnit).Hours
+        //   If(true, TextPosition, Align).Left
```
MikeStall, line 28: Once we block these cases, then we can eliminate this function altogether. 

Worst case, if we did need to have a way to OptionSet to RecordValue, we could have a derived RV that just wraps optionSet with virtual GetField() . Rather than fully hydrating the record with all fields.

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
```diff
@@ -57,9 +58,47 @@ internal FormulaType(DType type)
             _type = type;
         }
 
+        // Entites may be recursive and their Dytype is tagged with additional schema metadata.
```
lesaltzm, line 61: sp: dytype

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
```diff
@@ -57,9 +58,47 @@ internal FormulaType(DType type)
             _type = type;
         }
 
+        // Entites may be recursive and their Dytype is tagged with additional schema metadata. 
+        // Expand that metadata into a proper Dtype. 
+        private static DType GetExpandedEntityType(DType expandEntityType, string relatedEntityPath)
```
lesaltzm, line 63: Can relationship support be a separate PR?

src/tests/Microsoft.PowerFx.Interpreter.Tests/DisplayNameOptionSetTests.cs
```diff
@@ -47,6 +48,42 @@ public void OptionSetDisplayNames(string inputExpression, string outputExpressio
             }
         }
 
+        // Verify methods to go between OptionSet/Type/Value
+        [Fact]
+        public void TestHelpers()
```
MikeStall, line 53: >TestHelpers [](http://example.com/codeflow?start=20&length=11)

We could move this to Core if we moved OptionSet....
(or the test could add its own IExternalOptionSet impl)

src/libraries/Microsoft.PowerFx.Core/Public/Values/OptionSetValue.cs
```diff
@@ -6,36 +6,65 @@
...
+    /// A value within an option set. 
+    /// </summary>
+    [DebuggerDisplay("{ToString})")]
     public class OptionSetValue : FormulaValue
     {
         /// <summary>
         /// Logical name for this option set value.
         /// </summary>
-        public readonly string Option;
+        public string Option { get; private set; }
```
lesaltzm, line 23: Why not readonly anymore?

MikeStall: field --> property. 
Still readonly semantics.

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
MikeStall: ok.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/ResolvedObjectHelpers.cs
```diff
@@ -20,19 +21,24 @@ public static FormulaValue RecalcFormulaInfo(RecalcFormulaInfo fi)
             return fi._value;
         }
 
-        public static FormulaValue OptionSet(OptionSet optionSet, IRContext irContext)
+        // $$$ - this allows behavior we really don't want ... can we block in the IR?
+        // OptionSets can act like Records: 
+        //   TimeUnit.Hours
```
lesaltzm, line 26: Yeah. We don't enforce this for Option Sets, or legacy enums like TimeUnit/TextPosition/Align. 

For option sets, there's a use case in canvas: The function Choices(OptionSet) results in a list of all the members of the option set, for use in dropdowns and similar.  However, I wonder if we could treat that as a special case as opposed to allowing all these other things as a result of enabling it. 

For legacy enums there's no valid use case, and it seems buggy. We should add a feature flag and block it in binder while we discuss what to do with Canvas.

src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalOptionSet.cs
```diff
@@ -18,6 +22,8 @@ internal interface IExternalOptionSet : IExternalEntity
                 
         bool IsBooleanValued { get; }
 
-        bool IsConvertingDisplayNameMapping { get; } 
+        bool IsConvertingDisplayNameMapping { get; }
+
+        bool TryGetValue(DName fieldName, out OptionSetValue optionSetValue);
```
lesaltzm, line 27: tag PR as Breaking Internals?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/ResolvedObjectHelpers.cs
MikeStall: Choices() is a good example - we'll want that in interpreter eventually. But it could be special cased. Ie, Choices impl could get the IExternalOptionSet and operate on that. We don't need to hydrate to a RecordValue.

 So I'll update comment here, but won't make any tangible changes.

src/libraries/Microsoft.PowerFx.Core/Public/Values/OptionSetValue.cs
lesaltzm: Right, sorry. I meant, why the private set? I don't see that used anywhere.

src/libraries/Microsoft.PowerFx.Core/Public/Values/OptionSetValue.cs
```diff
@@ -6,36 +6,65 @@
...
+using Microsoft.PowerFx.Core.Utils;
 
 namespace Microsoft.PowerFx.Core.Public.Values
 {
-    [DebuggerDisplay("OptionSetValue ({Option})")]
+    /// <summary>
+    /// A value within an option set. 
+    /// </summary>
+    [DebuggerDisplay("{ToString})")]
     public class OptionSetValue : FormulaValue
```
lesaltzm, line 18: sealed?

lesaltzm (approved):

src/libraries/Microsoft.PowerFx.Core/Public/Values/OptionSetValue.cs
MikeStall: it's for the ctor on line 28.

src/libraries/Microsoft.PowerFx.Core/Public/Values/OptionSetValue.cs
MikeStall: I can do in a followup...

Issue #354, opened 2022-04-28
Parse (a - b) as sub BinaryOp instead of add+negate


lesaltzm:
Sub is currently represented by add + negate. This is unnecessary confusion, we should instead add a sub binary op.

anderson-joyle:
I can take this one if you haven't worked on it yet. Please confirm.

MikeStall:
At the parse level, 
(A- B) is parsed as `Add(A, Minus(b))`, rather than `Sub(A,B)`

At the parser level: 
- Parse Tree's BinaryOp has a 'Add' but no 'Sub'
https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Core/Lexer/BinaryOp.cs
- And Unary op has the minus: https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Core/Lexer/UnaryOp.cs 


The parser tree then goes through Binding, and then converted into an "execution tree" (called the IR Tree). 
See IRTranslator.cs.  
The IR's binary operators also just have Add, but no sub:

https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Core/IR/Nodes/BinaryOpKind.cs

Note that Syntax Binary op just has a single add operator because it maps to the + symbol. But the IR has multiple Add operators because now we know the types and can determine what kind of add it is.

MikeStall:
Once this is fixed, then we can clean up this spot:
https://github.com/microsoft/Power-Fx/blob/4b8a3e1e7bbe48f7e5870367cebbaaea10d633f7/src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs#L756

Introduced in #356.

anderson-joyle:
Thanks for taking the time to explain.

PR #356, opened 2022-04-28
DateDifference in IR shouldn't use negate

### LINKED ISSUE #355: Fix DateDifference IR to remove negate unary op

lesaltzm:
resolves #355 

Also fixes DateDifference implementation in interpreter to return a number of days.


ian-legler (approved):

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -756,9 +756,14 @@ private static IntermediateNode GetAddBinaryOp(IRTranslatorContext context, Texl
                             // Date + '-DateTime' => in days
                             // Date + '-Date' => in days
 
-                            // Ensure that this is really '-Date' - Binding should always catch this, but let's make sure...
-                            Contracts.Assert(node.Right.AsUnaryOpLit().VerifyValue().Op == UnaryOp.Minus);
-                            return new BinaryOpNode(context.GetIRContext(node), BinaryOpKind.DateDifference, left, right);
+                            // Binding produces this as '-Date'. This should be cleaned up when we switch to a proper sub op.
```
MikeStall, line 759: >cleaned up when we switch to a proper sub op.  [](http://example.com/codeflow?start=80&length=46)

Tagging [https://github.com/microsoft/Power-Fx/issues/354](https://github.com/microsoft/Power-Fx/issues/354) here.

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -756,9 +756,14 @@ private static IntermediateNode GetAddBinaryOp(IRTranslatorContext context, Texl
                             // Date + '-DateTime' => in days
                             // Date + '-Date' => in days
 
-                            // Ensure that this is really '-Date' - Binding should always catch this, but let's make sure...
-                            Contracts.Assert(node.Right.AsUnaryOpLit().VerifyValue().Op == UnaryOp.Minus);
-                            return new BinaryOpNode(context.GetIRContext(node), BinaryOpKind.DateDifference, left, right);
+                            // Binding produces this as '-Date'. This should be cleaned up when we switch to a proper sub op. 
+                            if (right is not UnaryOpNode unaryNode || unaryNode.Op != UnaryOpKind.Negate)
+                            {
+                                throw new NotSupportedException();
```
MikeStall, line 762: >throw new NotSupportedException(); [](http://example.com/codeflow?start=32&length=34)

Is there a legal expression that can land here?

Or will binding have prevented it and landing here is a compiler bug.

lesaltzm: Binding will have prevented this. It's always a bug if we land here.

lesaltzm: see 
image

PR #357, opened 2022-04-28
Implement Proper

DavidLuong98:
Implement Proper #284

I'm opened for some suggestions if it can be done differently or if there needs to be some adjustment to unit tests.


src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -313,6 +313,11 @@ public static FormulaValue Upper(IRContext irContext, StringValue[] args)
             return new StringValue(irContext, args[0].Value.ToUpper());
         }
 
+        public static FormulaValue Proper(IRContext irContext, StringValue[] args)
+        {
+            return new StringValue(irContext, CultureInfo.CurrentCulture.TextInfo.ToTitleCase(args[0].Value));
```
jcoc611-microsoft, line 318: @MikeStall should `IRContext` expose the parameterized CultureInfo?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Proper.txt
```diff
@@ -0,0 +1,14 @@
+>> Proper("e. e. cummings")
```
lesaltzm, line 1: strings with non ascii characters?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
lesaltzm, line 318: JS implementation was done with regex + using the current locale. Check in powerapps that this handles stuff like acronyms the same way. 
image

lesaltzm: It's available on EvalRunner. See Value() in LibraryText.cs as an example.

DavidLuong98: image
Not sure if i fully understand you. Did you mean something like this? Are we expecting "RSVP" to maintain as "RSPV" after the proper function? Let me know if i set this up wrong

DavidLuong98: I edited this and will resolve the comment. Let me know if you think this is not the correct way.

lesaltzm: I meant that the ToTitleCase function in C# would keep it as RSVP according to the docs https://docs.microsoft.com/en-us/dotnet/api/system.globalization.textinfo.totitlecase?view=net-6.0

So we have a difference in behavior here.

DavidLuong98: I made the change, let me know if you think this is ok

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Proper.txt
```diff
@@ -0,0 +1,38 @@
...
```
MikeStall, line 1: FYI - ran against JS tests and these all pass.

MikeStall (approved):
:shipit:

PR #360, opened 2022-04-29
Remove DedupedQueue from Repository

jack-work:
In the creation of this repository, utilities were moved unnecessarily.  Among those includes DedupedQueue, which is used exclusively by the code generator, the code for which is tracked by the internal PowerApps-Client repository.  A PR to that repo necessitated a change to DedupedQueue, and it was so moved to that repository.  This change removes the class from this repository.


lesaltzm (approved):

MikeStall (approved):
:shipit:

PR #361, opened 2022-04-29
Fixup namespace for publics. 

MikeStall:
Apply renames to publics.  See PublicSurfaceTest for principles:

            // The goal for public namespaces is to make the SDK easy for the consumer. 
            // Namespace principles for public classes:            // 
            // - prefer fewer namespaces. See C# for example: https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis
            // - For easy discovery, but Engine in "Microsoft.PowerFx".
            // - For sub areas with many related classes, cluster into a single subnamespace.
            // - Avoid nesting more than 1 level deep

This tries to just change using and namespaces - just update public surface area and minimize churn.


src/libraries/Microsoft.PowerFx.Core/Errors/ErrorSeverity.cs
```diff
@@ -0,0 +1,65 @@
...
+
+#pragma warning disable SA1300 // Element should begin with upper-case letter
+        _Min = Verbose,
+#pragma warning restore SA1300 // Element should begin with upper-case letter
+
+        /// <summary>
+        /// A suggestion about possible high-level improvements or refactoring that may help the user
+        /// get a better app experience.
+        /// Examples: performance changes
+        /// Verbose messages will not be part of 'ChoreUpdateRulesWithErrors' as they would be analysed when the dependency changes.
```
lesaltzm, line 22: Nits: Chore references don't belong in public facing doc comments (apply to only PAClient analysis)

lesaltzm (dismissed):

src/libraries/Microsoft.PowerFx.Core/Errors/ErrorSeverity.cs
MikeStall: removed.

lesaltzm (approved):

PR #362, opened 2022-04-29
Remove most enums from the default PowerFxConfig

bryancroteau-MSFT:
Remove most enums from the default enum store.

Also fixes a few tests in the LanguageServerTests.cs file, and makes them less fragile overall.


src/libraries/Microsoft.PowerFx.Interpreter/Environment/PowerFxConfigExtensions.cs
```diff
@@ -20,5 +20,15 @@ public static void AddFunction(this PowerFxConfig powerFxConfig, ReflectionFunct
...
+
+        public static void AddDefaultEnums(this PowerFxConfig powerFxConfig)
+        {
+            powerFxConfig.EnumStoreBuilder.WithDefaultEnums();
+        }
+
+        public static void AddRequiredEnums(this PowerFxConfig powerFxConfig)
+        {
+            powerFxConfig.EnumStoreBuilder.WithRequiredEnums(powerFxConfig.Functions);
+        }
```
MikeStall, line 32: This is in Interpreter.
Can these be moved to Core?

MikeStall:
>             var engine = new Engine(new PowerFxConfig());

Can we just change the definition here so that it's still setup the same way, and then not have any of the other changes in the file?

---
Refers to: src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/LanguageServerTests.cs:36 in ad45ed3. [](commit_id = ad45ed32f95cae6ca5758910f30b0f6eab3f4894, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Environment/PowerFxConfigExtensions.cs
```diff
@@ -20,5 +20,15 @@ public static void AddFunction(this PowerFxConfig powerFxConfig, ReflectionFunct
         {
             powerFxConfig.AddFunction(function.GetTexlFunction());
         }
+
+        public static void AddDefaultEnums(this PowerFxConfig powerFxConfig)
+        {
+            powerFxConfig.EnumStoreBuilder.WithDefaultEnums();
```
lesaltzm, line 26: What's the difference between required/default?

src/libraries/Microsoft.PowerFx.Interpreter/Environment/PowerFxConfigExtensions.cs
bryancroteau-MSFT: Moved

src/libraries/Microsoft.PowerFx.Interpreter/Environment/PowerFxConfigExtensions.cs
bryancroteau-MSFT: Default includes all defined enums. Required only pulls in the ones required for the current function list.

orcmid: I must say that's weird.  Generally, I would think that required means you must explicitly make an available choice; default is what you can get without making a choice if you're good with that.  

Is this a specialized meaning in the context of `powerFXConfig` ?

lesaltzm: So, in that case, Default is the set only used by PA-Client? That doesn't seem like something we need to expose publicly.

bryancroteau-MSFT: I have reduced the scope of this change to only add the required enums in the public constructor of the PowerFxConfig.

MikeStall (approved):
:shipit:

PR #366, opened 2022-05-02
Add other parameter types for custom functions.  

### LINKED ISSUE #69: additional data types for custom functions needed
only numerical valued custom functions with numrerical arguments are supported.  adding strings and bools would be great for starters.

many thanks for the important work being done here!

MikeStall:
Just support all types for PrimitiveValue<T>.

Fixes #69


LucGenetier (approved):
:shipit:

PR #368, opened 2022-05-03
Expose CheckInvocation

ivanradicek:
Exposes `CheckInvocation` to check whether an invocation of some function with arguments (as existing syntax nodes) is valid.


src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -389,13 +393,29 @@ public DType GetType(TexlNode node)
             return _typeMap[node.Id];
         }
 
+        public bool IsNodeValid(TexlNode node)
```
MikeStall, line 396: >public [](http://example.com/codeflow?start=8&length=6)

Nit: can we add a comment here explaining. 

// Ensure that this TexlNode is associated with this binding. 
// This is critical so that the node.Ids are valid.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -389,13 +393,29 @@ public DType GetType(TexlNode node)
...
+
         private void SetType(TexlNode node, DType type)
         {
             Contracts.AssertValue(node);
             Contracts.Assert(type.IsValid);
             Contracts.AssertIndex(node.Id, _typeMap.Length);
             Contracts.Assert(_typeMap[node.Id] == null || !_typeMap[node.Id].IsValid || type.IsError);
 
+            // TODO: Should we do this here? It's easy because it's called for each node, but then again it's not doing what it says is doing.
+            // TODO: Should we just rename the method?
```
MikeStall, line 417: I think this is ok.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -94,6 +95,52 @@ public void ThrowOnErrors()
...
+        /// Checks whether a call to a function with name <paramref name="fncName" /> is valid with argument list
+        /// <paramref name="args" />. Additionally returns (as an out parameter) the return type of this invocation.
+        /// 
+        /// Note: all arguments must belong to the formula that belongs to this <see cref="CheckResult" />.
+        /// </summary>
+        /// <param name="fncName"></param>
+        /// <param name="args"></param>
+        /// <param name="retType"></param>
+        /// <returns></returns>
+        public bool ValidateInvocation(string fncName, IReadOnlyList<TexlNode> args, out FormulaType retType)
```
MikeStall, line 108: >fncName [](http://example.com/codeflow?start=46&length=7)

functionName

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -94,6 +95,52 @@ public void ThrowOnErrors()
...
+        {
+            retType = null;
+
+            if (args is null)
```
MikeStall, line 112: >is [](http://example.com/codeflow?start=21&length=2)

nit: ==

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -94,6 +95,52 @@ public void ThrowOnErrors()
...
+            foreach (var arg in args)
+            {
+                if (arg is null)
+                {
+                    throw new ArgumentNullException(nameof(args));
+                }
+
+                if (!_binding.IsNodeValid(arg))
+                {
+                    throw new ArgumentException("Argument does not belong to this result");
```
MikeStall, line 126: >Argument [](http://example.com/codeflow?start=49&length=8)

nit: include arg index number to help debug

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -94,6 +95,52 @@ public void ThrowOnErrors()
...
+                }
+            }
+
+            var fnc = _binding.NameResolver.Functions
+                              .Where(fnc => fnc.Name == fncName && args.Count >= fnc.MinArity && args.Count <= fnc.MaxArity)
+                              .FirstOrDefault();
```
MikeStall, line 132: >.FirstOrDefault(); [](http://example.com/codeflow?start=30&length=18)

There could be multiple matches (like overloads like Sum function). We could just loop.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -94,6 +95,52 @@ public void ThrowOnErrors()
...
+                }
+
+                if (!_binding.IsNodeValid(arg))
+                {
+                    throw new ArgumentException("Argument does not belong to this result");
+                }
+            }
+
+            var fnc = _binding.NameResolver.Functions
+                              .Where(fnc => fnc.Name == fncName && args.Count >= fnc.MinArity && args.Count <= fnc.MaxArity)
```
MikeStall, line 131: >c.Name == fncName & [](http://example.com/codeflow?start=46&length=19)

namespace? 

Clock.AM

src/tests/Microsoft.PowerFx.Core.Tests/CheckInvocationTests.cs
```diff
@@ -0,0 +1,92 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+using Microsoft.PowerFx.Types;
+using Xunit;
+
+namespace Microsoft.PowerFx.Core.Tests
+{
+    public class CheckInvocationTests : PowerFxTest
```
MikeStall, line 10: >CheckInvocationTests [](http://example.com/codeflow?start=17&length=20)

Add tests for:

- overloads,  (is there  case where overloads return different types?)
- namespaces. 



If() is great test because return type is dynamic.

src/libraries/Microsoft.PowerFx.Core/Public/IntellisenseOperations.cs
```diff
@@ -0,0 +1,92 @@
...
+    /// </summary>
+    public class IntellisenseOperations
+    {
+        private readonly CheckResult _checkResult;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="IntellisenseOperations"/> class.
+        /// </summary>
+        /// <param name="result"></param>
+        public IntellisenseOperations(CheckResult result)
```
ivanradicek, line 24: Have no idea whether this is what we want?

src/libraries/Microsoft.PowerFx.Core/Public/IntellisenseOperations.cs
```diff
@@ -0,0 +1,92 @@
...
+
+                if (!_checkResult._binding.IsNodeValid(arg))
+                {
+                    throw new ArgumentException($"Argument {index} does not belong to this result");
+                }
+            }
+
+            var types = args.Select(node => _checkResult._binding.GetType(node)).ToArray();
+
+            // TODO: Horrible hack to get function identifier name - how to do this in idiomatic way?
```
ivanradicek, line 63: This is a nasty hack. How to do this properly?

MikeStall: Is this just to "parse" the functionName in case it contains a namespace? (essentially TexlParser.cs ParseInvocationWithNamespace )

src/libraries/Microsoft.PowerFx.Core/Public/IntellisenseOperations.cs
MikeStall: This is ok.

src/libraries/Microsoft.PowerFx.Core/Public/IntellisenseOperations.cs
ivanradicek: That works on the tokenized stuff, right? One (less hacky way) could be to tokenize the string and then construct an identifier manually, but this would still be hacky. The end goal is to convert dotted string to namespace and name (the opposite of `ToDottedSyntax()` of `DPath`).

MikeStall: Can we do this:

Keep line 64 as is, but move to its own dedicated function. 
That "quarantines" the hacky code, and gives us a place to document it and unit test. 

// Parse a function name string into a namespace and name. 
bool TryParseFunctionNameAndNamespace(string functionName, out Identifiier fncIdent)
{
 fncIdent = TexlParser.ParseScript($"{functionName}()").Root.AsCall()?.Head;
return fncIdent != null;
}

And then add a unit test for it. 
"Name" 
"Namespace.Name" 
error

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/IntellisenseOperations.cs
ivanradicek: Done

MikeStall (approved):
:shipit:

PR #371, opened 2022-05-04
Updates to support Dataverse.

MikeStall:
1. Allow FormulaType.Build to handle entities. 
2. Consolidate primitive marshalling helpers into a single class. 
3. Rename BuiltinFormulaTypeConversions --> PrimitiveValueConversions to be more consistent with PrimitiveValue<T>
Interpreter still has PrimitiveTypeMarshaller, but the converter code now lives in core since other scenarios need it. 
SQL runner can use this. 

4. Move tests into core, Add Tests. 
5. Allow calling interpreter directly on a binding result. This allows other resolvers to still run against the interpreter.


LucGenetier (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
```diff
@@ -62,9 +62,47 @@ internal FormulaType(DType type)
...
+            Contracts.AssertValue(relatedEntityPath);
+
+            var expandEntityInfo = expandEntityType.ExpandInfo;
+
+            if (expandEntityInfo.ParentDataSource is not IExternalTabularDataSource dsInfo)
+            {
+                return expandEntityType;
+            }
+
+            DType type;
```
bryancroteau-MSFT, line 80: Why is this being declared by itself?

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -65,6 +65,23 @@ private protected override SimpleResolver CreateResolver(PowerFxConfig alternate
...
         {
+            return CreateEvaluatorDirect(result);
+        }
+
+        /// <summary>
+        /// Create an evaluator over the existing binding. 
+        /// </summary>
+        /// <param name="result">A successful binding from a previous call to <see cref="Engine.Check(string, RecordType, ParserOptions)"/>. </param>
+        /// <returns></returns>
+        public static IExpression CreateEvaluatorDirect(CheckResult result)
```
bryancroteau-MSFT, line 76: Why is this public?

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
MikeStall: will fix.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
MikeStall: So that other Engines (notably dataverse) can run the interpreter on *their* IRs. 
The fundamental reason things need to derive from Engine is to override *the resolver*,  not to interpret.

Ie, there's nothing actually inherently coupled between the RecalcEngine and the interpreter.  (ditto for the other parts of the pipeline like parser, binder, etc)

bryancroteau-MSFT (dismissed):

bryancroteau-MSFT (approved):

PR #372, opened 2022-05-05
Move Span to the Microsoft.PowerFx.Syntax namespace.

MikeStall:
Move Span from : 
`Microsoft.PowerFx.Core.Localization.Span` to `Microsoft.PowerFx.Syntax.Span`.
This is part of a set of namespace changes. see #361 for details.  
This is part of the public  surface area and many products will depend on it. 

This is primarily provided by Syntax nodes, as well as for error ranges. 
This is a public class.
It also has [TransportAttribute] on it.


CarlosFigueiraMSFT:
This may have some implications for Canvas apps - which use spans in the client-side code as well. We should be careful when merging this change to that repository.

PR #375, opened 2022-05-06
Change SuggestUnqualifiedEnums to true

LucGenetier:


src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseData/IntellisenseData.cs
```diff
@@ -163,7 +163,7 @@ internal virtual bool DoesNameCollide(string name)
         /// <summary>
         /// Should unqualified enums be suggested.
         /// </summary>
-        internal virtual bool SuggestUnqualifiedEnums => true;
+        internal virtual bool SuggestUnqualifiedEnums => false;
```
MikeStall, line 166: There may need to be some flag here just to ensure that we don't break Canvas Apps.

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseFlags.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// <summary>
+        /// Default bahavior
+        /// </summary>
+        Default = 0x0,
+
+        /// <summary>
+        /// Suggests unqualified enums when possible
+        /// </summary>
+        SuggestUnqualifiedEnums = 0x1,
+    }
```
MikeStall, line 23: Can this by tied to the resolver?
Ie, somehow in PowerFx standalone, we already make a decision to require qualified enums ("TimeUnit.Hours"). 
(Whereas canvas apps still allows unqualified enums ("Hours"). 

Can we leverage that mechanism and avoid adding flags?

src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterSuggestTests.cs
```diff
@@ -61,5 +61,15 @@ public void TestSuggestOptionSets(string expression, params string[] expectedSug
             var actualSuggestions = SuggestStrings(expression, config, parameterType);
             Assert.Equal(expectedSuggestions, actualSuggestions);
         }
+
+        [Theory]
+        [InlineData("Hou|", "Hour", "TimeUnit.Hours")]
+        public void TestSuggestHour(string expression, params string[] expectedSuggestions)
```
MikeStall, line 67: >TestSuggestHour [](http://example.com/codeflow?start=20&length=15)

Great test!

src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs
```diff
@@ -42,20 +42,24 @@ internal class SimpleResolver : INameResolver
 
         IExternalEntity INameResolver.CurrentEntity => null;
 
+        public bool SuggestUnqualifiedEnums { get; private set; }
```
MikeStall, line 45: >public bool SuggestUnqualifiedEnums { get; private set; } [](http://example.com/codeflow?start=8&length=57)

This can just be fixed: 

public bool SuggestUnqualifiedEnums => true;

MikeStall: Canvas uses a different resolver instance.

src/libraries/Microsoft.PowerFx.Core/Binding/INameResolver.cs
```diff
@@ -36,6 +36,8 @@ internal interface INameResolver
 
         IEnumerable<TexlFunction> Functions { get; }
 
+        bool SuggestUnqualifiedEnums { get; }
```
MikeStall, line 39: >bool [](http://example.com/codeflow?start=8&length=4)

We should add a comment like:

This advertises whether the INameResolver instance will suggest unqualified enums ("Hours") or only qualified enums ("TimeUnit.Hours"). This must be consistent with how the other Lookup functions behave. 

Intellisense can use this when suggesting completion options.

src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs
LucGenetier: Right!

src/libraries/Microsoft.PowerFx.Core/Binding/INameResolver.cs
LucGenetier: Done

src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterSuggestTests.cs
LucGenetier: Thanks!

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseData/IntellisenseData.cs
```diff
@@ -51,6 +50,7 @@ public IntellisenseData(PowerFxConfig powerFxConfig, IEnumStore enumStore, IInte
             MissingTypes = missingTypes;
             BoundTo = string.Empty;
             CleanupHandlers = new List<ISpecialCaseHandler>();
+            SuggestUnqualifiedEnums = Binding.NameResolver.SuggestUnqualifiedEnums;
```
MikeStall, line 53: >SuggestUnqualifiedEnums [](http://example.com/codeflow?start=12&length=23)

This is the key fix.

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseData/IntellisenseData.cs
MikeStall: nit, can this be a 'private set'? Ie, we want to treat this as a readonly property and not let others set it. 

Or just => 
Binding.NameResolver.SuggestUnqualifiedEnums;

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseData/IntellisenseData.cs
LucGenetier: Yep. Took me a few seconds to get it ;)

MikeStall (approved):
:shipit:

PR #380, opened 2022-05-09
Parse (a - b) as sub BinaryOp.

anderson-joyle:
Issue #354. Please review. In theory there is nothing else missing. Sending as draft just to make sure I covered the current functions state.


src/tests/Microsoft.PowerFx.Core.Tests/ParseTests.cs
```diff
@@ -113,8 +113,8 @@ public void TexlParseStringLiteralsWithEscapableCharacters(string script)
         [InlineData("1 * 2 + 3 * 4")]
         [InlineData("1 * 2 * 3 + 4 * 5")]
         [InlineData("1 * 2 * 3 * 4 * 5")]
-        [InlineData("2 - 1", "2 + -1")]
-        [InlineData("2 - 1 - 2 - 3 - 4", "2 + -1 + -2 + -3 + -4")]
+        [InlineData("2 - 1", "2 - 1")]
```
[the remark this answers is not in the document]
anderson-joyle: Thanks for reviewing this, @MikeStall. In general, I would not remove tests. In this case, we are comparing an input expression string against its PrettyPrint string.
Since this fix is meant to replace the "add+negate" approach, this particular test case would fail.
Let me know if it makes sense.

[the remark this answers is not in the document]
MikeStall: Good point.  I see you're just changing the expected value. 
It would still be good to add these input cases:

2+ -1
-1

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryOperators.cs
```diff
@@ -35,6 +35,14 @@ internal static partial class Library
             returnBehavior: ReturnBehavior.AlwaysEvaluateAndReturnResult,
             targetFunction: NumericDiv);
 
+        public static readonly AsyncFunctionPtr OperatorBinarySub = StandardErrorHandling<NumberValue>(
+            expandArguments: NoArgExpansion,
+            replaceBlankValues: ReplaceBlankWithZero,
+            checkRuntimeTypes: ExactValueType<NumberValue>,
+            checkRuntimeValues: FiniteChecker,
+            returnBehavior: ReturnBehavior.AlwaysEvaluateAndReturnResult,
+            targetFunction: NumericSub);
```
MikeStall, line 44: to reduce copy & paste, I wonder if this could be wrapped in a helper like StandardMath, and then it's just

public static readonly AsyncFunctionPtr OperatorBinarySub = StandardMath(NumericSub);


And other operators here (like Add) could share that too.

MikeStall:
>                             }

Can this be cleaned up now? 
See #356 

---
Refers to: src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs:763 in a8e65bc. [](commit_id = a8e65bc5f4d7fcb437945f514aa136bf38571063, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryOperators.cs
anderson-joyle: > to reduce copy & paste, I wonder if this could be wrapped in a helper like StandardMath, and then it's just
> 
> public static readonly AsyncFunctionPtr OperatorBinarySub = StandardMath(NumericSub);
> 
> And other operators here (like Add) could share that too.


Is this close of what you are looking for:

Microsoft.PowerFx.Interpreter\Functions\StandardErrorHandling.cs
```
private static AsyncFunctionPtr StandardMath<T>(
            Func<IRContext, T[], FormulaValue> targetFunction)
            where T : FormulaValue
        {
            return StandardErrorHandling<T>(NoArgExpansion, ReplaceBlankWithZero, ExactValueType<NumberValue>, FiniteChecker, ReturnBehavior.AlwaysEvaluateAndReturnResult, (runner, symbolContext, irContext, args) =>
            {
                var result = targetFunction(irContext, args);
                return new ValueTask<FormulaValue>(result);
            });
        }
```

Microsoft.PowerFx.Interpreter\Functions\LibraryOperators.cs

```
public static readonly AsyncFunctionPtr OperatorBinaryAdd = StandardMath<NumberValue>(
    targetFunction: NumericAdd);

public static readonly AsyncFunctionPtr OperatorBinaryMul = StandardMath<NumberValue>(
    targetFunction: NumericMul);

public static readonly AsyncFunctionPtr OperatorBinaryDiv = StandardMath<NumberValue>(
    targetFunction: NumericDiv);

public static readonly AsyncFunctionPtr OperatorBinarySub = StandardMath<NumberValue>(
    targetFunction: NumericSub);
```


All Sub functions now invert the right arg sign and call the respective Add function. 

```
private static FormulaValue SubDateAndTime(IRContext irContext, FormulaValue[] args)
{
    var args1 = (TimeValue)args[1];

    args[1] = new TimeValue(irContext, -args1.Value);

    return AddDateAndTime(irContext, args);
}
```

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ParseJson.txt
```diff
@@ -88,7 +88,7 @@ Blank()
 Errors: Error 0-14: Invalid argument type. Expecting one of the following: Number, Text, Boolean.|Error 17-31: Invalid argument type. Expecting one of the following: Number, Text, Boolean.
 
 >> ParseJSON("5") - ParseJSON("5")
-Errors: Error 17-31: Invalid argument type. Expecting one of the following: Number, Text, Boolean.|Error 0-14: Invalid argument type. Expecting one of the following: Number, Text, Boolean.
+Errors: Error 0-14: Invalid argument type. Expecting one of the following: Number, Text, Boolean.|Error 17-31: Invalid argument type. Expecting one of the following: Number, Text, Boolean.
```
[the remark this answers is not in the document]
lesaltzm: it looks like they didn't, just the ordering. Makes sense, since we'd visit the args slightly differently.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4188,6 +4191,120 @@ void ReportInvalidOperation()
                 }
             }
 
+            private void PostVisitBinaryOpNodeSubtraction(BinaryOpNode node)
```
lesaltzm, line 4194: Can much of this logic be shared with the addition path? It looks mostly the same, and not dependent on it being subtract (excpet for DateDiff cases)

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -823,6 +815,67 @@ private static IntermediateNode GetAddBinaryOp(IRTranslatorContext context, Texl
...
+                    case DKind.Time:
+                        if (rightType == DType.Time)
+                        {
+                            // Time - Time => in ms
+                            return new BinaryOpNode(context.GetIRContext(node), BinaryOpKind.SubNumbers, left, right);
+                        }
+                        else
+                        {
+                            // Time + Number
+                            return new BinaryOpNode(context.GetIRContext(node), BinaryOpKind.AddTimeAndMilliseconds, left, right);
```
lesaltzm, line 846: shouldn't one of these be negated, or we use a sub binary op kind?

anderson-joyle: There are new sub binary op kinds.

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -740,85 +741,128 @@ private IntermediateNode InjectCoercion(IntermediateNode child, IRTranslatorCont
...
+                Contracts.Assert(node.Op == BinaryOp.Add || node.Op == BinaryOp.Sub);
 
                 switch (leftType.Kind)
                 {
                     case DKind.Date:
                         if (rightType == DType.DateTime || rightType == DType.Date)
                         {
-                            // Date + '-DateTime' => in days
-                            // Date + '-Date' => in days
+                            Contracts.Assert(node.Op == BinaryOp.Sub);
```
lesaltzm, line 754: Consider changing these asserts to Check. Assert only runs in debug build, check runs in all builds.

src/libraries/Microsoft.PowerFx.Core/IR/Nodes/BinaryOpKind.cs
```diff
@@ -17,6 +17,11 @@ internal enum BinaryOpKind
         AddDateTimeAndDay,
         AddTimeAndMilliseconds, // Time + Number (Milliseconds)
 
+        SubNumbers,
+        SubDateAndDay,
+        SubDateTimeAndDay,
+        SubTimeAndMilliseconds, // Not yet implemented
```
lesaltzm, line 23: is this used?

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
anderson-joyle: Done.

src/libraries/Microsoft.PowerFx.Core/IR/Nodes/BinaryOpKind.cs
anderson-joyle: The first 3 yes. The SubTimeAndMilliseconds op kind has been removed.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/arithmetic.txt
```diff
@@ -30,6 +30,9 @@
 >> 5-3
 2
 
+>> 5+-3
+2
+
```
lesaltzm, line 35: A test validating precedence would also be good, something like 
1 - 2 * 3

anderson-joyle: Extra tests have been added.

PR #381, opened 2022-05-09
Remove RequiresErrorContext from TexlFunction

CarlosFigueiraMSFT:
The RequiresErrorContext property is only applicable to the JS runtime, so it doesn't need to be in the base TexlFunction class.
Also ignored some warnings for function definitions.


MikeStall (approved):
:shipit:

PR #385, opened 2022-05-10
Allow (internally) to specify custom resolver and glue, and expose some new formula types

ivanradicek:


src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
```diff
@@ -148,7 +151,7 @@ internal static FormulaType Build(DType type)
                     return BindingError;
 
                 default:
-                    throw new NotImplementedException($"Not implemented type: {type}");
+                    throw new NotImplementedException($"Not implemented type: {type} ({type.Kind}) ({type?.GetType().FullName})");
```
ivanradicek, line 154: TODO: revert this change. Was useful for debugging.

src/libraries/Microsoft.PowerFx.Core/Public/Types/ControlType.cs
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the MIT license.
+
+using Microsoft.PowerFx.Core.Types;
+
+namespace Microsoft.PowerFx.Types
+{
+    /// <summary>
+    /// Representing UI control type.
+    /// </summary>
+    public class ControlType : AggregateType
```
ivanradicek, line 11: > `: AggregateType`

This will probably not work, as it requires `GetAllNames()` which (I believe) will not work for a control DType.

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
```diff
@@ -148,7 +151,7 @@ internal static FormulaType Build(DType type)
                     return BindingError;
 
                 default:
-                    throw new NotImplementedException($"Not implemented type: {type}");
+                    return new UnsupportedType(type);
```
ivanradicek, line 154: Don't know how sane this is, but it makes it really convenient for what we're trying to do.

ivanradicek: Note that this type does not have a public way of constructing it.

src/libraries/Microsoft.PowerFx.Core/Public/Types/ControlType.cs
```diff
@@ -0,0 +1,24 @@
...
```
MikeStall, line 11: >ControlType [](http://example.com/codeflow?start=17&length=11)

Do we need to expose this, or can it fall under Unsupprted?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -50,11 +50,21 @@ public IEnumerable<string> GetAllFunctionNames()
         /// </summary>
         /// <param name="alternateConfig">An alternate config that can be provided. Should default to engine's config if null.</param>
         /// <returns></returns>
-        private protected virtual SimpleResolver CreateResolver(PowerFxConfig alternateConfig = null)
+        private protected virtual SimpleResolver CreateSimpleResolver(PowerFxConfig alternateConfig = null)
```
MikeStall, line 53: >CreateSimpleResolver [](http://example.com/codeflow?start=49&length=20)

can we avoid the dual chaining virtuals? Maybe just have CreateResolver() return INameResolver? 
I see there are a few places that ask for SimplerResolver, but they can probably just take INameResolver

ivanradicek: Yes, that worked out.

src/libraries/Microsoft.PowerFx.Core/Public/Types/ControlType.cs
ivanradicek: The implementation was dropped.

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
MikeStall: I'd approve one this change is reverted.

ivanradicek: Note that this is an outdated change. All of this is replaced with unsupported type construction.

lesaltzm:
>                 null,

reuse the glue function?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs:236 in 6b439cc. [](commit_id = 6b439cc33c3f2729b793b81e543a714f935f7cb5, deletion_comment = False)

lesaltzm (dismissed):
:shipit:

MikeStall:
>                 null,

CreateBinderGlue() ? 

---
In reply to: [1129286005](https://github.com/microsoft/Power-Fx/pull/385#issuecomment-1129286005) [](http://example.com/codeflow?ancestors=1129286005,1129286005)

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs:236 in 6b439cc. [](commit_id = 6b439cc33c3f2729b793b81e543a714f935f7cb5, deletion_comment = False)

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Types/UnsupportedType.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+        }
+
+        /// <inheritdoc />
+        public override void Visit(ITypeVistor visitor)
+        {
+            visitor.Visit(this);
+        }
+
+        internal string Description => $"kind={_type.Kind} fullTypeName={_type.GetType().FullName}";
```
MikeStall, line 24: >Description [](http://example.com/codeflow?start=24&length=11)

nit: is this just for debugging? If so, you could use [DebuggerDisplay]

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Types/UnsupportedType.cs
ivanradicek: Done, please do check.

MikeStall (approved):

PR #386, opened 2022-05-10
Set function 

MikeStall:
Allow interpreter to add Set() function which just maps to RecalcEngine.UpdateVariable.

- must be explicitly enabled by the host (since many hosts will not want this). 
- only works for firstname nodes.  
- requires that the variable was already declared. 
- type or arg1 must match exactly  to arg0.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
@@ -117,8 +117,8 @@ namespace Microsoft.PowerFx
             return val;
         }
 
-        // Handle the Set() function - which is unique because
-        // it has an l-value for the first arg. 
+        // Handle the Set() function -
+        // Set is unique because it has an l-value for the first arg. 
         // Async params can't have out-params. 
         // Return null if not handled. Else non-null if handled.
         private async Task<FormulaValue> TryHandleSet(CallNode node, SymbolContext context)
@@ -134,14 +134,14 @@ namespace Microsoft.PowerFx
 
             var newValue = await arg1.Accept(this, context);
 
-            // ensure this is a firstname node            
+            // Binder has already ensured this is a first name node. 
             if (arg0 is ResolvedObjectNode obj)
             {
-                // $$$ Should we have an interface here instead?
-                // Set(variable, newValue);   
-                if (obj.Value is RecalcFormulaInfo fi)
+                if (obj.Value is ICanSetValue fi)
                 {
-                    fi._value = newValue;
+                    fi.SetValue(newValue);
+
+                    // Set() returns true. 
                     return FormulaValue.New(true);
                 }
             }
@@ -529,7 +529,7 @@ namespace Microsoft.PowerFx
         {
             return node.Value switch
             {
-                RecalcFormulaInfo fi => ResolvedObjectHelpers.RecalcFormulaInfo(fi),
+                ICanGetValue fi => fi.Value,
                 IExternalOptionSet optionSet => ResolvedObjectHelpers.OptionSet(optionSet, node.IRContext),
                 _ => ResolvedObjectHelpers.ResolvedObjectError(node),
             };
--- a/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
@@ -12,6 +12,7 @@ using Microsoft.PowerFx.Core.IR;
 using Microsoft.PowerFx.Core.IR.Nodes;
 using Microsoft.PowerFx.Core.IR.Symbols;
 using Microsoft.PowerFx.Functions;
+using Microsoft.PowerFx.Interpreter;
 using Microsoft.PowerFx.Types;
 using static Microsoft.PowerFx.Functions.Library;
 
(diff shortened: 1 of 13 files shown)


src/tests/Microsoft.PowerFx.Interpreter.Tests/SetFunctionTests.cs
```diff
@@ -0,0 +1,76 @@
...
+using Microsoft.PowerFx.Types;
+using Xunit;
+
+namespace Microsoft.PowerFx.Interpreter.Tests
+{
+    public class SetFunctionTests : PowerFxTest
+    {
+        private readonly ParserOptions _opts = new ParserOptions { AllowsSideEffects = true };
+
+        // $$$ Trigger a recalc?
```
MikeStall, line 14: todo

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -116,9 +117,48 @@ public override async ValueTask<FormulaValue> Visit(LazyEvalNode node, SymbolCon
...
             CheckCancel();
+            
+            var setResult = await TryHandleSet(node, context);
+            if (setResult != null)
+            {
+                return setResult;
+            }            
 
             // Sum(  [1,2,3], Value * Value)
             // return base.PreVisit(node);
```
Shpakh-MSFT, line 164: Do we still need this comment?

src/libraries/Microsoft.PowerFx.Interpreter/RecalcFormulaInfo.cs
```diff
@@ -30,5 +30,22 @@ internal class RecalcFormulaInfo
 
         // Cached value
         public FormulaValue _value;
```
Shpakh-MSFT, line 32: >public [](http://example.com/codeflow?start=8&length=6)

Can this be private?

src/strings/PowerFxResources.en-US.resx
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
Shpakh-MSFT, line 1: >standalone [](http://example.com/codeflow?start=37&length=10)

Is this intentional change?

src/strings/PowerFxResources.en-US.resx
```diff
@@ -6436,4 +6436,16 @@
   <data name="ErrNamedFormula_AlreadyDefined" xml:space="preserve">
     <value>NamedFormula '{0}' already exists.</value>
   </data>
+  <data name="ErrNeedValidVariableName_Arg" xml:space="preserve">
+    <value>The first argument of '{0}' should be a valid variable name, and cannot conflict with any existing control, screen, collection, or data source names. Found type '{1}'</value>
```
Shpakh-MSFT, line 6440: >'{1}' [](http://example.com/codeflow?start=172&length=5)

This will not be a type but name that is used by the end-user. I think Found type part is not needed.

I would change the message it

Invalid variable name {1} used as a first argument of function {0}. It should be a valid variable name, and cannot conflict with any existing control, screen, collection, or data source names.

src/tests/Microsoft.PowerFx.Interpreter.Tests/SetFunctionTests.cs
```diff
@@ -0,0 +1,76 @@
...
+            var r1 = engine.Eval("x", null, _opts); // 12
+            Assert.Equal(12.0, r1.ToObject());
+
+            var r2 = engine.Eval("Set(x, 15)", null, _opts);
+
+            r1 = engine.Eval("x"); // 15
+            Assert.Equal(15.0, r1.ToObject());
+
+            r1 = engine.GetValue("x");
+            Assert.Equal(15.0, r1.ToObject());
```
Shpakh-MSFT, line 33: Any test for
engine.UpdateVariable("x", FormulaValue.New(12));
engine.UpdateVariable("y", FormulaValue.New(12));
Set(y, x *2)
y should be 24

Also probably a test where value is a async task.

src/tests/Microsoft.PowerFx.Interpreter.Tests/SetFunctionTests.cs
```diff
@@ -0,0 +1,76 @@
...
+            Assert.False(result.IsSuccess);
+
+            // Fails, arg0 is not a settable object. 
+            result = engine.Check("Set({y:x}.y, 20)", null, _opts);
+            Assert.False(result.IsSuccess);
+        }
+
+        // Set() can only be called if it's enabled.
+        [Fact]
+        public void TestSetVarFailureEnabled()
```
Shpakh-MSFT, line 65: >TestSetVarFailureEnabled [](http://example.com/codeflow?start=20&length=24)

Also need a test where Set exists but config AllowsSideEffects is to false.

Shpakh-MSFT (dismissed):
:shipit:

src/strings/PowerFxResources.en-US.resx
MikeStall: I can update  - fwiw, I just copied from Canvas apps.

src/tests/Microsoft.PowerFx.Interpreter.Tests/SetFunctionTests.cs
MikeStall: that's case is on 51.   (AllowsSideEffects  is the default case for Check, we need to pass an extra _opts param to override)

src/strings/PowerFxResources.en-US.resx
MikeStall: no - just edited it via the default VS Resx editor.  I can revert this part.

src/tests/Microsoft.PowerFx.Interpreter.Tests/SetFunctionTests.cs
Shpakh-MSFT: Ah. Got it.

src/strings/PowerFxResources.en-US.resx
MikeStall: Ok, then I'll leave as is to stay consistent with Canvas.

src/tests/Microsoft.PowerFx.Interpreter.Tests/SetFunctionTests.cs
MikeStall: I'll add Async test to ParallelAsync.txt

LucGenetier (approved):

src/tests/Microsoft.PowerFx.Interpreter.Tests/SetFunctionTests.cs
MikeStall: Will tackle this separately...

PR #395, opened 2022-05-18
Add Locale-Specific Parse to PowerFx Engine

### LINKED ISSUE #359: GetInvariantExpression  should also translate tokens.
Ie, it should be able to translate from localeExpr to invariantExpr  and back. 

			// Fixup will adjust from invariant to current locale:
			//   Invariant name (new_field) --> Display Name  (field)
			//   token separator (,) --> current locale separator (;)
			//   Numbers (1.1) --> local numbers (1,1)

			string invariantExpr = "If(true, 1.1, new_field)";
			string localeExpr = "If(true; 1,1; field)"; 
			string localeName = "fr-FR"; // pass to engine.

lesaltzm:
This adds locale-specific parsing support to the PowerFx Engine, based on the locale set in PowerFxConfig. Also extends ConvertToInvariant/ConvertToDisplay to also update punctuators. 

Resolves #359


lesaltzm:
This is a behavioral breaking change if a host is passing user's Culture as part of PowerFxConfig. Expressions will now be parsed using the culture-appropriate punctuator set.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -344,6 +345,25 @@ public void CheckParseError()
             Assert.StartsWith("Error 4-4: Expected an operand", result.Errors.First().ToString());
         }
 
+        [Fact]
+        public void CheckParseErrorCommaSeparatedLocale()
+        {
+            var engine = new RecalcEngine(new PowerFxConfig(CultureInfo.GetCultureInfo("it-IT")));
```
MikeStall, line 351: >RecalcEngine [](http://example.com/codeflow?start=29&length=12)

Could these be with Engine (instead of RecalcEngine) and in Core?

MikeStall (dismissed):
:shipit:

MikeStall (approved):

PR #397, opened 2022-05-19
Basic connector support 

MikeStall:
Introduces a new project / nuget, **Microsoft.PowerFx.Connectors** which provides the ability to call http REST endpoints, including ASP.Net + Swashbuckle sites and Power Platform connectors. 

Connectors can be added via calling this new extension method on config:

`config.AddService("MSNWeather", swagger, httpClient);`

The httpClient handles the auth. 
This uses https://github.com/microsoft/OpenAPI.NET for parsing swagger and OpenAPI 3.0 documents. 
This avoids PAClient's WADL / XML legacy support and uses OpenAPI directly. 

Includes some strong tests that demonstrate the end-2-end flow for both a generic REST call and also a Connector. 

Next steps:
- use real ServiceFunction from PAClient. 
- more complete swagger parser handling - pulling from PAClient. This should help with better decoding and parsing help text for intellisense. 
- remove dependency on interpreter by leveraging IR rewriting.


src/tests/Microsoft.PowerFx.Connectors.Tests/BasicRestTests.cs
```diff
@@ -0,0 +1,62 @@
...
+            var config = new PowerFxConfig();
+            var apiDoc = testConnector._apiDocument;
+            
+            config.AddService("Test", apiDoc, httpClient);
+
+            var engine = new RecalcEngine(config);
+
+            testConnector.SetResponse("55");
+            var r1 = await engine.EvalAsync("Test.GetKey(\"Key1\")", CancellationToken.None);            
+            Assert.Equal(55.0, r1.ToObject());
```
MikeStall, line 39: Power Fx is translating a functional expression like:
`Test.GetKey("Key1") `

into a REST call using the swagger.

MikeStall:
The $$$ markers are todo followups...

src/libraries/Microsoft.PowerFx.Connectors/ConfigExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// Add functions for each operation in the <see cref="OpenApiDocument"/>. 
+        /// Functions names will be 'functionNamespace.operationName'.
+        /// Functions are invoked via rest via the httpClient. The client must handle authentication. 
+        /// </summary>
+        /// <param name="config">Config to add the functions to.</param>
+        /// <param name="functionNamespace">Namespace to place functions in.</param>
+        /// <param name="openApiDocument">An API document. This can represent multiple formats, including Swagger 2.0 and OpenAPI 3.0.</param>
+        /// <param name="httpClient">Required iff we want to invoke the API. A client to invoke the endpoints described in the api document. This must handle auth or any other tranforms the API expects.</param>
+        /// <param name="cache">A cache to avoid redundant HTTP gets.</param>
+        public static IReadOnlyList<FunctionInfo> AddService(this PowerFxConfig config, string functionNamespace, OpenApiDocument openApiDocument, HttpMessageInvoker httpClient, ICachingHttpClient cache = null)
```
MikeStall, line 23: >public [](http://example.com/codeflow?start=8&length=6)

This is the key entry point we expose. 
The OpenApiDocument is from [https://github.com/microsoft/OpenAPI.NET](https://github.com/microsoft/OpenAPI.NET) and acts as a unified object model to represent a variety of rest API formats, including swagger (powerplatform connectors) and Open API (which is what ASP+Swashbuckle produce).
It also handles both json and modern yml formats.

src/libraries/Microsoft.PowerFx.Connectors/OpenApiParser.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Net.Http;
+using Microsoft.OpenApi.Models;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Connectors
+{
+    internal class OpenApiParser
+    {
+        // Parse an OpenApiDocument and return functions. 
+        public static List<ServiceFunction> Parse(
```
MikeStall, line 15: >public [](http://example.com/codeflow?start=8&length=6)

@darrelmiller  - here's were we traverse OpenApiDocument

src/Microsoft.PowerFx.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 17
-VisualStudioVersion = 17.0.31717.71
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.32428.217
```
lesaltzm, line 4: >16.0.32428.217 [](http://example.com/codeflow?start=22&length=14)

Version stamps in source control? :grinning:

src/libraries/Microsoft.PowerFx.Connectors/CachingHttpClient.cs
```diff
@@ -0,0 +1,101 @@
...
+
+    /// <summary>
+    /// Simple cache for GETs. Cache is reset when a POST is made. 
+    /// </summary>
+    public class CachingHttpClient : ICachingHttpClient
+    {
+        // $$$ Empty on max size. Don't monotonically grow...
+
+        // For GETs, map of URL to response
+        private readonly Dictionary<string, Dictionary<string, FormulaValue>> _cache = new Dictionary<string, Dictionary<string, FormulaValue>>();
```
lesaltzm, line 51: >string [](http://example.com/codeflow?start=102&length=6)

specify comparator?

src/libraries/Microsoft.PowerFx.Connectors/ServiceFunction.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        public HttpFunctionInvoker _invoker;
+
+        public ServiceFunction(string functionNamespace, string name, FormulaType returnType, params FormulaType[] paramTypes)
+: this(functionNamespace, name, returnType._type, Array.ConvertAll(paramTypes, x => x._type))
+        {
+        }
+
+        public ServiceFunction(string functionNamespace, string name, DType returnType, params DType[] paramTypes)
+            : base(DPath.Root.Append(new DName(functionNamespace)), name, name, SG("Custom func " + name), FunctionCategories.MathAndStat, returnType, 0, paramTypes.Length, paramTypes.Length, paramTypes)
```
lesaltzm, line 31: >MathAndStat [](http://example.com/codeflow?start=126&length=11)

FunctionCategories.REST?

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -0,0 +1,138 @@
...
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Connectors
+{
+    internal static class OpenApiExtensions
+    {
+        public static string GetBasePath(this OpenApiDocument openApiDocument)
+        {
+            string basePath = null;
+            if (openApiDocument?.Servers.Count == 1)
```
lesaltzm, line 17: >openApiDocument?.Servers.Count == 1 [](http://example.com/codeflow?start=16&length=35)

Not super familiar with OpenApi, when would this be something other than 1?

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -0,0 +1,138 @@
...
+                    }
+
+                case "object":
+                    var obj = new RecordType();
+                    foreach (var kv in schema.Properties)
+                    {
+                        var propName = kv.Key;
+                        var propType = kv.Value.ToFormulaType();
+
+                        obj = obj.Add(propName, propType);
```
lesaltzm, line 85: >Add [](http://example.com/codeflow?start=34&length=3)

Any risk of key collisions, or is that enforced by the OpenApi parser?

src/tests/Microsoft.PowerFx.Connectors.Tests/Microsoft.PowerFx.Connectors.Tests.csproj
```diff
@@ -0,0 +1,46 @@
...
+		<OutputType>Library</OutputType>		
+	</PropertyGroup>
+	<ItemGroup>
+	  <None Remove="Swagger\MSNWeather.json" />
+	  <None Remove="Swagger\MSNWeather_Response.json" />
+	  <None Remove="Swagger\TestOpenAPI.json" />
+	</ItemGroup>
+	<ItemGroup>
+        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.11.0" />
+        <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
lesaltzm, line 17: >Newtonsoft.Json [](http://example.com/codeflow?start=35&length=15)

why newtonsoft and not stj?

lesaltzm: Actually, I don't even see this using newtonsoft anywhere

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
MikeStall: I have no idea, but just being defensive. @darrelmiller ?

src/Microsoft.PowerFx.sln
MikeStall: should I revert? This is just VS being VS.

lesaltzm: Doesn't really matter, imo. Fine to leave as is.

src/libraries/Microsoft.PowerFx.Connectors/ServiceFunction.cs
MikeStall: using real ServiceFunction now.

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
```diff
@@ -0,0 +1,413 @@
+﻿// <autogenerated>
```
lesaltzm, line 1: Add a comment that this isn't actually autogenerated, it's just to disable stylecop warnings.

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
```diff
@@ -0,0 +1,413 @@
...
+
+//------------------------------------------------------------------------------
+// <copyright file="ServiceFunction.cs" company="Microsoft Corporation">
```
lesaltzm, line 4: nit: wrong copyright header

lesaltzm: on all these shared files

src/libraries/Microsoft.PowerFx.Connectors/ArgumentMapper.cs
```diff
@@ -0,0 +1,192 @@
...
+    // Map arguments between a Swagger File and a Function signature. 
+    // Deals with: 
+    //  - internals, "connectionId"
+    //  - optional, required, default value 
+    //  - computing signature type
+    //
+    // Given a swagger definition, handles:
+    //  - signature information to create a ServiceFunction
+    //  - reverse mapping to map from function signature back to swagger.
+    internal class ArgumentMapper
```
MikeStall, line 26: >ArgumentMapper [](http://example.com/codeflow?start=19&length=14)

This class is critical. 
I really wish existing PAClient code had something like this where the logic was all centralized in one place.

src/libraries/Microsoft.PowerFx.Connectors/Shared/IService.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// </summary>
+        [TransportDisabled]
+        bool IsInUse { get; }
+
+        /// <summary>The runtime definition for this entity as stringified JSON.</summary>
+        string RuntimeDefinitionJson { get; }
+    }
+
+    [TransportType(TransportKind.ServerRemoted)]
+    public interface IAddinService : IService
```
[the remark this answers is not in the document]
lesaltzm: ah, used by the attribute on IService :(

src/libraries/Microsoft.PowerFx.Connectors/Shared/IService.cs
```diff
@@ -0,0 +1,61 @@
...
+//     Copyright (c) Microsoft Corporation.  All rights reserved.
+// </copyright>
+//------------------------------------------------------------------------------
+
+using System;
+using Microsoft.AppMagic.Transport;
+
+namespace Microsoft.AppMagic.Authoring
+{
+    [TransportType(TransportKind.Enum)]
```
MikeStall, line 14: >TransportType [](http://example.com/codeflow?start=5&length=13)

can we avoid this file? TransportType is problematic.

src/libraries/Microsoft.PowerFx.Connectors/Shared/IService.cs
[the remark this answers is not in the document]
MikeStall: I think we can remove...

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
MikeStall: There shouldn't be, but if there is a collision, we'll take the exception on import.

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -0,0 +1,165 @@
...
+            foreach (var param in _argMapper._openApiParameters)
+            {
+                if (map.TryGetValue(param.Name, out var paramValue))
+                {
+                    var valueStr = paramValue.ToObject().ToString();
+
+                    switch (param.In.Value)
+                    {
+                        case ParameterLocation.Path:
+                            path = path.Replace("{" + param.Name + "}", HttpUtility.UrlEncode(valueStr));
```
LucGenetier, line 69: Should we provide the Encoding for UrlEncode?

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -0,0 +1,165 @@
...
+
+                        case ParameterLocation.Query:
+                            query.Append((query.Length == 0) ? "?" : "&");
+                            query.Append(param.Name);
+                            query.Append('=');
+                            query.Append(HttpUtility.UrlEncode(valueStr));
+                            break;
+
+                        default:
+                            throw new NotImplementedException($"{param.In}");
```
LucGenetier, line 80: param.In.Value?

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -0,0 +1,165 @@
...
+            return request;
+        }   
+
+        public async Task<FormulaValue> DecodeResponseAsync(HttpResponseMessage response)
+        {
+            var json = await response.Content.ReadAsStringAsync();
+
+            if (!response.IsSuccessStatusCode)
+            {
+                var msg = $"Connector called failed {response.StatusCode}): " + json;
```
LucGenetier, line 98: typo? (called -> call)

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -0,0 +1,165 @@
...
+
+            if (request.Method != HttpMethod.Get)
+            {
+                _cache.Reset(cacheScope);
+                key = null; // don't bother caching
+            }
+                        
+            var result2 = await _cache.TryGetAsync(cacheScope, key, async () =>
+            {
+                var response = await _httpClient.SendAsync(request, cancel);
```
LucGenetier, line 139: Should we add .ConfigureAwait(false) ?

LucGenetier: Comment applies to all async calls

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+        public static bool IsTrigger(this OpenApiOperation op)
+        {
+            var isTrigger = op.Extensions.ContainsKey("x-ms-trigger");
+            return isTrigger;
+        }
+
+        public static bool TryGetDefaultValue(this OpenApiParameter param, out string defaultValue)
+        {
+            var x = param.Schema.Default;
+            if (x == null)
```
LucGenetier, line 76: OpenApiNull exists, not sure if we should include it here

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+                return true;
+            }
+
+            if (x is OpenApiInteger intVal)
+            {
+                defaultValue = intVal.Value.ToString();
+                return true;
+            }
+
+            if (x is OpenApiDouble dbl)
```
LucGenetier, line 94: Shall we consider OpenApiFloat as well?

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+                return true;
+            }
+
+            if (x is OpenApiBoolean b)
+            {
+                defaultValue = b.Value.ToString();
+                return true;
+            }
+
+            throw new NotImplementedException($"Unknown default value type {x.GetType().FullName}");
```
LucGenetier, line 106: OpenApiDate, OpenApiDateTime ?

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+                return false;
+            }
+
+            if (x is OpenApiString str)
+            {
+                defaultValue = str.Value;
+                return true;
+            }
+
+            if (x is OpenApiInteger intVal)
```
LucGenetier, line 88: What about OpenApiLong?

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+
+            throw new NotImplementedException($"Unknown default value type {x.GetType().FullName}");
+        }
+
+        public static bool HasDefaultValue(this OpenApiParameter param)
+        {
+            return param.Schema.Default != null;
+        }
+
+        public static bool IsInternal(this OpenApiParameter param)
```
LucGenetier, line 114: I'd prefer the shorter version `public static bool IsInternal(this OpenApiParameter param) => param.Extensions.TryGetValue("x-ms-visibility", out _);`

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
LucGenetier: is the schema case sensitive?

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+                return FormulaType.Blank;
+            }
+
+            // Responses is a list by content-type. Find "application/json"
+            foreach (var kv3 in response200.Content)
+            {
+                var mediaType = kv3.Key;
+                var response = kv3.Value;
+
+                if (mediaType == "application/json")
```
LucGenetier, line 213: Should we do a case insensitive comparison?
[link](https://stackoverflow.com/questions/7718476/are-http-headers-content-type-c-case-sensitive)

src/libraries/Microsoft.PowerFx.Connectors/OpenApiParser.cs
```diff
@@ -0,0 +1,138 @@
...
+                    };
+
+                    newFunctions.Add(sfunc);
+                }
+            }
+
+            return newFunctions;
+        }
+
+        private static bool IsSafeHttpMethod(HttpMethod httpMethod)
```
LucGenetier, line 131: HttpMethod.Options is safe too.

src/libraries/Microsoft.PowerFx.Connectors/PowerPlatformConnectorClient.cs
```diff
@@ -0,0 +1,92 @@
...
+            Endpoint = endpoint ?? throw new ArgumentNullException(nameof(endpoint));
+
+            // Must set to allow callers to invoke SendAsync() via other helper methods.
+            BaseAddress = new Uri("https://" + endpoint); // Uri.Parse will validate endpoint syntax. 
+        }
+                
+        public override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            var url = request.RequestUri.ToString();
+            if (url[0] != '/')
```
LucGenetier, line 60: Could we use !request.RequestUri.IsAbsoluteUri() ?

src/libraries/Microsoft.PowerFx.Connectors/ArgumentMapper.cs
```diff
@@ -0,0 +1,192 @@
...
+            _parameterTypes = GetParamTypes(_requiredParamInfo, _optionalParamInfo).ToArray();
+        }
+
+        #region ServiceFunction args
+
+        // Useful for passing into a ServiceFunction
+        public readonly ServiceFunctionParameterTemplate[] _requiredParamInfo;
+        public readonly ServiceFunctionParameterTemplate[] _optionalParamInfo;
+
+        public readonly Dictionary<string, Tuple<string, DType>> _parameterDefaultValues = new Dictionary<string, Tuple<string, DType>>();
```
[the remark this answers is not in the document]
MikeStall: These are types that can be passed directly into ServiceFunction - so there's an existing contract to match.

src/libraries/Microsoft.PowerFx.Connectors/ArgumentMapper.cs
```diff
@@ -0,0 +1,192 @@
...
+        public readonly IEnumerable<OpenApiParameter> _openApiParameters;
+
+        public ArgumentMapper(IEnumerable<OpenApiParameter> parameters)
+        {
+            _openApiParameters = parameters;
+
+            var requiredParams = new List<OpenApiParameter>();
+            var optionalParams = new List<OpenApiParameter>();
+
+            foreach (var param in parameters)
```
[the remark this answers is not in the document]
MikeStall: Do you happen to have some swaggers for this and we can add to tests?  (I'd like to keep code-coverage up as I add the checks). Else the null ref is fine for now (it's still a failure) and we can add checks as we get swaggers.

src/libraries/Microsoft.PowerFx.Connectors/ArgumentMapper.cs
```diff
@@ -0,0 +1,192 @@
...
+
+                var options = param.GetOptions();
+                if (options != null)
+                {
+                    var typedName = new TypedName(paramType._type, new DName(name));
+                    _parameterOptions[typedName] = new List<string>(options);
+                }
+            }
+
+            _optionalParamInfo = optionalParams.ConvertAll(x => Convert(x)).ToArray();
```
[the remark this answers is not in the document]
MikeStall: This is calling List.ConvertAll, List.ToArray 
Those are likely more performant than the generic IEnumerable operators because they know the list size.

src/libraries/Microsoft.PowerFx.Connectors/ArgumentMapper.cs
```diff
@@ -0,0 +1,192 @@
...
+        public Dictionary<string, FormulaValue> ConvertToSwagger(FormulaValue[] args)
+        {
+            // Type check should have caught this.
+            Contracts.Assert(args.Length >= _arityMin);
+            Contracts.Assert(args.Length <= _arityMax);
+
+            // First N are required params. 
+            // Last param is a record with each field being an optional.
+
+            var map = new Dictionary<string, FormulaValue>(StringComparer.Ordinal);
```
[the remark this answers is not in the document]
MikeStall: That's a fine suggestion. I prefer the explicit approach here:
- easier to inspect in debugger
- more efficient (directly creates the dictionary in a single pass, whereas the linq query has 2 passes)

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -0,0 +1,165 @@
...
+        internal static void VerifyCanHandle(ParameterLocation? location)
+        {
+            switch (location.Value)
+            {
+                case ParameterLocation.Path:
+                case ParameterLocation.Query:
+                    break;
+
+                default:
+                    throw new NotImplementedException($"Unsupported ParameterIn {location}");
```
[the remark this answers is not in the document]
MikeStall: Nullable.ToString calls Nullable.Value.ToString, so not needed.

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
MikeStall: can you elaborate?

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
MikeStall: per above, not needed.

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
MikeStall: Is it necessary? Else we have 1000+ awaits to go update.

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -0,0 +1,230 @@
...
+            return null;
+        }
+
+        public static bool IsTrigger(this OpenApiOperation op)
+        {
+            var isTrigger = op.Extensions.ContainsKey("x-ms-trigger");
+            return isTrigger;
+        }
+
+        public static bool TryGetDefaultValue(this OpenApiParameter param, out string defaultValue)
```
MikeStall, line 73: >string [](http://example.com/codeflow?start=79&length=6)

is 'string' the right thing here? That assumes a format. 
 Or should it be FormulaValue / Object?

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
MikeStall: Note any missing case will hit the exception on line 106. 
We can then fill them in as we get tests.

src/libraries/Microsoft.PowerFx.Connectors/OpenApiParser.cs
MikeStall: Just copying from Power Apps. 

Spec still only lists Get/Head
[https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html)

But Options does seem to be considered safe in practice:
[https://developer.mozilla.org/en-US/docs/Glossary/Safe/HTTP](https://developer.mozilla.org/en-US/docs/Glossary/Safe/HTTP)

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
MikeStall: good link. I'll just NotImpl > 1 and then we can sort that out when we have a case.

src/libraries/Microsoft.PowerFx.Connectors/Execution/CachingHttpClient.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }        
+
+    /// <summary>
+    /// Simple cache for GETs. Cache is reset when a POST is made. 
+    /// </summary>
+    public class CachingHttpClient : ICachingHttpClient
+    {
+        // For GETs, map of URL to response
+        private readonly Dictionary<string, Dictionary<string, FormulaValue>> _cache = new Dictionary<string, Dictionary<string, FormulaValue>>(StringComparer.Ordinal);
```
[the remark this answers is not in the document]
MikeStall: The challenge is that we're trying to synchronize 2 data structures (the outer and nested dictionary).- ie, see line 69 and 71, and 86+89.

src/tests/Microsoft.PowerFx.Connectors.Tests/Microsoft.PowerFx.Connectors.Tests.csproj
lesaltzm: it's an additional dependency, STJ is part of the standard for new .NET versions, and is faster.

src/libraries/Microsoft.PowerFx.Connectors/ArgumentMapper.cs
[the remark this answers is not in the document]
LucGenetier: Just applying basic (stupid?) rule: public => we test all input params

src/libraries/Microsoft.PowerFx.Connectors/ArgumentMapper.cs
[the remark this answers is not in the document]
LucGenetier: Tried a piece of code to compare the durations with 10M items to see a differnce and ConvertAll might be a winner but the difference is so small it's really hard to be sure.

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
LucGenetier: If we want to avoid deadlocks when loaded in asp.net ... yes

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
LucGenetier: [https://docs.microsoft.com/en-us/dotnet/api/system.web.httputility.urlencode?view=net-6.0#system-web-httputility-urlencode(system-string-system-text-encoding](https://docs.microsoft.com/en-us/dotnet/api/system.web.httputility.urlencode?view=net-6.0#system-web-httputility-urlencode(system-string-system-text-encoding)) points to [https://docs.microsoft.com/en-us/previous-versions/aspnet/a2a4yykt(v=vs.100](https://docs.microsoft.com/en-us/previous-versions/aspnet/a2a4yykt(v=vs.100))

LucGenetier (dismissed):
:shipit:

src/tests/Microsoft.PowerFx.Connectors.Tests/Microsoft.PowerFx.Connectors.Tests.csproj
```diff
@@ -0,0 +1,45 @@
...
+    </ItemGroup>
+
+    <ItemGroup>
+      <EmbeddedResource Include="Responses\*.json" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <ProjectReference Include="..\..\libraries\Microsoft.PowerFx.Connectors\Microsoft.PowerFx.Connectors.csproj" />
+      <ProjectReference Include="..\Microsoft.PowerFx.Core.Tests\Microsoft.PowerFx.Core.Tests.csproj" />
+      <ProjectReference Include="..\..\libraries\Microsoft.PowerFx.Interpreter\Microsoft.PowerFx.Interpreter.csproj" />
```
MikeStall, line 41: > <ProjectReference Include="..\..\libraries\Microsoft.PowerFx.Interpreter\Microsoft.PowerFx.Interpreter.csproj" /> [](http://example.com/codeflow?start=5&length=114)

We would live to remove this, such as by using an IR transform or by moving IAsyncTexlFunction into core.

src/libraries/Microsoft.PowerFx.Connectors/Microsoft.PowerFx.Connectors.csproj
```diff
@@ -0,0 +1,35 @@
...
+        <PackageReference Include="Microsoft.OpenApi.Readers" Version="1.3.1" />
+		<PackageReference Include="System.Text.Json" Version="5.0.2" />
+		<PackageReference Include="System.Collections.Immutable" Version="5.0.0" />
+		<PackageReference Include="System.Memory" Version="4.5.4" />
+		<PackageReference Include="Microsoft.PowerFx.Core" Condition=" '$(ReleasePackageVersion)' == '' " Version="$(LocalPackageVersion)" />
+	</ItemGroup>
+
+	<ItemGroup>
+		<ProjectReference Include="..\Microsoft.PowerFx.Core\Microsoft.PowerFx.Core.csproj" />
+		<ProjectReference Include="..\Microsoft.PowerFx.Interpreter\Microsoft.PowerFx.Interpreter.csproj" />
```
MikeStall, line 33: >..\Microsoft.PowerFx.Interpreter\Microsoft.PowerFx.Interpreter.csproj" /> [](http://example.com/codeflow?start=29&length=73)

We would live to remove this, such as by using an IR transform or by moving IAsyncTexlFunction into core.

src/tests/Microsoft.PowerFx.Connectors.Tests/BasicRestTests.cs
```diff
@@ -0,0 +1,140 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
```
lesaltzm, line 2: Optional: We should probably have a PublicSurfaceTest as well

src/libraries/Microsoft.PowerFx.Connectors/PowerPlatformConnectorClient.cs
```diff
@@ -0,0 +1,92 @@
...
+            req.Headers.Add("x-ms-client-session-id", SessionId);
+            req.Headers.Add("x-ms-request-method", method.ToString());
+            req.Headers.Add("authorization", "Bearer " + authToken);
+            req.Headers.Add("x-ms-client-environment-id", "/providers/Microsoft.PowerApps/environments/" + EnvironmentId);
+            req.Headers.Add("x-ms-user-agent", $"PowerFx/{Version}");
+
+            req.Headers.Add("x-ms-request-url", url); 
+
+            // $$$ Body? 
+
```
lesaltzm, line 87: What is the to-do here?

src/libraries/Microsoft.PowerFx.Connectors/Execution/CachingHttpClient.cs
```diff
@@ -0,0 +1,101 @@
...
+        // Skip cache if requestKey is null. 
+        Task<FormulaValue> TryGetAsync(string cacheScope, string requestKey, Func<Task<FormulaValue>> getter);
+    }
+
+    /// <summary>
+    /// Implementation that skips caching. 
+    /// </summary>
+    public class NonCachingClient : ICachingHttpClient
+    {
+        public static ICachingHttpClient Instance { get; } = new NonCachingClient();
```
lesaltzm, line 32: >NonCachingClient [](http://example.com/codeflow?start=65&length=16)

add a private constructor to force use of the singleton?

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -0,0 +1,235 @@
...
+                    {
+                        // Primitives get marshalled as a SingleColumn table.
+                        // Make sure this is consistent with invoker. 
+                        var r2 = new RecordType().Add(TableValue.ValueName, elementType);
+                        return r2.ToTable();
+                    }
+                    else
+                    {
+                        throw new NotImplementedException();                    
+                    }
```
lesaltzm, line 152: Swagger supports mixed type arrays, should map to UntypedObject in the future.

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -0,0 +1,235 @@
...
+                        var propName = kv.Key;
+                        var propType = kv.Value.ToFormulaType();
+
+                        obj = obj.Add(propName, propType);
+                    }
+
+                    return obj;
+
+                case null:
+                    // Missing a schema is ok - that signifies void.
```
lesaltzm, line 167: >that signifies void [](http://example.com/codeflow?start=48&length=19)

from [https://swagger.io/docs/specification/data-models/data-types/](https://swagger.io/docs/specification/data-models/data-types/) (Any Type) header, looks like an empty schema is dynamic, not void?

lesaltzm (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Connectors/PowerPlatformConnectorClient.cs
MikeStall: handle body parameter which may need to be copied over from incoming request to APIM request.

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -0,0 +1,235 @@
...
+        // Internal parameters are not showen to the user. 
+        // They can have a default value or be special cased by the infrastructure (like "connectionId").
+        public static bool IsInternal(this OpenApiParameter param) => param.Extensions.TryGetValue("x-ms-visibility", out _);
+
+        // See https://swagger.io/docs/specification/data-models/data-types/
+        public static FormulaType ToFormulaType(this OpenApiSchema schema)
+        {
+            switch (schema.Type)
+            {
+                case "string": return FormulaType.String;
```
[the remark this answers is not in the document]
MikeStall: yes - although any missing case will get the not-implemented on line 172 below.  So as we add test cases that excercise , we'll fill these in.

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
MikeStall: good point.

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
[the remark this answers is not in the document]
lesaltzm: This is a bit tricky from a back compat perspective, however. They'll currently parse as Strings, but later will change to DateTime. We're preview, so that's ok, but we should let consumers know about that.

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
MikeStall: The key thing is that we're calling UrlEncode at all.

Are you talking about passing the 2nd encoding parameter here as opposed to using the default?

src/libraries/Microsoft.PowerFx.Connectors/Execution/CachingHttpClient.cs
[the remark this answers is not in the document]
MikeStall: They're both equally suspectable and need to be protected.

src/libraries/Microsoft.PowerFx.Connectors/OpenApiParser.cs
```diff
@@ -131,6 +131,7 @@ internal class OpenApiParser
         private static bool IsSafeHttpMethod(HttpMethod httpMethod)
         {
             // HTTP/1.1 spec states that only GET and HEAD requests are 'safe' by default.
+            // https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html
             return httpMethod == HttpMethod.Get ||
```
LucGenetier, line 135: 2616 is a bit old
https://www.rfc-editor.org/rfc/rfc7231#section-4.2.1
   Of the request methods defined by this specification, the GET, HEAD,
   OPTIONS, and TRACE methods are defined to be safe.

LucGenetier (approved):

src/libraries/Microsoft.PowerFx.Connectors/OpenApiParser.cs
MikeStall: Good points - I wonder if PowerApps should be updated. will followup separaretly...

PR #398, opened 2022-05-19
make diagnostics startChar 0-based

LucGenetier:


src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/LanguageServerTests.cs
```diff
@@ -343,6 +344,30 @@ public void TestDidOpenSeverityFormula()
             TestPublishDiagnostics("powerfx://app", "textDocument/didOpen", formula, expectedDiagnostics);
         }
 
+        [Fact]
+        public void TestConcatenateStartAndEndChars()
```
MikeStall, line 348: >TestConcatenateStartAndEndChars [](http://example.com/codeflow?start=20&length=31)

Can we update to 
TestDidOpenWithErrors()

(it's not about concat). 

I'm seeing this is the only test that checks:
- in a case where the expression has errors. 
- on an /didOpen message, 
- Params.Diagnostics 
so we have a gap!!

LucGenetier: Done

MikeStall:
>                     var span = item.Span;

We really out to have a Contract.Assert here that the range is valid. 

Or even better, factor this into a helper method:

static Range GetRange(string expression, Span span);

And then we could unit test that directly. Looks like there are some interesting case here, especially with multi-line expressions. 


---
Refers to: src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs:432 in 685eee2. [](commit_id = 685eee29b5ef566749e1f51a6c5d186cec1e1dc4, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
```diff
@@ -55,6 +55,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\libraries\Microsoft.PowerFx.Core\Microsoft.PowerFx.Core.csproj" />
+    <ProjectReference Include="..\..\libraries\Microsoft.PowerFx.LanguageServerProtocol\Microsoft.PowerFx.LanguageServerProtocol.csproj" />
```
MikeStall, line 58: >LanguageServerProtocol [](http://example.com/codeflow?start=65&length=22)

Layering issue... WE should avoid adding this dependency in core. 

we can use .Interpreter Tests for now - it already has the dependency.  (in a pure world, we'd have LSp.Tests)

LucGenetier: Argh

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -468,6 +449,37 @@ private void PublishDiagnosticsNotification(string uri, string expression, Expre
                 }));
         }
 
+        /// <summary>
+        /// Construct a Range based on a Span for a given expression.
+        /// </summary>
+        /// <param name="expression">The expression.</param>
+        /// <param name="span">The Span.</param>
+        /// <returns>Generated Range.</returns>
+        public static Range GetRange(string expression, Span span)
```
MikeStall, line 458: >GetRange [](http://example.com/codeflow?start=28&length=8)

Can we add Contract.Asserts here?

(startLine < endLine) || 
(startLine == endLine  && startChar <= endChar)

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -468,6 +449,41 @@ private void PublishDiagnosticsNotification(string uri, string expression, Expre
...
+                    Line = startLine
+                },
+                End = new Position()
+                {
+                    Character = endChar,
+                    Line = endLine
+                }
+            };
+
+            Contracts.Assert(range.IsValid());
```
MikeStall, line 482: >Contracts.Assert(range.IsValid()); [](http://example.com/codeflow?start=12&length=34)

Yes!! With this assert, I think existing tests would have caught this bug.

MikeStall (approved):
:shipit:

MikeStall (approved):

lesaltzm (approved):

PR #400, opened 2022-05-19
Hygiene: Fix or disable all warnings, enable Warnings-As-Errors

lesaltzm:


src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineAsyncTests.cs
```diff
@@ -19,14 +19,16 @@ namespace Microsoft.PowerFx.Tests
...
     {
-        // Trivial async function that runs synchronously. 
+        // Intentionally trivial async function that runs synchronously. 
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
         private static async Task<FormulaValue> Worker(FormulaValue[] args, CancellationToken cancel)
         {
             var n = (NumberValue)args[0];
 
             var result = FormulaValue.New(n.Value * 2);
             return result;
```
CarlosFigueiraMSFT, line 29: >result [](http://example.com/codeflow?start=19&length=6)

You can use `return Task.FromResult(result);` - and that would remove the need to use the `async` method modifier

CarlosFigueiraMSFT (dismissed):
:shipit:

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineAsyncTests.cs
```diff
@@ -19,14 +19,16 @@ namespace Microsoft.PowerFx.Tests
     // Test async eval features. 
     public class RecalcEngineAsyncTests : PowerFxTest
     {
-        // Trivial async function that runs synchronously. 
+        // Intentionally trivial async function that runs synchronously. 
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
```
MikeStall, line 23: Can we remove this warning?
async keyword actually matters for how AggregateException is handled.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/TestRunner.cs
```diff
@@ -242,7 +242,7 @@ public TestRunFullResults RunTests()
                 {
                     var engineName = runner.GetName();
 
-                    var (result, msg) = runner.RunAsync(testCase).Result;
```
MikeStall, line 245: >Result [](http://example.com/codeflow?start=66&length=6)

double check?

lesaltzm: Intended, runasync isn't actually async (I'll fix the name)

MikeStall (changes requested):
:clock1:

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -316,14 +316,14 @@ private ValueTask<FormulaValue> VisitBinaryOpNode(BinaryOpNode node, SymbolConte
...
-                    return OperatorDynamicGetField(node, args);
+                    return new ValueTask<FormulaValue>(OperatorDynamicGetField(node, args));
 
                 default:
                     return new ValueTask<FormulaValue>(CommonErrors.UnreachableCodeError(node.IRContext));
             }
         }
 
-        private static async ValueTask<FormulaValue> OperatorDynamicGetField(BinaryOpNode node, FormulaValue[] args)
+        private static FormulaValue OperatorDynamicGetField(BinaryOpNode node, FormulaValue[] args)
```
CarlosFigueiraMSFT, line 326: >OperatorDynamicGetField [](http://example.com/codeflow?start=36&length=23)

Not related to your changes, but all other operators are async functions (using the error wrapper) - we should make this one as well (and simplify the blank/error handling on it. Can be dealt with in a separate PR

CarlosFigueiraMSFT (approved):
:shipit:

Issue #402, opened 2022-05-20
Exception while checking 'in' operator with table as left argument


ivanradicek:
## Repro steps

Use engine to check the following formulas (without any parameters):

```
Table() in ""
```

## Expected behavior

Result with an appropriate error (e.g., `Invalid argument type`).

## Actual behavior

Internal exception:

```
Visitor.CheckInArgTypes(TexlNode left, TexlNode right) line 2705
Visitor.PostVisit(BinaryOpNode node) line 4032
BinaryOpNode.Accept(TexlVisitor visitor) line 76
Visitor.Run() line 2473
TexlBinding.Run(IBinderGlue glue, IExternalRuleScopeResolver scopeResolver, DataSourceToQueryOptionsMap queryOptionsMap, TexlNode node, INameResolver resolver, BindingConfig bindingConfig, Boolean updateDisplayNames, DType ruleScope, Boolean forceUpdateDisplayNames, IExternalRule rule, Boolean useThisRecordForRuleScope) line 346
TexlBinding.Run(IBinderGlue glue, TexlNode node, INameResolver resolver, BindingConfig bindingConfig, DType ruleScope, Boolean useThisRecordForRuleScope) line 378
Engine.CheckInternal(ParseResult parse, RecordType parameterType, BindingConfig bindingConfig) line 125
Engine.Check(String expressionText, RecordType parameterType, ParserOptions options) line 101
```

PR #370, opened 2022-05-04
Adding support for MultiSelect optionset with In operator

kartikkr11:
#### Summary

MultiSelect Optionsets are currently not supported with the **in** operator, so customers didn't have an easy way to filter records based on the multi-select optionset column, with this change we will now be able to support In operator filtering on these columns in both delegated and non delegated queries.

example: If we need to filter a column - > states visited by a contact (multiselectoptionset) 

The formula would look like this:
**Filter(Contacts,multiselect_state in SelectedStates)**

with this PR we support the both delegated and non delegated scenarios:

1. Delegated query :https://org7c2cf9ed.crm10.dynamics.com/api/data/v9.0/contacts?$filter=Microsoft.Dynamics.CRM.ContainValues(PropertyName=%27cr7fe_multiselect_state%27,PropertyValues=[%27304050000%27])%20and%20Microsoft.Dynamics.CRM.ContainValues(PropertyName=%27cr7fe_multiselect_state%27,PropertyValues=[%27304050001%27])&$select=contactid,cr7fe_multiselect_state,fullname 

2. Non-Delegated: We would locally iterate over all the records and return the records that contain either of the selected optionset values.

#### Validation

Manual Tests.
TODO: Unit Tests.

#### Risks

New Feature, the risk is limited and we also have this under our Enhanced Delegation feature flag.

PowerApps Client uptake PR: https://msazure.visualstudio.com/OneAgile/_git/PowerApps-Client/pullrequest/6030875

PR #403, opened 2022-05-20
Fix null exception in binder

ivanradicek:
Fixes issue #402.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -2702,7 +2702,8 @@ namespace Microsoft.PowerFx.Core.Binding
                     return false;
                 }
 
-                if (_txb.Document.Properties.EnabledFeatures.IsEnhancedDelegationEnabled && typeLeft.IsTable)
+                if ((_txb.Document == null || _txb.Document.Properties.EnabledFeatures.IsEnhancedDelegationEnabled)
+                    && typeLeft.IsTable)
                 {
                     // Table in table: RHS must be a single column table with a compatible schema. No coercion is allowed.
                     if (typeRight.IsTable)
--- a/src/tests/Microsoft.PowerFx.Core.Tests/BindingEngineTests.cs
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/BindingEngineTests.cs
@@ -179,6 +179,18 @@ namespace Microsoft.PowerFx.Tests
             AssertContainsError(result, "Error 7-11: Name isn't valid. 'foo' isn't recognized");
         }
 
+        [Fact]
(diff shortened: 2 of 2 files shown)


src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -2702,7 +2702,8 @@ private bool CheckInArgTypes(TexlNode left, TexlNode right)
                     return false;
                 }
 
-                if (_txb.Document.Properties.EnabledFeatures.IsEnhancedDelegationEnabled && typeLeft.IsTable)
+                if ((_txb.Document == null || _txb.Document.Properties.EnabledFeatures.IsEnhancedDelegationEnabled)
```
ivanradicek, line 2705: Not sure whether this is correct. Perhaps it should be `&&`?

MikeStall:
It looks like this was introduced recently with #370.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
MikeStall, line 2705: >Document [](http://example.com/codeflow?start=26&length=8)

We use _txb.Document all over... I wonder if there are other places we could nullref...

_txb.Document just pulls from the resolver. 

Engine/Recalc engine produce a SimplerResolver that always has a null document.  
PowerApps has its own resolver with a non-nulldocument.

MikeStall: I think || is fine. 

We should assume that all feature flags will eventually become True.

CarlosFigueiraMSFT: `&&` would be safer. Currently the `in` operator doesn't work for tables in the interpreter for general tables, and even on Canvas it only works for a specific type of tables (multi-select option sets).

kartikkr11: Yea, I feel && is better too, since its specific to a type of table that we are adding support for and only when the flag is set.

kartikkr11: +1 , any references to _txb.Document should have null checks

ivanradicek: Changed the condition.

ivanradicek: I have used search to add couple of more null checks. Have no idea whether I covered all of them or whether they are correct.

lesaltzm (approved):

CarlosFigueiraMSFT (approved):
:shipit:

PR #409, opened 2022-05-30
Better error message for missing fields

omprakaash:
Added a new error message for missing fields. Not sure how to deal with implicit records with no aliases (Eg: [1,2,3], []). I default to the previous error message(ErrInvalidName) for now.


MikeStall:
@omprakaash - thanks, will take a look. Curious - are you consuming Fx in a project?

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -3666,7 +3666,16 @@ public override void PostVisit(DottedNameNode node)
                     }
                     else
                     {
-                        SetDottedNameError(node, TexlStrings.ErrInvalidName, node.Right.Name.Value);
+                        FirstNameNode lhsFirstNameNode;
+                        if ((lhsFirstNameNode = node.Left.AsFirstName()) != null)
```
MikeStall, line 3670: nit:
our convention is to avoid assignment inside expressions.

Can you change to:
FirstNameNode lhsFirstNameNode = node.Left.AsFirstName();
if (lhsFirstNameNode != null)

MikeStall: What about a cases where lhs is not a firstname node?
 a.b.missing
 Call(...).missing

lesaltzm: Generally, the pattern should be 
`if (node.Left is FirstNameNode lhsFirstNameNode)`..., combines the cast and null check.

omprakaash:
> @omprakaash - thanks, will take a look. Curious - are you consuming Fx in a project?

@MikeStall  Yep. We are using Power-Fx to implement formula columns in an ERP product.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -389,6 +389,53 @@ public void CheckDottedBindError2()
             Assert.StartsWith("Error 2-8: Name isn't valid. 'Value' isn't recognized", result.Errors.First().ToString());
         }
 
+        [Fact]
+        public void CheckDottedBindError3()
```
lesaltzm, line 393: These tests should probably live in BindingEngineTests.cs, since they're not specific to just RecalcEngine.

omprakaash: Got it. Moved newly added tests to BindingEngineTests.cs

lesaltzm (approved):

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5762,6 +5762,10 @@
     <value>https://go.microsoft.com/fwlink/?linkid=2132478</value>
     <comment>{Locked}</comment>
   </data>
+  <data name="ErrorResource_ErrInvalidField_ShortMessage" xml:space="preserve">
```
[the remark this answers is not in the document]
omprakaash: Got it. Duplicated error messages for InvalidField.

lesaltzm (approved):

PR #411, opened 2022-05-30
Generate xml docs with usual conventions

### LINKED ISSUE #410: XML docs aren't generataed correctly, VS doesn't show in intellisense
XML docs are not generated correctly for consumption in downstream projects, even though many public types and members in Power-FX have useful comments that would help users of the library.

Seems like just following usual conventions would avoid pain (i.e. let the standard build targets create the xml documentation file naturally via `<GenerateDocumentationFile>true</GenerateDocumentationFile>` instead of manually specifying prop `DocumentationFile` in each csproj.

Example:
https://github.com/microsoft/Power-Fx/blob/8f4ee9629a0ec10212933e38e2df12b0945d8215/src/libraries/Microsoft.PowerFx.Core/Microsoft.PowerFx.Core.csproj#L8

davidnx:
Fixes #410

### DIFF
--- a/src/libraries/Directory.Build.props
+++ b/src/libraries/Directory.Build.props
@@ -2,6 +2,7 @@
 <Project>
   <PropertyGroup>
     <GeneratePackageOnBuild>$(GeneratePackages)</GeneratePackageOnBuild>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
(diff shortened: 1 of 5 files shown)


lesaltzm (approved):

MikeStall:
@davidnx  - thanks for the fix!

PR #422, opened 2022-06-10
Update FirstN/LastN to treat blank in the second argument as 0

CarlosFigueiraMSFT:
Blank values should be treated as 0 in numeric context; this updates the FirstN and LastN functions to do that. It also updates the Count* functions to return 0 when given an empty table.


lesaltzm (approved):
:ship:

PR #426, opened 2022-06-14
Add body support for connectors

LucGenetier:


src/tests/Microsoft.PowerFx.Connectors.Tests/Helpers/LoggingTestServer.cs
```diff
@@ -61,6 +61,16 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
                 _log.AppendLine($" {headerName}: {value}");
             }
 
+            var httpContent = request?.Content;
+            if (httpContent != null)
+            {
+                var content = await httpContent.ReadAsStringAsync();
```
MikeStall, line 67: >httpContent [](http://example.com/codeflow?start=36&length=11)

Can we log content headers too? (especially content-type)

src/tests/Microsoft.PowerFx.Connectors.Tests/BasicRestTests.cs
```diff
@@ -19,7 +17,7 @@ public class BasicRestTests : PowerFxTest
     {
         // Must set the BaseAddress on an httpClient, even if we don't actually use it. 
         // All the Send() methods will enforce this. 
-        private static readonly Uri _fakeBaseAddress = new Uri("http://localhost:5000");
```
MikeStall, line 22: >Uri [](http://example.com/codeflow?start=59&length=3)

?

src/tests/Microsoft.PowerFx.Connectors.Tests/Microsoft.PowerFx.Connectors.Tests.csproj
```diff
@@ -8,10 +8,13 @@
...
-	  <None Remove="Swagger\TestOpenAPI.json" />
-	  <None Remove="Swagger\TestOpenAPI2.json" />
+    <None Remove="Responses\AzureBlobStorage_Response.json" />
+    <None Remove="Swagger\AzureBlobStorage.json" />
+    <None Remove="Swagger\MSNWeather.json" />
+    <None Remove="Swagger\MSNWeather_Response.json" />
+    <None Remove="Swagger\TestOpenAPI.json" />
+    <None Remove="Swagger\TestOpenAPI2.json" />
+    <None Remove="Swagger\TestOpenAPI3.json" />
 	</ItemGroup>
```
MikeStall, line 18: Do lines 34..40 cover this, can we remove this whole section?

src/tests/Microsoft.PowerFx.Connectors.Tests/PowerPlatformConnectorTests.cs
```diff
@@ -75,6 +78,68 @@ public async Task Connector1()
...
+            testConnector.SetResponseFromFile(@"Responses\AzureBlobStorage_Response.json");
+
+            var result = await engine.EvalAsync(
+                @"AzureBlobStorage.CreateFile(""container"", ""bora1.jpg"", ""abc"")",
+                CancellationToken.None, 
+                options: new ParserOptions() { AllowsSideEffects = true });
+
+            var res = (ExpandoObject)result.ToObject();
+            var propertyList = res.ToList();
+            var size = (double)propertyList.First(prop => prop.Key == "Size").Value;
```
MikeStall, line 116: Can we avoid this and either:

1. update expression on line 110, just append ".size"  and do this in Fx. 

or 
2. use C# dynamic 
dynamic res = result.ToObject();
res.size

src/tests/Microsoft.PowerFx.Connectors.Tests/PowerPlatformConnectorTests.cs
```diff
@@ -75,6 +78,68 @@ public async Task Connector1()
...
+            // Function we added where specified in MSNWeather.json
+            var funcNames = funcs.Select(func => func.Name).OrderBy(x => x).ToArray();
+            Assert.Equal(funcNames, new string[] { "AppendFile", "CopyFile", "CopyFile_Old", "CreateFile", "CreateFile_Old", "DeleteFile", "DeleteFile_Old", "ExtractFolder_Old", "ExtractFolderV2", "GetDataSetsMetadata", "GetFileContent", "GetFileContent_Old", "GetFileContentByPath", "GetFileContentByPath_Old", "GetFileMetadata", "GetFileMetadata_Old", "GetFileMetadataByPath", "GetFileMetadataByPath_Old", "ListAllRootFolders", "ListAllRootFoldersV2", "ListFolder", "ListFolder_Old", "ListFolderV2", "ListRootFolder", "ListRootFolder_Old", "ListRootFolderV2", "TestConnection", "UpdateFile", "UpdateFile_Old" });
+
+            // Now execute it...
+            var engine = new RecalcEngine(config);            
+            testConnector.SetResponseFromFile(@"Responses\AzureBlobStorage_Response.json");
+
+            var result = await engine.EvalAsync(
+                @"AzureBlobStorage.CreateFile(""container"", ""bora1.jpg"", ""abc"")",
```
MikeStall, line 110: >jpg [](http://example.com/codeflow?start=69&length=3)

for test sanity ... 
should this be a .txt extension?

src/tests/Microsoft.PowerFx.Connectors.Tests/BasicRestTests.cs
LucGenetier: no need :)

src/libraries/Microsoft.PowerFx.Connectors/OpenApiParser.cs
```diff
@@ -58,12 +60,7 @@ internal class OpenApiParser
                     }
 
                     var operationName = op.OperationId ?? path.Replace("/", string.Empty);
-
-                    if (op.RequestBody != null)
-                    {
-                        // RequestBody can be ambiguous- treat as a single record? splat as parameters?
-                        throw new NotImplementedException($"Request body");
-                    }
+                    var bodyType = null as FormulaType;
```
MikeStall, line 63: >null as FormulaType [](http://example.com/codeflow?start=35&length=19)

?
Is this just: 
FormulaType bodyType = null;

Although doesn't look like we use it

src/libraries/Microsoft.PowerFx.Connectors/ArgumentMapper.cs
```diff
@@ -82,6 +98,35 @@ public ArgumentMapper(IEnumerable<OpenApiParameter> parameters)
...
+            if (requestBody != null)
+            {
+                var bodyName = requestBody.GetBodyName() ?? BodyParameter;
+                FxOpenApiParameter bodyParameter;
+
+                if (requestBody.Content != null && requestBody.Content.Any())
+                {
+                    var schema = requestBody.Content.First().Value.Schema;                    
+
+                    if (schema.AllOf.Any() || schema.AnyOf.Any() || schema.Not != null || schema.Items != null || schema.AdditionalProperties != null)
```
MikeStall, line 110: >if (schema.AllOf.Any() || schema.AnyOf.Any() || schema.Not != null || schema.Items != null || schema.AdditionalProperties != null) [](http://example.com/codeflow?start=20&length=130)

great to check this!

src/tests/Microsoft.PowerFx.Connectors.Tests/Helpers/LoggingTestServer.cs
LucGenetier: Done

src/tests/Microsoft.PowerFx.Connectors.Tests/Microsoft.PowerFx.Connectors.Tests.csproj
LucGenetier: VS...
Fixed

src/tests/Microsoft.PowerFx.Connectors.Tests/PowerPlatformConnectorTests.cs
LucGenetier: sure

src/libraries/Microsoft.PowerFx.Connectors/ArgumentMapper.cs
```diff
@@ -82,6 +98,35 @@ public ArgumentMapper(IEnumerable<OpenApiParameter> parameters)
...
             }
 
+            if (requestBody != null)
+            {
+                var bodyName = requestBody.GetBodyName() ?? BodyParameter;
+                FxOpenApiParameter bodyParameter;
+
+                if (requestBody.Content != null && requestBody.Content.Any())
+                {
+                    var schema = requestBody.Content.First().Value.Schema;
```
MikeStall, line 108: >Content.First() [](http://example.com/codeflow?start=45&length=15)

Should we throw NotSupported unless Length=1?

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -63,31 +65,51 @@ public HttpRequestMessage BuildRequest(FormulaValue[] args)
...
+
+                    foreach (var property in param.Schema.Properties)
+                    {
+                        if (map.TryGetValue(property.Key, out var paramValue))
+                        {
+                            props.Add(property.Key, JsonSerializer.Serialize(paramValue.ToObject()));
+                        }
+                    }
+
+                    body = $"{{ {string.Join(", ", props.Select(p => $"{p.Key}: {p.Value}"))} }}";
```
MikeStall, line 85: >body = $"{{ {string.Join(", ", props.Select(p => $"{p.Key}: {p.Value}"))} }}"; [](http://example.com/codeflow?start=20&length=78)

??
Is this attempting Json Serialization?

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -63,31 +65,51 @@ public HttpRequestMessage BuildRequest(FormulaValue[] args)
             // Header names are not case sensitive.
             // From RFC 2616 - "Hypertext Transfer Protocol -- HTTP/1.1", Section 4.2, "Message Headers"
             var headers = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+            string body = null;
 
             var map = _argMapper.ConvertToSwagger(args);
             foreach (var param in _argMapper._openApiParameters)
             {
-                if (map.TryGetValue(param.Name, out var paramValue))
+                if (param.In == FxParameterLocation.Body && param.Schema.Properties != null && param.Schema.Properties.Any())
```
MikeStall, line 73: >if (param.In == FxParameterLocation.Body && param.Schema.Properties != null && param.Schema.Properties.Any()) [](http://example.com/codeflow?start=16&length=109)

What is this case handling? (compared to line 109)

src/libraries/Microsoft.PowerFx.Connectors/ArgumentMapper.cs
LucGenetier: Nope
There are usually multiple and from what I've observed, the ".Value.Schema" is the same for all
Not sure if that's a universal rule though

src/libraries/Microsoft.PowerFx.Connectors/ArgumentMapper.cs
LucGenetier: Thanks!

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
LucGenetier: I could try serializing a dictionary instead, that will be cleaner

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
LucGenetier: line 109 = raw body
here we have a description for its content

src/tests/Microsoft.PowerFx.Connectors.Tests/PowerPlatformConnectorTests.cs
MikeStall: and remove the dead code ;)

src/libraries/Microsoft.PowerFx.Connectors/FxParameterLocation.cs
```diff
@@ -0,0 +1,33 @@
...
+        public static FxParameterLocation? ToFxParameterLocation(this ParameterLocation? location)
+        {
+            return location switch
+            {
+                null => null,
+                ParameterLocation.Header => FxParameterLocation.Header,
+                ParameterLocation.Query => FxParameterLocation.Query,
+                ParameterLocation.Path => FxParameterLocation.Path,
+                ParameterLocation.Cookie => FxParameterLocation.Cookie,
+                _ => throw new NotImplementedException("Unknown location")
```
MikeStall, line 29: >"Unknown location" [](http://example.com/codeflow?start=55&length=18)

in general, would be good ot include location in this message for diagnostics. 

 throw new NotImplementedException($"Unknown location {location}")

{} will handle null.

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -63,31 +65,56 @@ public HttpRequestMessage BuildRequest(FormulaValue[] args)
...
+                    {
+                        if (map.TryGetValue(property.Key, out var paramValue))
+                        {
+                            props.Add(property.Key, paramValue.ToObject());
+                        }
+                    }
+
+                    body = _argMapper.ContentType switch
+                    {
+                       "application/x-www-form-urlencoded" => props.ToFormUrlEncoded(),
```
MikeStall, line 87: >application/x-www-form-urlencoded" [](http://example.com/codeflow?start=24&length=34)

we should make these constants somewhere
(see OpenApiExtensions._applicationJson) 

ARe these case sensntive?

MikeStall: actually ... since this is a specific list we need to implement (and not just a string we pass through), should this be an enum?

src/libraries/Microsoft.PowerFx.Connectors/Execution/SerializationExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+using System.Text.Json;
+using System.Text.RegularExpressions;
+using System.Web;
+using System.Xml.Linq;
+
+namespace Microsoft.PowerFx.Connectors.Execution
+{
+    public static class SerializationExtensions
+    {
+        public static string ToJson(this Dictionary<string, object> dic) => Regex.Replace(JsonSerializer.Serialize(dic), @"""(?<k>\w+)"":", "$1:");
```
MikeStall, line 16: >Regex.Replace [](http://example.com/codeflow?start=76&length=13)

We should absolutely not be using regex in parsing....

src/libraries/Microsoft.PowerFx.Connectors/Execution/SerializationExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        public static string ToFormUrlEncoded(this Dictionary<string, object> dic) => JsonDocument.Parse(JsonSerializer.Serialize(dic)).RootElement.JsonElementToString(null);
+
+        private static string JsonElementToString(this JsonElement je, string prefix) => je.ValueKind switch
```
MikeStall, line 20: >JsonElementToString [](http://example.com/codeflow?start=30&length=19)

What is this doing? Is there an existing json function to handle this?

src/tests/Microsoft.PowerFx.Connectors.Tests/PowerPlatformConnectorTests.cs
LucGenetier: Done

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
LucGenetier: Here we verify we have a body and check if we have a schema description for it

src/libraries/Microsoft.PowerFx.Connectors/FxOpenApiParameter.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+using Microsoft.OpenApi.Interfaces;
+using Microsoft.OpenApi.Models;
+
+namespace Microsoft.PowerFx.Connectors
+{
+    internal class FxOpenApiParameter
```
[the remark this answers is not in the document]
LucGenetier: Done

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -47,8 +42,10 @@ internal static void VerifyCanHandle(ParameterLocation? location)
                 case ParameterLocation.Path:
                 case ParameterLocation.Query:
                 case ParameterLocation.Header:
+                //case FxParameterLocation.Body:
```
LucGenetier, line 45: >//case FxParameterLocation.Body: [](http://example.com/codeflow?start=16&length=32)

Will remove

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -63,50 +60,97 @@ public HttpRequestMessage BuildRequest(FormulaValue[] args)
...
+            var request = new HttpRequestMessage(_method, url);            
 
             foreach (var kv in headers)
             {
                 request.Headers.Add(kv.Key, kv.Value);
             }
 
+            if (!string.IsNullOrEmpty(body))
+            {                
+                // $$$ Might need System.Net.Http.FormUrlEncodedContent()
```
LucGenetier, line 120: >// $$$ Might need System.Net.Http.FormUrlEncodedContent() [](http://example.com/codeflow?start=16&length=57)

Temporary comment, I think we can remove it
As we serialize the content to a string, calling StringContent is enough and we provide the right content type

src/libraries/Microsoft.PowerFx.Connectors/ArgumentMapper.cs
```diff
@@ -27,17 +27,37 @@ internal class ArgumentMapper
         // All connectors have an internal parameter named connectionId. 
         // This is handled specially and value passed by connector. 
         private const string ConnectionIdParamName = "connectionId";
+        public const string BodyParameter = "body";
```
MikeStall, line 30: >BodyParameter [](http://example.com/codeflow?start=28&length=13)

nit: rename to "DefaultBodyParameterName"

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -229,7 +236,50 @@ public static FormulaType GetReturnType(this OpenApiOperation op)
...
+            ContentType_ApplicationXml,            
+        };        
+
+        public static KeyValuePair<string, OpenApiMediaType> GetContentTypeAndSchema(this IDictionary<string, OpenApiMediaType> content)
+        {
+            List<KeyValuePair<string, OpenApiMediaType>> list = new ();
+
+            foreach (var ct in _knownContentTypes)
+            {
+                var kvp = content.FirstOrDefault(c => c.Key == ct);
```
MikeStall, line 263: >content.FirstOrDefault(c => c.Key == ct); [](http://example.com/codeflow?start=26&length=41)

avoid linear scan

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -229,7 +236,50 @@ public static FormulaType GetReturnType(this OpenApiOperation op)
...
+
+        private static readonly List<string> _knownContentTypes = new ()
+        {
+            ContentType_ApplicationJson,
+            ContentType_XWwwFormUrlEncoded,
+            ContentType_TextJson,
+            ContentType_ApplicationXml,            
+        };        
+
+        public static KeyValuePair<string, OpenApiMediaType> GetContentTypeAndSchema(this IDictionary<string, OpenApiMediaType> content)
```
MikeStall, line 257: >string [](http://example.com/codeflow?start=102&length=6)

For dictionaries of strings, we should comment what the KeyType is.

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -229,7 +236,50 @@ public static FormulaType GetReturnType(this OpenApiOperation op)
...
+
+                if (kvp.Key != null)
+                {
+                    if ((kvp.Key == ContentType_ApplicationJson && !kvp.Value.Schema.Properties.Any()) ||
+                        (kvp.Key == ContentType_TextJson && kvp.Value.Schema.Properties.Any()))
+                    {
+                        continue;
+                    }
+
+                    list.Add(kvp);
```
MikeStall, line 273: > list.Add(kvp); [](http://example.com/codeflow?start=19&length=15)

Can we just return here? do we need a List if we're just taking the first?

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -63,50 +59,97 @@ public HttpRequestMessage BuildRequest(FormulaValue[] args)
             // Header names are not case sensitive.
             // From RFC 2616 - "Hypertext Transfer Protocol -- HTTP/1.1", Section 4.2, "Message Headers"
             var headers = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+            string body = null;
```
MikeStall, line 62: >string [](http://example.com/codeflow?start=12&length=6)

This is fine now, but eventually we'll have non-string bodies.

LucGenetier: even if it's an image, it will be a string at the end
here, body variable stores the ready-for-network string as we call StringContent

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -63,50 +59,97 @@ public HttpRequestMessage BuildRequest(FormulaValue[] args)
...
+                                query.Append(param.Name);
+                                query.Append('=');
+                                query.Append(HttpUtility.UrlEncode(valueStr));
+                                break;
+
+                            case ParameterLocation.Header:
+                                headers.Add(param.Name, valueStr);
+                                break;                           
+
+                            case ParameterLocation.Cookie:
```
MikeStall, line 99: >case ParameterLocation.Cookie: [](http://example.com/codeflow?start=28&length=30)

nit ... we don't actually need this, it will be caught by default. 
But it's also a good reminder that we need to impl.

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -63,50 +59,97 @@ public HttpRequestMessage BuildRequest(FormulaValue[] args)
...
+            if (!string.IsNullOrEmpty(body))
+            {                
+                // $$$ Might need System.Net.Http.FormUrlEncodedContent()
+                request.Content = new StringContent(body, Encoding.Default, _argMapper.ContentType);                
+            }
+
             return request;
         }
 
+        private string GetBody(OpenApiParameter param, Dictionary<string, FormulaValue> map)
```
MikeStall, line 126: >string [](http://example.com/codeflow?start=16&length=6)

Should this just return the HttpContent directly? (that may scale as we move to non-string bodies)

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
LucGenetier: I know, I only use that as a reminder and make the code clearer

src/tests/Microsoft.PowerFx.Connectors.Tests/Helpers/LoggingTestServer.cs
```diff
@@ -61,6 +62,24 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
...
             }
 
+            var httpContent = request?.Content;
+            if (httpContent != null)
+            {
+                if (httpContent.Headers != null)
+                {
+                    foreach (var h in httpContent.Headers)
+                    {
+                        _log.AppendLine($" [header] {h.Key}: {string.Join(", ", h.Value)}");
```
MikeStall, line 72: >header [](http://example.com/codeflow?start=44&length=6)

super nit:  to a reader, this looks too much like a regular header. 
So at least, rename to "content-header" to avoid potential confusion.

src/libraries/Microsoft.PowerFx.Connectors/Execution/OpenApiJsonSerializer.cs
```diff
@@ -0,0 +1,201 @@
...
+using System.Dynamic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Text.Json;
+using Microsoft.OpenApi.Models;
+
+namespace Microsoft.PowerFx.Connectors.Execution
+{
+    internal class OpenApiJsonSerializer
```
MikeStall, line 16: >OpenApiJsonSerializer [](http://example.com/codeflow?start=19&length=21)

We should definitely add some direct unit tests for this class.  We can skip swagger.json and just new up the schema directly. 


Those unit tests should get 100% code-coverage on this file.

MikeStall: For reference, 
[https://github.com/microsoft/PowerApps-Language-Tooling/blob/master/src/PAModel/Utility/JsonNormalizer.cs](https://github.com/microsoft/PowerApps-Language-Tooling/blob/master/src/PAModel/Utility/JsonNormalizer.cs)

src/libraries/Microsoft.PowerFx.Connectors/Execution/OpenApiJsonSerializer.cs
```diff
@@ -0,0 +1,201 @@
...
+
+                    _writer.WriteEndArray();
+                    break;
+
+                case "null":
+                case "number":
+                case "boolean":
+                case "integer":
+                case "string":
+                    WriteProperty(property.Key, value);
```
MikeStall, line 87: >WriteProperty [](http://example.com/codeflow?start=20&length=13)

These shouldn't all merge into the same case. 

Ie, you can do directly, like:
case "boolean":
   _writer.WriteBoolean(name, b);
   break;

In other words, there shouldn't be any possible path where "string" type end could potentially call WriterNumber().

src/libraries/Microsoft.PowerFx.Connectors/Execution/OpenApiJsonSerializer.cs
```diff
@@ -0,0 +1,201 @@
...
+namespace Microsoft.PowerFx.Connectors.Execution
+{
+    internal class OpenApiJsonSerializer
+    {
+        private readonly Dictionary<string, object> _dic;
+        private readonly MemoryStream _stream;
+        private readonly Utf8JsonWriter _writer;
+        private readonly OpenApiSchema _schema;
+
+        internal OpenApiJsonSerializer(OpenApiSchema schema, Dictionary<string, object> dictionary)
```
MikeStall, line 23: >Dictionary<string, object> dictionary [](http://example.com/codeflow?start=61&length=37)

Just pass the FormulaValue directly. 
This lets you avoid FV.ToObject() and creating the intermediate objects.

src/libraries/Microsoft.PowerFx.Connectors/Execution/OpenApiJsonSerializer.cs
```diff
@@ -0,0 +1,201 @@
...
+using System.Linq;
+using System.Text;
+using System.Text.Json;
+using Microsoft.OpenApi.Models;
+
+namespace Microsoft.PowerFx.Connectors.Execution
+{
+    internal class OpenApiJsonSerializer
+    {
+        private readonly Dictionary<string, object> _dic;
```
MikeStall, line 18: > private readonly Dictionary<string, object> _dic; [](http://example.com/codeflow?start=7&length=50)

Pass as a parameter instead of as a field. This will be necessary for nested objects.

src/libraries/Microsoft.PowerFx.Connectors/Execution/OpenApiJsonSerializer.cs
```diff
@@ -0,0 +1,201 @@
...
+        private readonly MemoryStream _stream;
+        private readonly Utf8JsonWriter _writer;
+        private readonly OpenApiSchema _schema;
```
MikeStall, line 21: >private readonly OpenApiSchema _schema; [](http://example.com/codeflow?start=8&length=39)

This also can be  a parameter and not a field.

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -63,50 +59,97 @@ public HttpRequestMessage BuildRequest(FormulaValue[] args)
...
+            foreach (var property in param.Schema.Properties)
+            {
+                if (map.TryGetValue(property.Key, out var paramValue))
+                {
+                    props.Add(property.Key, paramValue.ToObject());
+                }
+            }
+
+            return props;
+        }
```
MikeStall, line 151: We should be able to remove this method altogether by just passing the FormulaValue directly into the FV-->json serializer.

src/libraries/Microsoft.PowerFx.Connectors/Execution/OpenApiFormUrlEncoder.cs
```diff
@@ -0,0 +1,194 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.Dynamic;
+using System.Linq;
+using System.Text;
+using Microsoft.OpenApi.Models;
+
+namespace Microsoft.PowerFx.Connectors.Execution
+{
+    internal class OpenApiFormUrlEncoder
```
MikeStall, line 14: >OpenApiFormUrlEncoder [](http://example.com/codeflow?start=19&length=21)

Is there an existing generic method to convert Json to FormUrl? If so, we could just use that.

src/libraries/Microsoft.PowerFx.Connectors/Execution/OpenApiJsonSerializer.cs
```diff
@@ -0,0 +1,201 @@
...
+
+        internal string ToJson()
+        {
+            _writer.WriteStartObject();
+            foreach (var prop in _schema.Properties)
+            {
+                WriteProperty(prop);
+            }
+
+            _writer.WriteEndObject();
```
MikeStall, line 39: This case can be merged with the case "object" below. 
Ie like this: , [https://github.com/microsoft/PowerApps-Language-Tooling/blob/dae85de910ebff06e7a5b9b8e10c50b322e8400b/src/PAModel/Utility/JsonNormalizer.cs#L39](https://github.com/microsoft/PowerApps-Language-Tooling/blob/dae85de910ebff06e7a5b9b8e10c50b322e8400b/src/PAModel/Utility/JsonNormalizer.cs#L39)

src/libraries/Microsoft.PowerFx.Connectors/Execution/OpenApiJsonSerializer.cs
```diff
@@ -0,0 +1,201 @@
...
+                case "integer":
+                case "string":
+                    WriteProperty(property.Key, value);
+                    break;
+
+                case "object":
+                    _writer.WriteStartObject(property.Key);
+                    foreach (var prop in property.Value.Properties)
+                    {
+                        WriteProperty(prop, value);
```
MikeStall, line 94: >WriteProperty [](http://example.com/codeflow?start=24&length=13)

See: 

[https://github.com/microsoft/PowerApps-Language-Tooling/blob/dae85de910ebff06e7a5b9b8e10c50b322e8400b/src/PAModel/Utility/JsonNormalizer.cs#L51](https://github.com/microsoft/PowerApps-Language-Tooling/blob/dae85de910ebff06e7a5b9b8e10c50b322e8400b/src/PAModel/Utility/JsonNormalizer.cs#L51)

src/libraries/Microsoft.PowerFx.Connectors/Execution/OpenApiJsonSerializer.cs
```diff
@@ -0,0 +1,201 @@
...
+                case "number":
+                case "boolean":
+                case "integer":
+                case "string":
+                    WriteProperty(property.Key, value);
+                    break;
+
+                case "object":
+                    _writer.WriteStartObject(property.Key);
+                    foreach (var prop in property.Value.Properties)
```
MikeStall, line 92: >property.Value.Properties [](http://example.com/codeflow?start=41&length=25)

This is another good example of why we want the schema in seriazliation :
you can be sure to only include properties actually specified in the schema.   A good test case would be where the FormulaValue has extra properties not included in the schema - we should not be serializing them.

src/libraries/Microsoft.PowerFx.Connectors/Execution/OpenApiJsonSerializer.cs
```diff
@@ -0,0 +1,201 @@
...
+                if (outerValue != null && outerValue is ExpandoObject eo)
+                {
+                    value = eo.FirstOrDefault(kvp => kvp.Key == property.Key);
+
+                    if (value != null && value is KeyValuePair<string, object> kvp)
+                    {
+                        value = kvp.Value;
+                    }
+                }
+            }
```
MikeStall, line 58: once we pass in FormulaValue directly, this can be removed.

src/libraries/Microsoft.PowerFx.Connectors/Execution/OpenApiJsonSerializer.cs
LucGenetier: Pending

src/libraries/Microsoft.PowerFx.Connectors/Execution/OpenApiJsonSerializer.cs
LucGenetier: I don't understand

src/libraries/Microsoft.PowerFx.Connectors/Execution/OpenApiJsonSerializer.cs
LucGenetier: I think this would add complexity as the string here (null/number/boolean...) is the schema type and not the FormulaValue inner type. Some strings can "hide" a date for example. See [https://swagger.io/docs/specification/data-models/data-types/](https://swagger.io/docs/specification/data-models/data-types/)

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
LucGenetier: done, except for XML where the work isn't finished yet

LucGenetier: I added a comment

src/libraries/Microsoft.PowerFx.Connectors/Execution/OpenApiFormUrlEncoder.cs
LucGenetier: Haven't found any
Anyway, we need FormulaValue array -> URL encoded string

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -63,50 +59,97 @@ public HttpRequestMessage BuildRequest(FormulaValue[] args)
...
-                            query.Append(HttpUtility.UrlEncode(valueStr));
-                            break;
-
-                        case ParameterLocation.Header:
-                            headers.Add(param.Name, valueStr);
-                            break;
-
-                        default:
-                            throw new NotImplementedException($"{param.In}");
+                        // No schema, we send the raw value
```
MikeStall, line 77: >/ No schema, we send the raw value [](http://example.com/codeflow?start=25&length=34)

What's an example that hits this?
I'm wary that we have 2 ways of getting the body (line 69, line 78)

src/libraries/Microsoft.PowerFx.Connectors/Execution/OpenApiJsonSerializer.cs
```diff
@@ -0,0 +1,100 @@
...
+            _writer = new Utf8JsonWriter(_stream, new JsonWriterOptions());
+        }
+
+        internal override string GetResult()
+        {
+            _writer.Flush();
+            return Encoding.UTF8.GetString(_stream.ToArray());
+        }
+
+        internal override void WritePropertyName(string name)
```
MikeStall, line 30: >internal [](http://example.com/codeflow?start=8&length=8)

Most of these should be 'private'

src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
```diff
@@ -0,0 +1,150 @@
...
+using Microsoft.OpenApi.Models;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Connectors.Execution
+{
+    internal abstract class FormulaValueSerializer
+    {
+        internal abstract string GetResult();
+
+        internal abstract void StartObject(string name = null);
```
MikeStall, line 17: >internal [](http://example.com/codeflow?start=8&length=8)

protected?

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
LucGenetier: [InlineData(3, @"Test.PostWeatherWithId({body: 5})", "POST [http://localhost:5000/weatherPost\r\n](http://localhost:5000/weatherPost\r\n) [content-header] Content-Type: text/json; charset=utf-8\r\n [body] 5")]

src/libraries/Microsoft.PowerFx.Connectors/Execution/OpenApiJsonSerializer.cs
```diff
@@ -0,0 +1,107 @@
...
+
+        protected override void StartObject(string name = null)
+        {
+            if (string.IsNullOrEmpty(name))
+            {
+                _writer.WriteStartObject();
+            }
+            else
+            {
+                _writer.WriteStartObject(JsonEncodedText.Encode(name));
```
MikeStall, line 71: >(JsonEncodedText.Encode [](http://example.com/codeflow?start=40&length=23)

Does WriteStartObject do the encoding for us? 
(add a test...)

src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
```diff
@@ -0,0 +1,188 @@
...
+            foreach (var property in schema.Properties)
+            {
+                var namedValue = fields.FirstOrDefault(nv => nv.Name.Equals(property.Key, StringComparison.OrdinalIgnoreCase));
+
+                if (namedValue == null)
+                {
+                    throw new NotImplementedException($"Missing property {property.Key}, object is too complex or not supported");
+                }
+
+                WriteProperty(property.Key, property.Value, namedValue.Value);
```
MikeStall, line 65: WritePropertyName should be here (lifted up, outside the recursion). 

[https://github.com/microsoft/PowerApps-Language-Tooling/blob/master/src/PAModel/Utility/JsonNormalizer.cs](https://github.com/microsoft/PowerApps-Language-Tooling/blob/master/src/PAModel/Utility/JsonNormalizer.cs)

MikeStall: this would also allow us to handle top-level cases that aren't objects.

src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
```diff
@@ -0,0 +1,188 @@
...
+
+        protected abstract void WriteDateTimeValue(DateTime dateTimeValue);
+
+        internal static readonly DType DType_Table = new (DKind.Table);
+        internal static readonly DType DType_Record = new (DKind.Record);
+
+        internal string Serialize(OpenApiSchema schema, IEnumerable<NamedValue> fields)
+        {
+            StartSerialization(schema);
+            WriteObject(null, schema, fields);
```
MikeStall, line 48: >WriteObject [](http://example.com/codeflow?start=12&length=11)

Could be WriteValue()

src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
```diff
@@ -0,0 +1,188 @@
...
+                case "number":
+                    // float, double
+                    AssertType(DType.Number, fv, propertyName);
+                    WritePropertyValue(propertyName, fv);
+                    break;
+
+                case "boolean":
+                    AssertType(DType.Boolean, fv, propertyName);
+                    WritePropertyValue(propertyName, fv);
+                    break;
```
MikeStall, line 114: case "boolean": 

if (value is BooleanValue booleanValue)
            {
                WriteBooleanValue(booleanValue.Value);                
            }  else {error!!!}

src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
```diff
@@ -0,0 +1,188 @@
...
+        protected abstract void WriteNumberValue(double numberValue);
+
+        protected abstract void WriteStringValue(string stringValue);
+
+        protected abstract void WriteBooleanValue(bool booleanValue);
+
+        protected abstract void WriteDateTimeValue(DateTime dateTimeValue);
+
+        internal static readonly DType DType_Table = new (DKind.Table);
+        internal static readonly DType DType_Record = new (DKind.Record);
```
MikeStall, line 43: that means we could also get rid of these Dtype usage, and just use FormulaType

src/libraries/Microsoft.PowerFx.Connectors/Execution/OpenApiXmlSerializer.cs
```diff
@@ -0,0 +1,106 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Xml.Linq;
+using Microsoft.OpenApi.Models;
+
+namespace Microsoft.PowerFx.Connectors.Execution
+{
+    internal class OpenApiXmlSerializer : FormulaValueSerializer
```
MikeStall, line 12: >OpenApiXmlSerializer [](http://example.com/codeflow?start=19&length=20)

We can remove this for now, and add back in once we have the connectors.

src/tests/Microsoft.PowerFx.Connectors.Tests/OpenApiJsonSerializerTests.cs
```diff
@@ -0,0 +1,144 @@
...
+            var fields = new List<NamedValue>() { new NamedValue("a", table) };
+
+            var ex = Assert.Throws<ArgumentException>(() => jsonSerializer.Serialize(schema, fields));
+            Assert.Equal("Incompatible Table for supporting array, RecordValue has more than one column - propertyName a, number of fields 2", ex.Message);
+        }
+
+        [Fact]
+        public void JsonSerializer_Date()
+        {
+            var now = DateTime.Now;
```
MikeStall, line 135: >DateTime.Now; [](http://example.com/codeflow?start=22&length=13)

use a constant date

src/tests/Microsoft.PowerFx.Connectors.Tests/OpenApiJsonSerializerTests.cs
```diff
@@ -0,0 +1,144 @@
...
+            var str = jsonSerializer.Serialize(schema, fields);
+            Assert.Equal(@"{""a"":true,""b"":false}", str);
+        }
+
+        [Fact]
+        public void JsonSerializer_Array()
+        {
+            var jsonSerializer = new OpenApiJsonSerializer();
+            var schema = new OpenApiSchema() { Properties = new Dictionary<string, OpenApiSchema>() { ["a"] = new OpenApiSchema() { Type = "array", Format = "int" } } };                            
+            var fields = new List<NamedValue>() { new NamedValue("a", FormulaValue.NewSingleColumnTable(FormulaValue.New(1), FormulaValue.New(2))) };
```
MikeStall, line 113: >NewSingleColumnTable [](http://example.com/codeflow?start=83&length=20)

This is good! Showing that a single-column-table serializes as an array (and no extra "Value" field shows up)

src/tests/Microsoft.PowerFx.Connectors.Tests/OpenApiJsonSerializerTests.cs
```diff
@@ -0,0 +1,144 @@
...
+
+            var str = jsonSerializer.Serialize(schema, fields);
+            Assert.Equal(@"{""a"":""abc""}", str);
+        }
+
+        [Fact]
+        public void JsonSerializer_Bool()
+        {
+            var jsonSerializer = new OpenApiJsonSerializer();
+            var schema = new OpenApiSchema() { Properties = new Dictionary<string, OpenApiSchema>() { ["a"] = new OpenApiSchema() { Type = "boolean" }, ["b"] = new OpenApiSchema() { Type = "boolean" } } };
```
MikeStall, line 101: Helper functions?
var schema = new OpenApiSchema()
.AddBoolean("a")
.AddBoolean("b");

src/tests/Microsoft.PowerFx.Connectors.Tests/OpenApiJsonSerializerTests.cs
```diff
@@ -0,0 +1,144 @@
...
+            Assert.Equal(@"{""a"":[1,2]}", str);
+        }
+
+        [Fact]
+        public void JsonSerializer_Array_Invalid()
+        {
+            var jsonSerializer = new OpenApiJsonSerializer();
+            var schema = new OpenApiSchema() { Properties = new Dictionary<string, OpenApiSchema>() { ["a"] = new OpenApiSchema() { Type = "array", Format = "int" } } };
+            var record = FormulaValue.NewRecordFromFields(new NamedValue("a", FormulaValue.New(1)), new NamedValue("b", FormulaValue.New("foo")));
+            var table = new InMemoryTableValue(IRContext.NotInSource(TableType.FromRecord(record.Type)), new List<RecordValue>() { record }.Select(r => DValue<RecordValue>.Of(r)));
```
MikeStall, line 125: >new InMemoryTableValue( [](http://example.com/codeflow?start=24&length=23)

FormulaValue.NewTable()

src/tests/Microsoft.PowerFx.Connectors.Tests/OpenApiJsonSerializerTests.cs
```diff
@@ -0,0 +1,144 @@
...
+        {
+            var now = DateTime.Now;
+            var jsonSerializer = new OpenApiJsonSerializer();
+            var schema = new OpenApiSchema() { Properties = new Dictionary<string, OpenApiSchema>() { ["a"] = new OpenApiSchema() { Type = "string", Format = "date-time" } } };
+            var fields = new List<NamedValue>() { new NamedValue("a", FormulaValue.New(now)) };
+
+            var str = jsonSerializer.Serialize(schema, fields);
+            Assert.Equal($@"{{""a"":""{now.ToString("o", CultureInfo.InvariantCulture).Replace("+", "\u002B")}""}}", str);
+        }
+    }
```
MikeStall, line 143: ideas:

- Formula Value has extra property.  (should not be sent) 
- missing / blank property
- nested objects.

src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
```diff
@@ -0,0 +1,209 @@
...
+
+        protected abstract void WriteNumberValue(double numberValue);
+
+        protected abstract void WriteStringValue(string stringValue);
+
+        protected abstract void WriteBooleanValue(bool booleanValue);
+
+        protected abstract void WriteDateTimeValue(DateTime dateTimeValue);
+
+        protected bool _schemaLessBody;
```
MikeStall, line 42: >protected [](http://example.com/codeflow?start=8&length=9)

nit: readonly. 
(we don't want derived class changing it) 

Although ... why do we need this?  (a comment would be good). 
Does this mean we don't have a OpenApiSchema schema parameter?

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
MikeStall: why ByteArrayContent instead of HttpContent?

src/libraries/Microsoft.PowerFx.Core/AssemblyProperties.cs
```diff
@@ -36,5 +36,6 @@
...
 [assembly: InternalsVisibleTo("Microsoft.PowerFx.Interpreter.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
+[assembly: InternalsVisibleTo("Microsoft.PowerFx.Connectors.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
```
MikeStall, line 39: Can we avoid this?

PR #432, opened 2022-06-15
Example Base64 Encode/Decode functions

bryancroteau-MSFT:
Example Base64 encode/decode string -> string functions.


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Base64.txt
```diff
@@ -0,0 +1,17 @@
+﻿>> Base64Encode("The quick brown fox jumps over the lazy dog")
```
LucGenetier, line 1: I would add non [a-z] characters

PR #435, opened 2022-06-20
Add resource string for named formula name conflicts

VamsiModem:


src/strings/PowerFxResources.en-US.resx
```diff
@@ -6436,6 +6436,9 @@
   <data name="ErrNamedFormula_AlreadyDefined" xml:space="preserve">
     <value>NamedFormula '{0}' already exists.</value>
   </data>
+  <data name="ErrNamedFormula_NameConflict" xml:space="preserve">
+    <value>NamedFormula '{0}' cannot conflict with any existing named formula, control, screen, collection, global variable or data source names.</value>
+  </data>
```
MikeStall, line 6441: Is this symmetric - what error do I get if I try to create a screen that conflicts with an existing named formula?

VamsiModem: Yeah, we are using a similar message for global variables

ErrNeedValidVariableName_Arg - The first argument of '{0}' should be a valid variable name, and cannot conflict with any existing control, screen, collection, or data source names. Found type '{1}'

If you try to create a screen with an existing named formula, you we see a popup saying 'This name is already being used. please try a different name.

MikeStall (approved):

PR #436, opened 2022-06-22
Connector - Increase code coverage and fixes

LucGenetier:


src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
```diff
@@ -92,6 +92,11 @@ private void WriteProperty(string propertyName, OpenApiSchema propertySchema, Fo
                             throw new ArgumentException($"Incompatible Table for supporting array, RecordValue has more than one column - propertyName {propertyName}, number of fields {rva.Fields.Count()}");
                         }
 
+                        if (rva.Fields.First().Name != "Value")
```
MikeStall, line 95: >"Value" [](http://example.com/codeflow?start=55&length=7)

we can use a constant: TableValue.ValueName

MikeStall (approved):
:shipit:

PR #437, opened 2022-06-22
Preliminary support for user-defined functions in RecalcEngine

MalekiRe:
also added some test cases for it in Recursion.txt, currently the limit is at 20.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs
@@ -99,7 +99,7 @@ namespace Microsoft.PowerFx.Core.Glue
             return false;
         }
 
-        public IEnumerable<TexlFunction> LookupFunctions(DPath theNamespace, string name, bool localeInvariant = false)
+        public virtual IEnumerable<TexlFunction> LookupFunctions(DPath theNamespace, string name, bool localeInvariant = false)
         {
             Contracts.Check(theNamespace.IsValid, "The namespace is invalid.");
             Contracts.CheckNonEmpty(name, "name");
--- a/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
@@ -170,9 +170,14 @@ namespace Microsoft.PowerFx
                 var result = await asyncFunc.InvokeAsync(args, _cancel);
                 return result;
             }
-            else if (func is CustomTexlFunction customFunc)
+            else if (func is UserDefinedTexlFunction udtf)
             {
-                var result = customFunc.Invoke(args);
+                var result = await udtf.InvokeAsync(args, _cancel, context.StackDepthCounter.Increment());
+                return result;
+            }
+            else if (func is CustomTexlFunction customTexlFunc)
+            {
+                var result = customTexlFunc.Invoke(args);
                 return result;
             }
             else
@@ -182,7 +187,6 @@ namespace Microsoft.PowerFx
                     var result = await ptr(this, context.IncrementStackDepthCounter(childContext), node.IRContext, args);
 
                     Contract.Assert(result.IRContext.ResultType == node.IRContext.ResultType || result is ErrorValue || result.IRContext.ResultType is BlankType);
-
                     return result;
                 }
 
--- a/src/libraries/Microsoft.PowerFx.Interpreter/ParsedExpression.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/ParsedExpression.cs
@@ -39,5 +39,14 @@ namespace Microsoft.PowerFx
                 return maxCallDepthException.ToErrorValue(_irnode.IRContext);
             }
         }
+
+        internal async Task<FormulaValue> EvalAsyncInternal(RecordValue parameters, CancellationToken cancel, StackDepthCounter stackMarker)
+        {
+            // We don't catch the max call depth exception here becuase someone could swallow the error with an "IfError" check.
+            // Instead we only catch at the top of parsed expression, which is the above function.
+            var ev2 = new EvalVisitor(_cultureInfo, cancel);
+            var newValue = await _irnode.Accept(ev2, new EvalVisitorContext(SymbolContext.NewTopScope(_topScopeSymbol, parameters), stackMarker));
+            return newValue;
+        }
     }
 }
--- a/src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
@@ -194,13 +194,7 @@ namespace Microsoft.PowerFx
             var func = new UserDefinedTexlFunction(definition.Name, definition.ReturnType, definition.Parameters, check);
             if (_customFuncs.ContainsKey(definition.Name))
             {
-                return new UDFLazyBinder(
-                    new ExpressionError()
-                {
-                    Message = "Function name already defined",
-                    Span = null,
-                    Kind = ErrorKind.Internal, //Todo: change this out for a new kind of ErrorKind error.
-                }, definition.Name);
+                throw new InvalidOperationException($"function {definition.Name} is already defined");
             }
 
             _customFuncs[definition.Name] = func;
--- a/src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
@@ -3,16 +3,21 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.PowerFx.Core.Binding;
+using Microsoft.PowerFx.Core.Functions;
 using Microsoft.PowerFx.Core.Glue;
 using Microsoft.PowerFx.Core.IR;
 using Microsoft.PowerFx.Core.Texl;
+using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
 using Microsoft.PowerFx.Functions;
 using Microsoft.PowerFx.Interpreter;
+using Microsoft.PowerFx.Interpreter.UDF;
 using Microsoft.PowerFx.Types;
+using static Microsoft.PowerFx.Interpreter.UDFHelper;
 
 namespace Microsoft.PowerFx
 {
@@ -22,6 +27,8 @@ namespace Microsoft.PowerFx
     public sealed class RecalcEngine : Engine, IScope, IPowerFxEngine
     {
         internal Dictionary<string, RecalcFormulaInfo> Formulas { get; } = new Dictionary<string, RecalcFormulaInfo>();
+
+        internal Dictionary<string, TexlFunction> _customFuncs = new Dictionary<string, TexlFunction>();
 
         /// <summary>
         /// Initializes a new instance of the <see cref="RecalcEngine"/> class.
@@ -167,6 +174,83 @@ namespace Microsoft.PowerFx
             var check = Check(expressionText, (RecordType)parameters.IRContext.ResultType, options);
             check.ThrowOnErrors();
             return await check.Expression.EvalAsync(parameters, cancel);
+        }
+
+        /// <summary>
+        /// For private use because we don't want anyone defining a function without binding it.
+        /// </summary>
+        /// <returns></returns>
+        private UDFLazyBinder DefineFunction(UDFDefinition definition)
+        {
+            // $$$ Would be a good helper function 
+            var record = RecordType.Empty();
+            foreach (var p in definition.Parameters)
+            {
+                record = record.Add(p);
+            }
+
+            var check = new CheckWrapper(this, definition.Body, record);
+
+            var func = new UserDefinedTexlFunction(definition.Name, definition.ReturnType, definition.Parameters, check);
+            if (_customFuncs.ContainsKey(definition.Name))
+            {
+                return new UDFLazyBinder(
+                    new ExpressionError()
+                {
+                    Message = "Function name already defined",
+                    Span = null,
+                    Kind = ErrorKind.Internal, //Todo: change this out for a new kind of ErrorKind error.
+                }, definition.Name);
+            }
+
+            _customFuncs[definition.Name] = func;
+            return new UDFLazyBinder(func, definition.Name);
+        }
+
+        private void RemoveFunction(string name)
+        {
+            _customFuncs.Remove(name);
+        }
+
+        /// <summary>
+        /// Tries to define and bind all the functions here. If any function names conflict returns an expression error. 
+        /// Also returns any errors from binding failing. All functions defined here are removed if any of them contain errors.
+        /// </summary>
+        /// <param name="udfDefinitions"></param>
+        /// <returns></returns>
+        internal IEnumerable<ExpressionError> DefineFunctions(IEnumerable<UDFDefinition> udfDefinitions)
+        {
+            var expressionErrors = new List<ExpressionError>();
+
+            var binders = new List<UDFLazyBinder>();
+            foreach (UDFDefinition definition in udfDefinitions)
+            {
+                binders.Add(DefineFunction(definition));
+            }
+            
+            foreach (UDFLazyBinder lazyBinder in binders)
+            {
+                var possibleErrors = lazyBinder.Bind();
+                if (possibleErrors.Any())
+                {
+                    expressionErrors.AddRange(possibleErrors);
+                }
+            }
+
+            if (expressionErrors.Any())
+            {
+                foreach (UDFLazyBinder lazyBinder in binders)
+                {
+                    RemoveFunction(lazyBinder.Name);
+                }
+            }
+
+            return expressionErrors;
+        }
+
+        internal IEnumerable<ExpressionError> DefineFunctions(params UDFDefinition[] udfDefinitions)
+        {
+            return DefineFunctions(udfDefinitions.AsEnumerable());
         }
 
         // Invoke onUpdate() each time this formula is changed, passing in the new value. 
--- a/src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngineResolver.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngineResolver.cs
@@ -6,6 +6,7 @@ using System.Collections.ObjectModel;
 using System.Linq;
 using Microsoft.PowerFx.Core.Binding;
 using Microsoft.PowerFx.Core.Binding.BindInfo;
+using Microsoft.PowerFx.Core.Functions;
 using Microsoft.PowerFx.Core.Glue;
 using Microsoft.PowerFx.Core.Utils;
 
@@ -31,6 +32,19 @@ namespace Microsoft.PowerFx
         {
             _parent = parent;
             _powerFxConfig = powerFxConfig;
+        }
+
+        public override IEnumerable<TexlFunction> LookupFunctions(DPath theNamespace, string name, bool localeInvariant = false)
+        {
+            if (theNamespace.IsRoot)
+            {
+                if (_parent._customFuncs.TryGetValue(name, out var func))
+                {
+                    return new TexlFunction[] { func };
+                }
+            }
+
+            return base.LookupFunctions(theNamespace, name, localeInvariant);
         }
 
         public override bool Lookup(DName name, out NameLookupInfo nameInfo, NameLookupPreferences preferences = NameLookupPreferences.None)
--- a/src/libraries/Microsoft.PowerFx.Interpreter/RuntimeMaxCallDepthException.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/RuntimeMaxCallDepthException.cs
@@ -1,0 +1,35 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.PowerFx.Interpreter
+{
+    [Serializable]
+    internal class RuntimeMaxCallDepthException : Exception
+    {
+        public RuntimeMaxCallDepthException()
+            : base()
+        {
+        }
+
+        public RuntimeMaxCallDepthException(string message)
+            : base(message)
+        { 
+        }
+
+        public RuntimeMaxCallDepthException(string message, Exception inner)
+            : base(message, inner)
+        {
+        }
+
+        protected RuntimeMaxCallDepthException(
+            System.Runtime.Serialization.SerializationInfo info,
+            System.Runtime.Serialization.StreamingContext context)
+            : base(info, context) 
+        { 
+        }
+    }
+}
--- a/src/libraries/Microsoft.PowerFx.Interpreter/UDF/CheckWrapper.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/UDF/CheckWrapper.cs
@@ -1,0 +1,26 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Interpreter.UDF
+{
+    internal class CheckWrapper
+    {
+        private readonly string _expressionText;
+        private readonly RecordType _parameterType;
+        private readonly RecalcEngine _engine;
+
+        public CheckWrapper(RecalcEngine engine, string expressionText, RecordType parameterType = null)
+        {
+            _engine = engine;
+            _expressionText = expressionText;
+            _parameterType = parameterType;
+        }
+
+        public CheckResult Get() => _engine.Check(_expressionText, _parameterType);
+    }
+}
--- a/src/libraries/Microsoft.PowerFx.Interpreter/UDF/CheckWrapper.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/UDF/CheckWrapper.cs
@@ -8,6 +8,9 @@ using Microsoft.PowerFx.Types;
 
 namespace Microsoft.PowerFx.Interpreter.UDF
 {
+     /// <summary>
+     /// CheckWrapper delays the evaluation of the body to the Get call while taking in all the parameters needed to make the Check call.
+     /// </summary>
     internal class CheckWrapper
     {
         private readonly string _expressionText;
--- a/src/libraries/Microsoft.PowerFx.Interpreter/UDF/UDFHelper.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/UDF/UDFHelper.cs
@@ -1,0 +1,29 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Interpreter
+{
+    internal class UDFHelper
+    {
+        internal static IEnumerable<NamedValue> Zip(NamedFormulaType[] parameters, FormulaValue[] args)
+        {
+            if (parameters.Length != args.Length)
+            {
+                throw new ArgumentException();
+            }
+
+            var result = new NamedValue[args.Length];
+            for (var i = 0; i < args.Length; i++)
+            {
+                result[i] = new NamedValue(parameters[i].Name, args[i]);
+            }
+
+            return result;
+        }
+    }
+}
--- a/src/libraries/Microsoft.PowerFx.Interpreter/UDF/UDFHelper.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/UDF/UDFHelper.cs
@@ -8,7 +8,7 @@ using Microsoft.PowerFx.Types;
 
 namespace Microsoft.PowerFx.Interpreter
 {
-    internal class UDFHelper
+    internal static class UDFHelper
     {
         internal static IEnumerable<NamedValue> Zip(NamedFormulaType[] parameters, FormulaValue[] args)
         {
--- a/src/libraries/Microsoft.PowerFx.Interpreter/UDF/UDFLazyBinder.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/UDF/UDFLazyBinder.cs
@@ -1,0 +1,40 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.PowerFx.Interpreter
+{
+    internal class UDFLazyBinder
+    {
+        private readonly UserDefinedTexlFunction _function;
+        
+        private readonly List<ExpressionError> _expressionError = new List<ExpressionError>();
+
+        public readonly string Name;
+
+        public UDFLazyBinder(UserDefinedTexlFunction texlFunction, string name)
+        { 
+            _function = texlFunction;
+            Name = name;
+        }
+
+        // Used when trying to define a function, there is an error before binding, such as the function name already being defined.
+        public UDFLazyBinder(ExpressionError expressionError, string name)
+        {
+            _expressionError.Add(expressionError);
+            Name = name;
+        }
+
+        public IEnumerable<ExpressionError> Bind()
+        {
+            if (_expressionError.Any())
+            {
+                return _expressionError;
+            }
+
+            return _function.Bind();
+        } 
+    }
+}
--- a/src/tests/Microsoft.PowerFx.Interpreter.Tests/Helpers/AsyncFunctionHelperBase.cs
+++ b/src/tests/Microsoft.PowerFx.Interpreter.Tests/Helpers/AsyncFunctionHelperBase.cs
@@ -8,7 +8,6 @@ using System.Threading.Tasks;
 using Microsoft.PowerFx.Core.Functions;
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
-using Microsoft.PowerFx.Interpreter;
 using Microsoft.PowerFx.Types;
 
 namespace Microsoft.PowerFx.Tests
--- a/src/tests/Microsoft.PowerFx.Interpreter.Tests/Helpers/AsyncFunctionHelperBase.cs
+++ b/src/tests/Microsoft.PowerFx.Interpreter.Tests/Helpers/AsyncFunctionHelperBase.cs
@@ -8,6 +8,7 @@ using System.Threading.Tasks;
 using Microsoft.PowerFx.Core.Functions;
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Interpreter;
 using Microsoft.PowerFx.Types;
 
 namespace Microsoft.PowerFx.Tests
--- a/src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineAsyncTests.cs
+++ b/src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineAsyncTests.cs
@@ -10,7 +10,6 @@ using Microsoft.PowerFx.Core.Localization;
 using Microsoft.PowerFx.Core.Tests;
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
-using Microsoft.PowerFx.Interpreter;
 using Microsoft.PowerFx.Types;
 using Xunit;
 using static Microsoft.PowerFx.Core.Localization.TexlStrings;
--- a/src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineAsyncTests.cs
+++ b/src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineAsyncTests.cs
@@ -10,6 +10,7 @@ using Microsoft.PowerFx.Core.Localization;
 using Microsoft.PowerFx.Core.Tests;
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Interpreter;
 using Microsoft.PowerFx.Types;
 using Xunit;
 using static Microsoft.PowerFx.Core.Localization.TexlStrings;
--- a/src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
+++ b/src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
@@ -9,6 +9,8 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.PowerFx.Core;
 using Microsoft.PowerFx.Core.Functions;
+using Microsoft.PowerFx.Core.Localization;
+using Microsoft.PowerFx.Core.Parser;
 using Microsoft.PowerFx.Core.Tests;
 using Microsoft.PowerFx.Core.Texl;
 using Microsoft.PowerFx.Core.Types.Enums;
@@ -17,6 +19,7 @@ using Microsoft.PowerFx.Interpreter;
 using Microsoft.PowerFx.Types;
 using Xunit;
 using Xunit.Sdk;
+using static Microsoft.PowerFx.Interpreter.UDFHelper;
 
 namespace Microsoft.PowerFx.Tests
 {
@@ -237,6 +240,131 @@ namespace Microsoft.PowerFx.Tests
             // Can't redefine an existing formula. 
             Assert.Throws<InvalidOperationException>(() =>
                engine.SetFormula("A", "3", OnUpdate));
+        }
+
+        [Fact]
+        public void DefFunc()
+        {
+            var config = new PowerFxConfig(null);
+            var recalcEngine = new RecalcEngine(config);
+
+            IEnumerable<ExpressionError> enumerable = recalcEngine.DefineFunctions(
+            new UDFDefinition(
+                "foo",
+                "x * y", 
+                FormulaType.Number, 
+                new NamedFormulaType("x", FormulaType.Number),
+                new NamedFormulaType("y", FormulaType.Number)));
+            Assert.False(enumerable.Any());
+            Assert.Equal(17.0, recalcEngine.Eval("foo(3,4) + 5").ToObject());
+        }
+
+        [Fact]
+        public void DefRecursiveFunc()
+        {
+            var config = new PowerFxConfig(null);
+            var recalcEngine = new RecalcEngine(config);
+            var body = @"If(x=0,foo(1),If(x=1,foo(2),If(x=2,2)))";
+            IEnumerable<ExpressionError> enumerable = recalcEngine.DefineFunctions(
+                new UDFDefinition(
+                    "foo",
+                    body,
+                    FormulaType.Number,
+                    new NamedFormulaType("x", FormulaType.Number)));
+            var result = recalcEngine.Eval("foo(0)");
+            Assert.Equal(2.0, result.ToObject());
+            Assert.False(enumerable.Any());
+        }
+
+        [Fact]
+        public void DefSimpleRecursiveFunc()
+        {
+            var config = new PowerFxConfig(null);
+            var recalcEngine = new RecalcEngine(config);
+            Assert.False(recalcEngine.DefineFunctions(
+                new UDFDefinition(
+                    "foo",
+                    "foo()",
+                    FormulaType.Blank)).Any());
+            var result = recalcEngine.Eval("foo()");
+            Assert.IsType<ErrorValue>(result);
+        }
+
+        [Fact]
+        public void DefHailstoneSequence()
+        {
+            var config = new PowerFxConfig(null)
+            {
+                MaxCallDepth = 100
+            };
+            var recalcEngine = new RecalcEngine(config);
+            var body = @"If(Not(x = 1), If(Mod(x, 2)=0, hailstone(x/2), hailstone(3*x+1)), x)";
+            var funcName = "hailstone";
+            var returnType = FormulaType.Number;
+            var variable = new NamedFormulaType("x", FormulaType.Number);
+
+            Assert.False(recalcEngine.DefineFunctions(
+                new UDFDefinition(funcName, body, returnType, variable)).Any());
+            Assert.Equal(1.0, recalcEngine.Eval("hailstone(192)").ToObject());
+        }
+
+        [Fact]
+        public void DefMutualRecursionFunc()
+        {
+            var config = new PowerFxConfig(null)
+            {
+                MaxCallDepth = 100
+            };
+            var recalcEngine = new RecalcEngine(config);
+            var bodyEven = @"If(number = 0, true, odd(Abs(number)-1))";
+            var bodyOdd = @"If(number = 0, false, even(Abs(number)-1))";
+
+            var udfOdd = new UDFDefinition(
+                "odd",
+                bodyOdd,
+                FormulaType.Boolean,
+                new NamedFormulaType("number", FormulaType.Number));
+            var udfEven = new UDFDefinition(
+                "even",
+                bodyEven,
+                FormulaType.Boolean,
+                new NamedFormulaType("number", FormulaType.Number));
+
+            Assert.False(recalcEngine.DefineFunctions(udfOdd, udfEven).Any());
+
+            Assert.Equal(true, recalcEngine.Eval("odd(17)").ToObject());
+            Assert.Equal(false, recalcEngine.Eval("even(17)").ToObject());
+        }
+
+        [Fact]
+        public async void RedefinitionError()
+        {
+            var config = new PowerFxConfig(null);
+            var recalcEngine = new RecalcEngine(config);
+            Assert.True(recalcEngine.DefineFunctions(
+                new UDFDefinition("foo", "foo()", FormulaType.Blank),
+                new UDFDefinition("foo", "x+1", FormulaType.Number)).Any());
+            await Assert.ThrowsAsync<InvalidOperationException>(() => recalcEngine.EvalAsync("foo()", CancellationToken.None));
+        }
+
+        [Fact]
+        public void UDFBodySyntaxErrorTest()
+        {
+            var config = new PowerFxConfig(null);
+            var recalcEngine = new RecalcEngine(config);
+            Assert.True(recalcEngine.DefineFunctions(new UDFDefinition("foo", "x[", FormulaType.Blank)).Any());
+        }
+
+        [Fact]
+        public async void UDFIncorrectParametersTest()
+        {
+            var config = new PowerFxConfig(null);
+            var recalcEngine = new RecalcEngine(config);
+            Assert.False(recalcEngine.DefineFunctions(new UDFDefinition("foo", "x+1", FormulaType.Number, new NamedFormulaType("x", FormulaType.Number))).Any());
+            Assert.False(recalcEngine.Check("foo(False)").IsSuccess);
+            Assert.False(recalcEngine.Check("foo(Table( { Value: \"Strawberry\" }, { Value: \"Vanilla\" } ))").IsSuccess);
+            Assert.True(recalcEngine.Check("foo(1)").IsSuccess);
+            await Assert.ThrowsAsync<InvalidOperationException>(async () => await recalcEngine.EvalAsync("foo(False)", CancellationToken.None));
         }
 
         [Fact]
--- a/src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
+++ b/src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
@@ -341,10 +341,9 @@ namespace Microsoft.PowerFx.Tests
         {
             var config = new PowerFxConfig(null);
             var recalcEngine = new RecalcEngine(config);
-            Assert.True(recalcEngine.DefineFunctions(
+            Assert.Throws<InvalidOperationException>(() => recalcEngine.DefineFunctions(
                 new UDFDefinition("foo", "foo()", FormulaType.Blank),
-                new UDFDefinition("foo", "x+1", FormulaType.Number)).Any());
-            await Assert.ThrowsAsync<InvalidOperationException>(() => recalcEngine.EvalAsync("foo()", CancellationToken.None));
+                new UDFDefinition("foo", "x+1", FormulaType.Number)));
         }
 
         [Fact]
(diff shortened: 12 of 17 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -153,6 +155,12 @@ private async Task<FormulaValue> TryHandleSet(CallNode node, SymbolContext conte
         public override async ValueTask<FormulaValue> Visit(CallNode node, SymbolContext context)
         {
             CheckCancel();
+
+            //Don't know if this is the best place to put this increment call.
+            if (!context.IncrementCallDepth(_maxRecursionDepth))
```
lesaltzm, line 160: So, this depth check is specific to UDFs with recursion (not impl yet). You'd want to increment it specifically when entering evaluation of a UDF.

lesaltzm:
Discussed:
The goal of the recursion depth limit would be specifically for UDFs, to avoid allowing authors to write themselves a stack overflow. Closing this PR, but the core is usable down the line.

MalekiRe:
This PR now includes the UDF functions code, not the parser/interpreter part, but just the functions to define functions, including recursive functions. After talking with Mike Stall he recommended keeping a recursion depth limit to prevent any user from overloading any PowerFx interpreter or server by spamming extremely large calls.

bryancroteau-MSFT (changes requested):

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -84,7 +90,109 @@ public static IExpression CreateEvaluatorDirect(CheckResult result)
...
+                _engine = engine;
+                _expressionText = expressionText;
+                _parameterType = parameterType;
+                _options = options;
+            }
+
+            public CheckResult Get() => _engine.Check(_expressionText, _parameterType, _options);
+        }
+
+        private class UDFTexlFunction : CustomTexlFunction, IAsyncTexlFunction
```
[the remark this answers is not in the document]
MalekiRe: Done

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -153,6 +155,12 @@ private async Task<FormulaValue> TryHandleSet(CallNode node, SymbolContext conte
         public override async ValueTask<FormulaValue> Visit(CallNode node, SymbolContext context)
         {
             CheckCancel();
+
+            //Don't know if this is the best place to put this increment call.
```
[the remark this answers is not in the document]
MalekiRe: Done

src/libraries/Microsoft.PowerFx.Interpreter/EvalContext.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                _mutex.ReleaseMutex();
+                return false;
+            }
+
+            _mutex.ReleaseMutex();
+            
+            // The following line is just debug to keep track of call depth. Its painful to have to google it everytime I test recursion so I'm leaving it here commented out.
+            // If this breaks some sort of convention I'll remove the comment.
+            //System.Diagnostics.Debug.WriteLine($"CurrentCallDepth: {_currentCallDepth}");
```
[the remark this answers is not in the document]
MalekiRe: Done

src/libraries/Microsoft.PowerFx.Interpreter/EvalContext.cs
```diff
@@ -0,0 +1,54 @@
...
+            _currentCallDepth++;
+            if (_currentCallDepth >= _maxCallDepth)
+            {
+                _mutex.ReleaseMutex();
+                return false;
+            }
+
+            _mutex.ReleaseMutex();
+            
+            // The following line is just debug to keep track of call depth. Its painful to have to google it everytime I test recursion so I'm leaving it here commented out.
```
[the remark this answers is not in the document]
MalekiRe: Done

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -24,6 +24,7 @@ public sealed class PowerFxConfig
         private readonly HashSet<TexlFunction> _extraFunctions = new HashSet<TexlFunction>();
         private readonly Dictionary<DName, IExternalEntity> _environmentSymbols;
         private DisplayNameProvider _environmentSymbolDisplayNameProvider;
+        private readonly int _maxCallDepth = 100;
```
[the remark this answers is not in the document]
MalekiRe: Now it's part of the constructor

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -3,24 +3,30 @@
...
+using Microsoft.PowerFx.Interpreter;
 using Microsoft.PowerFx.Types;
 
 namespace Microsoft.PowerFx
 {
     /// <summary>
     /// Holds a set of Power Fx variables and formulas. Formulas are recalculated when their dependent variables change.
     /// </summary>
-    public sealed class RecalcEngine : Engine, IScope, IPowerFxEngine
+    public sealed partial class RecalcEngine : Engine, IScope, IPowerFxEngine
```
lesaltzm, line 23: >partial [](http://example.com/codeflow?start=18&length=7)

Can we avoid partial classes? They can be a real maintenance headache.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -66,15 +72,16 @@ private protected override INameResolver CreateResolver(PowerFxConfig alternateC
...
         }
 
         /// <summary>
         /// Create an evaluator over the existing binding. 
         /// </summary>
         /// <param name="result">A successful binding from a previous call to <see cref="Engine.Check(string, RecordType, ParserOptions)"/>. </param>
+        /// <param name="maxCallDepth">Max number of function nesting calls.</param>
         /// <returns></returns>
-        public static IExpression CreateEvaluatorDirect(CheckResult result)
+        public static IExpression CreateEvaluatorDirect(CheckResult result, int maxCallDepth = 100)
```
lesaltzm, line 84: >int maxCallDepth = 100 [](http://example.com/codeflow?start=76&length=22)

Adding an optional param to a public method  is a breaking change, as we noted recently from the PROSE changes. Is this necessary, and if so, should CreateEvaluatorDirect just take a powerfxconfig with this set? Assuming we'd want to have other config items like this eventually.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -144,21 +168,35 @@ public void UpdateVariable(string name, FormulaValue value)
         /// <returns>The formula's result.</returns>
         public FormulaValue Eval(string expressionText, RecordValue parameters = null, ParserOptions options = null)
         {
-            return EvalAsync(expressionText, CancellationToken.None, parameters, options).Result;          
+            try
+            {
+                return EvalAsync(expressionText, CancellationToken.None, parameters, options).Result;
+            }
+            catch (RuntimeMaxCallDepthException)
```
lesaltzm, line 175: This is caught within evalAsync, it shouldn't reach here.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
MalekiRe: https://github.com/microsoft/Power-Fx/pull/437#discussion_r906320365

src/libraries/Microsoft.PowerFx.Interpreter/UDFTexlFunction.cs
```diff
@@ -0,0 +1,99 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.PowerFx.Core.Parser;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx
+{
+    public sealed partial class RecalcEngine
```
lesaltzm, line 13: Do these all need to be inner classes of RecalcEngine?

src/libraries/Microsoft.PowerFx.Interpreter/UDFTexlFunction.cs
```diff
@@ -0,0 +1,99 @@
...
+                _engine = engine;
+                _expressionText = expressionText;
+                _parameterType = parameterType;
+                _options = options;
+            }
+
+            public CheckResult Get() => _engine.Check(_expressionText, _parameterType, _options);
+        }
+
+        internal void DefineFunction(string name, string body, FormulaType returnType, params NamedFormulaType[] parameters)
```
lesaltzm, line 68: >DefineFunction [](http://example.com/codeflow?start=22&length=14)

This should probably be in recalcengine.cs

lesaltzm:
Can you change the title of this PR to be something more meaningful?
i.e. "Preliminary support for user-defined functions in RecalcEngine" or something along those lines.

src/libraries/Microsoft.PowerFx.Interpreter/RuntimeMaxCallDepthException.cs
```diff
@@ -0,0 +1,35 @@
...
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.PowerFx.Interpreter
+{
+    [Serializable]
+    internal class RuntimeMaxCallDepthException : Exception
```
lesaltzm, line 11: Curious, why [Serializable]?

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
```diff
@@ -38,11 +38,10 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
   </ItemGroup>
-  
-	<ItemGroup>
-	  <AdditionalFiles Include="..\..\stylecop.json" Link="stylecop.json" />
-	</ItemGroup>
```
lesaltzm, line 44: nit: add back the missing line to avoid diff noise.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs
```diff
@@ -18,6 +18,16 @@ public static ErrorValue RuntimeTypeMismatch(IRContext irContext)
             });
         }
 
+        public static ErrorValue MaxCallDepth(IRContext irContext)
+        {
+            return new ErrorValue(irContext, new ExpressionError()
+            {
+                Message = "Max call depth reached",
```
lesaltzm, line 25: >Max call depth reached [](http://example.com/codeflow?start=27&length=22)

// TODO: Validate error message with greg

src/libraries/Microsoft.PowerFx.Interpreter/SymbolContext.cs
```diff
@@ -2,7 +2,9 @@
 // Licensed under the MIT license.
 
 using System.Collections.Generic;
+using Microsoft.PowerFx.Core.IR;
```
lesaltzm, line 5: Un needed usings?

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -47,9 +48,10 @@ private PowerFxConfig(CultureInfo cultureInfo, EnumStoreBuilder enumStoreBuilder
         /// <summary>
         /// Initializes a new instance of the <see cref="PowerFxConfig"/> class.        
         /// </summary>
-        /// <param name="cultureInfo">Culture to use.</param>        
-        public PowerFxConfig(CultureInfo cultureInfo = null)
-            : this(cultureInfo, new EnumStoreBuilder().WithRequiredEnums(BuiltinFunctionsCore.BuiltinFunctionsLibrary)) 
+        /// <param name="cultureInfo">Culture to use.</param>
+        /// <param name="maxCallDepth"></param>        
+        public PowerFxConfig(CultureInfo cultureInfo = null, int maxCallDepth = 100)
```
lesaltzm, line 53: >int maxCallDepth = 100 [](http://example.com/codeflow?start=61&length=22)

Consider leaving this out of the public constructor for now. We may want to group a bunch of host-protection config under a single object, and if this is part of a signature, cleaning it up becomes a breaking change.

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -98,6 +101,15 @@ public IEnumerable<string> GetAllFunctionNames()
             return Functions.Select(func => func.Name).Distinct();
         }
 
+        /// <summary>
+        /// Gets the max number of recursive function calls allowed.
+        /// </summary>
+        /// <returns></returns>
+        public int GetMaxCallDepth()
```
lesaltzm, line 108: >public [](http://example.com/codeflow?start=8&length=6)

internal, imo

src/libraries/Microsoft.PowerFx.Interpreter/EvalContext.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// <summary>
+        /// Increments the runtime call depth counter.
+        /// Is threadsafe.
+        /// </summary>
+        /// <returns>False when you go over the recursive depth limit, true normally.</returns>
+        public bool IncrementCallDepth()
+        {
+            _mutex.WaitOne();
+            _currentCallDepth++;
+            if (_currentCallDepth >= _maxCallDepth)
```
lesaltzm, line 33: I don't think we need Mutex for this, Interlocked.Increment/Decrement should be sufficient: 
[https://docs.microsoft.com/en-us/dotnet/api/system.threading.interlocked.increment?view=net-6.0](https://docs.microsoft.com/en-us/dotnet/api/system.threading.interlocked.increment?view=net-6.0)

lesaltzm: I'm also not sure we need to be that careful about concurrent access here, at the moment EvalVistitor is single threaded.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -84,7 +91,24 @@ public static IExpression CreateEvaluatorDirect(CheckResult result)
             result.ThrowOnErrors();            
 
             (var irnode, var ruleScopeSymbol) = IRTranslator.Translate(result._binding);
-            return new ParsedExpression(irnode, ruleScopeSymbol);
+            return new ParsedExpression(irnode, ruleScopeSymbol, new Interpreter.EvalContext(maxCallDepth));
+        }
+
+        // would also be a good generic helper 
+        private static NamedValue[] Zip(NamedFormulaType[] parameters, FormulaValue[] args)
```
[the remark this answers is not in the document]
lesaltzm: Or to UDFTexlFunction

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngineResolver.cs
```diff
@@ -23,6 +25,19 @@ public RecalcEngineResolver(RecalcEngine parent, PowerFxConfig powerFxConfig)
             _powerFxConfig = powerFxConfig;
         }
 
+        public override IEnumerable<TexlFunction> LookupFunctions(DPath theNamespace, string name, bool localeInvariant = false)
+        {
+            if (theNamespace.IsRoot)
+            {
+                if (_parent._customFuncs.TryGetValue(name, out var func))
+                {
+                    return new TexlFunction[] { func };
```
lesaltzm, line 34: This is interesting, we could theoretically support multiple overloads from UDFs at some point.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -239,6 +241,99 @@ public void FormulaCantRedefine()
                engine.SetFormula("A", "3", OnUpdate));
         }
 
+        [Fact]
+        public void DefFunc()
```
lesaltzm, line 245: Great test cases! Consider also adding tests for 
Function Bodies with errors,
Definitions with the same name,
invocations with incorrect parameter count/types

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
MalekiRe: Done

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
MalekiRe: Done

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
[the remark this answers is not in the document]
MalekiRe: Done

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
MalekiRe: Done

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -144,21 +168,35 @@ public void UpdateVariable(string name, FormulaValue value)
...
+                var check = Check(expressionText, (RecordType)parameters.IRContext.ResultType, options);
+                check.ThrowOnErrors();
 
-            var newValue = await check.Expression.EvalAsync(parameters, cancel);
-            return newValue;
+                var newValue = await check.Expression.EvalAsync(parameters, cancel);
+                return newValue;
+            }
+            catch (RuntimeMaxCallDepthException)
+            {
```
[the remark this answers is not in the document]
MalekiRe: Fixed up.

src/libraries/Microsoft.PowerFx.Interpreter/UDFTexlFunction.cs
MalekiRe: Fixed

src/libraries/Microsoft.PowerFx.Interpreter/UDFTexlFunction.cs
MalekiRe: Fixed

src/libraries/Microsoft.PowerFx.Interpreter/RuntimeMaxCallDepthException.cs
MalekiRe: Any issues with this?

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
MalekiRe: Having issues with fixing this right now, going to try and resolve it as the very last thing.

src/libraries/Microsoft.PowerFx.Interpreter/UDFTexlFunction.cs
```diff
@@ -0,0 +1,99 @@
...
+        internal void BindAllUDFs()
+        {
+            foreach (var func in _funcsToBind)
+            {
+                func.LazyGetExpression();
+            }
+
+            _funcsToBind.Clear();
+        }
+    }
```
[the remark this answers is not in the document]
MalekiRe: Done in the way we discussed.

src/libraries/Microsoft.PowerFx.Interpreter/SymbolContext.cs
MalekiRe: Done

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
MalekiRe: Done

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
MalekiRe: Done

src/libraries/Microsoft.PowerFx.Interpreter/EvalContext.cs
MalekiRe: Changed to use Interlocked#Increment and Interlocked#Decrement

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
MalekiRe: Done

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -239,6 +242,145 @@ public void FormulaCantRedefine()
...
+        {
+            var config = new PowerFxConfig(null);
+            var recalcEngine = new RecalcEngine(config);
+            Assert.False(recalcEngine.DefineFunctions(new UDFDefinition("foo", "x+1", FormulaType.Number, new NamedFormulaType("x", FormulaType.Number))).Any());
+            try
+            {
+                var result = recalcEngine.Eval("foo(false)");
+                Assert.True(false);
+            }
+            catch (AggregateException)
```
lesaltzm, line 370: >AggregateException [](http://example.com/codeflow?start=19&length=18)

Why exception? Shouldn't this be a runtime error at worst?

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -239,6 +242,145 @@ public void FormulaCantRedefine()
...
+            var recalcEngine = new RecalcEngine(config);
+            Assert.True(recalcEngine.DefineFunctions(
+                new UDFDefinition("foo", "foo()", FormulaType.Blank),
+                new UDFDefinition("foo", "x+1", FormulaType.Number)).Any());
+            try
+            {
+                recalcEngine.Eval("foo()");
+                Assert.True(false);
+            }
+            catch (System.AggregateException)
```
lesaltzm, line 346: >AggregateException [](http://example.com/codeflow?start=26&length=18)

Same here, shouldn't this be either: Foo isn't defined, if we decided both were erroneous, or executing correctly with the 1st one?

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -239,6 +242,145 @@ public void FormulaCantRedefine()
...
+
+        [Fact]
+        public void UDFIncorrectParametersTest()
+        {
+            var config = new PowerFxConfig(null);
+            var recalcEngine = new RecalcEngine(config);
+            Assert.False(recalcEngine.DefineFunctions(new UDFDefinition("foo", "x+1", FormulaType.Number, new NamedFormulaType("x", FormulaType.Number))).Any());
+            try
+            {
+                var result = recalcEngine.Eval("foo(false)");
```
lesaltzm, line 367: >foo(false [](http://example.com/codeflow?start=48&length=9)

Try one passing a record to something that expects a scalar, as well.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
MalekiRe: All the exceptions turn into System.AggregateException for some reason when catching them in [Fact] functions. If you know how to resolve this it would be much better to catch the specific exception.

lesaltzm: This is because eval is async under the hood and the `eval()` function is just blocking on the result, which also winds up wrapping exceptions (@MikeStall, I recall you had opinions on this)

For your tests, you'd want to call EvalAsync and catch a specific exception, although, in these cases, these shouldn't throw, they should be either errors on the original call to register, or a compile time error on Eval.

MalekiRe: Hmmm, I wonder why they throw an error then.

src/libraries/Microsoft.PowerFx.Interpreter/UDFHelper.cs
```diff
@@ -0,0 +1,149 @@
...
+            }
+
+            // Used when trying to define a function, there is an error before binding, such as the function name already being defined.
+            public UDFLazyBinder(ExpressionError expressionError, string name)
+            {
+                _expressionError.Add(expressionError);
+                _name = name;
+            }
+
+            public string GetName()
```
lesaltzm, line 86: >GetName [](http://example.com/codeflow?start=26&length=7)

the general C# preference here is for a property

`public string Name {get;}` 

over GetName+ _name

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
MalekiRe: Wait no this throwing an error makes sense. It originally throws an error on the Register, but then if you try to use it it *also* throws an error.

src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs
```diff
@@ -99,7 +99,7 @@ public virtual bool Lookup(DName name, out NameLookupInfo nameInfo, NameLookupPr
             return false;
         }
 
-        public IEnumerable<TexlFunction> LookupFunctions(DPath theNamespace, string name, bool localeInvariant = false)
+        public virtual IEnumerable<TexlFunction> LookupFunctions(DPath theNamespace, string name, bool localeInvariant = false)
```
lesaltzm, line 102: >virtual [](http://example.com/codeflow?start=15&length=7)

@MikeStall this is a good change to make especially in light of discussions around composing resolvers

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -45,6 +45,15 @@ public IEnumerable<string> GetAllFunctionNames()
             return Config.GetAllFunctionNames();
         }
 
+        /// <summary>
+        /// Returns the maximum number of recursive function calls.
+        /// </summary>
+        /// <returns></returns>
+        public int GetMaxCallDepth()
```
lesaltzm, line 52: internal

src/libraries/Microsoft.PowerFx.Interpreter/UDFHelper.cs
```diff
@@ -0,0 +1,149 @@
...
+            var result = new NamedValue[args.Length];
+            for (var i = 0; i < args.Length; i++)
+            {
+                result[i] = new NamedValue(parameters[i].Name, args[i]);
+            }
+
+            return result;
+        }
+        
+        public class UDFDefinition
```
lesaltzm, line 51: nits: Each class in here should be a separate file, and not inner classes of UDFHelper

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
lesaltzm: remove the partial keyword?

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
lesaltzm: What changed here? Is this the right approach over a config object?

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -84,7 +91,7 @@ public static IExpression CreateEvaluatorDirect(CheckResult result)
             result.ThrowOnErrors();            
 
             (var irnode, var ruleScopeSymbol) = IRTranslator.Translate(result._binding);
-            return new ParsedExpression(irnode, ruleScopeSymbol);
+            return new ParsedExpression(irnode, ruleScopeSymbol, new Interpreter.EvalContext(maxCallDepth));
```
lesaltzm, line 94: >Interpreter [](http://example.com/codeflow?start=69&length=11)

nit: interpreter. shouldn't be needed, you already have a using for interpreter

src/libraries/Microsoft.PowerFx.Interpreter/UDFHelper.cs
```diff
@@ -0,0 +1,149 @@
...
+                return _function.Bind();
+            } 
+        }
+
+        internal class UDFTexlFunction : CustomTexlFunction, IAsyncTexlFunction
+        {
+            internal NamedFormulaType[] _parameterNames;
+            internal IExpression _expr;
+            internal LazyCheck _check;
+
```
lesaltzm, line 107: Does the function definition need a reference to LazyCheck?

src/libraries/Microsoft.PowerFx.Interpreter/UDFHelper.cs
```diff
@@ -0,0 +1,149 @@
...
+                if (_expressionError.Any())
+                {
+                    return _expressionError;
+                }
+
+                return _function.Bind();
+            } 
+        }
+
+        internal class UDFTexlFunction : CustomTexlFunction, IAsyncTexlFunction
```
lesaltzm, line 102: >UDFTexlFunction [](http://example.com/codeflow?start=23&length=15)

naming nit: UserDefinedTexlFunction (otherwise it's UserDefinedFunctionTexlFunction if you expand the acronym :smile: )

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -156,9 +163,96 @@ public async Task<FormulaValue> EvalAsync(string expressionText, CancellationTok
...
+            _customFuncs.Remove(name);
+        }
+
+        /// <summary>
+        /// Tries to define and bind all the functions here. If any function names conflict returns an expression error. 
+        /// Also returns any errors from binding failing. All functions defined here are removed if any of them contain errors.
+        /// </summary>
+        /// <param name="udfDefinitions"></param>
+        /// <returns></returns>
+        internal IEnumerable<ExpressionError> DefineFunctions(params UDFDefinition[] udfDefinitions)
```
lesaltzm, line 228: >IEnumerable<ExpressionError> [](http://example.com/codeflow?start=17&length=28)

Maybe this should be IEnumerable<IEnumerable<ExpressionErrors>>, with the order corresponding to the UDF definitions

lesaltzm, line 228: >params [](http://example.com/codeflow?start=62&length=6)

Probably IEnumerable instead of params []

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -156,9 +163,96 @@ public async Task<FormulaValue> EvalAsync(string expressionText, CancellationTok
...
+            {
+                record = record.Add(p);
+            }
+
+            var check = new LazyCheck(this, body, record);
+
+            var func = new UDFTexlFunction(name, returnType, parameters.Select(x => x.Type).ToArray())
+            {
+                _parameterNames = parameters,
+                _check = check,
```
lesaltzm, line 199: these should probably be args on the constructor

src/libraries/Microsoft.PowerFx.Interpreter/UDFHelper.cs
```diff
@@ -0,0 +1,149 @@
...
+namespace Microsoft.PowerFx.Interpreter
+{
+    internal class UDFHelper
+    {
+        public class LazyCheck
+        {
+            internal string _expressionText;
+            internal RecordType _parameterType;
+            internal ParserOptions _options;
+            internal Engine _engine;
```
lesaltzm, line 21: >_engine [](http://example.com/codeflow?start=28&length=7)

naming nit: internals should start with a capital letter. 
_ is only for privates.

lesaltzm: Actually, these should probably all be privates

src/libraries/Microsoft.PowerFx.Interpreter/UDFHelper.cs
```diff
@@ -0,0 +1,149 @@
...
+
+            public LazyCheck(Engine engine, string expressionText, RecordType parameterType = null, ParserOptions options = null)
```
lesaltzm, line 23: >RecordType parameterType = null, ParserOptions options = null [](http://example.com/codeflow?start=67&length=61)

Are these/should these be optional?

src/libraries/Microsoft.PowerFx.Interpreter/UDFHelper.cs
```diff
@@ -0,0 +1,149 @@
...
+                _expressionText = expressionText;
+                _parameterType = parameterType;
+                _options = options;
+            }
+
+            public CheckResult Get() => _engine.Check(_expressionText, _parameterType, _options);
+        }
+
+        // would also be a good generic helper 
+        private static NamedValue[] Zip(NamedFormulaType[] parameters, FormulaValue[] args)
```
lesaltzm, line 35: >NamedValue [](http://example.com/codeflow?start=23&length=10)

If you can, we prefer IEnumerable for interfaces over arrays.

src/libraries/Microsoft.PowerFx.Interpreter/UDFHelper.cs
```diff
@@ -0,0 +1,149 @@
...
+            {
+                result[i] = new NamedValue(parameters[i].Name, args[i]);
+            }
+
+            return result;
+        }
+        
+        public class UDFDefinition
+        {
+            internal string _name;
```
lesaltzm, line 53: >internal [](http://example.com/codeflow?start=12&length=8)

same comment about privates for these other classes as well.

src/libraries/Microsoft.PowerFx.Interpreter/UDFHelper.cs
```diff
@@ -0,0 +1,149 @@
...
+            } 
+        }
+
+        internal class UDFTexlFunction : CustomTexlFunction, IAsyncTexlFunction
+        {
+            internal NamedFormulaType[] _parameterNames;
+            internal IExpression _expr;
+            internal LazyCheck _check;
+
+            public override bool SupportsParamCoercion => false;
```
lesaltzm, line 108: >false [](http://example.com/codeflow?start=58&length=5)

If you set this to true, is there anything else you'd need to support coercion?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -69,6 +69,9 @@ internal abstract class TexlFunction : IFunction
 
         private string _cachedLocaleName;
 
+        // Return true if the function is a User Defined Function and false otherwise.
+        public virtual bool IsUDF => false;
```
MikeStall, line 73: Why does this matter? Can we remove this flag?

src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs
MikeStall: Let's chat more - my concern is that we currently get the function list via PowerFxConfig. This means we have 2 ways to provide functions.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -239,6 +242,145 @@ public void FormulaCantRedefine()
...
+                new UDFDefinition(
+                    "foo",
+                    "foo()",
+                    FormulaType.Blank)).Any());
+            var result = recalcEngine.Eval("foo()");
+            Assert.IsType<ErrorValue>(result);
+        }
+
+        [Fact]
+        public void DefHailstoneSequence()
```
MikeStall, line 294: >DefHailstoneSequence [](http://example.com/codeflow?start=20&length=20)

nit - Can you include a link to the Hailstone seqeuence?
Is there something significant about this; or is it just testing a popular recursive example (ala Fibonacci)

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
MikeStall: To get rid of the AggregateException, make this function async and use await EvalAsync()  instead. 
Then you can even do Assert.ThrowsAsync<> to test for the exception.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
MikeStall: Call .Check and that should have the failure w/o an exception. 


If we call Eval() on bad input, it will throw.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
MikeStall: Or even both (via an extension method)
params[] is nice for the caller :)

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
MalekiRe: Nothing particularly significant about it, it is just a popular example of a recursive function.
https://gist.github.com/stkade/3773810

src/libraries/Microsoft.PowerFx.Interpreter/UDFHelper.cs
MalekiRe: should I use getters or properties for this instead then?

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -156,9 +163,96 @@ public async Task<FormulaValue> EvalAsync(string expressionText, CancellationTok
...
+
+            if (_customFuncs.ContainsKey(name))
+            {
+                return new UDFLazyBinder(
+                    new ExpressionError()
+                {
+                    Message = "Function name already defined",
+                    Span = null,
+                    Kind = ErrorKind.Internal, //Todo: change this out for a new kind of ErrorKind error.
+                }, name);
```
MikeStall, line 210: Should this be an exception instead of runtime error?
Double adding a bug in the host.

src/libraries/Microsoft.PowerFx.Interpreter/UDFHelper.cs
MalekiRe: I honestly don't know, I based these off of Engine#Check, it isn't necessary for the specific use of LazyCheck that I am using right now, where it just takes in a string.

src/libraries/Microsoft.PowerFx.Interpreter/EvalContext.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Threading;
+
+namespace Microsoft.PowerFx.Interpreter
+{
+    internal class EvalContext
+    {
+        private int _currentCallDepth = 0;
+
+        // We can set this up to read from powerfx config, but it isn't nessecary yet.
+        private readonly int _maxCallDepth;
```
MikeStall, line 16: >_maxCallDepth [](http://example.com/codeflow?start=29&length=13)

If we just pass this as a stack parameter, then we don't need any locking or decrement calls.

src/libraries/Microsoft.PowerFx.Interpreter/UDFHelper.cs
MikeStall: +1 to using privates instead of internal. Especially if it's 'private readonly' and only set in ctor.

src/libraries/Microsoft.PowerFx.Interpreter/UDFHelper.cs
MalekiRe: It appeared to coerce values when this was previously true, so likely not.

src/libraries/Microsoft.PowerFx.Interpreter/UDFHelper.cs
```diff
@@ -0,0 +1,149 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Interpreter
+{
+    internal class UDFHelper
+    {
+        public class LazyCheck
```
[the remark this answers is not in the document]
MalekiRe: Good point, we actually don't need it to be lazy anymore, good name suggestions? I'm thinking "CheckWrapper" right now, but idk if that's the best.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
MalekiRe: Done

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
MalekiRe: Unneeded, recursion depth handled in seperate PR.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
MalekiRe: Unneeded, recursion depth handled in seperate PR.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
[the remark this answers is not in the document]
MalekiRe: Unneeded, recursion depth handled in seperate PR.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
MalekiRe: Unneeded, recursion depth handled in seperate PR.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
MalekiRe: Fixed

src/libraries/Microsoft.PowerFx.Interpreter/UDFHelper.cs
MalekiRe: Fixed

src/libraries/Microsoft.PowerFx.Interpreter/UDFHelper.cs
MalekiRe: Done.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
MalekiRe: Done

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
MalekiRe: Done.

src/libraries/Microsoft.PowerFx.Interpreter/UDFHelper.cs
MalekiRe: Done

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.2-preview.20220713-312200](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.2-preview.20220713-312200).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.RecalcEngine.CreateEvaluatorDirect(Microsoft.PowerFx.CheckResult)' does not exist in the implementation but it does exist in the contract.
```

</details>

BruceHaley:

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
MalekiRe: Now we have both.

src/libraries/Microsoft.PowerFx.Interpreter/UDFHelper.cs
MalekiRe: Removed ParserOptions as a parameter but left RecordType as it's used.

BruceHaley:

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -11,6 +11,7 @@
 using System.Web;
 using Microsoft.OpenApi.Models;
 using Microsoft.PowerFx.Connectors.Execution;
+using Microsoft.PowerFx.Interpreter;
```
lesaltzm, line 14: Why do we need the new using here and in ServiceFunction.cs?

MalekiRe: Oh yes! This was something I ended up undoing but I didn't remove the usings. I will go fix those.

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -1,219 +1,231 @@
...
+    /// <summary>
+    /// A container that allows for compiler customization.
+    /// </summary>
+    public sealed class PowerFxConfig
+    {
+        private bool _isLocked;
+        private readonly HashSet<TexlFunction> _extraFunctions = new HashSet<TexlFunction>();
+        private readonly Dictionary<DName, IExternalEntity> _environmentSymbols;
+        private DisplayNameProvider _environmentSymbolDisplayNameProvider;
+        private readonly int _maxCallDepth = 100;
```
lesaltzm, line 26: unused? (clean up from #470?)

src/libraries/Microsoft.PowerFx.Interpreter/EvalContext.cs
```diff
@@ -0,0 +1,44 @@
...
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+
+namespace Microsoft.PowerFx.Interpreter
+{
+    internal class EvalContext
```
lesaltzm, line 11: This should be unnecessary, after #470

src/libraries/Microsoft.PowerFx.Interpreter/UDF/UDFLazyBinder.cs
```diff
@@ -0,0 +1,43 @@
...
+
+namespace Microsoft.PowerFx.Interpreter
+{
+    internal class UDFLazyBinder
+    {
+        private readonly UserDefinedTexlFunction _function;
+        
+        private readonly List<ExpressionError> _expressionError = new List<ExpressionError>();
+
+        public string Name
```
lesaltzm, line 15: nit, can just be `public readonly string Name`

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
MalekiRe: Done

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -45,6 +45,15 @@ public IEnumerable<string> GetAllFunctionNames()
             return Config.GetAllFunctionNames();
         }
 
+        /// <summary>
+        /// Returns the maximum number of recursive function calls.
+        /// </summary>
+        /// <returns></returns>
+        internal int GetMaxCallDepth()
```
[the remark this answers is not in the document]
MalekiRe: Removed.

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
MalekiRe: Done.

src/libraries/Microsoft.PowerFx.Interpreter/EvalContext.cs
MalekiRe: Yep, removed.

src/libraries/Microsoft.PowerFx.Interpreter/UDF/UDFLazyBinder.cs
MalekiRe: Changed.

BruceHaley:

BruceHaley:

src/libraries/Microsoft.PowerFx.Interpreter/ParsedExpression.cs
```diff
@@ -39,5 +39,19 @@ public async Task<FormulaValue> EvalAsync(RecordValue parameters, CancellationTo
...
+
+        public async Task<FormulaValue> EvalAsync(RecordValue parameters, CancellationToken cancel, StackDepthCounter stackMarker)
+        {
+            var ev2 = new EvalVisitor(_cultureInfo, cancel);
+            try
+            {
+                var newValue = await _irnode.Accept(ev2, new EvalVisitorContext(SymbolContext.NewTopScope(_topScopeSymbol, parameters), stackMarker));
+                return newValue;
+            }
+            catch (MaxCallDepthException maxCallDepthException)
```
lesaltzm, line 51: This call is for evaluating a UDF, right? Should the exception be caught, or should it be thrown up the whole call stack to the top `EvalAsync`

BruceHaley:

BruceHaley:

src/libraries/Microsoft.PowerFx.Interpreter/ParsedExpression.cs
MikeStall: +1
Rename this overload to EvalInternalAsync
Remove catch here, keep the one on line 37. 

Add comment :)

MalekiRe: Fixed

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs
MalekiRe: Can be changed in another PR, just wanna get this through.

src/libraries/Directory.Build.props
```diff
@@ -1,7 +1,6 @@
 <!-- Contains common properties that apply to projects under the Libraries folder -->
```
bryancroteau-MSFT, line 1: Why is this change included?

BruceHaley:

src/libraries/Microsoft.PowerFx.Interpreter/UDF/CheckWrapper.cs
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Interpreter.UDF
+{
+    internal class CheckWrapper
```
bryancroteau-MSFT, line 11: Could we get a comment describing what this class is for?

src/libraries/Microsoft.PowerFx.Interpreter/UDF/UDFHelper.cs
```diff
@@ -0,0 +1,29 @@
...
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Interpreter
+{
+    internal class UDFHelper
```
bryancroteau-MSFT, line 11: This can be a static class right?

src/tests/Microsoft.PowerFx.Interpreter.Tests/Helpers/AsyncFunctionHelperBase.cs
```diff
@@ -8,6 +8,7 @@
 using Microsoft.PowerFx.Core.Functions;
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Interpreter;
```
bryancroteau-MSFT, line 11: Is this change needed?

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineAsyncTests.cs
```diff
@@ -10,6 +10,7 @@
 using Microsoft.PowerFx.Core.Tests;
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Interpreter;
```
bryancroteau-MSFT, line 13: Is this needed?

src/libraries/Microsoft.PowerFx.Interpreter/UDF/CheckWrapper.cs
MalekiRe: done

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -169,6 +176,83 @@ public async Task<FormulaValue> EvalAsync(string expressionText, CancellationTok
...
+            {
+                record = record.Add(p);
+            }
+
+            var check = new CheckWrapper(this, definition.Body, record);
+
+            var func = new UserDefinedTexlFunction(definition.Name, definition.ReturnType, definition.Parameters, check);
+            if (_customFuncs.ContainsKey(definition.Name))
+            {
+                return new UDFLazyBinder(
```
MikeStall, line 197: > return new UDFLazyBinder( [](http://example.com/codeflow?start=15&length=26)

I recommend we throw since this can be considered a host error (not necessarily user error). 
The Host shouldn't have to scan the expressionError() list , those ought to be passed straight through to the user and should be things the user can directly fix (like syntax errors).

BruceHaley:

bryancroteau-MSFT (approved):
Looks good to me.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
MalekiRe: Fixed

BruceHaley:
The binary compatibility report for library 'Microsoft.PowerFx.LanguageServerProtocol' wasn't generated. This may have happened because the NuGet library 'Microsoft.PowerFx.LanguageServerProtocol' for version '0.2.2-preview.20220721-313359' was unavailable or a connectivity issue.

PR #439, opened 2022-06-23
Prevent overwriting display name with logical name if we're in the middle of updating names for another type

lesaltzm:
Binding with _updateDisplayNames but not _forceUpdateDisplayNames is only supported in canvas. This fixes a long-standing bug where if we're refreshing one entity, other entities may get their display names overwritten with logical names if they aren't also marked for refresh.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Types/BindingErrorType.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Types/BindingErrorType.cs
@@ -17,6 +17,6 @@ namespace Microsoft.PowerFx.Types
         }
 
         /// <inheritdoc />
-        public override void Visit(ITypeVistor visitor) => visitor.Visit(this);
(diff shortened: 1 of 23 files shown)


src/tests/Microsoft.PowerFx.Core.Tests/DisplayNameOptionSetTests.cs
```diff
@@ -115,25 +115,5 @@ public void PowerFxConfigCollisionsThrow()
             Assert.True(config.TryGetSymbol(new DName("SomeDisplayName"), out _, out displayName));
             Assert.Equal("SomeDisplayName", displayName.Value);
         }
-
-        [Fact]
-        public void Sample()
```
lesaltzm, line 120: Removed, since it's a duplicate of a test in Engine Tests

VamsiModem (approved):

PR #440, opened 2022-06-23
Add option param to Engine.Check overload

ivanradicek:
Engine has two `Check` overloads:
- `Check(string, ...)` and
- `Check(ParseResult, ...)`

The first one takes `ParserOptions` as an argument, while the second does not. This seems intuitive since the second one does not perform any parsing. However, `ParserOptions` have `AllowsSideEffects` property, which is in the first case passed down to the binder, and in the second is not.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
@@ -108,9 +108,7 @@ namespace Microsoft.PowerFx
         public CheckResult Check(string expressionText, RecordType parameterType = null, ParserOptions options = null)
         {
             var parse = Parse(expressionText, options);
-
-            var bindingConfig = new BindingConfig(options?.AllowsSideEffects == true);
-            return CheckInternal(parse, parameterType, bindingConfig);
+            return Check(parse, parameterType, options);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -108,20 +108,20 @@ public static ParseResult Parse(string expressionText, ParserOptions options = n
...
 
         /// <summary>
         /// Type check a formula without executing it. 
         /// </summary>
         /// <param name="parse">the parsed expression. Obtain from <see cref="Parse(string, ParserOptions)"/>.</param>
         /// <param name="parameterType">types of additional args to pass.</param>
+        /// <param name="options">parser options to use.</param>
         /// <returns></returns>
-        public CheckResult Check(ParseResult parse, RecordType parameterType = null)
+        public CheckResult Check(ParseResult parse, RecordType parameterType = null, ParserOptions options = null)
```
ivanradicek, line 121: Since it's probably weird to have `ParserOptions` on a method that does not perform any parsing, an alternative would be to store them together with `ParseResult`. Thoughts?

src/tests/Microsoft.PowerFx.Core.Tests/BindingEngineTests.cs
```diff
@@ -199,6 +204,49 @@ public void TableInRegressionError()
...
+
+            var result1 = engine.Check(formula, options: options);
+            Assert.True(result1.IsSuccess);
+
+            var parseResult2 = engine.Parse(formula, options);
+            var result2 = engine.Check(parseResult2, options: options);
+            Assert.True(result2.IsSuccess);
+        }
+
+        private class DummyFunction : TexlFunction
```
ivanradicek, line 226: We need a function with behavior in the test above. However, I couldn't find any public function that has behavior and `Set` is defined in interpreter tests.

MikeStall: This approach is great, but Can we rename to "BehaviorFunction" or add a comment so that we know why we're defining a new function here.

ivanradicek: Did that

PR #444, opened 2022-06-28
[ { a: 1 } ] - [Breaking Change] Do not add "Value:" for arrays of records

LucGenetier:
This is part of the GA1.0 breaking changes for PVA's public preview.
Today [ { a: 1 } ] is equivalent to Table({ Value: { a:1 }}) and we want to be close to Json serialization in making it equivalent to Table({ a: 1 })
To use this new feature, you need to activate TableSyntaxDoesntWrapRecords feature flag


lesaltzm:
Can you update the PR Title/Description with some details about this change?

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -86,9 +86,8 @@ public override async ValueTask<FormulaValue> Visit(TableNode node, SymbolContex
                 var arg = await child.Accept(this, context);
                 args[i] = arg;
             }
-
-            // This is always a single column table
-            var tableValue = new InMemoryTableValue(node.IRContext, StandardTableNodeRecords(node.IRContext, args, forceSingleColumn: true));
+            
+            var tableValue = new InMemoryTableValue(node.IRContext, StandardTableNodeRecords(node.IRContext, args, forceSingleColumn: false));
```
lesaltzm, line 90: Is the forceSingleColumn flag needed anywhere else or can it be cleaned up?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Record.txt
```diff
@@ -15,8 +15,8 @@ false
 >> IsBlank({ x : Blank(), y : 2}.x)
 true
 
->> First([{a: 1}]).Value.a
+>> First([{a: 1}]).a
```
lesaltzm, line 18: Test cases showing that `.Value` is a compiler error in these cases would also be good.

lesaltzm (changes requested):
Thanks! This is a good change to make, we just need to be careful about it.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
LucGenetier: Used in LibraryMath for Sequence

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Record.txt
LucGenetier: Fair point

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -5836,9 +5836,8 @@ public override void PostVisit(TableNode node)
                     }
                 }
 
-                _txb.SetType(
-                    node,
-                    exprType.IsValid ? DType.CreateTable(new TypedName(exprType, TableValue.ValueDName)) : DType.EmptyTable);
+                DType tableType = exprType.IsValid ? (exprType.IsRecord ? DType.CreateTable(exprType.GetNames(DPath.Root)) : DType.CreateTable(new TypedName(exprType, TableValue.ValueDName))) : DType.EmptyTable;
```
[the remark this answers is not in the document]
MikeStall: +1 - we will definitely need some sort of flag. This is part of the changes for PVA before they go public preview. (ie, it needs to happen soon; whereas canvas will take a while)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Record.txt
```diff
@@ -15,8 +15,11 @@ false
...
```
MikeStall, line 18: We need to decide how we're managing the breaking change here. IE, these tests are shared by Canvas too; so they'll run against JS (which does not yet have this update; and it can't have it until PowerApps has a converter).

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -125,22 +125,34 @@ public HttpRequestMessage BuildRequest(FormulaValue[] args)
...
-            FormulaValueSerializer serializer = _argMapper.ContentType.ToLowerInvariant() switch
-            {
-                OpenApiExtensions.ContentType_XWwwFormUrlEncoded => new OpenApiFormUrlEncoder(schemaLessBody),                
-                OpenApiExtensions.ContentType_TextPlain => new OpenApiTextSerializer(schemaLessBody),
-                _ => new OpenApiJsonSerializer(schemaLessBody)
-            };
-            
-            serializer.StartSerialization(referenceId);
-            foreach (var kv in map)
+            FormulaValueSerializer serializer = null;
```
lesaltzm, line 128: Is this related to the `[]` syntax change? If not, can this be a separate PR?

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
```diff
@@ -53,6 +53,11 @@
     </Content>
   </ItemGroup>
 
+  <ItemGroup>
```
lesaltzm, line 56: unintentional .csproj change?

src/libraries/Microsoft.PowerFx.Core/FeatureFlags.cs
```diff
@@ -25,13 +31,28 @@ public static bool StringInterpolation
...
+                if (!value && !_inTests) 
                 { 
                     throw new NotSupportedException("Can't change field after it's set");
                 }
 
                 _stringInterpolation = value;
             }
         }
+
+        public static bool NoValueInRecordArrays
```
lesaltzm, line 43: naming: `TableSyntaxDoesntWrapRecords`? 
Basically, we don't have arrays, and we shouldn't mention "array" on our public surface.

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
LucGenetier: I'll create a separated PR then

src/libraries/Microsoft.PowerFx.Core/FeatureFlags.cs
LucGenetier: Ok. We'll soon have to support Arrays I think

lesaltzm: Discussed in Greg's recent xr: there is no plan to support arrays in PowerFx at the moment. If this change isn't sufficient, we may also support default fields within records, but beyond that, an array type is unlikely.

src/libraries/Microsoft.PowerFx.Core/FeatureFlags.cs
```diff
@@ -22,16 +28,23 @@ public static class FeatureFlags
...
-                // Once flipped, can't unflip. 
-                if (!value) 
-                { 
-                    throw new NotSupportedException("Can't change field after it's set");
-                }
+            set => _stringInterpolation = ProtectedSet(value);
+        }
 
-                _stringInterpolation = value;
+        public static bool TableSyntaxDoesntWrapRecords
```
MikeStall, line 34: >public [](http://example.com/codeflow?start=8&length=6)

nit, can you add a comment similar to the one on StringInterpolation

MikeStall:
>         public override async ValueTask<FormulaValue> Visit(TableNode node, SymbolContext context)

ideally, these semantics would all be entirely captured in the IR (and no change required in the interpreter)

(else this means other backends would also need to be updated) 

For example, when this mode is enabled, could the IR just do the rewrite:

[{a:1},{a:2}] 

to 
Table({a:1}, {a:2}

So a change in IRTranslator, but then literally no change to interpreter?


(@lesaltzm ?)


---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs:76 in ca3594d. [](commit_id = ca3594d874f56edd84c3ef11a5554e761bc77159, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/NoValueInRecordArrays/Index_NoValue.txt
```diff
@@ -0,0 +1,4 @@
+﻿// Feature NoValueInRecordArrays set
+
+>> Index([ { a: 10, b: "asdf"}, { a: 20, b: "second"}, { a: 30, b: "third"}], 1.1).b
```
MikeStall, line 3: >1.1 [](http://example.com/codeflow?start=78&length=3)

nit: why 1.1?   We can have a dedicated Index() test for testing non-integer values to this arg. 

Or even to keep test simple, can we just do First()

MikeStall, line 3: >1.1 [](http://example.com/codeflow?start=78&length=3)

Another test idea:

Explicitly including Value:

First([ { Value : 10 } ]).Value  // 10

MikeStall: We probably don't need separate tests for Index and First. 
Since this is a table change, it will impact all table functions the same way.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/ExclusiveRunner.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+
+namespace Microsoft.PowerFx.Core.Tests
+{
+    /// <summary>
+    /// Ensures only one instance can exist at a time.
+    /// </summary>
+    public abstract class ExclusiveRunner : IDisposable
```
MikeStall, line 14: >ExclusiveRunner [](http://example.com/codeflow?start=26&length=15)

Yikes... ... this is painful...

I guess this is because FeatureFlags is static?

MikeStall:
> // Feature NoValueInRecordArrays set

it would be ideal if we could model this as a #SETUP:

And then if other backends pick it up, they would just fail (or implement it!)

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/NoValueInRecordArrays/Index_NoValue.txt:1 in ca3594d. [](commit_id = ca3594d874f56edd84c3ef11a5554e761bc77159, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/ExclusiveRunner.cs
MikeStall: even this isn't fully correct since we may run against other tests that call RecalcEngine directly (outside of BaseRunner).

MikeStall: Is there something we can do at the Xunit level? ie: 
[https://xunit.net/docs/running-tests-in-parallel](https://xunit.net/docs/running-tests-in-parallel)

MikeStall (dismissed):
:clock1:

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/NoValueInRecordArrays/Index_NoValue.txt
LucGenetier: 1.1 --> typo
Index & First - >those are only examples that illustrate the change but I agree all Table calls are potentially impacted

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/ExclusiveRunner.cs
LucGenetier: Yes, this is due to static FeatureFlags
We should probably consider changing them to non-static, maybe in PowerFxConfig or elsewhere

Didn't know Xunit was providing "Collection" feature for tests, it's exactly what we need.

src/tests/Microsoft.PowerFx.Core.Tests/TestRunnerTests/TestRunnerTests.cs
```diff
@@ -337,13 +322,9 @@ public void TestRunnerError()
...
-            var errorValue = new ErrorValue(
-                IR.IRContext.NotInSource(FormulaType.Number),
-                new ExpressionError { Kind = ErrorKind.InvalidFunctionUsage });
-            var runner = new MockRunner
-            {
-                _hook = (expr, setup) => errorValue // error
-            };
+            var errorValue = new ErrorValue(IR.IRContext.NotInSource(FormulaType.Number), new ExpressionError { Kind = ErrorKind.InvalidFunctionUsage });
+            
+            var runner = new MockRunner { _hook = (expr, setup) => errorValue /* error */ };
```
MikeStall, line 327: >/* [](http://example.com/codeflow?start=78&length=2)

nit - we should avoid /* comments. 
(I thought we actually had a style cop rule for this)

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -104,7 +105,26 @@ public override IntermediateNode Visit(TexlTableNode node, IRTranslatorContext c
                 Contracts.AssertValue(context);
 
                 var children = node.Children.Select(child => child.Accept(this, context)).ToArray();
-                return MaybeInjectCoercion(node, new TableNode(context.GetIRContext(node), children), context);
+                var irContext = context.GetIRContext(node);
+
+                if (!Preview.FeatureFlags.TableSyntaxDoesntWrapRecords || children.First() is not RecordNode)
```
MikeStall, line 110: >children.First()  [](http://example.com/codeflow?start=74&length=17)

What if Children is empty?

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
```diff
@@ -9,21 +9,52 @@
 
 namespace Microsoft.PowerFx.Interpreter.Tests
 {
+    [Collection("Interpreter")]
```
MikeStall, line 12: >[Collection("Interpreter")] [](http://example.com/codeflow?start=4&length=27)

What does this do?

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -104,7 +105,26 @@ public override IntermediateNode Visit(TexlTableNode node, IRTranslatorContext c
...
+                        var rt = new RecordType().Add(TableValue.ValueName, childNode.IRContext.ResultType);
+
+                        recordNodes.Add(new RecordNode(new IRContext(childNode.IRContext.SourceContext, rt), values));
+                    }
+
+                    children = recordNodes.ToArray();                    
+                    irContext = new IRContext(node.GetCompleteSpan(), irContext.ResultType);
+                }
+
+                return MaybeInjectCoercion(node, new TableNode(irContext, children), context);
```
MikeStall, line 127: >new TableNode(irContext, children [](http://example.com/codeflow?start=49&length=33)

can we remove TableNode altogether and just replace with a call to Table() 

Ie, see how &&/And keywords becomes a call to And() function.  (search for BuiltinFunctionsCore.And)

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
LucGenetier: Fixed and added a test
>> []
Table()

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
LucGenetier: Done

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
LucGenetier: It groups all tests in this class with those in LanguageServerTests and have them run sequentially
[https://xunit.net/docs/running-tests-in-parallel](https://xunit.net/docs/running-tests-in-parallel)

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
```diff
@@ -9,21 +9,52 @@
...
+
+                case TestResult.Skip:
+                    Skip.If(true, prefix + msg);
+                    break;
+            }
+        }
+
+        [InterpreterTheory]
+        [TxtFileData(@"ExpressionTestCases\TableSyntaxDoesntWrapRecords", null, nameof(InterpreterRunner))]
+        public void InterpreterTestCase_TableSyntaxDoesntWrapRecords(ExpressionTestCase testCase)
```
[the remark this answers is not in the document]
LucGenetier: We would have to introduce a locking mechanism to ensure the static flags are constant when a group of tests runs

LucGenetier:
> // Feature NoValueInRecordArrays set

Tests are run in parallel and this flag is a static

---
In reply to: [1171765261](https://github.com/microsoft/Power-Fx/pull/444#issuecomment-1171765261) [](http://example.com/codeflow?ancestors=1171765261,1171765261)

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/NoValueInRecordArrays/Index_NoValue.txt:1 in ca3594d. [](commit_id = ca3594d874f56edd84c3ef11a5554e761bc77159, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/FeatureFlags.cs
```diff
@@ -13,25 +13,42 @@ namespace Microsoft.PowerFx.Preview
...
-                }
+            set => _stringInterpolation = ProtectedSet(value);
+        }
 
-                _stringInterpolation = value;
+        /// <summary>
+        /// Enable Table syntax to not add "Value:" extra layer.
+        /// Added on 1st July 2022.
+        /// </summary>
+        public static bool TableSyntaxDoesntWrapRecords
```
MikeStall, line 38: >TableSyntaxDoesntWrapRecords [](http://example.com/codeflow?start=27&length=28)

Let's move this to a new class in Power FX config , and make it an instance.

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
MikeStall: Call is good, but we can still remove TableNode

src/libraries/Microsoft.PowerFx.Core/Entities/AllowedValuesMetadata.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.PowerFx.Core.Entities
 {
     internal sealed class AllowedValuesMetadata
     {
-        private static readonly DName ValueColumnName = new DName("Value");
```
MikeStall, line 14: >private [](http://example.com/codeflow?start=8&length=7)

we can use TableValue.ValueName instead.  Already public

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
```diff
@@ -53,6 +53,13 @@
...
+    <Compile Remove="ExpressionTestCases\TableSyntaxDoesntWrapRecords\**" />
+    <Content Remove="ExpressionTestCases\TableSyntaxDoesntWrapRecords\**" />
+    <EmbeddedResource Remove="ExpressionTestCases\TableSyntaxDoesntWrapRecords\**" />
+    <None Remove="ExpressionTestCases\TableSyntaxDoesntWrapRecords\**" />
+  </ItemGroup>
```
MikeStall: ?

src/tests/Microsoft.PowerFx.Core.Tests/PublicSurfaceTests.cs
```diff
@@ -28,41 +28,39 @@ public void Test()
             var allowed = new HashSet<string>()
```
MikeStall, line 28: >allowed [](http://example.com/codeflow?start=16&length=7)

Why all the changes in this file?

src/tests/Microsoft.PowerFx.Core.Tests/ThreadingTests.cs
```diff
@@ -49,6 +49,8 @@ public void CheckFxCore()
                 "DataTypeInfo._validDataFormatsPerDKind",
 
                 "FeatureFlags._stringInterpolation",
+                "FeatureFlags._tableSyntaxDoesntWrapRecords",
```
MikeStall, line 52: >_tableSyntaxDoesntWrapRecords [](http://example.com/codeflow?start=30&length=29)

remove?

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/LanguageServerTests.cs
```diff
@@ -3,23 +3,23 @@
...
 using Microsoft.PowerFx.Intellisense;
 using Microsoft.PowerFx.LanguageServerProtocol;
 using Microsoft.PowerFx.LanguageServerProtocol.Protocol;
-using Microsoft.PowerFx.Syntax;
 using Microsoft.PowerFx.Types;
 using Xunit;
 
 namespace Microsoft.PowerFx.Tests.LanguageServiceProtocol.Tests
 {
     public class LanguageServerTests : PowerFxTest
```
MikeStall, line 21: Do we still need changes to this file?

src/tests/Microsoft.PowerFx.Core.Tests/PublicSurfaceTests.cs
LucGenetier: Alpha order

src/tests/Microsoft.PowerFx.Core.Tests/ThreadingTests.cs
LucGenetier: done

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/LanguageServerTests.cs
LucGenetier: Yes, see lines 851-...

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Config/Feature.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+
+namespace Microsoft.PowerFx.Core.Public.Config
```
lesaltzm, line 6: Should this just be Microsoft.PowerFx?

src/libraries/Microsoft.PowerFx.Core/Public/Config/FeatureExtensions.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+namespace Microsoft.PowerFx.Core.Public.Config
+{
+    public static class FeatureExtensions
```
lesaltzm, line 6: Internal?

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -104,7 +104,24 @@ public override IntermediateNode Visit(TexlTableNode node, IRTranslatorContext c
                 Contracts.AssertValue(context);
 
                 var children = node.Children.Select(child => child.Accept(this, context)).ToArray();
-                return MaybeInjectCoercion(node, new TableNode(context.GetIRContext(node), children), context);
+                var irContext = context.GetIRContext(node);
```
lesaltzm, line 107: This is great!

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -104,7 +104,24 @@ public override IntermediateNode Visit(TexlTableNode node, IRTranslatorContext c
...
+                {
+                    // Let's add "Value:" here                    
+                    children = children.Select(childNode =>
+                        new RecordNode(
+                            new IRContext(childNode.IRContext.SourceContext, new RecordType().Add(TableValue.ValueName, childNode.IRContext.ResultType)),
+                            new Dictionary<DName, IntermediateNode>
+                            {
+                                { TableValue.ValueDName, childNode }
+                            }))
+                        .ToArray();
```
lesaltzm, line 119: ToList? CallNode winds up calling ToList on the array anyway.

src/libraries/Microsoft.PowerFx.Core/Public/Config/Feature.cs
LucGenetier: Ok

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -132,6 +133,8 @@ internal sealed partial class TexlBinding
         /// </summary>
         internal DName ThisRecordDefaultName => new DName("ThisRecord");
 
+        public Feature Feature { get; }
```
lesaltzm, line 136: Wondering if these flag properties should be named `Features`, assuming we'll have more than one in the future.

LucGenetier: yes
[https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1714](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1714)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/BaseRunner.cs
```diff
@@ -99,38 +99,40 @@ public abstract class BaseRunner
         /// <returns>status from running.</returns>
         public (TestResult result, string message) RunTestCase(TestCase testCase)
         {
-            var result = TestResult.Fail;
-            string message = null;
+            var t = Task.Factory.StartNew(
```
MikeStall, line 102: >var t = Task.Factory.StartNew( [](http://example.com/codeflow?start=12&length=30)

Talked offline - this has a nearly 2x perf improvement on running tests.

MikeStall (approved):
:shipit:

lesaltzm (approved):
🚢

PR #445, opened 2022-06-28
[Draft] Add support for Lazy Type Marshalling

### LINKED ISSUE #225: Recursive type marshalling
1. Recursive types. 
2. Deferred type marshalling. If a type has 700 fields, but we only use 2, would be ideal to not have to import fields we don't use. 

In the same way that RecordValue has a virtual GetField, it would be nice if RecordType had some virtuals.

lesaltzm:
Adds support for Lazy Type Marshalling to the C# host marshalling interfaces, and extends `DType` with `LazyRecord` and `LazyTable` DKinds. 

The intent here is not only to provide support for hosts to lazily marshal types, but also to provide a path towards unifying the one-off DKinds that do similar type expansion today: Control, DataEntity, Attachment. 

The lazy type support is implemented through `LazyTypeProvider`, which contains a callback to retrieve the type of each field. For the C# host marshallers, there's a layer of indirection so that evaluating this callback also registers the appropriate Value marshalling function for that field.

Closes #225


src/libraries/Microsoft.PowerFx.Core/Public/Types/AggregateType.cs
```diff
@@ -19,16 +19,12 @@ internal AggregateType(DType type)
 
         public FormulaType MaybeGetFieldType(string fieldName)
         {
-            // $$$ Better lookup
```
lesaltzm, line 22: A long-standing todo that needed to be addressed to support lazy field access.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1030,15 +1032,31 @@ public IEnumerable<IExpandInfo> GetExpands()
...
         public DType ToRecord(ref bool fError)
         {
             AssertValid();
 
             switch (Kind)
             {
+                case DKind.LazyRecord:
                 case DKind.Record:
                     return this;
+                case DKind.LazyTable:
```
lesaltzm, line 1058: A LazyTable/LazyRecord that are backed by the same schema use the same LazyTypeProvider, so switching between them is just constructing a new DType with the appropriate DKind.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1324,10 +1334,15 @@ public DType Add(ref bool fError, DPath path, DName name, DType type)
...
         {
             AssertValid();
-            Contracts.Assert(IsAggregate);
+            Contracts.Assert(IsAggregate || IsLazyType);
             Contracts.Assert(name.IsValid);
             type.AssertValid();
 
+            if (IsLazyType)
+            {
+                return new DType(LazyTypeProvider.AddField(name, type), IsTable);
```
lesaltzm, line 1343: Field addition and removal are delegated to LazyTypeProvider, so that the table shaping functions don't affect the laziness of the underlying table.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1339,50 +1354,12 @@ public DType Add(DName name, DType type)
...
         {
             AssertValid();
-            Contracts.Assert(IsAggregate);
+            Contracts.Assert(IsAggregate || IsLazyType);
             Contracts.Assert(typedName.IsValid);
 
             return Add(typedName.Name, typedName.Type);
         }
 
-        // Return a new type based on this, with additional named member fields of some specified types.
```
lesaltzm, line 1348: I'm going to extract this (and some other DType cleanup) to a separate PR, it's not directly tied to Lazy Type Support

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1869,7 +1809,7 @@ public virtual bool Accepts(DType type, out KeyValuePair<string, DType> schemaDi
                 case DKind.Record:
                 case DKind.File:
                 case DKind.LargeImage:
-                    if (Kind == type.Kind)
+                    if (Kind == type.Kind || type.IsLazyType)
```
lesaltzm, line 1812: This ensures that `DType.EmptyRecord.Accepts(LazyRecord)`, so all our functions that operate on arbitrary table/record types behave correctly. 
We may also want to add support for `SomeSpecificRecord.Accepts(LazyRecord)`, however, that's not currently supported for Relationships/Controls/... and so that's left out until there is demand for it.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -2840,60 +2786,6 @@ public override string ToString()
...
-        /// <remarks>The representation is an object with a required member 't', of type string
-        /// (which maps to the _kind property) and an optional member 'c', of type object, with
-        /// keys named on the children properties for this DType, and values representing their
-        /// respective JS type:
-        /// export interface IJsonFunctionDataDefinition {
-        ///     t: string;   // Type (maps to DType.Kind)
-        ///     c?: HashTable.IJsonFunctionDataDefinition; // optional children
-        /// }.
-        /// </remarks>
-        internal string ToJsType(Func<DName, DType, bool> shouldBeIncluded = null)
```
lesaltzm, line 2856: more cleanup to be extracted

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1193,6 +1199,10 @@ private bool TryGetTypeCore(DName name, out DType type)
                 case DKind.OptionSet:
                 case DKind.View:
                     return TypeTree.TryGetValue(name, out type);
+
+                case DKind.LazyRecord:
+                case DKind.LazyTable:
+                    return LazyTypeProvider.TryGetFieldType(name, out type);
```
lesaltzm, line 1205: This is the core piece that enables this feature. All field accesses go through this function, and are delegated to the LazyTypeProvider.

src/libraries/Microsoft.PowerFx.Core/Types/LazyTypeProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        private LazyTypeProvider(ILazyTypeMetadata metadata, ImmutableDictionary<DName, FieldTypeGetter> fieldTypeGetters)
+        {            
+            LazyTypeMetadata = metadata;
+            _fieldTypeGetters = fieldTypeGetters;
+        }
+
+        internal bool TryGetFieldType(DName name, out DType type)
+        {
+            if (_expandedFields.TryGetValue(name, out type))
```
lesaltzm, line 38: We cache field access that has already been expanded, otherwise, we have a callback to get the type of each field in the type

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/LazyMarshalledTypeMetadata.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx
+{
+    [ThreadSafeImmutable]
+    internal class LazyMarshalledTypeMetadata : ILazyTypeMetadata
```
lesaltzm, line 14: This maintains the relatively simple notion of identity for a C# marshalled object. For Relationships, e.g, this would be implemented as the table name and dataset name instead.

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/ObjectMarshaller.cs
```diff
@@ -14,31 +18,26 @@ namespace Microsoft.PowerFx
     [DebuggerDisplay("ObjMarshal({Type})")]
     public class ObjectMarshaller : ITypeMarshaller
```
lesaltzm, line 19: Open question here:
Should we always default to lazy marshalling for C# objects? There's a memory/perf cost to creating all those callbacks.

lesaltzm: Useful to do for this PR, since it allows our marshalling tests to catch a bunch of cases.

src/libraries/Microsoft.PowerFx.Core/Types/LazyTypeProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+            {
+                return true;
+            }
+            else if (_fieldTypeGetters.TryGetValue(name, out var getter))
+            {
+                type = getter()._type;
+                _expandedFields.Add(name, type);
```
MikeStall, line 45: >_expandedFields.Add [](http://example.com/codeflow?start=16&length=19)

Do we know this is only accessed from 1 thread?

src/tests/Microsoft.PowerFx.Interpreter.Tests/MarshalTests.cs
```diff
@@ -704,6 +709,8 @@ public void FailWithoutCustomMarshaller()
                     Data = "B"
                 }
             };
+
+            // $$$ TODO: This no longer throws because marshalling the type of Widget is lazy
```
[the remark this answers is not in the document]
MikeStall: This makes sense. 
We could just update the test to do a deep traversal, and then that would hit it, like

  Dump(cache.Marshal(obj))

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/ObjectMarshaller.cs
MikeStall: It would be good if we were always lazy.  I'm concerned a bifurcation where we have lazy vs. non lazy. Can we find a way to be efficient about it?

src/tests/Microsoft.PowerFx.Interpreter.Tests/MarshalTests.cs
```diff
@@ -157,10 +157,15 @@ public void TestRecursion()
             var result3 = engine.Eval("x.Next.Next.Data");
             Assert.Equal(30.0, ((NumberValue)result3).Value);
 
-            // Recursion not supported - gets truncated...
-            // https://github.com/microsoft/Power-Fx/issues/225
-            var result4 = engine.Check("x.Next.Next.Next.Next.Next.Next");
-            Assert.False(result4.IsSuccess);
+            // Arbitrary recursion supported!
+            // Practically capped here at 50 due to parser depth limit
+            var result4 = engine.Check(
```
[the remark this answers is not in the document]
MikeStall: yes!!!!

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -572,6 +600,8 @@ public bool IsActivityPointer
 
         public bool HasExpandInfo => ExpandInfo != null;
 
+        public bool IsLazyType => Kind == DKind.LazyRecord || Kind == DKind.LazyTable;
```
Shpakh-MSFT, line 603: >Kind == DKind.LazyRecord || Kind == DKind.LazyTable; [](http://example.com/codeflow?start=34&length=52)

This should probably be LazyTypeProvider != null

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -474,6 +478,28 @@ private DType(DKind kind, string namedValueKind)
             AssertValid();
         }
 
+        internal DType(LazyTypeProvider provider, bool isTable)
+        {
+            Contracts.AssertValue(provider);
+            
+            LazyTypeProvider = provider;
+            Kind = isTable ? DKind.LazyTable : DKind.LazyRecord;
```
Shpakh-MSFT, line 486: >isTable ? DKind.LazyTable : DKind.LazyRecord [](http://example.com/codeflow?start=19&length=44)

Should this be a property on type provider?
Kind = provider.TypeKind

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1093,8 +1096,11 @@ public DType ToTable(ref bool fError)
...
             switch (Kind)
             {
+                case DKind.LazyTable:
                 case DKind.Table:
                     return this;
+                case DKind.LazyRecord:
+                    return new DType(LazyTypeProvider, isTable: true);
                 case DKind.Record:
                 case DKind.DataEntity:
                 case DKind.Control:
```
Shpakh-MSFT, line 1106: Do we still need below check which was removed by your change?

    if (ExpandInfo != null)
                    {
                        return new DType(DKind.Table, ExpandInfo, TypeTree);
                Contracts.Assert(false, "Bad source kind for ToTable");
            }

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1328,6 +1338,11 @@ public DType Add(DName name, DType type)
             Contracts.Assert(name.IsValid);
             type.AssertValid();
 
+            if (IsLazyType)
+            {
+                return new DType(LazyTypeProvider.AddField(name, type), IsTable);
```
Shpakh-MSFT, line 1343: >return new DType(LazyTypeProvider.AddField(name, type), IsTable); [](http://example.com/codeflow?start=16&length=65)

Why do we need to create a new type here?

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1528,86 +1510,19 @@ public DType DropMulti(ref bool fError, DPath path, params DName[] rgname)
...
-
-            var tree = typeOuter.TypeTree.RemoveItems(ref fError, rgname);
-
-            return SetType(ref fError, path, new DType(typeOuter.Kind, tree, AssociatedDataSources, DisplayNameProvider));
-        }
-
-        // Drop everything but the specified names/fields from path's type, and return the resulting type.
-        // If a name/field (that was specified to be kept) is missing, we are returning a new type
-        // with the type for the missing field as Error and fError will be true.
-        public DType KeepMulti(ref bool fError, DPath path, params DName[] rgname)
```
Shpakh-MSFT, line 1542: >KeepMulti [](http://example.com/codeflow?start=21&length=9)

Why are we removing keepMulti and Split? Are they not used / bad patterns?

lesaltzm: Yes, unused. I left a comment above about  this. Pulling this cleanup to a separate PR

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
lesaltzm: No, reusing the same provider for records/tables of the same object is useful, so we don't need a new provider instance for something like `First(LazyTableType)`

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
lesaltzm: Oh, ouch. I missed that these implementations of toTable diverged for ExpandInfo. That's not great...

lesaltzm:
Closing, putting up new PRs shortly.

PR #447, opened 2022-06-28
Add Boolean_T and number variants

bryancroteau-MSFT:
Adds Boolean_T. Also adds both Boolean and Boolean_T number variants. Note that coercion cannot be used here.


MikeStall (approved):
:shipit:

PR #448, opened 2022-06-30
Enable .Net analyzers in main projects and fix issues

LucGenetier:


src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -2810,7 +2809,7 @@ public override void Visit(NumLitNode node)
 
             public DName GetLogicalNodeNameAndUpdateDisplayNames(DType type, Identifier ident, bool isThisItem = false)
             {
-                return GetLogicalNodeNameAndUpdateDisplayNames(type, ident, out var unused, isThisItem);
+                return GetLogicalNodeNameAndUpdateDisplayNames(type, ident, out var _, isThisItem);
```
lesaltzm, line 2812: can just be `out _` even

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -5708,13 +5704,13 @@ private bool IsRecordScopeFieldName(DName name, out Scope scope)
 
                     // If scope type is a data source, the node may be a display name instead of logical.
                     // Attempt to get the logical name to use for type checking
-                    if (!scope.SkipForInlineRecords && (DType.TryGetConvertedDisplayNameAndLogicalNameForColumn(scope.Type, name.Value, out var maybeLogicalName, out var tmp) ||
+                    if (!scope.SkipForInlineRecords && (DType.TryGetConvertedDisplayNameAndLogicalNameForColumn(scope.Type, name.Value, out var maybeLogicalName, out var _) ||
```
lesaltzm, line 5707: same for these.

src/libraries/Microsoft.PowerFx.Core/IR/InvalidCoercionException.cs
```diff
@@ -11,5 +11,14 @@ public InvalidCoercionException(string message)
             : base(message)
         {
         }
+
+        public InvalidCoercionException()
```
lesaltzm, line 15: Why add an empty constructor?

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -350,25 +342,19 @@ public List<Token> GetTokens(string text)
...
-                case TokKind.KeyOr:
-                case TokKind.As:
-                    result = true;
-                    break;
-                default:
-                    result = false;
-                    break;
-            }
-
+                TokKind.True or
```
lesaltzm, line 347: neat!

src/libraries/Microsoft.PowerFx.Core/Microsoft.PowerFx.Core.csproj
```diff
@@ -21,7 +21,13 @@
 		<Summary>The core library for all Microsoft Power Fx Compiler targets</Summary>
 		<GeneratePackageOnBuild>True</GeneratePackageOnBuild>
 		<GenerateNeutralResourcesLanguageAttribute>false</GenerateNeutralResourcesLanguageAttribute>
-	</PropertyGroup>
+    <EnableNETAnalyzers>True</EnableNETAnalyzers>
```
lesaltzm, line 24: nit: run formatter, looks like a mixed spaces/tabs issue

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -29,6 +30,7 @@ public class CheckResult : IOperationStatus
         /// null if unavailable.  
         /// This is only valid when <see cref="IsSuccess"/> is true.
         /// </summary>
+        [SuppressMessage("Usage", "CA2227: Collection properties should be read only", Justification = "n/a")]
```
lesaltzm, line 33: This probably should be readonly, actually. (or at least `internal set`)

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
```diff
@@ -28,6 +27,7 @@ public abstract class FormulaType
 
         public static FormulaType Number { get; } = new NumberType();
 
+        [SuppressMessage("Naming", "CA1720:Identifiers should not contain type names", Justification="n/a")]
```
lesaltzm, line 30: We may want to disable CA1720 generally, this product is a compiler, and it's reasonable for some identifiers to be type names

src/libraries/Microsoft.PowerFx.Transport.Attributes/TransportConverterAttribute.cs
```diff
@@ -2,14 +2,16 @@
...
 
 namespace Microsoft.AppMagic.Transport
 {
     /// <summary>
     /// Applied to a converter class. Provides a conversion for a custom type.
     /// </summary>
     [AttributeUsage(AttributeTargets.Class)]
-    public class TransportConverterAttribute : Attribute
+    [SuppressMessage("Design", "CA1019: Define accessors for attribute arguments", Justification = "n/a")]
+    public sealed class TransportConverterAttribute : Attribute
```
lesaltzm, line 14: technically, these are a breaking change, but it should be safe. I don't think any uses in PA-Client derive from any of these attributes, and I'd be really surprised if someone else relied on this package.

lesaltzm (commented):
Thanks for doing this!

src/libraries/Microsoft.PowerFx.Connectors/Microsoft.PowerFx.Connectors.csproj
```diff
@@ -14,7 +14,13 @@
 		<Title>Microsoft Power Fx Connector Support</Title>
 		<Description>Connectors</Description>
 		<EnforceCodeStyleInBuild>false</EnforceCodeStyleInBuild>
+		<EnableNETAnalyzers>True</EnableNETAnalyzers>
+		<AnalysisLevel>latest-all</AnalysisLevel>
 	</PropertyGroup>
+  
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
```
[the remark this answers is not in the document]
MikeStall: +1
By default, Debug and Release should behave the same.

src/libraries/Microsoft.PowerFx.Connectors/PowerPlatformConnectorClient.cs
```diff
@@ -51,12 +52,13 @@ public PowerPlatformConnectorClient(string endpoint, string environmentId, strin
             // Must set to allow callers to invoke SendAsync() via other helper methods.
             BaseAddress = new Uri("https://" + endpoint); // Uri.Parse will validate endpoint syntax. 
         }
-                
+
+        [SuppressMessage("Reliability", "CA2000: Dispose objects before losing scope", Justification = "Callers must manage response properly")]
```
MikeStall, line 56: >objects [](http://example.com/codeflow?start=57&length=7)

which object is it complaining about?  Is there actually a missing dispose in here?

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -827,7 +834,7 @@ private DPath GetFunctionNamespace(CallNode node, TexlVisitor visitor)
 
         internal bool TryGetDataQueryOptions(out DataSourceToQueryOptionsMap tabularDataQueryOptionsMap)
         {
-            return TryGetDataQueryOptions(Top, false, out tabularDataQueryOptionsMap);
```
MikeStall, line 830: >false [](http://example.com/codeflow?start=47&length=5)

why are these params removed?

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -678,7 +685,7 @@ internal bool TryGetDataQueryOptions(TexlNode node, bool forCodegen, out DataSou
             if (node.Kind == NodeKind.FirstName
                 && Rule.TexlNodeQueryOptions.Count > 1)
             {
-                if (!(Rule.Document.GlobalScope.GetTabularDataSource(node.AsFirstName().Ident.Name) is IExternalTabularDataSource tabularDs))
+                if (Rule.Document.GlobalScope.GetTabularDataSource(node.AsFirstName().Ident.Name) is not IExternalTabularDataSource tabularDs)
```
MikeStall, line 688: >not [](http://example.com/codeflow?start=101&length=3)

Are these style preferences or is there a rule now flagging this?

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4708,7 +4705,7 @@ public override bool PreVisit(CallNode node)
                 var overloads = LookupFunctions(funcNamespace, node.Head.Name.Value);
                 if (!overloads.Any())
                 {
-                    if (funcNamespace.ToString() != string.Empty)
+                    if (!string.IsNullOrEmpty(funcNamespace.ToString()))
```
MikeStall, line 4708: >IsNullOrEmpty [](http://example.com/codeflow?start=32&length=13)

this change the behavior on null.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -25,9 +25,7 @@
 using Microsoft.PowerFx.Syntax;
 
 namespace Microsoft.PowerFx.Core.Functions
-{
-    using FunctionInfo = Microsoft.PowerFx.Core.Functions.TransportSchemas.FunctionInfo;
```
MikeStall, line 29: >FunctionInfo [](http://example.com/codeflow?start=10&length=12)

why remove? this looks like it made the code cleaner since we repeat it several times.

src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/DelegationValidationStrategy.cs
```diff
@@ -63,14 +64,14 @@ protected void SuggestDelegationHint(TexlNode node, TexlBinding binding, ErrorRe
         {
             Contracts.AssertValue(node);
             Contracts.AssertValue(binding);
-            Contracts.Assert(suggestionKey == null || suggestionKey?.Key != string.Empty);
```
MikeStall, line 66: > string.Empty [](http://example.com/codeflow?start=75&length=13)

Change change from Empty to null?

src/libraries/Microsoft.PowerFx.Core/IR/InvalidCoercionException.cs
MikeStall: This matters - C# won't automatically create a default ctor if there are already other ctors present. 

So before, 
new InvalidCoercionException() // illegal, must call a required ctor

src/libraries/Microsoft.PowerFx.Core/Localization/ErrorResourceKey.cs
```diff
@@ -11,7 +12,7 @@ namespace Microsoft.PowerFx.Core.Localization
     /// Existing keys for error messages are split between here (for general document errors) and Strings.cs (for Texl errors).
     /// </summary>
     [ThreadSafeImmutable]
-    public struct ErrorResourceKey
+    public struct ErrorResourceKey : IEquatable<ErrorResourceKey>
```
MikeStall, line 15: >IEquatable<ErrorResourceKey> [](http://example.com/codeflow?start=37&length=28)

Did a rule flag to add this?

src/libraries/Microsoft.PowerFx.Core/Localization/StringResources.cs
```diff
@@ -27,7 +25,7 @@ internal static class StringResources
 
         private static readonly ThreadSafeResouceManager _resourceManager = new ThreadSafeResouceManager();
 
-        [ThreadSafeProtectedByLockAttribute("_errorResources")]
```
MikeStall, line 30: >_errorResources [](http://example.com/codeflow?start=45&length=15)

why removing this arg?

src/libraries/Microsoft.PowerFx.Core/Public/Types/BlankType.cs
```diff
@@ -15,9 +15,9 @@ internal BlankType()
         {
         }
 
-        public override void Visit(ITypeVisitor vistor)
+        public override void Visit(ITypeVisitor visitor)
```
MikeStall, line 18: >visitor [](http://example.com/codeflow?start=48&length=7)

wow, impressed it caught this.

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
```diff
@@ -22,7 +22,7 @@ public abstract class CollectionTableValue<T> : TableValue
         private readonly IReadOnlyList<T> _sourceIndex; // maybe null. supports index. 
         private readonly IReadOnlyCollection<T> _sourceCount; // maybe null. supports count;
 
-        public CollectionTableValue(RecordType recordType, IEnumerable<T> source)
+        internal CollectionTableValue(RecordType recordType, IEnumerable<T> source)
```
MikeStall, line 25: >internal [](http://example.com/codeflow?start=8&length=8)

why are we demoting the visibility?

src/libraries/Microsoft.PowerFx.Connectors/Microsoft.PowerFx.Connectors.csproj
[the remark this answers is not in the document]
LucGenetier: I disagree. Debug is for developement and Release should be used on build servers (PR build included) to generate final code.
Logically Release could be more strict.
That said, I've protected .Net analyzers and this variable with TURN_OFF_WARNING_AS_ERROR variable

src/libraries/Microsoft.PowerFx.Core/AssemblyProperties.cs
```diff
@@ -37,4 +38,5 @@
 [assembly: InternalsVisibleTo("Microsoft.PowerFx.Core.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
 [assembly: InternalsVisibleTo("Microsoft.PowerFx.Interpreter.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
 
+[assembly: CLSCompliant(false)]
```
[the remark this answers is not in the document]
LucGenetier: We're not CLS compliant
- use of underscore in public field names
- use of params in constructors
- use of uint
(probably several others)

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
LucGenetier: good catch

src/libraries/Microsoft.PowerFx.Core/IR/InvalidCoercionException.cs
LucGenetier: [https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1032](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1032)

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
LucGenetier: thanks :)

src/libraries/Microsoft.PowerFx.Core/Microsoft.PowerFx.Core.csproj
LucGenetier: done

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
LucGenetier: Works fine

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
LucGenetier: Agreed, removed

src/libraries/Microsoft.PowerFx.Connectors/Execution/OpenApiJsonSerializer.cs
```diff
@@ -122,5 +123,32 @@ internal override void EndSerialization()
                 EndObject();
             }
         }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!_disposedValue)
```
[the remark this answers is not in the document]
LucGenetier: Not strictly equivalent 

Changed to 
            if (!_wasDisabled)
            {
                if (disposing)
                {
                    _writer?.Dispose();
                    _stream?.Dispose();
                }

                _wasDisabled = true;
            }

src/libraries/Microsoft.PowerFx.Transport.Attributes/TransportConverterAttribute.cs
LucGenetier: agreed, I don't expect any problem with this change

LucGenetier:
Thanks ;)

---
In reply to: [1025336172](https://github.com/microsoft/Power-Fx/pull/448#issuecomment-1025336172) [](http://example.com/codeflow?ancestors=1025336172,1025336172)

src/libraries/Microsoft.PowerFx.Connectors/PowerPlatformConnectorClient.cs
LucGenetier: This test isn't super clever I think.
It identifies we create a disposable object (HttpRequestMessage) that isn't disposed. 
This is ok but expected in this case as we return it!...
The callers will have to deal with this object and call Dispose when necessary.
Hence the need to suppress that message here.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
LucGenetier: look at line 632 in this file and you'll that that forCodegen is never used, so we can remove it

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
LucGenetier: This is from [https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0083](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0083)
Of course this is not a strict requirement but it makes the code much easier to read: natural reading + removes the need for extra parenthesis

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
LucGenetier: well, not really
if funcNamespace is null, it was throwning
if .ToString() would return null, it would if ToString would return null
In that case ToString maps to ToDottedSyntax() which never returns null.
We're 100% safe here.

That said, having code that differentiates null and an empty string is a real bad practice or you need a super good reason.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
LucGenetier: This is [https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0065](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0065) which isn't enforced but isn't a good idea/good practice.

I've changed it to the line below (out of namespace block) to make it more readable (and with a different name to avoid a conflict)

using FuncInfo = Microsoft.PowerFx.Core.Functions.TransportSchemas.FunctionInfo;

src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/DelegationValidationStrategy.cs
LucGenetier: I don't understand your question and this code (looks completely wrong to me)

I've added two exclusions that this fix was supposed to mitigate
[https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1820](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1820)
[https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1508](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1508)

src/libraries/Microsoft.PowerFx.Core/Localization/ErrorResourceKey.cs
LucGenetier: Yes
[https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1815](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1815)

src/libraries/Microsoft.PowerFx.Core/Localization/StringResources.cs
LucGenetier: it's never used in the code

src/libraries/Microsoft.PowerFx.Core/Public/Types/BlankType.cs
LucGenetier: So do I
[https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1725](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1725)

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
LucGenetier: [https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1012](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1012)

src/libraries/Microsoft.PowerFx.Core/AssemblyProperties.cs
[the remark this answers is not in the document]
LucGenetier: If we DO want to be CLSCompliant, we need to review all (breaking) changes

src/libraries/Microsoft.PowerFx.Core/Localization/ErrorResourceKey.cs
```diff
@@ -21,5 +22,24 @@ public ErrorResourceKey(string key)
...
+        {
+            return obj is ErrorResourceKey key && Equals(key);
+        }
+
+        public override int GetHashCode()
+        {
+            return Key.GetHashCode();
+        }
+
+        public static bool operator ==(ErrorResourceKey left, ErrorResourceKey right) => left.Equals(right);
```
CarlosFigueiraMSFT, line 41: >left.Equals(right); [](http://example.com/codeflow?start=89&length=19)

Does this break (NRE) if left is null?

    ErrorResourceKey erk1 = MethodThatCanReturnNull();
    ErrorResourceKey erk2 = new ErrorResourceKey("sth");
    return erk1 == erk2;

src/libraries/Microsoft.PowerFx.Core/Localization/InvalidXmlException.cs
```diff
@@ -52,5 +52,14 @@ private static int GetLinePosition(XObject xobj)
...
         }
+
+        public InvalidXmlException()
+        {
+        }
+
+        public InvalidXmlException(string message, System.Exception innerException) 
+            : base(message, innerException)
+        {
+        }
```
CarlosFigueiraMSFT, line 63: Can you move them closer to the other .ctors?

src/libraries/Microsoft.PowerFx.Core/Public/Types/Serialization/FormulaTypeSchema.cs
```diff
@@ -40,6 +39,7 @@ public enum ParamType
         /// <summary>
         /// Optional. For Records and Tables, contains the list of fields.
         /// </summary>
+        [SuppressMessage("Usage", "CA2227: Collection properties should be read only", Justification = "n/a")]
```
CarlosFigueiraMSFT, line 42: >[SuppressMessage("Usage", "CA2227: Collection properties should be read only", Justification = "n/a")] [](http://example.com/codeflow?start=8&length=102)

Can we make it a readonlydictionary instead? It's set on FormulaTypeToSchemaConverter, but only when creating a new FormulaTypeSchema, so we can make it part of hte .ctor

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
CarlosFigueiraMSFT, line 25: >internal [](http://example.com/codeflow?start=8&length=8)

Is there a possibility of a partner team wanting to derive from this class? If so, `protected` would be more appropriate.

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
```diff
@@ -11,7 +12,8 @@ namespace Microsoft.PowerFx.Types
 {
     /// <summary>
     /// Represent a Record. Records have named fields which can be other values. 
-    /// </summary>
+    /// </summary>    
+    [SuppressMessage("Design", "CA1012:Abstract type should not have public constructors", Justification = "n/a")]
```
CarlosFigueiraMSFT, line 16: >[SuppressMessage("Design", "CA1012:Abstract type should not have public constructors", Justification = "n/a")] [](http://example.com/codeflow?start=4&length=110)

We can make the .ctor protected and remove this suppression

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Calendar.cs
```diff
@@ -25,7 +22,7 @@ public CalendarFunction(string functionInvariantName, TexlStrings.StringGetter f
 
         public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
         {
-            yield return new TexlStrings.StringGetter[] { };
+            yield return System.Array.Empty<TexlStrings.StringGetter>();
```
CarlosFigueiraMSFT, line 25: >System.Array [](http://example.com/codeflow?start=25&length=12)

[minor] `using System` and only use `Array.Empty<T>` here

Shpakh-MSFT:
>                 return errors.Select(x => ExpressionError.New(x)).ToArray();

ToArray is not needed here.

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/ExpressionError.cs:76 in 21e1887. [](commit_id = 21e1887e93ce83bb9fe03488711a664a47794bc6, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/NameCollisionException.cs
```diff
@@ -16,5 +14,14 @@ public NameCollisionException(string collidingDisplayName)
...
+
+        public NameCollisionException()
+        {
+        }
+
+        public NameCollisionException(string message, Exception innerException) 
+            : base(message, innerException)
+        {
+        }
     }
```
Shpakh-MSFT, line 26: Why do we need to add these? Can these be private?

src/libraries/Microsoft.PowerFx.Core/Public/Values/BooleanValue.cs
```diff
@@ -5,7 +5,7 @@
 using Microsoft.PowerFx.Core.IR;
 
 namespace Microsoft.PowerFx.Types
-{
+{
```
Shpakh-MSFT, line 8: >     [](http://example.com/codeflow?start=1&length=4)

Nit. Remove extra space.

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Intellisense.cs
```diff
@@ -32,11 +33,12 @@ public Intellisense(PowerFxConfig config, IEnumStore enumStore, IReadOnlyList<IS
             _suggestionHandlers = suggestionHandlers;
         }
 
+        [SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "n/a")]
```
CarlosFigueiraMSFT, line 36: >n/a [](http://example.com/codeflow?start=99&length=3)

We can have something like `No need to crash the process in case of exception during intellisense; instead we return an empty result set along with exception for client use.` here.

src/libraries/Microsoft.PowerFx.Core/Types/ValueTree.cs
```diff
@@ -22,6 +23,8 @@ private ValueTree(RedBlackNode<EquatableObject> root)
         }
 
         [Conditional("PARANOID_VALIDATION")]
+        [SuppressMessage("Performance", "CA1822: Mark members as static", Justification = "n/a")]
+
         internal void AssertValid()
```
CarlosFigueiraMSFT, line 28: [nit] remove empty line

src/libraries/Microsoft.PowerFx.Core/Localization/ErrorResourceKey.cs
LucGenetier: I corrected this even if this is not possible, per line 21

src/libraries/Microsoft.PowerFx.Core/Public/Types/Serialization/FormulaTypeSchema.cs
LucGenetier: This causes some trouble to Json desialization, so I'd avoid this change for now

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
LucGenetier: Doesn't work well with static Empty() and other methods

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Calendar.cs
LucGenetier: Good catch!

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseSuggestion.cs
```diff
@@ -231,7 +231,7 @@ public int CompareTo(IntellisenseSuggestion other)
                 return 1;
             }
 
-            return SortPriority == other.SortPriority ? Text.CompareTo(other.Text) : (int)(other.SortPriority - SortPriority);
+            return SortPriority == other.SortPriority ? string.Compare(Text, other.Text, StringComparison.Ordinal) : (int)(other.SortPriority - SortPriority);
```
MikeStall, line 234: >string.Compare [](http://example.com/codeflow?start=56&length=14)

this breaks.

PR #450, opened 2022-07-05
Port TestAllBuiltinFunctionsHaveParameterDescriptions from PAClient

bryancroteau-MSFT:
This test can fail in PAClient even though the change was in PowerFx and the PowerFx build passed. This change adds this test to PowerFx so that we can catch the failures here, instead of in the PAClient pipeline.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
@@ -16,7 +16,7 @@ namespace Microsoft.PowerFx.Core.Texl
     {
         public static IEnumerable<TexlFunction> BuiltinFunctionsLibrary => _library;
 
-        public static IEnumerable<TexlFunction> AllBuiltinFunctions => _library.Concat(_featureGateFunctions);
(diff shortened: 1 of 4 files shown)


src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
```diff
@@ -15,8 +16,11 @@ internal class BuiltinFunctionsCore
     {
         public static IEnumerable<TexlFunction> BuiltinFunctionsLibrary => _library;
 
+        public static IEnumerable<TexlFunction> AllBuiltinFunctions => _library.Concat(_featureGateFunctions);
```
lesaltzm, line 19: `internal`, maybe name `TestOnly_AllBuiltinFunctions`?

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
bryancroteau-MSFT: Done

lesaltzm (approved):

PR #456, opened 2022-07-06
Enable .Net analyzers

LucGenetier:
The following rules are currently turned off and will be activated in future PRs

 CA1001: Types that own disposable fields should be disposable 
 CA1003: Use generic event handler instances 
 CA1012: Abstract types should not have public constructors 
 CA1014: Mark assemblies with CLSCompliantAttribute 
 CA1031: Do not catch general exception types 
 CA1032: Implement standard exception constructors 
 CA1051: Do not declare visible instance fields 
 CA1052: Static holder types should be Static or NotInheritable 
 CA1064: Exceptions should be public 
 CA1066: Implement IEquatable when overriding Equals 
 CA1068: CancellationToken parameters must come last 
 CA1304: Specify CultureInfo 
 CA1305: Specify IFormatProvider 
 CA1307: Specify StringComparison for clarity 
 CA1309: Use ordinal StringComparison 
 CA1310: Specify StringComparison for correctness 
 CA1507: Use nameof in place of string 
 CA1508: Avoid dead conditional code 
 CA1711: Identifiers should not have incorrect suffix 
 CA1716: Identifiers should not match keywords 
 CA1724: Type names should not match namespaces 
 CA1725: Parameter names should match base declaration 
 CA1802: Use Literals Where Appropriate 
 CA1805: Do not initialize unnecessarily   
 CA1806: Do not ignore method results 
 CA1810: Initialize reference type static fields inline 
 CA1812: Avoid uninstantiated internal classes 
 CA1813: Avoid unsealed attributes 
 CA1814: Prefer jagged arrays over multidimensional 
 CA1815: Override equals and operator equals on value types 
 CA1819: Properties should not return arrays 
 CA1820: Test for empty strings using string length 
 CA1822: Mark members as static 
 CA1823: Avoid unused private fields 
 CA1825: Avoid zero-length array allocations 
 CA1826: Use property instead of Linq Enumerable method 
 CA1827: Do not use Count/LongCount when Any can be used 
 CA1829: Use Length/Count property instead of Enumerable.Count method 
 CA1830: Prefer strongly-typed Append and Insert method overloads on StringBuilder. 
 CA1834: Use StringBuilder.Append(char) for single character strings 
 CA2000: Dispose objects before losing scope 
 CA2007: Do not directly await a Task 
 CA2012: Use ValueTasks correctly 
 CA2201: Do not raise reserved exception types 
 CA2208: Instantiate argument exceptions correctly 
 CA2227: Collection properties should be read only 
 CA2242: Test for NaN correctly 
 CA5394: Do not use insecure randomness 
 IDE0004: Cast is redundant 
 IDE0005: Remove unnecessary using directives 
 IDE0052: Remove unread private member 
 IDE0054: Use compound assignment 
 IDE0059: Remove unnecessary value assignment 
 IDE0060: Remove unused parameters 
 IDE0066: Use 'switch' expression 
 IDE0083: Use pattern matching


lesaltzm (commented):
Why the distinction with Tests?

LucGenetier:
We want to be more flexible with less rules for Tests

---
In reply to: [1030444371](https://github.com/microsoft/Power-Fx/pull/456#issuecomment-1030444371) [](http://example.com/codeflow?ancestors=1030444371,1030444371)

lesaltzm (approved):

MikeStall (approved):
:shipit:

PR #458, opened 2022-07-06
Fix 'in' function

LucGenetier:


src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryOperators.cs
```diff
@@ -419,7 +419,7 @@ private static BooleanValue NotEqual(IRContext irContext, FormulaValue[] args)
 
                         if (!exact && rhs is StringValue strRhs)
                         {
-                            right = strRhs.ToLower();
```
MikeStall, line 422: wow! how did this compile before?

LucGenetier: FormulaValue

MikeStall (approved):

PR #462, opened 2022-07-07
[.Net Analyzers] CA2007 - ConfigureAwait missing

LucGenetier:
Enable CA2007 .Net Analyzer - Do not directly await a Task


src/tests/Microsoft.PowerFx.Connectors.Tests/BasicRestTests.cs
```diff
@@ -72,7 +72,7 @@ public async void ValidateHttpCalls(int apiFileNumber, string fxQuery, string ht
             var checkResult = engine.Check(fxQuery, options: _optionsPost);
```
lesaltzm, line 72: These changes seem kind of unnecessary in tests. From what I can tell, it's only a concern in library code because we don't know if the caller will block.

LucGenetier: This is a general good practice and a (tiny) perf improvement but I agree it's not a drama if we would be missing those in tests.

PR #463, opened 2022-07-07
[.Net Analyzers] Enable CAxx rules - Globalization

LucGenetier:
Enable .Net analyzer globalization rules
CA1303: Do not pass literals as localized parameters
CA1304: Specify CultureInfo
CA1305: Specify IFormatProvider
CA1307: Specify StringComparison for clarity
CA1309: Use ordinal StringComparison
CA1310: Specify StringComparison for correctness


src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -104,10 +104,11 @@ public static ParseResult Parse(string expressionText, ParserOptions options = n
         /// <param name="expressionText">the expression in plain text. </param>
         /// <param name="parameterType">types of additional args to pass.</param>
         /// <param name="options">parser options to use.</param>
+        /// <param name="cultureInfo">culture to use.</param>
         /// <returns></returns>
-        public CheckResult Check(string expressionText, RecordType parameterType = null, ParserOptions options = null)
+        public CheckResult Check(string expressionText, RecordType parameterType = null, ParserOptions options = null, CultureInfo cultureInfo = null)
```
lesaltzm, line 109: breaking change, could have a diamond dependency issue with PROSE when integrating with canvas. @ivanradicek, are you making use of this function? If so, we'll need to coordinate another update like the other week.

ivanradicek: I believe we're using the other overload that takes `ParseResult` as the first argument.

PR #464, opened 2022-07-07
Initial work on Color and Color_UO

bryancroteau-MSFT:
Normal Functions: Color, GUID
UntypedObject Functions: Color, GUID, TimeValue


src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -532,5 +533,40 @@ public static FormulaValue TrimEnds(IRContext irContext, StringValue[] args)
 
             return new StringValue(irContext, result);
         }
+
+        private static readonly Regex Hex6 = new Regex("^#[0-9a-fA-F]{6,6}$");
```
LucGenetier, line 537: add RegexOptions.Compiled

src/libraries/Microsoft.PowerFx.Interpreter/Microsoft.PowerFx.Interpreter.csproj
```diff
@@ -32,6 +32,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="System.Drawing.Common" Version="6.0.0" />
```
MikeStall, line 35: > <PackageReference Include="System.Drawing.Common" Version="6.0.0" /> [](http://example.com/codeflow?start=3&length=69)

We need to be wary of package references...

Is this just for System.Drawing.ColorTranslator?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -545,6 +565,26 @@ internal static partial class Library
                     returnBehavior: ReturnBehavior.ReturnBlankIfAnyArgIsBlank,
                     targetFunction: ForAll)
             },
+            {
+                BuiltinFunctionsCore.GUIDPure,
```
lesaltzm, line 569: There's a tricky bit here: Guids are implicitly coercible to strings, but we don't mark that in the IR. We tried doing this earlier, but had to back it out due to compat concerns with Canvas: #240 / #245

Could be fine right now to be limited, but note that we'll have pretty big gaps, test cases like:
>> "TheGuidIs: " & GUID("SOME-VALID-GUID")

pass in JS, not C#

src/libraries/Microsoft.PowerFx.Interpreter/Microsoft.PowerFx.Interpreter.csproj
MikeStall, line 35: Can we avoid this new dependency?

bryancroteau-MSFT: Removed

bryancroteau-MSFT: Removed

MikeStall (approved):

PR #465, opened 2022-07-08
Remove max selects limit

kartikkr11:
Intention is to have the limit on selects handled on the client rather than the server in case we have a large url generated.
This change would help with an edge case where we don't see lookup fields in player when we have a large number of selects.


lesaltzm (dismissed):
Nice!

MikeStall (approved):

lesaltzm (approved):

PR #469, opened 2022-07-09
Fix a breaking change from #444.

MikeStall:
Fix a binary breaking change from #444 
 

Adding a default param to a ctor is a binary breaking change.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
@@ -50,9 +50,27 @@ namespace Microsoft.PowerFx
         /// Initializes a new instance of the <see cref="PowerFxConfig"/> class.        
         /// </summary>
         /// <param name="cultureInfo">Culture to use.</param>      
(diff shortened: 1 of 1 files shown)


src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -46,15 +46,33 @@ private PowerFxConfig(CultureInfo cultureInfo, EnumStoreBuilder enumStoreBuilder
...
+            : this(cultureInfo, new EnumStoreBuilder().WithRequiredEnums(BuiltinFunctionsCore.BuiltinFunctionsLibrary), features)
         {
-        }
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PowerFxConfig"/> class.        
+        /// </summary>
+        /// <param name="features">Features to use.</param>
+        public PowerFxConfig(Features features)
```
MikeStall, line 72: Needed this overload because:
- required params can't come after optional params 
- we had callers just passing {features}

lesaltzm (approved):

PR #470, opened 2022-07-11
Functions now have a maximum call depth at runtime.

MalekiRe:
Configurable through PowerFxConfig, which is also a breaking change because the config change is public.


src/libraries/Microsoft.PowerFx.Core/IR/IRNodeVisitor.cs
```diff
@@ -7,36 +7,36 @@ namespace Microsoft.PowerFx.Core.IR
 {
     internal abstract class IRNodeVisitor<TResult, TContext>
     {
-        public abstract TResult Visit(TextLiteralNode node, TContext context);
+        public abstract TResult Visit(TextLiteralNode node, TContext context, StackMarker stackMarker);
```
bryancroteau-MSFT, line 10: This is the wrong approach. We should not be modifying this interface in this way.

bryancroteau-MSFT (changes requested):

src/libraries/Microsoft.PowerFx.Core/IR/IRNodeVisitor.cs
MalekiRe: Changing this over to the proper way to implement the interface.

src/libraries/Microsoft.PowerFx.Core/IR/StackMarker.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            if (_depth >= _maxCallDepth)
+            {
+                throw new MaxCallDepthException();
+            }
+
+            return new StackMarker(_depth + 1, _maxCallDepth);
+        }
+
+        public StackMarker(int depth, int maxCallDepth)
```
bryancroteau-MSFT, line 23: Is there anywhere where depth is not 0? We might want to rename depth to currentDepth. We could swap these parameters, then make currentDepth = 0 a default value.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -130,10 +130,12 @@ private async Task<FormulaValue> TryHandleSet(CallNode node, SymbolContext conte
             return CommonErrors.UnreachableCodeError(node.IRContext);
         }
 
-        public override async ValueTask<FormulaValue> Visit(CallNode node, SymbolContext context)
+        public override async ValueTask<FormulaValue> Visit(CallNode node, (SymbolContext, StackMarker) context)
         {
             CheckCancel();
-            
+
+            context.Item2 = context.Item2.Inc();
```
bryancroteau-MSFT, line 137: Item1 and Item2 are not very meaningful. It might be a good idea to make a new class, EvalVisitorContext, which contains these 2 sub-contexts as properties.

bryancroteau-MSFT (changes requested):

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
MalekiRe: Fixed

src/libraries/Microsoft.PowerFx.Core/IR/StackMarker.cs
MalekiRe: Changed to this.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitorContext.cs
```diff
@@ -0,0 +1,22 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.PowerFx.Core.IR;
+
+namespace Microsoft.PowerFx.Interpreter
+{
+    internal struct EvalVisitorContext
+    {
+        public SymbolContext SymbolContext;
```
bryancroteau-MSFT, line 13: these should be properties

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitorContext.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.PowerFx.Core.IR;
+
+namespace Microsoft.PowerFx.Interpreter
```
bryancroteau-MSFT, line 9: EvalVisitor is in the Microsoft.PowerFx namespace. We should make this class consistent with EvalVisitor.

src/libraries/Microsoft.PowerFx.Interpreter/MaxCallDepthException.cs
```diff
@@ -0,0 +1,35 @@
...
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.PowerFx.Interpreter
+{
+    [Serializable]
+    internal class MaxCallDepthException : Exception
```
bryancroteau-MSFT, line 11: Why do we have 2 copies of this exception class?

bryancroteau-MSFT (changes requested):

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -10,6 +10,7 @@
 using Microsoft.PowerFx.Core.Texl;
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Interpreter;
```
bryancroteau-MSFT, line 13: Not needed anymore. Same for other files.

bryancroteau-MSFT (changes requested):

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
MalekiRe: Fixed

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitorContext.cs
MalekiRe: Is now consistent with EvalVisitor

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitorContext.cs
MalekiRe: Now are properties.

src/libraries/Microsoft.PowerFx.Interpreter/Values/LambdaFormulaValue.cs
```diff
@@ -6,6 +6,7 @@
 using System.Threading.Tasks;
 using Microsoft.PowerFx.Core.IR;
 using Microsoft.PowerFx.Core.IR.Nodes;
+using Microsoft.PowerFx.Interpreter;
```
bryancroteau-MSFT, line 9: Missed one

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
```diff
@@ -7,6 +7,7 @@
 using System.Threading.Tasks;
 using Microsoft.PowerFx.Core.Functions;
 using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Interpreter;
```
bryancroteau-MSFT, line 10: Missed one

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -550,6 +550,26 @@ public void EmptyEnumStoreTest()
...
+                MaxCallDepth = 5
+            };
+            var recalcEngine = new RecalcEngine(config);
+            try
+            {
+                await recalcEngine.EvalAsync("Abs(Abs(Abs(Abs(Abs(Abs(1))))))", new System.Threading.CancellationToken());
+                Assert.True(false);
+            }
+            catch (MaxCallDepthException)
+            {
```
bryancroteau-MSFT, line 567: In the body of this set a boolean flag to true. For example callDepthCaught = true. Then assert that it is true later.

bryancroteau-MSFT (changes requested):

src/libraries/Microsoft.PowerFx.Core/Utils/StackMarker.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+namespace Microsoft.PowerFx.Core.Utils
+{
+    internal struct StackMarker
```
MikeStall, line 6: >StackMarker [](http://example.com/codeflow?start=20&length=11)

Can you add a class comment?
I like the abstraction :)
And struct is the way to go (not class) since it's used as a stack parameter.

src/libraries/Microsoft.PowerFx.Core/Utils/StackMarker.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        private readonly int _depth;
+        private readonly int _maxCallDepth;
```
MikeStall, line 9: nit - this could actually be a single counter rather than 2. (could be relevant since we're passing as parameters on a stack). 

Start at _maxCallDepth and count towards 0. Fail if < 0 .

bryancroteau-MSFT: That might make it harder to debug right?

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -550,6 +550,29 @@ public void EmptyEnumStoreTest()
...
+        {
+            bool test;
+            var config = new PowerFxConfig(null)
+            {
+                MaxCallDepth = 5
+            };
+            var recalcEngine = new RecalcEngine(config);
+            try
+            {
+                await recalcEngine.EvalAsync("Abs(Abs(Abs(Abs(Abs(Abs(1))))))", new System.Threading.CancellationToken());
```
MikeStall, line 564: >new System.Threading.CancellationToken() [](http://example.com/codeflow?start=80&length=40)

nit:
CancellationToken.None

src/libraries/Microsoft.PowerFx.Core/Utils/StackMarker.cs
MalekiRe: That is **smart** why did I not think of that.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -550,6 +550,29 @@ public void EmptyEnumStoreTest()
...
+        [Fact]
+        public async void MaxRecursionDepthTest()
+        {
+            bool test;
+            var config = new PowerFxConfig(null)
+            {
+                MaxCallDepth = 5
+            };
+            var recalcEngine = new RecalcEngine(config);
+            try
```
MikeStall, line 562: >try [](http://example.com/codeflow?start=12&length=3)

nit:
Can you do
Assert.Throws<MaxCallDepthException>  ( () => ...

bryancroteau-MSFT: That might be better than the boolean assert

src/libraries/Microsoft.PowerFx.Core/Utils/StackMarker.cs
MikeStall, line 6: >StackMarker [](http://example.com/codeflow?start=20&length=11)

Nit - should this be in Interpreter project?

bryancroteau-MSFT: It should be in the same namespace as the exception. If neither class is used in this namespace, then both can be moved to the Interpreter project in the same namespace as EvalVisitor.

bryancroteau-MSFT: I think this class ended up in this namespace because it was being used in the Interface (an earlier PR comment fixed that problem however, so this might be in the wrong place)

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -87,7 +96,7 @@ public static IExpression CreateEvaluatorDirect(CheckResult result)
             result.ThrowOnErrors();            
 
             (var irnode, var ruleScopeSymbol) = IRTranslator.Translate(result._binding);
-            return new ParsedExpression(irnode, ruleScopeSymbol);
+            return new ParsedExpression(irnode, ruleScopeSymbol, new StackMarker(20));
```
MikeStall, line 99: >20 [](http://example.com/codeflow?start=81&length=2)

nit - should share this value.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -550,6 +550,29 @@ public void EmptyEnumStoreTest()
...
+                test = false;
+            }
+            catch (MaxCallDepthException)
+            {
+                test = true;
+            }
+
+            Assert.True(test);
+            await recalcEngine.EvalAsync("Abs(Abs(Abs(Abs(Abs(1)))))", new System.Threading.CancellationToken());
+        }
```
MikeStall, line 574: Can you try this test:

Sum(
  Sum(Sum(1),1),
  Sum(Sum(1),1),
  Sum(Sum(1),1),
)

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -69,7 +70,15 @@ private protected override INameResolver CreateResolver(PowerFxConfig alternateC
         /// <inheritdoc/>
         protected override IExpression CreateEvaluator(CheckResult result)
```
MikeStall, line 71: >CreateEvaluator [](http://example.com/codeflow?start=39&length=15)

nit: can this still share code with CreateEvaluatorDirect ?

MalekiRe: Would require a breaking change to CreateEvaluatorDirect. Do we wanna do that?

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
MalekiRe: It passes, if I remove the trailing comma.

MikeStall (changes requested):
:clock1:

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
MalekiRe: Second one fails with
Message: 
System.InvalidOperationException : Errors: Error 20-21: Expected operator. We expect an operator such as +, *, or & at this point in the formula.
Error 41-42: Expected operator. We expect an operator such as +, *, or & at this point in the formula.
Error 62-63: Expected operator. We expect an operator such as +, *, or & at this point in the formula.
Error 83-84: Expected operator. We expect an operator such as +, *, or & at this point in the formula.
Error 104-105: Expected operator. We expect an operator such as +, *, or & at this point in the formula.
Error 125-126: Expected operator. We expect an operator such as +, *, or & at this point in the formula.
Error 126-126: Expected an operand. The formula or expression expects a valid operand. For example, you can add the operand '2' to the expression ' 1 +_' so that the result is '3'. Or, you can add the operand "there" to the expression '"Hi "& _ ' so that the result is 'Hi there'.
Error 125-126: Expected operator. We expect an operator such as +, *, or & at this point in the formula.
Error 104-105: Expected operator. We expect an operator such as +, *, or & at this point in the formula.
Error 83-84: Expected operator. We expect an operator such as +, *, or & at this point in the formula.
Error 62-63: Expected operator. We expect an operator such as +, *, or & at this point in the formula.
Error 41-42: Expected operator. We expect an operator such as +, *, or & at this point in the formula.
Error 20-21: Expected operator. We expect an operator such as +, *, or & at this point in the formula.

src/libraries/Microsoft.PowerFx.Core/Utils/StackMarker.cs
MikeStall: Both assemblies contribute to the same Namespace, so we have freedom here.

src/libraries/Microsoft.PowerFx.Core/Utils/MaxCallDepthException.cs
```diff
@@ -0,0 +1,35 @@
...
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.PowerFx.Core.Utils
+{
+    [Serializable]
+    internal class MaxCallDepthException : Exception
```
MikeStall, line 11: >MaxCallDepthException [](http://example.com/codeflow?start=19&length=21)

What did we settle on for SDK model here ... 

Does Engine.Eval()  throw this exception (and hosts catch it)? Or does it get converter to some runtime error at top-level?

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -179,7 +181,7 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, SymbolContext
             {
                 if (FuncsByName.TryGetValue(func, out var ptr))
                 {
-                    var result = await ptr(this, childContext, node.IRContext, args);
+                    var result = await ptr(this, new EvalVisitorContext(childContext, context.StackMarker), node.IRContext, args);
```
MikeStall, line 184: >StackMarker [](http://example.com/codeflow?start=94&length=11)

Can this be:
context.StackMarker.Inc() 


And then remove this above:
  context.StackMarker = context.StackMarker.Inc();

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -130,10 +130,12 @@ private async Task<FormulaValue> TryHandleSet(CallNode node, SymbolContext conte
             return CommonErrors.UnreachableCodeError(node.IRContext);
         }
 
-        public override async ValueTask<FormulaValue> Visit(CallNode node, SymbolContext context)
+        public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorContext context)
         {
             CheckCancel();
-            
+
+            context.StackMarker = context.StackMarker.Inc();
```
MikeStall, line 137: >ontext.StackMarker = context.StackMarker.Inc(); [](http://example.com/codeflow?start=13&length=47)

Can we avoid this? (see suggestion below) 

The problem is that this is modifying a heap-based object, which means we have to reason about reuse, concurrency, etc.

src/libraries/Microsoft.PowerFx.Core/Utils/StackMarker.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        internal StackMarker Inc()
```
LucGenetier, line 11: >Inc [](http://example.com/codeflow?start=29&length=3)

Increment

src/libraries/Microsoft.PowerFx.Core/Utils/StackMarker.cs
LucGenetier: Could we name this StackDepthCounter or anything more meaningful?

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
LucGenetier: Probably needs EnableExpressionChaining flag

src/libraries/Microsoft.PowerFx.Core/Utils/MaxCallDepthException.cs
MalekiRe: We decided that it would get converted to a runtime error at the top level, top of the EvalAsync function. Reimplemented now.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
MalekiRe: Changed to a EvalVisitorContext#IncrementStackDepthCounter

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
MalekiRe: Changed instead to use EvalVisitorContext#IncrementStackDepthCounter everytime it as passed as a parameter in this function. Doesn't mutate state instead creates a new version of EvalVisitorContext with same reference to SymbolContext but with an entirely new StackDepthCounter.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
MalekiRe: Just used Sum for test.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
MalekiRe: Value is now shared in an internal readonly static int in PowerFxConfig#DefaultMaxCallDepth.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
MalekiRe: No longer necessary as I implemented the top level catch at EvalAsync to instead return an Error.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
MalekiRe: No longer necessary.

src/libraries/Microsoft.PowerFx.Core/Utils/StackMarker.cs
MalekiRe: Renamed to StackDepthCounter.

src/libraries/Microsoft.PowerFx.Interpreter/StackDepthCounter.cs
```diff
@@ -0,0 +1,28 @@
...
+// Licensed under the MIT license.
+
+namespace Microsoft.PowerFx.Interpreter
+{
+    /// <summary>
+    /// This keeps track of how many calls deep we are in a function.
+    /// </summary>
+    internal struct StackDepthCounter
+    {
+        private readonly int _depth;
```
MikeStall, line 11: >_depth [](http://example.com/codeflow?start=29&length=6)

nit: Rename to 
_remaining? 

This is counting down, not counting up.

MalekiRe: Also going to renaming the function calls to "Decrement" then instead of "Increment"

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -155,15 +155,15 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, SymbolContext
 
                 if (!isLambda)
                 {
-                    args[i] = await child.Accept(this, context);
+                    args[i] = await child.Accept(this, context.IncrementStackDepthCounter());
```
MikeStall, line 158: >IncrementStackDepthCounter [](http://example.com/codeflow?start=63&length=26)

Yes!! This is much easier to prove correctness.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -179,7 +179,7 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, SymbolContext
             {
                 if (FuncsByName.TryGetValue(func, out var ptr))
                 {
-                    var result = await ptr(this, childContext, node.IRContext, args);
+                    var result = await ptr(this, new EvalVisitorContext(childContext, context.StackDepthCounter).IncrementStackDepthCounter(), node.IRContext, args);
```
MikeStall, line 182: >new EvalVisitorContext(childContext, context.StackDepthCounter).IncrementStackDepthCounter() [](http://example.com/codeflow?start=49&length=92)

nit: consider adding a overload IncrementStackDepthCounter so we can just write: 

                    var result = await ptr(this, context.IncrementStackDepthCounter(childContext), node.IRContext, args);

src/libraries/Microsoft.PowerFx.Interpreter/StackDepthCounter.cs
MalekiRe: Actually no i'm not.

src/libraries/Microsoft.PowerFx.Interpreter/MaxCallDepthException.cs
MikeStall, line 11: >MaxCallDepthException [](http://example.com/codeflow?start=19&length=21)

nit: I'd add a comment that this is an interpreter internal excpetion and should be converted to a ErrorValue before returning to the host. 

Even consider adding the "convert to error value" as a helper method on this class to make that very clear.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -159,9 +166,14 @@ public async Task<FormulaValue> EvalAsync(string expressionText, CancellationTok
...
             var check = Check(expressionText, (RecordType)parameters.IRContext.ResultType, options);
             check.ThrowOnErrors();
-
-            var newValue = await check.Expression.EvalAsync(parameters, cancel);
-            return newValue;
+            try
+            {
+                return await check.Expression.EvalAsync(parameters, cancel);
+            }
+            catch (MaxCallDepthException)
```
MikeStall, line 173: >catch (MaxCallDepthException) [](http://example.com/codeflow?start=12&length=29)

This should be deeper - in the ParsedExpression.EvalAsync  call directly.  There are ways to call ParsedExpression that don't go through Engine.EvalAsync.  (such as getting it from Check).

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -159,9 +166,14 @@ public async Task<FormulaValue> EvalAsync(string expressionText, CancellationTok
...
+            {
+                return CommonErrors.MaxCallDepth(new IRContext(new Syntax.Span(0, expressionText.Length - 1), FormulaType.BindingError));
```
MikeStall, line 175: > return CommonErrors.MaxCallDepth(new IRContext(new Syntax.Span(0, expressionText.Length - 1), FormulaType.BindingError)); [](http://example.com/codeflow?start=15&length=122)

move as helper on MaxCallDepthException

Also, if we're passing in a source context, we should have a text that verifies the source span is what we expect. .

MikeStall, line 175: > FormulaType.BindingError [](http://example.com/codeflow?start=109&length=25)

This is the wrong error type. 

Use check.ReturnType. 

(and add a test :) )

MalekiRe: Unnecessary now that we have access to IRContext.

MalekiRe: Unnecessary now that we have access to IRContext.

src/libraries/Microsoft.PowerFx.Interpreter/MaxCallDepthException.cs
MalekiRe: Done.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
MalekiRe: Done.

src/libraries/Microsoft.PowerFx.Interpreter/StackDepthCounter.cs
MalekiRe: Done.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitorContext.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Text;
+using Microsoft.PowerFx.Interpreter;
+
+namespace Microsoft.PowerFx
+{
+    internal struct EvalVisitorContext
+    {
+        public SymbolContext SymbolContext { get; set; }
+
+        public StackDepthCounter StackDepthCounter { get; set; }
```
lesaltzm, line 15: Why set? For both of these.

MalekiRe: Removed.

lesaltzm (approved):

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1365,28 +1365,28 @@ public static FormulaValue IsNumeric(IRContext irContext, FormulaValue[] args)
             return new BooleanValue(irContext, arg0 is NumberValue);
         }
 
-        public static async ValueTask<FormulaValue> With(EvalVisitor runner, SymbolContext symbolContext, IRContext irContext, FormulaValue[] args)
+        public static async ValueTask<FormulaValue> With(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
```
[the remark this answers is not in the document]
LucGenetier: (Other use case) I was in need to share/add Features to all functions. A non static approach would be ideal. I found an alternative for what I'm working on but this need might come back.

Now, for this case, we would instantiate this Library class for every Eval and this context is no more than ++ or -- as you enter or leave one of these functions if I'm understanding this well. We are creating tons of objects for a simple counter...

I'm not understanding why this context is different for each function

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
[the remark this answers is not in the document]
lesaltzm: It also contains the current scope symbols

PR #476, opened 2022-07-15
[Draft] Add telemetry hooks to PowerFx

lesaltzm:
Adds telemetry capabilities to PowerFx. Hosts can construct a logger with PowerFx statically, which will receive log calls from within PowerFx that they can forward to whatever telemetry provider(s) they're already making use of.  

There are two main decision points here:
1. Global Singleton vs passing Logger to all methods.
2. ILogger [(.NET Extension)](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.ilogger?view=dotnet-plat-ext-6.0) vs custom interface

For 1, integrating this with Canvas and Dataverse is much easier with the singleton approach, although it may make tracking scopes trickier depending on the host usage. 

For 2, we would need to add a new nuget dependency to use the .NET extension, and going with a custom interface also allows us to integrate in hosts not using ILogger  more easily.


src/libraries/Microsoft.PowerFx.Core/Logging/Log.cs
```diff
@@ -0,0 +1,93 @@
...
+            var scenario = new LoggingScenario(scenarioName);
+            _logger?.StartScenario(scenario.ScenarioName, scenario.ScenarioInstance, serialized);
+            return scenario;
+        }
+
+        internal void EndScenario(LoggingScenario scenario, object customDimensions = null)
+        {
+            scenario.CloseScenario();
+            var serialized = customDimensions != null ? JsonSerializer.Serialize(customDimensions) : string.Empty;
+            _logger?.EndScenario(scenario.ScenarioName, scenario.ScenarioInstance, serialized);
```
Shpakh-MSFT, line 63: We probably should enforce some basic schema (Not sure how) for logged content. 

For example. 
timestamp
session id/request id, 
operation Id
etc.

lesaltzm: IMO, those are things that probably should get added by the host in their implementation of IPowerFxLogger. We don't really know much about what context we're running in within the Power Fx code, and every host might have different concepts of sessions/operations. 

For MSFT hosts, we can work with them on that, for others, it depends on if/how the hosts would want to make use of that telemetry.

Shpakh-MSFT: We should discuss how we want to track the usage. Having a consistent schema for MS internal teams will make it easy to track usage.

lesaltzm:
Logging in this style from PowerFx may not be the right directon. Providing extensions on CheckResult and other APIs to get consistently formatted telemetry may be a better approach.

src/libraries/Microsoft.PowerFx.Core/Logging/DebugLogger.cs
```diff
@@ -0,0 +1,63 @@
...
+                Log("Error", message);
+            }
+
+            public void Exception(Exception exception, string message)
+            {
+                Contracts.AssertValue(exception);
+                Log("Exception", $"{message}\n{exception}");
+            }
+
+            public void StartScenario(string scenarioName, Guid scenarioInstance, string customDimensions)
```
MikeStall, line 39: >StartScenario [](http://example.com/codeflow?start=24&length=13)

Would power fx ever actually call this? It's the host that knows the scenarios and determines the logging. 

Ie, why would the binder ever initiate logging.  For comparison, other "pure" SDKs, like Json.Net,  don't log.

src/libraries/Microsoft.PowerFx.Core/Logging/Log.cs
MikeStall: +1 to both points.  But I think we need to talk with the hosts first to make sure there's a e2e plan here - including that we can get meaningful data from wherever they're storing it.

PR #479, opened 2022-07-18
Separate out BinaryOp IR Matrix to be sharable with JS code generator

lesaltzm:
Today, the Coercion Matrix is shared with the JS code generation logic, without going through the IR. This separates out the Binary Op matrix as well, and enables sharing that with JS. 
This makes the eventual work of converging IR and JS easier, and also will make #380 easier to integrate.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/IR/BinaryOpMatrix.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/IR/BinaryOpMatrix.cs
@@ -1,0 +1,393 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+using Microsoft.PowerFx.Core.Binding;
+using Microsoft.PowerFx.Core.IR.Nodes;
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Syntax;
+
+namespace Microsoft.PowerFx.Core.IR
+{
+    internal class BinaryOpMatrix
+    {
+        public static BinaryOpKind GetBinaryOpKind(Syntax.BinaryOpNode node, TexlBinding binding)
+        {
+            var parsedBinaryOp = node.Op;
+            var leftType = binding.GetType(node.Left);
+            var rightType = binding.GetType(node.Right);
+
+            return parsedBinaryOp switch
+            {
+                BinaryOp.In or BinaryOp.Exactin => GetInOp(parsedBinaryOp, leftType, rightType),
+                BinaryOp.Power => BinaryOpKind.Power,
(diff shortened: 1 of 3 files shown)


src/libraries/Microsoft.PowerFx.Core/IR/Nodes/BinaryOpKind.cs
```diff
@@ -78,5 +80,16 @@ internal enum BinaryOpKind
         DynamicGetField,
 
         // And, Or, Pow, Concatenate get represented as FunctionNodes with lambdas to handle short-circuiting
+        // Included here to make the matrix cleaner, should not be generated in IR.
```
lesaltzm, line 83: Does this seem reasonable, or should we implement a separate enum that covers the matrix, and then convert it to the binaryopkind used in the IR (like we do for CoercionKind->UnaryOpKind)

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/IR/BinaryOpMatrix.cs
```diff
@@ -0,0 +1,393 @@
...
+                case DKind.Date:
+                    if (rightType == DType.DateTime || rightType == DType.Date)
+                    {
+                        // Date + '-DateTime' => in days
+                        // Date + '-Date' => in days
+
+                        // Binding produces this as '-Date'. This should be cleaned up when we switch to a proper sub op. 
+                        if (node.Right is not Syntax.UnaryOpNode { Op: UnaryOp.Minus })
+                        {
+                            throw new NotSupportedException();
```
bryancroteau-MSFT, line 280: Why are Exceptions being used in this method? The other methods don't throw exceptions, and instead return Invalid.

src/libraries/Microsoft.PowerFx.Core/IR/Nodes/BinaryOpKind.cs
```diff
@@ -5,6 +5,8 @@ namespace Microsoft.PowerFx.Core.IR.Nodes
 {
     internal enum BinaryOpKind
     {
+        Invalid,
```
bryancroteau-MSFT, line 8: Are we ever serializing these or otherwise using the ordinal values of this enum?

src/libraries/Microsoft.PowerFx.Core/IR/BinaryOpMatrix.cs
lesaltzm: Preserving existing behavior, but these scenarios are not representable from the parser/binder. Invalid is not valid, but still representable with errors from the parser/binder.

MikeStall (approved):
:shipit:

bryancroteau-MSFT (approved):

PR #483, opened 2022-07-20
Require column names as identifiers

LucGenetier:


src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -504,7 +520,7 @@ public virtual bool IsEcsExcemptedLambda(int index)
             return false;
         }
 
-        public virtual bool AllowsRowScopedParamDelegationExempted(int index)
+        public virtual bool AllowsRowScopedParamDelegationExempted(int index, Features features)
```
MikeStall, line 523: >, Features features [](http://example.com/codeflow?start=76&length=19)

Can we avoid this? 
has higher impact, and breaking (new parameter on method).

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -484,6 +492,14 @@ public virtual bool IsLambdaParam(int index)
             return _maskLambdas.TestBit(index);
         }
 
+        // Returns true if the parameter at the specified 0-based rank is an identifier, false otherwise.
+        public virtual bool IsIdentifier(int index)
```
MikeStall, line 496: >IsIdentifier [](http://example.com/codeflow?start=28&length=12)

nit: nameing:  IsIdentifierParam(int)?   
similar to IsLambaParam and others.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -484,6 +492,14 @@ public virtual bool IsLambdaParam(int index)
...
+        {
+            Contracts.AssertIndexInclusive(index, MaxArity);
+
+            return _maskIdentifiers.TestBit(index);
```
MikeStall, line 500: >_maskIdentifiers [](http://example.com/codeflow?start=19&length=16)

If this is virtual, can we just skip _maskIdentifiers altogether and defer to the derived function?

We'll need that anyways for varargs like AddColumns.

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/DottedNameNode.cs
```diff
@@ -15,7 +13,7 @@ namespace Microsoft.PowerFx.Syntax
     /// 
     /// <code>Left.Right</code>
     /// </summary>
-    public sealed class DottedNameNode : NameNode
+    public sealed class DottedNameNode : NameNode, IIdentifierNode
```
MikeStall, line 16: >IIdentifierNode [](http://example.com/codeflow?start=51&length=15)

We should find another way to do this.

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/IIdentifierNode.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+namespace Microsoft.PowerFx.Syntax
+{
+    // Interface used on Nodes that support identifiers (FirstNameNode and DottedNameNode) 
+    public interface IIdentifierNode
```
MikeStall, line 7: >IIdentifierNode [](http://example.com/codeflow?start=21&length=15)

let's discuss more offline

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/AddColumns_SupportIdentifiers.txt
```diff
@@ -0,0 +1,31 @@
...
+>> AddColumns(LastN([1, 2, 3, 4, 5], Blank()), Double, Value * 2)
+Table()
+
+>> AddColumns(If(1<0, [1, 2, 3]), Double, Value * 2)
+Blank()
+
+>> AddColumns(Table({a: 1}, {a: 2}), b, a * a * a)
+Table({a:1,b:1},{a:2,b:8})
+
+>> AddColumns(Table({a: 1, b: {c: 3}}, {a: 2, b: {c: 4}}), b.d, a * a * a)
```
MikeStall, line 24: >b.d [](http://example.com/codeflow?start=59&length=3)

This should be an error. 

Correct would be to enclose in single quotes, 

>> AddColumns(Table({a: 1, b: {c: 3}}, {a: 2, b: {c: 4}}), 'b.d', a * a * a)

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
MikeStall: Actually, this would also be very helpful in significantly reducing the code churn.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4939,7 +4950,7 @@ public override bool PreVisit(CallNode node)
                 {
                     Contracts.Assert(_currentScope == scopeNew || _currentScope == scopeNew.Parent);
 
-                    if (maybeFunc.AllowsRowScopedParamDelegationExempted(i))
+                    if (maybeFunc.AllowsRowScopedParamDelegationExempted(i, _features))
```
MikeStall, line 4953: >_features [](http://example.com/codeflow?start=76&length=9)

we can remove this. 

Our new binding is similar to row-scope, but should be its own code path because it has enough differences.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4954,13 +4965,30 @@ public override bool PreVisit(CallNode node)
                         _txb.AddVolatileVariables(args[i], volatileVariables);
                     }
 
+                    var isIdentifier = _features.HasFlag(Features.SupportIdentifiers) && maybeFunc.IsIdentifier(i);
```
MikeStall, line 4968: >isIdentifier [](http://example.com/codeflow?start=24&length=12)

if this is true, we just need to ensure the arg is a FirstNameNode and not do anything else. 

The rest of the work is then just in CheckInvocation and IR. 

And then CheckInvocation will verify that it's a "valid" identifier. 
For example, for DropCOlumns, it will ensure the identifier is a field on arg0. 
For AddColumns, probably ensures the identifier is not  yet on Arg0 (double check)

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -254,7 +261,17 @@ public override IntermediateNode Visit(TexlCallNode node, IRTranslatorContext co
...
-                    if (func.IsLazyEvalParam(i))
+
+                    var supportIdentifiers = _features.HasFlag(Features.SupportIdentifiers);
+                    if (supportIdentifiers && func.IsIdentifier(i))
+                    {
+                        var identifierNode = arg.AsIdentifierNode();
+                        Contracts.Assert(identifierNode != null);
+
+                        // Transform the identifier node as a string literal
+                        args.Add(new TextLiteralNode(context.GetIRContext(arg, DType.String), identifierNode.GetName()));
```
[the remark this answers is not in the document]
MikeStall: Actually, we'll need "ShowColumns()" for this. 
Since that arg1 references to an existing field on arg0 (which is a common pattern). 
Whereas AddColumns is special since it's creating a new column name (so it doesn't appear on arg 0).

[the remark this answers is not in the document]
MikeStall: see [https://github.com/microsoft/Power-Fx/blob/main/src/tests/Microsoft.PowerFx.Core.Tests/DisplayNameTests.cs](https://github.com/microsoft/Power-Fx/blob/main/src/tests/Microsoft.PowerFx.Core.Tests/DisplayNameTests.cs)

MikeStall:
A followup item here will be to influence intellisense so that we suggest these identifiers. 


ShowColumns(myTable, | 
at |, intellisense should show the fields on myTable  as suggestions.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/AddColumns_SupportIdentifiers.txt
```diff
@@ -0,0 +1,31 @@
+﻿#SETUP: SupportIdentifiers
+
+>> AddColumns([1,2,3,4,5], Double, Value * 2)
```
MikeStall, line 3: tests:
-  what if we add a column name that already exists? (is that an error) 

- binding:

>> With( { b : 15 },
  AddColumns([1,2], 
     b ,   <-- this is the new column name
     Value * b) <-- this 'b' binds against With's scope. 
)

[15, 30]

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
LucGenetier: We use the flag in AddColumns.cs, in public override bool AllowsRowScopedParamDelegationExempted(int index, Features features)

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
LucGenetier: Found an alternative

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
LucGenetier: Found an alternative

LucGenetier:
Good idea

---
In reply to: [1190843921](https://github.com/microsoft/Power-Fx/pull/483#issuecomment-1190843921) [](http://example.com/codeflow?ancestors=1190843921,1190843921)

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/DottedNameNode.cs
LucGenetier: You should explain why...

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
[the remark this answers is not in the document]
LucGenetier: I'd need your help to understand how to define a display name

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AddColumns.cs
```diff
@@ -177,11 +190,24 @@ public override bool IsLambdaParam(int index)
...
+            // Left to right mask (infinite): ...010101010 
+            return index >= 1 && ((index & 1) == 1);
+        }
+
         public override bool AllowsRowScopedParamDelegationExempted(int index)
+        {
+            throw new NotImplementedException("Do not call this method, test for IUsesFeatures interface and call AllowsRowScopedParamDelegationExempted(int index, Features features) instead");
+        }
+
+        public bool AllowsRowScopedParamDelegationExempted(int index, Features features)
```
MikeStall, line 206: >AllowsRowScopedParamDelegationExempted [](http://example.com/codeflow?start=20&length=38)

We shouldn't need *any* changes to AllowsRowScopedParamDelegationExempted.

src/libraries/Microsoft.PowerFx.Core/Functions/IHasIdentifiers.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.PowerFx.Core.Functions
+{
+    internal interface IHasIdentifiers
```
MikeStall, line 7: >IHasIdentifiers [](http://example.com/codeflow?start=23&length=15)

Why a new interface, rather than just a virtual on TexlFunction?

src/libraries/Microsoft.PowerFx.Core/Types/DKind.cs
```diff
@@ -89,7 +89,9 @@ internal enum DKind
         LazyRecord = 36,
         LazyTable = 37,
 
-        _Lim = 38,
+        // Identifier type
+        Identifier = 38,
```
MikeStall, line 93: don't add a new DKind.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AddColumns.cs
```diff
@@ -86,33 +89,43 @@ public override bool CheckInvocation(TexlBinding binding, TexlNode[] args, DType
                 // Verify we have a string literal for the column name. Accd to spec, we don't support
                 // arbitrary expressions that evaluate to string values, because these values contribute to
                 // type analysis, so they need to be known upfront (before AddColumns executes).
```
MikeStall, line 91: This is a relief! Makes our life much easier....

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AddColumns.cs
```diff
@@ -86,33 +89,43 @@ public override bool CheckInvocation(TexlBinding binding, TexlNode[] args, DType
...
+
+                if ((!supportIndentifiers && (nameArgType.Kind != DKind.String || (strLitNode = nameArg.AsStrLit()) == null)) ||
+                     (supportIndentifiers && (nameArgType.Kind != DKind.Identifier || (identifierNode = nameArg.AsIdentifierNode()) == null)))
                 {
                     fArgsValid = false;
+
+                    // Argument '{0}' is invalid, expected a text literal.
                     errors.EnsureError(DocumentErrorSeverity.Severe, nameArg, TexlStrings.ErrExpectedStringLiteralArg_Name, nameArg.ToString());
                     continue;
                 }
```
MikeStall, line 103: I'd leave this intact and just say:

string value;
if (!supportIndentifiers) { 
   // existing code, unchanged if
  value = nameNode.Value;
} else {
  // new code, 
  // verify it's a FirstNameNode (See SetFunction for example)
  value = nameNode.AsFirstNameNode().Ident.Name;
}

MikeStall (changes requested):
:clock1:

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AddColumns.cs
LucGenetier: Not sure why

src/libraries/Microsoft.PowerFx.Core/Functions/IUsesFeatures.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.PowerFx.Core.Functions
+{
+    internal interface IUsesFeatures
```
MikeStall, line 7: >IUsesFeatures [](http://example.com/codeflow?start=23&length=13)

can be removed,

src/libraries/Microsoft.PowerFx.Core/Functions/IHasIdentifiers.cs
MikeStall: with default impl that always returns false.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/TestRunner.cs
```diff
@@ -345,7 +345,18 @@ internal static void TestToString(FormulaValue result, StringBuilder sb)
                 foreach (var field in fields)
                 {
                     sb.Append(dil);
-                    sb.Append(field.Name);
+
+                    if (field.Name.Contains(' '))
```
MikeStall, line 349: >field.Name.Contains(' ')) [](http://example.com/codeflow?start=24&length=25)

is there an existing method to escape?

BruceHaley:

BruceHaley:

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/TestRunner.cs
LucGenetier: Haven't found any

MikeStall:
>     public sealed class FirstNameNode : NameNode

can we undo these changes?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/FirstNameNode.cs:14 in ce0d3ce. [](commit_id = ce0d3ce37b0f15358e393c015064d854d4292155, deletion_comment = False)

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DropColumns.cs
```diff
@@ -20,11 +22,14 @@ internal sealed class DropColumnsFunction : FunctionWithTableInput
     {
         public override bool IsSelfContained => true;
 
+        public override bool HasLambdas => true;
```
MikeStall, line 25: >HasLambdas [](http://example.com/codeflow?start=29&length=10)

?
DropBolumns doesn't have lambdas.

src/tests/Microsoft.PowerFx.Interpreter.Tests/DisplayNameTests.cs
```diff
@@ -0,0 +1,49 @@
...
+using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Types;
+using Xunit;
+
+namespace Microsoft.PowerFx.Interpreter.Tests
+{
+    public class DisplayNameTests
+    {
+        [Fact]
+        public void IdentifierAndDisplayNameTest()
```
MikeStall, line 19: >IdentifierAndDisplayNameTest [](http://example.com/codeflow?start=20&length=28)

nit: filename match class  name

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -504,6 +504,13 @@ public virtual bool IsEcsExcemptedLambda(int index)
             return false;
         }
 
+        public virtual bool IsIdentifierParam(int index)
```
MikeStall, line 507: >IsIdentifierParam [](http://example.com/codeflow?start=28&length=17)

nit: can we add a comment explaining wha tthis is.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DropColumns.cs
```diff
@@ -20,11 +22,14 @@ internal sealed class DropColumnsFunction : FunctionWithTableInput
...
+
         public override bool SupportsParamCoercion => false;
 
         public DropColumnsFunction()
             : base("DropColumns", TexlStrings.AboutDropColumns, FunctionCategories.Table, DType.EmptyTable, 0, 2, int.MaxValue, DType.EmptyTable)
         {
+            ScopeInfo = new FunctionScopeInfo(this);
```
MikeStall, line 32: >ScopeInfo [](http://example.com/codeflow?start=12&length=9)

?

BruceHaley:

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DropColumns.cs
LucGenetier: Needed for functions having lambdas or identifiers

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DropColumns.cs
```diff
@@ -75,23 +78,33 @@ public override bool CheckInvocation(TexlBinding binding, TexlNode[] args, DType
                 // Verify we have a string literal for the column name. Accd to spec, we don't support
                 // arbitrary expressions that evaluate to string values, because these values contribute to
                 // type analysis, so they need to be known upfront (before DropColumns executes).
-                StrLitNode nameNode;
-                if (nameArgType.Kind != DKind.String || (nameNode = nameArg.AsStrLit()) == null)
+                StrLitNode strLitNode = null;
+                FirstNameNode identifierNode = null;
```
MikeStall, line 82: >FirstNameNode [](http://example.com/codeflow?start=16&length=13)


Consider an approach like this:

                string value;
                if (!supportIdentifiers)
                {
                    StrLitNode nameNode;
                    if (nameArgType.Kind != DKind.String || (nameNode = nameArg.AsStrLit()) == null)
                    {
                        fArgsValid = false;
                        errors.EnsureError(DocumentErrorSeverity.Severe, nameArg, TexlStrings.ErrExpectedStringLiteralArg_Name, nameArg.ToString());
                        continue;
                    }

                    value = nameNode.Value;
                } 
                else
                {
                    FirstNameNode identifierNode = nameArg.AsFirstName();
                    if (identifierNode == null)
                    {
                        fArgsValid = false;
                        errors.EnsureError(DocumentErrorSeverity.Severe, nameArg, ????, nameArg.ToString());
                        continue;
                    }

                    value = identifierNode.Ident.Name;
                }

MikeStall: - only checks supportIndentifiers once
- simplify the logic. (
- limits the scope of the locals. 
- the current error ErrExpectedStringLiteralArg_Name is probably wrong for identifiers.

src/libraries/Microsoft.PowerFx.Core/Public/DependencyFinder.cs
```diff
@@ -31,6 +31,11 @@ public override void Visit(FirstNameNode node)
         {
             var info = _binding.GetInfo(node);
 
+            if (info == null)
+            {
+                return;
+            }
```
MikeStall, line 37: what's this for?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -75,6 +75,9 @@ internal abstract class TexlFunction : IFunction
         // Return true if the function expects lambda arguments, false otherwise.
         public virtual bool HasLambdas => !_maskLambdas.IsZero;
 
+        // Returns true if the function expect identifiers, false otherwise.
+        public virtual bool HasIdentifiers => false;
```
MikeStall, line 79: >HasIdentifiers [](http://example.com/codeflow?start=28&length=14)

Why do we need this?  (ie, why isn't IsIdentifierParam sufficient)

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4743,10 +4750,10 @@ public override bool PreVisit(CallNode node)
                     return false;
                 }
 
-                // If there are no overloads with lambdas, we can continue the visitation and
+                // If there are no overloads with lambdas or identifiers, we can continue the visitation and
                 // yield to the normal overload resolution.
-                var overloadsWithLambdas = overloads.Where(func => func.HasLambdas);
-                if (!overloadsWithLambdas.Any())
+                var overloadsWithLambdasOrIdentifiers = overloads.Where(func => func.HasLambdas || func.HasIdentifiers);
```
MikeStall, line 4755: > || func.HasIdentifiers [](http://example.com/codeflow?start=95&length=23)

Need to triple check the impact of this. 

func.HasIdentifiers will be true even if the feature flag is not enabled - so this will impact all code paths (bindings in Canvas apps)

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4954,13 +4961,25 @@ public override bool PreVisit(CallNode node)
...
-                    _txb.AddVolatileVariables(node, _txb.GetVolatileVariables(args[i]));
+                    if (!isIdentifier)
+                    {
+                        args[i].Accept(this);
+                        _txb.AddVolatileVariables(node, _txb.GetVolatileVariables(args[i]));
+                        argTypes[i] = _txb.GetType(args[i]);
+                    }
+                    else
+                    {
+                        argTypes[i] = DType.String;
```
MikeStall, line 4980: > DType.String; [](http://example.com/codeflow?start=37&length=14)

why string? 
Does anybody use this value?

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4983,7 +5002,19 @@ public override bool PreVisit(CallNode node)
                 _currentScope = scopeNew.Parent;
                 PostVisit(node.Args);
 
-                // Typecheck the invocation.
+                if (maybeFunc.HasIdentifiers && _features.HasFlag(Features.SupportIdentifiers))
+                {
+                    foreach (var arg in args.Select((a, i) => a is FirstNameNode firstNameNode && maybeFunc.IsIdentifierParam(i) ? (i, firstNameNode) : (-1, null)).Where(a => a.i != -1))
```
MikeStall, line 5007: >  foreach (var arg in args.Select((a, i) => a is FirstNameNode firstNameNode && maybeFunc.IsIdentifierParam(i) ? (i, firstNameNode) : (-1, null)).Where(a => a.i != -1)) [](http://example.com/codeflow?start=18&length=168)

This expression is too complicated.
can you unroll the for-loop, and just avoid the linq expression and ?: operator.

src/libraries/Microsoft.PowerFx.Core/Public/DependencyFinder.cs
LucGenetier: When we have a FirstNameNode which is an identifier, its ocrresponding FirstNameInfo is null, so this is to avoid a crash.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
LucGenetier: I don't think so. We need functions that have identifiers to be treated like those having lambdas, with specific scopes for these parameters

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
LucGenetier: You're right, this isn't correct.
I'll add feature flag consideration here.

LucGenetier: I'll test with Power Apps later

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
LucGenetier: Yes!
We need a valid type and you didn't want to add DType.Identifier

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
lesaltzm, line 79: Add doc comment that says IsIdentifierParam must also be overloaded if this is?

lesaltzm:
> <?xml version="1.0" encoding="UTF-8" standalone="yes"?>

Why all these changes to strings?

---
Refers to: src/strings/PowerFxResources.bg-BG.resx:1 in 51092a0. [](commit_id = 51092a0913204785e0bb2f3fae551ede55e538b0, deletion_comment = True)

lesaltzm:
> <?xml version="1.0" encoding="UTF-8" standalone="yes"?>

Note that only the en-us file is owned by us, the others are regenerated by the Localization team. 

---
In reply to: [1197082903](https://github.com/microsoft/Power-Fx/pull/483#issuecomment-1197082903) [](http://example.com/codeflow?ancestors=1197082903)

---
Refers to: src/strings/PowerFxResources.bg-BG.resx:1 in 51092a0. [](commit_id = 51092a0913204785e0bb2f3fae551ede55e538b0, deletion_comment = True)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/TestRunner.cs
lesaltzm: See TexlLexer.EscapeName

LucGenetier:
> <?xml version="1.0" encoding="UTF-8" standalone="yes"?>

I have done the translation for all languages (via ACS), then can fix that if they want

---
In reply to: [1197083318](https://github.com/microsoft/Power-Fx/pull/483#issuecomment-1197083318) [](http://example.com/codeflow?ancestors=1197083318,1197082903)

---
Refers to: src/strings/PowerFxResources.bg-BG.resx:1 in 51092a0. [](commit_id = 51092a0913204785e0bb2f3fae551ede55e538b0, deletion_comment = True)

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
lesaltzm: Consider: DType.Unknown, since we don't actually care what it is. DType.String may lead to someone depending on it.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4983,7 +5002,26 @@ public override bool PreVisit(CallNode node)
...
                 PostVisit(node.Args);
 
-                // Typecheck the invocation.
+                if (maybeFunc.HasIdentifiers && _features.HasFlag(Features.SupportIdentifiers))
+                {
+                    var i = 0;
+
+                    foreach (var arg in args)
+                    {
+                        i++;
```
lesaltzm, line 5011: if we're using a counter anyway, should we just do 
`for (i = 0; i < args.Length; i++)` instead of allocating an iterator?

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4983,7 +5002,26 @@ public override bool PreVisit(CallNode node)
...
+
+                        if (arg is FirstNameNode firstNameNode && maybeFunc.IsIdentifierParam(i))
+                        {
+                            if (DType.TryGetLogicalNameForColumn(argTypes[0], firstNameNode.Ident.Name, out var logicalName))
+                            {
```
lesaltzm, line 5016: if you use:
GetLogicalNodeNameAndUpdateDisplayNames(argTypes[0], firstNameNode.Iden), this will also handle the display name rewrite cases.

src/tests/Microsoft.PowerFx.Interpreter.Tests/DisplayNameTests.cs
```diff
@@ -0,0 +1,49 @@
...
+using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Types;
+using Xunit;
+
+namespace Microsoft.PowerFx.Interpreter.Tests
+{
+    public class DisplayNameTests
+    {
+        [Fact]
+        public void DisplayNameTest1()
```
lesaltzm, line 19: please add tests covering Engine.GetInvariantExpression/GetDisplayExpression

See the DisplayNameTests in Core for examples

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4983,7 +5002,26 @@ public override bool PreVisit(CallNode node)
...
+                                var typedName = argTypes[0].GetAllNames(DPath.Root).First(tn => tn.Name == logicalName);
+                                var fileNameInfo = FirstNameInfo.Create(firstNameNode, new NameLookupInfo(BindKind.NamedValue, typedName.Type, DPath.Root, 0, displayName: new DName(logicalName)));
```
lesaltzm, line 5018: >logicalName [](http://example.com/codeflow?start=181&length=11)

Why are we passing something called "logicalName" to a field for DisplayName?

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4954,13 +4961,25 @@ public override bool PreVisit(CallNode node)
...
 
+                    var isIdentifier =
+                        args[i] is FirstNameNode &&
+                        _features.HasFlag(Features.SupportIdentifiers) &&
+                        maybeFunc.IsIdentifierParam(i);
+
                     // Use the new scope only for lambda args.
-                    _currentScope = (maybeFunc.IsLambdaParam(i) && scopeInfo.AppliesToArgument(i)) ? scopeNew : scopeNew.Parent;
-                    args[i].Accept(this);
+                    _currentScope = ((maybeFunc.IsLambdaParam(i) || isIdentifier) && scopeInfo.AppliesToArgument(i)) ? scopeNew : scopeNew.Parent;
```
lesaltzm, line 4970: >isIdentifier [](http://example.com/codeflow?start=68&length=12)

does isIdentifier need to affect the scope here? Seems mostly irrelevant.

src/libraries/Microsoft.PowerFx.Core/Binding/BindInfo/FirstNameInfo.cs
```diff
@@ -18,6 +18,8 @@ internal sealed class FirstNameInfo : NameInfo
     {
         public override DName Name => Node.AsFirstName().Ident.Name;
 
+        public DName DisplayName { get; private set; }
```
lesaltzm, line 21: >private set [](http://example.com/codeflow?start=40&length=11)

`init`?

lesaltzm: Also, is this always the display name? under what circumstances is it set? Add a comment clarifying that.

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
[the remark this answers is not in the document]
lesaltzm: This also applies to DropColumns, those are referencing fields in the type and may have display names.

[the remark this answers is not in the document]
lesaltzm: see what we do for first names to ensure it's resolved correctly:

``` 
                var nodeName = context.Binding.TryGetReplacedIdentName(node.Ident, out var newIdent) ? new DName(newIdent) : node.Ident.Name;
```

src/libraries/Microsoft.PowerFx.Core/Public/DependencyFinder.cs
lesaltzm: Is this still true? it looks like you're setting a first name info on binder.cs:5020

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AddColumns.cs
```diff
@@ -78,6 +77,9 @@ public override bool CheckInvocation(TexlBinding binding, TexlNode[] args, DType
                 errors.EnsureError(DocumentErrorSeverity.Severe, args[0].Parent.CastList().Parent.CastCall(), TexlStrings.ErrBadArityOdd, count);
             }
 
+            var supportIndentifiers = binding.Features.HasFlag(Features.SupportIdentifiers);
+            string value = null;
```
lesaltzm, line 81: >value [](http://example.com/codeflow?start=19&length=5)

naming nit: `columnName`?

lesaltzm: Consider moving this declaration inside the `for` since we'd want to reset it each iteration.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AddColumns.cs
```diff
@@ -86,33 +88,59 @@ public override bool CheckInvocation(TexlBinding binding, TexlNode[] args, DType
...
-                if (nameArgType.Kind != DKind.String ||
-                    (nameNode = nameArg.AsStrLit()) == null)
+                if (supportIndentifiers)
                 {
-                    fArgsValid = false;
-                    errors.EnsureError(DocumentErrorSeverity.Severe, nameArg, TexlStrings.ErrExpectedStringLiteralArg_Name, nameArg.ToString());
-                    continue;
+                    FirstNameNode identifierNode = nameArg.AsFirstName();
+
+                    if (identifierNode == null)
```
lesaltzm, line 95: the `AsxName` functions are left over from older implementations. 

this can be more consise as
`if (identifierNode is FirstNameNode identifierNode)` ...

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AddColumns.cs
```diff
@@ -86,33 +88,59 @@ public override bool CheckInvocation(TexlBinding binding, TexlNode[] args, DType
...
+                    {
+                        fArgsValid = false;
+
+                        // Argument '{0}' is invalid, expected a text literal.
+                        errors.EnsureError(DocumentErrorSeverity.Severe, nameArg, TexlStrings.ErrExpectedStringLiteralArg_Name, nameArg.ToString());
```
lesaltzm, line 100: >ErrExpectedStringLiteralArg_Name [](http://example.com/codeflow?start=94&length=32)

this error message is not accurate after this change, should probably be a new message capturing that it's `expecting identifier` instead.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AddColumns.cs
```diff
@@ -86,33 +88,59 @@ public override bool CheckInvocation(TexlBinding binding, TexlNode[] args, DType
...
 
-                var columnName = new DName(nameNode.Value);
+                var columnName = new DName(value);
                 if (DType.TryGetDisplayNameForColumn(typeScope, columnName, out var colName))
                 {
                     columnName = new DName(colName);
                 }
 
                 // Verify that the name doesn't already exist as either a logical or display name
-                if (typeScope.TryGetType(columnName, out var columnType) || DType.TryGetLogicalNameForColumn(typeScope, columnName, out var unused))
```
lesaltzm, line 113: >|| DType.TryGetLogicalNameForColumn(typeScope, columnName, out var unused)) [](http://example.com/codeflow?start=73&length=75)

Why are we removing this collision check?

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
```diff
@@ -69,6 +69,7 @@ internal class BuiltinFunctionsCore
         public static readonly TexlFunction Day = _library.Append(new DayFunction());
         public static readonly TexlFunction Degrees = _library.Append(new DegreesFunction());
         public static readonly TexlFunction DegreesT = _library.Append(new DegreesTableFunction());
+        public static readonly TexlFunction DropColumns = _library.Append(new DropColumnsFunction());
         public static readonly TexlFunction EndsWith = _library.Append(new EndsWithFunction());
```
lesaltzm, line 73: note to whoever bumps PA client versions: this is a tiny breaking change, will require removing this from the paclient function list.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DropColumns.cs
```diff
@@ -74,24 +78,53 @@ public override bool CheckInvocation(TexlBinding binding, TexlNode[] args, DType
 
                 // Verify we have a string literal for the column name. Accd to spec, we don't support
                 // arbitrary expressions that evaluate to string values, because these values contribute to
-                // type analysis, so they need to be known upfront (before DropColumns executes).
-                StrLitNode nameNode;
-                if (nameArgType.Kind != DKind.String || (nameNode = nameArg.AsStrLit()) == null)
+                // type analysis, so they need to be known upfront (before DropColumns executes).                            
+                if (supportIndentifiers)
```
lesaltzm, line 82: generally the same set of comments to AddColumns

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DropColumns.cs
```diff
@@ -74,24 +78,53 @@ public override bool CheckInvocation(TexlBinding binding, TexlNode[] args, DType
...
+                    errors.EnsureError(DocumentErrorSeverity.Severe, nameArg, TexlStrings.ErrArgNotAValidIdentifier_Name, value);
                     continue;
                 }
 
-                var columnName = new DName(nameNode.Value);
+                DName? displayName;
+
+                var columnName = nameArg is FirstNameNode firstNameNode && ((displayName = binding.GetDisplayName(firstNameNode)) != null)
+                    ? displayName.Value
+                    : new DName(value);
```
lesaltzm, line 127: this seems incorrect: display names aren't in the type, by definition. Those are only logical names.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -1584,6 +1584,13 @@ public FirstNameInfo GetInfo(FirstNameNode node)
             return _infoMap[node.Id] as FirstNameInfo;
         }
 
+        public DName? GetDisplayName(FirstNameNode node)
```
lesaltzm, line 1587: What is this used for?

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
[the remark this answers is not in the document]
LucGenetier: I've done what you requested but can't say I've understood anything.
I've too many unanswered questions.
"see what we do for first names" -> if that's in Visit(FirstNameNode node, IRTranslatorContext context), I don't understand that code, don't know what LambdaFullRecord, LambdaField, OptionSet, PowerFxResolvedObject "kinds" are + What output is it generating?
"ensure it's resolved correctly" -> Am with you but I wish I'd understand what we're doing here in this method Visit(TexlCallNode node...) -  Why is the method called "Visit"? What is an "IntermediateNode" (apart from an IRContext?)? What is IRTranslatorContext? Why do we call MaybeInjectCoercion at the end?

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
LucGenetier: Not convinced that calling args[i] is more efficient.
Anyway for a loop with 3 or 4 args at max, will it make any difference?

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
LucGenetier: Am now using it but don't understand what it's really doing and how it's different from the one I used.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
LucGenetier: Please allow me this incongruity...
The code is full of that already, one more, one less, will it make any difference?
I've fixed it though, it's better to use a valid name (and all the code should be updated too)

Taking this one as it's in front of me, why is something that "visits" called "accept" ?!

        /// <summary>
        /// This method visits the node using the visitor and context provided.
        /// </summary>
        /// <typeparam name="TResult"></typeparam>
        /// <typeparam name="TContext"></typeparam>
        /// <param name="visitor">Visitor to use.</param>
        /// <param name="context">Context.</param>
        /// <returns></returns>
        public abstract TResult Accept<TResult, TContext>(IRNodeVisitor<TResult, TContext> visitor, TContext context);

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
LucGenetier: I don't know if we should or should not use isIdentifier here but tests show it's not necessary and I've removed it.

src/libraries/Microsoft.PowerFx.Core/Binding/BindInfo/FirstNameInfo.cs
LucGenetier: We mainly set it here 

                if (maybeFunc.HasIdentifiers && _features.HasFlag(Features.SupportIdentifiers))
                {
                    var i = 0;

                    foreach (var arg in args)
                    {
                        i++;

                        if (arg is FirstNameNode firstNameNode && maybeFunc.IsIdentifierParam(i))
                        {
                            var displayName = GetLogicalNodeNameAndUpdateDisplayNames(argTypes[0], firstNameNode.Ident, out _);

                            if (argTypes[0].TryGetType(displayName, out var fieldType))
                            {
                                var fileNameInfo = FirstNameInfo.Create(firstNameNode, new NameLookupInfo(BindKind.NamedValue, fieldType, DPath.Root, 0, displayName: new DName(displayName)));
                                _txb.SetInfo(firstNameNode, fileNameInfo);
                            }
                        }
                    }
                }

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AddColumns.cs
LucGenetier: if (nameArg is not FirstNameNode identifierNode)

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AddColumns.cs
LucGenetier: Good catch.
I can't explain...

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
LucGenetier: OK.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DropColumns.cs
LucGenetier: This is how I set the displayname in the type.

            var rt = RecordType.Empty()
                .Add(new NamedFormulaType("logicalA", FormulaType.Number, displayName: "displayName"))
                .Add(new NamedFormulaType("logicalB", FormulaType.Number));

Help me fixing the code if it's incorrect, I don't know what to do.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
LucGenetier: Get the DisplayName of that node
If  *display* name is not the right term, let me know what to use. Logical? Virtual? Illlogical? ...

BruceHaley:

src/strings/PowerFxResources.en-US.resx
```diff
@@ -6549,4 +6549,8 @@
     <value>value</value>
     <comment>function_parameter - Second argument to the Set function - any Power Apps value.</comment>
   </data>
+  <data name="ErrExpectedIdentifierArg_Name" xml:space="preserve">
```
lesaltzm, line 6552: + @CarlosFigueiraMSFT 

Files other than en-US should not be updated in feature PRs. The translations are done by the localization team on a regular basis.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
lesaltzm: Fair enough, but if you're using a counter anyway, it also looks cleaner, imo.

BruceHaley:

BruceHaley:

LucGenetier:
Closing this PR because it's superseded by PR # 863

PR #501, opened 2022-07-28
adding dropped curly braces back in interpolated string when formatting

akshar2401:
When clicking 'format text" on a formula that contains a string interpolated expression, any double curly brackets will be stripped down to just 1, resulting in an invalid expression.

Example: 
Set(str, $" {{ }} ");

Clicking "Format text' in the formula bar will result in:

Set(
    str,
    $" { } "
);

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Utils/CharacterUtils.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Utils/CharacterUtils.cs
@@ -158,7 +158,7 @@ namespace Microsoft.PowerFx.Core.Utils
             return InternalEscapeString(value, length, lengthForBuilder, ref sb, finalizeBuilder: true);
         }
 
-        public static string ExcelEscapeString(string value, bool isInterpolatedString = false)
+        public static string ExcelEscapeString(string value, bool isValueAnInterpolatedString = false)
         {
             Contracts.AssertValue(value);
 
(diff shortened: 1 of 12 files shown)


src/libraries/Microsoft.PowerFx.Core/Utils/CharacterUtils.cs
```diff
@@ -380,6 +384,18 @@ private static void UpdateEscapeInternals(string escapedValue, string input, int
             sb.Append(escapedValue);
         }
 
+        private static void HandleEscapedCurlyBraces(char curlyBrace, string input, bool isInterpolatedInput, int estimatedLength, int currentPosition, ref int charsToAdd, ref StringBuilder sb)
+        {
+            Contracts.AssertOneOfValueType(curlyBrace, '{', '}');
+
+            if (!isInterpolatedInput)
+            {
+                return;
```
bryancroteau-MSFT, line 393: This function does not need to accept this boolean parameter just to return if it's false. It would be better to have this if statement in the switch/case body.

bryancroteau-MSFT (changes requested):

src/libraries/Microsoft.PowerFx.Core/Utils/CharacterUtils.cs
akshar2401: Done. Removed this function and moved the code directly in the case block.

bryancroteau-MSFT (approved):

Issue #276, opened 2022-04-05
Use typesafe builder pattern for PowerFx config instead of lock


lesaltzm:

MikeStall:
Fixed with 503.

PR #503, opened 2022-07-28
Fixing symbol table

MikeStall:
Symbol Table improvements: 
1. Allow sharing config and chaining. A common config may hold all the connectors, and then a smaller config may chain to it and add a function. 
2. Allow mutating config to enable "sessions". Mutations during bind will be caught and the binding will fail - similar to how mutating a collection during enumeration will be caught and fail the enumeration. 
3. Allow passing in config per-expression (Rather than just engine wide) 
4. Allow passing in runtime state per expression to functions (useful for User(), Rand(), Now()/TimeZoneInfo(), etc). 
5. Unify globals and locals 

**>> Introducing SymbolTable and SymbolValues**
Introduce **SymbolTable** as a public facing implementation of INameResolver. (This follows the pattern of FormulaType/Dtype).
SymbolTable also has a ReadOnlySymbolTable that can be handed to binding. 
SymbolTable can also have a parent pointer to another SymbolTable for chaining/reuse. 

Introduce a **SymbolValues**, which is the runtime (interpreter-only) counterpart to SymbolValues. 
The SymbolValues can also provide a **SymbolTable** to help keep the runtime and design time in sync. 

SymbolValues/SymbolTable are a generalized pattern of the previous RecordValue/RecordType used by Check/Eval. 

Add Check/Eval overloads that accept these new Symbol* objects. 

Much of the state on PowerFxConfig now moves to SymbolTable. To keep compact, the PowerFxConfig holds a default symbol table and methods forward to it. 

**>> Simplifying CoreFunctions **
Before, we had an awkward dance where Engine would take in a config and SetCoreFunctions() to mutate the config.  This was how the engine described what functions it could support. 
Now, since symbol tables can compose, the engine has a SymbolTable for whatever core functions it supports. 
And that table is composed with other tables containing global state, etc. 

**>>Unifying Globals and Locals **
Before, locals were passed as RowScope. Globals were handled via the INameResolver.  And there was nothing in between. 
Now, everything is just "symbols", provided by a symbol table.  This did flush out some differences between rowscope and global handling. 

**>> This avoids breaking changes**!  But does add [Obsolete] to some existing interfaces. 
Notably, CheckResult.IExpression is now Obsolete - this was problematic since CheckResult is in Core, but IExpression is for evaluation (in the interpreter layer). So it was a layering violation. 
Now, interpreter has an extension method on CheckResult to create a new IExpressionEvaluator, which can be used for evaluation - but both the extension method and new interface live in Interpreter where they belong.


src/tests/Microsoft.PowerFx.Interpreter.Tests/ConfigTests.cs
```diff
@@ -0,0 +1,409 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
```
MikeStall, line 4: This is DRAFT. 
Just check out these test cases and ignore anything else.

MikeStall:
This also will fix #276

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -588,29 +588,34 @@ private void CheckThroughInterface(IPowerFxEngine engine)
         [Fact]
         public void RecalcEngineLocksConfig()
```
CarlosFigueiraMSFT, line 589: >RecalcEngineLocksConfig [](http://example.com/codeflow?start=20&length=23)

If we can mutate the config (by adding new functions / symbols), then we should rename this :-)

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -684,9 +689,9 @@ public void EmptyEnumStoreTest()
             Assert.True(checkResult.IsSuccess);
             Assert.IsType<StringType>(checkResult.ReturnType);
 
-            var enums = config.EnumStoreBuilder.Build().EnumSymbols;
-
-            Assert.True(enums.Count() > 0);
+            // $$$ Which symbol table?
```
CarlosFigueiraMSFT, line 692: >// $$$ Which symbol table? [](http://example.com/codeflow?start=12&length=26)

A union of all symbols?

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
MikeStall: +1. Will rename. We no longer lock config.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -588,29 +588,34 @@ private void CheckThroughInterface(IPowerFxEngine engine)
...
 
             var func = BuiltinFunctionsCore.AsType; // Function not already in engine
             Assert.DoesNotContain(func, config.Functions); // didn't get auto-added by engine.
 
+            // We can mutate config after engine is created.
             var optionSet = new OptionSet("foo", DisplayNameUtility.MakeUnique(new Dictionary<string, string>() { { "one key", "one value" } }));
-            Assert.Throws<InvalidOperationException>(() => config.AddFunction(func));
-            Assert.Throws<InvalidOperationException>(() => config.AddOptionSet(optionSet));
+            config.SymbolTable.AddFunction(func);
+            config.SymbolTable.AddEntity(optionSet);
```
CarlosFigueiraMSFT, line 605: After this line:

    Assert.Contains(func, config.Functions); // function was added to the config.

src/tests/Microsoft.PowerFx.Interpreter.Tests/ConfigTests.cs
```diff
@@ -0,0 +1,476 @@
...
+            var s1 = new SymbolTable();
+            s1.AddVariable("local", FormulaType.Number);
+
+            // Per expression.
+            var engine = new RecalcEngine();
+            engine.UpdateVariable("global", 2.0);
+
+            var runtimeConfig = new RuntimeConfig();
+            runtimeConfig.SymbolValues["local"] = FormulaValue.New(3);
+
```
CarlosFigueiraMSFT, line 38: It feels a little disconnected to have the symbol table nd the runtime config's symbol values separately. I hope we can find a better way to do it than using variable names as the "join key" between the type (in the symbol table) and the runtime value (in the runtime config).

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -1,228 +1,183 @@
...
-        private void CheckUnlocked()
-        {
-            if (_isLocked)
-            {
-                throw new InvalidOperationException("This PowerFxConfig instance is locked");
-            }
-        }
-    }
-}
+            throw new NotImplementedException(); // $$$
```
MikeStall, line 177: > throw new NotImplementedException(); // $$$ [](http://example.com/codeflow?start=11&length=44)

Need to resolve this. Used for RenameDriver tests.

src/libraries/Microsoft.PowerFx.Core/Public/Config/RuntimeConfig.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Globalization;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx
+{
+    /// <summary>
+    /// Runtime Values that correspond to bindings described in a <see cref="SymbolTable"/>.
+    /// These can also be reused across runs.
+    /// </summary>
+    public class RuntimeConfig
```
MikeStall, line 15: >RuntimeConfig [](http://example.com/codeflow?start=17&length=13)

This gets passed per-eval. 
supersedes passing a RecordValue as parameters.

src/tests/Microsoft.PowerFx.Interpreter.Tests/ConfigTests.cs
MikeStall: Agreed, but the problem is how do we
- declare something at bind time 
- and then eval it many times, passing in different parameters for each eval. 

Wondering if there's something elegant we can do for interpreter cases (where the binder has the runtime values), like have a runtime config double as a symbol table.

src/tests/Microsoft.PowerFx.Interpreter.Tests/ConfigTests.cs
```diff
@@ -0,0 +1,476 @@
...
+            var engine = new Engine(new PowerFxConfig());
+
+            Assert.Throws<InvalidOperationException>(
+                () => engine.Check2("x", symbolTable: s1));
+        }
+
+        // Add a function that gets different runtime state per expression invoke
+        [Fact]
+        public async Task LocalFunction()
```
MikeStall, line 235: This test demonstrates solving #536

MikeStall:
Feedback From Bryan:
1. ImmutableSymbolTable 
1. Find a way to gaurantee that SymbolTable nad RuntimeConfig have the same structure....

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -119,17 +126,41 @@ public CheckResult Check(string expressionText, RecordType parameterType = null,
...
-            return CheckInternal(parse, parameterType, bindingConfig);
+            var symbols = SymbolTable.NewFromRecord(parameterType);
+
+            return Check2(parse, bindingConfig, symbols);
+        }
+
+        public CheckResult Check2(
+            string expressionText,
+            ParserOptions parserOptions = null,
+            SymbolTable symbolTable = null)
```
LucGenetier, line 137: Should we have a class instead of 2 params
This could easily be extended w/o having a breaking change

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -155,6 +186,12 @@ private CheckResult CheckInternal(ParseResult parse, RecordType parameterType, B
                 result.Expression = CreateEvaluator(result);
             }
 
+            var endHash = symbolTable?.VersionHash;
+            if (startHash != endHash)
```
LucGenetier, line 190: Shouldn't this be in a Contracts.Assert(...)

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -1,228 +1,183 @@
...
+        // $$$ Make this "locked" so we can safely share. 
+        private static readonly SymbolTable _allBuiltinCoreFunctions = PowerFxConfig.NewDefault(Core.Texl.BuiltinFunctionsCore.BuiltinFunctionsLibrary);
+
+        /// <summary>
+        /// Default Functions. $$$ Move to engine! That's what determines this....
+        /// </summary>
+        public SymbolTable DefaultFunctions { get; set; } = _allBuiltinCoreFunctions;
+
+        // $$$
+        internal static SymbolTable NewDefault(IEnumerable<TexlFunction> coreFunctions)
```
LucGenetier, line 41: >NewDefault [](http://example.com/codeflow?start=36&length=10)

AddFunctions, WithFunctions ?

src/libraries/Microsoft.PowerFx.Core/Public/Config/RuntimeConfig.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        // Used for parsing functions like Text()/Value() 
+        public CultureInfo CultureInfo { get; }
+
+        // $$$ Days/Hours? Impacts TimeZoneOffset(), Today(), Now() 
+        public double? TimeZoneOffsetHours { get; set; }
```
LucGenetier, line 21: Needs to be TimeZoneInfo otherwise we won't be able to manage DST properly

src/libraries/Microsoft.PowerFx.Core/Public/Config/RuntimeConfig.cs
```diff
@@ -0,0 +1,59 @@
...
+        // $$$ Should this be IReadOnlyDict
+        public Dictionary<string, FormulaValue> SymbolValues { get; set; } = new Dictionary<string, FormulaValue>();
+        
+        // $$$ Should everything (culture, timezoneoffset,etc) just use this?
+        // Functions can grab runtime state via the Type handle. 
+        public Dictionary<Type, object> RuntimeState { get; set; }
+
+        // $$$ Null if missing? Or throw?
+        // Chain to parents?
+        public T GetState<T>() => (T)RuntimeState[typeof(T)];
```
LucGenetier, line 46: >(T)RuntimeState[typeof(T)]; [](http://example.com/codeflow?start=34&length=27)

No testing if typeof(T) is present?

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -0,0 +1,388 @@
...
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.Types.Enums;
+using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx
+{
+    // Can pass to symbol table at compile time, and then it has a handle
+    // if we use it for execution at runtime. 
+    public interface ICanGetValue
```
LucGenetier, line 25: IHasValue

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -0,0 +1,388 @@
...
+    {
+        FormulaValue Value { get; }
+    }
+
+    // NameInfo data to associate a symbol at bind time with a runtime config at runtime.
```
LucGenetier, line 30: NameSymbol?

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -0,0 +1,388 @@
...
+{
+    // Can pass to symbol table at compile time, and then it has a handle
+    // if we use it for execution at runtime. 
+    public interface ICanGetValue
+    {
+        FormulaValue Value { get; }
+    }
+
+    // NameInfo data to associate a symbol at bind time with a runtime config at runtime. 
+    internal class NameSymbol
```
LucGenetier, line 31: SymbolName?

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -0,0 +1,388 @@
...
+    [DebuggerDisplay("{DebugName}")]
+    public class SymbolTable : INameResolver, IGlobalSymbolNameResolver
+    {
+        // Changed on each update. 
+        // Host can use to ensure that a symbol table wasn't mutated on us.                 
+        private int _version = (int)DateTime.UtcNow.Ticks; // seed psuedo-random
+
+        public int VersionHash => _parent == null ? 
+            _version : 
+            (_version * -1521134295) + _parent.VersionHash;
```
LucGenetier, line 55: Unsecure and probably unsafe as two threads calling this code at the same time will produce the same result.

We could consider a fully secure option (but slow)

    using (var provider = new RNGCryptoServiceProvider()) 
    {
        var bytes = new byte[16];
        provider.GetBytes(bytes);

        return new Guid(bytes);
    }

or use a simple counter with Interlocked.Add with a static (very fast)

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -0,0 +1,388 @@
...
+        
+        public SymbolTable _parent;
+
+        // Helper in debugging. Useful when we have multiple symbol tables chained. 
+        public string DebugName { get; set; } = "SymbolTable";
```
LucGenetier, line 60: DebugText

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -0,0 +1,388 @@
...
+        /// </summary>
+        /// <param name="name"></param>
+        public void RemoveFunction(string name)
+        {
+            _version++;
+
+            // $$$ namespace?
+            _functions.RemoveAll(func => func.Name == name);
+        }
+
```
LucGenetier, line 209: We probably need RemoveFunction(func) as well to remove an overload and not all funcitons with the same name

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -0,0 +1,388 @@
...
+        internal void AddFunction(TexlFunction function)
+        {
+            _version++;
+            _functions.Add(function);
+            
+            // Add any associated enums 
+            EnumStoreBuilder?.WithRequiredEnums(new List<TexlFunction>() { function });
```
LucGenetier, line 216: >? [](http://example.com/codeflow?start=28&length=1)

Shouldn't we throw is EnumStoreBuilder is null?

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -0,0 +1,388 @@
...
+        
+        /// <summary>
+        /// Provide variable for binding only.
+        /// Value must be provided at runtime via <see cref="RuntimeConfig.SymbolValues"/>.
+        /// </summary>
+        /// <param name="name"></param>
+        /// <param name="type"></param>
+        public void AddVariable(string name, FormulaType type)
+        {
+            _version++;
```
LucGenetier, line 156: > _version++; [](http://example.com/codeflow?start=11&length=12)

unsafe if two thread are calling this in parallel
Should be in GetNextVersion() where we do the needful

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -0,0 +1,388 @@
...
+        }
+
+        IEnumerable<TexlFunction> INameResolver.LookupFunctions(DPath theNamespace, string name, bool localeInvariant)
+        {
+            Contracts.Check(theNamespace.IsValid, "The namespace is invalid.");
+            Contracts.CheckNonEmpty(name, "name");
+
+            // See TexlFunctionsLibrary.Lookup
+            // return _functionLibrary.Lookup(theNamespace, name, localeInvariant, null);            
+            var functionLibrary = _functions.Where(func => func.Namespace == theNamespace && name == (localeInvariant ? func.LocaleInvariantName : func.Name)); // Base filter
```
LucGenetier, line 332: >func.Namespace == theNamespace && name == (localeInvariant ? func.LocaleInvariantName : func.Name) [](http://example.com/codeflow?start=59&length=98)

equality should be done in functions that know if we are case sensitive or not

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -0,0 +1,388 @@
...
+            return functionLibrary;
+        }
+
+        IEnumerable<TexlFunction> INameResolver.LookupFunctionsInNamespace(DPath nameSpace)
+        {
+            Contracts.Check(nameSpace.IsValid, "The namespace is invalid.");
+
+            return _functions.Where(function => function.Namespace.Equals(nameSpace));
```
LucGenetier, line 340: >function.Namespace.Equals(nameSpace) [](http://example.com/codeflow?start=48&length=36)

Same

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTableEnumerator.cs
```diff
@@ -0,0 +1,47 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.PowerFx
+{
+    /// <summary>
+    /// Enumerate though a chain of symbol tables.
```
LucGenetier, line 10: typo?

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTableResolver.cs
```diff
@@ -0,0 +1,190 @@
...
+
+        public IReadOnlyDictionary<string, NameLookupInfo> GlobalSymbols
+        {
+            get
+            {
+                var map = new Dictionary<string, NameLookupInfo>();
+
+                foreach (var table in _symbolTables)
+                {
+                    if (table is IGlobalSymbolNameResolver x)
```
LucGenetier, line 71: >x [](http://example.com/codeflow?start=59&length=1)

nameResolver

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTableResolver.cs
```diff
@@ -0,0 +1,190 @@
...
+        // $$$ MAke this non-virtual 
+        public virtual bool Lookup(DName name, out NameLookupInfo nameInfo, NameLookupPreferences preferences = NameLookupPreferences.None)
+        {
+            foreach (INameResolver table in _symbolTables)
+            {
+                if (table.Lookup(name, out nameInfo, preferences))
+                {
+                    return true;
+                }
+            }
```
LucGenetier, line 100: Is Linq that slow to rewrite it here?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
MikeStall: no, it's a runtime error, and I also include a test to verify. 
(Contract.Assert can get optimized out in retail builds)

This is similar to "can't mutate a collection while enumerating it" exceptions.

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
MikeStall: These classes are not thread-safe - this is like the default BCL collections.  Host is responsible for locking if they're going to try and mutate from multiple threads.

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
MikeStall: not a thread-safe collection, by-design.

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
MikeStall: this is just a copy of existing code.

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTableResolver.cs
MikeStall: ?
The use of out-parameters discourages me from trying to do this in a linq query.

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTableResolver.cs
```diff
@@ -0,0 +1,200 @@
...
+using Microsoft.PowerFx.Core.Utils;
+
+namespace Microsoft.PowerFx
+{
+    /// <summary>
+    /// Basic implementation of INameResolver around a <see cref="PowerFxConfig"/> object. 
+    /// This aides in binding and intellisense. 
+    /// Host can override Lookup to provide additional symbols to the expression. 
+    /// </summary>
+    internal class SymbolTableResolver : ReadOnlySymbolTable, INameResolver, IGlobalSymbolNameResolver
```
MikeStall, line 23: >SymbolTableResolver [](http://example.com/codeflow?start=19&length=19)

nit: better name:

ComposedReadOnlySymbolTable

src/tests/Microsoft.PowerFx.Interpreter.Tests/ConfigTests.cs
MikeStall: Latest iterations have some improvements here.

BruceHaley:
:x: 2 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.2-preview.20220816-317288](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.2-preview.20220816-317288).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfig.FunctionInfos.get()' does not exist in the implementation but it does exist in the contract.
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfig.GetAllFunctionNames()' does not exist in the implementation but it does exist in the contract.
```

</details>
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.2-preview.20220816-317288](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.2-preview.20220816-317288).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.ReflectionFunction.Invoke(Microsoft.PowerFx.Types.FormulaValue[])' does not exist in the implementation but it does exist in the contract.
```

</details>

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseData/IntellisenseData.cs
```diff
@@ -16,7 +16,9 @@ namespace Microsoft.PowerFx.Intellisense.IntellisenseData
     // The IntellisenseData class contains the pre-parsed data for Intellisense to provide suggestions
     internal class IntellisenseData : IIntellisenseData
     {
+        // $$$ This should probably be symbol and not just config. 
         protected readonly PowerFxConfig _powerFxConfig;
+
         private readonly IEnumStore _enumStore;
```
MikeStall, line 22: Since Binding already has the NameResolver, why does it need the config too?

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -24,11 +24,12 @@ namespace Microsoft.PowerFx
     /// <summary>
     /// Holds a set of Power Fx variables and formulas. Formulas are recalculated when their dependent variables change.
     /// </summary>
-    public sealed class RecalcEngine : Engine, IScope, IPowerFxEngine
+    public sealed class RecalcEngine : Engine, IPowerFxEngine
     {
         internal Dictionary<string, RecalcFormulaInfo> Formulas { get; } = new Dictionary<string, RecalcFormulaInfo>();
 
-        internal Dictionary<string, TexlFunction> _customFuncs = new Dictionary<string, TexlFunction>();
```
MikeStall, line 31: >_customFuncs [](http://example.com/codeflow?start=50&length=12)

Win: Got to remove this and move to a symbol table

BruceHaley:
:x: 2 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.2-preview.20220816-317288](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.2-preview.20220816-317288).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfig.FunctionInfos.get()' does not exist in the implementation but it does exist in the contract.
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfig.GetAllFunctionNames()' does not exist in the implementation but it does exist in the contract.
```

</details>
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.2-preview.20220816-317288](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.2-preview.20220816-317288).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.ReflectionFunction.Invoke(Microsoft.PowerFx.Types.FormulaValue[])' does not exist in the implementation but it does exist in the contract.
```

</details>

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -223,7 +319,9 @@ public RenameDriver CreateFieldRenamer(RecordType parameters, DPath pathToRename
             ** but that we don't return any display names for them. Thus, we clone a PowerFxConfig but without 
             ** display name support and construct a resolver from that instead, which we use for the rewrite binding.
             */
-            return new RenameDriver(parameters, pathToRename, updatedName, this, CreateResolver(Config.WithoutDisplayNames()), CreateBinderGlue());
+
+            // $$$ Config.WithoutDisplayNames()
```
MikeStall, line 323: >  // $$$ Config.WithoutDisplayNames() [](http://example.com/codeflow?start=10&length=37)

@lesaltzm  - I'm passing all tests with this removed ... so maybe it wasn't necessary, or we may have a test hole?

If we do need this, I can solve it by passing a flag through CreateResolver (rather than creating whole new resolver)

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
MikeStall: reverted, kept non-public.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -31,27 +35,87 @@ public class Engine : IPowerFxEngine
...
 
         /// <summary>
         /// Create a resolver for use in binding. This is called from <see cref="Check(string, RecordType, ParserOptions)"/>.
         /// Base classes can override this is there are additional symbols not in the config.
         /// </summary>
-        /// <param name="alternateConfig">An alternate config that can be provided. Should default to engine's config if null.</param>        
+        /// <param name="localSymbols">An alternate config that can be provided. Should default to engine's config if null.</param>        
         /// <returns></returns>
-        private protected virtual INameResolver CreateResolver(PowerFxConfig alternateConfig = null)
+        private protected INameResolver CreateResolverInternal(ReadOnlySymbolTable localSymbols = null)
```
LucGenetier, line 104: >CreateResolverInternal [](http://example.com/codeflow?start=40&length=22)

Shall we add [Obsolete] here too?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -116,28 +180,52 @@ public CheckResult Check(string expressionText, RecordType parameterType = null,
...
+            ReadOnlySymbolTable symbolTable = null)
         {
             var bindingConfig = new BindingConfig(allowsSideEffects: options?.AllowsSideEffects == true);
-            return CheckInternal(parse, parameterType, bindingConfig);
+
+            return Check2(parse, bindingConfig, symbolTable);
         }
 
-        private CheckResult CheckInternal(ParseResult parse, RecordType parameterType, BindingConfig bindingConfig)
+        private CheckResult Check2(
```
LucGenetier, line 210: >Check2 [](http://example.com/codeflow?start=28&length=6)

CheckInternal?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -116,28 +180,52 @@ public CheckResult Check(string expressionText, RecordType parameterType = null,
...
+            
             var glue = CreateBinderGlue();
 
             var binding = TexlBinding.Run(
                 glue,
                 parse.Root,
                 resolver,
                 bindingConfig,
-                ruleScope: parameterType._type,
+                ruleScope: null,
```
LucGenetier, line 228: >null [](http://example.com/codeflow?start=27&length=4)

Is this intentional/correct?

src/tests/Microsoft.PowerFx.Interpreter.Tests/ConfigTests.cs
```diff
@@ -0,0 +1,517 @@
...
+        // Drection evaluation with values. 
+        [Fact]
+        public async Task BasicEvaluation()
+        {
+            // Per expression.
+            var engine = new RecalcEngine();
+            engine.UpdateVariable("global", 2.0);
+
+            var locals = new SymbolValues()
+                .AddVariable("local", FormulaValue.New(3));
```
LucGenetier, line 34: UpdateVariable vs. AddVariable
Shouldn't we use the same name?

src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs
```diff
@@ -15,184 +15,16 @@
...
 namespace Microsoft.PowerFx.Core.Glue
 {
-    /// <summary>
-    /// Basic implementation of INameResolver around a <see cref="PowerFxConfig"/> object. 
-    /// This aides in binding and intellisense. 
-    /// Host can override Lookup to provide additional symbols to the expression. 
-    /// </summary>
-    internal class SimpleResolver : INameResolver
+    [Obsolete("Use symbol table")]
+    internal class SimpleResolver : ComposedReadOnlySymbolTable
```
lesaltzm, line 19: This is to avoid breaking changes to Dataverse, yes?
Do we have a task to clean up the SQL target's uses and then remove these Obsolete members?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -31,27 +35,87 @@ public class Engine : IPowerFxEngine
...
+        // Derived engines may only support a subset of these builtins, 
+        // and they may add their own custom ones. 
+        private static readonly ReadOnlySymbolTable _allBuiltinCoreFunctions = NewDefault(Core.Texl.BuiltinFunctionsCore.BuiltinFunctionsLibrary);
+
+        /// <summary>
+        /// Builtin functions supported by this engine. 
+        /// </summary>
+        public ReadOnlySymbolTable SupportedFunctions { get; set; } = _allBuiltinCoreFunctions;
+
+        // Helper to creaet a ReadOnly symbol table around a set of core functions.
```
lesaltzm, line 51: >creaet [](http://example.com/codeflow?start=21&length=6)

sp: create

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -31,27 +35,87 @@ public class Engine : IPowerFxEngine
...
 
+        // All functions that powerfx core knows about. 
+        // Derived engines may only support a subset of these builtins, 
+        // and they may add their own custom ones. 
+        private static readonly ReadOnlySymbolTable _allBuiltinCoreFunctions = NewDefault(Core.Texl.BuiltinFunctionsCore.BuiltinFunctionsLibrary);
+
+        /// <summary>
+        /// Builtin functions supported by this engine. 
+        /// </summary>
+        public ReadOnlySymbolTable SupportedFunctions { get; set; } = _allBuiltinCoreFunctions;
```
lesaltzm, line 49: >set [](http://example.com/codeflow?start=61&length=3)

Should this be either protected or internal set?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -31,27 +35,87 @@ public class Engine : IPowerFxEngine
...
+
+        // Helper to creaet a ReadOnly symbol table around a set of core functions.
+        // Important that this is readonly so that it can be safely shared across engines. 
+        internal static ReadOnlySymbolTable NewDefault(IEnumerable<TexlFunction> coreFunctions)
```
lesaltzm, line 53: >NewDefault [](http://example.com/codeflow?start=44&length=10)

organization nit: this builder static seems to logically belong on ReadOnlySymbolTable, not Engine.

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
lesaltzm: No, this is a symbol that's a name. SymbolName would read as the name of a symbol, at least to me

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -0,0 +1,179 @@
...
+    /// This is a publicly facing class around a <see cref="INameResolver"/>.
+    /// </summary>
+    [DebuggerDisplay("{DebugName}")]
+    public class SymbolTable : ReadOnlySymbolTable
+    {
+        // Expose public setters
+        public new ReadOnlySymbolTable Parent
+        {
+            get => _parent;
+            set
```
lesaltzm, line 57: >set [](http://example.com/codeflow?start=12&length=3)

Consider: `init` so it's impossible to define loops, and this needs to be set either by the constructor or object initializer

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTableOverRecordType.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            _type = type;
+            _debugName = "per-eval";
+        }
+
+        IReadOnlyDictionary<string, NameLookupInfo> IGlobalSymbolNameResolver.GlobalSymbols
+        {
+            get
+            {
+                // $$$ Enumeration is for intellisense, should return in DisplayNames?
```
lesaltzm, line 28: >// $$$ Enumeration is for intellisense, should return in DisplayNames [](http://example.com/codeflow?start=16&length=69)

probably, or at least set the DisplayName in the NameLookupInfo constructed on line 55 and get intellisense to use that.

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTableOverRecordType.cs
```diff
@@ -0,0 +1,63 @@
...
+                }
+
+                return map;
+            }
+        }
+
+        internal override bool TryLookup(DName name, out NameLookupInfo nameInfo)
+        {
+            // $$$ Not handling DisplayNames here.... 
+            // Dtypoe.TryGetDisplayNameForColumn?
```
lesaltzm, line 42: Should be the same as other symbol table lookups, like what was previously in PfxConfig, but pulling the displaynameprovider from the _type
```
            if (_type._type.DisplayNameProvider.TryGetDisplayName(name, out displayName))
            {
                lookupName = name;
            }
            else if (_type._type.DisplayNameProvider.TryGetLogicalName(name, out var logicalName))
            {
                lookupName = logicalName;
                displayName = name;
            }

            __type.TryGetFieldType(lookupName, out var type)
```
and ensure the DisplayName is set in the NameLookupInfo that's returned

src/libraries/Microsoft.PowerFx.Core/Public/Config/VersionHash.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            _value++;
+        }
+
+        public VersionHash Combine(VersionHash other)
+        {
+            return new VersionHash(HashCombine(_value, other._value));
+        }
+
+        internal static int HashCombine(int a, int b)
```
lesaltzm, line 57: Should these methods be internal? When would a host need to create or increment their own hash?

Also, does this class need to be public, or should it also be internal?

src/libraries/Microsoft.PowerFx.Core/Utils/GetTokensUtils.cs
```diff
@@ -35,10 +35,10 @@ internal static class GetTokensUtils
                     {
                         case BindKind.Control:
                         case BindKind.OptionSet:
-                        case BindKind.PowerFxResolvedObject:
                             tokens[item.Name] = TokenResultType.HostSymbol;
                             break;
                         case BindKind.LambdaField:
+                        case BindKind.PowerFxResolvedObject:
```
lesaltzm, line 41: why this change?

lesaltzm:
>             var str = name.Value;

seems like this is only handling recalc engine variables now, comment out of date?

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngineResolver.cs:74 in 352ac7e. [](commit_id = 352ac7ecc86b038b569665bfcb4f97cb66f8a45b, deletion_comment = False)

src/tests/Microsoft.PowerFx.Interpreter.Tests/ConfigTests.cs
```diff
@@ -0,0 +1,517 @@
...
+            Assert.Equal(6.0, result.ToObject());
+
+            // But rebinding will pickup new ... 
+            result = await engine.EvalAsync("Multiply(2)", CancellationToken.None, symbolTable: s1);
+            Assert.Equal(8.0, result.ToObject());
+        }
+
+        // $$$ Shadowing!
+        // $$$ and with overloads!!
+
```
lesaltzm, line 159: Commenting on this, would love to see those tests

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTableOverRecordType.cs
lesaltzm: This is a little tricky if all the symbol tables need to handle it themselves. Maybe we add `DisplayNameProvider` to the SymbolTable interface, and handle it in a wrapper?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
lesaltzm: Yeah, looks like a test gap: I took another look at it, and added #562, removes the need for this.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
MikeStall: protected set is good. 
And then engines that can be flexible here can expose options to update it.  (say if you wanted to remove a function from the interpreter)

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
MikeStall: That's a good idea - good way to get compile-time support and we can always loosen it later. 

we have a test for loop detection, but compile support is better.

src/libraries/Microsoft.PowerFx.Core/Public/Config/VersionHash.cs
MikeStall: They probably could all be internal.  Can always make public later. 
Just needs to be public if a host wants to tell if a symbol table has changed, but I don't think they need to do that.

src/libraries/Microsoft.PowerFx.Core/Utils/GetTokensUtils.cs
MikeStall: This is a rule-scope vs. global name resolver thing. 

We have LSP tests checking these TokenResultTypes.

Before, all the locals where rule-scope, which meant LambdaField, which returns TokenResultType.Variable.

Now, all the locals are NameResolver, which is  BindKind.PowerFxResolvedObject:.

src/libraries/Microsoft.PowerFx.Core/Glue/SimpleResolver.cs
MikeStall: Right. 
I can file a task.

src/libraries/Microsoft.PowerFx.Core/Utils/GetTokensUtils.cs
lesaltzm: Right, that sort of makes me think they should also be HostSymbol token types, since really, they're coming from the host.

MikeStall: but there are existing tests that expect these engine globals to be a TokenResultType.Variable.

lesaltzm: Right, I was being lazy because we were reusing scopes. If we were doing it from scratch, they should be HostSymbol. I'd change the tests. 

All this does is drive formulabar colorizing, for what it's worth.

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTableOverRecordType.cs
MikeStall: discussed offlne. 
We do need a display name lookup here.  I'll do what rowscope does.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
MikeStall: Yes. 
We've removed the parameterType altogether.
This information is now passed in via symbolTable, and via INameResolver

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
MikeStall: no. This is the new one.

src/tests/Microsoft.PowerFx.Interpreter.Tests/ConfigTests.cs
MikeStall: I'll rename to just Add(). 

Add existing should fail. 

Engine.UpdateVariable is existing method.

BruceHaley:
:x: 2 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.2-preview.20220817-317421](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.2-preview.20220817-317421).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfig.FunctionInfos.get()' does not exist in the implementation but it does exist in the contract.
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfig.GetAllFunctionNames()' does not exist in the implementation but it does exist in the contract.
```

</details>
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.2-preview.20220817-317421](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.2-preview.20220817-317421).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.ReflectionFunction.Invoke(Microsoft.PowerFx.Types.FormulaValue[])' does not exist in the implementation but it does exist in the contract.
```

</details>

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTableOverRecordType.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            var hasDisplayName = DType.TryGetDisplayNameForColumn(_type._type, name, out var displayName);
+
+            return new NameLookupInfo(
+                   BindKind.PowerFxResolvedObject,
+                   type._type,
+                   DPath.Root,
+                   0,
+                   data: new NameSymbol(name),
+                   displayName: hasDisplayName ? new DName(displayName) : default);
```
MikeStall, line 70: >   displayName: hasDisplayName ? new DName(displayName) : default); [](http://example.com/codeflow?start=16&length=67)

This passes new cases from #569

BruceHaley:
:x: 2 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.2-preview.20220817-317421](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.2-preview.20220817-317421).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfig.FunctionInfos.get()' does not exist in the implementation but it does exist in the contract.
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfig.GetAllFunctionNames()' does not exist in the implementation but it does exist in the contract.
```

</details>
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.2-preview.20220817-317421](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.2-preview.20220817-317421).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.ReflectionFunction.Invoke(Microsoft.PowerFx.Types.FormulaValue[])' does not exist in the implementation but it does exist in the contract.
```

</details>

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
LucGenetier: Should it be NamedSymbol then? (with a d)

src/libraries/Microsoft.PowerFx.Core/Localization/ExpressionLocalizationHelper.cs
```diff
@@ -32,7 +32,7 @@ internal static string ConvertExpression(string expressionText, RecordType param
                 formula.ParseTree,
                 resolver,
                 config,
-                ruleScope: parameters._type,
+                ruleScope: parameters?._type,
                 updateDisplayNames: toDisplay,
```
lesaltzm, line 36: Should we be passing parameters at all here anymore?
If it's all been moved to resolver, this could be cleaned up.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -31,28 +35,77 @@ public class Engine : IPowerFxEngine
...
+            return FunctionInfos.Select(func => func.Name).Distinct();
+        }
+
+        // Additional symbols for the engine.
+        // A derived engine can replace this completely to inject engine-specific virtuals. 
+        // These symbols then feed into the resolver
+        protected SymbolTable EngineSymbols { get; set; }
+
+        // Move away
+
```
lesaltzm, line 74: nit: is this a todo comment, or something else?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
lesaltzm: it's an optional param with a null default, any reason not to just leave it out?

src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
```diff
@@ -0,0 +1,300 @@
...
+
+        IExternalEntity INameResolver.CurrentEntity => default;
+
+        DName INameResolver.CurrentProperty => default;
+
+        DPath INameResolver.CurrentEntityPath => default;
+
+        bool INameResolver.SuggestUnqualifiedEnums => false;
+
+        internal readonly Dictionary<DName, IExternalEntity> _environmentSymbols = new Dictionary<DName, IExternalEntity>();
```
lesaltzm, line 140: small readability thought:
Can we move all the default/notimpl left overs to the end/put it in a  region to hide or something like that?

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTableEnumerator.cs
```diff
@@ -0,0 +1,54 @@
...
+            var seen = new HashSet<ReadOnlySymbolTable>();
+
+            foreach (var table in _tables)
+            {
+                var t = table;
+                while (t != null)
+                {
+                    if (!seen.Add(t))
+                    {
+                        throw new InvalidOperationException($"Loop in symbol tables");
```
lesaltzm, line 40: This probably isn't an issue anymore, given that it's now `set; init;` on specifying parents. Still doesn't hurt to be defensive.

LucGenetier (approved):

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTableOverRecordType.cs
```diff
@@ -0,0 +1,73 @@
...
+
+                return map;
+            }
+        }
+
+        internal override bool TryLookup(DName name, out NameLookupInfo nameInfo)
+        {
+            // Lookup must handle display names
+            // This is consistent with RowScope. 
+            if (!DType.TryGetConvertedDisplayNameAndLogicalNameForColumn(_type._type, name, out var logicalName, out _))
```
lesaltzm, line 45: >out _ [](http://example.com/codeflow?start=113&length=5)

Note that this is safe to discard, as non-canvas PowerFx never attempts a display -> display rewrite.

src/libraries/Microsoft.PowerFx.Core/Localization/ExpressionLocalizationHelper.cs
MikeStall: future pr?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
MikeStall: I just wanted to be explicit - rule scope distinction was such a critical issue here.

src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
```diff
@@ -0,0 +1,300 @@
...
+        /// <summary>
+        /// This can be compared to determine if the symbol table was mutated during an operation. 
+        /// </summary>
+        internal virtual VersionHash VersionHash => _parent == null ?
+            _version : _version.Combine(_parent.VersionHash);
+
+        /// <summary>
+        /// Notify the symbol table has changed. 
+        /// </summary>
+        public void Inc()
```
bryancroteau-MSFT, line 43: Why is this public?

src/libraries/Microsoft.PowerFx.Interpreter/Environment/ReadOnlySymbolValues.cs
```diff
@@ -0,0 +1,83 @@
...
+            return _lastSnapshot;
+        }
+
+        private ReadOnlySymbolTable _lastSnapshot;
+
+        protected abstract ReadOnlySymbolTable GetSymbolTableSnapshotWorker();
+
+        public virtual bool TryGetValue(string name, out FormulaValue value)
+        {
+            value = null;
```
bryancroteau-MSFT, line 68: Why is this always returning false/null?

src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
MikeStall: it lets callers invalidate a symbol table. 
This can be useful if the symbol table refers to some external thing and that has changed.

src/libraries/Microsoft.PowerFx.Interpreter/Environment/ReadOnlySymbolValues.cs
MikeStall: Virtual function. 
Derived classes can override.

bryancroteau-MSFT: If that is the case, why not make this function abstract?

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
MikeStall: this is consistent with NameLookupInfo

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
MikeStall: I think this is a merge  conflict - I should remove altogether.

src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
MikeStall: yes!

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTableEnumerator.cs
MikeStall: right - definitely want to be defensive, especially around loops. 
 Still including the test to catch it.

src/libraries/Microsoft.PowerFx.Core/Localization/ExpressionLocalizationHelper.cs
lesaltzm: sure.

lesaltzm (approved):
:ship;

bryancroteau-MSFT (approved):
Symbol/Value tables look good

BruceHaley:
:x: 2 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.2-preview.20220818-317567](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.2-preview.20220818-317567).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfig.FunctionInfos.get()' does not exist in the implementation but it does exist in the contract.
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfig.GetAllFunctionNames()' does not exist in the implementation but it does exist in the contract.
```

</details>
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.2-preview.20220818-317567](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.2-preview.20220818-317567).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.ReflectionFunction.Invoke(Microsoft.PowerFx.Types.FormulaValue[])' does not exist in the implementation but it does exist in the contract.
```

</details>

PR #507, opened 2022-08-01
OData Delegation

jcoc611-microsoft:
Adds `QueryableTableValue` as a base class for all TableValue_s coming from external data sources and supporting operations that can be delegated (similar to `IQueryable` except the translation is from PowerFX IntermediateNode_s rather than from C# expressions).

Also adds `ODataQueryableTableValue` and a `ODataVisitor` for delegation to OData endpoints.

Supports `Filter`, `Sort`, `FirstN` in a composable manner.

This change is a first step towards delegable connectors, working at runtime to delegate key operations. In the future, we will add more support for parse/design-time detection of delegable operations.


src/libraries/Microsoft.PowerFx.Connectors/ODataVisitor.cs
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        public override string Visit(BooleanLiteralNode node, object context)
+        {
+            return node.LiteralValue.ToString(CultureInfo.InvariantCulture);
+        }
+
+        public override string Visit(ColorLiteralNode node, object context)
+        {
+            throw new System.NotImplementedException();
```
lesaltzm, line 34: We may want a specific `NonDelegableException` that we can catch and fall back to trying in-memory.

src/libraries/Microsoft.PowerFx.Connectors/ODataVisitor.cs
```diff
@@ -0,0 +1,185 @@
...
+                // BinaryOpKind.InScalarTable => expr,
+                // BinaryOpKind.ExactInScalarTable => expr,
+                // BinaryOpKind.InRecordTable => expr,
+                BinaryOpKind.AddNumbers => "+",
+                BinaryOpKind.AddDateAndTime => "+",
+                BinaryOpKind.AddDateAndDay => "+",
+                BinaryOpKind.AddDateTimeAndDay => "+",
+                BinaryOpKind.AddTimeAndMilliseconds => "+",
+                BinaryOpKind.DateDifference => "-",
+                BinaryOpKind.TimeDifference => "-",
```
lesaltzm, line 117: Some of these may be interesting, our date/time addition semantics may not be the same as what the target implements. We should check with Sagar/Reza how the JS runtime handles it.

src/libraries/Microsoft.PowerFx.Interpreter/Values/QueryableTableValue.cs
```diff
@@ -0,0 +1,53 @@
...
+        public sealed override IEnumerable<DValue<RecordValue>> Rows
+        {
+            get
+            {
+                // Not the best, but doesn't require major breaking changes to TableValue
+                lock (_lock)
+                {
+                    if (_taskRows == null)
+                    {
+                        _taskRows = GetRowsAsync();
```
MikeStall, line 45: >GetRowsAsync [](http://example.com/codeflow?start=36&length=12)

You can't call this under a lock. 
Call in a local and then just assign under the lock - possible that it may run twice, but that's uinevitable.

src/libraries/Microsoft.PowerFx.Connectors/ODataQueryableTableValue.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Web;
+using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Connectors
+{
+    public abstract class ODataQueryableTableValue : QueryableTableValue
+    {
+        private readonly Uri _uriBase;
+        internal readonly NameValueCollection ODataParams;
```
MikeStall, line 15: >NameValueCollection [](http://example.com/codeflow?start=26&length=19)

Is this like $filter, $select, etc? 
Could we make this strongly typed, or is it intrinsically dynamic?

src/tests/Microsoft.PowerFx.Connectors.Tests/ODataDelegationTests.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Threading.Tasks;
+using System.Web;
+using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Core.Tests;
+using Microsoft.PowerFx.Types;
+using Xunit;
+
+namespace Microsoft.PowerFx.Connectors.Tests
+{
+    public class ODataDelegationTests : PowerFxTest
```
MikeStall, line 16: great test!

src/libraries/Microsoft.PowerFx.Interpreter/Values/QueryableTableValue.cs
jcoc611-microsoft: Yeah I wasn't super sure about this one...do you recommend using `Lazy<ReadOnlyList<RecordValue>>` instead?

src/libraries/Microsoft.PowerFx.Connectors/ODataVisitor.cs
jcoc611-microsoft: Added

src/libraries/Microsoft.PowerFx.Connectors/ODataQueryableTableValue.cs
jcoc611-microsoft: Added struct type

src/libraries/Microsoft.PowerFx.Interpreter/Values/QueryableTableValue.cs
jcoc611-microsoft: Switched to `Lazy<Task...>` so it is only initialized once and needs no lock

src/libraries/Microsoft.PowerFx.Connectors/ODataVisitor.cs
```diff
@@ -0,0 +1,237 @@
...
+        {
+            return node.Child.Accept(this, runContext);
+        }
+
+        public override string Visit(CallNode node, DelegationRunContext runContext)
+        {
+            // Special cases for functions supported by OData
+            if (node.Function is NotFunction)
+            {
+                return $"NOT({node.Args[0].Accept(this, runContext)})";
```
lesaltzm, line 61: From OData Spec
not operator is lowercase: https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_Not

As are the others (startswith, endswith,...)

The implementation that you're testing against may not care about case, but we should match the spec for these.

src/libraries/Microsoft.PowerFx.Connectors/ODataVisitor.cs
```diff
@@ -0,0 +1,237 @@
...
+                StringValue stringValue => $"'{stringValue.Value}'",
+                TimeValue timeValue => throw new NotDelegableException(),
+                _ => throw new NotDelegableException(),
+            };
+        }
+
+        private static string ODataOpFromUnaryOp(UnaryOpKind op) =>
+            op switch
+            {
+                UnaryOpKind.Negate => "NOT",
```
lesaltzm, line 151: same here. Also, can this be consolidated with the Not Function Call impl?

src/libraries/Microsoft.PowerFx.Interpreter/NotDelegableException.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+
+namespace Microsoft.PowerFx.Interpreter
+{
+    public sealed class NotDelegableException : Exception
```
lesaltzm, line 8: internal?

lesaltzm, line 8: Add a comment that this is used to fail attempted delegation and fall back to in memory evaluation, and should always be caught.

src/libraries/Microsoft.PowerFx.Connectors/ODataQueryableTableValue.cs
```diff
@@ -0,0 +1,125 @@
...
+
+namespace Microsoft.PowerFx.Connectors
+{
+    public abstract class ODataQueryableTableValue : QueryableTableValue
+    {
+        private readonly Uri _uriBase;
+
+        private readonly ODataParams _odataParams;
+
+        // internal ODataQueryableTableValue(IRContext irContext, NameValueCollection odataParams)
```
lesaltzm, line 21: nit: clean up commented out code

src/libraries/Microsoft.PowerFx.Interpreter/Values/QueryableTableValue.cs
```diff
@@ -0,0 +1,53 @@
...
+        public bool HasCachedRows => _lazyTaskRows.IsValueCreated;
+
+        protected abstract Task<List<DValue<RecordValue>>> GetRowsAsync();
+
+        // Not the best, but doesn't require major breaking changes to TableValue
+        public sealed override IEnumerable<DValue<RecordValue>> Rows =>
+            _lazyTaskRows.Value.GetAwaiter().GetResult();
+    }
+
+    public readonly struct DelegationRunContext
```
lesaltzm, line 40: Separate file?

lesaltzm: Also, internal?

src/libraries/Microsoft.PowerFx.Connectors/ODataQueryableTableValue.cs
```diff
@@ -0,0 +1,125 @@
...
+            }
+        }
+
+        internal ODataParams WithCount() => new ODataParams(true, _filter, _orderby, _top);
+
+        internal ODataParams WithFilter(string filterNew)
+        {
+            if (_filter != null)
+            {
+                return new ODataParams(_count, $"{_filter} and {filterNew}", _orderby, _top);
```
MikeStall, line 115: >{_filter} and {filterNew} [](http://example.com/codeflow?start=49&length=25)

Do we need parenthesis here to protect against precedence?

A is "false" // false
B is "false or true" // true

"false and false or true" // true!!!

src/libraries/Microsoft.PowerFx.Connectors/ODataVisitor.cs
```diff
@@ -0,0 +1,237 @@
...
+                return $"endsWith({node.Args[0].Accept(this, runContext)}, {node.Args[1].Accept(this, runContext)})";
+            }
+
+            // Fallback: try to evaluate call node, fail if it tries to access a delegated scope
+            try
+            {
+                FormulaValue result = runContext.EvalAsync(node).Result;
+                return SerializeLiteralValue(result);
+            }
+            catch (KeyNotFoundException)
```
MikeStall, line 80: >KeyNotFoundException [](http://example.com/codeflow?start=19&length=20)

How do we get here?

src/libraries/Microsoft.PowerFx.Connectors/ODataVisitor.cs
```diff
@@ -0,0 +1,237 @@
...
+
+namespace Microsoft.PowerFx.Connectors
+{
+    internal class ODataVisitor : IRNodeVisitor<string, DelegationRunContext>
+    {
+        internal static readonly ODataVisitor I = new ODataVisitor();
+
+        public override string Visit(TextLiteralNode node, DelegationRunContext runContext)
+        {
+            return $"'{node.LiteralValue}'";
```
MikeStall, line 23: >{node.LiteralValue} [](http://example.com/codeflow?start=22&length=19)

Is there other escaping that needs to happen here? 
Eg, what if the node has a ' in it?

MikeStall: Should these *LiteralNode visitors share code with SerializeLiteralValue?  (especially around some of the formatting)

src/tests/Microsoft.PowerFx.Connectors.Tests/ODataDelegationTests.cs
```diff
@@ -0,0 +1,70 @@
...
+            : base(tableType, new Uri("https://contoso.com/api/list"))
+        {
+        }
+
+        private TestODataTableValue(TableType tableType, ODataParams odataParams)
+            : base(tableType, new Uri("https://contoso.com/api/list"), odataParams)
+        {
+        }
+
+        protected override ODataQueryableTableValue WithParameters(ODataParams odataParamsNew)
```
MikeStall, line 60: >WithParameters [](http://example.com/codeflow?start=52&length=14)

Is this what actually executes the query? 

And this should throw  NotDelegableException if the runtime source does not support delegation?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -70,6 +71,17 @@ public static FormulaValue FirstN(IRContext irContext, FormulaValue[] args)
...
+                try
+                {
+                    return queryableTable.FirstN((int)arg1.Value);
+                }
+                catch (NotDelegableException)
+                {
+                }
+            }
+
             var rows = arg0.Rows.Take((int)arg1.Value);
```
MikeStall, line 85: >arg0.Rows.Take((int)arg1.Value); [](http://example.com/codeflow?start=23&length=32)

For a future PR ...  (@anderson-joyle )
The symmetry between line 78 and 85 makes me wonder ...
What if FirstN was a virtual on TableValue, and the default impl was line 85. 

Then this method becomes trivial....

src/tests/Microsoft.PowerFx.Connectors.Tests/ODataDelegationTests.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            return new TestODataTableValue((TableType)Type, odataParamsNew);
+        }
+
+        protected override Task<List<DValue<RecordValue>>> GetRowsAsync()
```
MikeStall, line 65: >GetRowsAsync() [](http://example.com/codeflow?start=59&length=14)

So this is what actually executes the query? 
Does it need access to the odataparams? (it's private field ... expose so we can read here?)

src/libraries/Microsoft.PowerFx.Connectors/ODataVisitor.cs
jcoc611-microsoft: updated as lowercase, thanks!

src/libraries/Microsoft.PowerFx.Connectors/ODataVisitor.cs
jcoc611-microsoft: consolidated

src/libraries/Microsoft.PowerFx.Interpreter/NotDelegableException.cs
jcoc611-microsoft: made internal

jcoc611-microsoft: added

src/libraries/Microsoft.PowerFx.Connectors/ODataQueryableTableValue.cs
jcoc611-microsoft: removed

src/libraries/Microsoft.PowerFx.Interpreter/Values/QueryableTableValue.cs
jcoc611-microsoft: separated and made internal

src/libraries/Microsoft.PowerFx.Connectors/ODataQueryableTableValue.cs
jcoc611-microsoft: added parenthesis, thanks for the catch!

src/libraries/Microsoft.PowerFx.Connectors/ODataVisitor.cs
jcoc611-microsoft: good point, refactored to use the same methods...also added escaping of `'` with double single quotes `''` per spec

src/tests/Microsoft.PowerFx.Connectors.Tests/ODataDelegationTests.cs
jcoc611-microsoft: discussed offline...the actual query is executed by the derived class override `GetRowsAsync()`

src/libraries/Microsoft.PowerFx.Connectors/ODataQueryableTableValue.cs
```diff
@@ -0,0 +1,118 @@
...
+        protected ODataQueryableTableValue(TableType tableType, Uri uriBase, ODataParams odataParams = default)
+            : base(IRContext.NotInSource(tableType))
+        {
+            _uriBase = uriBase;
+            _odataParams = odataParams;
+        }
+
+        protected abstract ODataQueryableTableValue WithParameters(ODataParams odataParamsNew);
+
+        public Uri GetUri()
```
jcoc611-microsoft, line 29: @MikeStall do you still want me to remove _uriBase and instead have it as a parameter here to GetUri? or maybe even replace this method with `NameValueCollection GetQueryParameters()`?

src/libraries/Microsoft.PowerFx.Connectors/ODataVisitor.cs
jcoc611-microsoft: `KeyNotFoundException` is thrown when there is a variable accessed that is not defined in the current scope...effectively this happens when the PowerFX expression has a non-trivial reference of a would-be variable from the Table scope and we cannot delegate the execution

src/libraries/Microsoft.PowerFx.Connectors/ODataQueryableTableValue.cs
MikeStall: I'd remove _uriBase from here and keep in the derive class.

What about just expose ODataParams as a property - it's already a bublic type and a key field on this class - so is there a benefit to not expose? 

From there, the other methods can be helpers on that ODataParams:

1. AddTo(NameValueCollection query) is already present. 
2. ODataParams.GetUri(uriBase) - this is a convenience method.

src/libraries/Microsoft.PowerFx.Interpreter/Values/QueryableTableValue.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        private readonly Lazy<Task<List<DValue<RecordValue>>>> _lazyTaskRows;
+
+        internal QueryableTableValue(IRContext irContext)
+            : base(irContext)
+        {
+            _lazyTaskRows = new Lazy<Task<List<DValue<RecordValue>>>>(() => GetRowsAsync());
+        }
+
+        internal abstract TableValue Filter(LambdaFormulaValue lambda, EvalVisitor runner, EvalVisitorContext context);
```
MikeStall, line 21: >TableValue [](http://example.com/codeflow?start=26&length=10)

@lesaltzm  / @anderson-joyle  - I'm wondering if eventually all of these get pushed down to be virtuals on TableValue directly, or if it makes sense to stay here.

src/tests/Microsoft.PowerFx.Connectors.Tests/ODataDelegationTests.cs
```diff
@@ -0,0 +1,74 @@
...
+using Xunit;
+
+namespace Microsoft.PowerFx.Connectors.Tests
+{
+    public class ODataDelegationTests : PowerFxTest
+    {
+        [Theory]
+        [InlineData("Filter(Table, x > 10)", "https://contoso.com/api/list?$filter=x+gt+10")]
+        [InlineData("Filter(Table, x > 10 + 5)", "https://contoso.com/api/list?$filter=x+gt+15")]
+        [InlineData("Filter(Table, x > 2 * myNumber)", "https://contoso.com/api/list?$filter=x+gt+100")]
```
MikeStall, line 21: > x > 2 * myNumber [](http://example.com/codeflow?start=34&length=17)

This is a really great test - illustrating the cross over between OData filter and interpreter.

src/libraries/Microsoft.PowerFx.Connectors/ODataVisitor.cs
MikeStall: This is something that Fx team should followup on and find a better way.  We shouldn't rely on this specific exception

src/libraries/Microsoft.PowerFx.Interpreter/Values/DelegationRunContext.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Threading.Tasks;
+using Microsoft.PowerFx.Core.IR.Nodes;
+
+namespace Microsoft.PowerFx.Types
+{
+    internal readonly struct DelegationRunContext
```
MikeStall, line 9: >DelegationRunContext [](http://example.com/codeflow?start=29&length=20)

Can this move to PowerFx.Connectors? 

It looks like it's just used for the ODataVisitor?
And rename to ODataVisitorContext.cs?
And add  a parameter

/// Context for the <see=ODataVisitor>

jcoc611-microsoft: yep done

src/libraries/Microsoft.PowerFx.Connectors/ODataQueryableTableValue.cs
jcoc611-microsoft: exposed `ODataParams` and added convenience `GetUri(uriBase)`. Removed `_uriBase` from base class

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Connectors/ODataVisitor.cs
MikeStall: Filed #581 to track this.

PR #509, opened 2022-08-02
Simplify enum store builder logic.

MikeStall:
EnumStoreBuilder can just return the enum list directly.
Use a single cache.


src/libraries/Microsoft.PowerFx.Core/Types/Enums/EnumStoreBuilder.cs
```diff
@@ -13,8 +13,9 @@ namespace Microsoft.PowerFx.Core.Types.Enums
     /// <summary>
     /// Static class used to store built in Power Fx enums.
     /// </summary>
-    internal sealed class EnumStoreBuilder
+    internal sealed class EnumStoreBuilder : IEnumStore
     {
+        // These are used by Power Apps 
         #region Default Enums
         internal static IReadOnlyDictionary<string, string> DefaultEnums { get; } =
```
MikeStall, line 20: The fact that the TValue is string means that we're re-parsing the enum specs constantly.  If I don't cache, it's a huge 6x perf hit.

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Types/Enums/EnumStoreBuilder.cs
```diff
@@ -327,56 +341,47 @@ internal EnumStoreBuilder WithDefaultEnums()
...
-        internal EnumStore Build()
+        internal IEnumStore Build()
         {
-            return EnumStore.Build(new List<EnumSymbol>(EnumSymbols));
+            return this;
         }
         #endregion
 
-        private Dictionary<string, DType> RegenerateEnumTypes()
+        public IEnumerable<EnumSymbol> EnumSymbols
```
MikeStall, line 350: This is implementing IEnumStore. 
That interface forces a linear scan on every enum lookup. Maybe we need a more efficient lookup interface/

lesaltzm (approved):
Double check integration with PA, but this should be good!

src/libraries/Microsoft.PowerFx.Core/Types/Enums/EnumStoreBuilder.cs
LucGenetier: Is there any way to avoid "we're re-parsing the enum specs constantly"? 
Not just improving by 6x but resolving it all.

src/libraries/Microsoft.PowerFx.Core/Types/Enums/EnumStoreBuilder.cs
```diff
@@ -284,15 +285,23 @@ internal sealed class EnumStoreBuilder
             };
         #endregion
 
+        // Map from enum name to its type spec. 
         private readonly Dictionary<string, string> _workingEnums = new Dictionary<string, string>();
```
LucGenetier, line 289: >new Dictionary<string, string>(); [](http://example.com/codeflow?start=68&length=33)

Let's use a ConcurrentDictionary<string, string>, that will avoid the lock problem

src/libraries/Microsoft.PowerFx.Core/Types/Enums/EnumStoreBuilder.cs
LucGenetier: ConcurrentDictionary<string, string> manages it all for us

LucGenetier: I'd be interested in knowing the perf results you can get with it in your tests

LucGenetier (changes requested):
:clock1:

PR #510, opened 2022-08-02
 prevent support scalar in MultiselectOptionset table

kartikkr11:
This change basically restricts a scalar value like optionsetvalue to be used in the IN operator with a MultiselectOptionset Table, this has resulted in issues where formulas in App don't work in edit mode until they are forced to be reevaluated

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -2640,7 +2640,7 @@ namespace Microsoft.PowerFx.Core.Binding
                     }
 
                     // scalar in table: RHS must be a one column table. We'll allow coercion.
-                    if (typeRight.IsTable)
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -2717,13 +2717,13 @@ private bool CheckInArgTypes(TexlNode left, TexlNode right)
                         }
 
                         var typedName = names.Single();
-                        if (!typeLeft.CoercesTo(typedName.Type))
```
lesaltzm, line 2720: Why these changes? Can you leave a comment about the logic?

kartikkr11: updated it and added comments.

lesaltzm (approved):

PR #522, opened 2022-08-05
scalar in table optionset

kartikkr11:
From the test case, TableA.Optionsetcolumn in multiselectOptionset was supported previously as a delegable call to dataverse resulting in errors, this PR makes it non-delegable.

Ex: Filter(Accounts, optionsetValue in MultiselectTable)  // is now non delegated, instead of being a delegable call.
Filter(Accounts, 'Primary Contact'.optionsetcolumn in MultiselectTable) supported.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -2649,13 +2649,6 @@ namespace Microsoft.PowerFx.Core.Binding
                             return false;
                         }
 
-                        // restrict scalar in multiselect optionset table. (Not TableA.OptionsetColumn)
-                        if (typeRight.IsMultiSelectOptionSet() && typeLeft.AssociatedDataSources?.Count == 0)
-                        {
-                            _txb.ErrorContainer.EnsureError(DocumentErrorSeverity.Severe, left, TexlStrings.ErrBadType_Type, typeLeft.GetKindString());
-                            return false;
-                        }
-
                         var typedName = names.Single();
                         if (typedName.Type.Accepts(typeLeft) || typeLeft.Accepts(typedName.Type))
                         {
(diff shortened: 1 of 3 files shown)


src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -2649,6 +2649,13 @@ private bool CheckInArgTypes(TexlNode left, TexlNode right)
                             return false;
                         }
 
+                        // restrict scalar in multiselect optionset table. (Not TableA.OptionsetColumn)
+                        if (typeRight.IsMultiSelectOptionSet() && typeLeft.AssociatedDataSources?.Count == 0)
+                        {
+                            _txb.ErrorContainer.EnsureError(DocumentErrorSeverity.Severe, left, TexlStrings.ErrBadType_Type, typeLeft.GetKindString());
```
lesaltzm, line 2655: This seems arbitrary. Why do we care whether the lhs is from a data source or not? I understand for delegation, but what about in memory eval?

src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/OpDelegationStrategy.cs
```diff
@@ -299,6 +299,12 @@ public virtual bool IsSupportedOpNode(TexlNode node, OperationCapabilityMetadata
                 return true;
             }
 
+            if ((leftType.IsMultiSelectOptionSet() && !rightType.IsAggregate) || (rightType.IsMultiSelectOptionSet() && !leftType.IsAggregate))
```
lesaltzm, line 302: Shouldn’t this also check if it’s ‘in’ as opposed to a different binary op?

CarlosFigueiraMSFT (approved):
:shipit:

PR #531, opened 2022-08-09
Adding tests for If function code coverage.

anderson-joyle:


BruceHaley:

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
```diff
@@ -63,6 +63,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <None Remove="ExpressionTestCases\If_AllowsSideEffects.txt" />
     <None Remove="ExpressionTestCases\SqrtT.txt" />
   </ItemGroup>
```
MikeStall, line 68: can we remove? VS may have added accidentaly

anderson-joyle: Okay. Should I remove both?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/If_AllowsSideEffects.txt
```diff
@@ -0,0 +1,13 @@
+﻿#SETUP: EnableExpressionChaining
+
+>> If(false, "abc1", {x:3})
+{x:3}
```
MikeStall, line 4: Can we add a comment here to explain why this is ok?
(possibly reference IfFunction.IsArgTypeInconsequential) 

"This If() expression is allowed to have mismatching types between arg1 and arg2 because a) it's behavior, b) it's verified nobody is actually using the return value."

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -219,7 +220,11 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
                 {
                     var result = await ptr(this, context.IncrementStackDepthCounter(childContext), node.IRContext, args);
 
-                    Contract.Assert(result.IRContext.ResultType == node.IRContext.ResultType || result is ErrorValue || result.IRContext.ResultType is BlankType);
+                    if (IfFunction.CanCheckIfReturn(func))
```
MikeStall, line 223: >CanCheckIfReturn [](http://example.com/codeflow?start=35&length=16)

Note to future readers - this is needed to loosen the assert.
We can't fully determine when the type doesn't matter, so we just ignore If alttogether.

 If(false, "abc1", {x:3}) // Behavior

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
anderson-joyle: Done.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/If_AllowsSideEffects.txt
anderson-joyle: Done.

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/If.cs
```diff
@@ -157,6 +157,13 @@ public override bool HasSuggestionsForParam(int argumentIndex)
             return argumentIndex > 1;
         }
 
+        // When IsArgTypeInconsequential returns true, the runtime result of If may not match the binder's expectation.
+        // So use this helper to skip asserts comparing runtime and bind time types.
+        internal static bool CanCheckIfReturn(TexlFunction func)
```
LucGenetier, line 162: >CanCheckIfReturn [](http://example.com/codeflow?start=29&length=16)

Would write it as an extension method and name it "IsNotIfFunction()"

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -91,7 +91,7 @@ protected override async Task<RunResult> RunAsyncInternal(string expr, string se
 
                 if (iSetup.HandlerName != null)
                 {
-                    if (!SetupHandlers.TryGetValue(setupHandlerName, out var handler))
+                    if (!SetupHandlers.TryGetValue(iSetup.HandlerName, out var handler))
```
LucGenetier, line 94: >iSetup [](http://example.com/codeflow?start=51&length=6)

Why do we need this now?

anderson-joyle: This was suggested by @CarlosFigueiraMSFT. If I understood correctly, in the event where we have more than one setup handler in a text file (e.g. #SETUP: EnableExpressionChaining, AnotherHandler), the setupHandlerName will be a concat value of all handlers.
@CarlosFigueiraMSFT please correct me if this is not the case.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
LucGenetier: Check if node.IRContext.ResultType is DType.Unknown

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
CarlosFigueiraMSFT: This was the case when we needed to have two handlers, but in this PR this is not necessary. We can revert this change now, and add it when we do need it.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/If_AllowsSideEffects.txt
```diff
@@ -0,0 +1,20 @@
...
+
+>> If(false, "abc2", {x:3}; {x:2}); Value(100)
+100
+
+>> If(false, "abc3", {x:3}; If(true, Value(10); {x:3})); Value(101)
+101
+
+>> Value(If(false, "abc4", {x:3}; If(true, Value(10); {x:3})); Value(102))
+102
```
CarlosFigueiraMSFT, line 20: We should add tests where the If value is used (e.g., `If(1>0,100,{x:3}) + 1` - in this case we should have a compile-time error

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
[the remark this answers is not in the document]
LucGenetier: I proposed Anderson not to filter on some give function or set of functions (If and Switch now) but use the return type instead (which is DType.Unknown)

@CarlosFigueiraMSFT Do you think this is a valid approach?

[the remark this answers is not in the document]
anderson-joyle: What you think of the following approach? It doesn't seem to break any test.
```
if (!func.ReturnType.IsUnknown)
{
    ...
}
```

BruceHaley:

BruceHaley:

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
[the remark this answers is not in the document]
MikeStall: func.ReturnType is set statically  (see TexlFunction) and the same for all invocations of the function. This is meaningless in cases where the function's type depends on its args. In these cases, some functions set to Unknown (IfFunction does).  

But there are many times when TexlFunction.ReturnType is unknown, but we should still run this assert. 

That's because this assert is checking the binding type (computed once we know the parameter types) against the actual runtime type.  That should always be true, except in very rare cases where the binder is wrong. IfFunction() hits one of those rare cases. 


So the question I'd have is how many functions have this true? Is it just IfFunction()? or others?

BruceHaley:

BruceHaley:

BruceHaley:

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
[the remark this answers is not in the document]
anderson-joyle: Kept the original push approach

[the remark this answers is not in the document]
CarlosFigueiraMSFT: Answering Mike's question: other functions have the same issue (IfError currently does, Switch doesn't but it should)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/If_AllowsSideEffects.txt
```diff
@@ -0,0 +1,47 @@
...
+>> If(1<0, "abc2", {x:3}; {x:2})
+{x:2}
+
+>> If(false, "abc3", {x:3}; If(true, Value(10); {x:3})); Value(101)
+101
+
+>> Value(If(false, "abc4", {x:3}; If(true, Value(10); {x:3})); Value(102))
+102
+
+>> If(Value("1")>Value("0"), Blank(), 3)
```
CarlosFigueiraMSFT, line 37: >Value("1")>Value("0") [](http://example.com/codeflow?start=6&length=21)

Since this is a unit test for the If function, we don't need to use the Value calls here (and this is not relevant to this test). Something like `If(1>0,...` would do what we need, and we should have coverage for the Value function in its own test file.

PR #535, opened 2022-08-10
SetProperty check failing scenarios for record and table type

NavneetThekkumpat:
Engine check fails on setProperty() when I try to access Comboxbox or Dropdown or Listbox properties. These expect table and record type values.Engine.Check() Fails and throws " **behavior function in a non behavior property error**" .

**Example Engine check() passes**
 `SetProperty(TextInput1.Text, "hi");
 SetProperty(Toggle1.Value, true);` 
**Example Engine check() fails**
`SetProperty(Dropdown1.Selected, {Value:"2"});
SetProperty(ComboBox1.SelectedItems, Table({Value:"1"},{Value:"2"}));          
 SetProperty(ListBox1.SelectedItems, Table({Value:"1"},{Value:"3"}));`

Adding specific testcases where Check() fails.

Would need these Check() to pass

### DIFF
--- a/src/tests/Microsoft.PowerFx.Interpreter.Tests/CustomFunctions.cs
+++ b/src/tests/Microsoft.PowerFx.Interpreter.Tests/CustomFunctions.cs
@@ -1,9 +1,13 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+using System;
 using System.Collections.Generic;
+using System.Dynamic;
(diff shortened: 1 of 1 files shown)


MikeStall:
We talked offline and the issue was that the ListBox1 didn't have 'SelectedItems' property, so error was legimitate.

NavneetThekkumpat:
This PR no longer needed, fixes in TestEngine solved the above-mentioned problem.

PR #550, opened 2022-08-15
DRAFT: Failing test from pad team

MikeStall:
This is a test from the PAD team. Currently fails.  

We should understand why it fails, and either:
- If it's an Fx bug --> fix it 
- If it's user error --> what's the "right" way to do this?


LucGenetier (approved):

lesaltzm:
Cleared up by #551, now only RecordType is derivable. That should push folks in the right direction of using `ToTable` instead of constructing a derived TableType to wrap a record.

PR #553, opened 2022-08-15
Mark Date <-> Time Coercions in binder

lesaltzm:
Date/Time implicitly are comparable, however the binder does not mark this as a coercion to DateTime, even though it should. This fixes that, and allows the binary op matrix to correctly generate IR for this.

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Time.txt
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Time.txt
@@ -73,6 +73,9 @@ true
 >> Time( 11, 23, 45 ) > Date( 2019, 5, 16 )
 false
 
+>> Date(1899, 12, 30) < Time(12,0,0)
+true
+
 >> Time(1/0, 2, 3) <= Time(8,1,30)
 #Error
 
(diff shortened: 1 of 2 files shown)


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Time.txt
```diff
@@ -67,6 +67,12 @@ false
 >> Time(1/0, 2, 3) < Time(8,1,30)
 #Error
 
+>> Date( 2019, 5, 16 ) > Time( 11, 23, 45 )
+true
+
+>> Time( 11, 23, 45 ) > Date( 2019, 5, 16 )
+false
+
```
CarlosFigueiraMSFT, line 75: Can we have a comparison where the time is greater than the date? I believe something like `>> Date(1899, 12, 30) < Time(12,0,0)` would be true, since that date is the "date 0", and the time should be equivalent to 0.5

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Time.txt
```diff
@@ -67,6 +67,12 @@ false
 >> Time(1/0, 2, 3) < Time(8,1,30)
 #Error
 
+>> Date( 2019, 5, 16 ) > Time( 11, 23, 45 )
```
MikeStall, line 70: Can we add a comment about what does this mean?

lesaltzm: Unfortunately, we already allowed this in Canvas, and there's not a great way to remove that support.

CarlosFigueiraMSFT (approved):
:shipit:

BruceHaley:

PR #559, opened 2022-08-16
Date() returns a datetime

landonarogers:
The Date() function is returning a time as well. We would like it to truly return a Date type, not a DateTime.


landonarogers:
```
Assert.Equal() Failure
Expected: SetFieldDateFunction,1/1/2040
Actual:   SetFieldDateFunction,1/1/2040 12:00:00 AM 
```

PR #520, opened 2022-08-04
Move Attachments to use LazyType support

lesaltzm:
We have a bunch of special case recursive types in DType. This begins moving them on to the Lazy pattern.

PR #562, opened 2022-08-17
Add tests for Renamer when using globals with display names

lesaltzm:
From a comment on #503

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
@@ -200,16 +200,6 @@ namespace Microsoft.PowerFx
             return _environmentSymbols.TryGetValue(lookupName, out symbol);
         }
 
-        /// <summary>
-        /// Some scenarios require that lookups be done with logical names only.
-        /// This returns the same PowerFxConfig with the display name provider disabled.
-        /// </summary>
-        /// <returns></returns>
-        internal PowerFxConfig WithoutDisplayNames()
-        {
-            return new PowerFxConfig(this) { _environmentSymbolDisplayNameProvider = DisabledDisplayNameProvider.Instance };
-        }
-
         internal void Lock()
         {
             CheckUnlocked();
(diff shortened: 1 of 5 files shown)


src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -47,11 +47,10 @@ public IEnumerable<string> GetAllFunctionNames()
         /// Create a resolver for use in binding. This is called from <see cref="Check(string, RecordType, ParserOptions)"/>.
         /// Base classes can override this is there are additional symbols not in the config.
         /// </summary>
-        /// <param name="alternateConfig">An alternate config that can be provided. Should default to engine's config if null.</param>        
         /// <returns></returns>
-        private protected virtual INameResolver CreateResolver(PowerFxConfig alternateConfig = null)
```
MikeStall, line 52: >PowerFxConfig alternateConfig = null [](http://example.com/codeflow?start=63&length=36)

Great to remove this!
FYI, technically an internal breaking change - will require a trivia update in PAClient.

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -198,16 +198,6 @@ internal bool TryGetSymbol(DName name, out IExternalEntity symbol, out DName dis
...
 
             return _environmentSymbols.TryGetValue(lookupName, out symbol);
-        }
-
-        /// <summary>
-        /// Some scenarios require that lookups be done with logical names only.
-        /// This returns the same PowerFxConfig with the display name provider disabled.
-        /// </summary>
-        /// <returns></returns>
-        internal PowerFxConfig WithoutDisplayNames()
```
MikeStall, line 208: >WithoutDisplayNames [](http://example.com/codeflow?start=31&length=19)

yes!!!

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
lesaltzm: Yeah, will pick up with #520

PR #567, opened 2022-08-18
Collect function initial draft

anderson-joyle:
Collect function to enable cards usage
From @MikeStall: Note to others - this is an early draft. We're discussing, and will confirm semantics with GregLi.


BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
```diff
@@ -85,5 +87,17 @@ record = Marshal(item);
                 return base.TryGetIndex(index1, out record);
             }
         }
+
+        public bool Append(T record)
```
MikeStall, line 91: >bool [](http://example.com/codeflow?start=15&length=4)

This should be a virtual on TableValue (like Count, TryGetIndex). 
The default implementation can error. (for now, throwing is ok - we'll confirm the right error value later). 

Then this becomes:
            if (_sourceList == null)
            {
                return base.Append(record);
            }
            _sourceList.Add(record);

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNewTable.cs
```diff
@@ -29,7 +29,7 @@ public partial class FormulaValue
         /// <returns></returns>
         public static TableValue NewTable(RecordType recordType, params RecordValue[] values)
         {
-            return NewTable(recordType, (IEnumerable<RecordValue>)values);
+            return NewTable(recordType, values.ToList());
```
MikeStall, line 32: >values.ToList() [](http://example.com/codeflow?start=40&length=15)

This is subtle - 
if we pass in an array, we're making a copy into a mutable list.   So the return value will always be mutable. 

whereas if we call the overload on Line 35, we avoid copying, but instead querying for a mutable interface. and it's potentially not mutable. 

Not necessarily wrong, just raising awareness of the semantics. Good for now, we'll likely refine further.

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
```diff
@@ -0,0 +1,71 @@
...
+                new NamedValue("MyField1", FormulaValue.New(1)),
+                new NamedValue("MyField2", FormulaValue.New("Hello Earth!!!")));
+
+            RecordValue r2 = FormulaValue.NewRecordFromFields(
+                new NamedValue("MyField1", FormulaValue.New(2)),
+                new NamedValue("MyField2", FormulaValue.New("Hello Mars!!!")));
+
+            var list = new List<RecordValue>();
+
+            var t = (RecordsOnlyTableValue)FormulaValue.NewTable(r1.Type, list);
```
MikeStall, line 49: > (RecordsOnlyTableValue) [](http://example.com/codeflow?start=19&length=24)

we can remove the cast. (especially per other comment to move Append onto TableValue)

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            engine.UpdateVariable("list", t);
+            engine.UpdateVariable("r1", r1);
+
+            engine.Eval("Collect(list, r1)"); // [1]
+
+            t.Append(r2); // [2]
```
MikeStall, line 56: > t.Append(r2); // [2] [](http://example.com/codeflow?start=11&length=21)

we can move this line out into a separate unit test. 
(which might even in live in Core, since it's operating on TableValue directly and not needing eval?)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CollectFunction.cs
```diff
@@ -0,0 +1,362 @@
...
+    // The CollectFunction class was inspired (copied) by Set function.
+    // Implementation of a Set function which just chains to 
+    // RecalcEngine.UpdateVariable().
+    // Set has no return value. 
+    // Whereas PowerApps' Set() will implicitly define arg0,
+    //  this Set() requires arg0 was already defined and has a type.
+    //
+    // Called as:
+    //   Set(var,newValue)
+    internal class CollectFunction : BuiltinFunction, IAsyncTexlFunction
```
MikeStall, line 32: >CollectFunction [](http://example.com/codeflow?start=19&length=15)

this is copied from PowerApps?

It would be good to run under code-coverage and see:
- if there are sections here we can remove 
- new tests we should add around collect?

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
```diff
@@ -0,0 +1,71 @@
...
+
+namespace Microsoft.PowerFx.Interpreter.Tests
+{
+    public class CollectFunctionTests : PowerFxTest
+    {
+        private readonly ParserOptions _opts = new ParserOptions { AllowsSideEffects = true };
+        private static readonly TypeMarshallerCache _cache = new TypeMarshallerCache();
+
+        [Fact]
+        public void AppendRecord()
```
MikeStall, line 31: >AppendRecord [](http://example.com/codeflow?start=20&length=12)

This is a great test.
It's convenient to have some C# tests.  We'll also want to see if we can get a .txt test for it (will require some #444 
That will make it easier to more thoroughly test; as well as communicate the semantics.

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
```diff
@@ -21,6 +21,7 @@ public abstract class CollectionTableValue<T> : TableValue
         // Additional capabilities. 
         private readonly IReadOnlyList<T> _sourceIndex; // maybe null. supports index. 
         private readonly IReadOnlyCollection<T> _sourceCount; // maybe null. supports count;
+        private readonly List<T> _sourceList;
```
LucGenetier, line 24: >List [](http://example.com/codeflow?start=25&length=4)

IList<T>

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
```diff
@@ -35,6 +36,7 @@ internal CollectionTableValue(IRContext irContext, IEnumerable<T> source)
 
             _sourceIndex = source as IReadOnlyList<T>;
             _sourceCount = source as IReadOnlyCollection<T>;
+            _sourceList = source as List<T>;
```
LucGenetier, line 39: >List [](http://example.com/codeflow?start=36&length=4)

IList<T>

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
```diff
@@ -85,5 +87,17 @@ record = Marshal(item);
                 return base.TryGetIndex(index1, out record);
             }
         }
+
+        public bool Append(T record)
+        {
+            if (_sourceList == null)
+            {
+                throw new NotImplementedException();
```
LucGenetier, line 95: >NotImplementedException(); [](http://example.com/codeflow?start=26&length=26)

Add a message when throwing any exception

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
LucGenetier: Shouldn't this API be named TryAppend(...) instead?
If it's Append, it shouldn't return a bool, but void.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CollectFunction.cs
```diff
@@ -0,0 +1,362 @@
...
+            //            DName name = typedName.Name;
+            //            if (!dataSourceType.TryGetType(name, out DType dsNameType))
+            //                dataSourceType.ReportNonExistingName(FieldNameKind.Display, errors, name, args[i], DocumentErrorSeverity.Warning);
+            //        }
+            //    }
+            //}
+
+            // TASK: 75145: SPEC: what if the types align for arg0, but arg0 is not a name node? For example:
+            //      Collect( Filter(T,A<2), {A:10} )
+            // The current behavior is that Collect has no side effects for transient tables/collections.
```
LucGenetier, line 196: To be removed

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
MikeStall: The challenge is that we want Append (), and IList doesn't have that, only List. 
We may need to find yet another interface.

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
MikeStall: +1 - if we do end up with bool, it should be TryAppend. 
Some of the PowerFx functions (like Set) always just return 'true'; but even in that case, the C# signature can be void, and the power fx function impl can handle the return. 

Although, sorting out exact semantics with Greg - leaning towards returning a FormulaValue for the newly added record (with any server-side fields filled out); or an error value (say the server rejected the add)

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
lesaltzm: IList has Add(), which is what it looks like you're using.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CollectFunction.cs
```diff
@@ -0,0 +1,362 @@
...
+
+        public virtual DType GetCollectedType(DType argType)
+        {
+            Contracts.Assert(argType.IsValid);
+
+            return argType;
+        }
+
+        // Attempt to get the unified schema of the items being collected by an invocation.
+        public bool TryGetUnifiedCollectedType(TexlNode[] args, DType[] argTypes, IErrorContainer errors, out DType collectedType)
```
lesaltzm, line 99: We don't want to do implicit collection schema definition for this use, likely all this type checking is irrelevant.

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
anderson-joyle: Done.

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
anderson-joyle: Changed to ICollection<T>.

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
anderson-joyle: ICollection <T>

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CollectFunction.cs
anderson-joyle: It was copied from AP. The not covered code is currently at 47.23%. More test cases will be added as needed.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CollectFunction.cs
```diff
@@ -0,0 +1,362 @@
...
+    //   Set(var,newValue)
+    internal class CollectFunction : BuiltinFunction, IAsyncTexlFunction
+    {
+        public override bool SuppressIntellisenseForComponent => true;
+
+        public override bool ManipulatesCollections => true;
+
+        public override bool ModifiesValues => true;
+
+        public override bool AffectsCollectionSchemas => true;
```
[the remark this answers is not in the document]
anderson-joyle: Done.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CollectFunction.cs
```diff
@@ -0,0 +1,362 @@
...
+        //                Contracts.Assert(fieldName == Contracts.VerifyValue(CollectScalarFunction.GetInvariantNameForRecord(binding.GetType(args[i]).Kind)));
+        //#endif
+
+        //            var builder = new PAStringBuilder(fieldName.Length + 2);
+        //            builder.AppendAsPlainText(fieldName);
+
+        //            argFragments.Add(translator.CreateFragment(builder));
+        //        }
+
+        protected static string CreateInvariantFieldName(DKind dKind)
```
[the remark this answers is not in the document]
anderson-joyle: Apparently was used to determine field name to a tabular table depending on the DType. At this point it wasn't being called so it has been removed for now.

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
anderson-joyle: Agreed. I'm leaving as it currently is since there is still a discussion about what the function should return.

BruceHaley:

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
MikeStall: This should now be:
    _sourceList = source as ICollection<T>;

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CollectFunction.cs
```diff
@@ -0,0 +1,267 @@
...
+
+        // This method returns true if there are special suggestions for a particular parameter of the function.
+        public override bool HasSuggestionsForParam(int argumentIndex)
+        {
+            Contracts.Assert(argumentIndex >= 0);
+
+            return argumentIndex == 0;
+        }
+
+        public override bool TryGetDataSourceNodes(CallNode callNode, TexlBinding binding, out IList<FirstNameNode> dsNodes)
```
MikeStall, line 203: >TryGetDataSourceNodes [](http://example.com/codeflow?start=29&length=21)

You can remove this (and GetIdentifierOfModifiedValue) 
They're not used here. 
(just used in more complex powerapps scenarios)

src/tests/Microsoft.PowerFx.Core.Tests/CollectFunctionTests.cs
```diff
@@ -0,0 +1,37 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.PowerFx.Types;
+using Xunit;
+
+namespace Microsoft.PowerFx.Core.Tests
+{
+    public class CollectFunctionTests : PowerFxTest
```
MikeStall, line 13: >CollectFunctionTests [](http://example.com/codeflow?start=17&length=20)

rename:
  TableValueMutation tests 

ie, this is in Core and going against TableValue directly, bypassing the Collect() function 

This is nice to have the additional coverage in Core.

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
```diff
@@ -0,0 +1,58 @@
...
+            config.AddFunction(new CollectFunction());
+            config.EnableSetFunction();
+
+            var engine = new RecalcEngine(config);
+
+            RecordValue r1 = FormulaValue.NewRecordFromFields(
+                new NamedValue("MyField1", FormulaValue.New(1)),
+                new NamedValue("MyField2", FormulaValue.New("Hello World!!!")));
+
+            var t = (RecordsOnlyTableValue)FormulaValue.NewTable(r1.Type, new List<RecordValue>());
```
MikeStall, line 46: >(RecordsOnlyTableValue [](http://example.com/codeflow?start=20&length=22)

can we remove the cast?

In fact, i think this could just be:
     var t = FormulaValue.NewTable(r1.Type);


And it will bind to the mutable overload of NewTable.

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        [Theory]
+        [InlineData("Collect(t, {MyField1:2, MyField2:\"hello1\"});CountRows(t)", 1)]
+        [InlineData("Collect(t, r1);Collect(t, r1);Collect(t, r1);CountRows(t)", 3)]
+        [InlineData("Collect(t, r1);Collect(t, If(1>0, r1,Blank()));CountRows(t)", 2)]
+        public void AppendCountTest(string script, int expected)
+        {
+            var config = new PowerFxConfig();
+
+            config.AddFunction(new CollectFunction());
```
MikeStall, line 37: >    config.AddFunction(new CollectFunction()); [](http://example.com/codeflow?start=8&length=46)

TexlFunction (and CollectFunction) are internal classes.

we'll probably need to wrap this in a public  .EnableCollect() function call, similar to what we do with Set.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CollectFunction.cs
```diff
@@ -0,0 +1,267 @@
...
+
+        public override bool IsAsyncInvocation(CallNode callNode, TexlBinding binding)
+        {
+            Contracts.AssertValue(callNode);
+            Contracts.AssertValue(binding);
+
+            return Arg0RequiresAsync(callNode, binding);
+        }
+
+        public Task<FormulaValue> InvokeAsync(FormulaValue[] args, CancellationToken cancel)
```
MikeStall, line 254: >public Task<FormulaValue> [](http://example.com/codeflow?start=8&length=25)

use async keyword.  The consistency is important here, especially because there are subtle diferences with exception handling. 

That will also simplify line 264.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CollectFunction.cs
```diff
@@ -0,0 +1,267 @@
...
+        {
+            var arg0 = (RecordsOnlyTableValue)args[0];
```
MikeStall, line 256: >(RecordsOnlyTableValue) [](http://example.com/codeflow?start=23&length=23)

Just (TableValue)

Can't assume that it's anything more derived.

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
```diff
@@ -78,6 +79,16 @@ private static ErrorValue ArgumentOutOfRange(IRContext irContext)
             });
         }
 
+        // Return appended value 
+        // - Error, 
+        // - with updated values
+        // Async because derived classes may back this with a network call. 
+        public virtual async Task<DValue<RecordValue>> AppendAsync(RecordValue record)
+        {
+            // fails by default
```
MikeStall, line 88: >// fails by default [](http://example.com/codeflow?start=12&length=19)

Will need to confirm the exact error here. Will probably need to be a runtime ErrorValue (DValue allows that) rather than an exception

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
```diff
@@ -0,0 +1,58 @@
...
+using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Interpreter;
+using Microsoft.PowerFx.Syntax;
+using Microsoft.PowerFx.Types;
+using Xunit;
+using static Microsoft.PowerFx.Core.Localization.TexlStrings;
+
+namespace Microsoft.PowerFx.Interpreter.Tests
+{
+    public class CollectFunctionTests : PowerFxTest
```
MikeStall, line 25: >CollectFunctionTests [](http://example.com/codeflow?start=17&length=20)

See Set tests for some inspiration. 

More test ideas:
- verify that Collect() fails in a non-behavior context. 

- Type checking - what if we pass the wrong type to collect (too few fields, too many fields, not a record, etc). 

- PowerApps lets pass multiple records - this is determined by CheckInvocation, which we copied from PA.  We should either 
a) reject that as an error, 
b) support it. 
Collect(_c, {r:1}, {r:2})

- will need to confirm corner case behavior, ie, what if we pass errors, etc.  (We can probably do that in a followup PR, just to unblock usage now) 



- trying to collect on an Immutable object. We should get an error

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
anderson-joyle: EnableCollectFunction extension method created

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CollectFunction.cs
anderson-joyle: Removed.

src/tests/Microsoft.PowerFx.Core.Tests/CollectFunctionTests.cs
anderson-joyle: Renamed.

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
anderson-joyle: Done.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CollectFunction.cs
anderson-joyle: Done.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CollectFunction.cs
anderson-joyle: Done.

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
anderson-joyle: Got it. Added this to my personal notes as reminder.

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
lesaltzm: Also, type checking with marshalled types on both sides.

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
anderson-joyle: Done.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CollectFunction.cs
anderson-joyle: Thanks for pointing that. By removing this method, the following also becomes irrelevant and can also be delete (from CheckInvocation method).
```
            fValid &= TryGetUnifiedCollectedType(args, argTypes, errors, out DType collectedType);
            Contracts.Assert(collectedType.IsTable);

            // The item type must be compatible with the collection schema.
            var fError = false;
            returnType = DType.Union(ref fError, collectionType, collectedType, useLegacyDateTimeAccepts: true);
            if (fError)
            {
                fValid = false;
                if (!SetErrorForMismatchedColumns(collectionType, collectedType, args[1], errors))
                {
                    errors.EnsureError(DocumentErrorSeverity.Severe, args[0], TexlStrings.ErrNeedValidVariableName_Arg);
                }
            }
```

Make sense?

BruceHaley:

BruceHaley:

BruceHaley:

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CollectFunction.cs
```diff
@@ -0,0 +1,224 @@
...
+        {
+            var arg0 = (TableValue)args[0];
+            var arg1 = (RecordValue)args[1];
+
+            var mytype0 = arg0.GetType();
+            var mytype1 = arg1.GetType();
+
+            var result = arg0.AppendAsync(arg1);
+
+            return await Task.FromResult<FormulaValue>(FormulaValue.New(true));
```
MikeStall, line 221: >Task.FromResult<FormulaValue> [](http://example.com/codeflow?start=25&length=29)

With await, This can just be:
return await FormulaValue.New(true);

BruceHaley:

BruceHaley:

src/libraries/Microsoft.PowerFx.Interpreter/Environment/PowerFxConfigExtensions.cs
```diff
@@ -30,5 +30,14 @@ public static void EnableSetFunction(this PowerFxConfig powerFxConfig)
         {
             powerFxConfig.AddFunction(new RecalcEngineSetFunction());
         }
+
+        /// <summary>
+        /// Enable a Collect() function which allows scripts to append table records.
+        /// </summary>
+        /// <param name="powerFxConfig"></param>
+        public static void EnableCollectFunction(this PowerFxConfig powerFxConfig)
```
MikeStall, line 38: >this PowerFxConfig powerFxConfig [](http://example.com/codeflow?start=49&length=32)

Update: add to this SymbolTable symbolTable, instead

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        private readonly ParserOptions _opts = new ParserOptions { AllowsSideEffects = true };
+
+        [Theory]
+        [InlineData("Collect(t, {MyField1:2, MyField2:\"hello1\"});CountRows(t)", 1)]
+        [InlineData("Collect(t, r1);Collect(t, r1);Collect(t, r1);CountRows(t)", 3)]
+        [InlineData("Collect(t, r1);Collect(t, If(1>0, r1,Blank()));CountRows(t)", 2)]
```
MikeStall, line 32: >Collect(t, r1) [](http://example.com/codeflow?start=21&length=14)

New Test for return value 

Collect(t, r1)

src/libraries/Microsoft.PowerFx.Interpreter/Environment/PowerFxConfigExtensions.cs
anderson-joyle: Done.

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
anderson-joyle: Done.

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
MikeStall: fyi

src/tests/Microsoft.PowerFx.Core.Tests/TableValueMutationTest.cs
```diff
@@ -0,0 +1,37 @@
...
+
+            Assert.Equal(1, t.Count());
+            await t.AppendAsync(r2); // succeeds
+
+            Assert.Equal(2, t.Count());
+
+            // Future test case
+            //IEnumerable<RecordValue> source = new RecordValue[] { r1 };
+            //var t2 = FormulaValue.NewTable(r1.Type, source); // Immutabe
+            //await t2.AppendAsync(r2); // Fails
```
MikeStall, line 34: Let's just go ahead and add this - as a separate test case ("AppendToReadOnlyTableFails")

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
MikeStall: fyi

BruceHaley:

src/tests/Microsoft.PowerFx.Core.Tests/TableValueMutationTest.cs
anderson-joyle: Done.

BruceHaley:

src/tests/Microsoft.PowerFx.Core.Tests/TableValueMutationTest.cs
```diff
@@ -0,0 +1,37 @@
...
+            var t = FormulaValue.NewTable(r1.Type, r1); // Mutable 
+
+            Assert.Equal(1, t.Count());
+            await t.AppendAsync(r2); // succeeds
+
+            Assert.Equal(2, t.Count());
+
+            // Future test case
+            IEnumerable<RecordValue> source = new RecordValue[] { r1 };
+            var t2 = FormulaValue.NewTable(r1.Type, source); // Immutabe
```
MikeStall, line 33: >Immutabe [](http://example.com/codeflow?start=64&length=8)

typo

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
```diff
@@ -65,6 +73,20 @@ public override int Count()
             }
         }
 
+        public override async Task<DValue<RecordValue>> AppendAsync(RecordValue record)
+        {
+            if (_sourceList == null)
```
MikeStall, line 78: We can also check ICollection<T>.IsReadOnly 

[https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.icollection-1.isreadonly?view=net-6.0](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.icollection-1.isreadonly?view=net-6.0)

If it's readonly, we should just call base class (and let it fail)

       if (_sourceList == null || _sourceList.IsReadOnly)

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
anderson-joyle: After some changes:

> verify that Collect() fails in a non-behavior context.

Throw an exception


> Type checking - what if we pass the wrong type to collect (too few fields, too many fields, not a record, etc).

All these cases will return an exception

> - PowerApps lets pass multiple records - this is determined by CheckInvocation, which we copied from PA. We should either
> 
> a) reject that as an error,
> b) support it.
> Collect(_c, {r:1}, {r:2})

For now, limiting argument numbers to two: table and record


> will need to confirm corner case behavior, ie, what if we pass errors, etc. (We can probably do that in a followup PR, just to unblock usage now)

I'll wait further interactions on this.

> trying to collect on an Immutable object. We should get an error

This test case has been added.

src/tests/Microsoft.PowerFx.Core.Tests/TableValueMutationTest.cs
anderson-joyle: Fixed.

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
anderson-joyle: Done.

BruceHaley:

src/tests/Microsoft.PowerFx.Core.Tests/TableValueMutationTest.cs
anderson-joyle: Fixed.

BruceHaley:

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CollectFunction.cs
```diff
@@ -0,0 +1,232 @@
...
+    //
+    // Called as:
+    //   Set(var,newValue)
+    internal class CollectFunction : BuiltinFunction, IAsyncTexlFunction
+    {
+        public override bool ManipulatesCollections => true;
+
+        public override bool ModifiesValues => true;
+
+        public override bool IsSelfContained => false;
```
MikeStall, line 38: >public override bool IsSelfContained => false; [](http://example.com/codeflow?start=8&length=46)

why did this flag from first iteration?

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
```diff
@@ -0,0 +1,149 @@
...
+
+            var t = FormulaValue.NewTable(r1.Type, listT);
+
+            symbol.AddConstant("t", t);
+            symbol.AddConstant("r1", r1);
+
+            var engine = new RecalcEngine();
+
+            // options: _opts has been removed
+            await Assert.ThrowsAsync<InvalidOperationException>(async () => await engine.EvalAsync("Collect(t, r1)", CancellationToken.None, symbolTable: symbol)); // Fails
```
MikeStall, line 146: >EvalAsync [](http://example.com/codeflow?start=89&length=9)

we should actually get the error at design time (calling Check).

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
```diff
@@ -0,0 +1,149 @@
...
+                new NamedValue("MyField1", FormulaValue.New(1)),
+                new NamedValue("MyField2", FormulaValue.New("Hello World!!!")));
+
+            var t = FormulaValue.NewTable(r1.Type, listT);
+
+            symbol.AddConstant("t", t);
+            symbol.AddConstant("r1", r1);
+
+            var engine = new RecalcEngine();
+            await Assert.ThrowsAsync<InvalidOperationException>(async () => await engine.EvalAsync(script, CancellationToken.None, options: _opts, symbolTable: symbol)); // Fails
```
MikeStall, line 123: >EvalAsync [](http://example.com/codeflow?start=89&length=9)

also error at bind time, Check

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
```diff
@@ -0,0 +1,149 @@
...
```
MikeStall, line 101: >EvalAsync [](http://example.com/codeflow?start=89&length=9)

error at check

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CollectFunction.cs
anderson-joyle: From what I can tell, this is the flag thats controls if the function can run only in a side effect context. It is actually a different flag that validates this, but it depends directly on the IsSelfContained flag.

> public virtual bool IsBehaviorOnly => !IsSelfContained;

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
anderson-joyle: Done.

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
anderson-joyle: Done.

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
anderson-joyle: Done.

BruceHaley:

BruceHaley:

MikeStall (approved):
:shipit:

MikeStall:
Approving under a flag to unblock cards. Will continue to followup on surface area here and make some breaking changes.

BruceHaley:

BruceHaley:

PR #571, opened 2022-08-19
Added method to get field based on logical/ display name. (Fixes #556)

### LINKED ISSUE #556: DisplayNames: Add method to record type for looking up display name.
Today, RecordType has a [GetField(name)](https://github.com/microsoft/Power-Fx/blob/cd901242244d31c83b431bbaef278bd40c40e153/src/libraries/Microsoft.PowerFx.Core/Public/Types/AggregateType.cs#L30)  for logical names - see AggregateType.cs.

But we want a lookup for **display names** as well. This is used by the SDK. 
(Note that the IR should embed a logical name). 

Add RecordType: 

   bool TryGetFieldType(string displayOrLogical, out logical, out FormulaType type);

Accept either display or logical name. 
If it is a display name, it should return the logical name that it corresponds to. 
(If it is a logical name, just return itself) 

Also See #549 for another issue about display names. 

Things to consider:
- This may need to be virtual, like the existing GetField()
- adding method and a test.

jas-valgotar:
- Fixes #556 
- Added method "TryGetFieldType" to get field based on logical or display name!
- Added tests for the same in "TryGetFieldTypeLookupTest" and "TryGetFieldTypeLookupNullTest".


BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Public/Types/AggregateType.cs
```diff
@@ -46,6 +46,39 @@ public virtual bool TryGetFieldType(string name, out FormulaType type)
...
 
+        /// <summary>
+        /// Takes display or logical name as input and returns field and logical name if exists.
+        /// </summary>
+        /// <param name="displayOrLogicalName">Display or Logical name.</param>
+        /// <param name="logical">Logical name for the input.</param>
+        /// <param name="type">Type for the input Display or Logical name.</param>
+        /// <returns>true or false.</returns>
+        /// <exception cref="ArgumentNullException">Throws, if input displayOrLogicalName is empty.</exception>
+        public virtual bool TryGetFieldType(string displayOrLogicalName, out string logical, out FormulaType type)
```
lesaltzm, line 57: IMO, this shouldn't be virtual. It gets the logical name, and then calls `TryGetFieldType`, which is virtual. Not sure why we'd want to allow overriding this.

src/libraries/Microsoft.PowerFx.Core/Public/Types/AggregateType.cs
```diff
@@ -46,6 +46,39 @@ public virtual bool TryGetFieldType(string name, out FormulaType type)
...
+        {
+            if (string.IsNullOrEmpty(displayOrLogicalName))
+            {
+                throw new ArgumentNullException("Input parameter \"displayOrLogicalName\" cannot be empty or null");
```
MikeStall, line 61: >ArgumentNullException [](http://example.com/codeflow?start=26&length=21)

nit:
There's a nameof() operator that returns the string name for a C# identifier. 
And ArgException ctor will provide a message. 

So you can just say:
new ArgumentNullException(nameof(displayOrLogicalName))

src/libraries/Microsoft.PowerFx.Core/Public/Types/AggregateType.cs
MikeStall: +1 
The case to consider here is a derived RecordType that overrides FieldNames. 

TryGetFieldType is already virtual. 
And the display-->logical mapping is via the DisplayNameProvider, which also has virtuals. 

The piece that's missing though is how would the derived RecordType provide it sown DisplayNameProvider? 

Maybe we need to allow that as part of deriving.

src/tests/Microsoft.PowerFx.Core.Tests/TypeTests.cs
```diff
@@ -36,5 +37,34 @@ public void RecordTypeValidation()
 
             Assert.Equal(r1.GetHashCode(), r2.GetHashCode());
         }
+
+        [Theory]
+        [InlineData("Display1", true, "F1")]
+        [InlineData("F1", true, "F1")]
+        [InlineData("SomethingElse", false, null)]
+        public void TryGetFieldTypeLookupTest(string inputDisplayOrLogical, bool succeeds, string expectedLogical)
```
MikeStall, line 45: >TryGetFieldTypeLookupTest [](http://example.com/codeflow?start=20&length=25)

One more case.... conflicts!
What a field's logical name conflicts with another fields display name. 

Logical=F1 , Display=Display1
Logical=Display1, Display=Display2

jas-valgotar: Current Implementation prioritizes logical name.
Can't add the test case for the same coz, singleNameDisplayProvider handles the collision and would not let us add such records.

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Public/Types/AggregateType.cs
jas-valgotar: PR for how would the derived RecordType provide it sown DisplayNameProvider?
https://github.com/microsoft/Power-Fx/pull/593

MikeStall (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Types/AggregateType.cs
```diff
@@ -46,6 +46,48 @@ public virtual bool TryGetFieldType(string name, out FormulaType type)
...
+        /// <param name="logical">Logical name for the input.</param>
+        /// <param name="type">Type for the input Display or Logical name.</param>
+        /// <returns>true or false.</returns>
+        /// <exception cref="ArgumentNullException">Throws, if input displayOrLogicalName is empty.</exception>
+        public bool TryGetFieldType(string displayOrLogicalName, out string logical, out FormulaType type)
+        {
+            if (string.IsNullOrEmpty(displayOrLogicalName))
+            {
+                throw new ArgumentNullException(nameof(displayOrLogicalName));
+            }
```
lesaltzm, line 65: nit: we have a utility for this: Contracts.CheckNonEmpty(displayOrLogicalName)

The Contracts.Assert* functions are debug only, but Contracts.Check* stay in.

src/libraries/Microsoft.PowerFx.Core/Public/Types/AggregateType.cs
```diff
@@ -46,6 +46,48 @@ public virtual bool TryGetFieldType(string name, out FormulaType type)
...
+                logical = maybeLogical;
+            }
+            else
+            {
+                logical = null;
+                type = Blank;
+                return false;
+            }
+
+            if (!TryGetFieldType(logical, out type))
```
lesaltzm, line 82: This block could just be `return TryGetFieldType(logical, out type)`. Not necessary to clear logical before returning, if it's not found in the type, the derived class is violating a contract by having it in the DisplayNameProvider. You could even use the more strict `GetFieldType` if you wanted to verify that.

src/libraries/Microsoft.PowerFx.Core/Public/Types/AggregateType.cs
jas-valgotar: Thanks! I'll keep this in mind! updated the current code.

src/libraries/Microsoft.PowerFx.Core/Public/Types/AggregateType.cs
jas-valgotar: updated with return TryGetFieldType(logical, out type)

BruceHaley:

lesaltzm (approved):

BruceHaley:

PR #576, opened 2022-08-20
Update error for invalid language code + expected value for date tests

CarlosFigueiraMSFT:
The expected error for when an invalid language code is passed to functions should be BadLanguageCode. This also updates expected results in DateValue tests to not rely on being run on en-US locale.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs
@@ -69,6 +69,16 @@ namespace Microsoft.PowerFx.Functions
             });
         }
 
+        public static ErrorValue BadLanguageCode(IRContext irContext, string languageCode)
+        {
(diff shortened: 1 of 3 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs
```diff
@@ -69,6 +69,16 @@ public static ErrorValue OverflowError(IRContext irContext)
             });
         }
 
+        public static ErrorValue BadLanguageCode(IRContext irContext, string languageCode)
+        {
+            return new ErrorValue(irContext, new ExpressionError()
+            {
+                Message = $"Language code {languageCode} not supported",
```
LucGenetier, line 76: As this will happen when the language culture is invalid, i'd name this error BadCulture and fix the error message accordingly ("Invalid Culture {...

LucGenetier (approved):
:shipit:

MikeStall (approved):

PR #584, opened 2022-08-23
Remove navigation columns from query options selects

kartikkr11:


src/libraries/Microsoft.PowerFx.Core/Entities/QueryOptions/TabularDataQueryOptions.cs
```diff
@@ -96,6 +96,12 @@ public void AddRelatedColumns()
                     // Add the Annotated value in case a navigation field is referred in selects. (ex: if the Datasource is Accounts and primarycontactid is in selects also append _primarycontactid_value)
                     _selects.Add(additionalColumnName);
                 }
+
+                if (cdsDataSourceInfo.CanRemoveNavigationColumn(select))
```
lesaltzm, line 100: Is there a reason this needs to be implemented here instead of either in the code generator or runtime?

lesaltzm (changes requested):
IMO, this is a change that is external to PowerFx, and is better suited to code generation for the host-specific QueryOption requirements.

src/libraries/Microsoft.PowerFx.Core/Entities/QueryOptions/TabularDataQueryOptions.cs
kartikkr11: handling of selects in the expands would also need to be specially handled if we change it in the codegen layer right?

kartikkr11: Also the navigation selects are not used in canvas apps too so this kinda cleans up both call for canvas apps and custom pages, from my understanding it was always present without really being required.

PR #595, opened 2022-08-25
Patch function

anderson-joyle:
This is work in progress but I want to get some feedback from reviewers.
Some of my questions/concerns/comments:

- Two Patch implementations for now:

> Patch(record, record)
Not the focus although it is going to be implemented anytime soon.

> Patch(dataSource, record, arg1, arg2, ...)
The priority one.

- In order to update a record, I'm simply removing the old record (ICollection<T>.Remove(T)) and adding a new version of it. Is there a better/prettier way to do this?
- I'm still working on the CheckInvocation implementations.
- I'm still working on how to handle errors.
- A new handler has been added for test units. Is this the way or is there a better approach?
- Any other major concerns?


src/libraries/Microsoft.PowerFx.Interpreter/Environment/PowerFxConfigExtensions.cs
```diff
@@ -39,5 +39,23 @@ public static void EnableCollectFunction(this SymbolTable symbolTable)
         {
             symbolTable.AddFunction(new CollectFunction());
         }
+
+        /// <summary>
+        /// Enable a Patch() function which allows scripts to append table records.
+        /// </summary>
+        /// <param name="symbolTable"></param>
+        public static void EnablePatchRecordFunction(this SymbolTable symbolTable)
```
MikeStall, line 47: >EnablePatchRecordFunction [](http://example.com/codeflow?start=27&length=25)

nit: we can probably include this w/ Collect.
And just generalize to: 
EnableMutationFunctions()

Clients will use these altogether.

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
```diff
@@ -0,0 +1,204 @@
...
+                var namedValue = argFields.ContainsKey(field.Name) ? argFields[field.Name] : field.Value;
+
+                fieldList.Add(new NamedValue(field.Name, namedValue));
+            }
+
+            var newRecord = FormulaValue.NewRecordFromFields(fieldList);
+
+            // @@@JYL Whats a good way to replace the old record with the new one and "update" the TableValue object?
+
+            var deleted = await arg0.RemoveAsync(arg1);
```
MikeStall, line 189: This should probably be a "UpdateField" virtual on RecordValue. 

Not a remove+append.

src/libraries/Microsoft.PowerFx.Core/Localization/Strings.cs
```diff
@@ -448,6 +448,7 @@ internal static class TexlStrings
         public static StringGetter AboutIndex = (b) => StringResources.Get("AboutIndex", b);
         public static StringGetter IndexArg1 = (b) => StringResources.Get("IndexArg1", b);
         public static StringGetter IndexArg2 = (b) => StringResources.Get("IndexArg2", b);
+        public static StringGetter AboutPatch = (b) => StringResources.Get("AboutPatch", b);
```
CarlosFigueiraMSFT, line 451: >AboutPatch [](http://example.com/codeflow?start=76&length=10)

We need the resources for the function parameters as well (like IndexArg1, IndexArg2 above)

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
```diff
@@ -89,6 +89,14 @@ public virtual async Task<DValue<RecordValue>> AppendAsync(RecordValue record)
             throw new System.NotImplementedException("It is not possible to append to a TableValue directly.");
         }
 
+        // Return boolean value
+        // Async because derived classes may back this with a network call. 
+        public virtual async Task<DValue<BooleanValue>> RemoveAsync(RecordValue record)
+        {
+            // fails by default
+            throw new System.NotImplementedException("It is not possible to append to a TableValue directly.");
```
CarlosFigueiraMSFT, line 97: >append to a TableValue directly [](http://example.com/codeflow?start=76&length=31)

update a TableValue directly

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
```diff
@@ -89,6 +89,14 @@ public virtual async Task<DValue<RecordValue>> AppendAsync(RecordValue record)
             throw new System.NotImplementedException("It is not possible to append to a TableValue directly.");
         }
 
+        // Return boolean value
+        // Async because derived classes may back this with a network call. 
+        public virtual async Task<DValue<BooleanValue>> RemoveAsync(RecordValue record)
```
CarlosFigueiraMSFT, line 94: >RemoveAsync [](http://example.com/codeflow?start=56&length=11)

Why remove if we are implementing an update?

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
```diff
@@ -0,0 +1,204 @@
...
+        public override bool IsSelfContained => false;
+
+        public PatchRecordFunction()
+            : base("Patch", TexlStrings.AboutPatch, FunctionCategories.Table | FunctionCategories.Behavior, DType.EmptyRecord, 0, 2, int.MaxValue, DType.EmptyRecord, DType.EmptyRecord)
+        {
+        }
+
+        public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
+        {
+            yield return new[] { TexlStrings.AboutPatch, TexlStrings.AboutPatch, TexlStrings.AboutPatch };
```
CarlosFigueiraMSFT, line 76: >AboutPatch [](http://example.com/codeflow?start=45&length=10)

This should be the arguments, not the about for the function

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
```diff
@@ -0,0 +1,204 @@
...
+        {
+            if (arity > 3)
+            {
+                return GetGenericSignatures(arity, TexlStrings.AboutPatch, TexlStrings.AboutPatch, TexlStrings.AboutPatch);
+            }
+
+            return base.GetSignatures(arity);
+        }
+
+        public Task<FormulaValue> InvokeAsync(FormulaValue[] args, CancellationToken cancel)
```
CarlosFigueiraMSFT, line 90: >InvokeAsync [](http://example.com/codeflow?start=34&length=11)

Most functions have their definitions split from their implementation in the interpreter. Any reason why we are not doing this with Patch?

src/libraries/Microsoft.PowerFx.Core/Localization/Strings.cs
anderson-joyle: Is there a good way to test these prior to push?

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
anderson-joyle: Resolved to both PatchRecordFunction and PatchFunction functions. Let me know a way to test these locally is any.

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
anderson-joyle: For this first phase we a looking for a quick way to get other team members unblocked. This will be addressed soon.

MikeStall: Yes - another reason is that these are all enabled under a flag, and not built in directly to the interpreter yet.

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
MikeStall, line 97: >append [](http://example.com/codeflow?start=76&length=6)

wording update

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
MikeStall: Patch shouldn't call Remove.  But ok to add this for the upcoming Remove function

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
MikeStall: another reason is that Remove/append will change the order of the collection.

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.2-preview.20220825-318698](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.2-preview.20220825-318698).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfigExtensions.EnableCollectFunction(Microsoft.PowerFx.SymbolTable)' does not exist in the implementation but it does exist in the contract.
```

</details>

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.2-preview.20220825-318698](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.2-preview.20220825-318698).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfigExtensions.EnableCollectFunction(Microsoft.PowerFx.SymbolTable)' does not exist in the implementation but it does exist in the contract.
```

</details>

anderson-joyle:
@MikeStall about this [binary compatibility issue](https://github.com/microsoft/Power-Fx/pull/595#issuecomment-1228922040), this is happening because Microsoft.PowerFx.PowerFxConfigExtensions.EnableCollectFunction was previously merged into main and now it has been removed due to the new EnableMutationFunctions method. Should I bring back the EnableCollectFunction method just for compatibility purposes?

MikeStall:
This breaking change is ok. We just introduced the function, PowerApps doesn't depend on it yet.  

---
In reply to: [1228934844](https://github.com/microsoft/Power-Fx/pull/595#issuecomment-1228934844) [](http://example.com/codeflow?ancestors=1228934844,1228934844)

src/libraries/Microsoft.PowerFx.Core/Public/Values/InMemoryRecordValue.cs
```diff
@@ -41,5 +43,18 @@ protected override bool TryGetField(FormulaType fieldType, string fieldName, out
...
         }
+
+        public override void UpdateField(NamedValue newNamedValue)
+        {
+            if (newNamedValue == null || _sourceDict.IsReadOnly)
+            {
+                base.UpdateField(newNamedValue);
+            }
+
+            if (_sourceDict.ContainsKey(newNamedValue.Name))
```
MikeStall, line 54: > if (_sourceDict.ContainsKey(newNamedValue.Name)) [](http://example.com/codeflow?start=11&length=49)

Why do we do this check?

We should be able to set blank values to non-blank values.

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
```diff
@@ -161,5 +161,10 @@ public sealed override void Visit(IValueVisitor visitor)
         {
             visitor.Visit(this);
         }
+
+        public virtual void UpdateField(NamedValue newNamedValue)
```
MikeStall, line 165: >UpdateField(NamedValue newNamedValue) [](http://example.com/codeflow?start=28&length=37)

We should also have an overload:
  Update(string name, FormulaValue newValue)

Note that only 1 overload should be virtual, the others should chain.

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -18,7 +18,8 @@ public class ExpressionEvaluationTests : PowerFxTest
     {
         internal static Dictionary<string, Func<PowerFxConfig, (RecalcEngine engine, RecordValue parameters)>> SetupHandlers = new ()
         {
-            { "OptionSetTestSetup", OptionSetTestSetup }
+            { "OptionSetTestSetup", OptionSetTestSetup },
+            { "PatchTestSetup", PatchTestSetup }
```
MikeStall, line 22: >PatchTestSetup [](http://example.com/codeflow?start=15&length=14)

Can we reuse this to for .txt files for Collect ?

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.2-preview.20220826-318850](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.2-preview.20220826-318850).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfigExtensions.EnableCollectFunction(Microsoft.PowerFx.SymbolTable)' does not exist in the implementation but it does exist in the contract.
```

</details>

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
```diff
@@ -0,0 +1,187 @@
...
+            Contracts.Assert(args.Length == argTypes.Length);
+            Contracts.AssertValue(errors);
+
+            var isValid = CheckInvocation(args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
+
+            return isValid;
+        }
+    }
+
+    internal class PatchRecordFunction : PatchAndValidateRecordFunctionBase, IAsyncTexlFunction
```
MikeStall, line 65: >PatchRecordFunction [](http://example.com/codeflow?start=19&length=19)

Can you add a comment for which Patch() overload this is.

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
```diff
@@ -0,0 +1,187 @@
...
+        {
+            if (arity > 2)
+            {
+                return GetGenericSignatures(arity, TexlStrings.PatchBaseRecordArg, TexlStrings.PatchChangeRecordsArg);
+            }
+
+            return base.GetSignatures(arity);
+        }
+
+        public Task<FormulaValue> InvokeAsync(FormulaValue[] args, CancellationToken cancel)
```
MikeStall, line 90: >InvokeAsync [](http://example.com/codeflow?start=34&length=11)

Can you check with Jas if there are displayname issues here?

src/libraries/Microsoft.PowerFx.Core/Public/Values/InMemoryRecordValue.cs
anderson-joyle: From my tests in PA, when you are trying to Patch a base record from data source, the function should not add new fields to the base record. Only fields that exists in the base record will be modified.

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
MikeStall: (else, impl here looks good)

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
anderson-joyle: Yes. I'll change the handler's name to something more generic.

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
```diff
@@ -0,0 +1,187 @@
...
+                    continue;
+                }
+                else if (arg is ErrorValue)
+                {
+                    // @@@JYL How to handle error value arguments?
+                    return arg;
+                }
+                else if (arg is RecordValue record)
+                {
+                    // @@@JYL What is a ellegant way to avoid the nested loop?
```
MikeStall, line 168: > // @@@JYL What is a ellegant way to avoid the nested loop? [](http://example.com/codeflow?start=19&length=59)

nested loop is ok.  It could be extracted into its own function.

Although I wonder - can this code (153...174) be shared with the other overload of patch? via some shared helper

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
```diff
@@ -0,0 +1,187 @@
...
+                    {
+                        argFields[field.Name] = field.Value;
+                    }
+                }
+            }
+
+            foreach (KeyValuePair<string, FormulaValue> kvp in argFields)
+            {
+                arg1.UpdateField(new NamedValue(kvp));
+            }
```
MikeStall, line 179: UpdateField() will be a network call - so we don't want to call that in a loop. 

Instead, could we just change the signature to UpdateField to allow an IEnum<KeyValuePair<string, FormulaValue>> ? 
Then it's 1 network call. 

It can still have other simpler overloads that chain to that.

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
```diff
@@ -0,0 +1,187 @@
...
+
+            // @@@JYL Check if record exists. Append if it doesn't
```
MikeStall, line 181: > // @@@JYL Check if record exists. [](http://example.com/codeflow?start=11&length=34)

Is this correct semantics?
If it doesn't exist, do we just nop (or error)? 

Else, if they wanted append semantics, they needed to pass default() else for arg1  - and we don't need to support that yet (we can use Collect)

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
anderson-joyle: Also added to PatchFunction function

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
anderson-joyle: > Although I wonder - can this code (153...174) be shared with the other overload of patch? via some shared helper

I believe we can hold for now. The upcoming implementations of Patch don't have the same signature and we are also planning to move the implementation to the interpreter. Let me know if you agree @MikeStall .

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
anderson-joyle: > Is this correct semantics?

From [documentation](https://docs.microsoft.com/en-us/power-platform/power-fx/reference/function-patch#modify-or-create-a-set-of-records-in-a-data-source-1), record should be created if doesn't exist. 

Not supporting it for the time being.

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
anderson-joyle: Collect test cases have been moved to Collect.txt.

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.2-preview.20220826-318850](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.2-preview.20220826-318850).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfigExtensions.EnableCollectFunction(Microsoft.PowerFx.SymbolTable)' does not exist in the implementation but it does exist in the contract.
```

</details>

src/libraries/Microsoft.PowerFx.Core/Public/Values/InMemoryRecordValue.cs
MikeStall: "add new fields" can mean two things:
1. add a field that is not already on the Type
2. change an existing blank value to non-blank. 

I suspect 1 is not allowed, but I think 2 should be (we can double check).  This will block 2.

MikeStall: ClearCollect(_x, {f1:1, f2:2}, {f1:1, f3:3})
IsBlank(First(_x).f3) // true

Patch(_x, First(_x), { f3:33}) // Succeeds, now First(_x).f3 = 33

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
MikeStall: It looks like this can be shared immediately with the code at line 92.

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.2-preview.20220828-318998](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.2-preview.20220828-318998).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfigExtensions.EnableCollectFunction(Microsoft.PowerFx.SymbolTable)' does not exist in the implementation but it does exist in the contract.
```

</details>

src/libraries/Microsoft.PowerFx.Core/Public/Values/InMemoryRecordValue.cs
```diff
@@ -24,6 +25,7 @@ public InMemoryRecordValue(IRContext irContext, IReadOnlyDictionary<string, Form
             Contract.Assert(IRContext.ResultType is RecordType);
 
             _fields = fields;
+            _sourceDict = fields as IDictionary<string, FormulaValue>;
```
LucGenetier, line 28: Why do we need to store the data twice?
And the names are misleading as the "source" is 'fields'

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
```diff
@@ -0,0 +1,182 @@
...
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Syntax;
+using Microsoft.PowerFx.Types;
+using static Microsoft.PowerFx.Core.Localization.TexlStrings;
+
+namespace Microsoft.PowerFx.Interpreter
+{
+    /*
+         * !JYL!
```
LucGenetier, line 29: >!JYL! [](http://example.com/codeflow?start=11&length=5)

What's the meaning? Sorry if obvious...

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
```diff
@@ -0,0 +1,182 @@
...
+            }
+
+            return base.GetSignatures(arity);
+        }
+
+        public Task<FormulaValue> InvokeAsync(FormulaValue[] args, CancellationToken cancel)
+        {
+            var fieldsDict = new Dictionary<string, FormulaValue>();
+
+            var arg0 = args[0] as RecordValue;
```
LucGenetier, line 95: >arg0 [](http://example.com/codeflow?start=16&length=4)

What is this used for?

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
```diff
@@ -0,0 +1,182 @@
...
+                    // !JYL! How to handle error value arguments?
+                    return arg;
+                }
+                else if (arg is RecordValue record)
+                {
+                    foreach (var field in record.Fields)
+                    {
+                        argFields[field.Name] = field.Value;
+                    }
+                }
```
LucGenetier, line 174: >} [](http://example.com/codeflow?start=16&length=1)

We should add tests where these args are neither Blank, nor an ErrorValue nor a RecordValue

src/libraries/Microsoft.PowerFx.Interpreter/Environment/PowerFxConfigExtensions.cs
```diff
@@ -32,12 +32,15 @@ public static void EnableSetFunction(this PowerFxConfig powerFxConfig)
         }
 
         /// <summary>
-        /// Enable a Collect() function which allows scripts to append table records.
+        /// Enable all multation function which allows scripts to execute side effect behavior.
```
LucGenetier, line 35: >function [](http://example.com/codeflow?start=33&length=8)

with an 's'?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Collect.txt
```diff
@@ -0,0 +1,36 @@
...
+Errors: Error 0-10: Invalid number of arguments: received 1, expected 2.
+
+>> Collect(t, r1, r1)
+Errors: Error 0-18: Invalid number of arguments: received 3, expected 2.
+
+>> Collect(t, r1, 1; 2; r1)
+Errors: Error 0-24: Invalid number of arguments: received 3, expected 2.
+
+>> Collect(t, "x")
+Errors: Error 11-14: The first argument of '{0}' should be a valid variable name, and cannot conflict with any existing control, screen, collection, or data source names. Found type '{1}'|Error 0-15: The function 'Collect' has some invalid arguments.
```
LucGenetier, line 36: >'{0}' [](http://example.com/codeflow?start=43&length=5)

Is this normal to have '{0}' in an error message?

LucGenetier, line 36: >'{1}' [](http://example.com/codeflow?start=182&length=5)

Same question here

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch.txt
```diff
@@ -0,0 +1,33 @@
+﻿#SETUP: EnableExpressionChaining,MutationFunctionsTestSetup
+
+// PatchTestSetup handle does the following:
+// - Creates a record: r1 => {Field1:1,Field2:2}
+// - Creates a record: r2 => {Field1:11,Field2:22}
+// - Creates a record: r3 => {Field1:111,Field2:222}
+// - Creates a record: r4 => {Field1:1111,Field2:2222}
```
LucGenetier, line 7: >- Creates a record: r4 => {Field1:1111,Field2:2222} [](http://example.com/codeflow?start=3&length=51)

Try adding an empty record
Try adding records with other types (strings, dates, ...)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Collect.txt
CarlosFigueiraMSFT: No, this is wrong. It should have been replaced with 'Collect' in this case.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5004,6 +5004,22 @@
     <value>Contrast theme</value>
     <comment>Display text representing the Contrast value of TeamsTheme enum (TeamsTheme_Contrast_Name). The possible values for this enumeration are: Default, Dark, Contrast.</comment>
   </data>
+  <data name="AboutPatch" xml:space="preserve">
+    <value>Applies the updates to the specified row and changes the row in the specified datasource. Returns the updated row.</value>
+    <comment>Description of 'Patch' function.</comment>
+  </data>
+  <data name="PatchDataSourceArg" xml:space="preserve">
+    <value>Data source</value>
+    <comment>Required. The data source that contains the records that you want to modify or will contain the records that you want to create.</comment>
```
CarlosFigueiraMSFT, line 5013: >Required. The data source that contains the records that you want to modify or will contain the records that you want to create. [](http://example.com/codeflow?start=13&length=128)

The comment field is not something that is shown to the users; instead, it is context that we give to the translators so that they are able to translate the <value> correctly.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5004,6 +5004,22 @@
     <value>Contrast theme</value>
     <comment>Display text representing the Contrast value of TeamsTheme enum (TeamsTheme_Contrast_Name). The possible values for this enumeration are: Default, Dark, Contrast.</comment>
   </data>
+  <data name="AboutPatch" xml:space="preserve">
+    <value>Applies the updates to the specified row and changes the row in the specified datasource. Returns the updated row.</value>
+    <comment>Description of 'Patch' function.</comment>
+  </data>
+  <data name="PatchDataSourceArg" xml:space="preserve">
+    <value>Data source</value>
```
CarlosFigueiraMSFT, line 5012: >Data source [](http://example.com/codeflow?start=11&length=11)

`source` - we use single words for parameter names. If we do need multiple words, we concatenate them with `_`

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5004,6 +5004,22 @@
...
+    <comment>Required. The data source that contains the records that you want to modify or will contain the records that you want to create.</comment>
+  </data>
+  <data name="PatchBaseRecordArg" xml:space="preserve">
+    <value>Base record</value>
```
CarlosFigueiraMSFT, line 5016: >Base record [](http://example.com/codeflow?start=11&length=11)

`record` - we use single words for parameter names. If we do need multiple words, we concatenate them with `_`

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5004,6 +5004,22 @@
...
+    <comment>Required. The record to modify or create. If the record came from a data source, the record is found and modified. If the result of Defaults is used, a record is created.</comment>
+  </data>
+  <data name="PatchChangeRecordsArg" xml:space="preserve">
+    <value>Change record(s)</value>
```
CarlosFigueiraMSFT, line 5020: >Change record(s) [](http://example.com/codeflow?start=11&length=16)

`update` - we use single words for parameter names. If we do need multiple words, we concatenate them with `_`

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
anderson-joyle: This is a personal "to-do" tag that I can easily search across the code. These will be removed when code is moved from draft to ready to review.

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
anderson-joyle: It was used in the previous interactions. Removed.

src/libraries/Microsoft.PowerFx.Core/Public/Values/InMemoryRecordValue.cs
bryancroteau-MSFT: What was the resolution for this comment? I have the same question.

src/libraries/Microsoft.PowerFx.Core/Public/Values/InMemoryRecordValue.cs
```diff
@@ -41,5 +43,21 @@ protected override bool TryGetField(FormulaType fieldType, string fieldName, out
...
+        public override void UpdateFields(IEnumerable<KeyValuePair<string, FormulaValue>> record)
+        {
+            if (_sourceDict.IsReadOnly)
+            {
+                base.UpdateFields(record);
+            }
+
+            foreach (var field in record)
+            {
+                if (_sourceDict.ContainsKey(field.Key))
```
bryancroteau-MSFT, line 56: What is the behavior in the case where a field is being "patched" which does not exist in the record? Is this possible? (either at Compile time or Runtime?)

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
```diff
@@ -0,0 +1,182 @@
...
+using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Core.Localization;
+using Microsoft.PowerFx.Core.Texl.Builtins;
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Syntax;
+using Microsoft.PowerFx.Types;
+using static Microsoft.PowerFx.Core.Localization.TexlStrings;
+
+namespace Microsoft.PowerFx.Interpreter
```
bryancroteau-MSFT, line 26: Why is a texl function being added in this namespace? Why is this not under Texl/Builtins in Core?

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
```diff
@@ -0,0 +1,182 @@
...
+        {
+            if (arity > 2)
+            {
+                return GetGenericSignatures(arity, TexlStrings.PatchBaseRecordArg, TexlStrings.PatchChangeRecordsArg);
+            }
+
+            return base.GetSignatures(arity);
+        }
+
+        public Task<FormulaValue> InvokeAsync(FormulaValue[] args, CancellationToken cancel)
```
bryancroteau-MSFT, line 91: Why is the implementation defined in the same file as the front-end/type checking? Why are there no changes to Library.cs? Why does these functions not have standard error handling wrappers?

src/libraries/Microsoft.PowerFx.Core/Public/Values/InMemoryRecordValue.cs
anderson-joyle: @bryancroteau-MSFT I kept _fields and changed it from IReadOnlyDictionary to IDictionary.

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
MikeStall: These are interpreter-only implementations for Cards and need to be opted in.  

We would like to share them soon, but the binding implementation in PowerApps refer to PA-only concepts not in Core - so that will require some refactoring.

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
```diff
@@ -161,5 +161,10 @@ public sealed override void Visit(IValueVisitor visitor)
         {
             visitor.Visit(this);
         }
+
+        public virtual void UpdateFields(IEnumerable<KeyValuePair<string, FormulaValue>> record)
```
MikeStall, line 165: >void [](http://example.com/codeflow?start=23&length=4)

Patch() can fail. So this may need return a FormulaValue (like Append did) 

Try this in PowerApps, it would be a good test case:

IfError(Patch(Table({f2:5}), {f2:5}, { f2:6}), Notify(FirstError.Message))

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
MikeStall: Those should be binding errors. 
So we could just have an else here and throw an exception. 
(it's ok to throw exceptions when somebody circumvents the binder by calling SDK methods direclty)

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.2-preview.20220829-319146](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.2-preview.20220829-319146).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfigExtensions.EnableCollectFunction(Microsoft.PowerFx.SymbolTable)' does not exist in the implementation but it does exist in the contract.
```

</details>

anderson-joyle:
Move Patch (and Collect) to \Functions\Mutation folder.
Change their namespaces to Microsoft.PowerFx.Functions

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.2-preview.20220829-319193](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.2-preview.20220829-319193).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfigExtensions.EnableCollectFunction(Microsoft.PowerFx.SymbolTable)' does not exist in the implementation but it does exist in the contract.
```

</details>

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
```diff
@@ -1,155 +0,0 @@
...
-using Microsoft.PowerFx.Core.Utils;
-using Microsoft.PowerFx.Interpreter;
-using Microsoft.PowerFx.Syntax;
-using Microsoft.PowerFx.Types;
-using Xunit;
-using static Microsoft.PowerFx.Core.Localization.TexlStrings;
-
-namespace Microsoft.PowerFx.Interpreter.Tests
-{
-    public class CollectFunctionTests : PowerFxTest
```
MikeStall, line 25: >CollectFunctionTests [](http://example.com/codeflow?start=17&length=20)

deleted?

anderson-joyle: These test cases have been moved to Collect.txt

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.2-preview.20220830-319326](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.2-preview.20220830-319326).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfigExtensions.EnableCollectFunction(Microsoft.PowerFx.SymbolTable)' does not exist in the implementation but it does exist in the contract.
```

</details>

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.2-preview.20220830-319360](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.2-preview.20220830-319360).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfigExtensions.EnableCollectFunction(Microsoft.PowerFx.SymbolTable)' does not exist in the implementation but it does exist in the contract.
```

</details>

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.2-preview.20220830-319375](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.2-preview.20220830-319375).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfigExtensions.EnableCollectFunction(Microsoft.PowerFx.SymbolTable)' does not exist in the implementation but it does exist in the contract.
```

</details>

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
```diff
@@ -1,155 +0,0 @@
...
-            var t = FormulaValue.NewTable(r1.Type, listT);
-
-            symbol.AddConstant("t", t);
-            symbol.AddConstant("r1", r1);
-
-            var engine = new RecalcEngine();
-            var resultCount = await engine.EvalAsync(script, CancellationToken.None, options: _opts, symbolTable: symbol);
-            
-            Assert.Equal(expected, ((NumberValue)resultCount).Value);
-            Assert.Equal(expected, listT.Count);
```
MikeStall, line 53: > Assert.Equal(expected, listT.Count); [](http://example.com/codeflow?start=11&length=37)

It's good to keep this test, at least for this assert - which tests at the SDK level, so it's testing something that the .txts can't.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/PatchFunction.cs
```diff
@@ -0,0 +1,241 @@
...
+            Contracts.AssertValue(argTypes);
+            Contracts.Assert(args.Length == argTypes.Length);
+            Contracts.AssertValue(errors);
+
+            var isValid = CheckInvocation(args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
+
+            return isValid;
+        }
+
+        protected Dictionary<string, FormulaValue> CreateRecordFromArgsDict(FormulaValue[] args, int startFrom, out ErrorValue errorValue)
```
MikeStall, line 54: >protected [](http://example.com/codeflow?start=8&length=9)

nit - I think this could be static.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/PatchFunction.cs
```diff
@@ -0,0 +1,241 @@
...
+                    errorValue = error;
+                    return null;
+                }
+                else if (arg is RecordValue record)
+                {
+                    foreach (var field in record.Fields)
+                    {
+                        retFields[field.Name] = field.Value;
+                    }
+                }
```
MikeStall, line 77: >   } [](http://example.com/codeflow?start=13&length=4)

else throw. 
Binder will protect us, so exception is ok.

src/libraries/Microsoft.PowerFx.Core/Public/Values/InMemoryRecordValue.cs
```diff
@@ -41,5 +41,34 @@ protected override bool TryGetField(FormulaType fieldType, string fieldName, out
...
+                {
+                    return errorValue;
+                }
+
+                _fields[fieldName] = field.Value;
+
+                fields.Add(new NamedValue(field.Key, field.Value));
+            }
+
+            return NewRecordFromFields(fields);
```
MikeStall, line 71: >NewRecordFromFields(fields); [](http://example.com/codeflow?start=19&length=28)

we should get clarity about what the return value needs to be.

src/libraries/Microsoft.PowerFx.Core/Public/Values/InMemoryRecordValue.cs
```diff
@@ -41,5 +41,34 @@ protected override bool TryGetField(FormulaType fieldType, string fieldName, out
...
+            {
+                var fieldName = field.Key;
+
+                // Throws if field is missing
+                Type.GetFieldType(fieldName);
+
+                if (field.Value is ErrorValue errorValue)
+                {
+                    return errorValue;
+                }
```
MikeStall, line 64: is this right? 
Difference between:
- the whole record is an error 
- a field within the record is an error

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
MikeStall: This should probably be an Error Value (not exception), since it can happen at runtime. 

An exception is for SDK level errors and will abort the entire execution (and not return the error to the expression).

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/PatchFunction.cs
```diff
@@ -0,0 +1,241 @@
...
+                else if (isSafeToUnion)
+                {
+                    retType = DType.Union(retType, curType);
+                }
+            }
+
+            return isValid;
+        }
+
+        public async Task<FormulaValue> InvokeAsync(FormulaValue[] args, CancellationToken cancel)
```
MikeStall, line 227: >args [](http://example.com/codeflow?start=67&length=4)

It's suspicious that this function never uses args[0].

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/PatchFunction.cs
```diff
@@ -0,0 +1,241 @@
...
+        {
+            var argFields = CreateRecordFromArgsDict(args, 1, out ErrorValue errorValue);
+
+            var arg1 = args[1] as RecordValue;
```
MikeStall, line 231: >arg1 [](http://example.com/codeflow?start=16&length=4)

what if arg1 is blank or error?

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTests.cs
anderson-joyle: Moved back some SDK test cases.

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.2-preview.20220830-319390](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.2-preview.20220830-319390).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfigExtensions.EnableCollectFunction(Microsoft.PowerFx.SymbolTable)' does not exist in the implementation but it does exist in the contract.
```

</details>

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.2-preview.20220830-319441](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.2-preview.20220830-319441).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfigExtensions.EnableCollectFunction(Microsoft.PowerFx.SymbolTable)' does not exist in the implementation but it does exist in the contract.
```

</details>

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/PatchFunction.cs
```diff
@@ -0,0 +1,267 @@
...
+                else if (arg is RecordValue record)
+                {
+                    foreach (var field in record.Fields)
+                    {
+                        retFields[field.Name] = field.Value;
+                    }
+                }
+                else
+                {
+                    throw new ArgumentException($"Cann't handler {arg.Type} argument type.");
```
MikeStall, line 80: >Cann [](http://example.com/codeflow?start=50&length=4)

nit: typo

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/PatchFunction.cs
```diff
@@ -0,0 +1,267 @@
...
+            {
+                return args[1];
+            }
+
+            var argFields = CreateRecordFromArgsDict(args, 1, out ErrorValue errorValue);
+
+            if (errorValue != null)
+            {
+                return errorValue;
+            }
```
MikeStall, line 260: nit: now that we're checking for ErrorValue on line 244, we probably don't need the check in CreateRecordFromArgsDict. 
This could simplify it (remove the errorValue out param)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/PatchFunction.cs
```diff
@@ -0,0 +1,267 @@
...
+
+            var arg1 = args[1] as RecordValue;
+
+            return arg1.UpdateFields(argFields);
```
MikeStall, line 264: the more defensive coding would be:

 var arg1 = (RecordValue) args[1]; 

Before, if for some reason it's not a RecordValue, then 'as' operator returns null, and we'd get a null ref on line 264 -and null refs an be hard to track down. 

Whereas the cast will generated an explicit cast exception.

src/tests/Microsoft.PowerFx.Interpreter.Tests/CollectFunctionTest.cs
```diff
@@ -0,0 +1,56 @@
...
+using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Interpreter;
+using Microsoft.PowerFx.Syntax;
+using Microsoft.PowerFx.Types;
+using Xunit;
+using static Microsoft.PowerFx.Core.Localization.TexlStrings;
+
+namespace Microsoft.PowerFx.Interpreter.Tests
+{
+    public class CollectFunctionTests : PowerFxTest
```
MikeStall, line 25: >CollectFunctionTests [](http://example.com/codeflow?start=17&length=20)

nit - filename should be plural (will also make it matche the class name)

src/libraries/Microsoft.PowerFx.Core/Public/Values/InMemoryRecordValue.cs
```diff
@@ -41,5 +43,29 @@ protected override bool TryGetField(FormulaType fieldType, string fieldName, out
         {
             return _fields.TryGetValue(fieldName, out result);
         }
+
+        public override FormulaValue UpdateFields(IEnumerable<KeyValuePair<string, FormulaValue>> record)
+        {
+            if (_fieldsDict == null || _fieldsDict.IsReadOnly)
```
MikeStall, line 49: > || _fieldsDict.IsReadOnly [](http://example.com/codeflow?start=35&length=26)

to consider:
we could check this once in the ctor. 

if (_fieldsDict.IsReadOnly) {
 _fieldsDict = null;
}

And then elsewhere we're just doing
if ( _fields == null)

src/libraries/Microsoft.PowerFx.Core/Public/Values/InMemoryRecordValue.cs
```diff
@@ -12,6 +12,7 @@ namespace Microsoft.PowerFx.Types
     internal class InMemoryRecordValue : RecordValue
     {
         private readonly IReadOnlyDictionary<string, FormulaValue> _fields;
+        private readonly IDictionary<string, FormulaValue> _fieldsDict;
```
MikeStall, line 15: >_fieldsDict [](http://example.com/codeflow?start=59&length=11)

nit: consider renaming to:
  _mutableFields

That name captures the intent (that it's about mutability))

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
```diff
@@ -161,5 +161,23 @@ public sealed override void Visit(IValueVisitor visitor)
         {
             visitor.Visit(this);
         }
+
+        public virtual FormulaValue UpdateFields(IEnumerable<KeyValuePair<string, FormulaValue>> record)
```
MikeStall, line 165: >FormulaValue [](http://example.com/codeflow?start=23&length=12)

This probably needs to be async. (ie, for dataverse, it will make a network call). 

Type should also be: 

Task<DValue<RecordValue>>

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/PatchFunction.cs
MikeStall: Or we could go the other direction ... 
we could roll the error check on 242...248 into CreateRecordFromArgsDict. It gets passed the entire array, so it could scan everything.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/PatchFunction.cs
MikeStall: the real item here is making sure that arg1 is a member or arg0.

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.2-preview.20220830-319453](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.2-preview.20220830-319453).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfigExtensions.EnableCollectFunction(Microsoft.PowerFx.SymbolTable)' does not exist in the implementation but it does exist in the contract.
```

</details>

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.2-preview.20220830-319453](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.2-preview.20220830-319453).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfigExtensions.EnableCollectFunction(Microsoft.PowerFx.SymbolTable)' does not exist in the implementation but it does exist in the contract.
```

</details>

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.2-preview.20220830-319453](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.2-preview.20220830-319453).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.PowerFxConfigExtensions.EnableCollectFunction(Microsoft.PowerFx.SymbolTable)' does not exist in the implementation but it does exist in the contract.
```

</details>

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/PatchFunction.cs
```diff
@@ -0,0 +1,264 @@
...
+        }
+
+        protected static bool CheckArgs(FormulaValue[] args, out FormulaValue faultyArg)
+        {
+            // If any args are error, propagate up.
+            foreach (var arg in args)
+            {
+                if (arg is ErrorValue)
+                {
+                    faultyArg = (FormulaValue)arg;
```
MikeStall, line 61: >(FormulaValue) [](http://example.com/codeflow?start=32&length=14)

nit: cast unecessary

MikeStall (approved):
:shipit:

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch.txt
```diff
@@ -0,0 +1,56 @@
...
+>> Patch(t1, r1, {Field3:DateTime(2022,12,12,0,0,0,0)}, {Field4:false});t1
+Table({Field1:1,Field2:"earth",Field3:DateTime(2022,12,12,0,0,0,0),Field4:false})
+
+>> Patch(t1, r1, r2);t1
+Table({Field1:2,Field2:"moon",Field3:DateTime(2022,2,1,0,0,0,0),Field4:false})
+
+>> Patch(t1, r1, {Field5:"Field5"}); t1
+Errors: Error 0-32: The function 'Patch' has some invalid arguments.|Error 14-31: The specified column 'Field5' does not exist. The column with the most similar name is 'Field1'.
+
+// This test would fail in PA.
```
LucGenetier, line 45: >This test would fail in PA. [](http://example.com/codeflow?start=3&length=27)

Why?...
Add some explanation

LucGenetier (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Interpreter/PatchFunction.cs
anderson-joyle: It is going to be fixed on #611

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch.txt
anderson-joyle: "The data source supplied to the function is invalid"
I'll add the explanation in a future interaction.

PR #596, opened 2022-08-25
Move A* tests from NotYetReady to Ready

CarlosFigueiraMSFT:
Starting to move tests from the "NotYetReady" folder to the place where they can actually run. Starting with files starting with A. This found a missing functionality (DateTime to Number unary conversion) which was not implemented, so added it as well.


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Abs.txt
```diff
@@ -76,4 +76,12 @@ false
 >> Abs("false")
 #Error(Kind=InvalidArgument)
 
+// ******** Date values and functions with dates ********
```
lesaltzm, line 79: This'll likely cause test failures when we bump the version for JS, as a heads up. 
Might want to either do that yourself, or keep a doc somewhere of the changed test results so it's easy to fix.

CarlosFigueiraMSFT: Yes, I will bring the PFx version to JS when this is done, and handle the appropriate overrides.

lesaltzm (approved):

PR #618, opened 2022-09-01
Move B*,C* tests from NotYetReady to Ready

CarlosFigueiraMSFT:
Continuing to move tests from the "NotYetReady" folder to the place where they can actually run. Files starting with B, and C. Also update how time values are compared in test results.


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Coalesce.txt
```diff
@@ -8,12 +8,24 @@
...
+"a"
+
 >> Coalesce( Blank(), "", Blank(), "", "3", "4" )
 "3"
 
+>> Coalesce(Blank(), "", Blank(), "", Blank(), " ", "3", "4")
+" "
+
 >> Coalesce("")
 Blank()
```
Ananya-Gupta24, line 25: Unrelated to your changes, could you add a case here for an input with only Blanks and empty strings and verify that it returns Blank()?

CarlosFigueiraMSFT: Added

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Count.txt
```diff
@@ -1,20 +1,76 @@
...
+
+>> Count(Table({a:Blank()},{a:Blank()},{a:Blank()}))
+0
+
+// ******** BLANK TABLE ********
+
+>> Count(If(1<0,[1,2,3]))
+0
+
+// ******** ERRORS ********
```
Ananya-Gupta24, line 60: Should we have error cases here for when the first argument is an empty string or Blank() followed by a string or number value like Count(["", 1]) and Count([Blank(), "1"])?

CarlosFigueiraMSFT: Good point, will add some cases for (compiler) errors where the table doesn't have a numeric column (which is the case in your examples)

Ananya-Gupta24 (approved):

Ananya-Gupta24 (approved):

PR #624, opened 2022-09-02
Updated single table wrapper to use the wrapped scalar function

CarlosFigueiraMSFT:
Currently some of the tabular versions of scalar functions are not calling the scalar version properly - instead of going to the function as exposed to the user (the wrapper that handles errors and blank values), it is going to the inner implementation. This PR updates the tabular versions for single argument functions (multi-argument functions will come later).

This also fixes the Char function to treat Blank values as 0, like Excel does.


src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -32,8 +32,28 @@ private static T GetService<T>(this IServiceProvider services, T defaultService)
 
         public static IEnumerable<TexlFunction> FunctionList => FuncsByName.Keys;
 
+        private static readonly object LockObject = new object();
+
+        public static readonly IReadOnlyDictionary<TexlFunction, AsyncFunctionPtr> FuncsByName;
```
lesaltzm, line 37: nit: Funcs by name isn't quite right, the key here isn't the name it's the full function object

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -32,8 +32,28 @@ private static T GetService<T>(this IServiceProvider services, T defaultService)
 
         public static IEnumerable<TexlFunction> FunctionList => FuncsByName.Keys;
 
+        private static readonly object LockObject = new object();
```
lesaltzm, line 35: Is this used?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
CarlosFigueiraMSFT: True - this is what we had before, I didn't want to change more than necessary... I can update it, though

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
CarlosFigueiraMSFT: It was in an initial iteration, not anymore. I'll remove it.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
CarlosFigueiraMSFT: Actually, it's a public field, so it's possible that someone else is using it; I won't update it (at least not in this PR)

MikeStall: It's safe - 
public field on *internal* class ... 
And in Interpreter, so it's not included by PAClient (and products don't have internals access to interpreter)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -34,8 +34,26 @@ private static T GetService<T>(this IServiceProvider services, T defaultService)
...
         public static IEnumerable<TexlFunction> FunctionList => FuncsByName.Keys;
 
+        public static readonly IReadOnlyDictionary<TexlFunction, AsyncFunctionPtr> FuncsByName;
+
+        static Library()
+        {
+            var funcsByName = new Dictionary<TexlFunction, AsyncFunctionPtr>();
+            foreach (var func in SimpleFuncsByName)
+            {
+                funcsByName.Add(func.Key, func.Value);
```
MikeStall, line 44: >func.Key [](http://example.com/codeflow?start=32&length=8)

This could be an issue for overloads / name conflicts.

CarlosFigueiraMSFT: The name of this field is misleading - it's not based on name, but based on the function implementation itself. But since this is a public field, it's possible that someone has a dependency on it, so I don't want to change it right now.

CarlosFigueiraMSFT: Ah, just saw your other comment; will update it in this PR then.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
CarlosFigueiraMSFT: Updated the function name

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1425,6 +1353,100 @@ private static T GetService<T>(this IServiceProvider services, T defaultService)
...
+                    expandArguments: NoArgExpansion,
+                    replaceBlankValues: DoNotReplaceBlank,
+                    checkRuntimeTypes: ExactValueTypeOrBlank<TableValue>,
+                    checkRuntimeValues: DeferRuntimeValueChecking,
+                    returnBehavior: ReturnBehavior.ReturnBlankIfAnyArgIsBlank,
+                    targetFunction: StandardSingleColumnTable<NumberValue>(SimpleFunctionImplementations[BuiltinFunctionsCore.BooleanN]))
+            },
+            {
+                BuiltinFunctionsCore.CharT,
+                StandardErrorHandlingAsync(
```
MikeStall, line 1391: >StandardErrorHandlingAsync [](http://example.com/codeflow?start=16&length=26)

Most of these are                 StandardErrorHandlingAsync<TableValue>, 

should this be too?

CarlosFigueiraMSFT: Created another wrapper for tabular functions - code is much simpler now :-)

lesaltzm (approved):

src/tests/Microsoft.PowerFx.Interpreter.Tests/FunctionDefinitionTests.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Microsoft.PowerFx.Tests
+{
+    public class FunctionDefinitionTests : PowerFxTest
+    {
+        [Fact]
+        public void TabularOverloadListIsOverloadOfSingleFunction()
+        {
+            var type = typeof(Microsoft.PowerFx.Functions.Library);
+            var simpleFunctions = type.GetProperty("SimpleFunctionImplementations", BindingFlags.Static | BindingFlags.NonPublic).GetValue(null)
```
MikeStall, line 22: >SimpleFunctionImplementations [](http://example.com/codeflow?start=52&length=29)

nit - this could probably be nameof

PR #626, opened 2022-09-02
Add tests for custom record type that allows access to undefined fields

susosuso:
By default, Power FX is strongly + static typed. References to any undefined fields will result in error state at type checking time.

This set of unit tests gives an example which overrides that behavior by writing a custom `RecordType`, which always returns `FormulaType.UntypedObject` from `TryGetField()` when the targeted field is not defined.

Assume an expression `"obj.field1"`. If field1 is not defined, by default it will throw at `Check()`. With this change, it will pass type checking, and result in `BlankValue` (but of `UntypedObject` type) during runtime.


src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/CustomRecordTypeWithUndefinedFields.cs
```diff
@@ -0,0 +1,201 @@
...
+                }
+            };
+        }
+
+        [Theory]
+        [InlineData("obj.prop1", 1.0)]
+        [InlineData("IsError(obj.prop2)", true)] // type mismatch should fail at runtime
+        [InlineData("IsBlank(obj.missing)", "#exception")] // type checking fails when referencing undefined type.
+        [InlineData("IsBlank(obj.missing.missing)", "#exception")]
+        [InlineData("IsBlank(obj.prop_not_defined_in_type)", "#exception")]
```
MikeStall, line 48: >#exception [](http://example.com/codeflow?start=62&length=10)

where is this string coming from?

lesaltzm: Yes, instead of comparing the "#exception" string, can you compare to ex.Message?

src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/CustomRecordTypeWithUndefinedFields.cs
```diff
@@ -0,0 +1,201 @@
...
+                Console.WriteLine(ex.Message);
+            }
+        }
+
+        [Theory]
+        [InlineData("obj.prop1", 1.0)]
+        [InlineData("IsError(obj.prop2)", true)] // type mismatch should fail at runtime.
+        [InlineData("IsBlank(obj.missing)", true)] // fields not defined in RecordType is treated as Blank.
+        [InlineData("IsBlank(obj.missing.missing)", true)]
+        [InlineData("IsBlank(obj.prop_not_defined_in_type)", true)] // fields not defined in RecordType is always Blank, regardless of whether it is defined in actual data.
```
lesaltzm, line 77: note: this is the behavior defined on lines 117-121, not innate to the system. 
A different implementation could be returning an UntypedObjectValue for these.

susosuso: yes, I explicitly define that because I want to able to have a "blank if not exist" behavior in the expressions.

src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/CustomRecordTypeWithUndefinedFields.cs
susosuso: Sure, let me update. I was following the existing format from [another scenario test ](https://github.com/microsoft/Power-Fx/blob/a38e155bf82024f4c03989c0782506007c61f4a3/src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/ScenarioDotNetObjectWrapper.cs#L48)in this folder.

susosuso: Updated. I didn't end up using ex.Message because the actual exception message is outside of scope of this test. Let me know if you'd like another approach.

src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/CustomRecordTypeWithUndefinedFields.cs
```diff
@@ -0,0 +1,200 @@
...
+            var symbolTable = new SymbolTable();
+            var symbolValues = new SymbolValues();
+
+            symbolTable.AddVariable("obj", _customRecordType);
+            symbolValues.Add("obj", new TestCustomRecordValue(_testObj, _customRecordType));
+
+            try
+            {
+                var result = await engine.EvalAsync(expr, default, null, symbolTable, symbolValues).ConfigureAwait(false);
+                Assert.False(hasException);
```
MikeStall, line 90: > Assert.False(hasException); [](http://example.com/codeflow?start=15&length=28)

Note we could actually split the eval up into parts to be more specific and avoid the exception:

CheckResult check = engine.Check(expr, ...) 
Assert.Equals(check.IsSuccess, !hasException);

if (check.IsSuccess) {
    // Now we eval it, 
   var run = eval.GetEvaluator();
   var result = run.Eval(symbolValues);
   Assert.Equal(expected, result.ToObject());
}


This could actually be important if you want to use the same compiled expression against multiple inputs.

MikeStall (approved):
:shipit:

src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/CustomRecordTypeWithUndefinedFields.cs
susosuso: Thanks! Updated accordingly.

MikeStall (approved):
:shipit:

PR #633, opened 2022-09-07
Remove function initial draft

anderson-joyle:
Initial push for [**Remove( DataSource, Record1 [, Record2, ... ] [, All ] )**](https://docs.microsoft.com/en-us/power-platform/power-fx/reference/function-remove-removeif#syntax) implementation.

- Still working on a better way to find matching records.
- Working on adding extra test cases for SDK, compiler and interpreter.
- RemoveAsync returns a Boolean. Maybe there is a better way.
- Good approach to get a range of elements from array.
- Any other improvements.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
@@ -127,15 +127,15 @@ namespace Microsoft.PowerFx.Types
             {
                 var deleteList = new List<T>();
 
-                foreach (var current in _enumerator)
+                foreach (var item in _enumerator)
                 {
                     cancel.ThrowIfCancellationRequested();
 
-                    var currentRecord = current as RecordValue;
+                    var dRecord = Marshal(item);
 
-                    if (Matches(currentRecord, recordToRemove))
+                    if (Matches(dRecord.Value, recordToRemove))
                     {
-                        deleteList.Add(current);
+                        deleteList.Add(item);
 
                         if (!all)
                         {
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
@@ -108,6 +108,41 @@ namespace Microsoft.PowerFx.Types
             }
         }
 
+        public override async Task<DValue<BooleanValue>> RemoveAsync(IEnumerable<FormulaValue> recordsToRemove, bool all = false)
+        {
+            var ret = false;
+
+            if (_sourceList == null || _sourceList.IsReadOnly)
+            {
+                return await base.RemoveAsync(recordsToRemove, all);
+            }
+
+            // !JYL! Needs improvement
+            foreach (RecordValue record in recordsToRemove)
+            {
+                while (true)
+                {
+                    var actual = Find(record);
+
+                    if (actual != null)
+                    {
+                        _sourceList.Remove(MarshalInverse(actual));
+
+                        if (!all)
+                        {
+                            break;
+                        }
+                    }
+                    else
+                    {
+                        break;
+                    }                    
+                }
+            }
+
+            return DValue<BooleanValue>.Of(New(ret));
+        }
+
         protected override async Task<DValue<RecordValue>> PatchCoreAsync(RecordValue baseRecord, RecordValue changeRecord)
         {
             var actual = Find(baseRecord);
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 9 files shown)


BruceHaley:

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/RemoveFunction.cs
```diff
@@ -0,0 +1,218 @@
...
+                if (lastArgValue.ToUpper() == "ALL")
+                {
+                    all = true;
+                    toExclude = 2;
+                }
+            }
+
+            var datasource = (TableValue)args[0];
+
+            // !JYL! Better approach?
```
MikeStall, line 210: > // !JYL! Better approach? [](http://example.com/codeflow?start=11&length=26)

I think this is actually ok. 
It's still a O[n], traversing the enumeration once. 

The alternative is probably a call to Array.Copy
[https://docs.microsoft.com/en-us/dotnet/api/system.array.copy?view=net-6.0](https://docs.microsoft.com/en-us/dotnet/api/system.array.copy?view=net-6.0)

MikeStall:
>                 Message = $"It is not possible to call {methodName} method from TableValue directly.",

nit: 
better error message, 

"{methodName} is not supported on this table instance."



---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs:91 in 53bdf07. [](commit_id = 53bdf07658e027e3cdd82311cf170251cf3c3f4a, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
```diff
@@ -108,6 +108,41 @@ record = Marshal(item);
             }
         }
 
+        public override async Task<DValue<BooleanValue>> RemoveAsync(IEnumerable<FormulaValue> recordsToRemove, bool all = false)
+        {
+            var ret = false;
+
+            if (_sourceList == null || _sourceList.IsReadOnly)
```
MikeStall, line 115: >_sourceList.IsReadOnly) [](http://example.com/codeflow?start=39&length=23)

nit: we should check this in the ctor
if (_sourceList.IsReadOnly) { _sourceList  = null}

that can simplify the check on line 78 as well.

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
```diff
@@ -108,6 +108,41 @@ record = Marshal(item);
             }
         }
 
+        public override async Task<DValue<BooleanValue>> RemoveAsync(IEnumerable<FormulaValue> recordsToRemove, bool all = false)
```
MikeStall, line 111: >RemoveAsync [](http://example.com/codeflow?start=57&length=11)

Looking at these loops reminds me, we should be taking in a CancellationToken parameter (passed in from caller), 
and call 
cancel.ThrowIfCancellationRequested() inside the loop. 
Derived async impls will pass this cancellation token into their network calls.

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
```diff
@@ -108,6 +108,41 @@ record = Marshal(item);
...
+            // !JYL! Needs improvement
+            foreach (RecordValue record in recordsToRemove)
+            {
+                while (true)
+                {
+                    var actual = Find(record);
+
+                    if (actual != null)
+                    {
+                        _sourceList.Remove(MarshalInverse(actual));
```
MikeStall, line 129: > _sourceList.Remove(MarshalInverse(actual)); [](http://example.com/codeflow?start=23&length=44)

This may not work reliably. 

MarshalInverse can return a new instance of T, which may not Object.Equals match to the T in the list. 


We may need to iterate on _enumerator directly. 

var deleteList = new List<T>();
foreach(var item in _enumerator)
{
  if (Matches(item, record) {
     deleteList.Add(item); 
  }
}

foreach(var delete in deleteList)
{
  // delete is the same object instance as what was in the list. 
  _sourceList.Remove(delete);
}

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
```diff
@@ -108,6 +114,45 @@ record = Marshal(item);
...
+
+            foreach (RecordValue recordToRemove in recordsToRemove)
+            {
+                var deleteList = new List<T>();
+
+                foreach (var current in _enumerator)
+                {
+                    cancel.ThrowIfCancellationRequested();
+
+                    var currentRecord = current as RecordValue;
```
MikeStall, line 134: >r currentRecord = current as RecordValue; [](http://example.com/codeflow?start=22&length=41)

This will frequently fail. 
Marshal() instead?

MikeStall: To see this, 
can you add a test to  DataTableEvalTest that uses Remove on one of those tables?
That test derives from CollectionTableValue where T is not a RecordValue, and should fail here.

BruceHaley:

MikeStall (approved):

PR #636, opened 2022-09-07
Move D* tests from NotYetReady to Ready

CarlosFigueiraMSFT:
Continuing to move tests from the "NotYetReady" folder to the place where they can actually run. Files starting with D. This is also fixing some issues in the interpreter:
- DateDiff was not honoring date part boundaries
- Proper handling of blank values in DateAdd/DateDiff/Date functions
- Fix bug with DateAdd/DateDiff across daylight saving time boundaries
- Better errors for invalid dates


src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -336,10 +338,10 @@ private static T GetService<T>(this IServiceProvider services, T defaultService)
                 BuiltinFunctionsCore.Date,
                 StandardErrorHandling<NumberValue>(
                     expandArguments: NoArgExpansion,
-                    replaceBlankValues: DoNotReplaceBlank,
+                    replaceBlankValues: ReplaceBlankWithZero,
                     checkRuntimeTypes: ExactValueTypeOrBlank<NumberValue>,
                     checkRuntimeValues: ExactSequence(
-                        PositiveNumericNumberChecker,
+                        FiniteChecker,
```
Ananya-Gupta24, line 344: What additional scenario(s) is this accounting for?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
```diff
@@ -145,49 +158,59 @@ public static FormulaValue DateDiff(EvalVisitor runner, EvalVisitorContext conte
...
             // as we could even change of year if the date we have is 1st of Jan and we are in a UTC+N time zone (N positive)
-            switch (units.Value.ToLower())
+            switch (timeUnit)
             {
                 case "months":
                     double months = ((end.Year - start.Year) * 12) + end.Month - start.Month;
                     return new NumberValue(irContext, months);
                 case "quarters":
-                    var quarters = ((end.Year - start.Year) * 4) + Math.Floor(end.Month / 3.0) - Math.Floor(start.Month / 3.0);
+                    var quarters = ((end.Year - start.Year) * 4) + Math.Floor((end.Month - 1) / 3.0) - Math.Floor((start.Month - 1) / 3.0);
```
Ananya-Gupta24, line 173: Could you please add a comment here explaining the decrement to start and end months?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DateValue.txt
```diff
@@ -1,6 +1,12 @@
 ﻿
 // Use of DateValue, DateTimeValue
 
+>> DateValue("")
```
Ananya-Gupta24, line 4: Could you add a case for Blank() as a parameter?

Ananya-Gupta24 (changes requested):

Ananya-Gupta24 (dismissed):

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
CarlosFigueiraMSFT: The validation is already done in the implementation. And the PositiveNumericNumberChecker returns a Error(Kind=Numeric), which should be used for numeric functions only; the actual error type should be InvalidArgument

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DateValue.txt
CarlosFigueiraMSFT: Good idea, added as well

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
CarlosFigueiraMSFT: Added

Ananya-Gupta24 (approved):

PR #638, opened 2022-09-07
Allow lookup field delegation with 'in' operator

sagarbakliwal:


src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/InOpDelegationStrategy.cs
```diff
@@ -190,7 +194,7 @@ public override bool IsOpSupportedByTable(OperationCapabilityMetadata metadata,
             if (!(_binaryOpNode.Right.Kind == NodeKind.FirstName ||
                 binding.IsFullRecordRowScopeAccess(_binaryOpNode.Right) ||
                 (metadata.IsDelegationSupportedByTable(DelegationCapability.CdsIn) &&
-                (_binaryOpNode.Left.Kind == NodeKind.FirstName || binding.IsFullRecordRowScopeAccess(_binaryOpNode.Left)) &&
+                (_binaryOpNode.Left.Kind == NodeKind.FirstName || binding.IsFullRecordRowScopeAccess(_binaryOpNode.Left) || (_binaryOpNode.Left.Kind == NodeKind.DottedName && binding.GetType((_binaryOpNode.Left as DottedNameNode).Left).HasExpandInfo)) &&
```
Shpakh-MSFT, line 197: It would be good to simplify this condition. It's getting way too complicated.

src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/OpDelegationStrategy.cs
```diff
@@ -77,27 +77,28 @@ private bool IsSupportedNode(TexlNode node, OperationCapabilityMetadata metadata
...
+            //  4) it is a single column table and
+            //  5) metadata belongs to cds datasource that supports delegation of CdsIn
+            // If this check fails, verify if it is simply a valid node..
+            // Eg of valid delegation functions -
+            // Filter(Accounts, 'Account Name' in ["Foo", Bar"]) - Direct table use
+            // Set(Names, ["Foo", Bar"]); Filter(Accounts, 'Account Name' in Names) - Using variable of type table
+            // ClearCollect(Names, Accounts); Filter(Accounts, 'Account Name' in Names.'Account Name') - using column from collection.
+            // This won't be delegated - Filter(Accounts, 'Account Name' in Accounts.'Account Name') as Accounts.'Account Name' is async.
+            if (binding.Document.Properties.EnabledFeatures.IsEnhancedDelegationEnabled &&
+                isRHSNode && (opDelStrategy as BinaryOpDelegationStrategy)?.Op == BinaryOp.In && !binding.IsAsync(node) && binding.GetType(node).IsTable && binding.GetType(node).IsColumn &&
```
Shpakh-MSFT, line 93: >(opDelStrategy as BinaryOpDelegationStrategy)?.Op == BinaryOp.I [](http://example.com/codeflow?start=29&length=63)

Shouldn't this be part of InOpDelegationStrategy if it is "in" operator specific?

src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/OpDelegationStrategy.cs
```diff
@@ -77,27 +77,28 @@ private bool IsSupportedNode(TexlNode node, OperationCapabilityMetadata metadata
...
-                    isRHSNode && (opDelStrategy as BinaryOpDelegationStrategy)?.Op == BinaryOp.In && !binding.IsAsync(node) && binding.GetType(node).IsTable && binding.GetType(node).IsColumn &&
-                    metadata.IsDelegationSupportedByTable(DelegationCapability.CdsIn)) ||
-                    IsValidNode(node, binding))
-                {
-                    return true;
-                }
+                return true;
+            }
+
+            if (!binding.IsRowScope(node) && IsValidNode(node, binding))
```
Shpakh-MSFT, line 99: >!binding.IsRowScope(node) [](http://example.com/codeflow?start=16&length=25)

What is the reason for moving this check below? Just wanted to validate that this is not an oversight.

sagarbakliwal: I moved the special logic for cdsin away from three and made the code exactly same as it used to be before any in delegation

Shpakh-MSFT (approved):
:shipit:

PR #645, opened 2022-09-08
Fork: Remove 'Delete old compat results'

BruceHaley:
DO NOT MERGE. TESTING ONLY.
This is to see whether Southworks' github-pr-comment is working correctly.

Supposedly, task 'Publish compat results to GitHub' was changed to overwrite its previous publish. That means the second time [experimental]PowerFx-CI runs against a PR, the second published API compatibility comment does not create a duplicate. That makes the workaround, task 'Delete old compat results', redundant.

The fix:
Remove redundant task 'Delete old compat results'.


BruceHaley:
Testing completed

PR #646, opened 2022-09-08
Fix description for the DateTime function

CarlosFigueiraMSFT:
The description for the DateTime function is incorrect, this fixes it.


Shpakh-MSFT (approved):
:shipit:

PR #650, opened 2022-09-11
Fix crash in enumbuilder.

MikeStall:
If the config has a chained symbol table:

  powerFxConfig.SymbolTable = new SymbolTable() { Parent = dataverseConnection.Symbols };

Then expressions like this get a null ref:
(IEnumStore) config.EnumStoreBuilder.Build(),

because config.EnumStoreBuilder is just config.SymbolTable.EnumStoreBuilder, which is null.

It's not enough to just initialize SymbolTable.EnumStoreBuilder to non-null, we need to ensure the symbols are actually composing.

This runs all Suggest tests in both configurations.

Marking config.EnumStoreBuilder as obsolete.
PowerApps doesn't call config.EnumStoreBuilder.
Dataverse does, but those call should config.EnumStore instead.

This was related to #503.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
@@ -28,8 +28,11 @@ namespace Microsoft.PowerFx
         {
             DebugName = "DefaultConfig"
         };
-                
+
+        [Obsolete("Use Config.EnumStore or symboltable directly")]
         internal EnumStoreBuilder EnumStoreBuilder => SymbolTable.EnumStoreBuilder;
+
+        internal IEnumStore EnumStore => ReadOnlySymbolTable.Compose(SymbolTable);
 
         public CultureInfo CultureInfo { get; }
 
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
@@ -25,7 +25,7 @@ namespace Microsoft.PowerFx
     /// </summary>
     [DebuggerDisplay("{_debugName}")]
     [ThreadSafeImmutable]
-    public class ReadOnlySymbolTable : INameResolver, IGlobalSymbolNameResolver
+    public class ReadOnlySymbolTable : INameResolver, IGlobalSymbolNameResolver, IEnumStore
     {
         // Changed on each update. 
         // Host can use to ensure that a symbol table wasn't mutated on us.                 
@@ -103,6 +103,7 @@ namespace Microsoft.PowerFx
 
         // Which enums are available. 
         // These do not compose - only bottom one wins. 
+        // ComposedReadOnlySymbolTable will handle composition by looking up in each symbol table. 
         private protected EnumStoreBuilder _enumStoreBuilder;
         private EnumSymbol[] _enumSymbolCache;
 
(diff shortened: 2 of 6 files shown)


jas-valgotar (approved):

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseProvider.cs
```diff
@@ -27,7 +27,7 @@ internal static class IntellisenseProvider
 
         internal static IIntellisense GetIntellisense(PowerFxConfig config)
         {
-            return new Intellisense(config, config.EnumStoreBuilder.Build(), SuggestionHandlers);
```
MikeStall, line 30: >EnumStoreBuilder [](http://example.com/codeflow?start=51&length=16)

config.EnumStoreBuilder is 
config.SymbolTable.EnumStoreBuilder, which is null. 

Since this is intellisense, it's just reading from the store (doesn't need a full builder). 
That's also important because a reader can compose across multiple symbol tables, whereas a builder can't.

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -28,9 +28,12 @@ public sealed class PowerFxConfig
         {
             DebugName = "DefaultConfig"
         };
-                
+
+        [Obsolete("Use Config.EnumStore or symboltable directly")]
         internal EnumStoreBuilder EnumStoreBuilder => SymbolTable.EnumStoreBuilder;
```
MikeStall, line 34: We should remove this. 
Can't remove quite yet  - it's used in DataverseIntellisense. DataverseIntellisenseData (although not when we have composed symbol tables, so they don't hit the bug) 

PowerApps doesn't use it.

src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
```diff
@@ -25,7 +25,7 @@ namespace Microsoft.PowerFx
     /// </summary>
     [DebuggerDisplay("{_debugName}")]
     [ThreadSafeImmutable]
-    public class ReadOnlySymbolTable : INameResolver, IGlobalSymbolNameResolver
+    public class ReadOnlySymbolTable : INameResolver, IGlobalSymbolNameResolver, IEnumStore
```
MikeStall, line 28: >IEnumStore [](http://example.com/codeflow?start=81&length=10)

IEnumStore is conceptually part of the symbol table, just like functions, names, etc.

src/tests/Microsoft.PowerFx.Core.Tests/IntellisenseTests/SuggestTest.cs
```diff
@@ -212,10 +224,26 @@ public void TestSuggestEmptyEnumList(string expression, params string[] expected
...
+            // With adjusted config 
+            AdjustConfig(config);
+            actualSuggestions = SuggestStrings(expression, config);
             Assert.Equal(expectedSuggestions, actualSuggestions);
         }
 
+        // Add an extra (empy) symbol table into the config and ensure we get the same results. 
+        private void AdjustConfig(PowerFxConfig config)
+        {
+            config.SymbolTable = new SymbolTable
```
MikeStall, line 240: >SymbolTable [](http://example.com/codeflow?start=37&length=11)

This new symbol table has a null EnumStoreBuilder 
Hence, causing config.EnumStoreBuilder to return null and hit the nullref. 

Fix is to avoid calling config.EnumStoreBuilder all together, and instead call config.EnumStore  - and that enum store can properly enumerate from the full chain.

PR #658, opened 2022-09-13
add start of FormulaValue Serializer

MikeStall:


MikeStall:
Close in favor of #690

PR #659, opened 2022-09-13
Fixes for Date/time operations and functions

CarlosFigueiraMSFT:
This PR moves more date/time-related tests from the "notYetReady" folder to run, and fixes a bunch of issues with date/time operations and functions:
- Adding / subtracting numbers to date values now honors the fractional part of the numbers (and the resulting type is DateTime)
- Similar fix for the DateAdd function with Date values
- Fixed time difference (timeValue1 - timeValue2)
- Allowed the addition of DateTime and Time values
- Allowed the addition of Time values
- Allowed the subtraction of time values from date[time] values


src/libraries/Microsoft.PowerFx.Core/IR/Nodes/BinaryOpKind.cs
```diff
@@ -17,11 +17,16 @@ internal enum BinaryOpKind
         AddDateAndTime, // Date + Time
         AddDateAndDay, // Date + Number (Days)
         AddDateTimeAndDay,
-        AddTimeAndMilliseconds, // Time + Number (Milliseconds)
+        AddTimeAndNumber, // Time + Number (typically fraction of a day)
```
CarlosFigueiraMSFT, line 20: >AddTimeAndNumber [](http://example.com/codeflow?start=8&length=16)

Since this is an internal enumeration I believe it's ok to rename it, but will double check

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryOperators.cs
```diff
@@ -562,7 +675,38 @@ private static FormulaValue TimeDifference(IRContext irContext, FormulaValue[] a
...
+                    return CommonErrors.RuntimeTypeMismatch(irContext);
+            }
+
+            var arg1 = (TimeValue)args[1];
+
+            var result = arg0.Subtract(arg1.Value);
+            return new DateTimeValue(irContext, result);
+        }
+
+        private static FormulaValue SubtractNumberAndDate(IRContext irContext, FormulaValue[] args)
```
CarlosFigueiraMSFT, line 702: >SubtractNumberAndDate [](http://example.com/codeflow?start=36&length=21)

This operation (and the next one) are currently accepted by the compiler/binder. But their results are wrong (both in C# and in JS). So we'll keep them working in the compiler (otherwise it would be a breaking change for Power Apps), but always return an error.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -164,7 +164,7 @@ public FormulaValue GetResult(IRContext irContext)
             }
         }
 
-        private class MinDateTimeAgg : IAggregator
+        private class MinDateAndDateTimeAgg : IAggregator
```
CarlosFigueiraMSFT, line 167: >MinDateAndDateTimeAgg [](http://example.com/codeflow?start=22&length=21)

Merged the aggregator for date and datetime values, since the logic (and underlying data type, DateTime) is the same.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -296,7 +279,7 @@ public FormulaValue GetResult(IRContext irContext)
             }
         }
 
-        private class MaxDateAgg : IAggregator
+        private class MaxDateAndDateTimeAgg : IAggregator
```
CarlosFigueiraMSFT, line 282: >MaxDateAndDateTimeAgg [](http://example.com/codeflow?start=22&length=21)

Merged the aggregator for date and datetime values, since the logic (and underlying data type, DateTime) is the same.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryOperators.cs
```diff
@@ -125,6 +125,30 @@ internal static partial class Library
...
 
+        public static readonly AsyncFunctionPtr OperatorAddTimeAndNumber = StandardErrorHandling<FormulaValue>(
+            expandArguments: NoArgExpansion,
+            replaceBlankValues: ReplaceBlankWith(
+                new TimeValue(IRContext.NotInSource(FormulaType.Time), TimeSpan.Zero),
+                new NumberValue(IRContext.NotInSource(FormulaType.Number), 0)),
+            checkRuntimeTypes: ExactSequence(
+                TimeOrDateTime,
+                ExactValueType<NumberValue>),
+            checkRuntimeValues: DeferRuntimeValueChecking,
```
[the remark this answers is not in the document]
CarlosFigueiraMSFT: Yep - we need to do finite value checking for all inputs; I have a PR coming up right after this one that will do that (this one is already quite large)

bryancroteau-MSFT (approved):

Ananya-Gupta24 (dismissed):

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs
```diff
@@ -168,6 +168,16 @@ public static ErrorValue NotYetImplementedError(IRContext irContext, string mess
             });
         }
 
+        public static ErrorValue NotSupportedError(IRContext irContext)
+        {
+            return new ErrorValue(irContext, new ExpressionError()
+            {
+                Message = $"Not supported",
```
LucGenetier, line 175: >Not supported [](http://example.com/codeflow?start=28&length=13)

Shouldn't this error be more "Invalid Operation" as it's used when trying to a number and a date or a number and a time.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1197,7 +1197,7 @@ private static T GetService<T>(this IServiceProvider services, T defaultService)
                     expandArguments: NoArgExpansion,
                     replaceBlankValues: ReplaceBlankWithZero,
                     checkRuntimeTypes: ExactValueTypeOrBlank<NumberValue>,
-                    checkRuntimeValues: PositiveNumericNumberChecker,
+                    checkRuntimeValues: DeferRuntimeTypeChecking,
```
LucGenetier, line 1200: >DeferRuntimeTypeChecking [](http://example.com/codeflow?start=40&length=24)

Why? Do we want to support Sqrt(Time)? 
[you could add a comment]

LucGenetier:
>         public static FormulaValue DateDiff(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)

(Maybe out of scope but as we're fixing date and time operations)

We should probably use runner.GetService<TimeZoneInfo>() 

And test

#SETUP: TimeZoneInfo("Pacific Standard Time")
>>  DateDiff(DateTimeValue("2022-03-27 1am"),DateTimeValue("2022-03-27 4am"), Minutes)
120


---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs:126 in a5df486. [](commit_id = a5df486e85eda23f4a4192c3c6bdff1d3c80a3b0, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Date.txt
```diff
@@ -214,9 +215,32 @@ true
...
+
+>> Date(2001, 2, 14) + 2.3
+DateTime(2001,2,16,7,12,0,0)
+
+>> Date(2001, 2, 14) - 2.3
+DateTime(2001,2,11,16,48,0,0)
+
+// Across DST
+>> Date(2022, 1, 1) + 180
+DateTime(2022,6,30,0,0,0,0)
```
LucGenetier, line 234: The result cannot be at midnight due to DST

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Date.txt
```diff
@@ -225,8 +249,35 @@ Date(2001,2,16)
 >> Date(2022, 1, 15) - Date(2022, 1, 30)
 -15
 
->> Date(2022, 1/0, 15) - Date(2022, 1, 30)
-#Error(Kind=Div0)
+>> Date(2022, 10, 1) + Time(12, 34, 56)
```
LucGenetier, line 252: >Time(12, 34, 56) [](http://example.com/codeflow?start=23&length=16)

Try with Time(30, 0, 0) as well

LucGenetier:
> >> Date(2022,2,29)

XL doesn't accept this

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Date.txt:284 in a5df486. [](commit_id = a5df486e85eda23f4a4192c3c6bdff1d3c80a3b0, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DateAdd.txt
```diff
@@ -88,16 +88,16 @@ false
 #Error
 
 >> DateAdd(Date(1, 2, 3), 3)
-Date(1,2,6)
+DateTime(1,2,6,0,0,0,0)
 
 >> Day(DateAdd(Date(2001,1,1), 4, TimeUnit.Days))
 5
 
 >> DateAdd(Date(2011,1,15), 100000000, "milliseconds")
```
LucGenetier, line 96: >100000000 [](http://example.com/codeflow?start=28&length=9)

 Will it overflow if we use 10^12 ms? (it should work as it's ~31 years and a few months)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DateAdd.txt
```diff
@@ -126,10 +126,8 @@ DateTime(2000,1,1,0,0,0,0)
 >> DateAdd(DateTime(2000,1,1,0,0,0),-1,TimeUnit.Milliseconds)
 DateTime(1999,12,31,23,59,59,999)
 
-// TO BE DETERMINED: Adding fractional values
-// Subtracting 1.9 ms truncated to 1ms
-// >> DateAdd(DateTime(2000,1,1,0,0,0),-1.9,TimeUnit.Milliseconds)
-// DateTime(1999,12,31,23,59,59,999)
+>> DateAdd(DateTime(2000,1,1,0,0,0),-1.9,TimeUnit.Milliseconds)
```
LucGenetier, line 129: >-1.9 [](http://example.com/codeflow?start=36&length=4)

I'd also try with -1.4 and -1.5

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DateDiff.txt
```diff
@@ -232,14 +232,59 @@
...
+1
+
+>> DateDiff(DateTime(2011,12,31,23,59,59,999),Date(2012,1,1), TimeUnit.Months)
+1
+
+>> DateDiff(DateTime(2011,12,31,23,59,59,999),Date(2012,1,1), TimeUnit.Quarters)
+1
+
+>> DateDiff(DateTime(2011,12,31,23,59,59,999),Date(2012,1,1), TimeUnit.Years)
+1
```
LucGenetier, line 266: I really like this one :)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Time.txt
```diff
@@ -88,17 +88,23 @@ true
...
 
->> Text(Time(10, -30, 360))
-"9:36 AM"
+>> Text(Time(10, 30, -360), "HH:mm:ss")
+"10:24:00"
 
->> Text(Time(10, 30, -360))
-"10:24 AM"
+>> Text(Time(10, -30, -360), "HH:mm:ss")
+"09:24:00"
```
LucGenetier, line 101: We should try Time(23, ...) and maybe with "hh:mm:ss" as well + other characters like YY, ffff, g, tt...

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Time.txt
```diff
@@ -134,4 +140,203 @@ true
...
+>> Time(12,0,0) - Date(2022,10,10)
+Errors: Error 13-14: This operation isn't valid on these types: Time, Date.
+
+>> Time(12,0,0) - DateTime(2022,10,10,0,0,0)
+Errors: Error 13-14: This operation isn't valid on these types: Time, DateTime.
+
+>> 0 - Time(2,30,45)
+#Error(Kind=NotSupported)
+
+// *********** COMPONENTS FROM OVERFLOWS **********
```
LucGenetier, line 198: >OVERFLOWS [](http://example.com/codeflow?start=31&length=9)

I'd try
>> Time(12,50,45,950) + Time(21,15,35,750)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Time.txt
```diff
@@ -134,4 +140,203 @@ true
...
+>> Second(If(1<0,Time(1,1,1)))
+0
+
+>> Second(If(1<0,DateTime(2000,1,1,1,1,1)))
+0
+
+>> Time(1/0, 34, 56)
+#Error(Kind=Div0)
+
+>> Time(12, Sqrt(-1), 56)
```
LucGenetier, line 270: >Time(12, Sqrt(-1), 56) [](http://example.com/codeflow?start=3&length=22)

That should be Time(12, i, 56), no? :)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs
CarlosFigueiraMSFT: Updated the message. Also moved this to LibraryOperators, since this is only used in that file anyway.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
CarlosFigueiraMSFT: The runtime type check should take care of different values being passed to it. And we should be able to support Sqrt(Time) - after all, a Time value can be coerced to a number - so `Sqrt(Time(6,0,0))` should be `0.5` (since the numeric value of 06:00:00 is 0.25)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Date.txt
```diff
@@ -278,6 +329,16 @@ Date(2022,1,1)
 >> Year(44811)
 2022
 
+// Boolean to date (should fail)
+>> Day(true)
```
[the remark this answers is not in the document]
CarlosFigueiraMSFT: This is already a compile-time error, so repeating this test for a different boolean balue should not provide extra coverage. I will change one of them to be false, though, just to cover it as well.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Date.txt
CarlosFigueiraMSFT: Actually, when we are adding days / months / quarters / years, this is the behavior that we get. For example, `Console.WriteLine(new DateTime(2022, 1, 1, 0, 0, 0, DateTimeKind.Local).AddDays(180).ToString("yyyy-MM-dd HH:mm:ss"));` will print out "2022-06-30 00:00:00". I added two new test files for both the US Pacific and Romance time zones just in case.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Date.txt
CarlosFigueiraMSFT: Great idea - it found a bug in our Time implementation. Excel wraps it to be the same as Time(6,0,0), we weren't doing it. Fixed it.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DateAdd.txt
CarlosFigueiraMSFT: That works fine. Even 10^13 (the result will be around 2327) and 10^14 (5179) would work. Only when we go beyond 9999-12-31 is that we fail

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DateDiff.txt
CarlosFigueiraMSFT: :-D

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Time.txt
CarlosFigueiraMSFT: I'll leave it to a future PR; what I wanted to do in this one was to make the tests more locale-independent (since "9:24 AM" is very US-centric). There is still more work to do in other test files

CarlosFigueiraMSFT: But I'll add some sanity cases for larger times as well

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Time.txt
CarlosFigueiraMSFT: Added

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Time.txt
CarlosFigueiraMSFT: We don't have to worry about this - at least not yet. Do you want to be the one to introduce complex numbers to Power Fx? :-D

CarlosFigueiraMSFT:
>                 Message = "Unknown error",

We do need to take a pass in all error messages - and move them to resources (so that they are properly localized). Mike should have this in his radar.

---
In reply to: [1249115516](https://github.com/microsoft/Power-Fx/pull/659#issuecomment-1249115516) [](http://example.com/codeflow?ancestors=1249115516,1249115516)

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs:155 in a5df486. [](commit_id = a5df486e85eda23f4a4192c3c6bdff1d3c80a3b0, deletion_comment = False)

CarlosFigueiraMSFT:
>         public static FormulaValue DateDiff(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)

Added this case to DateTimeValue_TimeZone_Seattle.txt. We are getting the time zone service for sub-day subtractions - see below

---
In reply to: [1249141733](https://github.com/microsoft/Power-Fx/pull/659#issuecomment-1249141733) [](http://example.com/codeflow?ancestors=1249141733,1249141733)

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs:126 in a5df486. [](commit_id = a5df486e85eda23f4a4192c3c6bdff1d3c80a3b0, deletion_comment = False)

CarlosFigueiraMSFT:
> >> Date(2022,2,29)

I got this case from Excel :-)

---
In reply to: [1249158708](https://github.com/microsoft/Power-Fx/pull/659#issuecomment-1249158708) [](http://example.com/codeflow?ancestors=1249158708,1249158708)

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Date.txt:284 in a5df486. [](commit_id = a5df486e85eda23f4a4192c3c6bdff1d3c80a3b0, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DateAdd.txt
CarlosFigueiraMSFT: May add in a future PR; the value 1.9 is nice because it shows that:
- If we accept the decimal part, we don't round or truncate
- If we do accept the decimal part, we truncate, and not round up

Ananya-Gupta24 (approved):

PR #674, opened 2022-09-16
Tests for error kinds, move all finite checking to the error handler

CarlosFigueiraMSFT:
New test to validate error kinds for all functions / operations. Also moved all finite checking to the function wrapper instead of having each function declare it


src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUntypedObject.cs
```diff
@@ -76,7 +76,7 @@ public static FormulaValue Text_UO(IRContext irContext, UntypedObjectValue[] arg
                 return new StringValue(irContext, str);
             }
 
-            return CommonErrors.RuntimeTypeMismatch(irContext);
+            return GetTypeMismatchError(irContext, "Text", "String", impl);
```
lesaltzm, line 79: These should probably be `BuiltinFunctionsCore.Text.Name` instead of string literals

lesaltzm: For types, I'd suggest using `DType.<string|number|...>.GetKindString()`

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
```diff
@@ -528,6 +551,65 @@ private static FormulaValue TimeOrDateTime(IRContext irContext, int index, Formu
...
+                if (arg is NumberValue nv && IsInvalidDouble(nv.Value))
+                {
+                    if (errors == null)
+                    {
+                        errors = new List<ErrorValue>();
+                    }
+
+                    errors.Add(new ErrorValue(irContext, new ExpressionError()
+                    {
+                        Message = $"Arguments to the {functionName} function must be finite.",
```
Shpakh-MSFT, line 578: >Arguments [](http://example.com/codeflow?start=36&length=9)

Should the error message indicate which argument? I don't know if we have this information to put in error message here.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
```diff
@@ -528,6 +551,65 @@ private static FormulaValue TimeOrDateTime(IRContext irContext, int index, Formu
...
+                    }));
+                }
+            }
+
+            if (errors != null)
+            {
+                return ErrorValue.Combine(irContext, errors);
+            }
+
+            return null;
```
Shpakh-MSFT, line 590: return errors != null 
            ? ErrorValue.Combine(irContext, errors) 
             : null;

Shpakh-MSFT (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
CarlosFigueiraMSFT: While we could pass this information through (which argument), we should never have a non-finite value flowing into the functions (none of the functions can return it), so I don't think we need to have different error messages for different arguments.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
CarlosFigueiraMSFT: Updated

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -729,26 +784,29 @@ static Library()
...
                     targetFunction: Ln)
             },
             {
                 BuiltinFunctionsCore.Log,
                 StandardErrorHandling<NumberValue>(
+                    BuiltinFunctionsCore.Log.Name,
                     expandArguments: InsertDefaultValues(outputArgsCount: 2, fillWith: new NumberValue(IRContext.NotInSource(FormulaType.Number), 10)),
                     replaceBlankValues: ReplaceBlankWithZero,
                     checkRuntimeTypes: ExactValueTypeOrBlank<NumberValue>,
-                    checkRuntimeValues: StrictNumericPositiveNumberChecker,
```
bryancroteau-MSFT, line 745: Why is this being removed?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -729,26 +784,29 @@ static Library()
             {
                 BuiltinFunctionsCore.Ln,
                 StandardErrorHandling<NumberValue>(
+                    BuiltinFunctionsCore.Ln.Name,
                     expandArguments: NoArgExpansion,
                     replaceBlankValues: ReplaceBlankWithZero,
                     checkRuntimeTypes: ExactValueTypeOrBlank<NumberValue>,
-                    checkRuntimeValues: StrictNumericPositiveNumberChecker,
+                    checkRuntimeValues: DeferRuntimeValueChecking,
```
bryancroteau-MSFT, line 791: We should still do Positive number checks here right?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
CarlosFigueiraMSFT: There is no need for an additional check - the function will return a non-finite number, and the wrapper will create the appropriate error

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
CarlosFigueiraMSFT: There is no need for an additional check - the function will return a non-finite number, and the wrapper will create the appropriate error

bryancroteau-MSFT (approved):

PR #678, opened 2022-09-18
Check invocation result fail prior to get tabular result

CarlosFigueiraMSFT:
A recent change caused an exception to be thrown if an invalid tabular DateAdd function was used. This fixes it and adds a new test to check this scenario.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DateTime.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DateTime.cs
@@ -484,10 +484,13 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
                 // for DateTime alone is sufficient.
                 fValid &= CheckDateColumnType(type0, args[0], errors, ref nodeToCoercedTypeMap);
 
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DateTime.cs
```diff
@@ -484,10 +484,13 @@ public override bool CheckInvocation(TexlBinding binding, TexlNode[] args, DType
...
 
-                var resultColumnType = binding.Features.HasFlag(Features.ConsistentOneColumnTableResult)
-                    ? ColumnName_Value
-                    : type0.GetNames(DPath.Root).Single().Name;
-                returnType = DType.CreateTable(new TypedName(DType.DateTime, resultColumnType));
+                if (fValid)
+                {
+                    var resultColumnType = binding.Features.HasFlag(Features.ConsistentOneColumnTableResult)
+                        ? ColumnName_Value
+                        : type0.GetNames(DPath.Root).Single().Name;
```
CarlosFigueiraMSFT, line 491: If the datecolumntype check above failed, it's possible that the table does not have a single column, so the call to `Single` would throw an exception.

MikeStall (approved):

PR #690, opened 2022-09-23
FormulaValue serializer.

anderson-joyle:
Looking forward to hearing reviewers' comments on top of my questions within the code (**!JYL!** tag).


src/tests/Microsoft.PowerFx.Interpreter.Tests/FormulaValueSerializerTests.cs
```diff
@@ -0,0 +1,150 @@
...
+            var guidValue = FormulaValue.New(newGuid);
+            var stringValue = FormulaValue.New("*SERIALIZER*");
+            var timeValue = FormulaValue.New(dateTimeNow.TimeOfDay);
+
+            Assert.Equal("1", numberValue.ToExpression());
+            Assert.Equal("true", boolValue.ToExpression());
+            Assert.Equal("\"*SERIALIZER*\"", stringValue.ToExpression());
+            Assert.Equal("DateTime(2022,12,25,12,34,56)", dateTimeValue.ToExpression());
+            Assert.Equal("Date(2022,12,25)", dateValue.ToExpression());
+            Assert.Equal("Time(12,34,56)", timeValue.ToExpression());
```
LucGenetier, line 58: I would make this test a [Theory] instead

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
```diff
@@ -181,5 +183,25 @@ public Task<DValue<RecordValue>> UpdateField(string name, FormulaValue value)
...
+
+            foreach (var field in Fields)
+            {
+                sb.Append($"'{field.Name}'");
+                sb.Append(":");
+                sb.Append(field.Value.ToExpression());
+                sb.Append(",");
+            }
+
+            sb.Remove(sb.Length - 1, 1);
```
LucGenetier, line 201: >sb.Remove(sb.Length - 1, 1); [](http://example.com/codeflow?start=12&length=28)

Not a big fan

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateTimeValue.cs
```diff
@@ -23,5 +23,10 @@ public override void Visit(IValueVisitor visitor)
         {
             visitor.Visit(this);
         }
+
+        public override string ToExpression()
+        {
+            return $"DateTime({Value.ToString("yyyy,M,d,H,m,s")})";
```
LucGenetier, line 29: >Value.ToString("yyyy,M,d,H,m,s [](http://example.com/codeflow?start=31&length=30)

This expression depends on the local locale and would result in "2022,9,23,18,46,24" when in French.

You should convert to a neutral format (ISO 8601) and UTC

if (dt.Kind != DateTimeKind.Utc)
 dt = dt.ToUniversalTime();
dt.ToString("o")

This gives "2022-09-23T16:49:02.1901316Z"

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateValue.cs
```diff
@@ -24,5 +24,10 @@ public override void Visit(IValueVisitor visitor)
         {
             visitor.Visit(this);
         }
+
+        public override string ToExpression()
+        {
+            return $"Date({Value.ToString("yyyy,M,d")})";
```
LucGenetier, line 30: >alue.ToString("yyyy,M,d") [](http://example.com/codeflow?start=28&length=25)

Same

src/libraries/Microsoft.PowerFx.Core/Public/Values/ErrorValue.cs
```diff
@@ -68,5 +68,16 @@ private static IEnumerable<ExpressionError> CombineErrors(IEnumerable<ErrorValue
                 }
             }
         }
+
+        public override string ToExpression()
+        {
+            // !JYL!
+            /* Few questions on this:
+             * Should this be serialized?
```
LucGenetier, line 76: >Should this be serialized? [](http://example.com/codeflow?start=15&length=26)

I don't see why we wouldn't

src/libraries/Microsoft.PowerFx.Core/Public/Values/OptionSetValue.cs
```diff
@@ -61,6 +61,12 @@ public string DisplayName
 
                 return logicalName;
             }
-        }        
+        }
+
+        public override string ToExpression()
+        {
+            // !JYL! Do we need to serialize this?
```
LucGenetier, line 68: >/ !JYL! Do we need to serialize this? [](http://example.com/codeflow?start=13&length=37)

Yes :)

src/libraries/Microsoft.PowerFx.Core/Public/Values/GuidValue.cs
```diff
@@ -19,5 +20,10 @@ public override void Visit(IValueVisitor visitor)
         {
             visitor.Visit(this);
         }
+
+        public override string ToExpression()
+        {
+            return $"GUID({CharacterUtils.ToPlainText(Value.ToString("D"))})";
```
LucGenetier, line 26: >"D" [](http://example.com/codeflow?start=69&length=3)

What is this doing?

src/libraries/Microsoft.PowerFx.Core/Public/Values/PrimitiveValue.cs
```diff
@@ -30,5 +30,10 @@ public override object ToObject()
         {
             return _value;
         }
+
+        public override string ToExpression()
+        {
+            return Value.ToString();
```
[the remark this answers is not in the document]
LucGenetier: Do we need to escape strings having quotes, spaces, non printable characters...?

src/libraries/Microsoft.PowerFx.Core/Public/Values/StringValue.cs
```diff
@@ -23,5 +23,10 @@ internal StringValue ToLower()
         {
             return new StringValue(IRContext.NotInSource(FormulaType.String), Value.ToLowerInvariant());
         }
+
+        public override string ToExpression()
+        {
+            return $"\"{Value}\"";
```
LucGenetier, line 29: >$"\"{Value}\""; [](http://example.com/codeflow?start=19&length=15)

What happens if the string contains quotes, spaces, non-printable characters like \t or CR, LF?

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
```diff
@@ -181,5 +183,28 @@ public override void Visit(IValueVisitor visitor)
         {
             visitor.Visit(this);
         }
+
+        public override string ToExpression()
+        {
+            if (Rows.Count() == 0)
+            {
+                throw new InvalidDataContractException("It's not possible to serialize an empty table.");
```
LucGenetier, line 191: >hrow new InvalidDataContractException("It's not possible to serialize an empty table."); [](http://example.com/codeflow?start=17&length=88)

We should support it

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
```diff
@@ -181,5 +183,28 @@ public override void Visit(IValueVisitor visitor)
...
+
+            sb.Append("Table(");
+
+            foreach (var row in Rows)
+            {
+                sb.Append(row.Value.ToExpression());
+                sb.Append(",");
+            }
+
+            sb.Remove(sb.Length - 1, 1);
```
LucGenetier, line 204: >sb.Remove(sb.Length - 1, 1); [](http://example.com/codeflow?start=12&length=28)

Not a big fan

src/libraries/Microsoft.PowerFx.Core/Public/Values/GuidValue.cs
LucGenetier: To make it shorter you can use "n" format

src/libraries/Microsoft.PowerFx.Interpreter/Values/LambdaFormulaValue.cs
```diff
@@ -43,5 +43,11 @@ public override void Visit(IValueVisitor visitor)
         {
             return _tree.Accept(visitor, context);
         }
+
+        public override string ToExpression()
+        {
+            // !JYL! Do we need to serialize this?
+            throw new NotImplementedException("LambdaFormulaValue cannot be serialized.");
```
LucGenetier, line 50: >throw new NotImplementedException("LambdaFormulaValue cannot be serialized."); [](http://example.com/codeflow?start=12&length=78)

We probably should

src/tests/Microsoft.PowerFx.Interpreter.Tests/DatabaseSimulationTests.cs
```diff
@@ -142,6 +142,12 @@ public override void Visit(IValueVisitor visitor)
             {
                 throw new NotImplementedException("TestEntityValue.Visit");
             }
+
+            public override string ToExpression()
+            {
+                // !JYL! Do we need to serialize this?
```
LucGenetier, line 148: >/ !JYL! Do we need to serialize this? [](http://example.com/codeflow?start=17&length=37)

No need for this one

BruceHaley:
:x: 8 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.3-preview.20220920-322403](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.3-preview.20220920-322403).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.Types.CollectionTableValue<T>.AppendAsync(Microsoft.PowerFx.Types.RecordValue)' does not exist in the implementation but it does exist in the contract.
MembersMustExist : Member 'Microsoft.PowerFx.Types.CollectionTableValue<T>.Matches(Microsoft.PowerFx.Types.RecordValue, Microsoft.PowerFx.Types.RecordValue)' does not exist in the implementation but it does exist in the contract.
MembersMustExist : Member 'Microsoft.PowerFx.Types.CollectionTableValue<T>.PatchCoreAsync(Microsoft.PowerFx.Types.RecordValue, Microsoft.PowerFx.Types.RecordValue)' does not exist in the implementation but it does exist in the contract.
MembersMustExist : Member 'Microsoft.PowerFx.Types.RecordValue.UpdateField(System.String, Microsoft.PowerFx.Types.FormulaValue)' does not exist in the implementation but it does exist in the contract.
MembersMustExist : Member 'Microsoft.PowerFx.Types.RecordValue.UpdateFieldsAsync(Microsoft.PowerFx.Types.RecordValue)' does not exist in the implementation but it does exist in the contract.
MembersMustExist : Member 'Microsoft.PowerFx.Types.TableValue.AppendAsync(Microsoft.PowerFx.Types.RecordValue)' does not exist in the implementation but it does exist in the contract.
MembersMustExist : Member 'Microsoft.PowerFx.Types.TableValue.PatchAsync(Microsoft.PowerFx.Types.RecordValue, Microsoft.PowerFx.Types.RecordValue)' does not exist in the implementation but it does exist in the contract.
MembersMustExist : Member 'Microsoft.PowerFx.Types.TableValue.PatchCoreAsync(Microsoft.PowerFx.Types.RecordValue, Microsoft.PowerFx.Types.RecordValue)' does not exist in the implementation but it does exist in the contract.
```

</details>

src/libraries/Microsoft.PowerFx.Core/Public/Values/ErrorValue.cs
MikeStall: agreed, yes, we should serialize
We have an Error() function we can use.  Can you ask Carlos for details - he'll want to review this.

src/libraries/Microsoft.PowerFx.Core/Public/Values/UntypedObjectValue.cs
```diff
@@ -59,5 +59,11 @@ public override void Visit(IValueVisitor visitor)
         {
             visitor.Visit(this);
         }
+
+        public override string ToExpression()
+        {
+            // !JYL! Do we need to serialize this?
```
MikeStall, line 65: >// !JYL! Do we need to serialize this? [](http://example.com/codeflow?start=12&length=38)

yes ...  but this may be tricky (note that  ToObject isn't impl either). 

We may need to defer it to:  IUntypedObject and add a method to that. 

The primary use case for this is ParseJson().

src/libraries/Microsoft.PowerFx.Interpreter/Values/LambdaFormulaValue.cs
MikeStall: this one is tricky.... we can actually probably punt on it for now. 

This would be like arg1 passed into a filter - a lambda callback. 
Since those can't get passed around like regular values, the need to serialize is mitigated.

src/tests/Microsoft.PowerFx.Interpreter.Tests/DatabaseSimulationTests.cs
MikeStall: this is in testing. 
The good news is this is actuall an example for custom serialization :) 
Ie, when we serialize a DataverseRecordValue, we'll want to override and serialize as `Lookup(table, id=Guid(id))` 
. That's an ideal compact form, rather than serialize as a full record.

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValue.cs
```diff
@@ -43,5 +43,7 @@ internal FormulaValue(IRContext irContext)
         public abstract object ToObject();
 
         public abstract void Visit(IValueVisitor visitor);
+
+        public abstract string ToExpression();
```
MikeStall, line 47: > public abstract string ToExpression(); [](http://example.com/codeflow?start=7&length=39)

We should decide whether we just make this ToString() or create a new method.

src/libraries/Microsoft.PowerFx.Core/Public/Values/OptionSetValue.cs
lesaltzm: Yes, would need to be `$"{Type._type.OptionSetInfo.EntityName}.{Option}`, the host would be responsible for making sure the option set is registered under the same name

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
lesaltzm: yes, definitely prefer [String.Join](https://learn.microsoft.com/en-us/dotnet/api/system.string.join?view=net-6.0)

src/libraries/Microsoft.PowerFx.Interpreter/Values/LambdaFormulaValue.cs
lesaltzm: Yeah, this is a hack for the interpreter, and can't meaningfully show up in a place where the host would want to serialize it. notice that this type is also `internal` unlike the other FormulaValues

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
MikeStall: or use a flag to avoid the sb.Append(,) on the first iteration.

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateTimeValue.cs
MikeStall: That's a very good point - I think the challenge is that we don't have a good way to deserialize ISO - which we should.

src/libraries/Microsoft.PowerFx.Core/Public/Values/PrimitiveValue.cs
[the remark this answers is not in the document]
MikeStall: Avoid this method altogether.  Keep it abstract here and force derived classes to handle. 

We just don't know the right answer here and its too easy to get wrong.

src/libraries/Microsoft.PowerFx.Core/Public/Values/StringValue.cs
MikeStall: yes, use one of the existing parser helpers to escape this.

MikeStall: Actually ... this may get very tricky... 
Quotes can be escaped as double quotes.
Newlines can exist as-is!
But other chars would need to be Char() function. 
Like:
"abc\tdef" // C# 
"abc" & Char(9) & "def" // Power Fx

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
MikeStall: Definitely - this was a core thing for Cards.
The challenge is that we don't have a good way to do it today  since we don't have type literals, and Table() is ambiguous. 

This is the same problem Blank() had. 
Need to brainstorm ....

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValue.cs
MikeStall: Also, we probably need the virtual tp accept a TextWriter or StringBuilder. 
And then have a ToString() helper for conveneince that creates the StringBuilder for us and calls that and returns the string,

BruceHaley:
:x: 8 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.3-preview.20220920-322403](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.3-preview.20220920-322403).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.Types.CollectionTableValue<T>.AppendAsync(Microsoft.PowerFx.Types.RecordValue)' does not exist in the implementation but it does exist in the contract.
MembersMustExist : Member 'Microsoft.PowerFx.Types.CollectionTableValue<T>.Matches(Microsoft.PowerFx.Types.RecordValue, Microsoft.PowerFx.Types.RecordValue)' does not exist in the implementation but it does exist in the contract.
MembersMustExist : Member 'Microsoft.PowerFx.Types.CollectionTableValue<T>.PatchCoreAsync(Microsoft.PowerFx.Types.RecordValue, Microsoft.PowerFx.Types.RecordValue)' does not exist in the implementation but it does exist in the contract.
MembersMustExist : Member 'Microsoft.PowerFx.Types.RecordValue.UpdateField(System.String, Microsoft.PowerFx.Types.FormulaValue)' does not exist in the implementation but it does exist in the contract.
MembersMustExist : Member 'Microsoft.PowerFx.Types.RecordValue.UpdateFieldsAsync(Microsoft.PowerFx.Types.RecordValue)' does not exist in the implementation but it does exist in the contract.
MembersMustExist : Member 'Microsoft.PowerFx.Types.TableValue.AppendAsync(Microsoft.PowerFx.Types.RecordValue)' does not exist in the implementation but it does exist in the contract.
MembersMustExist : Member 'Microsoft.PowerFx.Types.TableValue.PatchAsync(Microsoft.PowerFx.Types.RecordValue, Microsoft.PowerFx.Types.RecordValue)' does not exist in the implementation but it does exist in the contract.
MembersMustExist : Member 'Microsoft.PowerFx.Types.TableValue.PatchCoreAsync(Microsoft.PowerFx.Types.RecordValue, Microsoft.PowerFx.Types.RecordValue)' does not exist in the implementation but it does exist in the contract.
```

</details>

src/libraries/Microsoft.PowerFx.Core/Public/Values/OptionSetValue.cs
anderson-joyle: What about $"{Type._type.OptionSetInfo.EntityName}.{**_DisplayName_**}"?

lesaltzm: We always serialize in the invariant locale (logical names), so we'd want to use Option in this case. The host may provide different display names when we're running this, but the logical names would be the same.

src/libraries/Microsoft.PowerFx.Core/Public/Values/TimeValue.cs
```diff
@@ -19,5 +19,10 @@ public override void Visit(IValueVisitor visitor)
         {
             visitor.Visit(this);
         }
+
+        public override string ToExpression()
+        {
+            return $"Time({Value.Hours},{Value.Minutes},{Value.Seconds})";
```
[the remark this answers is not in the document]
anderson-joyle: The number of digits for milliseconds needs to be the same as the input.

BruceHaley:

src/tests/Microsoft.PowerFx.Interpreter.Tests/FormulaValueSerializerTests.cs
anderson-joyle: We changed the approach of how serialization is tested.

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateTimeValue.cs
anderson-joyle: DateTimeValue(ISO value)

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
```diff
@@ -233,5 +233,10 @@ public override int GetHashCode()
         #endregion // Equality
 
         public abstract void Visit(ITypeVisitor vistor);
+
+        internal virtual string DefaultExpressionValue()
```
MikeStall, line 237: >internal virtual string DefaultExpressionValue() [](http://example.com/codeflow?start=8&length=48)

we need this as a form of getting inline types in an expression. 
These expressions may eventually get superceded by the inline type work.

src/libraries/Microsoft.PowerFx.Core/Public/Types/RecordType.cs
```diff
@@ -96,5 +97,30 @@ public static RecordType Empty()
         {
             return new KnownRecordType();
         }
+
+        internal override string DefaultExpressionValue()
```
MikeStall, line 101: >DefaultExpressionValue [](http://example.com/codeflow?start=33&length=22)

This could be

  internal override void DefaultExpressionValue(StringBuilder sb);

that would compose more efficiently

src/libraries/Microsoft.PowerFx.Core/Public/Types/TimeType.cs
```diff
@@ -21,5 +22,12 @@ public override string ToString()
         {
             return "Time";
         }
+
+        internal override string DefaultExpressionValue()
+        {
+            var timeSpanMin = TimeSpan.MinValue;
+
+            return $"Time({timeSpanMin.Hours},{timeSpanMin.Minutes},{timeSpanMin.Seconds},{timeSpanMin.Milliseconds})";
```
MikeStall, line 30: >Time [](http://example.com/codeflow?start=21&length=4)

These could all be string constants.

src/libraries/Microsoft.PowerFx.Core/Public/Values/BlankValue.cs
```diff
@@ -28,5 +28,17 @@ public override void Visit(IValueVisitor visitor)
...
             visitor.Visit(this);
         }
+
+        public override string ToExpression()
+        {
+            if (Type is BlankType)
+            {
+                return Type.DefaultExpressionValue();
+            }
+            else
```
MikeStall, line 38: can we remove this special check?

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
```diff
@@ -24,7 +24,7 @@ public abstract class CollectionTableValue<T> : TableValue
         // Additional capabilities. 
         private readonly IReadOnlyList<T> _sourceIndex; // maybe null. supports index. 
         private readonly IReadOnlyCollection<T> _sourceCount; // maybe null. supports count;
-        private readonly ICollection<T> _sourceList;
+        private readonly ICollection<T> _sourceList; // maybe null. supports multation;
```
MikeStall, line 27: >multation [](http://example.com/codeflow?start=77&length=9)

typo?

src/libraries/Microsoft.PowerFx.Core/Public/Values/ErrorValue.cs
```diff
@@ -68,5 +72,40 @@ private static IEnumerable<ExpressionError> CombineErrors(IEnumerable<ErrorValue
                 }
             }
         }
+
+        public override string ToExpression()
```
MikeStall, line 76: >    public override string ToExpression() [](http://example.com/codeflow?start=4&length=41)

This is an example of why this overload would be good:
 public override void ToExpression(StringBuilder sb)

And then we can have a convenience method:
public string ToExpression()
{
 var sb = new StringBuilder();
 this.ToExpression(sb);
 return sb.ToString();
}

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
```diff
@@ -213,6 +216,30 @@ public virtual async Task<DValue<RecordValue>> UpdateFieldsAsync(RecordValue cha
...
+            foreach (var field in Fields)
+            {
+                if (!flag)
+                {
+                    sb.Append(",");
+                }
+
+                flag = false;
+
+                sb.Append($"'{field.Name}':{field.Value.ToExpression()}");
```
MikeStall, line 237: >{field.Name} [](http://example.com/codeflow?start=29&length=12)

also need to escape

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
```diff
@@ -190,5 +193,36 @@ public override void Visit(IValueVisitor visitor)
         {
             visitor.Visit(this);
         }
+
+        public override string ToExpression()
+        {
+            if (!Rows.Any())
+            {
```
MikeStall, line 200: suggested comment:

// Table() is not legal, so we need an alternate expression to capture the table's type

src/libraries/Microsoft.PowerFx.Interpreter/Values/LambdaFormulaValue.cs
MikeStall: Nit - I'd expand the comment here to explain why we don't impl this.

MikeStall:
>     public class StringType : FormulaType

We should be able to add a good generic test here:


for each T in StringType, NumberType, etc...
(any non-aggregate)

var expr = new T().DefaultExpressionValue();
var check = engine.check(expr);
Assert.Equals(check.ReturnType , T);

See PrimitiveTests.TestAll




---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/Types/StringType.cs:9 in 1334f62. [](commit_id = 1334f622ac930ca826bbc88e6636009e7cff05ce, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Public/Values/BlankValue.cs
anderson-joyle: Without it we get a weird type: **Blank() ({?})**

BruceHaley:

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
```diff
@@ -233,5 +234,19 @@ public override int GetHashCode()
         #endregion // Equality
 
         public abstract void Visit(ITypeVisitor vistor);
+
+        internal virtual void DefaultExpressionValue(StringBuilder sb)
+        {
+            throw new NotSupportedException();
```
MikeStall, line 240: >NotSupportedException [](http://example.com/codeflow?start=22&length=21)

nit:
we can improve the error message here by including the derived type. 

$"{this.GetType().FullName} doesn't implement DefaultExpressionValue"

src/libraries/Microsoft.PowerFx.Core/Public/Values/BlankValue.cs
MikeStall: Wouldn't it just be:
If(false, Blank())

Does that have a different type than:
Blank()?

src/libraries/Microsoft.PowerFx.Core/Public/Values/ErrorValue.cs
```diff
@@ -68,5 +72,36 @@ private static IEnumerable<ExpressionError> CombineErrors(IEnumerable<ErrorValue
                 }
             }
         }
+
+        public override void ToExpression(StringBuilder sb)
+        {            
+            var concat = string.Empty;
```
MikeStall, line 78: >concat [](http://example.com/codeflow?start=16&length=6)

unused?

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValue.cs
```diff
@@ -43,5 +44,16 @@ internal FormulaValue(IRContext irContext)
         public abstract object ToObject();
 
         public abstract void Visit(IValueVisitor visitor);
+
+        public abstract void ToExpression(StringBuilder sb);
+
+        public string ToExpression()
```
MikeStall, line 50: >public string ToExpression() [](http://example.com/codeflow?start=8&length=28)

Can we add an xml comment?

/// Provides serialization. Return an expression that, when evaluated, recreates an equivalent formula value, including its type. 
/// This may not be the same as the expression used to originally create this type.

src/libraries/Microsoft.PowerFx.Core/Public/Types/RecordType.cs
```diff
@@ -96,5 +97,29 @@ public static RecordType Empty()
...
+            foreach (var field in GetFieldTypes())
+            {
+                if (!flag)
+                {
+                    sb.Append(",");
+                }
+
+                flag = false;
+                
+                // !JYL! Check how to escape field name
```
MikeStall, line 116: > // !JYL! Check how to escape field name [](http://example.com/codeflow?start=15&length=40)

todo

MikeStall: The fact that we still pass means that we have a test gap in the .txts. 
Can we add a test that hits this failure, and then we can fix it.

BruceHaley:

MikeStall:
> >> {'abc \ " \0 \b \t \n \v \f \r \u0085 \u2028 \u2029def':1}

This is great. 

Can we also add a test that accesses the field:
>> {'abc \ " \0 \b \t \n \v \f \r \u0085 \u2028 \u2029def':1}.'abc \ " \0 \b \t \n \v \f \r \u0085 \u2028 \u2029def'
1


---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Escaping.txt:1 in 1cdf334. [](commit_id = 1cdf3345019dee94e6a3b22fdb314d848dc91e79, deletion_comment = False)

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -257,7 +261,25 @@ protected override async Task<RunResult> RunAsyncInternal(string expr, string se
...
+                    return new RunResult(newValue);
+                }
+
+                // Get serialized expression. DateTimeValue needs timezone information.
+                var serializedExpression = newValue is DateTimeValue dtv ? dtv.ToExpression(iSetup.TimeZoneInfo) : newValue.ToExpression();
+
+                // Serialization test. Serialized expression must produce an identical result.
+                var checkSerialized = engine.Check(serializedExpression, parameters.Type, options: iSetup.Flags.ToParserOptions());
+                if (!check.IsSuccess)
+                {
```
MikeStall, line 276: How could this fail?
Why do we need to set parser options?

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateTimeValue.cs
```diff
@@ -22,5 +24,22 @@ public override void Visit(IValueVisitor visitor)
         {
             visitor.Visit(this);
         }
+
+        public override void ToExpression(StringBuilder sb)
+        {
+            sb.Append($"DateTimeValue({CharacterUtils.ToPlainText(Value.ToString("o"))})");
+        }
+
+        public string ToExpression(TimeZoneInfo timezoneInfo)
```
MikeStall, line 33: > public string ToExpression(TimeZoneInfo timezoneInfo) [](http://example.com/codeflow?start=7&length=54)

Who calls this?

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
anderson-joyle: The idea is to run the serialized expression under the very same scenario when the original eval was performed.

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -257,7 +261,25 @@ protected override async Task<RunResult> RunAsyncInternal(string expr, string se
...
 
-                return new RunResult(newValue);
+                // UntypedObjectType type is currently not supported for serialization.
+                if (newValue.Type is UntypedObjectType)
+                {
+                    return new RunResult(newValue);
+                }
+
+                // Get serialized expression. DateTimeValue needs timezone information.
+                var serializedExpression = newValue is DateTimeValue dtv ? dtv.ToExpression(iSetup.TimeZoneInfo) : newValue.ToExpression();
```
MikeStall, line 271: This will have issues if the DTV is nested in  a record.
Let's discuss.

src/libraries/Microsoft.PowerFx.Core/Public/Types/RecordType.cs
anderson-joyle: Added test cases for this.

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Public/Config/FormulaValueSerializerSettings.cs
```diff
@@ -0,0 +1,16 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.PowerFx
+{
+    /// <summary>
+    /// Provide settings to FormulaValues in order to serialize values in different situations.
+    /// </summary>
+    public class FormulaValueSerializerSettings
```
MikeStall, line 13: >FormulaValueSerializerSettings [](http://example.com/codeflow?start=17&length=30)

Adding now to avoid breaking changes. Some potential future settings:
            // Max Depth?  
            // Overrides.   
            // .txt ... short form... "Blank()"  instead of "If(false,1)"
            // Pretty printing ... indenting?

src/libraries/Microsoft.PowerFx.Core/Public/Values/ColorValue.cs
```diff
@@ -19,5 +20,10 @@ public override void Visit(IValueVisitor visitor)
         {
             visitor.Visit(this);
         }
+
+        public override void ToExpression(StringBuilder sb, FormulaValueSerializerSettings settings)
+        {
+            throw new System.NotImplementedException();
```
MikeStall, line 26: >throw new System.NotImplementedException(); [](http://example.com/codeflow?start=12&length=43)

We can followup in future PR, - this could be an RGB expression. 
(we should hit this with future tests in color.txt)

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateTimeValue.cs
MikeStall: After discussion, I think just naively serializing the DateTime is sufficient.  That will include the timezone specifier and will deserialize faithfully. 

Our primary serialization scenario is deserializing across multiple requests to a server - so the target timezone will be constant.

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Values/BlankValue.cs
anderson-joyle: The following test cases illustrate this scenario:
image

src/libraries/Microsoft.PowerFx.Core/Public/Values/ColorValue.cs
anderson-joyle: Created issue #723 to track this pendency.

BruceHaley:

src/tests/Microsoft.PowerFx.Interpreter.Tests/SerializationTests.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    public class SerializationTests : PowerFxTest
+    {
+        [Fact]
+        public void DateTimeSerializationTests()
+        {
+            var engine = new RecalcEngine(new PowerFxConfig());
+
+            // Local datetime
+            var now = DateTime.Now;
```
MikeStall, line 22: > var now = DateTime.Now; [](http://example.com/codeflow?start=11&length=24)

super nit - can we put this in a for-loop, so we reuse lines 23...26

MikeStall (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Values/BlankValue.cs
MikeStall: Can we add that as a comment? 
// if(False, Blank()) actually generates an unknown type, not blank type.

BruceHaley:

MikeStall (approved):
:shipit:

PR #700, opened 2022-09-28
Rename ThisItem to ThisRecord in ForAll_UO

bryancroteau-MSFT:
Rename ThisItem to ThisRecord for ForAll_UO, because of a conversation with Greg today.


CarlosFigueiraMSFT (approved):
:shipit:

PR #702, opened 2022-09-28
Make CultureInfo of ParserOptions public

### LINKED ISSUE #687: EvalAsync() with invariant formulas
```EvalAsync()``` method from ```RecalEngine``` class should be able to eval invariant formulas.

Repro step :
- Set **fr-FR** culture from ```PowerFxConfig``` of ```RecalcEngine```
- Write ```Concatenate("abc";"def")``` formula
- Store this formula into invariant format that corresponds to ```Concatenate("abc","def")``` (```,``` insteand of ```;```)
- ```EvalAsync()``` of the invariant formula gives the following error :  Errors: Error 17-18: Caractères inattendus. La formule contient « ParenClose » alors que « Error » est attendu.
Error 17-18: Caractères inattendus. Des caractères sont utilisés dans la formule de manière inattendue. 

I don't want to use two engines with different culture or set a culture before every evaluation.

Maybe you could add an invariant flag into the ```EvalAsync()``` method or add a new metohd ```EvalInvariantAsync()```.

ahestin:
Fix #687

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Parser/ParserOptions.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Parser/ParserOptions.cs
@@ -20,7 +20,7 @@ namespace Microsoft.PowerFx
         /// </summary>
         public bool AllowsSideEffects { get; set; }
                 
-        internal CultureInfo Culture { get; set; }
+        public CultureInfo Culture { get; set; }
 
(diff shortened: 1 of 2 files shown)


ahestin:
@microsoft-github-policy-service agree company="divalto"

MikeStall (approved):

PR #705, opened 2022-09-28
Demonstrate creative use of symbol tables

jorisdg:
Demonstrate creative use of symbol tables to make a reflection function with a dynamic return type

### DIFF
--- a/src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/ScenarioSymbolTables.cs
+++ b/src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/ScenarioSymbolTables.cs
@@ -1,0 +1,192 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Core.Tests;
+using Microsoft.PowerFx.Syntax;
+using Microsoft.PowerFx.Types;
+using Xunit;
+using Xunit.Sdk;
+
+namespace Microsoft.PowerFx.Interpreter.Tests.Scenarios
+{
+    // Demonstrate creative use of symbol tables to make a reflection function with a dynamic return type
+    public class ScenarioSymbolTables : PowerFxTest
+    {
+        [Theory]
+        [InlineData("FindRecord(\"Customer\")", typeof(CustomerTable))]
+        [InlineData("FindRecord(\"Vendor\")", typeof(VendorTable))]
+        public void DynamicRecordTypeChecking(string expr, Type expectedType)
+        {
+            var typeCache = new TypeMarshallerCache();
+
+            var config = new PowerFxConfig();
+            config.AddFunction(new FindRecord(typeCache));
+
+            var engine = new RecalcEngine(config);
+
+            var parseResult = engine.Parse(expr);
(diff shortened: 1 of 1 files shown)


src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/ScenarioSymbolTables.cs
```diff
@@ -0,0 +1,192 @@
...
+
+            public override void PostVisit(VariadicOpNode node)
+            {
+            }
+
+            public override void PostVisit(CallNode node)
+            {
+                if (node.Head.Name.Equals("FindRecord"))
+                {
+                    _recordTypes.Add((node.Args.ChildNodes.First() as StrLitNode).Value);
```
lesaltzm, line 139: This could be a null ref if the first arg is not a string literal.

src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/ScenarioSymbolTables.cs
```diff
@@ -0,0 +1,192 @@
...
+            var findRecords = new FindRecordVisitor();
+            parseResult.Root.Accept(findRecords);
+
+            var symbolTable = new SymbolTable();
+            foreach (var tableName in findRecords.RecordTypes)
+            {
+                var record = BaseRecord.CreateRecord(tableName);
+                var recordValue = typeCache.Marshal(record, record.GetType()) as RecordValue;
+
+                symbolTable.AddFunction(new FindRecord(typeCache, recordValue.Type));
```
lesaltzm, line 42: Does this work correctly if there's more than one FindRecord call in a single expression with different args? Our overload resolution is based on types, and I THINK this would always pick the first FindRecord function registered each time

src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/ScenarioSymbolTables.cs
jorisdg: Yes indeed. As you mentioned, there's no way to set the type for the reflection function arg is a literal, only FormulaType.String...

src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/ScenarioSymbolTables.cs
jorisdg: That was my assumption that it would work, but I haven't actually tried...

src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/ScenarioSymbolTables.cs
lesaltzm: Right, at the very least, we should check and just return, instead of throwing a nullref, though.

BruceHaley:

src/tests/Microsoft.PowerFx.Interpreter.Tests/Scenarios/ScenarioSymbolTables.cs
jorisdg: I tried, and it doesn't work...
`FindRecord(\"Vendor\").VendorId; FindRecord(\"Customer\").CustomerId;`
results in:
"Name isn't valid. 'CustomerId' isn't recognized."

BruceHaley:

BruceHaley:

PR #716, opened 2022-09-30
Display name fix.

MikeStall:
DisplayName looks were failing when the record came from dataverse.  

Avoid calling TryGetConvertedDisplayNameAndLogicalNameForColumn, it doesn't work on dataverse entities. 
Call TryGetDisplayNameForColumn instead, which is conveniently used by RecordType.TryGetFieldType. 

Also include helper for setting display names on symbol values.


src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -2434,6 +2434,7 @@ internal static bool TryGetLogicalNameForColumn(DType type, string displayName,
         }
 
         /// <summary>
+        /// This API is very specific for Canvas. Don't call it unless you know exactly what you're doing.
```
lesaltzm, line 2437: Good to add haha

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTableOverRecordType.cs
```diff
@@ -40,31 +40,26 @@ internal override bool TryLookup(DName name, out NameLookupInfo nameInfo)
         {
             // Lookup must handle display names
             // This is consistent with RowScope.
```
lesaltzm, line 42: Do we still need this comment, if we've pushed the handling for display names into TryGetFieldType?

lesaltzm (approved):

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTableOverRecordType.cs
MikeStall: There's so much confusion around all this, better to be explicit.

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Interpreter** compared against [version 0.2.3-preview.20220930-323792](https://www.nuget.org/packages/Microsoft.PowerFx.Interpreter/0.2.3-preview.20220930-323792).
<details>

```
MembersMustExist : Member 'Microsoft.PowerFx.ReadOnlySymbolValues.NewRowScope(Microsoft.PowerFx.Types.RecordValue, Microsoft.PowerFx.ReadOnlySymbolValues)' does not exist in the implementation but it does exist in the contract.
```

</details>

PR #740, opened 2022-10-07
Draft PR. Switching to different languages during runtime and within the same process.

anderson-joyle:


src/tests/Microsoft.PowerFx.Interpreter.Tests/SerializationTests.cs
```diff
@@ -32,7 +33,25 @@ public void DateTimeSerializationTests()
...
+
+            var ptBR_opt = new ParserOptions() { Culture = new CultureInfo("pt-BR") };
+            var frFR_opt = new ParserOptions() { Culture = new CultureInfo("fr-FR") };
+            var enGB_opt = new ParserOptions() { Culture = new CultureInfo("en-GB") };
+
+            var ptBR_check = engine.Check(expression, options: ptBR_opt);
+            var frFR_check = engine.Check(expression, options: frFR_opt);
+            var enGB_check = engine.Check(expression, options: enGB_opt);
+
+            Assert.True(ptBR_check.IsSuccess);
```
MikeStall, line 54: >ptBR_check.IsSuccess); [](http://example.com/codeflow?start=24&length=22)

Success?

src/tests/Microsoft.PowerFx.Interpreter.Tests/SerializationTests.cs
```diff
@@ -32,7 +33,25 @@ public void DateTimeSerializationTests()
...
+        {
+            var engine = new RecalcEngine(new PowerFxConfig());
+
+            var expression = "If(true,";
+
+            var ptBR_opt = new ParserOptions() { Culture = new CultureInfo("pt-BR") };
+            var frFR_opt = new ParserOptions() { Culture = new CultureInfo("fr-FR") };
+            var enGB_opt = new ParserOptions() { Culture = new CultureInfo("en-GB") };
+
+            var ptBR_check = engine.Check(expression, options: ptBR_opt);
```
MikeStall, line 50: >ptBR_check [](http://example.com/codeflow?start=16&length=10)

Can we add a test that looks at the error message in the various locales, and verifies it's actually translated. 

That's the most critical customer-facing part.

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -56,8 +57,15 @@ private TexlParser(IReadOnlyList<Token> tokens, Flags flags)
             _depth = 0;
             _curs = new TokenCursor(tokens);
             _flagsMode = new Stack<Flags>();
-            _flagsMode.Push(flags);
-        }
+            _flagsMode.Push(flags);
+            _locale = CultureInfo.CurrentCulture;
```
MikeStall, line 61: > CultureInfo.CurrentCulture; [](http://example.com/codeflow?start=21&length=28)

we're trying to remove calls to CultureInfo.CurrentCulture :) 

Parser is conveniently internal - so we can change the signature. Can we make the param required and avoid a call to CultureInfo.CurrentCulture;

src/libraries/Microsoft.PowerFx.Core/Errors/BaseError.cs
```diff
@@ -173,7 +178,7 @@ private string FormatMessage(string message, params object[] args)
             {
                 try
                 {
-                    sb.AppendFormat(CultureInfo.CurrentCulture, message, args);
```
MikeStall, line 176: >CultureInfo.CurrentCulture [](http://example.com/codeflow?start=36&length=26)

Great to remove!

src/libraries/Microsoft.PowerFx.Core/Errors/BaseError.cs
```diff
@@ -111,12 +111,17 @@ internal abstract class BaseError : IDocumentError
...
         private static readonly string HowToFixSuffix = "_HowToFix";
 
+        internal BaseError(IDocumentError innerError, Exception internalException, DocumentErrorKind kind, DocumentErrorSeverity severity, ErrorResourceKey errKey, CultureInfo locale, params object[] args)
+            : this(innerError, internalException, kind, severity, errKey, textSpan: null, sinkTypeErrors: null, locale, args: args)
+        {
+        }
+
         internal BaseError(IDocumentError innerError, Exception internalException, DocumentErrorKind kind, DocumentErrorSeverity severity, ErrorResourceKey errKey, params object[] args)
-            : this(innerError, internalException, kind, severity, errKey, textSpan: null, sinkTypeErrors: null, args: args)
+            : this(innerError, internalException, kind, severity, errKey, textSpan: null, sinkTypeErrors: null, CultureInfo.CurrentCulture, args: args)
```
MikeStall, line 120: >CultureInfo.CurrentCulture [](http://example.com/codeflow?start=112&length=26)

can we avoid this - and require the updates?

src/libraries/Microsoft.PowerFx.Core/Errors/BaseError.cs
```diff
@@ -141,7 +146,7 @@ internal BaseError(IDocumentError innerError, Exception internalException, Docum
             // that haven't yet been converted to an ErrorResource in the Resources.pares file.
             string shortMessage;
             string longMessage;
-            if (!StringResources.TryGetErrorResource(errKey, out var errorResource))
+            if (!StringResources.TryGetErrorResource(errKey, out var errorResource, locale.Name))
```
MikeStall, line 149: >, locale.Name) [](http://example.com/codeflow?start=82&length=14)

It looks like this is the critical fix.

src/tests/Microsoft.PowerFx.Interpreter.Tests/SerializationTests.cs
anderson-joyle: Just testing a few things. This test case will be changed/removed.

src/libraries/Microsoft.PowerFx.Core/Errors/BaseError.cs
anderson-joyle: Not all. This is also being called from ErrorContainer which doesn't contain culture information.
https://github.com/microsoft/Power-Fx/blob/60838bf8b7024a15402ae44fdbf0f3f30a80db5e/src/libraries/Microsoft.PowerFx.Core/App/ErrorContainers/ErrorContainer.cs#L114

src/tests/Microsoft.PowerFx.Interpreter.Tests/SerializationTests.cs
anderson-joyle: These are temporary tests. The real ones are in the src/tests/Microsoft.PowerFx.Core.Tests/ParseTests.cs file.

Issue #728, opened 2022-10-04
[Serializer] Apply serializer to .txt tests


MikeStall:
Today, testing has several FormulaValue to string serializers.  These should use the new expression serializer in the product. 

**Key features here:**
- Note that we may need a "Compact" mode here, since we want these .txt to be readable. So serialize Blank() as `Blank()` (losing type info) rather than `If(false, 1)`. 
- They also need to be canonical - serialize the record fields in deterministic (ie, alphabetical order)

[1] .txt tests:
https://github.com/microsoft/Power-Fx/blob/6334544bcbe1a7750d96a3f5e81077b6663dc3fa/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/TestRunner.cs#L254

[2] Tests
https://github.com/microsoft/Power-Fx/blob/07eed1bfe117735b6149167d763858dfb5100bc4/src/tests/Microsoft.PowerFx.Interpreter.Tests/ValueTests.cs#L349

Other usages - these each also have a fully copy.  
[3] Console REPL 

[4] Web Demo.

PR #742, opened 2022-10-07
Updated serialization to properly escape record property values

CarlosFigueiraMSFT:
The record serialization was escaping property names incorrectly. This fixes it.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
@@ -12,6 +12,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.PowerFx.Core.IR;
 using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Syntax;
 
 namespace Microsoft.PowerFx.Types
 {
@@ -233,7 +234,8 @@ namespace Microsoft.PowerFx.Types
 
                 flag = false;
 
-                sb.Append($"'{CharacterUtils.Escape(field.Name)}':");
+                sb.Append(IdentToken.MakeValidIdentifier(field.Name));
+                sb.Append(':');
 
                 field.Value.ToExpression(sb, settings);
             }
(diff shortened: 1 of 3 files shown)


src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
```diff
@@ -233,7 +234,8 @@ public override void ToExpression(StringBuilder sb, FormulaValueSerializerSettin
 
                 flag = false;
 
-                sb.Append($"'{CharacterUtils.Escape(field.Name)}':");
+                sb.Append(IdentToken.MakeValidIdentifier(field.Name));
```
anderson-joyle, line 237: >IdentToken.MakeValidIdentifier [](http://example.com/codeflow?start=26&length=30)

Should this be called from a helper (CharacterUtils?) so we have a centralized place to handle these scenarios?

CarlosFigueiraMSFT: CharacterUtils is about escaping string values. IdentToken is the place where we deal with identifiers (which is the case for record property names). I agree it's confusing, we can consider unifying them all in a single place in the future.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/TestRunner.cs
```diff
@@ -345,7 +346,8 @@ internal static void TestToString(FormulaValue result, StringBuilder sb)
                 foreach (var field in fields)
                 {
                     sb.Append(dil);
-                    sb.Append(field.Name);
+                    var escapedName = IdentToken.MakeValidIdentifier(field.Name);
+                    sb.Append(escapedName);
                     sb.Append(':');
```
MikeStall, line 351: ah, a bug in the test code!
Another reason to do:  #728

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
MikeStall: Bonus that IdentToken.MakeValidIdentifier is public, so derived classes can call it.

MikeStall (approved):
:shipit:

lesaltzm (approved):

Issue #698, opened 2022-09-27
Disable row-scope disambiguation syntax for C# hosts


lesaltzm:
`table[@Field]` instead of As is currently still supported in the C# host. We should remove this support before shipping V1. 

Now, user would need to use
`Filter(A, ThisRecord.Value = 2)` or `Filter(A As Foo, Foo.Value = 2)`
instead of
`Filter(A, A[@Value] = 2)`

lesaltzm:
![image](https://user-images.githubusercontent.com/69215460/192655374-e0b93223-60bf-46fc-ad48-25a34f6639bc.png)

lesaltzm:
Luckily this is easy to clean up. returning early from Binder.cs::IsRowScopeAlias() based on a flag would disable this.

MikeStall:
Note that _just_ returning false from IsRowScopeAlias will cause us to hit other cases and still attempt to bind the expressions. So we should at just give an error if IsRowScopeAlias  is true. 

Or better yet - IsRowScopeAlias is at the binder - can we disable at the parser level?

Should also confirm that **intellisense _doesn't_ suggest it.** 

![image](https://user-images.githubusercontent.com/1538900/194601161-8c97d2eb-f36f-4d13-acf7-46e577ef592a.png)

jas-valgotar:
Do we also disable it in the case of unqualified field name?
e.g. Filter([1,2,3],ThisRecord[@Value]=2)

PR #747, opened 2022-10-11
Make `Identifier.HasAtToken` public

ivanradicek:


MikeStall:
fyi, another case of @ is #698.

lesaltzm (approved):

Issue #731, opened 2022-10-04
[Serializer] OptionSetValueType.DefaultExpressionValue


MikeStall:
OptionSetValueType.DefaultExpressionValue isn't implemented. 
We hit this trying to serialize blank dataverse records.

MikeStall:
See #761  - that has some tests that should be enabled once this is ready. 

https://github.com/microsoft/Power-Fx/blob/854158e1025b886868b837cf07734a99068026b3/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Sort.txt#L15

PR #761, opened 2022-10-13
Allow option set columns to be used with Sort and SortByColumns

sagarbakliwal:
Dataverse allows Option set (Choice) columns to be sorted. Only limitation is multiselect option sets (Choices) which are registered as Table and won't be allowed in Sort and SortByColumns

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Sort.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Sort.cs
@@ -62,7 +62,7 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
             returnType = argTypes[0];
 
             var exprType = argTypes[1];
-            if (!exprType.IsPrimitive || exprType.IsOptionSet)
+            if (!exprType.IsPrimitive)
             {
                 fValid = false;
                 errors.EnsureError(args[1], TexlStrings.ErrSortWrongType);
@@ -210,6 +210,9 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
 
             switch (type.Kind)
             {
+                case DKind.OptionSet:
+                case DKind.OptionSetValue:
+                    return "3";
                 case DKind.Boolean:
                     return "0";
                 case DKind.Number:
@@ -243,7 +246,7 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
 
             foreach (var column in allColumns)
             {
-                if (column.Type.IsPrimitive && !column.Type.IsOptionSet)
+                if (column.Type.IsPrimitive)
                 {
                     primitiveColumnsAndComparatorIds.AppendFormat(
                         "{0}\"{1}\":{2}",
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Sort.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Sort.cs
@@ -204,64 +204,6 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
             return true;
         }
 
-        public static string GetSortComparatorIdForType(DType type)
-        {
-            Contracts.AssertValid(type);
-
-            switch (type.Kind)
-            {
-                case DKind.OptionSet:
-                case DKind.OptionSetValue:
-                    return "3";
-                case DKind.Boolean:
-                    return "0";
-                case DKind.Number:
-                case DKind.Color:
-                case DKind.Currency:
-                case DKind.Date:
-                case DKind.Time:
-                case DKind.DateTime:
-                    return "1";
-                case DKind.String:
-                case DKind.Image:
-                case DKind.PenImage:
-                case DKind.Hyperlink:
-                case DKind.Media:
-                case DKind.Blob:
-                default:
-                    Contracts.Assert(DType.String.Accepts(type, exact: false));
-                    return "2";
-            }
-        }
-
-        internal static string GenerateColumnNamesMappingForSortByColumns(DType sourceType)
-        {
-            Contracts.Assert(sourceType.IsTable);
-
-            var allColumns = sourceType.GetNames(DPath.Root);
-            var separator = string.Empty;
-
-            var primitiveColumnsAndComparatorIds = new StringBuilder();
-            primitiveColumnsAndComparatorIds.Append("{");
-
-            foreach (var column in allColumns)
-            {
-                if (column.Type.IsPrimitive)
-                {
-                    primitiveColumnsAndComparatorIds.AppendFormat(
-                        "{0}\"{1}\":{2}",
-                        separator,
-                        CharacterUtils.EscapeString(column.Name),
-                        GetSortComparatorIdForType(column.Type));
-                    separator = ",";
-                }
-            }
-
-            primitiveColumnsAndComparatorIds.Append("}");
-
-            return primitiveColumnsAndComparatorIds.ToString();
-        }
-
         private bool IsSortOrderSuppportedByColumn(TexlNode node, TexlBinding binding, string order, SortOpMetadata metadata, DPath columnPath)
         {
             Contracts.AssertValue(node);
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
@@ -173,10 +173,10 @@ namespace Microsoft.PowerFx.Functions
 
             if (arg0 is TableValue table)
             {
-                var errors = table.Rows.Where(r => r.IsError).Select(r => r.Error);
-                if (errors.Any())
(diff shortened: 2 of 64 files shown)


lesaltzm:
Option sets have support in the interpreter, please add *.txt test cases covering the semantics of sorting on option set values.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Sort.cs
```diff
@@ -210,6 +210,9 @@ public static string GetSortComparatorIdForType(DType type)
 
             switch (type.Kind)
```
lesaltzm, line 211: Not necessarily as part of this change, but this should be pulled back to PA-Client, it only affects JSTranslator's impl of Sort/SortByColumns.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -441,10 +442,14 @@ public static async ValueTask<FormulaValue> SortTable(EvalVisitor runner, EvalVi
             {
                 return SortValueType<DateTimeValue, DateTime>(pairs, irContext, compareToResultModifier);
             }
-            else
+            else if (allDates)
             {
                 return SortValueType<DateValue, DateTime>(pairs, irContext, compareToResultModifier);
             }
+            else
```
lesaltzm, line 449: else if (allOptionSets)?
And return an ErrorValue of some sort if no conditions match

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -476,6 +481,27 @@ private static bool IsValueTypeErrorOrBlank<T>(FormulaValue val)
             return new InMemoryTableValue(irContext, pairs.Select(pair => pair.row));
         }
 
+        private static FormulaValue SortOtionSet(List<(DValue<RecordValue> row, FormulaValue sortValue)> pairs, IRContext irContext, int compareToResultModifier)
```
lesaltzm, line 484: sp: SortOptionSet

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Sort.txt
```diff
@@ -1,4 +1,12 @@
-﻿>> Sort([-2, -1, 5, 1, 2], Value)
+﻿#SETUP: OptionSetSortTestSetup
+
+>> Sort(t1, OptionSetField1, SortOrder.Ascending)
+Table({OptionSetField1:OptionSet.option_1},{OptionSetField1:OptionSet.option_2})
```
lesaltzm, line 4: could you add a test with multiple rows that have the same option set value?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Sort.cs
sagarbakliwal: Included it as part of this change and added it to PA-Client pull request

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -56,6 +57,46 @@ private static (RecalcEngine engine, RecordValue parameters) OptionSetTestSetup(
...
+            optionSet.TryGetValue(new DName("option_1"), out var o1Val);
+            optionSet.TryGetValue(new DName("option_2"), out var o2Val);
+
+            var r1 = FormulaValue.NewRecordFromFields(new NamedValue("OptionSetField1", o1Val), new NamedValue("StrField1", FormulaValue.New("test1")));
+            var r2 = FormulaValue.NewRecordFromFields(new NamedValue("OptionSetField1", o2Val), new NamedValue("StrField1", FormulaValue.New("test2")));
+            var r3 = FormulaValue.NewRecordFromFields(new NamedValue("OptionSetField1", o1Val), new NamedValue("StrField1", FormulaValue.New("test3")));
+            var r4 = FormulaValue.NewRecordFromFields(new NamedValue("OptionSetField1", o2Val), new NamedValue("StrField1", FormulaValue.New("test4")));
+            
+            // Testing with missing/blank option set field is throwing an exception
+            var r5 = FormulaValue.NewRecordFromFields(new NamedValue("StrField1", FormulaValue.New("test5")));
```
sagarbakliwal, line 79: @MikeStall record r5 is missing option set field and it throws an exception Microsoft.PowerFx.Types.OptionSetValueType doesn't implement DefaultExpressionValue

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Sort.txt
```diff
@@ -1,4 +1,24 @@
...
+>> Sort(t1, OptionSetField1, SortOrder.Descending)
+Table({OptionSetField1:OptionSet.option_2,StrField1:"test2"},{OptionSetField1:OptionSet.option_1,StrField1:"test1"})
+
+>> Sort(t2, OptionSetField1, SortOrder.Ascending)
+Table({OptionSetField1:OptionSet.option_1,StrField1:"test1"},{OptionSetField1:OptionSet.option_1,StrField1:"test3"},{OptionSetField1:OptionSet.option_2,StrField1:"test2"},{OptionSetField1:OptionSet.option_2,StrField1:"test4"})
+
+>> Sort(t2, OptionSetField1, SortOrder.Descending)
+Table({OptionSetField1:OptionSet.option_2,StrField1:"test2"},{OptionSetField1:OptionSet.option_2,StrField1:"test4"},{OptionSetField1:OptionSet.option_1,StrField1:"test1"},{OptionSetField1:OptionSet.option_1,StrField1:"test3"})
+
+>> Sort(t3, OptionSetField1, SortOrder.Ascending)
```
sagarbakliwal, line 15: @MikeStall  This test fails due to blank option set issue

sagarbakliwal: I have commented out the failing/unsupported test case for now

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
sagarbakliwal: Commented the failing/unsupported test case for now

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -56,6 +57,46 @@ private static (RecalcEngine engine, RecordValue parameters) OptionSetTestSetup(
...
+            var r4 = FormulaValue.NewRecordFromFields(new NamedValue("OptionSetField1", o2Val), new NamedValue("StrField1", FormulaValue.New("test4")));
+            
+            // Testing with missing/blank option set field is throwing an exception. Once that is resolved uncomment and fix the test case in Sort.txt
+            var r5 = FormulaValue.NewRecordFromFields(new NamedValue("StrField1", FormulaValue.New("test5")));
+
+            var rType = RecordType.Empty()
+                .Add(new NamedFormulaType("OptionSetField1", FormulaType.OptionSetValue, "DisplayNameField1"))
+                .Add(new NamedFormulaType("StrField1", FormulaType.String, "DisplayNameField2"));
+
+            var t1 = new RecordsOnlyTableValue(Core.IR.IRContext.NotInSource(rType.ToTable()), new List<RecordValue>() { r1, r2 });
```
MikeStall, line 85: >var t1 = new RecordsOnlyTableValue(Core.IR.IRContext.NotInSource(rType.ToTable()), new List<RecordValue>() { r1, r2 }); [](http://example.com/codeflow?start=12&length=119)

nit:  we have a public helper to simplify this
var t1 = FormulaValue.NewTable(rType, r1, r2);

(IRContext is internal)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Sort.txt
MikeStall: This is hitting #731. 
But it's just happening in the test harness when trying to print the table  w/ option set. 
If we wrap in a Concat(), we avoid the test bug and still excercise the code. 

Can you adjust test case like this:
>> Concat(Sort(t1, OptionSetField1, SortOrder.Ascending), ThisRecord.StrField1 & ",")
"test1,test2,"

Then the output is a string, which the harness handles. 
I'd mention 731 in the test comments too.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Sort.cs
lesaltzm: Thanks!

lesaltzm (dismissed):
🚢

lesaltzm (approved):

PR #770, opened 2022-10-17
Update blank value handling for text functions in the C# interpreter

CarlosFigueiraMSFT:
Blank values should be coerced to empty strings (`""`) or zero (`0`) when passed as inputs to string or numeric arguments in functions. This change updates text functions to do so.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
@@ -408,6 +408,11 @@ namespace Microsoft.PowerFx.Functions
             var count = ((NumberValue)args[2]).Value;
             var replacement = ((StringValue)args[3]).Value;
 
+            if (start <= 0 || count < 0)
+            {
+                return CommonErrors.ArgumentOutOfRange(irContext);
+            }
+
             if (start >= int.MaxValue)
             {
                 start = source.Length + 1;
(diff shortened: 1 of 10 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -439,19 +444,19 @@ public static FormulaValue Split(IRContext irContext, StringValue[] args)
...
             {
                 return source;
             }
 
-            var match = (StringValue)args[1];
-            var replacement = (StringValue)args[2];
-
             var instanceNum = -1;
-            if (args[3] is NumberValue nv)
+            if (args.Length > 3)
```
Ananya-Gupta24, line 456: Probably not a frequently hit scenario but should we short circuit and return source if args.Length <= 3 and (match == replacement )?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/EndsWith.txt
```diff
@@ -142,13 +142,13 @@ true
...
 false
 
-//Null end case
+//Null end case - coerced to ""
 >> EndsWith("Hello",Blank())
-false
+true
 
-//Null both case
+//Null both case - both coerced to ""
```
Ananya-Gupta24, line 149: Could you add another case here that verifies the coercion- EndsWith("", Blank())?

Ananya-Gupta24 (approved):

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/EndsWith.txt
CarlosFigueiraMSFT: Added here and in StartsWith as well

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
CarlosFigueiraMSFT: This is a corner case that is already handled by the existing logic, so I'd rather have less code than more for a (negligible) performance improvement.

Ananya-Gupta24 (approved):

PR #793, opened 2022-10-28
Custom function error handling. Fixes #777

### LINKED ISSUE #777: Error handling in custom function
For custom functions. we should check for errors and blanks before invoking. 

Add to CustomFunction.cs , which define s afunction:
public static StringValue Execute(NumberValue x, BooleanValue b)

Since the function expects a NumberValue, and error like 1/0 is an ErrorValue and will 
If the function wanted to handle errors, it could take a formulaValue or DValue<NumberValue> . 

        [Fact]
        public void CustomFunctionError()
        {
            var config = new PowerFxConfig(null);
            config.AddFunction(new TestCustomFunction());
            var engine = new RecalcEngine(config);

            // Shows up in enuemeration
            var func = engine.GetAllFunctionNames().First(name => name == "TestCustom");
            Assert.NotNull(func);

            // Can be invoked. 
            var result = engine.Eval("TestCustom(1/0,true)");
            Assert.Equal("3,True", result.ToObject());
        }

Also consider Blank. And for NumberValue, should we just coerce Blank to 0?

jas-valgotar:
- Fixes #777 
- When Custom function will have error as Arg we would report those errors before trying to invoke the function.
- If Custom function has Blank() as arg and expected type is NumberValue then we would treat it as 0, if expected type is StringValue then we would treat it as empty string.
- Added test to demo the same.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
@@ -364,11 +364,17 @@ namespace Microsoft.PowerFx
                 }
                 else if (arg is BlankValue)
                 {
+                    if (errors == null)
+                    {
+                        errors = new List<ErrorValue>();
+                    }
+
                     errors.Add(CommonErrors.RuntimeTypeMismatch(IRContext.NotInSource(FormulaType.Blank)));
                 }
                 else if (arg is LambdaFormulaValue lambda)
                 {
-                    arg = async () => (BooleanValue)await lambda.EvalAsync();
+                    Func<Task<BooleanValue>> argLambda = async () => (BooleanValue)await lambda.EvalAsync();
+                    arg = argLambda;
                 }
 
                 args2.Add(arg);
--- a/src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
@@ -15,6 +15,7 @@ using Microsoft.PowerFx.Core.IR;
 using Microsoft.PowerFx.Core.Localization;
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Functions;
 using Microsoft.PowerFx.Syntax;
 using Microsoft.PowerFx.Types;
 using static Microsoft.PowerFx.Core.Localization.TexlStrings;
@@ -361,6 +362,10 @@ namespace Microsoft.PowerFx
                 {
                     arg = FormulaValue.New(string.Empty);
                 }
+                else if (arg is BlankValue)
+                {
+                    errors.Add(CommonErrors.RuntimeTypeMismatch(IRContext.NotInSource(FormulaType.Blank)));
+                }
                 else if (arg is LambdaFormulaValue lambda)
                 {
                     arg = async () => (BooleanValue)await lambda.EvalAsync();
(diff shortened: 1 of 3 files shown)


BruceHaley:

src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
```diff
@@ -336,19 +337,44 @@ public async Task<FormulaValue> InvokeAsync(IServiceProvider serviceProvider, Fo
...
                 else
                 {
-                    args2.Add(arg);
+                    args2.Add(args[i]);
                 }
             }
 
+            if (errors != null)
+            {
+                return ErrorValue.Combine(IRContext.NotInSource(FormulaType.BindingError), errors);
```
MikeStall, line 375: >FormulaType.BindingError [](http://example.com/codeflow?start=64&length=24)

This should be the function's return Type. 

Errors can be typed. 
Ie, Abs(1/0) returns an error of type number.

src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
```diff
@@ -336,19 +337,44 @@ public async Task<FormulaValue> InvokeAsync(IServiceProvider serviceProvider, Fo
...
+                if (args[i] is ErrorValue ev)
+                {
+                    if (errors == null)
+                    {
+                        errors = new List<ErrorValue>();
+                    }
+
+                    errors.Add(ev);
+                }
+                else if (args[i] is BlankValue && _info.ParamTypes[i] is NumberType)
```
MikeStall, line 352: >BlankValue [](http://example.com/codeflow?start=36&length=10)

There are  lot of cases here to consider (eg, what if I pass Blank to Date, Bool, record ...etc) .

Maybe all those are just failures now, but we should test it. 

In the future,  if we bind to a nullable BooleanValue? , I could imagine that accepts blanks. 

This overlaps the policies that we have in StandardErrorHandling uses:
replaceBlankValues: 
 DoNotReplaceBlank,
 ReplaceBlankWithZero
 ReplaceBlankWithEmptyString

src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
```diff
@@ -336,19 +337,44 @@ public async Task<FormulaValue> InvokeAsync(IServiceProvider serviceProvider, Fo
...
+                {
+                    args[i] = FormulaValue.New(0);
+                    args2.Add(args[i]);
```
MikeStall, line 355: >args2 [](http://example.com/codeflow?start=20&length=5)

From a "defensive coding" perspective, this pattern is easy for one of the else-branches to miss an args2.add(). 

Consider:
- introduce a local, FormulaValue adjustedArg, inside the loop body 
- having a single args2.add(adjustedArg)   at the bottom of the loop.   Then we know it's only added once. And the compiler will verify that adjustedArg was assigned to.

src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
```diff
@@ -336,19 +337,44 @@ public async Task<FormulaValue> InvokeAsync(IServiceProvider serviceProvider, Fo
                 }
             }
 
-            foreach (var arg in args)
+            List<ErrorValue> errors = null;
+            for (var i = 0; i < args.Length; i++)
```
MikeStall, line 341: > for (var i = 0; i < args.Length; i++) [](http://example.com/codeflow?start=11&length=38)

nit, 
I see why we switched away from for-each, but consider using locals at the top:

var arg = args[i];
var expectedType = _info.Params[i];

MikeStall (changes requested):
:clock1:

BruceHaley:

BruceHaley:

PR #796, opened 2022-10-28
Add distinct function

### LINKED ISSUE #732: Missing Distinct function
Hi all, the distinct function that is documented [here](https://learn.microsoft.com/en-us/power-platform/power-fx/reference/function-distinct) does not exist in the engine today. Ideally, that function could make it over to the C# version of the engine.

nathansoz:
Adding the `Distinct` function as documented [here](https://learn.microsoft.com/en-us/power-platform/power-fx/reference/function-distinct). Fixes #732

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
@@ -81,7 +81,6 @@ namespace Microsoft.PowerFx.Core.Texl
         public static readonly TexlFunction Day = _library.Append(new DayFunction());
         public static readonly TexlFunction Degrees = _library.Append(new DegreesFunction());
         public static readonly TexlFunction DegreesT = _library.Append(new DegreesTableFunction());
-        public static readonly TexlFunction Distinct = _library.Append(new DistinctFunction());
         public static readonly TexlFunction DropColumns = _library.Append(new DropColumnsFunction());
         public static readonly TexlFunction EndsWith = _library.Append(new EndsWithFunction());
         public static readonly TexlFunction Error = _library.Append(new ErrorFunction());
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
@@ -81,6 +81,7 @@ namespace Microsoft.PowerFx.Core.Texl
         public static readonly TexlFunction Day = _library.Append(new DayFunction());
         public static readonly TexlFunction Degrees = _library.Append(new DegreesFunction());
         public static readonly TexlFunction DegreesT = _library.Append(new DegreesTableFunction());
+        public static readonly TexlFunction Distinct = _library.Append(new DistinctFunction());
         public static readonly TexlFunction DropColumns = _library.Append(new DropColumnsFunction());
         public static readonly TexlFunction EndsWith = _library.Append(new EndsWithFunction());
         public static readonly TexlFunction Error = _library.Append(new ErrorFunction());
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Distinct.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Distinct.cs
@@ -21,7 +21,7 @@ using Microsoft.PowerFx.Types;
 
 namespace Microsoft.PowerFx.Core.Texl.Builtins
 {
-    // Sort(source:*, valueFunc:b, [order:s])
+    // Distinct(source:*, valueFunc:b)
     internal sealed class DistinctFunction : FunctionWithTableInput
     {
         public override bool IsSelfContained => true;
@@ -39,7 +39,7 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
             yield return new[] { TexlStrings.DistinctArg1, TexlStrings.DistinctArg2 };
         }
 
-        protected override bool CheckInvocation(CheckTypesContext context, TexlNode[] args, DType[] argTypes, IErrorContainer errors, out DType returnType, out Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
+        public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DType[] argTypes, IErrorContainer errors, out DType returnType, out Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
         {
             Contracts.AssertValue(args);
             Contracts.AssertAllValues(args);
@@ -48,7 +48,7 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
             Contracts.AssertValue(errors);
             Contracts.Assert(MinArity <= args.Length && args.Length <= MaxArity);
 
-            var fValid = CheckInvocation(args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
+            var fValid = CheckTypes(args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
             Contracts.Assert(returnType.IsTable);
 
             var lambdaType = argTypes[1];
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Distinct.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Distinct.cs
@@ -51,15 +51,14 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
             var fValid = CheckTypes(args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
             Contracts.Assert(returnType.IsTable);
 
-            var lambdaType = argTypes[1];
-
-            returnType = DType.CreateTable(new TypedName(new DType(lambdaType.Kind), GetOneColumnTableResultName(context.Features)));
-
             var exprType = argTypes[1];
+
+            returnType = DType.CreateTable(new TypedName(exprType, GetOneColumnTableResultName(context.Features)));
+
             if (!exprType.IsPrimitive || exprType.IsOptionSet)
             {
                 fValid = false;
-                errors.EnsureError(args[1], TexlStrings.ErrSortWrongType);
+                errors.EnsureError(args[1], TexlStrings.ErrDistinctWrongType);
             }
 
             return fValid;
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Distinct.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Distinct.cs
@@ -1,0 +1,76 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Microsoft.PowerFx.Core.App.ErrorContainers;
+using Microsoft.PowerFx.Core.Binding;
+using Microsoft.PowerFx.Core.Binding.BindInfo;
+using Microsoft.PowerFx.Core.Entities;
+using Microsoft.PowerFx.Core.Functions;
+using Microsoft.PowerFx.Core.Functions.Delegation;
+using Microsoft.PowerFx.Core.Functions.Delegation.DelegationMetadata;
+using Microsoft.PowerFx.Core.Functions.FunctionArgValidators;
+using Microsoft.PowerFx.Core.Localization;
+using Microsoft.PowerFx.Core.Logging.Trackers;
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Syntax;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Core.Texl.Builtins
+{
+    // Sort(source:*, valueFunc:b, [order:s])
+    internal sealed class DistinctFunction : FunctionWithTableInput
+    {
+        public override bool IsSelfContained => true;
+
+        public override bool SupportsParamCoercion => false;
+
+        public DistinctFunction()
+            : base("Distinct", TexlStrings.AboutDistinct, FunctionCategories.Table, DType.EmptyTable, 0x02, 2, 2, DType.EmptyTable)
+        {
+            ScopeInfo = new FunctionScopeInfo(this);
+        }
+
+        public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
+        {
+            yield return new[] { TexlStrings.DistinctArg1, TexlStrings.DistinctArg2 };
+        }
+
+        protected override bool CheckInvocation(CheckTypesContext context, TexlNode[] args, DType[] argTypes, IErrorContainer errors, out DType returnType, out Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
+        {
+            Contracts.AssertValue(args);
+            Contracts.AssertAllValues(args);
+            Contracts.AssertValue(argTypes);
+            Contracts.Assert(args.Length == argTypes.Length);
+            Contracts.AssertValue(errors);
+            Contracts.Assert(MinArity <= args.Length && args.Length <= MaxArity);
+
+            var fValid = CheckInvocation(args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
+            Contracts.Assert(returnType.IsTable);
+
+            var lambdaType = argTypes[1];
+
+            returnType = DType.CreateTable(new TypedName(new DType(lambdaType.Kind), GetOneColumnTableResultName(context.Features)));
+
+            var exprType = argTypes[1];
+            if (!exprType.IsPrimitive || exprType.IsOptionSet)
+            {
+                fValid = false;
+                errors.EnsureError(args[1], TexlStrings.ErrSortWrongType);
+            }
+
+            return fValid;
+        }
+
+        // This method returns true if there are special suggestions for a particular parameter of the function.
+        public override bool HasSuggestionsForParam(int argumentIndex)
+        {
+            Contracts.Assert(argumentIndex >= 0);
+
+            return argumentIndex == 0;
+        }
+    }
+}
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Distinct.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Distinct.cs
@@ -55,11 +55,11 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
 
             returnType = DType.CreateTable(new TypedName(exprType, GetOneColumnTableResultName(context.Features)));
 
-            if (!exprType.IsPrimitive || exprType.IsOptionSet)
-            {
-                fValid = false;
-                errors.EnsureError(args[1], TexlStrings.ErrDistinctWrongType);
-            }
+            //if (!exprType.IsPrimitive || exprType.IsOptionSet)
+            //{
+            //    fValid = false;
+            //    errors.EnsureError(args[1], TexlStrings.ErrDistinctWrongType);
+            //}
 
             return fValid;
         }
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Distinct.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Distinct.cs
@@ -55,12 +55,6 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
 
             returnType = DType.CreateTable(new TypedName(exprType, GetOneColumnTableResultName(context.Features)));
 
-            //if (!exprType.IsPrimitive || exprType.IsOptionSet)
-            //{
-            //    fValid = false;
-            //    errors.EnsureError(args[1], TexlStrings.ErrDistinctWrongType);
-            //}
-
             return fValid;
         }
 
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
@@ -9,6 +9,7 @@ using Microsoft.PowerFx.Core.Binding;
 using Microsoft.PowerFx.Core.Functions;
 using Microsoft.PowerFx.Core.IR;
 using Microsoft.PowerFx.Core.Texl;
+using Microsoft.PowerFx.Core.Texl.Builtins;
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
 using Microsoft.PowerFx.Interpreter;
@@ -18,6 +19,12 @@ namespace Microsoft.PowerFx.Functions
 {
     internal static partial class Library
     {
+        /// <summary>
+        /// This isn't part of <see cref="BuiltinFunctionsCore"/> since PA has different implementation of
+        /// Texl Instance of <see cref="DistinctFunction"/>.
+        /// </summary>
+        public static readonly TexlFunction DistinctInterpreterFunction = new DistinctFunction();
+
         private static readonly DateTime _epoch = new DateTime(1899, 12, 30, 0, 0, 0, 0);
 
         // Helper to get a service or fallback to a default if the service is missing.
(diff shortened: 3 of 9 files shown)


src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Distinct.cs
```diff
@@ -0,0 +1,76 @@
...
+using Microsoft.PowerFx.Core.Localization;
+using Microsoft.PowerFx.Core.Logging.Trackers;
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Syntax;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Core.Texl.Builtins
+{
+    // Sort(source:*, valueFunc:b, [order:s])
```
jas-valgotar, line 24: >// Sort(source:*, valueFunc:b, [order:s]) [](http://example.com/codeflow?start=4&length=41)

This should be for Distinct now.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -377,6 +377,76 @@ private static async Task<(DValue<RecordValue> row, FormulaValue sortValue)> App
...
+            if (!row.IsValue)
+            {
+                return (row, row.ToFormulaValue());
+            }
+
+            var childContext = context.SymbolContext.WithScopeValues(row.Value);
+            var distinctValue = await lambda.EvalInRowScopeAsync(context.NewScope(childContext));
+
+            return (row, distinctValue);
+        }
```
jas-valgotar, line 391: Can the ApplySortLambda() be resused? maybe rename that one?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -377,6 +377,76 @@ private static async Task<(DValue<RecordValue> row, FormulaValue sortValue)> App
...
+        {
+            var arg0 = (TableValue)args[0];
+            var arg1 = (LambdaFormulaValue)args[1];
+
+            var pairs = (await Task.WhenAll(arg0.Rows.Select(row => ApplyDistinctLambda(runner, context, row, arg1)))).ToList();
+
+            var errors = new List<ErrorValue>();
+            bool allNumbers = true, allStrings = true, allBooleans = true, allDatetimes = true, allDates = true;
+
+            foreach (var (row, sortValue) in pairs)
```
jas-valgotar, line 403: >sortValue [](http://example.com/codeflow?start=31&length=9)

nit: rename to distinctValue / Value?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -462,6 +532,44 @@ private static bool IsValueTypeErrorOrBlank<T>(FormulaValue val)
...
+            where TDotNetPrimitive : IComparable<TDotNetPrimitive>
+        {
+            var values = new Dictionary<TDotNetPrimitive, DValue<RecordValue>>();
+            var name = ((TableType)irContext.ResultType).SingleColumnFieldName;
+            var type = ((TableType)irContext.ResultType).SingleColumnFieldType;
+
+            // Have to handle null
+            var addBlankValue = false;
+
+            foreach (var (row, sortValue) in pairs)
```
jas-valgotar, line 546: >sortValue [](http://example.com/codeflow?start=31&length=9)

nit: rename to distinctValue / Value?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -462,6 +532,44 @@ private static bool IsValueTypeErrorOrBlank<T>(FormulaValue val)
...
+            {
+                if (sortValue?.ToObject() == null)
```
jas-valgotar, line 548: >sortValue?.ToObject() == null [](http://example.com/codeflow?start=20&length=29)

this can be rewritten as, `if(sortValue is BlankValue)`

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
```diff
@@ -78,6 +78,7 @@ internal class BuiltinFunctionsCore
         public static readonly TexlFunction Day = _library.Append(new DayFunction());
         public static readonly TexlFunction Degrees = _library.Append(new DegreesFunction());
         public static readonly TexlFunction DegreesT = _library.Append(new DegreesTableFunction());
+        public static readonly TexlFunction Distinct = _library.Append(new DistinctFunction());
```
anderson-joyle, line 81: >public static readonly TexlFunction Distinct = _library.Append(new DistinctFunction()); [](http://example.com/codeflow?start=8&length=87)

By adding this function here it means that PA team also need to implement it from their side. Lets check with them first and then decide if we add the function to the _library or _featureGateFunctions collection.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Distinct.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            ScopeInfo = new FunctionScopeInfo(this);
+        }
+
+        public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
+        {
+            yield return new[] { TexlStrings.DistinctArg1, TexlStrings.DistinctArg2 };
+        }
+
+        protected override bool CheckInvocation(CheckTypesContext context, TexlNode[] args, DType[] argTypes, IErrorContainer errors, out DType returnType, out Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
```
anderson-joyle, line 42: >CheckInvocation [](http://example.com/codeflow?start=32&length=15)

CheckInvocation has been split into 2 different checks: CheckTypes and CheckSemantics. Check SortFunction changes as reference.

anderson-joyle, line 42: >args [](http://example.com/codeflow?start=86&length=4)

Are we checking if column name (args[1]) exists in the table?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Distinct.txt
```diff
@@ -0,0 +1,59 @@
+>> Distinct(Blank(), true)
+Blank()
+
+>> Distinct([], true)
+Table()
+
+>> Distinct([{A: 1}, {A: 2}, {A: 3}, {A: 4}], true)
+Table({Result:true})
```
jas-valgotar, line 8: Should this return the original source? since formula is always true, meaning everything would evaluate to distinct?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Distinct.txt
```diff
@@ -0,0 +1,59 @@
...
+
+>> Distinct([{A: 1}, {A: 2}, {A: 3}, {A: 4}], false)
+Table({Result:false})
```
jas-valgotar, line 11: Should this return the empty table? since formula is always false, meaning everything would not evaluate as distinct?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Distinct.cs
nathansoz: Yep, good catch.

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
nathansoz: Did this get resolved?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Distinct.cs
nathansoz: Updated to checktypes

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
nathansoz: Updated.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
nathansoz: Yeah. I think I forked it thinking they would have to be different and then they weren't. Good catch.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
nathansoz: Updated.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
nathansoz: Updated.

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
jas-valgotar: @anderson-joyle can we add it to _featureGateFunctions for now, at least that unblocks Nathan.

anderson-joyle: Lets get it included to _featureGateFunctions for now.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -462,6 +519,44 @@ private static bool IsValueTypeErrorOrBlank<T>(FormulaValue val)
...
+
+            foreach (var (row, distinctValue) in pairs)
+            {
+                if (distinctValue is BlankValue)
+                {
+                    addBlankValue = true;
+                    continue;
+                }
+
+                var key = (TDotNetPrimitive)distinctValue?.ToObject();
```
MikeStall, line 541: We probably need some platform functionality for this. 
It's important that this is limited to just primitives (ToObject() on records is not reliable).

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
jas-valgotar: turns out this is already in PA, keeping it under _library

jas-valgotar (commented):
/azp run PowerFx-PR

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
MikeStall: My concern is that PAClient has its own implementation of this, [https://msazure.visualstudio.com/DefaultCollection/OneAgile/_git/PowerApps-Client?path=/src/Cloud/DocumentServer.Core/Document/Texl/Builtins/Distinct.cs&version=GBmaster](https://msazure.visualstudio.com/DefaultCollection/OneAgile/_git/PowerApps-Client?path=/src/Cloud/DocumentServer.Core/Document/Texl/Builtins/Distinct.cs&version=GBmaster)

So adding to this list may hide that implementation. 

Let's just move this as a static field on Library.cs, and then it's only used by interpreter.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -493,6 +493,19 @@ static Library()
                     returnBehavior: ReturnBehavior.AlwaysEvaluateAndReturnResult,
                     targetFunction: Dec2Hex)
             },
+            {
+                 BuiltinFunctionsCore.Distinct,
```
MikeStall, line 497: > BuiltinFunctionsCore.Distinct, [](http://example.com/codeflow?start=16&length=31)

Pull from static field on Library.cs, rather than shared BuiltinFunctionsCore.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Distinct.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            ScopeInfo = new FunctionScopeInfo(this);
+        }
+
+        public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
+        {
+            yield return new[] { TexlStrings.DistinctArg1, TexlStrings.DistinctArg2 };
+        }
+
+        public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DType[] argTypes, IErrorContainer errors, out DType returnType, out Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
```
MikeStall, line 42: >CheckTypes [](http://example.com/codeflow?start=29&length=10)

double check this impl against PAClient's.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Distinct.txt
```diff
@@ -0,0 +1,82 @@
...
+
+>> Distinct(Table({a:1,b:2},{a:3,b:4},{a:5,b:6}), Trunc(a / 2))
+Table({Result:0},{Result:1},{Result:2})
+
+// Distinct values for an expression that generates same result
+>> Distinct([-2,-1,0,1,0],Value^2)
+Table({Result:4},{Result:1},{Result:0})
+
+// Distinct values with blanks
+>> Distinct(Table({FirstName:"Harry"},{FirstName:Blank()},{FirstName:"Sally"},{FirstName:"Harry"},{FirstName:Blank()},{FirstName:"Alice"},{FirstName:"Sally"}), FirstName)
```
MikeStall, line 32: >"Harry" [](http://example.com/codeflow?start=29&length=7)

Can we add a test for case insensitivity. 

"Harry", "HARRY",  ... same of different?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Distinct.txt
```diff
@@ -0,0 +1,82 @@
...
+
+// Distinct values where expression handles errors
+>> Distinct(Table({a:1},{a:2},{a:If(1/0<2,{a:2}).a},{a:7},{a:Sqrt(-1)},{a:-13},{a:2}), IfError(a,-FirstError.Kind))
+Table({Result:1},{Result:2},{Result:-13},{Result:7},{Result:-24})
+
+// Distinct which ignores columns with errors
+>> Distinct(Table({a:10,b:1/0},{a:9,b:Sqrt(-1)},{a:8,b:3},{a:10,b:1/0},{a:8,b:Ln(-1)}), a)
+Table({Result:10},{Result:9},{Result:8})
+
+>> Distinct(Blank(), true)
```
MikeStall, line 51: can we add a test where arg0 is a table, but a runtime error?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -493,6 +504,33 @@ private static bool IsValueTypeErrorOrBlank<T>(FormulaValue val)
...
+            var type = irContext.ResultType;
+
+            foreach (var (row, distinctValue) in pairs)
+            {
+                if (distinctValue is ErrorValue errorValue)
+                {
+                    return errorValue;
+                }
+
+                var key = distinctValue.ToObject();
```
MikeStall, line 521: >distinctValue.ToObject(); [](http://example.com/codeflow?start=26&length=25)

Can we gaurante this is a primitive object?

jas-valgotar: Yes, CheckTypes guarantees it.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Distinct.txt
jas-valgotar: added

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Distinct.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            var exprType = argTypes[1];
+
+            returnType = DType.CreateTable(new TypedName(exprType, GetOneColumnTableResultName(context.Features)));
+
+            //if (!exprType.IsPrimitive || exprType.IsOptionSet)
+            //{
+            //    fValid = false;
+            //    errors.EnsureError(args[1], TexlStrings.ErrDistinctWrongType);
+            //}
```
MikeStall, line 62: nit: remove commented code?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -493,6 +504,40 @@ private static bool IsValueTypeErrorOrBlank<T>(FormulaValue val)
...
+
+                if(key != null && key is not object)
```
MikeStall, line 523: >key is not object [](http://example.com/codeflow?start=34&length=17)

How do we hit this?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Distinct.txt
```diff
@@ -0,0 +1,92 @@
+// Blank table
+>> Distinct(If(1<0,[1,2]), true)
+Blank()
+
+// Empty table
+>> Distinct(Filter([1,2,3],Value>10), true)
+Table()
+
+// Constant expressions
```
MikeStall, line 9: Sounds like For In-memory collections, existing JS implementation is indeed ordered.  We should assume our 1+ million customers rely on that behavior even if it’s not in the docs. 
But delegated calls are not necessarily ordered

So I'd comment about ordering and add these test cases: 

Distinct([1,2,3,2,1], Value)
1,2,3

Distinct([3,2,1,3,3,3,3], Value)
3,2,1

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Distinct.txt
```diff
@@ -0,0 +1,92 @@
+// Blank table
+>> Distinct(If(1<0,[1,2]), true)
+Blank()
+
+// Empty table
```
MikeStall, line 5: It looks like PowerApps does allow record comparison!

CountRows(Distinct(Table({x:1,y:2},{x:1,y:2}), ThisRecord)) // 1

First(Distinct(Table({x:1,y:2},{x:10,y:2}), ThisRecord)).Result.x // 1

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
MikeStall: How is CheckTypes  guaranteeing that?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -493,6 +504,40 @@ private static bool IsValueTypeErrorOrBlank<T>(FormulaValue val)
...
+                {
+                    return CommonErrors.RuntimeTypeMismatch(irContext);
+                }
+
+                var keySerialized = System.Text.Json.JsonSerializer.Serialize(key);
```
MikeStall, line 528: >System.Text.Json.JsonSerializer.Serialize(key); [](http://example.com/codeflow?start=36&length=47)

We should avoid the json call (see #884).
Beware that Json isn't necessarily canonical -the same value can serializer as multiple different strings. 

But we could use FormulaValue.ToExpressionValue!
That also gets a string value.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -493,6 +504,40 @@ private static bool IsValueTypeErrorOrBlank<T>(FormulaValue val)
             return val is T || val is BlankValue || val is ErrorValue;
         }
 
+        private static FormulaValue DistinctValueType(List<(DValue<RecordValue> row, FormulaValue distinctValue)> pairs, IRContext irContext)
+        {
+            var lookup = new HashSet<string>();
+            var result = new List<DValue<RecordValue>>();
+            var name = ((TableType)irContext.ResultType).SingleColumnFieldName;
+            var type = irContext.ResultType;
```
MikeStall, line 512: >      var type = irContext.ResultType; [](http://example.com/codeflow?start=6&length=38)

nit, not used. 
(although could be used on line 511)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
MikeStall: Although ... why do we need to stringify at all? Can't we just use `HashSet<object>(); ` and object.equals?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
jas-valgotar: I removed the check since PA supports Record for distinct.

jas-valgotar: Added a check in implementation now.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
jas-valgotar: Since now we are not using Agreegate types, moved back to `HashSet<object>();`

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Distinct.txt
jas-valgotar: added the test as mentioned.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Distinct.txt
jas-valgotar: Added test demoing we only support primitive types.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Distinct.cs
jas-valgotar: removed

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -408,6 +408,76 @@ private static async Task<(DValue<RecordValue> row, FormulaValue sortValue)> App
...
+
+            return (row, distinctValue);
+        }
+
+        public static async ValueTask<FormulaValue> DistinctTable(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
+        {
+            var arg0 = (TableValue)args[0];
+            var arg1 = (LambdaFormulaValue)args[1];
+
+            var pairs = (await Task.WhenAll(arg0.Rows.Select(row => ApplyDistinctLambda(runner, context, row, arg1)))).ToList();
```
victoriaquasar: Could we have a well structured variable tree? Easier to debug via breakpoints.

Ex (not tested):

```cs
var values = arg0.Rows.Select(row => ApplyDistinctLambda(runner, context, row, arg1));
var pairResults = await Task.WhenAll(values);
var pairs = pairResults.ToList();
```

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #802, opened 2022-11-02
Adds support for partial input type checking 

### LINKED ISSUE #720: Add the ability to perform type checking validation for partial input
#### Our scenario:
We have two input variables that can get referenced in an expression, `A` and `B`.
The schema of `A` is known to us at authoring time, whereas the the schema of `B` is only known to us at runtime. 

#### Ask:
We would like to be able to validate the expression against the schema of `A` as a first pass, without having to wait until runtime to do the full `Check()` with schema of both `A` and `B`.

examples:
Suppose we know `A`'s schema is
```c#
RecordType().Empty()
     .Add(new NamedFormulaType("prop1", FormulaType.Number)
```

`Power(A.prop1 + 10, B.prop2) ` --> In the first pass, we should be able to verify that A.prop1 is a Number, and that the first argument of Power() is correct, without having to know anything about B.

negative cases:
 `Power("hello", B.prop2) ` --> Error in the first pass, as the first argument of Power() should be a Number..
 
`A.prop1.blah + B.prop2` --> Error in the first pass while type checking against `A`. We will learn `A.prop1.blah` is invalid because `prop1` is a primitive number without nested fields.

jas-valgotar:
- Fixes #720 

**Need:**
Some teams would like to perform partial type checking at authoring time.
e.g. `deferred + number => valid` but, `deferred + record => error`

NOTE: In this phase, The functions that take lambda expressions as input still does not support `deferred (Unknown Type)`
For that the below options were considered:

**Input type consideration:**
1. Introduce new type (e.g. deferredType) 
This could work, but we might miss some cases along the Binder and cause the issue. Plus, it would require lot of changes.
2. Use an existing type if feasible:
Type Considered:
1._UtypedObject or Error as deferred type_: The Pro is these types of nodes can already be generated by user input so we could leverage already existing checks. But The issue here is we cannot differentiate between if user intended to use `actual UntypedObject` or `UntypedObject as deferred type`. And for error we can't distinguish between the `actual Error` or `Error as deferred type`.
Also, all existing type would not accept this type for type checking.
\
2._Unknown as deferred type_: This type looks promising, as this type is already accepted by All Primitive types and hence can be used as `deferred type`. Also, its being currently used only as return type of few functions like `If`, `Coalesce` and hence there are no clashes on input as no function currently accepts `UnknownType` as input. So that makes it easy to distinguish that user did intend to use it as `deferred type`. As it does pretty well satisfy all we need, I have decided to use this one as input type.

**Implementation Consideration:**
1. Changes in CheckTypes:
Pro to this approach being we can achieve a true type checking for all the function. But that would require changes in 100s of places and could turn into headache to maintain and can potentially introduce bugs.

2. Post process the error that was generated by CheckTypes:
In this approach, we would check all the errors generated by CheckTypesand if all of them were generated due to `Unknown` Type argument and if indeed they were, we consider the type checking to be successful and discard those errors. This would require changes in only few places in binder. As this would be a low risk, we decided to go with this approach.

**What users will achieve?**
Using this `Unkown` type. User will be able to partially type check expressions at authoring time and will get warnings where `Unknown` type is used. So, if schema of any argument X is unknown, user can still get meaningful error for other arguments whose types are known.
e.g.
`Power(10, X.field1)` would just give a warning at return time.
`Power({x:1}, X.field1)` would fail as record is not a valid input.


BruceHaley:

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
```diff
@@ -3038,6 +3038,48 @@ public void TestAttachmentIf()
...
+        [InlineData("\"The\" in X")]
+        [InlineData("123 in X")]
+        [InlineData("123 in Table(X)")]
+        [InlineData("IsBlank(X)")]
+        [InlineData("IsError(X)")]
+        [InlineData("Text(X)")]
+        [InlineData("Value(X)")]
+        [InlineData("Boolean(X)")]
+        [InlineData("Index([1,2,3], X)")]
+        public void DeferredTypeTest(string script)
```
MikeStall, line 3060: >DeferredTypeTest [](http://example.com/codeflow?start=20&length=16)

Can we add a comment to each of these like:

// Ensure expression binds without any errors - but issues a warning for the deferred type

or

// Ensure expression issues an error , despite the deferred type.

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
```diff
@@ -3038,6 +3038,48 @@ public void TestAttachmentIf()
...
+        [InlineData("X < 2")]
+        [InlineData("\"The\" in X")]
+        [InlineData("123 in X")]
+        [InlineData("123 in Table(X)")]
+        [InlineData("IsBlank(X)")]
+        [InlineData("IsError(X)")]
+        [InlineData("Text(X)")]
+        [InlineData("Value(X)")]
+        [InlineData("Boolean(X)")]
+        [InlineData("Index([1,2,3], X)")]
```
MikeStall, line 3059: > [InlineData("Index([1,2,3], X)")] [](http://example.com/codeflow?start=7&length=34)

Can we make it stronger:
Index([1,2,3], X).Value

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
```diff
@@ -3038,6 +3038,48 @@ public void TestAttachmentIf()
...
+        [InlineData("$\"Test {X} {XR}\"")]
+        [InlineData("Table(X, XN)")]
+        [InlineData("X.field + XN.field")]
+        [InlineData("Index([1,2,3], X).missing")]
+        public void DeferredTypeTest_Negative(string script)
+        {
+            Preview.FeatureFlags.StringInterpolation = true;
+            var symbolTable = new SymbolTable();
+            symbolTable.AddVariable("X", FormulaType.Unknown);
+            symbolTable.AddVariable("XN", FormulaType.Number);
```
MikeStall, line 3078: >XN [](http://example.com/codeflow?start=37&length=2)

minor nit: for convention in these tests, we could have X be the Unknown type. 

For the non-deferred types, can we avoid the 'X' prefix - we call them "N" or "R"

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
```diff
@@ -3119,5 +3161,33 @@ private static void TestSimpleBindingSuccess(string script, DType expectedType,
...
+            {
+                SymbolTable = symbolTable
+            };
+
+            var engine = new Engine(config);
+            var result = engine.Check(script);
+
+            Assert.True(result.IsSuccess);
+            Assert.True(result._binding.ErrorContainer.HasErrors());
+            Assert.True(result._binding.ErrorContainer.GetErrors().All(error => error.MessageKey.Equals(TexlStrings.WarnUnknownType.Key)));
```
MikeStall, line 3177: >result._binding.ErrorContainer.GetErrors().All(error => error.MessageKey.Equals(TexlStrings.WarnUnknownType.Key))); [](http://example.com/codeflow?start=24&length=115)

Can we assert this on result.Errors directly (rather than result._binding.ErrorContainer...)

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
```diff
@@ -3038,6 +3038,48 @@ public void TestAttachmentIf()
...
+        [InlineData("X.field + XN.field")]
+        [InlineData("Index([1,2,3], X).missing")]
+        public void DeferredTypeTest_Negative(string script)
+        {
+            Preview.FeatureFlags.StringInterpolation = true;
+            var symbolTable = new SymbolTable();
+            symbolTable.AddVariable("X", FormulaType.Unknown);
+            symbolTable.AddVariable("XN", FormulaType.Number);
+            symbolTable.AddVariable("XR", RecordType.Empty());
+            TestBindingError(script, symbolTable);
```
MikeStall, line 3080: >TestBindingError [](http://example.com/codeflow?start=12&length=16)

we could make the assert stronger here by passing in a fragment of the expected error message and ensuring that shows up.

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -138,8 +138,41 @@ internal static bool TryGetBestOverload(TexlBinding txb, CallNode node, DType[]
 
                 var localWarnings = new LimitedSeverityErrorContainer(txb.ErrorContainer, DocumentErrorSeverity.Warning);
 
+                // This error container is used as temporary container so we can trap type mismatch kind of error for
+                // deferred (unknown) type args and validate all the errors were caused due to deferred(unknown) type.
```
MikeStall, line 142: >deferred [](http://example.com/codeflow?start=19&length=8)

Changes in binder.cs look similar - can this code be shared?

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -669,7 +702,7 @@ private static BinderCheckTypeResult CheckEqualArgTypesCore(IErrorContainer erro
 
             // EqualOp is only allowed on primitive types, polymorphic lookups, and control types.
             if (!(typeLeft.IsPrimitive && typeRight.IsPrimitive) && !(typeLeft.IsPolymorphic && typeRight.IsPolymorphic) && !(typeLeft.IsControl && typeRight.IsControl)
-                && !(typeLeft.IsPolymorphic && typeRight.IsRecord) && !(typeLeft.IsRecord && typeRight.IsPolymorphic))
+                && !(typeLeft.IsPolymorphic && typeRight.IsRecord) && !(typeLeft.IsRecord && typeRight.IsPolymorphic) && !(typeLeft.IsUnknown || typeRight.IsUnknown))
```
MikeStall, line 705: >&& !(typeLeft.IsUnknown || typeRight.IsUnknown) [](http://example.com/codeflow?start=118&length=47)

how do we hit this?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -229,7 +229,8 @@ public CheckResult Check(ParseResult parse, RecordType parameterType, ParserOpti
                 CultureInfo = Config.CultureInfo
             };
 
-            if (result.IsSuccess)
+            // If the result is not sucess or has warning related to DeferredArgs we don't want to calculate return type or generate the IR Tree.
+            if (result.IsSuccess && !result.HasDeferredArgsWarning)
```
MikeStall, line 233: >&& !result.HasDeferredArgsWarning [](http://example.com/codeflow?start=33&length=33)

Note that PowerApps will call the binder directly and not go through this method. 
And we want to ensure that we're not changing error cases into passing cases. 

So we need to be sure that either:
1. Any check that we "loosened" to no longer create an error ("X.field") was a case that didn't happen in Power Apps. 
2. Have a binder flag so that PowerApps issues error. 
3. Possibly post-process - always add the error (so PowerApps gets it), but then remove error in this path.

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
jas-valgotar: This is hit on equality operators, here, if one of the operands is unknown, we still don't want to give an error.
e.g. unknown = number => bool

BruceHaley:

BruceHaley:

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -324,13 +357,13 @@ private static BinderCheckTypeResult CheckInArgTypesCore(IErrorContainer errorCo
 
             var coercions = new List<BinderCoercionResult>();
 
-            if (!typeLeft.IsValid || typeLeft.IsUnknown || typeLeft.IsError)
```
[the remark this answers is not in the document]
jas-valgotar: You are correct! The assumption was since no function produces the Unknown Type, it should be fine. But to be on safer side I have now added the flag

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
jas-valgotar: I have now added the flag to mitigate the possible risk to PowerApps.

BruceHaley:

BruceHaley:

src/tests/Microsoft.PowerFx.Core.Tests/TypeSystemTests/TypeTests.cs
```diff
@@ -82,5 +84,79 @@ public void TryGetFieldTypeLookupNullTest()
...
+        }
+
+        [Theory]
+        [InlineData("$\"Test {X} {R}\"")]
+        [InlineData("Table(X, N)")]
+        [InlineData("X.field + N.field")]
+        [InlineData("Index([1,2,3], X).missing")]
+
+        // Ensures expression issues an error if it exists, despite the deferred type.
+        public void DeferredTypeTest_Negative(string script)
```
MikeStall, line 124: >DeferredTypeTest_Negative [](http://example.com/codeflow?start=20&length=25)

we could make the assert stronger here by passing in a fragment of the expected error message and ensuring that shows up.

src/tests/Microsoft.PowerFx.Interpreter.Tests/FlagTests.cs
```diff
@@ -46,5 +46,88 @@ public void Parser_DisableRowScopeDisambiguationSyntax(string expression)
...
+        [InlineData("123 in Table(X)", false)]
+        [InlineData("IsBlank(X)")]
+        [InlineData("IsError(X)", false)]
+        [InlineData("Text(X)")]
+        [InlineData("Value(X)")]
+        [InlineData("Boolean(X)")]
+        [InlineData("Index([1,2,3].Value, X)")]
+
+        // Ensures expression binds without any errors - but issues a warning for the deferred(unknown) type.
+        public void DeferredTypeTest_EnableDeferredType(string script, bool isEvalAllowed = true)
```
MikeStall, line 71: >DeferredTypeTest_EnableDeferredType [](http://example.com/codeflow?start=20&length=35)

what's relation of these tests to TypeTests?

jas-valgotar: Ahh, I just missed to remove it while I was moving the test to different place. Removed now...

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -184,6 +194,39 @@ internal static bool TryGetBestOverload(TexlBinding txb, CallNode node, DType[]
             return false;
         }
 
+        /// <summary>
+        /// Typecheck the invocation and infer the return type. Error on deferred(unknown) arg is not considered.
+        /// </summary>
+        internal static bool HandleCheckInvocationWithUnknown(TexlFunction maybeFunc, TexlBinding txb, TexlNode[] args, DType[] argTypes, out ErrorContainer checkInvocationErrors, out DType returnType, out Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
```
MikeStall, line 200: >HandleCheckInvocationWithUnknown [](http://example.com/codeflow?start=29&length=32)

@bryancroteau-MSFT  - fyi

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -184,6 +194,39 @@ internal static bool TryGetBestOverload(TexlBinding txb, CallNode node, DType[]
...
+        {
+            // This error container is used as temporary container so we can trap type mismatch kind of error for
+            // deferred (unknown) type args and validate all the errors were caused due to deferred(unknown) type.
+            checkInvocationErrors = new ErrorContainer();
+
+            var typeCheckSucceeded = maybeFunc.HandleCheckInvocation(txb, args, argTypes, checkInvocationErrors, out returnType, out nodeToCoercedTypeMap);
+
+            // If type check failed and errors were due to Unknown type node we would like to consider the typeChecking passed.
+            if (txb.Features.HasFlag(Features.EnableDeferredType) && !typeCheckSucceeded && checkInvocationErrors.HasErrors() && checkInvocationErrors.GetErrors().All(error => txb.GetType(error.Node).IsUnknown))
```
bryancroteau-MSFT, line 209: We should not be using txb as part of type checking ever.

bryancroteau-MSFT:
As a general comment, we should prefer to use CheckTypes in isolation for new features, and not call CheckSemantics. CheckInvocation is deprecated, and HandleCheckInvocation will be removed. HandleCheckInvocation was never intended to be permanent. CheckTypes is currently protected but the access modifier can be changed.

We should not be storing types in TexlBinding. If you want to store the types from the output of CheckTypes, do so in an immutable data structure. Storing types in a mutable data structure like TexlBinding is a very bad idea and we should never be doing that in code moving forward.

src/tests/Microsoft.PowerFx.Interpreter.Tests/DefeeredTypeTests.cs
```diff
@@ -0,0 +1,95 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Microsoft.PowerFx.Core.Localization;
+using Microsoft.PowerFx.Types;
+using Xunit;
+
+namespace Microsoft.PowerFx.Interpreter
+{
+    public class DefeeredTypeTests
```
lesaltzm, line 14: nit: spelling, deferred

src/tests/Microsoft.PowerFx.Interpreter.Tests/DefeeredTypeTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            var symbolTable = new SymbolTable();
+            symbolTable.AddVariable("X", FormulaType.Unknown);
+            TestDeferredTypeBindingWarning(script, Features.EnableDeferredType, symbolTable);
+        }
+
+        [Theory]
+        [InlineData("$\"Test {X} {R}\"", "Invalid argument type (Record). Expecting a Text value instead.")]
+        [InlineData("Table(X, N)", "Cannot use a non-record value in this context")]
+        [InlineData("X.field + N.field", "Invalid use of '.'")]
+        [InlineData("Index([1,2,3], X).missing", "Name isn't valid. 'missing' isn't recognized")]
```
lesaltzm, line 49: also, `With(x, ...)`, `ForAll(x, ...)`

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
lesaltzm, line 209: Features is interesting: this is more of a binding "mode" than a feature, perhaps it belongs in binding config rather than as a feature flag?

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -321,6 +322,13 @@ public override IntermediateNode Visit(FirstNameNode node, IRTranslatorContext c
                 Contracts.AssertValue(node);
                 Contracts.AssertValue(context);
 
+                var nodeType = context.Binding.GetType(node);
+
+                if (nodeType.IsUnknown)
```
lesaltzm, line 327: Should we check and fail earlier (in Translate()) if the binding was done in the "deferred" mode?

src/strings/PowerFxResources.en-US.resx
```diff
@@ -2160,6 +2160,10 @@
     <value>Warning: Select "Capture Schema" at the bottom of the expanded formula bar to set and refresh this method's result schema. Otherwise this method will return no result.</value>
     <comment>Warning given when service function returns dynamic metadata.</comment>
   </data>
+  <data name="WarnUnknownType" xml:space="preserve">
+    <value>Warning: unknown type provided.</value>
```
lesaltzm, line 2164: nit: `Unknown`

lesaltzm:
My primary concern is that it's something now that every change/error added to the binder needs to be aware of, and another place to be concerned about breaking changes: previously, all these cases were errors, and we can make breaking changes to error cases without back compat concerns. Does this change that contract?

src/tests/Microsoft.PowerFx.Interpreter.Tests/DefeeredTypeTests.cs
```diff
@@ -0,0 +1,95 @@
+﻿// Copyright (c) Microsoft Corporation.
```
susosuso, line 1: NP: file name :)

src/tests/Microsoft.PowerFx.Interpreter.Tests/DefeeredTypeTests.cs
jas-valgotar: In current iteration, we are not allowing deferred type on functions which takes formula as an input in any arg.

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
jas-valgotar: Added a Binder config for this now.

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
jas-valgotar: We can't, because even with deferred mode on user can choose to not pass any deferred args.

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
lesaltzm: Disagree, even if the user didn’t reference a deferred arg, binding in deferred mode not being able to generate IR is a useful invariant.

susosuso:
Thank you for making the change! I have a few clarification questions just to make sure I understand how we may use this.

1. At authoring time, we spin up an Engine with PowerFxConfig's feature flag enabled.
2. Any type that is not available at this time, we pass it in as FormulaType.Unknown
3. The Check() performed at this stage here is only for this partial validation, and should not be used for downstream Eval()

MikeStall:
@susosuso  - yes. 
Note that the feature flags are all temporary - eventually they're always be on by default.

src/libraries/Microsoft.PowerFx.Core/Parser/ParserOptions.cs
```diff
@@ -15,6 +15,11 @@ namespace Microsoft.PowerFx
     /// </summary>
     public class ParserOptions
     {
+        /// <summary>
+        /// If true, This would allow users to do partial type checking by enabling support for Unknown type in input.
+        /// </summary>
+        public bool AllowDeferredType { get; set; }
```
lesaltzm, line 21: Why does this need to be in ParserOptions? The parser shouldn't care about type checking features.

MikeStall:
We should close this in favor of the approach in #823 .

MikeStall:
Closing in favor of #823 .

PR #798, opened 2022-11-01
Add record and replay feature to LimitedSeverityErrorContainer

bryancroteau-MSFT:
Adds a "record and replay" feature to the error container that is being used by TryGetBestOverload. This is so that, when functions want to use CheckTypes/CheckSemantics instead of CheckInvocation, semantic errors that do not contribute to a type checking error can be "replayed" in the case where type checking succeeds.

A new error container is created per-function, so only the errors generated for a single function are added to the list.

This should greatly reduce the number of failures in PAClient test labs as we migrate functions to use the new methods.

PR #823, opened 2022-11-09
Adds support for partial input type checking

### LINKED ISSUE #720: Add the ability to perform type checking validation for partial input

jas-valgotar:
- Fixes #720 

**Assumptions on Deferred Type:**
1. You can't construct deferred type around aggregate type e.g. Record type, TableType
2. Function calls discard all the errors if any arg present is deferred type. e.g. Sum(deferred, Record) would not give any error for Record Type.
3. Function Overload resolution is biased to scaler type, because we discard error and make the first overload valid it matches to scaler one and does not try and go to Tabular overload. e.g. Sum(deferred) return type is number, and not the table of numbers.
**Need:**
Some teams would like to perform partial type checking at authoring time.
e.g. `deferred + number => valid` but, `deferred + record => error`

**Input type consideration:**
1. Introduce new type (e.g. deferredType) 
This could work, but we might miss some cases along the Binder and cause the issue. Plus, it would require lot of changes.
2. Use an existing type if feasible:
Type Considered:
1._UtypedObject or Error as deferred type_: The Pro is these types of nodes can already be generated by user input so we could leverage already existing checks. But The issue here is we cannot differentiate between if user intended to use `actual UntypedObject` or `UntypedObject as deferred type`. And for error we can't distinguish between the `actual Error` or `Error as deferred type`.
Also, all existing type would not accept this type for type checking.
\
2._Unknown as deferred type_: This type looks promising, as this type is already accepted by All Primitive types and hence can be used as `deferred type`. Also, its being currently used only as return type of few functions like `If`, `Coalesce` and hence there are no clashes on input as no function currently accepts `UnknownType` as input. So that makes it easy to distinguish that user did intend to use it as `deferred type`. As it does pretty well satisfy all we need, but has potential to brake the PA.
\
3._Introducing new DKind_: Since all the exisiting types did not satisfy our needs we have decided to inroduce a new DKind called `Deferred`.

**Implementation Consideration for function calls:**
1. Changes in CheckTypes:
Pro to this approach being we can achieve a true type checking for all the function. But that would require changes in 100s of places and could turn into headache to maintain and can potentially introduce bugs.

2. Post process the error that was generated by CheckTypes:
In this approach, we would check all the errors generated by CheckTypes and if all of them were generated due to `Deferred` Type argument and if indeed they were, we consider the type checking to be successful and discard all the errors. This would require changes in only few places in binder. As this would be a low risk, we decided to go with this approach.

**What users will achieve?**

Using this `Deferred` type. User will be able to partially type check expressions at authoring time and will get warnings where `Deferred` type is used. So, if schema of any argument X is deferred, user can still get meaningful error for other arguments whose types are known.
e.g.
`Deferred + 1` would just give a warning at compile time.
`Deferred + Record` would fail as record is not a valid input.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/App/ErrorContainers/ErrorContainer.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/App/ErrorContainers/ErrorContainer.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using System.Collections.Generic;
+using System.Linq;
 using Microsoft.PowerFx.Core.Errors;
 using Microsoft.PowerFx.Core.Localization;
 using Microsoft.PowerFx.Core.Types;
@@ -16,12 +17,14 @@ namespace Microsoft.PowerFx.Core.App.ErrorContainers
 
         public DocumentErrorSeverity DefaultSeverity => DocumentErrorSeverity.Critical;
 
-        public void ConcatErrors(IEnumerable<TexlError> errors)
+        public void MergeErrors(IEnumerable<TexlError> errors)
         {
             if (_errors == null)
             {
                 _errors = new List<TexlError>();
             }
+
+            errors = errors.Where(e => !HasErrors(e.Node, e.Severity)).ToList();
 
             _errors.AddRange(errors);
         }
--- a/src/libraries/Microsoft.PowerFx.Core/App/ErrorContainers/ErrorContainer.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/App/ErrorContainers/ErrorContainer.cs
@@ -16,6 +16,16 @@ namespace Microsoft.PowerFx.Core.App.ErrorContainers
 
         public DocumentErrorSeverity DefaultSeverity => DocumentErrorSeverity.Critical;
 
+        public void ConcatErrors(IEnumerable<TexlError> errors)
+        {
+            if (_errors == null)
+            {
+                _errors = new List<TexlError>();
+            }
+
+            _errors.AddRange(errors);
+        }
+
         public bool HasErrors()
         {
             return CollectionUtils.Size(_errors) > 0;
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -4360,13 +4360,14 @@ namespace Microsoft.PowerFx.Core.Binding
                 _currentScope = scopeNew.Parent;
                 PostVisit(node.Args);
 
-                // Typecheck the invocation.
+                // Temporary error container which can be discarded if deferred type arg is present.
+                var checkErrorContainer = new ErrorContainer();
 
                 // Typecheck the invocation and infer the return type.
-                fArgsValid &= maybeFunc.HandleCheckInvocation(_txb, args, argTypes, _txb.ErrorContainer, out var returnType, out var nodeToCoercedTypeMap);
+                fArgsValid &= maybeFunc.HandleCheckInvocation(_txb, args, argTypes, checkErrorContainer, out var returnType, out var nodeToCoercedTypeMap);
 
                 // If type check failed and errors were due to Unknown type arg we would like to consider the typeChecking passed and discard all the errors.
-                (fArgsValid, returnType) = CheckDeferredType(argTypes, returnType, fArgsValid);
+                (fArgsValid, returnType) = CheckDeferredType(argTypes, returnType, fArgsValid, checkErrorContainer, _txb.ErrorContainer);
 
                 // This is done because later on, if a CallNode has a return type of Error, you can assert HasErrors on it.
                 // This was not done for UnaryOpNodes, BinaryOpNodes, CompareNodes.
@@ -4688,11 +4689,14 @@ namespace Microsoft.PowerFx.Core.Binding
                 var argTypes = args.Select(_txb.GetType).ToArray();
                 bool fArgsValid;
 
+                // Temporary error container which can be discarded if deferred type arg is present.
+                var checkErrorContainer = new ErrorContainer();
+
                 // Typecheck the invocation and infer the return type.
-                fArgsValid = func.HandleCheckInvocation(_txb, args, argTypes, _txb.ErrorContainer, out returnType, out _);
+                fArgsValid = func.HandleCheckInvocation(_txb, args, argTypes, checkErrorContainer, out returnType, out _);
 
                 // If type check failed and errors were due to Unknown type arg we would like to consider the typeChecking passed and discard all the errors.
-                (fArgsValid, returnType) = CheckDeferredType(argTypes, returnType, fArgsValid);
+                (fArgsValid, returnType) = CheckDeferredType(argTypes, returnType, fArgsValid, checkErrorContainer, _txb.ErrorContainer);
 
                 if (!fArgsValid)
                 {
@@ -4860,20 +4864,13 @@ namespace Microsoft.PowerFx.Core.Binding
 
                 // This error container is used as temporary container so we can trap type mismatch kind of error for
                 // deferred (unknown) type args and validate all the errors were caused due to deferred(unknown) type.
-                var checkInvocationErrors = new ErrorContainer();
+                var checkErrorContainer = new ErrorContainer();
 
                 // Typecheck the invocation and infer the return type.
-                fArgsValid = func.HandleCheckInvocation(_txb, args, argTypes, checkInvocationErrors, out returnType, out var nodeToCoercedTypeMap);
-
-                var isDeferredArgPresent = argTypes.Any(type => type.IsDeferred);
+                fArgsValid = func.HandleCheckInvocation(_txb, args, argTypes, checkErrorContainer, out returnType, out var nodeToCoercedTypeMap);
 
                 // If type check failed and errors were due to Unknown type arg we would like to consider the typeChecking passed and discard all the errors.
-                (fArgsValid, returnType) = CheckDeferredType(argTypes, returnType, fArgsValid);
-
-                if(!isDeferredArgPresent)
-                {
-                    _txb.ErrorContainer.MergeErrors(checkInvocationErrors.GetErrors());
-                }
+                (fArgsValid, returnType) = CheckDeferredType(argTypes, returnType, fArgsValid, checkErrorContainer, _txb.ErrorContainer);
 
                 if (!fArgsValid && !func.HasPreciseErrors)
                 {
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -4861,20 +4861,9 @@ namespace Microsoft.PowerFx.Core.Binding
                 var isDeferredArgPresent = argTypes.Any(type => type.IsDeferred);
 
                 // If type check failed and errors were due to Unknown type node we would like to consider the typeChecking passed and discard all the errors.
-                if (!fArgsValid && isDeferredArgPresent)
-                {
-                    fArgsValid = true;
+                (fArgsValid, returnType) = CheckDeferredType(argTypes, returnType, fArgsValid);
 
-                    // If one of the arg was deferred and
-                    // return type could not be calculated and was error, we assign it to deferred as safeguard.
-                    // returnType was EmptyTable, we assign it to deferred as safeguard e.g. Table(Deferred) => deferred,
-                    // this is because we don't want to embed deferred type inside of any aggregate type.
-                    if (returnType.IsError || returnType.Equals(DType.EmptyTable))
-                    {
-                        returnType = DType.Deferred;
-                    }
-                }
-                else
+                if(!isDeferredArgPresent)
                 {
                     _txb.ErrorContainer.MergeErrors(checkInvocationErrors.GetErrors());
                 }
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -4852,11 +4852,10 @@ namespace Microsoft.PowerFx.Core.Binding
                 bool fArgsValid;
 
                 // Typecheck the invocation and infer the return type.
-                // fArgsValid = func.HandleCheckInvocation(_txb, args, argTypes, _txb.ErrorContainer, out returnType, out var nodeToCoercedTypeMap);
 
                 fArgsValid = HandleCheckInvocationWithDeferred(func, _txb, args, argTypes, out var checkInvocationErrors, out returnType, out var nodeToCoercedTypeMap);
 
-                if (!fArgsValid)
+                if (checkInvocationErrors.HasErrors())
                 {
                     _txb.ErrorContainer.MergeErrors(checkInvocationErrors.GetErrors());
                 }
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -3442,6 +3442,11 @@ namespace Microsoft.PowerFx.Core.Binding
                     _txb.FlagPathAsAsync(node);
                 }
 
+                if (typeRhs.IsDeferred)
+                {
+                    _txb.ErrorContainer.EnsureError(DocumentErrorSeverity.Warning, node, TexlStrings.WarnDeferredType);
+                }
+
                 // Set the type for the dotted node itself.
                 if (leftType.IsEnum)
                 {
@@ -4849,11 +4854,11 @@ namespace Microsoft.PowerFx.Core.Binding
                 // Typecheck the invocation and infer the return type.
                 // fArgsValid = func.HandleCheckInvocation(_txb, args, argTypes, _txb.ErrorContainer, out returnType, out var nodeToCoercedTypeMap);
 
-                fArgsValid = HandleCheckInvocationWithUnknown(func, _txb, args, argTypes, out var checkInvocationErrors, out returnType, out var nodeToCoercedTypeMap);
+                fArgsValid = HandleCheckInvocationWithDeferred(func, _txb, args, argTypes, out var checkInvocationErrors, out returnType, out var nodeToCoercedTypeMap);
 
                 if (!fArgsValid)
                 {
-                    _txb.ErrorContainer.ConcatErrors(checkInvocationErrors.GetErrors());
+                    _txb.ErrorContainer.MergeErrors(checkInvocationErrors.GetErrors());
                 }
 
                 if (!fArgsValid && !func.HasPreciseErrors)
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -4365,6 +4365,9 @@ namespace Microsoft.PowerFx.Core.Binding
                 // Typecheck the invocation and infer the return type.
                 fArgsValid &= maybeFunc.HandleCheckInvocation(_txb, args, argTypes, _txb.ErrorContainer, out var returnType, out var nodeToCoercedTypeMap);
 
+                // If type check failed and errors were due to Unknown type arg we would like to consider the typeChecking passed and discard all the errors.
+                (fArgsValid, returnType) = CheckDeferredType(argTypes, returnType, fArgsValid);
+
                 // This is done because later on, if a CallNode has a return type of Error, you can assert HasErrors on it.
                 // This was not done for UnaryOpNodes, BinaryOpNodes, CompareNodes.
                 // This doesn't need to be done on the other nodes (but can) because their return type doesn't depend
@@ -4687,6 +4690,10 @@ namespace Microsoft.PowerFx.Core.Binding
 
                 // Typecheck the invocation and infer the return type.
                 fArgsValid = func.HandleCheckInvocation(_txb, args, argTypes, _txb.ErrorContainer, out returnType, out _);
+
+                // If type check failed and errors were due to Unknown type arg we would like to consider the typeChecking passed and discard all the errors.
+                (fArgsValid, returnType) = CheckDeferredType(argTypes, returnType, fArgsValid);
+
                 if (!fArgsValid)
                 {
                     _txb.ErrorContainer.Error(DocumentErrorSeverity.Severe, node, TexlStrings.ErrInvalidArgs_Func, func.Name);
@@ -4860,7 +4867,7 @@ namespace Microsoft.PowerFx.Core.Binding
 
                 var isDeferredArgPresent = argTypes.Any(type => type.IsDeferred);
 
-                // If type check failed and errors were due to Unknown type node we would like to consider the typeChecking passed and discard all the errors.
+                // If type check failed and errors were due to Unknown type arg we would like to consider the typeChecking passed and discard all the errors.
                 (fArgsValid, returnType) = CheckDeferredType(argTypes, returnType, fArgsValid);
 
                 if(!isDeferredArgPresent)
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -2802,6 +2802,11 @@ namespace Microsoft.PowerFx.Core.Binding
                     }
                 }
 
+                if (lookupType.IsDeferred)
+                {
+                    _txb.ErrorContainer.EnsureError(DocumentErrorSeverity.Warning, node, TexlStrings.WarnDeferredType);
+                }
+
                 // Make a note of this global's type, as identifier by the resolver.
                 _txb.SetType(node, lookupType);
 
@@ -3166,7 +3171,7 @@ namespace Microsoft.PowerFx.Core.Binding
 
                 var leftType = _txb.GetType(node.Left);
 
-                if (!leftType.IsControl && !leftType.IsAggregate && !leftType.IsEnum && !leftType.IsOptionSet && !leftType.IsView && !leftType.IsUntypedObject)
+                if (!leftType.IsControl && !leftType.IsAggregate && !leftType.IsEnum && !leftType.IsOptionSet && !leftType.IsView && !leftType.IsUntypedObject && !leftType.IsDeferred)
                 {
                     SetDottedNameError(node, TexlStrings.ErrInvalidDot);
                     return;
@@ -3367,7 +3372,7 @@ namespace Microsoft.PowerFx.Core.Binding
                         _txb.FlagPathAsAsync(node);
                     }
                 }
-                else if (!leftType.TryGetType(nameRhs, out typeRhs) && !leftType.IsUntypedObject)
+                else if (!leftType.TryGetType(nameRhs, out typeRhs) && !leftType.IsUntypedObject && !leftType.IsDeferred)
                 {
                     // We may be in the case of dropDown!Selected!RHS
                     // In this case, Selected embeds a meta field whose v-type encapsulates localization info
@@ -3456,6 +3461,10 @@ namespace Microsoft.PowerFx.Core.Binding
                 else if (leftType.IsUntypedObject)
                 {
                     _txb.SetType(node, DType.UntypedObject);
+                }
+                else if (leftType.IsDeferred)
+                {
+                    _txb.SetType(node, DType.Deferred);
                 }
                 else if (leftType.IsTable)
                 {
@@ -4838,7 +4847,14 @@ namespace Microsoft.PowerFx.Core.Binding
                 bool fArgsValid;
 
                 // Typecheck the invocation and infer the return type.
-                fArgsValid = func.HandleCheckInvocation(_txb, args, argTypes, _txb.ErrorContainer, out returnType, out var nodeToCoercedTypeMap);
+                // fArgsValid = func.HandleCheckInvocation(_txb, args, argTypes, _txb.ErrorContainer, out returnType, out var nodeToCoercedTypeMap);
+
+                fArgsValid = HandleCheckInvocationWithUnknown(func, _txb, args, argTypes, out var checkInvocationErrors, out returnType, out var nodeToCoercedTypeMap);
+
+                if (!fArgsValid)
+                {
+                    _txb.ErrorContainer.ConcatErrors(checkInvocationErrors.GetErrors());
+                }
 
                 if (!fArgsValid && !func.HasPreciseErrors)
                 {
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -4851,11 +4851,30 @@ namespace Microsoft.PowerFx.Core.Binding
                 var argTypes = args.Select(_txb.GetType).ToArray();
                 bool fArgsValid;
 
+                // This error container is used as temporary container so we can trap type mismatch kind of error for
+                // deferred (unknown) type args and validate all the errors were caused due to deferred(unknown) type.
+                var checkInvocationErrors = new ErrorContainer();
+
                 // Typecheck the invocation and infer the return type.
+                fArgsValid = func.HandleCheckInvocation(_txb, args, argTypes, checkInvocationErrors, out returnType, out var nodeToCoercedTypeMap);
 
-                fArgsValid = HandleCheckInvocationWithDeferred(func, _txb, args, argTypes, out var checkInvocationErrors, out returnType, out var nodeToCoercedTypeMap);
+                var isDeferredArgPresent = argTypes.Any(type => type.IsDeferred);
 
-                if (checkInvocationErrors.HasErrors())
+                // If type check failed and errors were due to Unknown type node we would like to consider the typeChecking passed and discard all the errors.
+                if (!fArgsValid && isDeferredArgPresent)
+                {
+                    fArgsValid = true;
+
+                    // If one of the arg was deferred and
+                    // return type could not be calculated and was error, we assign it to deferred as safeguard.
+                    // returnType was EmptyTable, we assign it to deferred as safeguard e.g. Table(Deferred) => deferred,
+                    // this is because we don't want to embed deferred type inside of any aggregate type.
+                    if (returnType.IsError || returnType.Equals(DType.EmptyTable))
+                    {
+                        returnType = DType.Deferred;
+                    }
+                }
+                else
                 {
                     _txb.ErrorContainer.MergeErrors(checkInvocationErrors.GetErrors());
                 }
@@ -5059,6 +5078,10 @@ namespace Microsoft.PowerFx.Core.Binding
                         {
                             _txb.ErrorContainer.EnsureError(DocumentErrorSeverity.Severe, node.Children[i], TexlStrings.ErrMultipleValuesForField_Name, displayName);
                         }
+                        else if (_txb.GetType(node.Children[i]).IsDeferred)
+                        {
+                            _txb.ErrorContainer.EnsureError(DocumentErrorSeverity.Severe, node.Children[i], TexlStrings.ErrRecordDoesNotAcceptThisType);
+                        }
                         else
                         {
                             nodeType = nodeType.Add(fieldName, _txb.GetType(node.Children[i]));
@@ -5084,15 +5107,15 @@ namespace Microsoft.PowerFx.Core.Binding
                     var childType = _txb.GetType(child);
                     isSelfContainedConstant &= _txb.IsSelfContainedConstant(child);
 
-                    if (!exprType.IsValid)
+                    if (!childType.IsDeferred && !exprType.IsValid)
                     {
                         exprType = childType;
                     }
-                    else if (exprType.CanUnionWith(childType))
+                    else if (!childType.IsDeferred && exprType.CanUnionWith(childType))
                     {
                         exprType = DType.Union(exprType, childType);
                     }
-                    else if (childType.CoercesTo(exprType))
+                    else if (!childType.IsDeferred && childType.CoercesTo(exprType))
                     {
                         _txb.SetCoercedType(child, exprType);
                     }
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
@@ -138,8 +138,20 @@ namespace Microsoft.PowerFx.Core.Binding
 
                 var localWarnings = new LimitedSeverityErrorContainer(errors, DocumentErrorSeverity.Warning);
 
+                var checkTypeErrors = new ErrorContainer();
+
                 // Typecheck the invocation and infer the return type.
-                typeCheckSucceeded = maybeFunc.CheckTypes(context, args, argTypes, localWarnings, out returnType, out nodeToCoercedTypeMap);
+                typeCheckSucceeded = maybeFunc.CheckTypes(context, args, argTypes, checkTypeErrors, out returnType, out nodeToCoercedTypeMap);
+
+                // Adding back all the errors to local warnings
+                foreach (var error in checkTypeErrors.GetErrors())
+                {
+                    //// Only add error for nodes apart from Unknown.
+                    //if (!txb.GetType(error.Node).IsDeferred)
+                    //{
+                    //    localWarnings.EnsureError(error.Node, error.ErrorResourceKey);
+                    //}
+                }
 
                 if (typeCheckSucceeded)
                 {
@@ -178,6 +190,39 @@ namespace Microsoft.PowerFx.Core.Binding
             nodeToCoercedTypeMap = null;
             returnType = null;
             return false;
+        }
+
+        /// <summary>
+        /// Typecheck the invocation and infer the return type. Error on deferred(unknown) arg is not considered.
+        /// </summary>
+        internal static bool HandleCheckInvocationWithUnknown(TexlFunction maybeFunc, TexlBinding txb, TexlNode[] args, DType[] argTypes, out ErrorContainer checkInvocationErrors, out DType returnType, out Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
+        {
+            // This error container is used as temporary container so we can trap type mismatch kind of error for
+            // deferred (unknown) type args and validate all the errors were caused due to deferred(unknown) type.
+            checkInvocationErrors = new ErrorContainer();
+
+            var typeCheckSucceeded = maybeFunc.HandleCheckInvocation(txb, args, argTypes, checkInvocationErrors, out returnType, out nodeToCoercedTypeMap);
+
+            // If type check failed and errors were due to Unknown type node we would like to consider the typeChecking passed.
+            if (!typeCheckSucceeded && checkInvocationErrors.HasErrors() && checkInvocationErrors.GetErrors().All(error => txb.GetType(error.Node).IsDeferred))
+            {
+                // If one of the arg was unknown and that generated error (e.g. type mismatch)
+                // and return type could not be calculated and was error we assign it as unknown.
+                // and if return type was Table, we assign it to be table of unknown, so operation like In can work.
+                switch (returnType.Kind)
+                {
+                    case DKind.Error:
+                        returnType = DType.Deferred;
+                        break;
+                    case DKind.Table:
+                        returnType = DType.EmptyRecord.Add(new TypedName(DType.Deferred, new DName(TexlFunction.ColumnName_ValueStr))).ToTable();
+                        break;
+                }
+
+                return true;
+            }
+
+            return typeCheckSucceeded;
         }
 
         /// <summary>
@@ -665,7 +710,7 @@ namespace Microsoft.PowerFx.Core.Binding
 
             // EqualOp is only allowed on primitive types, polymorphic lookups, and control types.
             if (!(typeLeft.IsPrimitive && typeRight.IsPrimitive) && !(typeLeft.IsPolymorphic && typeRight.IsPolymorphic) && !(typeLeft.IsControl && typeRight.IsControl)
-                && !(typeLeft.IsPolymorphic && typeRight.IsRecord) && !(typeLeft.IsRecord && typeRight.IsPolymorphic))
+                && !(typeLeft.IsPolymorphic && typeRight.IsRecord) && !(typeLeft.IsRecord && typeRight.IsPolymorphic) && !(typeLeft.IsDeferred || typeRight.IsDeferred))
             {
                 var leftTypeDisambiguation = typeLeft.IsOptionSet && typeLeft.OptionSetInfo != null ? $"({typeLeft.OptionSetInfo.EntityName})" : string.Empty;
                 var rightTypeDisambiguation = typeRight.IsOptionSet && typeRight.OptionSetInfo != null ? $"({typeRight.OptionSetInfo.EntityName})" : string.Empty;
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
@@ -185,7 +185,16 @@ namespace Microsoft.PowerFx.Core.Binding
             return false;
         }
 
-        internal static (bool typeCheckSucceeded, DType returnType) CheckDeferredType(DType[] argTypes, DType returnType, bool typeCheckSucceeded)
+        /// <summary>
+        /// if <paramref name="typeCheckSucceeded"/> failed and DeferredType arg is present, discard all the error and succeed the type check.
+        /// Keep <paramref name="errorContainer"/> and <paramref name="checkErrorContainer"/> null if no need to merge errors.
+        /// </summary>
+        /// <param name="argTypes"> types of all the args.</param>
+        /// <param name="returnType"> return type determined by function's check type.</param>
+        /// <param name="typeCheckSucceeded"> function's check type succeeded or not.</param>
+        /// <param name="checkErrorContainer"> Temp error container used for type checking only.</param>
+        /// <param name="errorContainer"> Binder's error container.</param>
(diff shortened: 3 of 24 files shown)


src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -321,6 +321,13 @@ public override IntermediateNode Visit(FirstNameNode node, IRTranslatorContext c
                 Contracts.AssertValue(node);
                 Contracts.AssertValue(context);
 
+                var nodeType = context.Binding.GetType(node);
```
MikeStall, line 324: also check a field on dotted name node.

MikeStall:
Also enforce that custom functions can't add deferred types.

src/libraries/Microsoft.PowerFx.Core/Public/Types/DeferredType.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Types
+{
+    /// <summary>
+    /// An unknown type, attached to syntax nodes whose type cannot be determined.
```
lesaltzm, line 10: consider also commenting that it's not valid to even attempt to execute an expression that references a deferred type

src/tests/Microsoft.PowerFx.Core.Tests/TypeSystemTests/DTypeTests.cs
```diff
@@ -205,7 +208,36 @@ public void UnknownIsSubtypeOfAll()
             Assert.True(MultiSelectOptionSetType.Accepts(DType.Unknown));
             Assert.True(DType.Polymorphic.Accepts(DType.Unknown));
         }
-        
+
+        [Fact]
+        public void DeferredIsSubtypeOfAll()
```
lesaltzm, line 213: Also good to test other type operations:
Supertype, Union, CoercesTo, ...

and operations where deferred is a field of an aggregate (table/record)

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -185,6 +197,39 @@ internal static bool TryGetBestOverload(TexlBinding txb, CallNode node, DType[]
...
+            checkInvocationErrors = new ErrorContainer();
+
+            var typeCheckSucceeded = maybeFunc.HandleCheckInvocation(txb, args, argTypes, checkInvocationErrors, out returnType, out nodeToCoercedTypeMap);
+
+            // If type check failed and errors were due to Unknown type node we would like to consider the typeChecking passed.
+            if (!typeCheckSucceeded && checkInvocationErrors.HasErrors() && checkInvocationErrors.GetErrors().All(error => txb.GetType(error.Node).IsDeferred))
+            {
+                // If one of the arg was unknown and that generated error (e.g. type mismatch)
+                // and return type could not be calculated and was error we assign it as unknown.
+                // and if return type was Table, we assign it to be table of unknown, so operation like In can work.
```
lesaltzm, line 216: nit: table of deferred

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -185,6 +197,39 @@ internal static bool TryGetBestOverload(TexlBinding txb, CallNode node, DType[]
...
+                switch (returnType.Kind)
+                {
+                    case DKind.Error:
+                        returnType = DType.Deferred;
+                        break;
+                    case DKind.Table:
```
lesaltzm, line 222: is there any reason we'd care about `Record` results here?

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -140,7 +140,19 @@ internal static bool TryGetBestOverload(TexlBinding txb, CallNode node, DType[]
                 var localWarnings = new LimitedSeverityErrorContainer(txb.ErrorContainer, DocumentErrorSeverity.Warning);
 
                 // Typecheck the invocation and infer the return type.
-                typeCheckSucceeded = maybeFunc.HandleCheckInvocation(txb, args, argTypes, localWarnings, out returnType, out nodeToCoercedTypeMap);
+                //typeCheckSucceeded = maybeFunc.HandleCheckInvocation(txb, args, argTypes, localWarnings, out returnType, out nodeToCoercedTypeMap);
+
+                typeCheckSucceeded = HandleCheckInvocationWithDeferred(maybeFunc, txb, args, argTypes, out var checkInvocationErrors, out returnType, out nodeToCoercedTypeMap);
+
+                // Adding back all the errors to local warnings
+                foreach (var error in checkInvocationErrors.GetErrors())
```
lesaltzm, line 148: Should this use the record/replay feature bryan added in #798 ?

lesaltzm: Looks like this is creating 2 layers of throwaway error containers, can it be more efficient?

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -140,7 +140,19 @@ internal static bool TryGetBestOverload(TexlBinding txb, CallNode node, DType[]
                 var localWarnings = new LimitedSeverityErrorContainer(txb.ErrorContainer, DocumentErrorSeverity.Warning);
 
                 // Typecheck the invocation and infer the return type.
-                typeCheckSucceeded = maybeFunc.HandleCheckInvocation(txb, args, argTypes, localWarnings, out returnType, out nodeToCoercedTypeMap);
+                //typeCheckSucceeded = maybeFunc.HandleCheckInvocation(txb, args, argTypes, localWarnings, out returnType, out nodeToCoercedTypeMap);
+
+                typeCheckSucceeded = HandleCheckInvocationWithDeferred(maybeFunc, txb, args, argTypes, out var checkInvocationErrors, out returnType, out nodeToCoercedTypeMap);
```
MikeStall, line 145: >HandleCheckInvocationWithDeferred [](http://example.com/codeflow?start=37&length=33)

Tracking for future improvement once dataflow is ready: 
[https://github.com/microsoft/Power-Fx/issues/837](https://github.com/microsoft/Power-Fx/issues/837)

src/tests/Microsoft.PowerFx.Core.Tests/TypeSystemTests/DTypeTests.cs
jas-valgotar: Since DeferredType is SubType of all and we have a test for that. SuperType, Union would be redundant. DefeeredType is not usable pass Type checking so there would be no need to check CoercesTo either...

jas-valgotar: Since DeferredType is SubType of all and we have a test for that. SuperType, Union would be redundant. DefeeredType is not usable pass Type checking so there would be no need to check CoercesTo either...

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
jas-valgotar: That is so operator like `In` can work.
e.g. 123 in Table(X) would only work if Table(X) resolves to single column deferred type table. Because then type of one column table can be accepted by number type.

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.3-preview.20221116-329654](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.3-preview.20221116-329654).
<details>

```
InterfacesShouldHaveSameMembers : Interface member 'Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.DeferredType)' is present in the implementation but not in the contract.
```

</details>

src/tests/Microsoft.PowerFx.Interpreter.Tests/DeferredTypeTests.cs
```diff
@@ -0,0 +1,122 @@
...
+            TestDeferredTypeBindingWarning(script, Features.None, symbolTable);
+        }
+
+        [Theory]
+        [InlineData("$\"Test {X} {R}\"", "Invalid argument type (Record). Expecting a Text value instead.")]
+        [InlineData("X + R", "Invalid argument type. Expecting one of the following: Number, Text, Boolean.")]
+        [InlineData("Table(X, N)", "Cannot use a non-record value in this context")]
+        [InlineData("X.field + N.field", "Invalid use of '.'")]
+        [InlineData("Index([1,2,3], X).missing", "Name isn't valid. 'missing' isn't recognized")]
+
```
susosuso, line 75: Can we consider adding these tests as well? (just to cover some of our common use cases)

`"X < \"2021-12-09T20:28:52Z\""`
`"true And X.field Or "string""`

src/tests/Microsoft.PowerFx.Interpreter.Tests/DeferredTypeTests.cs
```diff
@@ -0,0 +1,122 @@
...
+        [InlineData("X + 1")]
+        [InlineData("X + \"1\"")]
+        [InlineData("X + DateTime(2022, 11, 10, 0, 0, 0)")]
+        [InlineData("X + Date(2022, 11, 10)")]
+        [InlineData("X + Time(0, 0, 0)")]
+        [InlineData("1 + X")]
+        [InlineData("DateTime(2022, 11, 10, 0, 0, 0) + X")]
+        [InlineData("Date(2022, 11, 10) + X")]
+        [InlineData("Time(0, 0, 0) + X")]
```
susosuso, line 34: Consider adding `X < DateTime(2022,11,10,0,0,0)` to cover temporal comparison

src/tests/Microsoft.PowerFx.Core.Tests/TypeSystemTests/DTypeTests.cs
lesaltzm: specifically add tests that cover that, then

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.3-preview.20221118-329922](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.3-preview.20221118-329922).
<details>

```
InterfacesShouldHaveSameMembers : Interface member 'Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.DeferredType)' is present in the implementation but not in the contract.
```

</details>

src/tests/Microsoft.PowerFx.Interpreter.Tests/DeferredTypeTests.cs
jas-valgotar: `"true And X.field Or \"string\""` is considered syntactically valid expression since we coerce string param to Boolean.
e.g. `false Or "true"` => `true`
Added test for other one.

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.3-preview.20221118-329922](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.3-preview.20221118-329922).
<details>

```
InterfacesShouldHaveSameMembers : Interface member 'Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.DeferredType)' is present in the implementation but not in the contract.
```

</details>

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -138,8 +139,41 @@ internal static bool TryGetBestOverload(CheckTypesContext context, IErrorContain
...
                 var localWarnings = new LimitedSeverityErrorContainer(errors, DocumentErrorSeverity.Warning);
 
+                var checkTypeErrors = new ErrorContainer();
+
                 // Typecheck the invocation and infer the return type.
-                typeCheckSucceeded = maybeFunc.CheckTypes(context, args, argTypes, localWarnings, out returnType, out nodeToCoercedTypeMap);
+                typeCheckSucceeded = maybeFunc.CheckTypes(context, args, argTypes, checkTypeErrors, out returnType, out nodeToCoercedTypeMap);
+
+                if (!typeCheckSucceeded && checkTypeErrors.HasErrors() &&
+                    checkTypeErrors.GetErrors().All(error => context.NameResolver.Lookup(new DName(error.Tok.ToString()), out NameLookupInfo nameLookup) && nameLookup.Type.IsDeferred))
```
lesaltzm, line 148: Why are we iterating the error container/calling Lookup?
We have the types of the args in `argTypes`

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -138,8 +139,41 @@ internal static bool TryGetBestOverload(CheckTypesContext context, IErrorContain
...
+
+                // Adding back all the errors to local warnings
+                foreach (var error in checkTypeErrors.GetErrors())
+                {
+                    var isLookupSuccess = context.NameResolver.Lookup(new DName(error.Tok.ToString()), out var nameLookupInfo);
+
+                    // Only add error for nodes apart from Unknown type argument.
+                    if (!isLookupSuccess || nameLookupInfo.Type.IsDeferred)
+                    {
+                        localWarnings.EnsureError(error.Node, error.ErrorResourceKey);
```
lesaltzm, line 174: This could throw, what if the error is a format string? Looks like it loses the rest of the error context

src/tests/Microsoft.PowerFx.Interpreter.Tests/DeferredTypeTests.cs
```diff
@@ -0,0 +1,124 @@
...
+        [InlineData("123 in Table(X)")]
+        [InlineData("IsBlank(X)")]
+        [InlineData("IsError(X)")]
+        [InlineData("Text(X)")]
+        [InlineData("Value(X)")]
+        [InlineData("Boolean(X)")]
+        [InlineData("Index([1,2,3].Value, X)")]
+
+        // Ensures expression binds without any errors - but issues a warning for the deferred(unknown) type.
+        public void DeferredTypeTest(string script)
```
MikeStall, line 56: we should add some tests with overload resolution. 
Sum, Filter 

Sum(?,?) < 5 // succeed, with warning 
First(Sum(?,?)).Field // also succeed, with waring

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4838,7 +4852,13 @@ private void PreVisitBottomUp(CallNode node, int argCountVisited, Scope scopeNew
                 bool fArgsValid;
 
                 // Typecheck the invocation and infer the return type.
-                fArgsValid = func.HandleCheckInvocation(_txb, args, argTypes, _txb.ErrorContainer, out returnType, out var nodeToCoercedTypeMap);
+
+                fArgsValid = HandleCheckInvocationWithDeferred(func, _txb, args, argTypes, out var checkInvocationErrors, out returnType, out var nodeToCoercedTypeMap);
+
+                if (checkInvocationErrors.HasErrors())
```
MikeStall, line 4858: >if (checkInvocationErrors.HasErrors()) [](http://example.com/codeflow?start=16&length=38)

this this check necessary?

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.3-preview.20221122-330455](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.3-preview.20221122-330455).
<details>

```
InterfacesShouldHaveSameMembers : Interface member 'Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.DeferredType)' is present in the implementation but not in the contract.
```

</details>

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -61,6 +62,8 @@ internal void SetErrors(IEnumerable<IDocumentError> errors)
         /// </summary>
         public bool IsSuccess => !Errors.Any(x => !x.IsWarning);
 
+        public bool HasDeferredArgsWarning => Errors.Any(x => x.IsWarning && x.MessageKey.Equals(TexlStrings.WarnDeferredType.Key));
```
MikeStall, line 65: >public [](http://example.com/codeflow?start=8&length=6)

can this start as internal?

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4837,8 +4851,30 @@ private void PreVisitBottomUp(CallNode node, int argCountVisited, Scope scopeNew
                 var argTypes = args.Select(_txb.GetType).ToArray();
                 bool fArgsValid;
 
+                // This error container is used as temporary container so we can trap type mismatch kind of error for
+                // deferred (unknown) type args and validate all the errors were caused due to deferred(unknown) type.
+                var checkInvocationErrors = new ErrorContainer();
+
                 // Typecheck the invocation and infer the return type.
-                fArgsValid = func.HandleCheckInvocation(_txb, args, argTypes, _txb.ErrorContainer, out returnType, out var nodeToCoercedTypeMap);
+                fArgsValid = func.HandleCheckInvocation(_txb, args, argTypes, checkInvocationErrors, out returnType, out var nodeToCoercedTypeMap);
```
MikeStall, line 4859: >HandleCheckInvocation [](http://example.com/codeflow?start=34&length=21)

We have several other instances in this file calling HandleCheckInvocation.

Do they need to be updated too?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Table.cs
```diff
@@ -60,7 +60,7 @@ public override bool CheckTypes(TexlNode[] args, DType[] argTypes, IErrorContain
             for (var i = 0; i < argTypes.Length; i++)
             {
                 var argType = argTypes[i];
-                if (!argType.IsRecord)
+                if (!(argType.IsRecord || argType.IsDeferred))
```
MikeStall, line 63: >|| argType.IsDeferred [](http://example.com/codeflow?start=39&length=21)

Do we need this?

Ie, why does Table() function need a special deferred check , but the other functions don't.

src/tests/Microsoft.PowerFx.Core.Tests/TypeSystemTests/DTypeTests.cs
MikeStall, line 213: >DeferredIsSubtypeOfAll [](http://example.com/codeflow?start=20&length=22)

Could we have a DType[] of all types and use a for-loop here?

This may also help with the above functions.

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
```diff
@@ -51,6 +52,8 @@ public abstract class FormulaType
 
         public static FormulaType Unknown { get; } = new UnknownType();
 
+        public static FormulaType Deferred { get; } = new DeferredType();
```
MikeStall, line 55: >Deferred [](http://example.com/codeflow?start=34&length=8)

Can we verify a custom function can't attempt to return one of these? 

More broadly, the challenge  is to ensure that anywhere these deferred types first "Show up" in an expression, we trap it and issue a warning. 

FirstName node is the most obvious place. 
DottedName node is a 2nd spot (since somebody may try to pass it in via a type (can we just fail this at cosntruction?)

CustomFunctions is another possible entry point.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
jas-valgotar: We don't need to update them in current interation, we may need to update them when we extend support of deferred to functions with lamda args

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Table.cs
jas-valgotar: Since Table() is the only function whose checkType is written in a way that it would not compute a correct return type if these changes are not made.

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.3-preview.20221122-330455](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.3-preview.20221122-330455).
<details>

```
InterfacesShouldHaveSameMembers : Interface member 'Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.DeferredType)' is present in the implementation but not in the contract.
```

</details>

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
jas-valgotar: Now Aggregate type fails at construction, Custom Function would not be an issue as there is no `DeferredValue` define.

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.3-preview.20221123-330655](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.3-preview.20221123-330655).
<details>

```
InterfacesShouldHaveSameMembers : Interface member 'Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.DeferredType)' is present in the implementation but not in the contract.
```

</details>

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.3-preview.20221123-330655](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.3-preview.20221123-330655).
<details>

```
InterfacesShouldHaveSameMembers : Interface member 'Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.DeferredType)' is present in the implementation but not in the contract.
```

</details>

src/libraries/Microsoft.PowerFx.Core/Public/Types/RecordType.cs
```diff
@@ -85,6 +85,11 @@ public virtual RecordType Add(NamedFormulaType field)
         /// </summary>
         public RecordType Add(string logicalName, FormulaType type, string optionalDisplayName = null)
         {
+            if (type._type.IsDeferred)
```
MikeStall, line 88: > if (type._type.IsDeferred) [](http://example.com/codeflow?start=11&length=27)

nit - this check should actually be down a layer in DType so that it can't be circumvented. 

FormulaType is just a lightweight wrapper

MikeStall: ditto for checks in AggregateType / TableType

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -141,6 +143,22 @@ internal static bool TryGetBestOverload(CheckTypesContext context, IErrorContain
...
+                    typeCheckSucceeded = true;
+
+                    // If one of the arg was deferred and
+                    // return type could not be calculated and was error, we assign it to deferred as safeguard.
+                    // returnType was EmptyTable, we assign it to deferred as safeguard e.g. Table(Deferred) => deferred,
+                    // this is because we don't want to embed deferred type inside of any aggregate type.
+                    if (returnType.IsError || returnType.Equals(DType.EmptyTable))
+                    {
+                        returnType = DType.Deferred;
+                    }
```
MikeStall, line 159: Can this block of code be in a static helper and shared with binder.cs?

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.3-preview.20221123-330655](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.3-preview.20221123-330655).
<details>

```
InterfacesShouldHaveSameMembers : Interface member 'Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.DeferredType)' is present in the implementation but not in the contract.
```

</details>

MikeStall:
>                 if (TryGetBestOverload(_txb.CheckTypesContext, _txb.ErrorContainer, node, argTypes, overloads, out var function, out var nodeToCoercedTypeMap, out var returnType))

@bryancroteau-MSFT  - 

This swallows type check errors when an argument is deferred, effectively shortcircuiting overload resolution and causing it to always succeed here, but propagating returnType as deferred. 

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs:4925 in f9bb65d. [](commit_id = f9bb65d7dd93633810111f29b430c284d8be845b, deletion_comment = False)

src/tests/Microsoft.PowerFx.Interpreter.Tests/DeferredTypeTests.cs
```diff
@@ -0,0 +1,145 @@
...
+using Microsoft.PowerFx.Core.Tests.Helpers;
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Types;
+using Xunit;
+
+namespace Microsoft.PowerFx.Interpreter
+{
+    public class DeferredTypeTests
+    {
+        [Theory]
```
MikeStall, line 18: >Theory [](http://example.com/codeflow?start=9&length=6)

More tests:
 
Sum(X,X) // both args are deferred 
X.Field1.Field2    // cascading field deref

ForAll([1,2,3], X)
ForAll(ParseJson("[1]"), X) 


If(true, X) // with IF
If(true, X, 1)
If(true, X, X)

jas-valgotar: Added tests except for ForAll(). Since in current iteration, we are not supporting the functions which can take expression as an input.

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.3-preview.20221123-330655](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.3-preview.20221123-330655).
<details>

```
InterfacesShouldHaveSameMembers : Interface member 'Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.DeferredType)' is present in the implementation but not in the contract.
```

</details>

src/tests/Microsoft.PowerFx.Interpreter.Tests/DeferredTypeTests.cs
jas-valgotar: Made changes to allow functions with lambda arg, Added tests for ForAll() as well now.

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.3-preview.20221123-330655](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.3-preview.20221123-330655).
<details>

```
InterfacesShouldHaveSameMembers : Interface member 'Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.DeferredType)' is present in the implementation but not in the contract.
```

</details>

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.3-preview.20221123-330655](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.3-preview.20221123-330655).
<details>

```
InterfacesShouldHaveSameMembers : Interface member 'Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.DeferredType)' is present in the implementation but not in the contract.
```

</details>

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1339,6 +1342,12 @@ public DType Add(TypedName typedName)
             Contracts.Assert(IsAggregate);
             Contracts.Assert(typedName.IsValid);
 
+            // We don't want to allow building aggregate types around deferred type.
```
MikeStall, line 1345: > // We don't want to allow building aggregate types around deferred type. [](http://example.com/codeflow?start=11&length=73)

This is a great simplifying assumption. 
We can expand later if needed.

src/tests/Microsoft.PowerFx.Interpreter.Tests/DeferredTypeTests.cs
```diff
@@ -0,0 +1,155 @@
...
+        [InlineData("Sum(X, T) < 5", "b")] // Since we discard all errors for function calls, Function calls are biased to non tabular overload.
+
+        [InlineData("Sum(X, X)", "n")]
+        [InlineData("X.Field1.Field1", "X")]
+        [InlineData("If(true, X)", "X")]
+        [InlineData("If(true, X, 1)", "n")]
+        [InlineData("If(true, X, X)", "X")]
+        [InlineData("ForAll([1,2,3], X)", "X")]
+        [InlineData("ForAll(ParseJSON(\"[1]\"), X)", "X")]
+        [InlineData("Abs(Table(X))", "n")]
```
MikeStall, line 64: >"n" [](http://example.com/codeflow?start=37&length=3)

Can you add a comment to explain why this is n (and not X)

MikeStall (dismissed):
:shipit:

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.3-preview.20221123-330655](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.3-preview.20221123-330655).
<details>

```
InterfacesShouldHaveSameMembers : Interface member 'Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.DeferredType)' is present in the implementation but not in the contract.
```

</details>

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4351,6 +4365,9 @@ public override bool PreVisit(CallNode node)
                 // Typecheck the invocation and infer the return type.
                 fArgsValid &= maybeFunc.HandleCheckInvocation(_txb, args, argTypes, _txb.ErrorContainer, out var returnType, out var nodeToCoercedTypeMap);
```
bryancroteau-MSFT, line 4366: What happens to the errors generated on this line?

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4673,6 +4690,10 @@ private void PreVisitMetadataArg(CallNode node, TexlFunction func)
 
                 // Typecheck the invocation and infer the return type.
                 fArgsValid = func.HandleCheckInvocation(_txb, args, argTypes, _txb.ErrorContainer, out returnType, out _);
```
bryancroteau-MSFT, line 4692: Same question for this line, what happens to the errors generated by this call?

bryancroteau-MSFT (changes requested):

src/tests/Microsoft.PowerFx.Interpreter.Tests/DeferredTypeTests.cs
jas-valgotar: Function overload resolution will always stay biased to scaler, and Abs scaler's returnType is number hence number

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1339,6 +1342,12 @@ public DType Add(TypedName typedName)
...
+            if(typedName.Type.IsDeferred)
```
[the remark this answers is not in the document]
jas-valgotar: I believe this covers all the aggregate type creation.

It is NOT possible to have deeply nested record or table type, and we don't want to allow anybody to create Aggregate types using DType API.
e.g. all of the below would produces an error.
[X,..]
{a : X }
{ a : { a : { a : X } } } 
Table( { a : X })
Table( { a : { a : { a : X } } } ).

src/libraries/Microsoft.PowerFx.Core/App/ErrorContainers/ErrorContainer.cs
```diff
@@ -16,6 +17,18 @@ internal class ErrorContainer : IErrorContainer
 
         public DocumentErrorSeverity DefaultSeverity => DocumentErrorSeverity.Critical;
 
+        public void MergeErrors(IEnumerable<TexlError> errors)
```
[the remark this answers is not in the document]
jas-valgotar: Correct, I did not see any other container needing this method so did not put it in IErrorContainer

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
jas-valgotar: You are correct, there should have been temp error container here.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1889,18 +1899,20 @@ public virtual bool Accepts(DType type, out KeyValuePair<string, DType> schemaDi
...
+                    accepts = (Kind != type.Kind && (type.Kind == DKind.Unknown || type.Kind == DKind.Deferred)) ||
                               (EnumSuperkind == type.EnumSuperkind && EnumTreeAccepts(ValueTree, type.ValueTree, exact));
                     break;
 
                 case DKind.Unknown:
                     accepts = type.Kind == DKind.Unknown;
                     break;
-
+                case DKind.Deferred:
+                    accepts = type.Kind == DKind.Deferred || type.Kind == DKind.Unknown;
```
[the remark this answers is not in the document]
jas-valgotar: Just in case of future use cases, as we want Deferred type to be sub type of all the type. So it can accept Unknown as well.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
jas-valgotar: You are correct, there should have been temp error container here.

src/tests/Microsoft.PowerFx.Interpreter.Tests/DeferredTypeTests.cs
```diff
@@ -0,0 +1,162 @@
...
```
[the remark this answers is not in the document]
jas-valgotar: Those test cases are now added to Negative cases.

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.3-preview.20221123-330655](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.3-preview.20221123-330655).
<details>

```
InterfacesShouldHaveSameMembers : Interface member 'Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.DeferredType)' is present in the implementation but not in the contract.
```

</details>

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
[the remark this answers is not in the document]
bryancroteau-MSFT: Do we have code coverage for this line? If not, can we add it? Otherwise should this be an error in the short term What code path _could_ ever hit this?

We certainly would not want to have Deferred appear in the argument type list of a TexlFunction. The branches of If/Switch must all be the same, how does Deferred work there?

src/tests/Microsoft.PowerFx.Interpreter.Tests/DeferredTypeTests.cs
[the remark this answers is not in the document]
bryancroteau-MSFT: Please add 1 test for Switch as well.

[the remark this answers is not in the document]
bryancroteau-MSFT: Also, can we add a test for Set? What should the semantics of Set be? For example, Set(Y, X), what is the type of Y? Should this be an error?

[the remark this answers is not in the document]
bryancroteau-MSFT: If you do Set(Y, X); Set(Y, "5"), does the Number type "clobber" the deferred type?

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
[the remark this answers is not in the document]
jas-valgotar: We have a test in src/tests/Microsoft.PowerFx.Core.Tests/TypeSystemTests/DTypeTests.cs/DeferredIsSubtypeOfAll().

For computing return type of If/Switch, it takes supertype of all possible branch and since deferred is subtype of all if any other supertype is present that becomes the return type.

src/tests/Microsoft.PowerFx.Interpreter.Tests/DeferredTypeTests.cs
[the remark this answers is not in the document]
jas-valgotar: Set(Y, X) would retain type of Y. I have added test for Switch and Set to demonstrate.

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.3-preview.20221123-330655](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.3-preview.20221123-330655).
<details>

```
InterfacesShouldHaveSameMembers : Interface member 'Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.DeferredType)' is present in the implementation but not in the contract.
```

</details>

bryancroteau-MSFT (dismissed):

MikeStall (approved):

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -617,6 +656,13 @@ BinderCheckTypeResult ReportInvalidOperation()
                             // Regular Addition
                             var leftResAdd = CheckTypeCore(errorContainer, node.Left, leftType, DType.Number, /* coerced: */ DType.String, DType.Boolean);
                             var rightResAdd = CheckTypeCore(errorContainer, node.Right, rightType, DType.Number, /* coerced: */ DType.String, DType.Boolean);
+                            
+                            // Deferred + number or number + Deferred
+                            if (leftKind == DKind.Deferred || rightKind == DKind.Deferred)
+                            {
```
lesaltzm, line 662: Is this missing in the other cases for leftKind? 
What about `Time(,...) + deferred)`?

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
lesaltzm: That's not entirely true. `ReflectionFunction` has a constructor that can specify return types.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<?xml version="1.0" encoding="utf-8"?>
 <root>
```
lesaltzm, line 2: why the header change?

src/tests/Microsoft.PowerFx.Interpreter.Tests/DeferredTypeTests.cs
```diff
@@ -0,0 +1,176 @@
...
+        [InlineData("$\"Test {X} {R}\"", "Invalid argument type (Record). Expecting a Text value instead.")]
+        [InlineData("X + R", "Invalid argument type. Expecting one of the following: Number, Text, Boolean.")]
+        [InlineData("X.field + N.field", "Invalid use of '.'")]
+        [InlineData("Index([1,2,3], X).missing", "Name isn't valid. 'missing' isn't recognized")]
+        [InlineData("X < \"2021-12-09T20:28:52Z\"", "Invalid argument type. Expecting one of the following: Number, Date, Time, DateTime.")]
+        [InlineData("First(Sum(X, 1))", "Invalid argument type (Number). Expecting a Table value instead.")]
+
+        // Can't create aggregates around Deferred type.
+        [InlineData("[X]", "Incompatible type. The item you are trying to put into a table has a type that is not compatible with the table.")]
+        [InlineData("{test: X}", "Incompatible type. The item you are trying to put into a record has a type that is not compatible with the record.")]
```
lesaltzm, line 105: These are en-US error messages you're comparing against, it'll fail when @LucGenetier tries to run them. 

Generally, best practice for error tests is to compare the Message Key

lesaltzm:
>             TestUnion("n", "n", "n");

tests for union (see your earlier comment about If())

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/TypeSystemTests/DTypeTests.cs:2047 in 9e7ecac. [](commit_id = 9e7ecac73ab5aea12f79b348709bf230afd5ce63, deletion_comment = False)

lesaltzm:
>             var type1 = DType.CreateTable(

tests for supertype

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/TypeSystemTests/DTypeTests.cs:926 in 9e7ecac. [](commit_id = 9e7ecac73ab5aea12f79b348709bf230afd5ce63, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
jas-valgotar: Thanks for pointing that out, I have updated it accordingly.

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
jas-valgotar: No, Time(,...) + deferred is not missed, we have test for that also.

jas-valgotar:
>             Assert.True(DType.Boolean.CoercesTo(DType.String));

Added the test, Coercion should be true as Deferred is subtype of all.

---
In reply to: [1331506640](https://github.com/microsoft/Power-Fx/pull/823#issuecomment-1331506640) [](http://example.com/codeflow?ancestors=1331506640,1331506640)

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/TypeSystemTests/DTypeTests.cs:1585 in 9e7ecac. [](commit_id = 9e7ecac73ab5aea12f79b348709bf230afd5ce63, deletion_comment = False)

jas-valgotar:
>             TestUnion("n", "n", "n");

Added.

---
In reply to: [1331506904](https://github.com/microsoft/Power-Fx/pull/823#issuecomment-1331506904) [](http://example.com/codeflow?ancestors=1331506904,1331506904)

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/TypeSystemTests/DTypeTests.cs:2047 in 9e7ecac. [](commit_id = 9e7ecac73ab5aea12f79b348709bf230afd5ce63, deletion_comment = False)

jas-valgotar:
>             var type1 = DType.CreateTable(

added

---
In reply to: [1331507193](https://github.com/microsoft/Power-Fx/pull/823#issuecomment-1331507193) [](http://example.com/codeflow?ancestors=1331507193,1331507193)

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/TypeSystemTests/DTypeTests.cs:926 in 9e7ecac. [](commit_id = 9e7ecac73ab5aea12f79b348709bf230afd5ce63, deletion_comment = False)

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.3-preview.20221123-330655](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.3-preview.20221123-330655).
<details>

```
InterfacesShouldHaveSameMembers : Interface member 'Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.DeferredType)' is present in the implementation but not in the contract.
```

</details>

src/tests/Microsoft.PowerFx.Interpreter.Tests/DeferredTypeTests.cs
jas-valgotar: Great point! updated the tests.

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.3-preview.20221123-330655](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.3-preview.20221123-330655).
<details>

```
InterfacesShouldHaveSameMembers : Interface member 'Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.DeferredType)' is present in the implementation but not in the contract.
```

</details>

MikeStall:
Updating the visitor is a breaking change. Flagging - we should do a buddy build with PAclient.

MikeStall (approved):
:shipit:

src/tests/Microsoft.PowerFx.Interpreter.Tests/DeferredTypeTests.cs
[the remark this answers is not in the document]
MikeStall: Note that at the SDK level, Set(Y,X)  requires that Y's type was already established. Set() is not allocated variables - just assigning.  "Something" (like Analysis) had to take a pre-pass to allocate Y.

jas-valgotar:
I did a buddy build and it was successful. 

---
In reply to: [1334631197](https://github.com/microsoft/Power-Fx/pull/823#issuecomment-1334631197) [](http://example.com/codeflow?ancestors=1334631197,1334631197)

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.3-preview.20221130-331506](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.3-preview.20221130-331506).
<details>

```
InterfacesShouldHaveSameMembers : Interface member 'Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.DeferredType)' is present in the implementation but not in the contract.
```

</details>

src/strings/PowerFxResources.en-US.resx
```diff
@@ -6615,4 +6619,8 @@
     <value>This feature is deprecated and is no longer supported.</value>
     <comment>An error message for deprecated features.</comment>
   </data>
+  <data name="WarnDeferredType" xml:space="preserve">
+    <value>Warning: Deferred type provided.</value>
+    <comment>Warning given when the input has deffered Type.</comment>
```
lesaltzm, line 6624: nit, spelling deffered  -> deferred  . Also, add a bit of context on what "Deferred" means, so the translators can have an easier time.

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -321,6 +321,13 @@ public override IntermediateNode Visit(FirstNameNode node, IRTranslatorContext c
...
+
+                if (nodeType.IsDeferred)
+                {
+                    throw new NotSupportedException("Deferred(Unknown) is not supported");
```
lesaltzm, line 328: nit: this string is repeated 2 places, define a constant?
Maybe add a Host Exceptions file if this is a pattern we'll follow more?
Also, consider adding a bit more context. We want this exception message to deflect incidents from reaching our team if a host is using this wrong. 

something like:
"Deferred(Unknown) is not supported in expressions to be evaluated. This is always an error, deferred is only valid when calling Check"

lesaltzm (approved):
looks good, 2 minor comments!

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
jas-valgotar: Added a constant.

bryancroteau-MSFT (approved):

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.3-preview.20221202-331786](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.3-preview.20221202-331786).
<details>

```
InterfacesShouldHaveSameMembers : Interface member 'Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.DeferredType)' is present in the implementation but not in the contract.
```

</details>

MikeStall (approved):

BruceHaley:
:x: 1 Binary Compatibility issues for **Microsoft.PowerFx.Core** compared against [version 0.2.3-preview.20221202-331815](https://www.nuget.org/packages/Microsoft.PowerFx.Core/0.2.3-preview.20221202-331815).
<details>

```
InterfacesShouldHaveSameMembers : Interface member 'Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.DeferredType)' is present in the implementation but not in the contract.
```

</details>

PR #858, opened 2022-11-22
Fix null reference in OptionSetValueType

ivanradicek:


src/libraries/Microsoft.PowerFx.Core/Public/Types/OptionSetValueType.cs
```diff
@@ -40,7 +40,7 @@ public override void Visit(ITypeVisitor vistor)
         /// <summary>
         /// List of logical names within this option set. 
         /// </summary>
-        public IEnumerable<DName> LogicalNames => _type.OptionSetInfo.OptionNames; 
+        public IEnumerable<DName> LogicalNames => _type.OptionSetInfo?.OptionNames;
```
MikeStall, line 43: this is consistent with the check above on line 19.

PR #863, opened 2022-11-24
Require column names as identifiers (new)

LucGenetier:
Remove use of string literals for column names arguments in functions: AddColumns, DropColumns

From `AddColumns(Table({a: 1}, {a: 2}), "b", a * a * a)`  to `AddColumns(Table({a: 1}, {a: 2}), b, a * a * a)` (notice the change around 'b')
This functionality is enabled via `SupportIdentifiers` feature.
TexlFunction has a new HasIdentifiers bool and IsIdentifierParam(index) function.


BruceHaley:

src/tests/Microsoft.PowerFx.Interpreter.Tests/DisplayNameTests.cs
```diff
@@ -0,0 +1,53 @@
...
+            var tv = TableValue.NewTable(recordType, rv1, rv2, rv3);
+
+            var parameters = RecordValue.NewRecordFromFields(
+                new NamedValue("myTable", tv));
+
+            var result = engine.Eval("DropColumns(myTable, displayName)", parameters);
+           
+            var sb = new StringBuilder();
+            result.ToExpression(sb, new FormulaValueSerializerSettings() { UseCompactRepresentation = true });
+            var output = sb.ToString();
```
MikeStall, line 42: >sb.ToString(); [](http://example.com/codeflow?start=25&length=14)

nit: there's an overload to simplify:
output = result.ToExpression();

src/tests/Microsoft.PowerFx.Interpreter.Tests/DisplayNameTests.cs
```diff
@@ -0,0 +1,53 @@
...
+using Microsoft.PowerFx.Core.Tests;
+using Microsoft.PowerFx.Types;
+using Xunit;
+
+namespace Microsoft.PowerFx.Interpreter.Tests
+{
+    public class DisplayNameTests
+    {
+        [Fact]
+        public void DisplayNameTest1()
```
MikeStall, line 14: This is a great test. 
Can you do a similar one for AddColumns? That's interesting since the argument can't have a display name (since it's new and so doesn't map to the existing type).

src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterSuggestTests.cs
```diff
@@ -170,5 +170,41 @@ public void TestSuggestVariableName(string suggestion)
...
+        [InlineData("DIs", "displayName")]
+        [InlineData("display", "displayName")]
+        [InlineData("displayname", "displayName")]
+        [InlineData("l", "logicalB")]
+        [InlineData("L", "logicalB")]
+        [InlineData("lo", "logicalB")]
+        [InlineData("LO", "logicalB")]
+        [InlineData("logical", "logicalB")]
+        [InlineData("logicalB", "logicalB")]
+        public void TestSuggestIdentifier(string txt, string expected)
```
MikeStall, line 189: >TestSuggestIdentifier [](http://example.com/codeflow?start=20&length=21)

Can this be in Core  (not interpreter) 

See existing tests at: 
[https://github.com/microsoft/Power-Fx/blob/main/src/tests/Microsoft.PowerFx.Core.Tests/IntellisenseTests/SuggestTest.cs](https://github.com/microsoft/Power-Fx/blob/main/src/tests/Microsoft.PowerFx.Core.Tests/IntellisenseTests/SuggestTest.cs)

That also has a pattern that can simplify the test authoring (and avoid things like hardcoding the constant 21),

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4317,13 +4317,26 @@ public override bool PreVisit(CallNode node)
...
+                    if (!isIdentifier)
+                    {
+                        args[i].Accept(this);
+                        _txb.AddVolatileVariables(node, _txb.GetVolatileVariables(args[i]));
+                        argTypes[i] = _txb.GetType(args[i]);
+                    }
+                    else
+                    {
+                        // This is an identifier, no associated type 
+                        argTypes[i] = DType.Unknown;
```
MikeStall, line 4337: >DType.Unknown [](http://example.com/codeflow?start=38&length=13)

I think this is fine - since we should never actually use the type.  (ie, curious - could we set it to null and not hit anything?) 

We could also "poison pill" it by making it error.

src/libraries/Microsoft.PowerFx.Core/Public/Config/Features.cs
```diff
@@ -28,5 +29,11 @@ public enum Features : int
         /// Filter(A, A[@Value] = 2)
         /// </summary>
         DisableRowScopeDisambiguationSyntax = 0x4,
+
+        /// <summary>
+        /// Enable Identifier support for describing column names
+        /// Added on 20ieth July 2022.
+        /// </summary>
+        SupportIdentifiers = 0x8,
```
MikeStall, line 37: >SupportIdentifiers [](http://example.com/codeflow?start=8&length=18)

nit: "identifier support" is too broad... 
Although "Identifier Args instead of Strings for field names" is too verbose.  Maybe something in the middle?

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -57,6 +57,19 @@ public PowerFxConfig(CultureInfo cultureInfo = null)
         {
         }
 
+        internal PowerFxConfig(CultureInfo culture, SymbolTable symbols, Features features)
```
MikeStall, line 60: >PowerFxConfig [](http://example.com/codeflow?start=17&length=13)

why a new ctor?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DropColumns.cs
```diff
@@ -65,33 +66,63 @@ public override bool CheckTypes(TexlNode[] args, DType[] argTypes, IErrorContain
...
+
+                    // Argument '{0}' is not a valid identifier.
+                    errors.EnsureError(DocumentErrorSeverity.Severe, nameArg, TexlStrings.ErrArgNotAValidIdentifier_Name, expectedColumnName);
                     continue;
                 }
 
-                var columnName = new DName(nameNode.Value);
+                var columnName = nameArg is FirstNameNode && 
+                                 argTypes[0].DisplayNameProvider != null && 
+                                 argTypes[0].DisplayNameProvider.TryGetLogicalName(new DName(expectedColumnName), out var logicalDName)
```
MikeStall, line 123: There's some broader helper on DType we should use here. Ie, see what RecordType is using.
DNP is only one of the 3 ways of handling Display Names.

src/tests/Microsoft.PowerFx.Interpreter.Tests/DisplayNameTests.cs
LucGenetier: It's not fully equivalent as is usesUseCompactRepresentation=false (and the API is missing a bool param)
That said, it doesn't change the test and i'll fix it.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
LucGenetier: I think Unknown is the most appropriate here
- null will immediately generate a NRE at Contracts.Assert(argTypes[i].IsValid)
- DType.Error will make it fail on "check.ThrowOnErrors();" a bit later in the execution

Notice that the real type of the new column will be determined later when calling maybeFunc.HandleCheckInvocation(_.., out var returnType, ...);

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
LucGenetier: Need it to copy an existing config & change its locale
Used in WithCulture below, itself used in RenameDriver.cs.
Notice that we haven't any constructor taking a SymbolTable

src/tests/Microsoft.PowerFx.Core.Tests/LocalizationTests.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.PowerFx.Core.Localization;
+using Xunit;
+using System.Reflection;
+using System.Linq;
+
+namespace Microsoft.PowerFx.Core.Tests
+{
+    public class LocalizationTests
```
MikeStall, line 14: >LocalizationTests [](http://example.com/codeflow?start=17&length=17)

? git merge issue?

MikeStall: This and the corresponding .resx* changes should definitely be in a separate PR.

bryancroteau-MSFT:
Can you update the description to include your strategy/approach to solving this?

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
MikeStall: SymbolTable has a mutable setter.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
MikeStall: "I think Unknown is the most appropriate here" 
We need a stronger argument, especially because, despite the name, Unknown's semantics are somewhat ambiguous (we hit this exact issue on the deferred work for #720 ). 

If other code is actually reading this value - then it's possible that assigning it Unknown could create bugs. 

Since nobody should actually be reading this value, we should look at the exact locations that are reading this -  and understand why. Should they not be?

src/tests/Microsoft.PowerFx.Core.Tests/LocalizationTests.cs
LucGenetier: We are adding ErrExpectedIdentifierArg_Name message

LucGenetier: This test verifies we are not forgetting any message (which was the case), hence a couple of renames

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
LucGenetier: You're right.
I've set it to null and verified where it's used.
There are only 2 places:
- right after this code in Contracts.Assert
- in TexfFuntion.CheckTypesCore where we can easily avoid the problem

BruceHaley:

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4317,14 +4317,30 @@ public override bool PreVisit(CallNode node)
...
+                    {
+                        // This is an identifier, no associated type
+                        argTypes[i] = null;
+                    }
 
-                    argTypes[i] = _txb.GetType(args[i]);
-                    Contracts.Assert(argTypes[i].IsValid);
+                    if (!isIdentifier)
+                    {
+                        Contracts.Assert(argTypes[i].IsValid);
```
MikeStall, line 4342: > Contracts.Assert(argTypes[i].IsValid); [](http://example.com/codeflow?start=23&length=39)

nit - just move to line 4332 inside the existing isIdentifier check.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -78,6 +78,13 @@ internal abstract class TexlFunction : IFunction
         // Return true if the function expects lambda arguments, false otherwise.
         public virtual bool HasLambdas => !_maskLambdas.IsZero;
 
+        /// <summary>
+        /// Returns true if the function expect identifiers, false otherwise.
+        /// Needs to be overloaded for functions having identifier parameters.
+        /// Also overload IsIdentifierParam method. 
+        /// </summary>
+        public virtual bool HasIdentifiers => false;
```
MikeStall, line 86: >HasIdentifiers [](http://example.com/codeflow?start=28&length=14)

nit - update name to be consistent with flags. 
Like "HasColumnIdentifiers"

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -297,7 +304,18 @@ public override IntermediateNode Visit(TexlCallNode node, IRTranslatorContext co
...
-                    if (func.IsLazyEvalParam(i))
+
+                    var supportColumnNamesAsIdentifiers = _features.HasFlag(Features.SupportColumnNamesAsIdentifiers);
+                    if (supportColumnNamesAsIdentifiers && func.IsIdentifierParam(i))
+                    {
+                        var identifierNode = arg.AsFirstName();
+                        Contracts.Assert(identifierNode != null);
+
+                        // Transform the identifier node as a string literal
+                        var nodeName = context.Binding.TryGetReplacedIdentName(identifierNode.Ident, out var newIdent) ? new DName(newIdent) : identifierNode.Ident.Name;
```
MikeStall, line 315: >TryGetReplacedIdentName [](http://example.com/codeflow?start=55&length=23)

Why this API TryGetReplacedIdentName instead of the binding info (FirstNameInfo)?

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -445,12 +452,19 @@ private bool CheckTypesCore(TexlNode[] args, DType[] argTypes, IErrorContainer e
...
+            Contracts.AssertValue(argTypes);            
             Contracts.Assert(args.Length == argTypes.Length);
             Contracts.AssertValue(errors);
             Contracts.Assert(MinArity <= args.Length && args.Length <= MaxArity);
 
+            for (int i = 0; i < argTypes.Length; i++)
+            {
+                if (!IsIdentifierParam(i))
+                {
+                    Contracts.AssertValid(argTypes[i]);
```
bryancroteau-MSFT, line 464: Did we not determine what the type of these args would be?

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4066,10 +4066,10 @@ public override bool PreVisit(CallNode node)
                     return false;
                 }
 
-                // If there are no overloads with lambdas, we can continue the visitation and
+                // If there are no overloads with lambdas or identifiers, we can continue the visitation and
                 // yield to the normal overload resolution.
-                var overloadsWithLambdas = overloads.Where(func => func.HasLambdas);
-                if (!overloadsWithLambdas.Any())
+                var overloadsWithLambdasOrIdentifiers = overloads.Where(func => func.HasLambdas || func.HasIdentifiers);
```
bryancroteau-MSFT, line 4071: As mentioned in the teams thread, I am concerned about functions which take lambdas also taking identifiers in the exact same argument slot. Right now, this is not the case. However, we need to add asserts to make absolutely sure.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4317,14 +4317,30 @@ public override bool PreVisit(CallNode node)
...
+                    if (!isIdentifier)
+                    {
+                        args[i].Accept(this);
+                        _txb.AddVolatileVariables(node, _txb.GetVolatileVariables(args[i]));
+                        argTypes[i] = _txb.GetType(args[i]);
+                    }
+                    else
+                    {
+                        // This is an identifier, no associated type
+                        argTypes[i] = null;
```
bryancroteau-MSFT, line 4337: I do not think we should be doing this.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -6615,4 +6615,8 @@
     <value>This feature is deprecated and is no longer supported.</value>
     <comment>An error message for deprecated features.</comment>
   </data>
+  <data name="ErrExpectedIdentifierArg_Name" xml:space="preserve">
```
lesaltzm, line 6618: We have a loc team. Only change en-US, the others will get overwritten

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AddColumns.cs
```diff
@@ -78,41 +77,68 @@ public override bool CheckTypes(TexlNode[] args, DType[] argTypes, IErrorContain
...
+                        continue;
+                    }
+
+                    expectedColumnName = identifierNode.Ident.Name;
+                }
+                else
+                {
+                    StrLitNode strLitNode = nameArg.AsStrLit();
+
+                    if (nameArgType.Kind != DKind.String && strLitNode == null)
```
jas-valgotar, line 108: >&& [](http://example.com/codeflow?start=57&length=2)

This should be an Or `||` condition.

jas-valgotar: Otherwise we would miss error and throw null pointer on arg which is FirstName node with string type.
since in that case, `nameArgtype.Kind != DKind.String` would be `false`

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DropColumns.cs
```diff
@@ -65,33 +66,61 @@ public override bool CheckTypes(TexlNode[] args, DType[] argTypes, IErrorContain
...
+                        continue;
+                    }
+
+                    expectedColumnName = identifierNode.Ident.Name;
+                }
+                else
+                {
+                    StrLitNode strLitNode = nameArg.AsStrLit();
+
+                    if (nameArgType.Kind != DKind.String && strLitNode == null)
```
jas-valgotar, line 99: >&& [](http://example.com/codeflow?start=57&length=2)

should be `||`, same as I mentioned for AddColumns()

src/libraries/Microsoft.PowerFx.Core/Binding/BindInfo/FirstNameInfo.cs
```diff
@@ -18,6 +18,10 @@ internal sealed class FirstNameInfo : NameInfo
     {
         public override DName Name => Node.AsFirstName().Ident.Name;
 
+        // DisplayName is set in Create methods
+        // Mostly used with identifiers in PreVisit(CallNode node)
+        public DName DisplayName { get; init; }
```
MikeStall, line 23: Does anybody actually use this? 
Binder is usually translating to *logical* names  and away from DisplayNames.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
MikeStall: Per teams chat today, this should be DType.Invalid.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4346,7 +4362,26 @@ public override bool PreVisit(CallNode node)
...
                 PostVisit(node.Args);
 
-                // Typecheck the invocation.
+                if (maybeFunc.HasIdentifiers && _features.HasFlag(Features.SupportColumnNamesAsIdentifiers))
+                {
+                    var i = 0;
+
+                    foreach (var arg in args)
+                    {
+                        i++;
```
MikeStall, line 4371: >i++; [](http://example.com/codeflow?start=24&length=4)

This means arg is args[i-1] . 
so we're out of sync with IsIdentifierParam(i)?

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4118,16 +4118,16 @@ public override bool PreVisit(CallNode node)
...
 
                 // We support a single overload with lambdas. Otherwise we have a conceptual chicken-and-egg
                 // problem, whereby in order to bind the lambda args we need the precise overload (for
                 // its lambda mask), which in turn requires binding the args (for their types).
-                Contracts.Assert(overloadsWithLambdas.Count() == 1, "Incorrect multiple overloads with lambdas.");
-                var maybeFunc = overloadsWithLambdas.Single();
-                Contracts.Assert(maybeFunc.HasLambdas);
+                Contracts.Assert(overloadsWithLambdasOrIdentifiers.Count() == 1, "Incorrect multiple overloads with lambdas.");
+                var maybeFunc = overloadsWithLambdasOrIdentifiers.Single();
+                Contracts.Assert(maybeFunc.HasLambdas || maybeFunc.HasIdentifiers);
```
MikeStall, line 4130: >|| [](http://example.com/codeflow?start=54&length=2)

xor

MikeStall:
>                     fArgsValid = scopeInfo.CheckInput(nodeInput, typeInput, out typeScope);

Under what conditions can scopeInfo now be null? 

if scope can now be null on 4247 - does that mean it could be null here?


---
Refers to: src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs:4255 in f542454. [](commit_id = f542454cc5ee23240a5cbadda6603ec4a469530a, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4346,7 +4362,26 @@ public override bool PreVisit(CallNode node)
...
+                    {
+                        i++;
+
+                        if (arg is FirstNameNode firstNameNode && maybeFunc.IsIdentifierParam(i))
+                        {
+                            var displayName = GetLogicalNodeNameAndUpdateDisplayNames(argTypes[0], firstNameNode.Ident, out _);
+
+                            if (argTypes[0].TryGetType(displayName, out var fieldType))
+                            {
+                                var fileNameInfo = FirstNameInfo.Create(firstNameNode, new NameLookupInfo(BindKind.NamedValue, fieldType, DPath.Root, 0, displayName: new DName(displayName)));
```
MikeStall, line 4379: >BindKind.NamedValue [](http://example.com/codeflow?start=106&length=19)

How did you decide to use this value? (Does it actually matter)
There are some other interesting candidates, like LambdaField

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DropColumns.cs
MikeStall: We shouldn't be referencing 
argTypes[0].DisplayNameProvider at all.  (DNP is only one of the 3 ways of handling Display Names.  )

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4346,7 +4362,26 @@ public override bool PreVisit(CallNode node)
...
+                                _txb.SetInfo(firstNameNode, fileNameInfo);
```
MikeStall, line 4380: >_txb.SetInfo(firstNameNode, fileNameInfo); [](http://example.com/codeflow?start=32&length=42)

Is there anything that actually uses this FirstNameInfo that we set here?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DropColumns.cs
LucGenetier: Using DType.TryGetLogicalNameForColumn now

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DropColumns.cs
```diff
@@ -65,33 +66,61 @@ public override bool CheckTypes(TexlNode[] args, DType[] argTypes, IErrorContain
...
                     fArgsValid = false;
-                    errors.EnsureError(DocumentErrorSeverity.Severe, nameArg, TexlStrings.ErrArgNotAValidIdentifier_Name, nameNode.Value);
+
+                    // Argument '{0}' is not a valid identifier.
+                    errors.EnsureError(DocumentErrorSeverity.Severe, nameArg, TexlStrings.ErrArgNotAValidIdentifier_Name, expectedColumnName);
                     continue;
                 }
 
-                var columnName = new DName(nameNode.Value);
+                var columnName = new DName(nameArg is FirstNameNode && argTypes[0].DisplayNameProvider != null && DType.TryGetLogicalNameForColumn(argTypes[0], expectedColumnName, out var logicalName)
```
MikeStall, line 121: >argTypes[0].DisplayNameProvider != null [](http://example.com/codeflow?start=71&length=39)

this part is the problem -  DisplayNameProvider  is only 1 of several ways DisplayNames are handled.  For example, this probably would fail on a dataverse table. 

If you're already using the DType API, you shouldn't have to check this.

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
LucGenetier: Same API as the one used at line 350 with FirstNameNode. I don't see what other API you'd like to call and why.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
LucGenetier: Am replacing AssertAllValid with this code (line 449 in previous code)

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
LucGenetier: line 4367 should start at -1 instead

LucGenetier: This i++ is wrongly placed, you're right
It should be after the 'if' condition

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
LucGenetier: After checking ... no!
I'll remove that part

src/strings/PowerFxResources.en-US.resx
LucGenetier: We need to provide strings for other languages as we'd get a "Resource not found" exception if we don't provide those, when this message will be used.
I've checked with the localization team and they will ignore the other resource strings, so this is fine to keep those.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AddColumns.cs
LucGenetier: Good eyes!

src/libraries/Microsoft.PowerFx.Core/Binding/BindInfo/FirstNameInfo.cs
LucGenetier: You're correct, this isn't necessary anymore

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
LucGenetier: Code removed, see other comments

LucGenetier:
>                     fArgsValid = scopeInfo.CheckInput(nodeInput, typeInput, out typeScope);

I don't remember when/why I had to change line 4247
I changed it back to what it was 

---
In reply to: [1330158508](https://github.com/microsoft/Power-Fx/pull/863#issuecomment-1330158508) [](http://example.com/codeflow?ancestors=1330158508,1330158508)

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs:4255 in f542454. [](commit_id = f542454cc5ee23240a5cbadda6603ec4a469530a, deletion_comment = False)

BruceHaley:

BruceHaley:

BruceHaley:

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
MikeStall: but now that argType is no longer null ... can we revert this change?

LucGenetier: This wouldn't work as DType.Invalid will make Contracts.Assert fail

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
LucGenetier: Added new code in PowerFxConfig AddFunction method to test this + extra UTs

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
MikeStall: discussed offline - Identifier Params are DKind.Invalid, which would still fail the assert.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AddColumns.cs
```diff
@@ -78,41 +77,68 @@ public override bool CheckTypes(TexlNode[] args, DType[] argTypes, IErrorContain
...
                 // Verify that the name is valid.
-                if (!DName.IsValidDName(nameNode.Value))
+                if (!DName.IsValidDName(expectedColumnName))
                 {
                     fArgsValid = false;
-                    errors.EnsureError(DocumentErrorSeverity.Severe, nameArg, TexlStrings.ErrArgNotAValidIdentifier_Name, nameNode.Value);
+
+                    // Argument '{0}' is not a valid identifier.
+                    errors.EnsureError(DocumentErrorSeverity.Severe, nameArg, TexlStrings.ErrArgNotAValidIdentifier_Name, expectedColumnName);
                     continue;
```
anderson-joyle, line 127: Test coverage is not reaching this portion.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AddColumns.cs
```diff
@@ -78,41 +77,68 @@ public override bool CheckTypes(TexlNode[] args, DType[] argTypes, IErrorContain
...
+                {
+                    StrLitNode strLitNode = nameArg.AsStrLit();
+
+                    if (nameArgType.Kind != DKind.String || strLitNode == null)
+                    {
+                        fArgsValid = false; 
+
+                        // Argument '{0}' is invalid, expected a text literal.
+                        errors.EnsureError(DocumentErrorSeverity.Severe, nameArg, TexlStrings.ErrExpectedStringLiteralArg_Name, nameArg.ToString());
+                        continue;
```
anderson-joyle, line 114: Test coverage is not reaching this portion.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DropColumns.cs
```diff
@@ -65,33 +66,61 @@ public override bool CheckTypes(TexlNode[] args, DType[] argTypes, IErrorContain
...
+                {
+                    StrLitNode strLitNode = nameArg.AsStrLit();
+
+                    if (nameArgType.Kind != DKind.String || strLitNode == null)
+                    {
+                        fArgsValid = false;
+
+                        // Argument '{0}' is invalid, expected a text literal.
+                        errors.EnsureError(DocumentErrorSeverity.Severe, nameArg, TexlStrings.ErrExpectedStringLiteralArg_Name, nameArg.ToString());
+                        continue;
```
anderson-joyle, line 105: Test coverage is not reaching this portion.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DropColumns.cs
```diff
@@ -65,33 +66,61 @@ public override bool CheckTypes(TexlNode[] args, DType[] argTypes, IErrorContain
...
                 // Verify that the name is valid.
-                if (!DName.IsValidDName(nameNode.Value))
+                if (!DName.IsValidDName(expectedColumnName))
                 {
                     fArgsValid = false;
-                    errors.EnsureError(DocumentErrorSeverity.Severe, nameArg, TexlStrings.ErrArgNotAValidIdentifier_Name, nameNode.Value);
+
+                    // Argument '{0}' is not a valid identifier.
+                    errors.EnsureError(DocumentErrorSeverity.Severe, nameArg, TexlStrings.ErrArgNotAValidIdentifier_Name, expectedColumnName);
                     continue;
```
anderson-joyle, line 118: Test coverage is not reaching this portion.

BruceHaley:

src/strings/PowerFxResources.en-US.resx
lesaltzm: If that's happening, we have a separate bug. We should be falling back to en-us if the string isn't found.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
Shpakh-MSFT, line 1: >tandalone="yes"? [](http://example.com/codeflow?start=38&length=16)

Is this intentional change?

src/libraries/Microsoft.PowerFx.Core/Public/DependencyFinder.cs
```diff
@@ -31,6 +31,11 @@ public override void Visit(FirstNameNode node)
         {
             var info = _binding.GetInfo(node);
 
+            if (info == null)
+            {
+                return;
+            }
```
Shpakh-MSFT, line 37: Can you add some comment about when this can be null?

src/libraries/Microsoft.PowerFx.Core/Public/Config/Features.cs
```diff
@@ -28,5 +29,11 @@ public enum Features : int
         /// Filter(A, A[@Value] = 2)
         /// </summary>
         DisableRowScopeDisambiguationSyntax = 0x4,
+
+        /// <summary>
+        /// Enable Identifier support for describing column names
+        /// Added on 20ieth July 2022.
```
Shpakh-MSFT, line 35: >Added on 20ieth July 2022. [](http://example.com/codeflow?start=12&length=26)

Why do we have this comment?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DropColumns_SupportColumnNamesAsIdentifiers.txt
```diff
@@ -0,0 +1,28 @@
...
+>> DropColumns(Table({a:1,b:2},{a:3,b:4}), b)
+Table({a:1},{a:3})
+
+>> DropColumns(Table({a:1,b:2},{a:3,b:4}), b, a)
+Table({},{})
+
+>> DropColumns(Table({a:1,b:2},{a:3,b:4}), a, b)
+Table({},{})
+
+>> DropColumns(DropColumns(Table({a:1,b:2},{a:3,b:4}), a), b)
```
Shpakh-MSFT, line 12: >DropColumns(DropColumns(Table({a:1,b:2},{a:3,b:4}), a), b) [](http://example.com/codeflow?start=3&length=58)

Can you add a case for DropColumns(AddColumns(...?

src/strings/PowerFxResources.en-US.resx
LucGenetier: Absolutely not :)
I don't know what "standalone" adds here but i'll have it back

src/libraries/Microsoft.PowerFx.Core/Public/DependencyFinder.cs
LucGenetier: Sure - Occurs when an identifier is used

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DropColumns_SupportColumnNamesAsIdentifiers.txt
LucGenetier: >> DropColumns(AddColumns(Table({a:1,b:2},{a:3,b:4}), c, a * b), a)
Table({b:2,c:2},{b:4,c:12})

>> DropColumns(AddColumns(Table({a:1,b:2},{a:3,b:4}), c, a * b), b)
Table({a:1,c:2},{a:3,c:12})

>> DropColumns(AddColumns(Table({a:1,b:2},{a:3,b:4}), c, a * b), c)
Table({a:1,b:2},{a:3,b:4})

>> AddColumns(DropColumns(Table({a:1,b:2},{a:3,b:4}), a), a, b * b * b)
Table({a:8,b:2},{a:64,b:4})

BruceHaley:

src/strings/PowerFxResources.en-US.resx
MikeStall: fwiw - I'm seeing this is in a bunch of PRs ... some VS thing is automatically doing this.

VamsiModem (approved):

MikeStall (dismissed):

MikeStall (changes requested):
:clock1:

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DropColumns.cs
MikeStall: please fix this.

MikeStall:
>                 returnType = returnType.Drop(ref fError, DPath.Root, columnName);

@Shpakh-MSFT  - this is the answer to your question from chat. 
Return type is computed the same way regardless of whether column name comes from string form or symbol form. 

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DropColumns.cs:135 in c340c01. [](commit_id = c340c01b3ca4bb0d8802006ec6b5871734133166, deletion_comment = False)

BruceHaley:

src/tests/Microsoft.PowerFx.Interpreter.Tests/DisplayNameTests.cs
```diff
@@ -0,0 +1,87 @@
...
+            var result = engine.Eval("DropColumns(myTable, displayName)", parameters);                      
+            var output = result.ToExpression();
+
+            Assert.Equal("Table({logicalB:4},{logicalB:5},{logicalB:6})", output);
+
+            var displayExpression = engine.GetDisplayExpression("DropColumns(myTable, displayName)", parameters.Type);
+            var invariantExpression = engine.GetInvariantExpression("DropColumns(myTable, displayName)", parameters.Type);
+
+            Assert.Equal("DropColumns(myTable, displayName)", displayExpression);
+            Assert.Equal("DropColumns(myTable, logicalA)", invariantExpression);
```
Shpakh-MSFT, line 47: >DropColumns [](http://example.com/codeflow?start=26&length=11)

Not related to your change but 
ShowColumns is very interesting function and we should fix it.

Today following expression returns a type *[createdby: s]
Showcolumns(Accounts, “createdby”) 

After your change same expression customers will write as 
ShowColumns(Accounts, ‘Created By’)
But the return type still contains logical field name.
*[createdby: s]

This would be even more confusing after this change.

But DropColumns actually returns a type with display name in it.

src/tests/Microsoft.PowerFx.Interpreter.Tests/DisplayNameTests.cs
```diff
@@ -0,0 +1,87 @@
...
+            var rv3 = RecordValue.NewRecordFromFields(
+                new NamedValue("logicalA", FormulaValue.New(3)),
+                new NamedValue("logicalB", FormulaValue.New(6)));
+
+            var tv = TableValue.NewTable(recordType, rv1, rv2, rv3);
+
+            var parameters = RecordValue.NewRecordFromFields(
+                new NamedValue("myTable", tv));
+
+            var result = engine.Eval("AddColumns(myTable, newColumn, displayName * logicalB)", parameters);
```
Shpakh-MSFT, line 75: >logicalB [](http://example.com/codeflow?start=83&length=8)

Do we have a negative test to validate that logicalname identifier is invalid? 

For example,
AddColumns(myTable, newColumn, logicalA)

MikeStall:
Although this would be a good thing to test. 
are we allowed to pass logical name as arg1?

If so, can we verify that logical and display name give same results?

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AddColumns.cs
```diff
@@ -78,41 +77,68 @@ public override bool CheckTypes(TexlNode[] args, DType[] argTypes, IErrorContain
...
+                    }
+
+                    expectedColumnName = strLitNode.Value;
                 }
 
                 // Verify that the name is valid.
-                if (!DName.IsValidDName(nameNode.Value))
+                if (!DName.IsValidDName(expectedColumnName))
```
Shpakh-MSFT, line 121: >Name.IsValidDName( [](http://example.com/codeflow?start=22&length=18)

This check is probably not needed when supportColumnNamesAsIdentifiers = true

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DropColumns.cs
```diff
@@ -5,26 +5,27 @@
...
 using Microsoft.PowerFx.Core.Utils;
 using Microsoft.PowerFx.Syntax;
 
-#pragma warning disable SA1402 // File may only contain a single type
-#pragma warning disable SA1649 // File name should match first type name
-
 namespace Microsoft.PowerFx.Core.Texl.Builtins
 {
     // DropColumns(source:*[...], name:s, name:s, ...)
     internal sealed class DropColumnsFunction : FunctionWithTableInput
```
Shpakh-MSFT, line 17: >DropColumnsFunction [](http://example.com/codeflow?start=26&length=19)

Why not update ShowColumns as well?

src/tests/Microsoft.PowerFx.Interpreter.Tests/DisplayNameTests.cs
LucGenetier: I'll test the return types with display names

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AddColumns.cs
LucGenetier: I prefer keeping that test as there is no risk.
If later we have functions that return identifiers and if they could return invalid identifiers, this check will be a valid safety net.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DropColumns.cs
LucGenetier: I considered it out of scope for this PR 
We can address it in a future PR for sure.

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DropColumns.cs
MikeStall: Ditto - future PR. This is laying a foundation, and it would even enabling doing the other functions in parallel - similar to what Bryan did for migrating to CheckType

MikeStall (approved):
:shipit:

BruceHaley:

Shpakh-MSFT (approved):
:shipit:

PR #866, opened 2022-11-29
Update static format check for Text function

CarlosFigueiraMSFT:
The Text function has a static check that the format argument should not have both numeric and date/time specifiers. This logic will change (once we support `0` as the specifier for fractional seconds, as Excel does), and it isn't implemented correctly. This change updates it to cover more constant values. Notice that this should not be a replacement for a runtime check, since the format can come from expressions, variables, which are not captured by a static check.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
@@ -2,7 +2,9 @@
 // Licensed under the MIT license.
 
 using System.Collections.Generic;
+using System.Text.RegularExpressions;
 using Microsoft.PowerFx.Core.App.ErrorContainers;
+using Microsoft.PowerFx.Core.Binding;
 using Microsoft.PowerFx.Core.Errors;
 using Microsoft.PowerFx.Core.Functions;
 using Microsoft.PowerFx.Core.Localization;
@@ -34,7 +36,7 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
             yield return new[] { TexlStrings.TextArg1, TexlStrings.TextArg2, TexlStrings.TextArg3 };
         }
 
-        public override bool CheckTypes(TexlNode[] args, DType[] argTypes, IErrorContainer errors, out DType returnType, out Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
+        public override bool CheckTypes(CheckTypesContext checkTypesContext, TexlNode[] args, DType[] argTypes, IErrorContainer errors, out DType returnType, out Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
         {
             Contracts.AssertValue(args);
             Contracts.AssertAllValues(args);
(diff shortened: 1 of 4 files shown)


MikeStall:
"this should be a runtime check, not a static one, since the format can come from expressions,"
True, although these are frequently literal constants - and it is helpful to give warnings if they're wrong in the case when they are literals.

CarlosFigueiraMSFT:
Fair point. Doing the proper validation is a little expensive (and it will be done again at runtime), but I added the validation back - and this time made it work properly, not only for literals but also for other "constant" expressions.

---
In reply to: [1331096059](https://github.com/microsoft/Power-Fx/pull/866#issuecomment-1331096059) [](http://example.com/codeflow?ancestors=1331096059,1331096059)

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
```diff
@@ -99,33 +101,40 @@ public override bool CheckTypes(TexlNode[] args, DType[] argTypes, IErrorContain
...
-                var hasDateTimeFmt = fmt.IndexOfAny(new char[] { 'm', 'd', 'y', 'h', 'H', 's', 'a', 'A', 'p', 'P' }) >= 0;
-                var hasNumericFmt = fmt.IndexOfAny(new char[] { '0', '#' }) >= 0;
+                var hasDateTimeFmt = formatArg.IndexOfAny(new char[] { 'm', 'd', 'y', 'h', 'H', 's', 'a', 'A', 'p', 'P' }) >= 0;
+                var hasNumericFmt = formatArg.IndexOfAny(new char[] { '0', '#' }) >= 0;
                 if (hasDateTimeFmt && hasNumericFmt)
                 {
-                    errors.EnsureError(DocumentErrorSeverity.Moderate, formatNode, TexlStrings.ErrIncorrectFormat_Func, Name);
+                    // Check if the date time format contains '0's after the seconds specifier, which
+                    // is used for fractional seconds - in which case it is valid
+                    var formatWithoutZeroSubseconds = Regex.Replace(formatArg, @"[sS]\.?(0+)",
```
MikeStall, line 130: >Regex.Replace [](http://example.com/codeflow?start=54&length=13)

we always have to be wary of RegEx on the server. 
But I think this is ok since the regex is constant and avoids backtracing, etc (and not coming from user input)

MikeStall (approved):
:shipit:

MikeStall (dismissed):

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
```diff
@@ -99,33 +101,40 @@ public override bool CheckTypes(TexlNode[] args, DType[] argTypes, IErrorContain
...
-                if (fmt.StartsWith("[$-"))
+                if (formatArg.StartsWith("[$-"))
                 {
-                    var end = fmt.IndexOf(']', 3);
+                    var end = formatArg.IndexOf(']', 3);
                     if (end > 0)
                     {
-                        fmt = fmt.Substring(end + 1);
+                        formatArg = formatArg.Substring(end + 1);
                     }
```
anderson-joyle, line 121: No test case is reaching this portion.

CarlosFigueiraMSFT: This hasn't changed, but since I'm touching this function I added a couple of cases that exercise this code as well.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
```diff
@@ -34,7 +36,7 @@ public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
             yield return new[] { TexlStrings.TextArg1, TexlStrings.TextArg2, TexlStrings.TextArg3 };
         }
 
-        public override bool CheckTypes(TexlNode[] args, DType[] argTypes, IErrorContainer errors, out DType returnType, out Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
+        public override bool CheckTypes(CheckTypesContext checkTypesContext, TexlNode[] args, DType[] argTypes, IErrorContainer errors, out DType returnType, out Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
```
anderson-joyle, line 39: >args [](http://example.com/codeflow?start=88&length=4)

One possible argument is ResultLanguageTag (CultureInfo name). Should ResultLanguageTag argument get validated to check if is a valid one?

CarlosFigueiraMSFT: The check for the third argument should be done at runtime

anderson-joyle (approved):
:shipit:

PR #868, opened 2022-11-30
Move the DateTime function to the core library

CarlosFigueiraMSFT:
The function is already implemented in Power Apps, so it can be moved to the core library.


MikeStall (approved):
:shipit:

PR #870, opened 2022-12-01
Allow coercions in the Concat function

CarlosFigueiraMSFT:
Currently the Concat function disallows coercions to its parameters, but there is nothing that should prevent that from happening. This change removes that restriction, and adds tests to validate that coercions work as intended.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Lookup.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Lookup.cs
@@ -45,6 +45,11 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
 
             var fValid = base.CheckTypes(args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
 
+            if (!fValid)
+            {
+                return false;
+            }
+
             // The return type is dictated by the last argument (projection) if one exists. Otherwise it's based on first argument (source).
             returnType = args.Length == 2 ? argTypes[0].ToRecord() : argTypes[2];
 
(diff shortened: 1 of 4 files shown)


src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Lookup.cs
```diff
@@ -45,6 +45,11 @@ public override bool CheckTypes(TexlNode[] args, DType[] argTypes, IErrorContain
 
             var fValid = base.CheckTypes(args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
 
+            if (!fValid)
+            {
+                return false;
+            }
```
MikeStall, line 51: Why this?

This could be significant because it skips potential logic for determining a returnType. 
We still want returnType even if CheckTypes returns false (@bryancroteau-MSFT )

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Concat.txt
```diff
@@ -111,3 +111,24 @@ Errors: Error 43-48: Name isn't valid. 'Value' isn't recognized.
 // Errors in the lambda
 >> Concat([2, 1, 0, -1, -2], Text(1 / Value), ",")
 Error({Kind:ErrorKind.Div0})
+
+// Coercions on lambda
```
MikeStall, line 115: These tests are cases that failed before and now they pass?

It seems reasonably, but it's technically a change in semantics, we should get a confirmation from Greg.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Lookup.cs
CarlosFigueiraMSFT: This change wasn't supposed to be in this PR, thanks for catching it. Reverted

MikeStall (approved):
:shipit:

PR #871, opened 2022-12-01
Allow coercions in the CountIf function

CarlosFigueiraMSFT:
Currently the CountIf function disallows coercions to its parameters, but there is nothing that should prevent that from happening. This change removes that restriction and adds tests to validate that coercions work as intended.

This also fixes the CountIf function to work with multiple predicates.


src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/CountIf.cs
```diff
@@ -76,8 +76,15 @@ public override bool CheckTypes(TexlNode[] args, DType[] argTypes, IErrorContain
                 }
```
Shpakh-MSFT, line 76: Do we need this separate check for OptionSetValue in that case?

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
```diff
@@ -280,21 +280,30 @@ public void TexlFunctionTypeSemanticsCountA()
                 symbol);
         }
 
-        [Fact]
-        public void TexlFunctionTypeSemanticsCountIf()
+        [Theory]
+        [InlineData("CountIf(Table, A < 10)")]
+        [InlineData("CountIf(Table, A < 10, A > 0, A <> 2)")]
+        [InlineData("CountIf([1,2,3], Value) // Coercion from number to boolean")]
+        [InlineData("CountIf([\"false\",\"true\",\"false\"], Value) // Coercion from text to boolean")]
```
Shpakh-MSFT, line 287: >[\"false\",\"true\",\"false\ [](http://example.com/codeflow?start=29&length=28)

Any case for error scenario?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/CountIf.cs
CarlosFigueiraMSFT: We already have a check for OptionSetValue. This extends it to other types.

Shpakh-MSFT:
This change looks to be bigger than just CountIf function as description says.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/CountIf.cs
CarlosFigueiraMSFT: Ah, good question. Let me try it out

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/CountIf.cs
```diff
@@ -70,14 +70,17 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
             // Ensure that all the args starting at index 1 are booleans or coerece to boolean if they are OptionSetValues.
```
Shpakh-MSFT, line 70: >coerece to boolean if they are OptionSetValues [](http://example.com/codeflow?start=76&length=46)

Can you update this comment?

CarlosFigueiraMSFT: Updated

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
```diff
@@ -292,21 +294,63 @@ public void TexlFunctionTypeSemanticsCountA()
...
+
+            bool IExternalOptionSet.IsBooleanValued => true;
+        }
+
+        [Theory]
+        [InlineData("CountIf(Table, A < 10)")]
+        [InlineData("CountIf(Table, A < 10, A > 0, A <> 2)")]
+        [InlineData("CountIf([1,2,3], Value) // Coercion from number to boolean")]
+        [InlineData("CountIf([\"false\",\"true\",\"false\"], Value) // Coercion from text to boolean")]
+        [InlineData("CountIf([1,2,3], BoolOptionSet.Yes) // Coercion from 2-valued option set to boolean")]
```
Shpakh-MSFT, line 312: >BoolOptionSet.Yes [](http://example.com/codeflow?start=38&length=17)

Can you add below case as well?

CountIf([1,2,3], If(true, 0 > 1, "true"))

Shpakh-MSFT (dismissed):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

Shpakh-MSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #878, opened 2022-12-02
Add support for display name for top level variables.

### LINKED ISSUE #779: Allow display names for top level variables.
Thinking more about display names ...  hitting some issues with Rules. 

Basically, I see 2 models:
1. **multiple orgs**  - like canvas apps. 
a. Need a "variable name" to distinguish accounts in multiple orgs, like "Accounts_1". 
b. The variable names are not localized. 
c. explicit add gesture - mainly so that we can get the variable name.

2. **single org** - like business rules / dataverse logic .  
a. Happy to just use "display name" 
b. would like them localized. 
c. don't want explicit add gesture - that just gets in their way. They just want the entire org.  
 

I think the practical feature here is : display names for top-level symbols. 
Today, we only do display names for 
- record fields. And so we leverage row scope.
- Option sets.  https://github.com/microsoft/Power-Fx/blob/2e9130f2352b3a04a910a4ca86c8e8cb44c4e2c3/src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs#L164  

This works fine for the multi-org case. But not so good for dataverse. 

Ie, in case 1, they say: 
`dataverse.AddTable("Accounts_1", "accounts"); ` // creates non-localizable symbol called "Accounts_1"

whereas in case 2, they just want to say:
`dataverse.AddEverything();` // creates localilzable symbols for every table in the org.  "Accounts/account", "Contacts/contact", etc ...  

I think practically, this means that when we're adding a symbol, we should be able to add it with a display name. 
And that this should also get translated with Engine.GetInvariant().

jas-valgotar:
- Adds support for display name for top level variable via Symbol Table.
- Added tests in src/tests/Microsoft.PowerFx.Interpreter.Tests/ConfigTests.cs/TopLevelVariableDisplayName and src/tests/Microsoft.PowerFx.Core.Tests/DisplayNameTests.cs/CollisionsThrow
- I have ran a buddy build with PA Client and have verified there's no impact of removing/changing the internal methods.

User's can now supply display name via Symbol Table as below:
var symbol = new Symbol Table();
symbol.AddVariable("logicalName", FormulaType, "DisplayName");

Fix #779

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -2749,6 +2749,10 @@ namespace Microsoft.PowerFx.Core.Binding
                     {
                         _txb.NodesToReplace.Add(new KeyValuePair<Token, string>(node.Token, entity.EntityName));
                     }
+                    else if(lookupInfo.Data is NameSymbol nameSymbol)
+                    {
+                        _txb.NodesToReplace.Add(new KeyValuePair<Token, string>(node.Token, nameSymbol.Name));
+                    }
                 }
 
                 // Internal control references are not allowed in component input properties.
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
@@ -126,7 +126,10 @@ namespace Microsoft.PowerFx
             }
         }
 
-        internal IEnumerable<NameLookupInfo> GetSymbols() => SymbolTable._variables.Values;
+        internal bool GetSymbols(string name, out NameLookupInfo symbol) => SymbolTable._variables.TryGetValue(name, out symbol);
+
+        internal IEnumerable<string> GetSuggestableSymbolName() => SymbolTable._variables.Keys;
+
 
         internal void AddEntity(IExternalEntity entity, DName displayName = default)
             => SymbolTable.AddEntity(entity, displayName);
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using Microsoft.PowerFx.Core;
+using Microsoft.PowerFx.Core.Binding.BindInfo;
 using Microsoft.PowerFx.Core.Entities;
 using Microsoft.PowerFx.Core.Functions;
 using Microsoft.PowerFx.Core.Texl;
@@ -125,10 +126,7 @@ namespace Microsoft.PowerFx
             }
         }
 
-        internal IEnumerable<IExternalEntity> GetSymbols() => SymbolTable._environmentSymbols.Values;
-
-        internal string GetSuggestableSymbolName(IExternalEntity entity)
-            => SymbolTable.GetSuggestableSymbolName(entity);
+        internal IEnumerable<NameLookupInfo> GetSymbols() => SymbolTable._variables.Values;
 
         internal void AddEntity(IExternalEntity entity, DName displayName = default)
             => SymbolTable.AddEntity(entity, displayName);
@@ -151,7 +149,7 @@ namespace Microsoft.PowerFx
             AddEntity(optionSet, optionalDisplayName);
         }
 
-        internal bool TryGetSymbol(DName name, out IExternalEntity symbol, out DName displayName)
-            => SymbolTable.TryGetSymbol(name, out symbol, out displayName);
+        internal bool TryGetVariable(DName name, out DName displayName)
+            => SymbolTable.TryGetVariable(name, out _, out displayName); 
     }
 }
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
@@ -197,13 +197,7 @@ namespace Microsoft.PowerFx
             return s;
         }
 
-        // https://github.com/microsoft/Power-Fx/issues/779
-        // _environmentSymbols have the display name support (they update the _environmentSymbolDisplayNameProvider)
-        // Merge _variables with _environmentSymbols to provide display name support for variables.. 
-        // Both ultimately serve in INameResolver.Lookup 
         internal readonly Dictionary<string, NameLookupInfo> _variables = new Dictionary<string, NameLookupInfo>();
-
-        internal readonly Dictionary<DName, IExternalEntity> _environmentSymbols = new Dictionary<DName, IExternalEntity>();
 
         internal DisplayNameProvider _environmentSymbolDisplayNameProvider = new SingleSourceDisplayNameProvider();
 
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
@@ -201,7 +201,7 @@ namespace Microsoft.PowerFx
         // _environmentSymbols have the display name support (they update the _environmentSymbolDisplayNameProvider)
         // Merge _variables with _environmentSymbols to provide display name support for variables.. 
         // Both ultimately serve in INameResolver.Lookup 
-        private protected readonly Dictionary<string, NameLookupInfo> _variables = new Dictionary<string, NameLookupInfo>();
+        internal readonly Dictionary<string, NameLookupInfo> _variables = new Dictionary<string, NameLookupInfo>();
 
         internal readonly Dictionary<DName, IExternalEntity> _environmentSymbols = new Dictionary<DName, IExternalEntity>();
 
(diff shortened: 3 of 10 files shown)


src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
```diff
@@ -298,45 +288,12 @@ bool INameResolver.Lookup(DName name, out NameLookupInfo nameInfo, NameLookupPre
...
-                        0,
-                        symbol,
-                        displayName);
-
-                    return true;
-                }
-                else
-                {
-                    throw new NotImplementedException($"{symbol.GetType().Name} not supported.");
-                }
```
jas-valgotar, line 339: This is now moved to SymbolTable.AddEntity()

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -92,16 +93,19 @@ internal void ValidateAccepts(ISymbolSlot slot, FormulaType newType)
...
-            if (displayName != null)
-            {
-                // Include parameter so that it's not a breaking change when we enable.
-                // https://github.com/microsoft/Power-Fx/issues/779
-                throw new NotImplementedException("DisplayName support for variables not implemented yet");
-            }
-
+        {
             Inc();
             ValidateName(name);
```
Shpakh-MSFT, line 98: >ValidateName(name); [](http://example.com/codeflow?start=12&length=19)

var dName = ValidateName(name);

You can use this are line 107 and 136

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -92,16 +93,19 @@ internal void ValidateAccepts(ISymbolSlot slot, FormulaType newType)
         /// <param name="mutable"></param>
         /// <param name="displayName"></param>
         public ISymbolSlot AddVariable(string name, FormulaType type, bool mutable = false, string displayName = null)
```
Shpakh-MSFT, line 95: >AddVariable [](http://example.com/codeflow?start=27&length=11)

Can you add a comment that this can throw in case of conflict?

MikeStall:
>         // https://github.com/microsoft/Power-Fx/issues/779

Can we delete this comment since this PR is fixing the issue?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs:197 in 918aafa. [](commit_id = 918aafabc0ced97fc3788c478046dca88c6a224a, deletion_comment = False)

src/tests/Microsoft.PowerFx.Interpreter.Tests/ConfigTests.cs
```diff
@@ -82,6 +82,33 @@ public void DisplayNameBinding()
...
+
+            var engine = new RecalcEngine(config);
+
+            var check = engine.Check(expression);
+            Assert.True(check.IsSuccess);
+
+            var eval = await engine.EvalAsync(expression,CancellationToken.None, runtimeConfig: r1);
+            Assert.Equal(expected, eval.ToObject());
+
+            var actualInvariantExpression = engine.GetInvariantExpression(expression, null);
```
MikeStall, line 108: >engine.GetInvariantExpression [](http://example.com/codeflow?start=44&length=29)

Can we run other direction too and test engine.GetDisplay?

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -151,7 +149,7 @@ public void AddOptionSet(OptionSet optionSet, DName optionalDisplayName = defaul
             AddEntity(optionSet, optionalDisplayName);
         }
 
-        internal bool TryGetSymbol(DName name, out IExternalEntity symbol, out DName displayName)
```
MikeStall, line 154: >TryGetSymbol [](http://example.com/codeflow?start=22&length=12)

This is potentially internal-breaking; we should confirm with PAClient build.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -2749,6 +2749,10 @@ public override void Visit(FirstNameNode node)
                     {
                         _txb.NodesToReplace.Add(new KeyValuePair<Token, string>(node.Token, entity.EntityName));
                     }
+                    else if(lookupInfo.Data is NameSymbol nameSymbol)
```
MikeStall, line 2752: >NameSymbol [](http://example.com/codeflow?start=47&length=10)

Should NameSymbol just implement IExternalEntity? 
(and then this check wouldn't be necessary)

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -92,16 +93,19 @@ internal void ValidateAccepts(ISymbolSlot slot, FormulaType newType)
...
             ValidateName(name);
+            DName displayDName = default;
+
+            if(displayName != null)
+            {
+                displayDName = ValidateName(displayName);
+            }
+            else
+            {
+                displayDName = new DName(name);
```
MikeStall, line 107: >displayDName = new DName(name); [](http://example.com/codeflow?start=16&length=31)

If there's no displayName, can we leave this null?
Or are we saying now that everything has a displayName, it's just that it may be same as logical name?

src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
```diff
@@ -198,7 +198,7 @@ public SymbolTable GetMutableCopyOfFunctions()
         // _environmentSymbols have the display name support (they update the _environmentSymbolDisplayNameProvider)
         // Merge _variables with _environmentSymbols to provide display name support for variables.. 
         // Both ultimately serve in INameResolver.Lookup 
-        private protected readonly Dictionary<string, NameLookupInfo> _variables = new Dictionary<string, NameLookupInfo>();
+        internal readonly Dictionary<string, NameLookupInfo> _variables = new Dictionary<string, NameLookupInfo>();
 
         internal readonly Dictionary<DName, IExternalEntity> _environmentSymbols = new Dictionary<DName, IExternalEntity>();
```
MikeStall, line 203: >_environmentSymbols [](http://example.com/codeflow?start=61&length=19)

Is this still used? 

It looks like we always use  _variables now?

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/SingleSourceDisplayNameProvider.cs
```diff
@@ -77,5 +77,24 @@ public override bool TryGetDisplayName(DName logicalName, out DName displayName)
         {
             return _logicalToDisplay.TryGetValue(logicalName, out displayName);
         }
+
+        public DisplayNameProvider RemoveField(DName lookupName)
```
MikeStall, line 81: >RemoveField [](http://example.com/codeflow?start=35&length=11)

Do we get full code-coverage on this method?

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseData/IntellisenseData.cs
```diff
@@ -314,7 +314,7 @@ internal virtual void AddCustomSuggestionsForGlobals()
         {
             foreach (var global in _powerFxConfig.GetSymbols())
             {
-                IntellisenseHelper.AddSuggestion(this, _powerFxConfig.GetSuggestableSymbolName(global), SuggestionKind.Global, SuggestionIconKind.Other, global.Type, requiresSuggestionEscaping: true);
```
MikeStall, line 317: >GetSuggestableSymbolName [](http://example.com/codeflow?start=70&length=24)

nice that we can get rid of GetSuggestableSymbolName and just use DisplayName!

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
jas-valgotar: Great point, I'll move 98 to 103, but will keep 136 as is since its already been validated.

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
jas-valgotar: I did do a buddy build which ran successfully.

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/SingleSourceDisplayNameProvider.cs
lesaltzm, line 81: Why return DisplayNameProvider and not SingleSourceDisplayNameProvider?

jas-valgotar: I have added test for the same.

jas-valgotar: you are correct! Thanks!

BruceHaley:

src/tests/Microsoft.PowerFx.Interpreter.Tests/ConfigTests.cs
jas-valgotar: updated.

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
jas-valgotar: "we saying now that everything has a displayName", I was planning to go with that approach but realized it can have side effects so moving away from that.

MikeStall:
>         private readonly ImmutableDictionary<DName, DName> _logicalToDisplay;

as an aside - now that we're adding remove, there's a concern about whether this dictionary can monotonically remove. 

Ie, would calling this in a loop leak memory? Or does ImmutableDictionary have some magic to stay compact?
  { Add(x); Remove(x); }


---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/SingleSourceDisplayNameProvider.cs:17 in 247dd0f. [](commit_id = 247dd0f44b62797700e45e7dfd09da1eb655d0c8, deletion_comment = False)

MikeStall (approved):
:shipit:

BruceHaley:

Issue #869, opened 2022-12-01
Text function doesn't handle date format specifiers like Excel


anderson-joyle:
PA 16176554

Excel's date format specifiers are case insensitive. The "mm" (or "MM") specifier should represent the hour if near a minute specifier, or month otherwise. The "hh" (or "HH") specifier should represent the hour in a 24-hour clock, except if an "am/pm" specifier is used.

Excel:
`=TEXT(DATE(2022,12,23) + TIME(17,30,59), "YYYY-MM-DD HH:MM:SS")` => `2022-12-23 17:30:59`
`=TEXT(DATE(2022,12,23) + TIME(17,30,59), "YYYY-MM-DD HH:MM:SS a/p")` => `2022-12-23 05:30:59 p`

Power Fx currently:
`Text(Date(2022,12,23) + Time(17,30,59), "YYYY-MM-DD HH:MM:SS")` => `"YYYY-12-DD 17:12:SS"`
`Text(Date(2022,12,23) + Time(17,30,59), "YYYY-MM-DD HH:MM:SS a/p")` => `"YYYY-12-DD 17:12:SS a/p"`

PR #890, opened 2022-12-07
Text function doesn't handle date format specifiers like Excel

anderson-joyle:
Issue #869. 

This PR expanded from its original scope and now is handling 4 bugs:

> Text function: millisecond format specifier should be '0' not 'f'

> Text function can't handle numbers formatted as dates or dates formatted as numbers

> Text function doesn't handle date format specifiers like Excel

> Text function should return an error if both numeric and date format specifiers are passed to it

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
@@ -110,31 +110,9 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
             {
                 // Verify statically that the format string doesn't contain BOTH numeric and date/time
                 // format specifiers. If it does, that's an error according to Excel and our spec.
-
-                // But firstly skip any locale-prefix
-                if (formatArg.StartsWith("[$-"))
+                if (!TextFormatUtils.IsValidFormatArg(formatArg, out _, out _))
                 {
-                    var end = formatArg.IndexOf(']', 3);
-                    if (end > 0)
-                    {
-                        formatArg = formatArg.Substring(end + 1);
-                    }
-                }
-
-                var hasDateTimeFmt = formatArg.IndexOfAny(new char[] { 'm', 'd', 'y', 'h', 'H', 's', 'a', 'A', 'p', 'P' }) >= 0;
-                var hasNumericFmt = formatArg.IndexOfAny(new char[] { '0', '#' }) >= 0;
-                if (hasDateTimeFmt && hasNumericFmt)
-                {
-                    // Check if the date time format contains '0's after the seconds specifier, which
-                    // is used for fractional seconds - in which case it is valid
-                    var formatWithoutZeroSubseconds = Regex.Replace(formatArg, @"[sS]\.?(0+)",
-                        m => m.Groups[1].Success ? "" : m.Groups[1].Value);
-                    hasNumericFmt = formatWithoutZeroSubseconds.IndexOfAny(new char[] { '0', '#' }) >= 0;
-                }
-
-                if (hasDateTimeFmt && hasNumericFmt)
-                {
-                    errors.EnsureError(DocumentErrorSeverity.Moderate, args[1], TexlStrings.ErrIncorrectFormat_Func, Name);
+                    errors.EnsureError(DocumentErrorSeverity.Severe, args[1], TexlStrings.ErrIncorrectFormat_Func, Name);
                     isValid = false;
                 }
             }
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
@@ -112,7 +112,7 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
                 // format specifiers. If it does, that's an error according to Excel and our spec.
                 if (!TextFormatUtils.IsValidFormatArg(formatArg, out _, out _))
                 {
-                    errors.EnsureError(DocumentErrorSeverity.Severe, args[1], TexlStrings.ErrIncorrectFormat_Func, Name);
+                    errors.EnsureError(DocumentErrorSeverity.Moderate, args[1], TexlStrings.ErrIncorrectFormat_Func, Name);
                     isValid = false;
                 }
             }
--- a/src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
@@ -46,25 +46,5 @@ namespace Microsoft.PowerFx.Core.Utils
 
             return true;
         }
-
-        public static bool IsNumericFormat(string formatArg)
-        {
-            if (formatArg == null)
-            {
-                return false;
-            }
-
-            return IsValidFormatArg(formatArg, out _, out var hasNumericFmt) && hasNumericFmt;
-        }
-
-        public static bool IsDateTimeFormat(string formatArg)
-        {
-            if (formatArg == null)
-            {
-                return false;
-            }
-
-            return IsValidFormatArg(formatArg, out var hasDateTimeFmt, out _) && hasDateTimeFmt;
-        }
     }
 }
--- a/src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
@@ -11,7 +11,7 @@ namespace Microsoft.PowerFx.Core.Utils
 {
     internal sealed class TextFormatUtils
     {
-        static public DateTime BaseDateTime => new DateTime(1899, 12, 30, 0, 0, 0, 0);
+        static public DateTime BaseDateTime => new DateTime(1899, 12, 30, 0, 0, 0, 0, DateTimeKind.Local);
 
         public static bool IsValidFormatArg(string formatArg, out bool hasDateTimeFmt, out bool hasNumericFmt)
         {
@@ -65,15 +65,6 @@ namespace Microsoft.PowerFx.Core.Utils
             }
 
             return IsValidFormatArg(formatArg, out var hasDateTimeFmt, out _) && hasDateTimeFmt;
-        }
-
-        public static DateTime NumberValueToDateTime(NumberValue numberValue)
-        {
-            var millisecondsPerDay = 86400000; // hours * minutes * seconds * 1000
-            var intPart = (int)Math.Floor(numberValue.Value);
-            var fracPart = numberValue.Value - intPart;
-
-            return BaseDateTime.AddDays(intPart).AddMilliseconds((int)(fracPart * millisecondsPerDay));
         }
     }
 }
--- a/src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
@@ -11,7 +11,7 @@ namespace Microsoft.PowerFx.Core.Utils
 {
     internal sealed class TextFormatUtils
     {
-        static public DateTime BaseDateTime => new DateTime(1899, 12, 30, 0, 0, 0, 0, DateTimeKind.Local);
+        static readonly private Regex _formatWithoutZeroSubsecondsRegex = new Regex(@"[sS]\.?(0+)", RegexOptions.Compiled);
 
         public static bool IsValidFormatArg(string formatArg, out bool hasDateTimeFmt, out bool hasNumericFmt)
         {
@@ -34,7 +34,7 @@ namespace Microsoft.PowerFx.Core.Utils
             {
                 // Check if the date time format contains '0's after the seconds specifier, which
                 // is used for fractional seconds - in which case it is valid
-                var formatWithoutZeroSubseconds = Regex.Replace(formatArg, @"[sS]\.?(0+)",
+                var formatWithoutZeroSubseconds = _formatWithoutZeroSubsecondsRegex.Replace(formatArg,
                     m => m.Groups[1].Success ? "" : m.Groups[1].Value);
                 hasNumericFmt = formatWithoutZeroSubseconds.IndexOfAny(new char[] { '0', '#' }) >= 0;
             }
--- a/src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
@@ -1,0 +1,79 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Text.RegularExpressions;
+using Microsoft.PowerFx.Core.Errors;
+using Microsoft.PowerFx.Core.Localization;
+using System.Xml.Linq;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Core.Utils
+{
+    internal sealed class TextFormatUtils
+    {
+        static public DateTime BaseDateTime => new DateTime(1899, 12, 30, 0, 0, 0, 0);
+
+        public static bool IsValidFormatArg(string formatArg, out bool hasDateTimeFmt, out bool hasNumericFmt)
+        {
+            // Verify statically that the format string doesn't contain BOTH numeric and date/time
+            // format specifiers. If it does, that's an error according to Excel and our spec.
+
+            // But firstly skip any locale-prefix
+            if (formatArg.StartsWith("[$-"))
+            {
+                var end = formatArg.IndexOf(']', 3);
+                if (end > 0)
+                {
+                    formatArg = formatArg.Substring(end + 1);
+                }
+            }
+
+            hasDateTimeFmt = formatArg.IndexOfAny(new char[] { 'm', 'M', 'd', 'D', 'y', 'Y', 'h', 'H', 's', 'S', 'a', 'A', 'p', 'P' }) >= 0;
+            hasNumericFmt = formatArg.IndexOfAny(new char[] { '0', '#' }) >= 0;
+            if (hasDateTimeFmt && hasNumericFmt)
+            {
+                // Check if the date time format contains '0's after the seconds specifier, which
+                // is used for fractional seconds - in which case it is valid
+                var formatWithoutZeroSubseconds = Regex.Replace(formatArg, @"[sS]\.?(0+)",
+                    m => m.Groups[1].Success ? "" : m.Groups[1].Value);
+                hasNumericFmt = formatWithoutZeroSubseconds.IndexOfAny(new char[] { '0', '#' }) >= 0;
+            }
+
+            if (hasDateTimeFmt && hasNumericFmt)
+            {
+                return false;
+            }
+
+            return true;
+        }
+
+        public static bool IsNumericFormat(string formatArg)
+        {
+            if (formatArg == null)
+            {
+                return false;
+            }
+
+            return IsValidFormatArg(formatArg, out _, out var hasNumericFmt) && hasNumericFmt;
+        }
+
+        public static bool IsDateTimeFormat(string formatArg)
+        {
+            if (formatArg == null)
+            {
+                return false;
+            }
+
+            return IsValidFormatArg(formatArg, out var hasDateTimeFmt, out _) && hasDateTimeFmt;
+        }
+
+        public static DateTime NumberValueToDateTime(NumberValue numberValue)
+        {
+            var millisecondsPerDay = 86400000; // hours * minutes * seconds * 1000
+            var intPart = (int)Math.Floor(numberValue.Value);
+            var fracPart = numberValue.Value - intPart;
+
+            return BaseDateTime.AddDays(intPart).AddMilliseconds((int)(fracPart * millisecondsPerDay));
+        }
+    }
+}
--- a/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
@@ -33,6 +33,8 @@ namespace Microsoft.PowerFx
 
         private readonly CancellationToken _cancellationToken;
 
+        internal CancellationToken CancelationToken => _cancellationToken;
+
         public EvalVisitor(CultureInfo cultureInfo, CancellationToken cancellationToken, ReadOnlySymbolValues runtimeConfig = null)
         {
             _defaultCultureInfo = cultureInfo;
--- a/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
@@ -33,7 +33,7 @@ namespace Microsoft.PowerFx
 
         private readonly CancellationToken _cancellationToken;
 
-        internal CancellationToken CancelationToken => _cancellationToken;
+        internal CancellationToken CancellationToken => _cancellationToken;
 
         public EvalVisitor(CultureInfo cultureInfo, CancellationToken cancellationToken, ReadOnlySymbolValues runtimeConfig = null)
         {
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs
@@ -108,6 +108,16 @@ namespace Microsoft.PowerFx.Functions
             });
         }
 
+        public static ErrorValue GenericInvalidArgument(IRContext irContext, string message)
+        {
+            return new ErrorValue(irContext, new ExpressionError()
+            {
+                Message = message,
+                Span = irContext.SourceContext,
+                Kind = ErrorKind.InvalidArgument
+            });
+        }
+
         public static ErrorValue InvalidBooleanFormatError(IRContext irContext)
         {
             return new ErrorValue(irContext, new ExpressionError()
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs
@@ -98,11 +98,11 @@ namespace Microsoft.PowerFx.Functions
             });
         }
 
-        public static ErrorValue GenericInvalidArgument(IRContext irContext, string message = null)
+        public static ErrorValue GenericInvalidArgument(IRContext irContext, string message = "Invalid Argument")
         {
             return new ErrorValue(irContext, new ExpressionError()
             {
-                Message = message ?? "Invalid Argument",
+                Message = message,
                 Span = irContext.SourceContext,
                 Kind = ErrorKind.InvalidArgument
             });
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs
@@ -98,21 +98,11 @@ namespace Microsoft.PowerFx.Functions
             });
         }
 
-        public static ErrorValue GenericInvalidArgument(IRContext irContext)
+        public static ErrorValue GenericInvalidArgument(IRContext irContext, string message = null)
         {
             return new ErrorValue(irContext, new ExpressionError()
             {
-                Message = "Invalid Argument",
-                Span = irContext.SourceContext,
-                Kind = ErrorKind.InvalidArgument
-            });
-        }
-
-        public static ErrorValue GenericInvalidArgument(IRContext irContext, string message)
-        {
-            return new ErrorValue(irContext, new ExpressionError()
-            {
-                Message = message,
+                Message = message ?? "Invalid Argument",
                 Span = irContext.SourceContext,
                 Kind = ErrorKind.InvalidArgument
             });
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
@@ -209,7 +209,7 @@ namespace Microsoft.PowerFx.Functions
                     {
                         // It's a number, formatted as date/time. Let's convert it to a date/time value first
                         var newDateTime = Library.NumberToDateTime(runner, context, IRContext.NotInSource(FormulaType.DateTime), new NumberValue[] { num });
-                        resultString = ExpandDateTimeExcelFormatSpecifiers(formatString, "g", newDateTime.Value, culture, runner.CancelationToken);
+                        resultString = ExpandDateTimeExcelFormatSpecifiers(formatString, "g", newDateTime.Value, culture, runner.CancellationToken);
                     }
                     else
                     {
@@ -242,7 +242,7 @@ namespace Microsoft.PowerFx.Functions
                     }
                     else
                     {
-                        resultString = ExpandDateTimeExcelFormatSpecifiers(formatString, "g", dateTimeValue.Value, culture, runner.CancelationToken);
+                        resultString = ExpandDateTimeExcelFormatSpecifiers(formatString, "g", dateTimeValue.Value, culture, runner.CancellationToken);
                     }
                     break;
             }
@@ -326,12 +326,14 @@ namespace Microsoft.PowerFx.Functions
 
         private static string ResolveDateTimeFormatAmbiguities(string format, DateTime dateTime, CultureInfo culture, CancellationToken cancellationToken)
         {
-            format = ReplaceDoubleQuotedStrings(format, out var replaceList, cancellationToken);        
-            format = TokenizeDatetimeFormat(format, cancellationToken);
-            format = DetokenizeDatetimeFormat(format, dateTime, culture);
-            format = RestoreDoubleQuotedStrings(format, replaceList, cancellationToken);
+            var resultString = format;
 
-            return format;
+            resultString = ReplaceDoubleQuotedStrings(resultString, out var replaceList, cancellationToken);
+            resultString = TokenizeDatetimeFormat(resultString, cancellationToken);
+            resultString = DetokenizeDatetimeFormat(resultString, dateTime, culture);
+            resultString = RestoreDoubleQuotedStrings(resultString, replaceList, cancellationToken);
+
+            return resultString;
         }
 
         private static string RestoreDoubleQuotedStrings(string format, List<string> replaceList, CancellationToken cancellationToken)
@@ -375,19 +377,19 @@ namespace Microsoft.PowerFx.Functions
 
             // Day component            
             format = _daysDetokenizeRegex.Replace(format, dateTime.ToString("dddd", culture))
-                          .Replace("\u0004\u0004\u0004", dateTime.ToString("dddd", culture).Substring(0, 3))
+                          .Replace("\u0004\u0004\u0004", dateTime.ToString("ddd", culture))
                           .Replace("\u0004\u0004", dateTime.ToString("dd", culture))
                           .Replace("\u0004", dateTime.ToString("%d", culture));
 
             // Month component
             format = _monthsDetokenizeRegex.Replace(format, dateTime.ToString("MMMM", culture))
-                          .Replace("\u0003\u0003\u0003", dateTime.ToString("MMMM", culture).Substring(0, 3))
+                          .Replace("\u0003\u0003\u0003", dateTime.ToString("MMM", culture))
                           .Replace("\u0003\u0003", dateTime.ToString("MM", culture))
                           .Replace("\u0003", dateTime.ToString("%M", culture));
 
             // Year component
             format = _yearsDetokenizeRegex.Replace(format, dateTime.ToString("yyyy", culture));
-            format = _years2DetokenizeRegex.Replace(format, dateTime.ToString("yyyy", culture).Substring(2, 2));
+            format = _years2DetokenizeRegex.Replace(format, dateTime.ToString("yy", culture));
 
             // Hour component
             format = _hoursDetokenizeRegex.Replace(format, hasAmPm ? dateTime.ToString("hh", culture) : dateTime.ToString("HH", culture))
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
@@ -22,20 +22,20 @@ namespace Microsoft.PowerFx.Functions
 {
     internal static partial class Library
     {
-        static readonly Regex _ampmReplaceRegex = new Regex("[aA][mM]\\/[pP][mM]", RegexOptions.Compiled);
-        static readonly Regex _apReplaceRegex = new Regex("[aA]\\/[pP]", RegexOptions.Compiled);
-        static readonly Regex _minutesBeforeSecondsRegex = new Regex("[mM][^dDyYhH]+[sS]", RegexOptions.Compiled);
-        static readonly Regex _minutesAfterHoursRegex = new Regex("[hH][^dDyYmM]+[mM]", RegexOptions.Compiled);
-        static readonly Regex _minutesRegex = new Regex("[mM]", RegexOptions.Compiled);
-        static readonly Regex _internalStringRegex = new Regex("([\"][^\"]*[\"])", RegexOptions.Compiled);
-        static readonly Regex _daysDetokenizeRegex = new Regex("[\u0004][\u0004][\u0004][\u0004]+", RegexOptions.Compiled);
-        static readonly Regex _monthsDetokenizeRegex = new Regex("[\u0003][\u0003][\u0003][\u0003]+", RegexOptions.Compiled);
-        static readonly Regex _yearsDetokenizeRegex = new Regex("[\u0005][\u0005][\u0005]+", RegexOptions.Compiled);
-        static readonly Regex _years2DetokenizeRegex = new Regex("[\u0005]+", RegexOptions.Compiled);
-        static readonly Regex _hoursDetokenizeRegex = new Regex("[\u0006][\u0006]+", RegexOptions.Compiled);
-        static readonly Regex _minutesDetokenizeRegex = new Regex("[\u000A][\u000A]+", RegexOptions.Compiled);
-        static readonly Regex _secondsDetokenizeRegex = new Regex("[\u0008][\u0008]+", RegexOptions.Compiled);
-        static readonly Regex _milisecondsDetokenizeRegex = new Regex("[\u000e][\u000e][\u000e]+", RegexOptions.Compiled);
+        static readonly private Regex _ampmReplaceRegex = new Regex("[aA][mM]\\/[pP][mM]", RegexOptions.Compiled);
+        static readonly private Regex _apReplaceRegex = new Regex("[aA]\\/[pP]", RegexOptions.Compiled);
+        static readonly private Regex _minutesBeforeSecondsRegex = new Regex("[mM][^dDyYhH]+[sS]", RegexOptions.Compiled);
+        static readonly private Regex _minutesAfterHoursRegex = new Regex("[hH][^dDyYmM]+[mM]", RegexOptions.Compiled);
+        static readonly private Regex _minutesRegex = new Regex("[mM]", RegexOptions.Compiled);
+        static readonly private Regex _internalStringRegex = new Regex("([\"][^\"]*[\"])", RegexOptions.Compiled);
+        static readonly private Regex _daysDetokenizeRegex = new Regex("[\u0004][\u0004][\u0004][\u0004]+", RegexOptions.Compiled);
+        static readonly private Regex _monthsDetokenizeRegex = new Regex("[\u0003][\u0003][\u0003][\u0003]+", RegexOptions.Compiled);
+        static readonly private Regex _yearsDetokenizeRegex = new Regex("[\u0005][\u0005][\u0005]+", RegexOptions.Compiled);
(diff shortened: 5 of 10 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -255,6 +257,76 @@ private static string ReplaceWith24HourClock(string format)
             return Regex.Replace(format, pattern, "HH${nonHours}", RegexOptions.IgnorePatternWhitespace);
         }
 
+        /// <summary>
+        /// Resolves month/minute ambiguity and 24h vs AM/PM.
+        /// </summary>
+        /// <param name="format">Format string.</param>
+        /// <returns></returns>
+        private static string ResolveAmbiguities(string format)
```
CarlosFigueiraMSFT, line 265: >ResolveAmbiguities [](http://example.com/codeflow?start=30&length=18)

Suggestion: ResolveDateTimeFormatAmbiguities

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_Format.txt
```diff
@@ -0,0 +1,200 @@
+﻿>> Text(DateTime(2022,12,1,23,45,59), "YYYY-MM-DD HH:MM:SS AM/PM")
+"2022-12-01 11:45:59 PM"
+
+>> Text(DateTime(2022,12,1,23,45,59), "yYyY-Mm-dd Hh:Mm:Ss Am/pM")
+"2022-12-01 11:45:59 PM"
+
+>> Text(DateTime(2022,12,1,23,45,59), "YYYY-MM-DD HH:MM:SS A/P")
+"2022-12-01 11:45:59 P"
+
+>> Text(DateTime(2022,12,1,23,45,59), "YYYY-MM-DD HH:MM:SS a/pa/p")
```
CarlosFigueiraMSFT, line 10: >a/pa/p [](http://example.com/codeflow?start=59&length=6)

It's nice that this is working in C# - it doesn't work in Power Apps, we'll need to fix it there.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -255,6 +257,76 @@ private static string ReplaceWith24HourClock(string format)
...
+            format = Regex.Replace(format, "[aA]\\/[pP]", "t");
+            format = Regex.Replace(format, "[hH]", time12hMatch.Success ? "h" : "H");
+            format = Regex.Replace(format, "[yY]", "y");
+            format = Regex.Replace(format, "[dD]", "d");
+            format = Regex.Replace(format, "[sS]", "s");
+
+            // All "m" char to upper case to express months
+            format = Regex.Replace(format, "[mM]", "M");
+
+            // Find all "m" chrs for minutes, before seconds
```
CarlosFigueiraMSFT, line 307: >chrs [](http://example.com/codeflow?start=28&length=4)

[typo] `char`

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -255,6 +257,76 @@ private static string ReplaceWith24HourClock(string format)
...
+            var minutesBeforeSecondsRegex = "[M]+[^dDyYhHmM]+[sS]";
+            var minutesBeforeSecondsMatch = Regex.Match(format, minutesBeforeSecondsRegex);
+
+            if (minutesBeforeSecondsMatch.Success)
+            {
+                var afterBeforeMinutes = format.Substring(minutesBeforeSecondsMatch.Index, minutesBeforeSecondsMatch.Length);
+                format = format.Substring(0, minutesBeforeSecondsMatch.Index) + afterBeforeMinutes.Replace('M', 'm') + format.Substring(minutesBeforeSecondsMatch.Index + minutesBeforeSecondsMatch.Length);
+            }
```
CarlosFigueiraMSFT, line 315: If we have more than one time format, this logic will not catch it. For example, `Text(Now(), "hh:mm:ss HH:MM:SS")`. Same with the next one.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -255,6 +257,76 @@ private static string ReplaceWith24HourClock(string format)
...
+            format = Regex.Replace(format, "[hH]", time12hMatch.Success ? "h" : "H");
+            format = Regex.Replace(format, "[yY]", "y");
+            format = Regex.Replace(format, "[dD]", "d");
+            format = Regex.Replace(format, "[sS]", "s");
+
+            // All "m" char to upper case to express months
+            format = Regex.Replace(format, "[mM]", "M");
+
+            // Find all "m" chrs for minutes, before seconds
+            var minutesBeforeSecondsRegex = "[M]+[^dDyYhHmM]+[sS]";
```
CarlosFigueiraMSFT, line 308: >"[M]+[^dDyYhHmM]+[sS]" [](http://example.com/codeflow?start=44&length=22)

[minor, perf] Since the Text function can be called multiple times, it may be worth saving those regular expressions in Regex objects so that they are compiled only once.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -255,6 +257,76 @@ private static string ReplaceWith24HourClock(string format)
...
+            var minutesBeforeSecondsMatch = Regex.Match(format, minutesBeforeSecondsRegex);
+
+            if (minutesBeforeSecondsMatch.Success)
+            {
+                var afterBeforeMinutes = format.Substring(minutesBeforeSecondsMatch.Index, minutesBeforeSecondsMatch.Length);
+                format = format.Substring(0, minutesBeforeSecondsMatch.Index) + afterBeforeMinutes.Replace('M', 'm') + format.Substring(minutesBeforeSecondsMatch.Index + minutesBeforeSecondsMatch.Length);
+            }
+
+            // Find all "m" chrs for minutes, after hours
+            var minutesAfterHoursRegex = "[hH][^dDyYmM]+[mM]+";
```
CarlosFigueiraMSFT, line 318: >mM [](http://example.com/codeflow?start=57&length=2)

Only need to search for uppercase M here - if it's already lower case we don't need to replace it again

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -255,6 +257,76 @@ private static string ReplaceWith24HourClock(string format)
...
+                return format;
+            }
+
+            var time12hMatch = Regex.Match(format, "[aA][mM]\\/[pP][mM]|[aA]\\/[pP]");
+
+            // Replace reserved chars
+            format = Regex.Replace(format, "[aA][mM]\\/[pP][mM]", "tt");
+            format = Regex.Replace(format, "[aA]\\/[pP]", "t");
+            format = Regex.Replace(format, "[hH]", time12hMatch.Success ? "h" : "H");
+            format = Regex.Replace(format, "[yY]", "y");
```
CarlosFigueiraMSFT, line 300: >Regex.Replace(format, "[yY]", "y") [](http://example.com/codeflow?start=21&length=34)

`format.Replace('Y', 'y')` does the same, and is less expensive than using regex. Same with other similar cases below.

MikeStall:
> >> Text(DateTime(2022,12,1,23,45,59), "YYYY-MM-DD HH:MM:SS AM/PM")

do these pass against JS backend too? 

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_Format.txt:1 in 757e98c. [](commit_id = 757e98c0d0d6f103b1bf2aa74c8b4d42e4bcaac3, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/NotYetReady/Text_Format.txt
```diff
@@ -0,0 +1,489 @@
+﻿>> Text(30470, "DDDD, MMM ""D"" DD, YYY") // excelDate22
+#skip
```
MikeStall, line 2: We don't run NotYetRady, so we don't need # skip here.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -255,6 +257,76 @@ private static string ReplaceWith24HourClock(string format)
...
+            if (format.Length == 0)
+            {
+                return format;
+            }
+
+            // Handling especial cases
+            if (format.Length == 1)
+            {
+                // Cases not supported by Excel are not handled ie 'G'
+                var singleCharMatch = Regex.Match(format, "[dDhHmMsSyY]");
```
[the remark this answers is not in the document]
anderson-joyle: Good point. Thanks.

anderson-joyle:
> >> Text(DateTime(2022,12,1,23,45,59), "YYYY-MM-DD HH:MM:SS AM/PM")

They do. The majority of these tests came from JS backend.

---
In reply to: [1341370797](https://github.com/microsoft/Power-Fx/pull/890#issuecomment-1341370797) [](http://example.com/codeflow?ancestors=1341370797,1341370797)

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_Format.txt:1 in 757e98c. [](commit_id = 757e98c0d0d6f103b1bf2aa74c8b4d42e4bcaac3, deletion_comment = False)

BruceHaley:

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -255,6 +257,66 @@ private static string ReplaceWith24HourClock(string format)
...
+        /// <param name="format">Format string.</param>
+        /// <returns></returns>
+        private static string ResolveDateTimeFormatAmbiguities(string format)
+        {
+            if (format.Length == 0)
+            {
+                return format;
+            }
+
+            // Handling especial cases
```
CarlosFigueiraMSFT, line 272: >Handling especial cases [](http://example.com/codeflow?start=15&length=23)

It is good to add that we are adding the '%' because that's what .NET requires (I didn't know that until I tried it today)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -255,6 +257,66 @@ private static string ReplaceWith24HourClock(string format)
...
+                    default:
+                        return $"%{format}";
+                }
+            }
+
+            Regex minutesBeforeSecondsRegex = new Regex("[M]+[^dDyYhHmM]+[sS]", RegexOptions.Compiled);
+            Regex minutesAfterHoursRegex = new Regex("[hH][^dDyYmM]+[M]+", RegexOptions.Compiled);
+            Regex time12hRegex = new Regex("[aA][mM]\\/[pP][mM]|[aA]\\/[pP]", RegexOptions.Compiled);
+            Regex ampmReplaceRegex = new Regex("[aA][mM]\\/[pP][mM]", RegexOptions.Compiled);
+            Regex apReplaceRegex = new Regex("[aA]\\/[pP]", RegexOptions.Compiled);
```
CarlosFigueiraMSFT, line 289: Move those to static readonly members in the class, so that the regex compilation only happens once per app domain (instead of every call to the Text function)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -255,6 +257,66 @@ private static string ReplaceWith24HourClock(string format)
...
+
+            format = ampmReplaceRegex.Replace(format, "tt");
+            format = apReplaceRegex.Replace(format, "t");
+            format = format.Replace('Y', 'y')
+                           .Replace('m', 'M') // All "m" char to upper case to express months
+                           .Replace('D', 'd')
+                           .Replace('H', hourChar)
+                           .Replace('h', hourChar)
+                           .Replace('S', 's')
+                           .Replace('F', 'f');
```
CarlosFigueiraMSFT, line 301: >.Replace('F', 'f'); [](http://example.com/codeflow?start=27&length=19)

The sub-second specifier should be `0`, not `f` (compatible with Excel).

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -255,6 +257,66 @@ private static string ReplaceWith24HourClock(string format)
             return Regex.Replace(format, pattern, "HH${nonHours}", RegexOptions.IgnorePatternWhitespace);
         }
 
+        /// <summary>
+        /// Resolves month/minute ambiguity and 24h vs AM/PM.
+        /// </summary>
+        /// <param name="format">Format string.</param>
+        /// <returns></returns>
+        private static string ResolveDateTimeFormatAmbiguities(string format)
```
MikeStall, line 265: >ResolveDateTimeFormatAmbiguities [](http://example.com/codeflow?start=30&length=32)

This looks like a complicated function ... can we create malformed inputs that will cause it to crash? or take long times?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
anderson-joyle: Right. Since PA is case insensitive, a single `d` would result in a totally different format than a day number in .NET.
I'll make sure to add the comment.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
anderson-joyle: This was intentionally left out since my next fix will involve the milliseconds formation.

BruceHaley:

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
anderson-joyle: This has been fixed within my latest push.

BruceHaley:

MikeStall:
> >> Text(0, "mm/dd/yy") // excelDate1

Great tests. 
It's nice this is in a separate file - makes it easier to organized (and disable on backends that don't support all these flags) 

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_Format.txt:1 in fb81cb7. [](commit_id = fb81cb7cd46002b219e5075faf60c3faf3556820, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -175,25 +194,37 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
                 }
             }
 
+            // We limit the format string size to 100 chars
+            if (formatString != null && formatString.Length > 100)
```
MikeStall, line 198: >100 [](http://example.com/codeflow?start=62&length=3)

this should be a const field. 
Then we don't have to repeat the literal on line 201.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -175,25 +194,37 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
...
+            {
+                // !JYL! do we need a 'translatable' error message?
```
MikeStall, line 200: >translatable [](http://example.com/codeflow?start=39&length=12)

there may be an existing one... or a generic "out or range" we can use.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -207,52 +238,209 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
...
+            format = ReplaceDoubleQuotedStrings(runner, format, out var replaceList);        
+            format = TokenizeDatetimeFormat(runner, format);
+            format = DetokenizeDatetimeFormat(format, dateTime, culture);
+            format = RestoreDoubleQuotedStrings(runner, format, replaceList);
 
             return format;
         }
 
-        private static string ReplaceWith24HourClock(string format)
+        private static string RestoreDoubleQuotedStrings(EvalVisitor runner, string format, List<string> repleaceList)
```
MikeStall, line 326: >EvalVisitor runner, [](http://example.com/codeflow?start=57&length=19)

Least privilege - do we need to pass EvalVisitor (a large powerful object); or can we just pass the CancellationToken?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -207,52 +238,209 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
             return CommonErrors.NotYetImplementedError(irContext, $"Text format for {args[0]?.GetType().Name}");
         }
 
-        internal static string ExpandDateTimeFormatSpecifiers(string format, CultureInfo culture)
+        internal static string ExpandDateTimeExcelFormatSpecifiers(EvalVisitor runner, string format, string defaultFormat, DateTime dateTime, CultureInfo culture)
```
MikeStall, line 241: >ExpandDateTimeExcelFormatSpecifiers [](http://example.com/codeflow?start=31&length=35)

This code is all ported from JS? 
Ie, so same algorithm we're using in Power Apps client-side?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
MikeStall, line 326: >repleaceList [](http://example.com/codeflow?start=105&length=12)

typo?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
anderson-joyle: This have been removed and is not long part of this PR.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
anderson-joyle: JS test cases have been provided with some samples of formats.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
anderson-joyle: Haven't found it. Ended up creating a new one.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
anderson-joyle: CancellationToken was not available from EvalVisitor. Now it is available as a field.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
anderson-joyle: Only the code that handles custom formats were ported from JS. All JS tests have been successful. @CarlosFigueiraMSFT please double-check if there is anything you would change.

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        public static DateTime NumberValueToDateTime(NumberValue numberValue)
+        {
+            var millisecondsPerDay = 86400000; // hours * minutes * seconds * 1000
+            var intPart = (int)Math.Floor(numberValue.Value);
+            var fracPart = numberValue.Value - intPart;
+
+            return BaseDateTime.AddDays(intPart).AddMilliseconds((int)(fracPart * millisecondsPerDay));
+        }
```
anderson-joyle, line 77: @CarlosFigueiraMSFT please double-check this against JS implementation.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -207,52 +238,209 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
...
             {
-                return format;
+                return dateTime.ToString(defaultFormat, culture);
+            }
+
+            // Is a number format?
+            if (TextFormatUtils.IsNumericFormat(format))
+            {
+                return (dateTime - TextFormatUtils.BaseDateTime).TotalDays.ToString(format ?? defaultFormat, culture);
+            }
```
anderson-joyle, line 252: @CarlosFigueiraMSFT please double-check this against JS implementation.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -1798,6 +1798,10 @@
     <value>Numeric value is too large.</value>
     <comment>Error Message.</comment>
   </data>
+  <data name="ErrTextFormatTooLarge" xml:space="preserve">
+    <value>Format string size is too large.</value>
```
MikeStall, line 1802: ><value>Format string size is too large.</value> [](http://example.com/codeflow?start=4&length=47)

nit - can we parameterize this

`Format string size can't be more than {0}  characters`

MikeStall (approved):

BruceHaley:

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -0,0 +1,79 @@
...
+                }
+            }
+
+            hasDateTimeFmt = formatArg.IndexOfAny(new char[] { 'm', 'M', 'd', 'D', 'y', 'Y', 'h', 'H', 's', 'S', 'a', 'A', 'p', 'P' }) >= 0;
+            hasNumericFmt = formatArg.IndexOfAny(new char[] { '0', '#' }) >= 0;
+            if (hasDateTimeFmt && hasNumericFmt)
+            {
+                // Check if the date time format contains '0's after the seconds specifier, which
+                // is used for fractional seconds - in which case it is valid
+                var formatWithoutZeroSubseconds = Regex.Replace(formatArg, @"[sS]\.?(0+)",
```
CarlosFigueiraMSFT, line 37: >@"[sS]\.?(0+)" [](http://example.com/codeflow?start=75&length=14)

Make this a private readonly static member, so that this regex is only compiled once.

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -0,0 +1,79 @@
...
+using Microsoft.PowerFx.Core.Errors;
+using Microsoft.PowerFx.Core.Localization;
+using System.Xml.Linq;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Core.Utils
+{
+    internal sealed class TextFormatUtils
+    {
+        static public DateTime BaseDateTime => new DateTime(1899, 12, 30, 0, 0, 0, 0);
```
CarlosFigueiraMSFT, line 14: >) [](http://example.com/codeflow?start=84&length=1)

`, DateTimeKind.Local`

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -0,0 +1,79 @@
...
+            if (formatArg.StartsWith("[$-"))
+            {
+                var end = formatArg.IndexOf(']', 3);
+                if (end > 0)
+                {
+                    formatArg = formatArg.Substring(end + 1);
+                }
+            }
+
+            hasDateTimeFmt = formatArg.IndexOfAny(new char[] { 'm', 'M', 'd', 'D', 'y', 'Y', 'h', 'H', 's', 'S', 'a', 'A', 'p', 'P' }) >= 0;
```
CarlosFigueiraMSFT, line 31: >new char[] { 'm', 'M', 'd', 'D', 'y', 'Y', 'h', 'H', 's', 'S', 'a', 'A', 'p', 'P' } [](http://example.com/codeflow?start=50&length=83)

[minor] consider making this a static readonly field, so that this array is only created once. Same with the next one

MikeStall (approved):
:shipit:

src/strings/PowerFxResources.en-US.resx
```diff
@@ -1798,6 +1798,10 @@
     <value>Numeric value is too large.</value>
     <comment>Error Message.</comment>
   </data>
+  <data name="ErrTextFormatTooLarge" xml:space="preserve">
+    <value>Format string size can't be more than {0} characters.</value>
```
CarlosFigueiraMSFT, line 1802: >can't [](http://example.com/codeflow?start=30&length=5)

We are very inconsistent with our usage of `can't` and `cannot`. We should choose one format and stick with it. Not needed for this PR, though.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
```diff
@@ -110,31 +110,9 @@ public override bool CheckTypes(CheckTypesContext checkTypesContext, TexlNode[]
...
-                if (formatArg.StartsWith("[$-"))
-                {
-                    var end = formatArg.IndexOf(']', 3);
-                    if (end > 0)
-                    {
-                        formatArg = formatArg.Substring(end + 1);
-                    }
-                }
-
-                var hasDateTimeFmt = formatArg.IndexOfAny(new char[] { 'm', 'd', 'y', 'h', 'H', 's', 'a', 'A', 'p', 'P' }) >= 0;
```
MikeStall, line 124: @CarlosFigueiraMSFT  - 

Most of this code moved to the new helper, but this line is changed to include more characters:
            hasDateTimeFmt = formatArg.IndexOfAny(new char[] { 'm', 'M', 'd', 'D', 'y', 'Y', 'h', 'H', 's', 'S', 'a', 'A', 'p', 'P' }) >= 0;

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_Format.txt
```diff
@@ -0,0 +1,531 @@
...
+
+>> Text(DateTime(1983, 6, 3, 2, 53, 9, 123), "YYYY-MM-DD HH:MM:SS.000") // datetime2
+"1983-06-03 02:53:09.123"
+
+>> Text(30470.01953125, "YYYY-MM-DD HH:MM:SS.000") // excelDatetime2
+"1983-06-03 00:28:07.500"
+
+>> Text(1, 
+  "mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm")
+Error({Kind:ErrorKind.Custom})
```
CarlosFigueiraMSFT, line 528: >Custom [](http://example.com/codeflow?start=22&length=6)

It should be an InvalidArgument error, not Custom

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -175,27 +198,34 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
                 }
             }
 
+            // We limit the format string size
+            if (formatString != null && formatString.Length > formatSize)
+            {
+                var customErrorMessage = StringResources.Get(TexlStrings.ErrTextFormatTooLarge, culture.Name);
+                return CommonErrors.CustomError(irContext, string.Format(customErrorMessage, formatSize));
```
CarlosFigueiraMSFT, line 205: >CustomError [](http://example.com/codeflow?start=36&length=11)

This should be an InvalidArgument error, not Custom (those are used by users)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -175,27 +198,34 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
...
+                    }
+                    else
+                    {
+                        resultString = num.Value.ToString(formatString ?? "g", culture);
+                    }                    
                     break;
                 case DateValue d:
-                    formatString = ExpandDateTimeFormatSpecifiers(formatString, culture);
-                    resultString = d.Value.ToString(formatString ?? "M/d/yyyy", culture);
+                    resultString = ExpandDateTimeExcelFormatSpecifiers(formatString, "M/d/yyyy", d.Value, culture, runner.CancelationToken);
```
CarlosFigueiraMSFT, line 222: >ExpandDateTimeExcelFormatSpecifiers [](http://example.com/codeflow?start=35&length=35)

We should also be able to format Date/Time values as numbers. For example, `Text(DateTime(2022, 12, 14, 18, 0, 0), "0.000")` should return `"44909.75"`.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
```diff
@@ -110,31 +110,9 @@ public override bool CheckTypes(CheckTypesContext checkTypesContext, TexlNode[]
...
-                    var formatWithoutZeroSubseconds = Regex.Replace(formatArg, @"[sS]\.?(0+)",
-                        m => m.Groups[1].Success ? "" : m.Groups[1].Value);
-                    hasNumericFmt = formatWithoutZeroSubseconds.IndexOfAny(new char[] { '0', '#' }) >= 0;
-                }
-
-                if (hasDateTimeFmt && hasNumericFmt)
+                if (!TextFormatUtils.IsValidFormatArg(formatArg, out _, out _))
                 {
-                    errors.EnsureError(DocumentErrorSeverity.Moderate, args[1], TexlStrings.ErrIncorrectFormat_Func, Name);
+                    errors.EnsureError(DocumentErrorSeverity.Severe, args[1], TexlStrings.ErrIncorrectFormat_Func, Name);
```
MikeStall, line 115: >Severe [](http://example.com/codeflow?start=61&length=6)

double check?

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
CarlosFigueiraMSFT, line 77: We should already have this logic in the unary operators, in the NumberToDateTime coercion.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
MikeStall: We should add as a test case ...
Text(1, "M#")

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -207,52 +237,209 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
...
+
+                match = _minutesAfterHoursRegex.Match(format);
+            }
+
+            format = Regex.Replace(format, "[mM]", "\u0003");
+            format = Regex.Replace(format, "[dD]", "\u0004");
+            format = Regex.Replace(format, "[yY]", "\u0005");
+            format = Regex.Replace(format, "[hH]", "\u0006");
+            format = Regex.Replace(format, "[sS]", "\u0008");
+            format = format.Replace('0', '\u000E');
```
CarlosFigueiraMSFT, line 440: That's a lot of regex operations. Not sure how much we will gain, but we can optimize it with something like

    var sb = new StringBuilder();
    foreach (var c in format) {
      switch(c) {
        case 'm': case 'M': sb.Append('\u0003'); break;
        case 'd': case 'D': sb.Append('\u0004'); break;
        case 'y': case 'Y': sb.Append('\u0005'); break;
        case 'h': case 'H': sb.Append('\u0006'); break;
        case 's': case 'S': sb.Append('\u0008'); break;
        case '0': sb.Append('\u000E'); break;
        default: sb.Append(c); break;
      }
    }
    format = sb.ToString();

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
anderson-joyle: Thanks for letting me know.

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
anderson-joyle: This regex is responsible for getting rid of the ambiguity between datatime format and numeric format. I'm under the impression that this regex is checking if we have seconds char (sS) followed by dot + zero(s). A "0" (zero) under this circumstances means this is a datetime format.

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
anderson-joyle: This code was moved from Text class so it can be re-used. I'm trying to keep it close to original as much as possible.

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
anderson-joyle: Per @CarlosFigueiraMSFT suggestion this has been replaced by existing library unary function (NumberToDateTime).

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
anderson-joyle: This code also doesn't handle well scoped chars. For example, the following should be handled as a datetime format string:
`Text(1, "M""#""")`

Binder doesn't take into account that `#` is scoped. Also happening on make.test.
Are we looking to fix this before merging this PR?

BruceHaley:

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
CarlosFigueiraMSFT: @MikeStall we talked about this validation before. I think we don't need to do it at compile time, leaving it to the runtime only - exactly because the validation is 1) incorrect (like in the example Anderson mentioned above) and 2) incomplete (for example, if the format comes from a variable).

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
CarlosFigueiraMSFT: Yes, you're right. But it can be done more efficiently without having specific logic to replace each captured group. Also, we should avoid using the static Regex.Replace(text, regularExpression), instead defining the regexes once as static members so that they are not recomputed every time the function is run.

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -0,0 +1,70 @@
...
+using Microsoft.PowerFx.Core.Errors;
+using Microsoft.PowerFx.Core.Localization;
+using System.Xml.Linq;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Core.Utils
+{
+    internal sealed class TextFormatUtils
+    {
+        static public DateTime BaseDateTime => new DateTime(1899, 12, 30, 0, 0, 0, 0, DateTimeKind.Local);
```
CarlosFigueiraMSFT, line 14: >BaseDateTime [](http://example.com/codeflow?start=31&length=12)

This is not needed anymore

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs
```diff
@@ -108,6 +108,16 @@ public static ErrorValue GenericInvalidArgument(IRContext irContext)
...
+        public static ErrorValue GenericInvalidArgument(IRContext irContext, string message)
+        {
+            return new ErrorValue(irContext, new ExpressionError()
+            {
+                Message = message,
+                Span = irContext.SourceContext,
+                Kind = ErrorKind.InvalidArgument
+            });
+        }
+
```
CarlosFigueiraMSFT, line 120: Consider merging this method with the one above (making the message parameter have a default value)

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -31,7 +31,9 @@ internal class EvalVisitor : IRNodeVisitor<ValueTask<FormulaValue>, EvalVisitorC
 
         private readonly ReadOnlySymbolValues _runtimeConfig;
 
-        private readonly CancellationToken _cancellationToken;
+        private readonly CancellationToken _cancellationToken;
+
+        internal CancellationToken CancelationToken => _cancellationToken;
```
CarlosFigueiraMSFT, line 36: >CancelationToken [](http://example.com/codeflow?start=35&length=16)

[spelling] `CancellationToken` (2 L's)

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
CarlosFigueiraMSFT: At a minimum - i.e., if we still want to keep this validation here - we must also do it at runtime, for expressions such as `With({format:"0.00 dd/mm/yyyy"},Text(123, format))` which are not evaluated at compile time.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -206,53 +254,214 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
...
+        }
+
+        private static string ResolveDateTimeFormatAmbiguities(string format, DateTime dateTime, CultureInfo culture, CancellationToken cancellationToken)
+        {
+            format = ReplaceDoubleQuotedStrings(format, out var replaceList, cancellationToken);        
+            format = TokenizeDatetimeFormat(format, cancellationToken);
+            format = DetokenizeDatetimeFormat(format, dateTime, culture);
+            format = RestoreDoubleQuotedStrings(format, replaceList, cancellationToken);
 
             return format;
```
CarlosFigueiraMSFT, line 334: >format [](http://example.com/codeflow?start=19&length=6)

[minor] What is returned isn't the format, but the actual formatted string; consider renaming it, or at least adding some comments to the method to clarify that.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs
anderson-joyle: Is there a practical technical again? Or just pattern?

CarlosFigueiraMSFT: Less code is almost always better... If we want to change this error kind in the future, for example, we would only have to update one place instead of 2 if it was a single method.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -206,53 +254,214 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
...
+
+            // Day component            
+            format = _daysDetokenizeRegex.Replace(format, dateTime.ToString("dddd", culture))
+                          .Replace("\u0004\u0004\u0004", dateTime.ToString("dddd", culture).Substring(0, 3))
+                          .Replace("\u0004\u0004", dateTime.ToString("dd", culture))
+                          .Replace("\u0004", dateTime.ToString("%d", culture));
+
+            // Month component
+            format = _monthsDetokenizeRegex.Replace(format, dateTime.ToString("MMMM", culture))
+                          .Replace("\u0003\u0003\u0003", dateTime.ToString("MMMM", culture).Substring(0, 3))
```
CarlosFigueiraMSFT, line 384: >dateTime.ToString("MMMM", culture).Substring(0, 3) [](http://example.com/codeflow?start=57&length=50)

`dateTime.ToString("MMM", culture)` - DateTime.ToString already handles abbreviated months

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -206,53 +254,214 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
...
+
+                format = format.Substring(0, pos) + "\u000A" + format.Substring(pos + 1);
+
+                match = _minutesAfterHoursRegex.Match(format);
+            }
+
+            var sb = new StringBuilder();
+            foreach (var c in format)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
```
CarlosFigueiraMSFT, line 450: >cancellationToken.ThrowIfCancellationRequested(); [](http://example.com/codeflow?start=16&length=49)

I don't think we need this code here. This is a very fast loop - it goes through at most 100 characters and appends a character into a StringBuilder instance. No need to check for cancellation for every single char.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -206,53 +254,214 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
...
+            return _internalStringRegex.Replace(format, "\u0011");
+        }
+
+        private static string DetokenizeDatetimeFormat(string format, DateTime dateTime, CultureInfo culture)
+        {
+            var hasAmPm = format.Contains('\u0001') || format.Contains('\u0002');
+
+            // Day component            
+            format = _daysDetokenizeRegex.Replace(format, dateTime.ToString("dddd", culture))
+                          .Replace("\u0004\u0004\u0004", dateTime.ToString("dddd", culture).Substring(0, 3))
```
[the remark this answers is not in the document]
anderson-joyle: Yes, I just wanted to be as much close to JS implementation as possible.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
anderson-joyle: Yes, I just wanted to be as much close to JS implementation as possible.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -206,53 +254,214 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
...
+
+            // Month component
+            format = _monthsDetokenizeRegex.Replace(format, dateTime.ToString("MMMM", culture))
+                          .Replace("\u0003\u0003\u0003", dateTime.ToString("MMMM", culture).Substring(0, 3))
+                          .Replace("\u0003\u0003", dateTime.ToString("MM", culture))
+                          .Replace("\u0003", dateTime.ToString("%M", culture));
+
+            // Year component
+            format = _yearsDetokenizeRegex.Replace(format, dateTime.ToString("yyyy", culture));
+            format = _years2DetokenizeRegex.Replace(format, dateTime.ToString("yyyy", culture).Substring(2, 2));
```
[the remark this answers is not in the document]
anderson-joyle: Yes, I just wanted to be as much close to JS implementation as possible.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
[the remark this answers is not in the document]
CarlosFigueiraMSFT: In JS we didn't have that ability to properly format date/time parts, in C# we do. We should use what the platform gives us.

BruceHaley:

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_Format.txt
```diff
@@ -0,0 +1,534 @@
...
+"1983-06-03 00:28:07.500"
+
+>> Text(1, 
+  "mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm")
+Error({Kind:ErrorKind.InvalidArgument})
+
+>> Text(1, "00.00 dd-mm-yyyy")
+Errors: Error 0-27: The function 'Text' has some invalid arguments.|Warning 8-26: Incorrect format specifier for 'Text'.
+
+>> Text(1, "M#")
```
MikeStall, line 533: This is a good test - let's confirm that it passes JSRunner tests in PAClient.

CarlosFigueiraMSFT (approved):
:shipit:

BruceHaley:

MikeStall (approved):
:shipit:

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_Format.txt
anderson-joyle: Confimed

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
MikeStall: Consider this test case:
Text(123.466, "[$-en-US]$#0.0M") // $123.5M

PR #892, opened 2022-12-08
Boolean(true) and Boolean([true]) should be valid.

### LINKED ISSUE #489: Boolean(true) shouldn't be invalid
Type conversion typically allow a value to convert to its current type. Looks like Power FX allows that for `Value` and `Text`, but not `Boolean`:

`Value(10)` returns `10`
`Text("hello")` returns `"hello"`

But
`Boolean(true)` is invalid.

This feels like a bug to me. Thoughts?

jas-valgotar:
- Fixes #489 
- Now `Boolean()` can take `Boolean` and `Table of Boolean` as a valid input.
- Changed and added new tests in src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Boolean.txt

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Boolean.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Boolean.cs
@@ -1,8 +1,11 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+using System;
 using System.Collections.Generic;
+using System.Linq;
 using Microsoft.PowerFx.Core.App.ErrorContainers;
+using Microsoft.PowerFx.Core.Errors;
 using Microsoft.PowerFx.Core.Functions;
 using Microsoft.PowerFx.Core.Localization;
 using Microsoft.PowerFx.Core.Types;
@@ -27,6 +30,30 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
         public BooleanFunction()
             : base(BooleanInvariantFunctionName, TexlStrings.AboutBoolean, FunctionCategories.Text, DType.Boolean, 0, 1, 1, DType.String)
         {
+        }
+
+        public override bool CheckTypes(TexlNode[] args, DType[] argTypes, IErrorContainer errors, out DType returnType, out Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
+        {
+            Contracts.AssertValue(args);
+            Contracts.AssertAllValues(args);
+            Contracts.AssertValue(argTypes);
+            Contracts.AssertAllValid(argTypes);
+            Contracts.Assert(args.Length == argTypes.Length);
+            Contracts.AssertValue(errors);
+            Contracts.Assert(MinArity <= args.Length && args.Length <= MaxArity);
+
+            nodeToCoercedTypeMap = null;
+
+            var isValid = true;
+            var argType = argTypes[0];
+            if (!(DType.Boolean.Accepts(argType) || DType.String.Accepts(argType)))
+            {
+                errors.EnsureError(DocumentErrorSeverity.Severe, args[0], TexlStrings.ErrNumberOrStringExpected);
+                isValid = false;
+            }
+
+            returnType = DType.Boolean;
+            return isValid;
         }
 
         public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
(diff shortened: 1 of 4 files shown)


BruceHaley:

anderson-joyle:
> Error({Kind:ErrorKind.InvalidArgument})

.NET allows this parse. For instance, the following is valid:
var validParse = bool.TryParse("  true", out var result);

validParse is true
result is true

Shouldn't PFx also allows the same parsing?
I know this is not part of your changes but it brought me the question.

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Boolean.txt:14 in af0e53b. [](commit_id = af0e53ba008c51643075937eefd2468151f6c15f, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
```diff
@@ -313,9 +313,21 @@ public static NumberValue BooleanToNumber(IRContext irContext, BooleanValue[] ar
...
+            {
+                return bv;
+            }
+            else if (args[0] is StringValue s)
+            {
+                val = s.Value;
+            }
+            else
+            {
+                CommonErrors.RuntimeTypeMismatch(irContext);
```
anderson-joyle, line 329: >CommonErrors.RuntimeTypeMismatch(irContext); [](http://example.com/codeflow?start=16&length=44)

Test coverage is not reaching this portion.

anderson-joyle:
Dont think this has to do with your changes but maybe is worth to ask.

The expression `Boolean([1,"True",true])` returns the following:
`Table({Value:true},{Value:Error({Kind:ErrorKind.InvalidArgument})},{Value:true})`

I'm wondering why `"True"` in this context is an invalid argument.

jas-valgotar:
Because we are a case sensitive language, only "true" is valid and Not "True".

---
In reply to: [1342925002](https://github.com/microsoft/Power-Fx/pull/892#issuecomment-1342925002) [](http://example.com/codeflow?ancestors=1342925002,1342925002)

anderson-joyle:
What about this test case:
```
>> Boolean(["true", "True", "TRUE", "false", "False", "FALSE"])
Table({Value:true},{Value:true},{Value:true},{Value:false},{Value:false},{Value:false})
```

There are different cases and none of them is invalid.

---
In reply to: [1342929108](https://github.com/microsoft/Power-Fx/pull/892#issuecomment-1342929108) [](http://example.com/codeflow?ancestors=1342929108,1342925002)

jas-valgotar:
I now see what's the issue, Table are always homogenous,
try, `[1, "True", true]` you'll see it is `Table({Value:"True"},{Value:"true"},{Value:"1"})`.  Because table will try to coerce everything to number and since string can't be coerced to a number that value is error, it essentially has nothing to do `Boolean()` in this case.

---
In reply to: [1342945287](https://github.com/microsoft/Power-Fx/pull/892#issuecomment-1342945287) [](http://example.com/codeflow?ancestors=1342945287,1342925002)

MikeStall:
> Error({Kind:ErrorKind.InvalidArgument})

Keep current semantics - they are rationale. 

If we do want to change the .txt, we should double check with Greg/Carlos. 



---
In reply to: [1342938174](https://github.com/microsoft/Power-Fx/pull/892#issuecomment-1342938174) [](http://example.com/codeflow?ancestors=1342938174,1342884861)

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Boolean.txt:14 in af0e53b. [](commit_id = af0e53ba008c51643075937eefd2468151f6c15f, deletion_comment = False)

BruceHaley:

BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Boolean.cs
```diff
@@ -143,7 +190,7 @@ public override bool CheckTypes(TexlNode[] args, DType[] argTypes, IErrorContain
             var arg = args[0];
             var argType = argTypes[0];
             fValid &= CheckNumericColumnType(argType, arg, errors, ref nodeToCoercedTypeMap);
-
+
```
bryancroteau-MSFT, line 193: Whitespace change?

MikeStall:
Adding Breaking Internal since we will need to update JS Translator to handle this. 

We could do this an IR transform. Practically that would mean:
- same test .txt updates
- but no changes to library*.cs files - because it's handled in IR.

jas-valgotar:
Closing this in favor of #907

PR #902, opened 2022-12-13
Provides a way to check if variable exists in RecalcEngine.

### LINKED ISSUE #882: Provide a way for RecalcEngine to check variable
Since Set() requires a variable to exist in the symbol table already, the host is responsible for ensuring that is the case somehow. However, RecalcEngine does not expose any methods for checking existence or type from the symbol table (the SymbolTable class does have a TryLookupSlot(), but recalc engine doesn't expose that).

Mike suggested something along the lines of `bool RecalcEngine.TryGetVariableType(string name, out FormulaType type)`

jas-valgotar:
-Fixes #882

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
@@ -394,9 +394,8 @@ namespace Microsoft.PowerFx
         public bool TryGetVariableType(string name, out FormulaType type)
         {
             type = default;
-            var symbols = CreateResolverInternal();
-            
-            if (symbols.Lookup(new DName(name), out var nameLookupInfo))
+
+            if (_symbolTable.TryGetVariable(new DName(name), out var nameLookupInfo, out _))
             {
                 type = FormulaType.Build(nameLookupInfo.Type);
                 return true;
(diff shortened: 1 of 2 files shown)


BruceHaley:

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -386,5 +390,18 @@ internal bool TryGetByName(string name, out RecalcFormulaInfo info)
             info = null;
             return false;
         }
+
+        public bool TryGetVariableType(string name, out FormulaType type)
+        {
+            type = default;
+            var symbols = CreateResolverInternal();
```
MikeStall, line 397: >var symbols = CreateResolverInternal(); [](http://example.com/codeflow?start=12&length=39)

This should have more symmetry with above methods like TryGetByName. 

We can probably just check against _symbolTable directly rather than call CreateResolverInternal() 

This means we're just operating on the current variables added here. 
If we want something more general, it should be on Engine base class.

jas-valgotar: updated.

MikeStall (approved):

BruceHaley:

PR #903, opened 2022-12-13
Fix [Add/Drop]Columns tests

CarlosFigueiraMSFT:
Some of the recently added tests for those functions have an extra `;` which is making the error that the test was intended to cover.


CarlosFigueiraMSFT:
There are some extra files in this PR; will close this one and create a new one from scratch.

PR #904, opened 2022-12-13
Fix [Add/Drop]Columns tests

CarlosFigueiraMSFT:
Some of the recently added tests for those functions have an extra `;`, which is making the error that the test was intended to cover.


CarlosFigueiraMSFT:
Something is wrong - I updated only two files, but the PR has 8 of them :|

PR #914, opened 2022-12-15
Enable DelegationValidationStrategy to determine if a function can be delegated without reporting warnings

ian-legler:
In order to better understand the challenges that users have in using functions and data sources that can't be delegated today, we need to collect failure telemetry without causing breaking UI changes.

This change adds the ability for a function to create delegation validation strategies that do not generate warnings.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
@@ -702,7 +702,7 @@ namespace Microsoft.PowerFx.Core.Functions
 
         // Returns true if function is row scoped and supports delegation.
         // Needs to be overriden by functions (For example, IsBlank) which are not server delegatable themselves but can become one when scoped inside a delegatable function.
-        public virtual bool IsRowScopedServerDelegatable(CallNode callNode, TexlBinding binding, OperationCapabilityMetadata metadata)
+        public virtual bool IsRowScopedServerDelegatable(CallNode callNode, TexlBinding binding, OperationCapabilityMetadata metadata, bool generateHints = true)
         {
             Contracts.AssertValue(callNode);
             Contracts.AssertValue(binding);
@@ -1209,7 +1209,7 @@ namespace Microsoft.PowerFx.Core.Functions
             return false;
         }
 
-        public virtual IOpDelegationStrategy GetOpDelegationStrategy(BinaryOp op, BinaryOpNode opNode)
+        public virtual IOpDelegationStrategy GetOpDelegationStrategy(BinaryOp op, BinaryOpNode opNode, bool generateHints = true)
         {
             Contracts.AssertValueOrNull(opNode);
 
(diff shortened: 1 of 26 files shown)


BruceHaley:

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -702,7 +702,7 @@ public virtual bool SupportsPaging(CallNode callNode, TexlBinding binding)
 
         // Returns true if function is row scoped and supports delegation.
         // Needs to be overriden by functions (For example, IsBlank) which are not server delegatable themselves but can become one when scoped inside a delegatable function.
-        public virtual bool IsRowScopedServerDelegatable(CallNode callNode, TexlBinding binding, OperationCapabilityMetadata metadata)
+        public virtual bool IsRowScopedServerDelegatable(CallNode callNode, TexlBinding binding, OperationCapabilityMetadata metadata, bool generateHints = true)
```
MikeStall, line 705: >generateHints [](http://example.com/codeflow?start=140&length=13)

Where do we call with generateHints = False?

src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
```diff
@@ -348,8 +348,9 @@ bool INameResolver.LookupEnumValueByTypeAndLocName(DType enumType, DName locName
 
         // Methods from INameResolver that we default / don't implement
         IExternalDocument INameResolver.Document => default;
-
-        IExternalEntityScope INameResolver.EntityScope => default;
+
+        internal virtual IExternalEntityScope InternalEntityScope => default;
```
MikeStall, line 352: >InternalEntityScope [](http://example.com/codeflow?start=46&length=19)

Why do we need this? 

PowerApps bypasses the SymbolTable classes and uses INameResolver directly.

MikeStall:
Is there some way to avoid the viral updates for GenerateHints, such as by logging them but ignoring them? like what we do in overload resolution and here:
https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Core/App/ErrorContainers/LimitedSeverityErrorContainer.cs

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
ian-legler: It will only be called from functions that are in PowerApps today

src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
ian-legler: This is needed for the unit tests, since the delegation validation logic is still using the SymbolTable

MikeStall:
Superceded by #960 ?

ian-legler:
Abandoning

PR #916, opened 2022-12-16
Copy Repl project into PowerFx repo 

MikeStall:
From Samples:
https://github.com/microsoft/power-fx-host-samples/blob/main/Samples/ConsoleREPL/ConsoleREPL.cs

We still want Samples copy since that goes against actual nugets. 
But a copy in the PowerFx repo is convenient for directly testing the SDK. 

Lots of whitespace differences because I had to bring up to coding conventions for this repo. 

Also leverage some recent functionality:
- replace a hack Set() RegEx with using official APIs (parse, etc) 
- FormulaValue.ToExpression to get  string from arbitrary expression.

### DIFF
--- a/src/tools/Repl/Program.cs
+++ b/src/tools/Repl/Program.cs
@@ -1,0 +1,559 @@
+﻿// <copyright file="Program.cs" company="PlaceholderCompany">
+// Copyright (c) PlaceholderCompany. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using Microsoft.PowerFx;
+using Microsoft.PowerFx.Core;
+using Microsoft.PowerFx.Types;
+
+namespace PowerFxHostSamples
+{
+    public static class ConsoleRepl
+    {
+        private static RecalcEngine engine;
+        private static bool _formatTable = true;
+        private const string OptionFormatTable = "FormatTable";
+
+        private static void ResetEngine()
+        {
+            Features toenable = 0;
+            foreach (Features feature in (Features[])Enum.GetValues(typeof(Features)))
+            {
+                toenable |= feature;
+            }
+
+            var config = new PowerFxConfig(toenable);
+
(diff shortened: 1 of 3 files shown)


src/tools/Repl/Program.cs
```diff
@@ -0,0 +1,601 @@
...
+
+            Console.WriteLine($"Experimental features enabled:{enabled}");
+
+            Console.WriteLine($"Enter Excel formulas.  Use \"Help()\" for details.");
+
+            REPL(Console.In, false);
+        }
+
+        // Pattern match for Set(x,y) so that we can define the variable
+        public static bool TryMatchSet(string expr, out string arg0name, out FormulaValue varValue)
```
MikeStall, line 80: This removes a hacky Set regex and uses "Official" APIs. 


Also see
https://github.com/microsoft/Power-Fx/discussions/889

src/tools/Repl/Program.cs
```diff
@@ -0,0 +1,601 @@
...
+                try
+                {
+                    // variable assignment: Set( <ident>, <expr> )
+                    if (TryMatchSet(expr, out var varName, out var varValue))
+                    {
+                        Console.WriteLine(varName + ": " + PrintResult(varValue));
+                    }
+
+                    // formula definition: <ident> = <formula>
+                    else if ((match = Regex.Match(expr, @"^\s*(?<ident>\w+)\s*=(?<formula>.*)$", RegexOptions.Singleline)).Success)
```
MikeStall, line 141: Would like to get rid of these RegEx, but the corresponding Parse functions are still internal...

src/tools/Repl/Program.cs
```diff
@@ -0,0 +1,601 @@
...
+                            separator = ", ";
+                        }
+
+                        resultString += "]";
+                    }
+                }
+            }
+            else
+            {
+                resultString = value.ToExpression();
```
MikeStall, line 433: This is also a nice simplification - we now have more reliable ToString functions for FormulaValues.

LucGenetier (dismissed):

BruceHaley:

LucGenetier (approved):

PR #931, opened 2022-12-20
Revert "Text function doesn't handle date format specifiers like Excel"

MikeStall:
Reverts microsoft/Power-Fx#890. This is causing a failure that's blocking updating powerapps. 

This updates Text.cs function with additional checks that cause new failures on this expression:

`Text(123.466, "[$-en-US]$#0.0M") // $123.5M`

But the new PR makes it an error because it treats the 'M' as a format string and not a arbitrary character.  It looks like "unknown" characters just get copied to the output.  And so basically any change to Text() is a breaking change,


BruceHaley:

MikeStall:
Trying more tactical fix  #933

Issue #932, opened 2022-12-20
Text(123.466, "[$-en-US]$#0.0M")


MikeStall:
PowerApps supports this:
`Text(123.466, "[$-en-US]$#0.0M") // $123.5M`

The 'M' is treated as an unrecognized character and emitted directly to output. 

PowerApps changes this for Excel compact. 
#890  is the corresponding change in the interpreter.

MikeStall:
Not sure what the right fix is here (is this by-design, is there a bug , etc). Determine correct behavior and handle accordingly.

gregli-msft:
This appears to have been fixed.  It should produce an error, same as Canvas and Excel.

```
> Text(123.466, "[$-en-US]$#0.0M")
Error: InvalidArgument - Incorrect format specifier for 'Text'.
```

PR #933, opened 2022-12-20
Revert portion of #890

MikeStall:
#890 makes some Text() format arguments be compile-time errors (if they're literal).  This is handled in Text.cs and shared by all backends -so it impacts PowerApps.
Previously, they may have been runtime errors - which means each backend handles separately. 

Revert the change to Text.cs because that is shared with PowerApps - the change is making this a compile-time error (instead of runtime), which is causing PA to fail at compile time.  
Reverting this so that in effect 890 has no impact on PowerApps (the rest of 890 is interpreter-only changes). 

Here is a specific expression causing a problem: 
`Text(123.466, "[$-en-US]$#0.0M")`

Interpreter doesn't implement it yet (#932), so adding override to call it out.


BruceHaley:

rick-nguyen (approved):

PR #937, opened 2022-12-22
CustomFunction - missing async case Fixes #935

### LINKED ISSUE #935: CustomFunction - missing async case
When a custom function derives from base ctor directly, it skips skips calling Scan() and gets the parameters from the ctor directly. 

This is skipping setting _info.async . So if it's Execute() is async, we will still treat it as sync and crash. 

    public class XValidateEmailFunction : ReflectionFunction
    {   
        public XValidateEmailFunction() 
            : base("XValidateEmail", FormulaType.Number, FormulaType.String) // Causes Scan to be skipped
        {
        }

        public async Task<FormulaValue> Execute(StringValue email)
        {
        }
    }

We missed a case in #40 . 

https://github.com/microsoft/Power-Fx/blob/9c92baa7791ef52d46a9180b6506d95e3b3cd7b3/src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs#L182

jas-valgotar:
-Fixes #935

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
@@ -181,7 +181,7 @@ namespace Microsoft.PowerFx
 
             var isAsync = m.ReturnType.BaseType == typeof(Task);
 
-            var configType = ConfigType ?? default(Type);
+            var configType = ConfigType ?? default;
 
             _info = new FunctionDescr(name, m, returnType, paramTypes, configType, BigInteger.Zero, isAsync);
         }
@@ -241,7 +241,7 @@ namespace Microsoft.PowerFx
 
                 var configType = default(Type);
 
-                BigInteger lamdaParamMask = default(BigInteger);
+                BigInteger lamdaParamMask = default;
                 for (var i = 0; i < parameters.Length; i++)
                 {
                     if (i == parameters.Length - 1 && isAsync)
@@ -280,8 +280,6 @@ namespace Microsoft.PowerFx
                         throw new InvalidOperationException($"Unknown parameter type: {parameters[i].Name}, {parameters[i].ParameterType}");
                     }
                 }
-
-                var paramTypes = paramTypes.ToArray();
 
                 _info = new FunctionDescr(name, m, returnType, paramTypes.ToArray(), configType, lamdaParamMask, isAsync);
             }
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
```diff
@@ -184,8 +184,9 @@ protected ReflectionFunction(string name, FormulaType returnType, params Formula
                 Name = name,
                 RetType = returnType,
                 ParamTypes = paramTypes,
-                _method = m
-            };
+                _method = m,
+                _isAsync = m.ReturnType.BaseType == typeof(Task)
```
MikeStall, line 188: >_isAsync [](http://example.com/codeflow?start=16&length=8)

Just noticed we're also missing _configType.

Maybe we should make FunctionDescr have a ctor 
and make these properties private  - to enforce they're getting set in both places.

MikeStall:
>                     throw new InvalidOperationException($"Call to {_info.Name} is missing config type {_info._configType.FullName}");

When my custom function (which uses the direct ctor instead of Scan()) also take a config, but that config is missing, 

I'm hitting in a null ref here because _configType is null. 

Actually, maybe this just should be ConfigType


---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs:342 in 9c26970. [](commit_id = 9c26970f6de9350dec5329a22d530d29437266be, deletion_comment = False)

BruceHaley:

LucGenetier:
✅ No public API change.

BruceHaley:

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/CustomFunctions.cs
```diff
@@ -68,6 +68,11 @@ public void CustomFunctionErrorOrBlank(string script, string expectedResult, boo
         // Must have "Function" suffix. 
         private class TestCustomFunction : ReflectionFunction
         {
+            public TestCustomFunction():
```
MikeStall, line 71: Did you intend to add this?
This will cause Scan() to be skipped for this function and change test cases.

src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
jas-valgotar: updated to use Ctor

src/tests/Microsoft.PowerFx.Interpreter.Tests/CustomFunctions.cs
jas-valgotar: you are right! this was by mistake...

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

Issue #895, opened 2022-12-09
DateTimeValue(UntypedObject) should allow higher precision of decimal fraction than 3


susosuso:
When parsing an UntypedObject string to DateTime object, I noticed it has very strict precision for second decimal fractions.

```DateTimeValue(ParseJSON("""2011-01-30T08:00:00.000Z"""))``` works, but

```DateTimeValue(ParseJSON("""2011-01-30T08:00:00.0005Z"""))``` will fail.

My guess is that this was intended to keep compatibility with MySql ? but my two cents is that Power Fx should be more flexible, especially when feeding it with plain string work just fine:
`DateTimeValue("2011-01-30T08:00:00.0005Z")` works

https://github.com/microsoft/Power-Fx/blob/2e9130f2352b3a04a910a4ca86c8e8cb44c4e2c3/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUntypedObject.cs#L22

Not super important, as we can drop the fractions if needed, despite losing some precision.

susosuso:
similarly, for TimeValue:

`TimeValue(ParseJSON("""11:12:00.000"""))` works, but
`TimeValue(ParseJSON("""11:12:00"""))` fails.

https://github.com/microsoft/Power-Fx/blob/2e9130f2352b3a04a910a4ca86c8e8cb44c4e2c3/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUntypedObject.cs#L184

MikeStall:
From Carlos: "We should allow that. Since we decided that we will support up to 7 fractional digits in date/time, that should work as well."

We should update the .txt tests as well to reflect this.

PR #939, opened 2022-12-27
Supporting millisenconds with up to 7 digits.

anderson-joyle:
Issue #895.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUntypedObject.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUntypedObject.cs
@@ -218,9 +218,7 @@ namespace Microsoft.PowerFx.Functions
             {
                 var s = impl.GetString();
 
-                TimeSpan res;
-
-                if (TimeSpan.TryParseExact(s, @"hh\:mm\:ss\.FFFFFFF", CultureInfo.InvariantCulture, TimeSpanStyles.None, out res) ||
+                if (TimeSpan.TryParseExact(s, @"hh\:mm\:ss\.FFFFFFF", CultureInfo.InvariantCulture, TimeSpanStyles.None, out var res) ||
                     TimeSpan.TryParseExact(s, @"hh\:mm\:ss", CultureInfo.InvariantCulture, TimeSpanStyles.None, out res))
                 {
                     return new TimeValue(irContext, res);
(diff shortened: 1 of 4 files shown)


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUntypedObject.cs
```diff
@@ -217,7 +217,11 @@ public static FormulaValue TimeValue_UO(IRContext irContext, UntypedObjectValue[
             if (impl.Type == FormulaType.String)
             {
                 var s = impl.GetString();
-                if (TimeSpan.TryParseExact(s, @"hh\:mm\:ss\.FFF", CultureInfo.InvariantCulture, TimeSpanStyles.None, out TimeSpan res))
+
+                TimeSpan res;
+
+                if (TimeSpan.TryParseExact(s, @"hh\:mm\:ss\.FFFFFFF", CultureInfo.InvariantCulture, TimeSpanStyles.None, out res) ||
```
MikeStall, line 223: Is this logic shared with DateTimeValue(string)? (And should it be?)

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUntypedObject.cs
anderson-joyle: Good question. DateTimeValue and TimeValue functions take locale as arguments. On the other hand, the untyped versions don't and always expect an invariant. That makes me think we don't need to share this logic with DateTimeValue /TimveValue.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUntypedObject.cs
```diff
@@ -217,7 +217,9 @@ public static FormulaValue TimeValue_UO(IRContext irContext, UntypedObjectValue[
             if (impl.Type == FormulaType.String)
             {
                 var s = impl.GetString();
-                if (TimeSpan.TryParseExact(s, @"hh\:mm\:ss\.FFF", CultureInfo.InvariantCulture, TimeSpanStyles.None, out TimeSpan res))
+
+                if (TimeSpan.TryParseExact(s, @"hh\:mm\:ss\.FFFFFFF", CultureInfo.InvariantCulture, TimeSpanStyles.None, out var res) ||
```
MikeStall, line 221: Possible future item - can / should this code be shared with the none UO time function?

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #938, opened 2022-12-27
Collect to return a record type.

anderson-joyle:
Issues #915 and #920.

PR #949, opened 2023-01-03
Extra test cases to test scenario that was failing previously.

anderson-joyle:
Previous to PR #938 the following test case was not working:
`Collect(x.t,{Value:200})`

Assuming `x` is `{v:1,t:[1,2,3]}`


LucGenetier:
✅ No public API change.

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #950, opened 2023-01-03
Allow more than 3 subsecond digits in the Text function

CarlosFigueiraMSFT:
.NET supports up to 7 sub-second digits, so we should support it in the Text function as well.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
@@ -219,14 +219,7 @@ namespace Microsoft.PowerFx.Functions
                     {
                         // It's a number, formatted as date/time. Let's convert it to a date/time value first
                         var newDateTime = Library.NumberToDateTime(runner, context, IRContext.NotInSource(FormulaType.DateTime), new NumberValue[] { num });
-                        try
-                        {
-                            resultString = ExpandDateTimeExcelFormatSpecifiers(formatString, "g", newDateTime.Value, culture, runner.CancellationToken);
-                        }
-                        catch (FormatException)
-                        {
-                            return CommonErrors.GenericInvalidArgument(irContext, StringResources.Get(TexlStrings.ErrTextInvalidFormat, culture.Name));
-                        }
+                        return ExpandDateTimeExcelFormatSpecifiers(irContext, formatString, "g", newDateTime.Value, culture, runner.CancellationToken);
                     }
                     else
                     {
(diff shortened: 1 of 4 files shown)


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -219,7 +219,14 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
...
                         // It's a number, formatted as date/time. Let's convert it to a date/time value first
                         var newDateTime = Library.NumberToDateTime(runner, context, IRContext.NotInSource(FormulaType.DateTime), new NumberValue[] { num });
-                        resultString = ExpandDateTimeExcelFormatSpecifiers(formatString, "g", newDateTime.Value, culture, runner.CancellationToken);
+                        try
+                        {
+                            resultString = ExpandDateTimeExcelFormatSpecifiers(formatString, "g", newDateTime.Value, culture, runner.CancellationToken);
+                        }
+                        catch (FormatException)
+                        {
+                            return CommonErrors.GenericInvalidArgument(irContext, StringResources.Get(TexlStrings.ErrTextInvalidFormat, culture.Name));
```
anderson-joyle, line 228: >return CommonErrors.GenericInvalidArgument(irContext, StringResources.Get(TexlStrings.ErrTextInvalidFormat, culture.Name)); [](http://example.com/codeflow?start=28&length=123)

This portion is not being reach by test coverage. Can we have the following test case added:
```
>> Text(30470.01953125, "yyyy-mm-dd hh:mm:ss.00000000")
```

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -219,7 +219,14 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
                     {
                         // It's a number, formatted as date/time. Let's convert it to a date/time value first
                         var newDateTime = Library.NumberToDateTime(runner, context, IRContext.NotInSource(FormulaType.DateTime), new NumberValue[] { num });
-                        resultString = ExpandDateTimeExcelFormatSpecifiers(formatString, "g", newDateTime.Value, culture, runner.CancellationToken);
+                        try
```
anderson-joyle, line 222: >try [](http://example.com/codeflow?start=24&length=3)

Nice touch adding this exception handler since it is possible to get an invalid format.
Would be a good idea (or worth the effort) to move this try/catch into ExpandDateTimeExcelFormatSpecifiers?

CarlosFigueiraMSFT: Done

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
CarlosFigueiraMSFT: Added

LucGenetier:
✅ No public API change.

anderson-joyle (approved):

MikeStall (approved):

PR #960, opened 2023-01-05
Update filter delegation validation to validate without generating warnings

ian-legler:
In order to better understand the challenges that users have in using functions and data sources that can't be delegated today, we need to collect failure telemetry without causing breaking UI changes.

This change adds the ability filter delegation validation to swallow warnings

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -121,7 +121,7 @@ namespace Microsoft.PowerFx.Core.Binding
 
         public bool HasLocalScopeReferences { get; private set; }
 
-        public ErrorContainer ErrorContainer { get; } = new ErrorContainer();
+        public ErrorContainer ErrorContainer { get; set; } = new ErrorContainer();
 
         /// <summary>
         /// The maximum number of selects in a table that will be included in data call.
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/FilterDelegationBase.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/FilterDelegationBase.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using System.Numerics;
+using Microsoft.PowerFx.Core.App.ErrorContainers;
 using Microsoft.PowerFx.Core.Binding;
 using Microsoft.PowerFx.Core.Functions.Delegation;
 using Microsoft.PowerFx.Core.Functions.Delegation.DelegationMetadata;
@@ -62,7 +63,7 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
             return true;
         }
 
-        protected bool IsValidDelegatableFilterPredicateNode(TexlNode dsNode, TexlBinding binding, FilterOpMetadata filterMetadata)
+        protected bool IsValidDelegatableFilterPredicateNode(TexlNode dsNode, TexlBinding binding, FilterOpMetadata filterMetadata, bool generateHints = true)
         {
             Contracts.AssertValue(dsNode);
             Contracts.AssertValue(binding);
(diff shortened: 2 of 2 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -121,7 +121,7 @@ internal sealed partial class TexlBinding
 
         public bool HasLocalScopeReferences { get; private set; }
 
-        public ErrorContainer ErrorContainer { get; } = new ErrorContainer();
+        public ErrorContainer ErrorContainer { get; set; } = new ErrorContainer();
```
MikeStall, line 124: @bryancroteau-MSFT  - fyi , adding setter

pakempar (approved):

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/FilterDelegationBase.cs
```diff
@@ -75,78 +76,97 @@ protected bool IsValidDelegatableFilterPredicateNode(TexlNode dsNode, TexlBindin
...
-                case NodeKind.BinaryOp:
-                    {
-                        var opNode = dsNode.AsBinaryOp();
-                        var binaryOpNodeValidationStrategy = GetOpDelegationStrategy(opNode.Op, opNode);
-                        Contracts.AssertValue(opNode);
+                // if hints should not be generated, create a temporary error container
+                if (!generateHints)
+                {
+                    originalErrorContainer = binding.ErrorContainer;
+                    binding.ErrorContainer = new ErrorContainer();
```
Shpakh-MSFT, line 86: How this would be used? I am little concern about making errorcontainer property mutable.

ian-legler: The scenario is that we want to allow delegation validation in the short term that does not result in warnings shown to the end-user, purely for telemetry.  The current validation code directly adds warnings to the ErrorContainer on the binder.  A previous PR (https://github.com/microsoft/Power-Fx/pull/914) changed the logic to limit emitting warnings, but it was a more significant code change that was fragile.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
bryancroteau-MSFT: Shouldn't this be set in the constructor?

ian-legler: @bryancroteau-MSFT Not sure if there's a reason to prefer one pattern over the other, I wanted to limit this PR to purely changes needed for my functionality (and testing)

lesaltzm, line 124: Instead of making this mutable, couldn't the error container instead be passed as an arg to the delegation strategy checkers, so that it puts errors there rather than using the binder's error container? I agree with the concerns others have raised here

Shpakh-MSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
ian-legler: The issue is that the delegation strategy checkers delegate their functionality to other methods other objects (e.g. IsRowScopedServerDelegatable), which would also need to take an error container instead of using the one attached to the binder.  This would introduce additional fragility.

bryancroteau-MSFT: Is it possible for multiple methods or objects to modify the mutable property? If so, how can you guarantee that the error container behavior returns to the way it was before it was modified? For example, is there ever a case where there should be a stack of error containers instead of a mutable property?

ian-legler: I've replaced the mutable property with a specific override method, to attempt to limit misuse.  I've also added a unit test to cover nested scenarios.

LucGenetier:
✅ No public API change.

lesaltzm (approved):
signing off on this, but if we see this pattern showing up more, we should revisit this

LucGenetier:
✅ No public API change.

PR #963, opened 2023-01-05
Check in latest localization

toshio-msft:


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #965, opened 2023-01-05
Ensure that Date + Number operations have a Date result

CarlosFigueiraMSFT:
We made a change a few months ago to make all operations with dates and number to return a DateTime value. This way, something like `Date(2022, 1, 2) + 2.5` would return `DateTime(2022, 1, 4, 12, 0, 0, 0)`. This is not the typical behavior in Excel, however, so we are changing it back to have those operations return a Date value.

In most applications it won't matter, but if we use the value in the Text function (without specifying a format) we will see a difference. For example, `Text(Date(2022, 1, 2) + 2.5))` will return `"1/6/2022"` (assuming en-US default format, m/d/yyyy).

Notice that although the static type is Date, the internal representation will have the time part and no information will be lost, as shown in the following expressions:

- `Text(Date(2022, 1, 2) + 2.5, "yyyy-mm-dd hh:mm:ss")` will return `"2022-01-04 12:00:00"`
- `Text(Date(2022, 1, 2) + 2.5 + 0.5)` will return `"1/5/2022"` (assuming en-US default format, m/d/yyyy)


LucGenetier:
✅ No public API change.

MikeStall (dismissed):

Shpakh-MSFT (dismissed):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

Shpakh-MSFT (approved):
:shipit:

PR #968, opened 2023-01-06
Improve Benchmark

LucGenetier:
Improve benchmark project
Add reference tests
Add script to run locally easily
Add script to read results quickly
Add script to write results to SQL database


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Performance.Tests/Reference.cs
```diff
@@ -0,0 +1,60 @@
...
+        public int FixedDuration()
+        {
+            // Fixed duration, independant of CPU speed 
+            Thread.Sleep(100);
+            return 0;
+        }
+
+        [Benchmark]
+        [MethodImpl(MethodImplOptions.NoOptimization | MethodImplOptions.NoInlining)]
+        public long Loop()
```
MikeStall, line 49: Can we rename to "FixedLoop".

And add a comment in this class that the goal for all these tests is that they should take constant time (independent of any Power Fx changes), and so act as a baseline reference for performance.

src/Directory.Build.props
```diff
@@ -73,7 +73,7 @@
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(Project)'!='Microsoft.PowerFx.Performance.Tests'">
+  <PropertyGroup Condition=" '$(ProjectName)' != 'Microsoft.PowerFx.Performance.Tests' ">
```
MikeStall, line 76: Curious: what is project vs. projectname?

MikeStall (approved):

MikeStall (dismissed):

src/Directory.Build.props
LucGenetier: ProjectName is documented
https://learn.microsoft.com/en-us/cpp/build/reference/common-macros-for-build-commands-and-properties?redirectedfrom=MSDN&view=msvc-170

src/tests/Microsoft.PowerFx.Performance.Tests/Reference.cs
LucGenetier: Done

MikeStall (approved):

LucGenetier:
✅ No public API change.

Issue #971, opened 2023-01-06
More robust implementation of Substitute


MikeStall:

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/1119

nguhoa:
Fixed in PR https://github.com/microsoft/Power-Fx/pull/1129

PR #972, opened 2023-01-07
Fix 730 - better protection for expressions that generate large memory usage.

### LINKED ISSUE #730: Add memory constraint
There should be a way to configure an upper bound of memory usage in Power FX engine/expressions, to protect the host from reaching their memory capacity and potential DoS attacks.

MikeStall:
Fix #730 - better protection for expressions that generate large memory usage. 

Add "governor" service that can abort on large memory usage. Host can implement this and use GC functions to check memory usage and abort if an expression consumes too much memory.    Substitute() function has an inefficient implementation which provides one way to quickly consume memory (see #971 )
This coexists with other robustness checks like cancellation token and stack overflow. 

Functions can also "pre-poll" to calculate their memory usage and fail before they start executing. 

Refactored similar tests into a single location, "SandboxTests"

Add #setup so that it can be disabled by tests which intentionally use large memory.

We'll still need to review the rest of the code and enforce the checks are all rpesent.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Environment/Governor.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Environment/Governor.cs
@@ -22,17 +22,25 @@ namespace Microsoft.PowerFx
 {
     /// <summary>
     /// Throttling object to allow a host to control limits of memory and hardware resources.
+    /// These virtuals are called by the evaluator, and exceptions thrown by them will abort the evaluation. 
+    /// This is a service passed to <see cref="RuntimeConfig"/>, and a derived class can override and throw their own host exceptions if hardware limits are exceeded.
     /// </summary>
     public class Governor
     {
-        // Called by evaluator when allocating large memory blocks 
-        // to ensure there's storage to proceed. 
+        /// <summary>
+        /// Called by evaluator before allocating large memory blocks 
+        /// to ensure there's storage to proceed. 
+        /// This allows failure before we start to eval a memory intensive operation. 
+        /// </summary>
+        /// <param name="allocateBytes">Predicted number of bytes this function may need in order to execute. </param>
         public virtual void PollMemory(long allocateBytes)
         {
         }
 
-        // Called throughout engine. This can monitor system resources (like memory)
-        // and throw an exception to abort execution. 
+        /// <summary>
+        /// Called throughout engine. This can monitor system resources (like memory) 
+        /// and throw an exception to abort execution. 
+        /// </summary>
         public virtual void Poll()
         {
         }
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Environment/Governor.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Environment/Governor.cs
@@ -1,0 +1,40 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.Contracts;
+using System.Globalization;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.PowerFx.Core.Entities;
+using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Core.IR.Nodes;
+using Microsoft.PowerFx.Core.IR.Symbols;
+using Microsoft.PowerFx.Core.Texl.Builtins;
+using Microsoft.PowerFx.Functions;
+using Microsoft.PowerFx.Interpreter;
+using Microsoft.PowerFx.Types;
+using static Microsoft.PowerFx.Functions.Library;
+
+namespace Microsoft.PowerFx
+{
+    /// <summary>
+    /// Throttling object to allow a host to control limits of memory and hardware resources.
+    /// </summary>
+    public class Governor
+    {
+        // Called by evaluator when allocating large memory blocks 
+        // to ensure there's storage to proceed. 
+        public virtual void PollMemory(long allocateBytes)
+        {
+        }
+
+        // Called throughout engine. This can monitor system resources (like memory)
+        // and throw an exception to abort execution. 
+        public virtual void Poll()
+        {
+        }
+    }
+}
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Environment/Governor.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Environment/Governor.cs
@@ -33,8 +33,15 @@ namespace Microsoft.PowerFx
         /// This allows failure before we start to eval a memory intensive operation. 
         /// </summary>
         /// <param name="allocateBytes">Predicted number of bytes this function may need in order to execute. </param>
-        public virtual void PollMemory(long allocateBytes)
+        public virtual void CanAllocateBytes(long allocateBytes)
         {
+        }
+
+        public virtual void CanAllocateString(long stringLength)
+        {
+            // .Net strings are an array of characters.
+            var allocateBytes = stringLength * sizeof(char);
+            this.CanAllocateBytes(allocateBytes);
         }
 
         /// <summary>
--- a/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
@@ -38,6 +38,8 @@ namespace Microsoft.PowerFx
 
         public CultureInfo CultureInfo { get; private set; }
 
+        public Governor Governor { get; private set; }
+        
         public EvalVisitor(IRuntimeConfig config, CancellationToken cancellationToken)
         {
             _symbolValues = config.Values; // may be null 
@@ -45,6 +47,8 @@ namespace Microsoft.PowerFx
 
             _services = config.ServiceProvider ?? new BasicServiceProvider();
 
+            Governor = GetService<Governor>() ?? new Governor();
+            
             CultureInfo = GetService<CultureInfo>();
         }
 
@@ -69,6 +73,8 @@ namespace Microsoft.PowerFx
         {
             // Throws OperationCanceledException exception
             _cancellationToken.ThrowIfCancellationRequested();
+
+            Governor.Poll();
         }
 
         // Helper to eval an arg that might be a lambda.
@@ -232,38 +238,40 @@ namespace Microsoft.PowerFx
 
             var childContext = context.SymbolContext.WithScope(node.Scope);
 
+            FormulaValue result;
             if (func is IAsyncTexlFunction asyncFunc)
             {
-                var result = await asyncFunc.InvokeAsync(args, _cancellationToken);
-                return result;
+                result = await asyncFunc.InvokeAsync(args, _cancellationToken);
             }
             else if (func is UserDefinedTexlFunction udtf)
             {
                 // $$$ Should add _runtimeConfig
-                var result = await udtf.InvokeAsync(args, _cancellationToken, context.StackDepthCounter.Increment());
-                return result;
+                result = await udtf.InvokeAsync(args, _cancellationToken, context.StackDepthCounter.Increment());
             }
             else if (func is CustomTexlFunction customTexlFunc)
             {
-                var result = await customTexlFunc.InvokeAsync(FunctionServices, args, _cancellationToken);
-                return result;
+                // If custom function throws an exception, don't catch it - let it propagate up to the host.
+                result = await customTexlFunc.InvokeAsync(FunctionServices, args, _cancellationToken);
             }
             else
             {
                 if (FunctionImplementations.TryGetValue(func, out var ptr))
                 {
-                    var result = await ptr(this, context.IncrementStackDepthCounter(childContext), node.IRContext, args);
+                    result = await ptr(this, context.IncrementStackDepthCounter(childContext), node.IRContext, args);
 
                     if (IfFunction.CanCheckIfReturn(func))
                     {
                         Contract.Assert(result.IRContext.ResultType == node.IRContext.ResultType || result is ErrorValue || result.IRContext.ResultType is BlankType);
                     }
-
-                    return result;
                 }
-
-                return CommonErrors.NotYetImplementedError(node.IRContext, $"Missing func: {func.Name}");
+                else
+                {
+                    result = CommonErrors.NotYetImplementedError(node.IRContext, $"Missing func: {func.Name}");
+                }
             }
+
+            CheckCancel();
+            return result;
         }
 
         public override async ValueTask<FormulaValue> Visit(BinaryOpNode node, EvalVisitorContext context)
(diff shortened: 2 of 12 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -703,12 +698,55 @@ private static FormulaValue Substitute(IRContext irContext, FormulaValue[] args)
                 instanceNum = (int)nv.Value;
             }
 
+            // Compute max possible memory this operation may need.
```
MikeStall, line 701: @bryancroteau-MSFT  - We should apply these sort of pre-checks to our other string,table functions.

src/libraries/Microsoft.PowerFx.Interpreter/Environment/Governor.cs
```diff
@@ -0,0 +1,40 @@
...
+using Microsoft.PowerFx.Interpreter;
+using Microsoft.PowerFx.Types;
+using static Microsoft.PowerFx.Functions.Library;
+
+namespace Microsoft.PowerFx
+{
+    /// <summary>
+    /// Throttling object to allow a host to control limits of memory and hardware resources.
+    /// </summary>
+    public class Governor
```
MikeStall, line 28: This is the critical new service.

src/tests/Microsoft.PowerFx.Interpreter.Tests/Helpers/BasicGovernor.cs
```diff
@@ -0,0 +1,75 @@
...
+    // Host also calls eval and controls what's caught. 
+    internal class GovernorException : Exception
+    {
+        public GovernorException(string msg)
+            : base(msg)
+        {
+        }
+    }
+
+    internal class BasicGovernor : Governor
```
MikeStall, line 20: This is a suggested default implementaiton.
Note that GC.GetAllocatedBytesForCurrentThread() requires NetStandard2.1, but Interpreter uses Net2.0.  Hence this helper class needs to live in the host.

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -302,6 +302,16 @@ protected override async Task<RunResult> RunAsyncInternal(string expr, string se
...
                 }
 
+                // Ensure tests can run with governor on. 
+                // Some tests that use large memory can disable via:
+                //    #SETUP: DisableMemChecks
+                if (!iSetup.DisableMemChecks)
+                {
+                    var kbytes = 1000;
+                    var mem = new BasicGovernor(10 * 1000 * kbytes);
+                    runtimeConfig.AddService<Governor>(mem);
```
MikeStall, line 312: Ensure that all our .txts run with this on. 
A few tests do Sequence(100000) and that blows the limit, hence the #setup to disable it.

src/tests/Microsoft.PowerFx.Interpreter.Tests/SandboxTests.cs
```diff
@@ -0,0 +1,174 @@
...
+using Microsoft.PowerFx.Core.Functions;
+using Microsoft.PowerFx.Core.Localization;
+using Microsoft.PowerFx.Core.Tests;
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Types;
+using Xunit;
+using static Microsoft.PowerFx.Core.Localization.TexlStrings;
+
+namespace Microsoft.PowerFx.Tests
```
MikeStall, line 18: Grouping related "sandbox" tests into a single place - this doubles as documentation for how hosts can protect against various abuses.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT:
> #SETUP: DisableMemChecks

What happens to tests with this setup step in other runtimes (such as when running with Power Apps)? Will they be skipped, or still run?

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/MinMax.txt:1 in 1359766. [](commit_id = 1359766d72505c11c3993153bacf5ed64c3f2e6d, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -703,12 +698,55 @@ private static FormulaValue Substitute(IRContext irContext, FormulaValue[] args)
...
+            return result; 
+        }
+
+        private static StringValue SubstituteWorker(EvalVisitor eval, StringValue source, StringValue match, StringValue replacement, int instanceNum)
+        {
+            if (string.IsNullOrEmpty(match.Value))
+            {
+                return source;
+            }
+
```
CarlosFigueiraMSFT, line 741: Why is it moved inside the worker? No new memory would be allocated in this case.

MikeStall: Every call to SubString() is an alloc!
This whole method needs to be rewritten  - #971.

src/tests/Microsoft.PowerFx.Interpreter.Tests/Helpers/BasicGovernor.cs
MikeStall, line 20: >BasicGovernor [](http://example.com/codeflow?start=19&length=13)

rename SingleThreadedGovernor - to emphasize it's tracking per thread.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -230,40 +236,42 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
...
+                        Contract.Assert(result.IRContext.ResultType == node.IRContext.ResultType || result is ErrorValue || result.IRContext.ResultType is BlankType);
+                    }
+                }
+                else
+                {
+                    result = CommonErrors.NotYetImplementedError(node.IRContext, $"Missing func: {func.Name}");
+                }
+            }
+
+            CheckCancel();
```
Shpakh-MSFT, line 273: Why add here just before returning the result?

MikeStall: We're returning from this function, but we're still in the middle of an execution. 
This also checks for memory usage - so checking after a function called keeps us close to the original point of failure and a small window.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -746,7 +786,7 @@ private static FormulaValue Substitute(IRContext irContext, FormulaValue[] args)
                 }
             }
 
-            return new StringValue(irContext, sourceValue);
+            return FormulaValue.New(sourceValue);
```
CarlosFigueiraMSFT, line 789: >return FormulaValue.New(sourceValue); [](http://example.com/codeflow?start=12&length=37)

We should keep the irContext in the values that we use - this may help if this new value is used in an expression that generates an error

MikeStall: The IR source spans are wrong and we can't meaningfully use them.

MikeStall:
> #SETUP: DisableMemChecks

there is an update to PAClient's JS Runner to ignore this #setup and run these tests. 

---
In reply to: [1375054099](https://github.com/microsoft/Power-Fx/pull/972#issuecomment-1375054099) [](http://example.com/codeflow?ancestors=1375054099,1375039652)

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/MinMax.txt:1 in 1359766. [](commit_id = 1359766d72505c11c3993153bacf5ed64c3f2e6d, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
MikeStall: Re-added to stay consistent. We can remove later.

src/libraries/Microsoft.PowerFx.Interpreter/Environment/Governor.cs
```diff
@@ -0,0 +1,48 @@
...
+using static Microsoft.PowerFx.Functions.Library;
+
+namespace Microsoft.PowerFx
+{
+    /// <summary>
+    /// Throttling object to allow a host to control limits of memory and hardware resources.
+    /// These virtuals are called by the evaluator, and exceptions thrown by them will abort the evaluation. 
+    /// This is a service passed to <see cref="RuntimeConfig"/>, and a derived class can override and throw their own host exceptions if hardware limits are exceeded.
+    /// </summary>
+    public class Governor
```
LucGenetier, line 28: >Governor [](http://example.com/codeflow?start=17&length=8)

We want to define if this class will work for Sync and/or Async Power Fx APIs

When a governor is used, we should validate if it's compatible with the API that is being used.

LucGenetier, line 28: >Governor [](http://example.com/codeflow?start=17&length=8)

I would use a more precise name: MemoryGovernor

src/tests/Microsoft.PowerFx.Core.Tests/TestRunnerTests/InternalSetup.cs
```diff
@@ -18,6 +18,8 @@ internal class InternalSetup
 
         internal TimeZoneInfo TimeZoneInfo { get; set; }
 
+        internal bool DisableMemChecks { get; set; }
```
LucGenetier, line 21: DisableMemoryChecks

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
LucGenetier: The name of this function isn't related to what it's really doing then
+ Shouldn't memory allocations be checked before allocating?
+ Checking for cancellation token at this point is not the right place.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -703,12 +698,55 @@ private static FormulaValue Substitute(IRContext irContext, FormulaValue[] args)
...
+                // Replace all instances. 
+                // Maximum possible length of Substitute, convert all the Match to Replacement. 
+                // Unicode, so 2B per character.
+                if (matchLen == 0)
+                {
+                    maxLenChars = sourceLen;
+                }
+                else
+                {
+                    maxLenChars = sourceLen / matchLen * replacementLen;
```
LucGenetier, line 718: >matchLen [](http://example.com/codeflow?start=46&length=8)

if matchLen > sourceLan, this becomes 0

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -703,12 +698,55 @@ private static FormulaValue Substitute(IRContext irContext, FormulaValue[] args)
...
+            var sourceLen = source.Value.Length;
+            var matchLen = match.Value.Length;
+            var replacementLen = replacement.Value.Length;
+                        
+            int maxLenChars;
+            if (instanceNum < 0)
+            {
+                // Replace all instances. 
+                // Maximum possible length of Substitute, convert all the Match to Replacement. 
+                // Unicode, so 2B per character.
```
LucGenetier, line 711: >Unicode, so 2B per character. [](http://example.com/codeflow?start=19&length=29)

Could be 3 byte long

src/libraries/Microsoft.PowerFx.Interpreter/Environment/Governor.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        /// <summary>
+        /// Called by evaluator before allocating large memory blocks 
+        /// to ensure there's storage to proceed. 
+        /// This allows failure before we start to eval a memory intensive operation. 
+        /// </summary>
+        /// <param name="allocateBytes">Predicted number of bytes this function may need in order to execute. </param>
+        public virtual void PollMemory(long allocateBytes)
```
LucGenetier, line 36: >PollMemory [](http://example.com/codeflow?start=28&length=10)

DeclareMemoryAllocation?

LucGenetier:
> #SETUP: DisableMemChecks

Why do we need this?

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Lookup.txt:1 in 57c14bb. [](commit_id = 57c14bb96c4a150c35e4865e44c0a53ee0abf495, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
MikeStall: True - it depends on the encoding. 

.Net uses a 2-byte unicode encoding:
[https://stackoverflow.com/questions/3787826/internal-representation-of-strings-in-c-sharp](https://stackoverflow.com/questions/3787826/internal-representation-of-strings-in-c-sharp)

Although ultimately, this is just an estimate - I can update the comments.

MikeStall:
> >> LookUp(Sequence(30000), Value > 25000, Value * 2)

We need the #setup above because these large sequences consume a lot of memory - because sequence operator is creating an in-memory table. 
These should be optimized and then we can remove the setup.


---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Lookup.txt:64 in 57c14bb. [](commit_id = 57c14bb96c4a150c35e4865e44c0a53ee0abf495, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
MikeStall: I'll use sizeof(char)

src/libraries/Microsoft.PowerFx.Interpreter/Environment/Governor.cs
MikeStall: This might expand to handle other resource checks.

MikeStall:
> #SETUP: DisableMemChecks

see comment below

---
In reply to: [1381458311](https://github.com/microsoft/Power-Fx/pull/972#issuecomment-1381458311) [](http://example.com/codeflow?ancestors=1381458311,1381458311)

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Lookup.txt:1 in 57c14bb. [](commit_id = 57c14bb96c4a150c35e4865e44c0a53ee0abf495, deletion_comment = False)

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
MikeStall: We check both before and after. 
If a function managed to allocate and go over the limit, this is an opportunity to catch it immediately.

LucGenetier (approved):

LucGenetier:
✅ No public API change.

PR #991, opened 2023-01-12
Fix display name bug for Single Column Table references

lesaltzm:
Single Column Table expressions (dotted names, with the lhs that's a table), are not supported in the interpreter. In Canvas, they have a subtle bug where the display name set for the resulting type is set to the field type rather than the base table's. This is incorrect, and leads to bugs when the base table was not associated with any child record/tables.


LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #992, opened 2023-01-15
Batch file for refreshing local nugets

MikeStall:
New batch file - buildAndRefresh.bat  , which will:
- clear the machine locale nuget cache for 0.3.0 so that the newly built local nugets will get picked up. (not an issue for PAClient since that has its own private nuget cache)
- call existing batch file to build nugets.
- print some instructions for how to consume them.


LucGenetier:
✅ No public API change.

LucGenetier (approved):

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

Issue #749, opened 2022-10-11
Add the ability to restrict certain data types from being used


susosuso:
Certain data types are not applicable to our scenarios -- for examples, `Color`. If possible, we'd like to be able to block off the ability to use those types in Power FX expressions. Perhaps exposing the settings in PowerFxConfig?

anderson-joyle:
Just for curiosity, is there a scenario where you expect a certain data type but ends up getting an unwanted one (like `Color`)?

MikeStall:
Some options:
1. You can walk the tree (after check) and reject "bad" types, using a visitor on 
2. you can restrict incoming Functions and Enums to block generating types. 

To be airtight, this would have to be done after the "deferred types" are substituted - else a deferred type make introduce one of the banned types. 

Fx can consider something like an overload on Engine that's called by Check() to ensure this walker gets invoked. That would cover both check and intellisense scenarios.

MikeStall:
Fixed with #993

Issue #990, opened 2023-01-11
Remove CultureInfo from PowerFxConfig


MikeStall:
Removing the config from PowerFxConfig. 

See #918, and discussion in #872 .

We have 3 uses of culture:
- Parser mode
- errors 
- runtime culture

These can often be different values. 
Setting this once in config may seem convenient, but it becomes a problem when we need to set different values. 

we should avoid a global setting on config, and just rely on the caller being explicit. 

Related - we should never be getting this from thread's current culture. Again, require the caller to be explicit.

MikeStall:
Different usages for culture: Parser /  Runtime / Error
All can be set independently. 
	Parser - via ParserOptions 
	Error
	Runtime - via ServiceProvider

Config.Culture is a problem because which culture does it refer to?

May need to be updated:
https://github.com/microsoft/Power-Fx/discussions/872

See ParseAndErrorLocaleAreDifferent test. 

These have culture via ParserOptions:
	Engine
	CheckResult 
	
So any instance method on these already has culture and doesn't need a new param. 

TBD - IntellisenseData?

	
CheckWrapper is UDF stuff; just ignore it.

MikeStall:
See #1292

PR #993, opened 2023-01-15
CheckResult rewrite, add hooks for derived engine like SQL.

MikeStall:
[1] Add hooks so that SqlEngine can be a real engine derived from PowerFx.Engine.  The benefits are:
- eliminates duplicated code around display name conversion and intellisense. 
- enabled SqlEngine to leverage CodeFixHandlers. 

[2] Also cleanup CheckResult. Better enforcement of invariants. Also lets us avoid duplicate or unnecessary work; such  as getting intellisense competition suggestions (just need the _binding) without having to compute dependency info (TopLevelIdentifiers). 

[3] Enforce "Text too long" as a parser options.  SqlEngine used this.
Before, this check occurred after binding. It should occur immediately before we even try to lex. 

[4] Leverage the new work on CheckResult to remove some Obsolete and some non-sensical APIs. For example, we shouldn't take a ParseResult and ParseOptions, since the Result already has the options. 

### New hooks on Engine
This required some new hooks for derived engines:
- ability to set rule scope to binder (SqlEngine used this to represent the current record)
- Ability for an engine to consistently apply per-engine constraints (also requested in #749). Sql engine has many restrictions (due to generating as SQL) beyond just which functions are allowed.  Add a Engine.PostCheck() hook which gets invoked and allows engines to provide additional errors.
- better granularity around CheckResult (see below). We need to coordinate when the SQL visitor walks , and avoid double walking. 

### Major refactor on CheckResult to be more granular.
Before - the class had organically grown and was unclear when certain fields got filled out. 
we have many different "stages" and each stage requires certain work.  But we don't want to repeat and duplicate the work.
For example: 

- Parsing requires input for text and ParserOptions. 
- Binding requires input for symbol info. And prereq is parsing. 
- IR has prereq of Parsing, and Binding. 
- Final Error pass required Parsing and Binding errors; possible invoking Engine's custom hook; possible IR.  
- Several other features require binding, but not errors or IR, such as: Intellisense Suggestions, Dependency Analysis, ToInvariant, ToDisplay, 

So the list is already at 7+ kinds of work, and I expect this list to grow. 

Specific engines also have requirements:
- Interpreter (RecalcEngine) requires an IR for evaluation. 
- SQL engine requires final error pass, which requires a full IR tree traversal.  But again, we don't want to do that on Intellisense, etc. 

Now, for CheckResult, the design is:
- all CheckResult are tied back to an Engine and can use the engine to compute work. 
- input is first provided by SetXX() functions, notably SetText() and SetBindingInfo(). Splitting into separate Set() functions also avoids the combinatorial explosion of parameters we were getting on Engine. 
- ApplyXXX() functions then compute (based on their engine) and save results on CheckResult. They nop if already computed. These fail if Set() wasn't called.
- properties return results computed by the apply function.   these fail if Apply() wasn't first called.


LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.CheckResult.#ctor<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.get_Expression<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.get_Symbols<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.set_Errors(System.Collections.Generic.IEnumerable{Microsoft.PowerFx.ExpressionError})<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.set_Expression(Microsoft.PowerFx.IExpression)<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.set_Parse(Microsoft.PowerFx.ParseResult)<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.set_Symbols(Microsoft.PowerFx.ReadOnlySymbolTable)<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.set_TopLevelIdentifiers(System.Collections.Generic.HashSet{System.String})<br />❌CP0002 M:Microsoft.PowerFx.Engine.Check(Microsoft.PowerFx.ParseResult,Microsoft.PowerFx.ParserOptions,Microsoft.PowerFx.ReadOnlySymbolTable)<br />❌CP0002 M:Microsoft.PowerFx.Engine.CreateEvaluator(Microsoft.PowerFx.CheckResult)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.CheckResult.#ctor<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.get_Expression<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.get_Symbols<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.set_Expression(Microsoft.PowerFx.IExpression)<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.set_Parse(Microsoft.PowerFx.ParseResult)<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.set_Symbols(Microsoft.PowerFx.ReadOnlySymbolTable)<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.set_TopLevelIdentifiers(System.Collections.Generic.HashSet{System.String})<br />❌CP0002 M:Microsoft.PowerFx.Engine.Check(Microsoft.PowerFx.ParseResult,Microsoft.PowerFx.ParserOptions,Microsoft.PowerFx.ReadOnlySymbolTable)<br />❌CP0002 M:Microsoft.PowerFx.Engine.CreateEvaluator(Microsoft.PowerFx.CheckResult)

LucGenetier:
>         public static ErrorResourceKey ErrColonExpected = new ErrorResourceKey("ErrColonExpected");

We should alpha-sort this long list at some point

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Localization/Strings.cs:559 in 90e97e1. [](commit_id = 90e97e1db5fc1772e235f9d13df961dd29e3d180, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Parser/ParseResult.cs
```diff
@@ -46,6 +46,9 @@ public class ParseResult : IOperationStatus
 
         internal SourceList After { get; }
 
+        // Options used ot create these results.
+        public ParserOptions Options { get; internal set; }
```
LucGenetier, line 50: >Options [](http://example.com/codeflow?start=29&length=7)

ParResult is an out paramaters
Why would in parameters be there?

src/libraries/Microsoft.PowerFx.Core/Parser/ParseResult.cs
```diff
@@ -55,6 +58,32 @@ public class ParseResult : IOperationStatus
...
+
+            var errKey = Core.Localization.TexlStrings.ErrTextTooLarge;
+            var err = new TexlError(tok, DocumentErrorSeverity.Critical, errKey, maxAllowed, text.Length);
+
+            List<TexlError> errors = new List<TexlError>()
+            {
+                err
+            };
+
+            int id = 0;
```
LucGenetier, line 80: >int id = 0; [](http://example.com/codeflow?start=12&length=11)

I think you can remove this line and use "ref int id" in next line

src/libraries/Microsoft.PowerFx.Core/Parser/ParseResult.cs
```diff
@@ -55,6 +58,32 @@ public class ParseResult : IOperationStatus
...
+            TexlNode root = new ErrorNode(ref id, tok, err.ShortMessage);
+            List<CommentToken> comments = new List<CommentToken>();
```
LucGenetier, line 82: >new List<CommentToken> [](http://example.com/codeflow?start=42&length=22)

I know you don't like it but you can remove that part for brievty

src/libraries/Microsoft.PowerFx.Core/Parser/ParserOptions.cs
```diff
@@ -20,18 +20,39 @@ public class ParserOptions
         /// </summary>
         public bool AllowsSideEffects { get; set; }
 
+        /// <summary>
+        /// The culture that an expression will parse with. 
+        /// This primarily determines numeric decimal separator character
+        /// as well as chaining operator.
```
LucGenetier, line 26: any impact on date formatting at this stage?

src/libraries/Microsoft.PowerFx.Core/Parser/ParserOptions.cs
```diff
@@ -20,18 +20,39 @@ public class ParserOptions
...
+        public int MaxExpressionLength { get; set; }
+
         internal ParseResult Parse(string script)
         {
             return Parse(script, Features.None);
         }
 
         internal ParseResult Parse(string script, Features features)
         {
+            if (MaxExpressionLength > 0 && script.Length > MaxExpressionLength)
```
LucGenetier, line 43: >MaxExpressionLength > 0 [](http://example.com/codeflow?start=16&length=23)

This test should be in the setter

src/libraries/Microsoft.PowerFx.Core/Parser/ParserOptions.cs
```diff
@@ -20,18 +20,39 @@ public class ParserOptions
...
+        /// This primarily determines numeric decimal separator character
+        /// as well as chaining operator. 
+        /// </summary>
         public CultureInfo Culture { get; set; }
 
+        /// <summary>
+        /// If greater than 0, enforces a maximum length on a single expression. 
+        /// It is an immediate parser error if the expression is too long. 
+        /// </summary>
+        public int MaxExpressionLength { get; set; }
```
LucGenetier, line 34: >set [](http://example.com/codeflow?start=46&length=3)

test the incoming value here

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.CheckResult.#ctor<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.get_Expression<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.get_Symbols<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.set_Expression(Microsoft.PowerFx.IExpression)<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.set_Parse(Microsoft.PowerFx.ParseResult)<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.set_Symbols(Microsoft.PowerFx.ReadOnlySymbolTable)<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.set_TopLevelIdentifiers(System.Collections.Generic.HashSet{System.String})<br />❌CP0002 M:Microsoft.PowerFx.Engine.Check(Microsoft.PowerFx.ParseResult,Microsoft.PowerFx.ParserOptions,Microsoft.PowerFx.ReadOnlySymbolTable)<br />❌CP0002 M:Microsoft.PowerFx.Engine.CreateEvaluator(Microsoft.PowerFx.CheckResult)

src/libraries/Microsoft.PowerFx.Core/Parser/ParserOptions.cs
```diff
@@ -20,18 +20,39 @@ public class ParserOptions
...
+                var result2 = ParseResult.ErrorTooLarge(script, MaxExpressionLength);
+                result2.Options = this;
+                return result2;
+            }
+
             var flags = AllowsSideEffects ? TexlParser.Flags.EnableExpressionChaining : TexlParser.Flags.None;
 
-            return TexlParser.ParseScript(script, features, Culture, flags);
+            var result = TexlParser.ParseScript(script, features, Culture, flags);
+            result.Options = this;
```
LucGenetier, line 54: >result.Options = this; [](http://example.com/codeflow?start=12&length=22)

Why are we including the options (input param) in the results?

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -17,115 +19,413 @@
...
+        private bool _setBindingCalled;
+        internal ReadOnlySymbolTable _symbols; // can be null
+        private VersionHash _symbolHash; // hash of _symbols at time of assignment
+
+        #endregion
+
+        [Obsolete("Use public constructor")]
+        internal CheckResult()
+        {
+        }
```
LucGenetier, line 52: Is it time to remove this one?

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -17,115 +19,413 @@
...
+        /// The source engine this was created against. 
+        /// This is critical for calling back to populate the rest of the results. 
         /// </summary>
-        public FormulaType ReturnType { get; set; }
+        private readonly Engine _engine;
+       
+        // The raw expression test. 
+        private string _expression;
+        
+        private ParserOptions _parserOptions;
```
LucGenetier, line 40: >private ParserOptions _parserOptions; [](http://example.com/codeflow?start=8&length=37)

Do we NEED this input parameter in the results?

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -17,115 +19,413 @@
...
 
         /// <summary>
-        /// Names of fields that this formula uses. 
-        /// null if unavailable.  
-        /// This is only valid when <see cref="IsSuccess"/> is true.
+        /// Initializes a new instance of the <see cref="CheckResult"/> class.
         /// </summary>
-        public HashSet<string> TopLevelIdentifiers { get; set; }
+        /// <param name="source">Engine used to handle Apply operations.</param>
+        public CheckResult(Engine source)
```
LucGenetier, line 58: >CheckResult [](http://example.com/codeflow?start=15&length=11)

Should we initialize _errors here?

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -17,115 +19,413 @@
...
         /// <summary>
-        /// List of errors and warnings. Check <see cref="ExpressionError.IsWarning"/>.
-        /// Not null, but empty on success.
+        /// Initializes a new instance of the <see cref="CheckResult"/> class.
+        /// Create a "failed" check around a set of errors. 
+        /// Can't do any other operations. 
         /// </summary>
-        public IEnumerable<ExpressionError> Errors { get; set; }
+        /// <param name="extraErrors"></param>
+        public CheckResult(IEnumerable<ExpressionError> extraErrors)
```
LucGenetier, line 71: >CheckResult [](http://example.com/codeflow?start=15&length=11)

Should we initialize _engine here?

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -17,115 +19,413 @@
...
+                throw new ArgumentNullException(nameof(parse));    
+            }
+
+            if (_expression != null)
+            {
+                throw new InvalidOperationException($"Can only call {nameof(SetText)} once.");
+            }
+
+            _expression = parse.Text;
+            _parserOptions = parse.Options;
```
LucGenetier, line 91: Do we need? (thiking at memory)

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -17,115 +19,413 @@
...
+        public CheckResult SetText(ParseResult parse)
+        {
+            if (parse == null)
+            {
+                throw new ArgumentNullException(nameof(parse));    
+            }
+
+            if (_expression != null)
+            {
+                throw new InvalidOperationException($"Can only call {nameof(SetText)} once.");
```
LucGenetier, line 87: >$"Can only call {nameof(SetText)} once." [](http://example.com/codeflow?start=52&length=40)

This message should come from resources

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -17,115 +19,413 @@
...
+        /// or any custom errors passes explicit to the ctor.
+        /// Not null, but empty on success.
         /// </summary>
-        public ParseResult Parse { get; set; }
+        public IEnumerable<ExpressionError> Errors
+        {
+            get => _errors;
+
+            [Obsolete("use constructor to set errors")]
+            set => _errors.AddRange(value);
```
LucGenetier, line 230: Can we remove it?

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -17,115 +19,413 @@
...
+
+        /// <summary>
+        /// Helper to throw if <see cref="IsSuccess"/> is false.
+        /// </summary>
+        public void ThrowOnErrors()
+        {
+            if (!IsSuccess)
+            {
+                var msg = string.Join("\r\n", Errors.Select(err => err.ToString()).ToArray());
+                throw new InvalidOperationException($"Errors: " + msg);
```
LucGenetier, line 246: >$"Errors: " + msg [](http://example.com/codeflow?start=52&length=17)

Should come from resources, like all other exception messages

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -17,115 +19,413 @@
...
+        /// May be null. 
+        /// </summary>
+        internal ReadOnlySymbolTable AllSymbols { get; private set; }
 
         /// <summary>
         /// Parameters are the subset of symbols that must be passed in Eval() for each evaluation. 
         /// This lets us associated the type in Check()  with the values in Eval().
         /// </summary>
-        internal ReadOnlySymbolTable Parameters { get; set; }
+        internal ReadOnlySymbolTable Parameters => _symbols;
```
LucGenetier, line 263: >Parameters => _symbols [](http://example.com/codeflow?start=37&length=22)

We need to align terms/variable names to avoid confusions.
This should be "_parameters"

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -17,115 +19,413 @@
...
         /// Culture info passed to this binding. May be null. 
         /// </summary>
-        internal CultureInfo CultureInfo { get; set; }
+        internal CultureInfo CultureInfo => this.Source.Config.CultureInfo;
+
+        internal void ThrowIfSymbolsChanged()
+        {
+            if (_symbols != null)
+            {
+                var endHash = _symbols.VersionHash;
```
LucGenetier, line 274: >endHash [](http://example.com/codeflow?start=20&length=7)

useless variable

LucGenetier: we can merge the two 'if' conditions for clarity

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -17,115 +19,413 @@
...
+            if (_binding == null)
+            {
+                if (!this._setBindingCalled)
+                {
+                    throw new InvalidOperationException($"Must call {nameof(SetBindingInfo)} before calling {nameof(ApplyBinding)}.");
+                }
+
+                (var binding, var combinedSymbols) = Source.ComputeBinding(this);
+
+                this.ThrowIfSymbolsChanged();
```
LucGenetier, line 333: >this [](http://example.com/codeflow?start=16&length=4)

no need, creates confusion (is there a base.Throw... ?)

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -17,115 +19,413 @@
...
-        /// Names of fields that this formula uses. 
-        /// null if unavailable.  
-        /// This is only valid when <see cref="IsSuccess"/> is true.
+        /// Initializes a new instance of the <see cref="CheckResult"/> class.
         /// </summary>
-        public HashSet<string> TopLevelIdentifiers { get; set; }
+        /// <param name="source">Engine used to handle Apply operations.</param>
+        public CheckResult(Engine source)
+        {
+            this._engine = source ?? throw new ArgumentNullException(nameof(source));
```
LucGenetier, line 60: >this [](http://example.com/codeflow?start=12&length=4)

This is creating confusion & not necessary.
When I see "this.xyz", I imagine there is a "base.xyz" somewhere and we need to distinguish the two with this prefix...

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -17,115 +19,413 @@
...
+
+                this.ThrowIfSymbolsChanged();
+                
+                // Don't modify any fields until after we've verified the symbols haven't change.
+
+                this._binding = binding;
+                this.AllSymbols = combinedSymbols;
+
+                // Add the errors
+                var bindingErrors = ExpressionError.New(binding.ErrorContainer.GetErrors(), CultureInfo);
```
LucGenetier, line 341: >var [](http://example.com/codeflow?start=16&length=3)

I'd clarify the type, knowning ExpressionError.New is not a constructor

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -17,115 +19,413 @@
...
+                _errors.AddRange(bindingErrors);
+
+                if (this.IsSuccess && !this.HasDeferredArgsWarning)
+                {
+                    // TODO: Fix FormulaType.Build to not throw exceptions for Enum types then remove this check
+                    if (binding.ResultType.Kind != DKind.Enum)
```
LucGenetier, line 347: Can we remove this check now?

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -17,115 +19,413 @@
...
 
-        internal CheckResult(ParseResult parse, CultureInfo locale, TexlBinding binding = null)
+        /// <summary>
+        /// Compute the dependencies. Called after binding. 
+        /// </summary>
+        public void ApplyDependencyAnalysis()
         {
-            Parse = parse ?? throw new ArgumentNullException(nameof(parse));
-            CultureInfo = locale;
+            var binding = this.Binding; // will throw if binding wasn't run
```
LucGenetier, line 362: >will throw if binding wasn't run [](http://example.com/codeflow?start=43&length=32)

Is the asignment really going to throw?

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -17,115 +19,413 @@
...
+        // Flag to ensure Post Checks are only invoked once. 
+        private bool _invokingPostCheck;
 
-            Errors = Parse.Errors.Concat(BindingErrors);
+        /// <summary>
+        /// Calculate all errors. 
+        /// Invoke Binding and any engine-specific errors via <see cref="Engine.PostCheck(CheckResult)"/>. 
+        /// </summary>
+        /// <returns></returns>
+        public IEnumerable<ExpressionError> ApplyErrors()
```
LucGenetier, line 374: >ApplyErrors [](http://example.com/codeflow?start=44&length=11)

Shouldn't this be names "GetErrors()" instead?

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -17,115 +19,413 @@
...
+                var extraErrors = Source.InvokePostCheck(this);
+
+                _errors.AddRange(extraErrors);
+            }
+            
+            return this.Errors;
         }
 
-        public void ThrowOnErrors()
+        internal IRResult ApplyIR()
```
LucGenetier, line 393: >ApplyIR [](http://example.com/codeflow?start=26&length=7)

GetIR() ?

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
LucGenetier: And maybe Errors { get [ ...

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -28,6 +28,14 @@ public class Engine : IPowerFxEngine
         /// </summary>
         public PowerFxConfig Config { get; }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Engine"/> class.
+        /// </summary>
+        public Engine()
+            : this(new PowerFxConfig())
+        {
+        }
```
LucGenetier, line 37: Couldn't we change the new constructor with

public Engine(PowerFxConfig powerFxConfig = null)
{
    Config = powerFxConfig ?? new PowerFxConfig();
}

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -108,6 +116,16 @@ private protected virtual IBinderGlue CreateBinderGlue()
             return new Glue2DocumentBinderGlue();
         }
 
+        public virtual ParserOptions GetDefaultParserOptionsCopy()
```
LucGenetier, line 119: >GetDefaultParserOptionsCopy [](http://example.com/codeflow?start=37&length=27)

CloneParserOptions() ?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -124,6 +142,7 @@ public IReadOnlyList<Token> Tokenize(string expressionText)
         /// <returns></returns>
         public ParseResult Parse(string expressionText, ParserOptions options = null)
         {
+            options ??= this.GetDefaultParserOptionsCopy();
```
LucGenetier, line 145: >this [](http://example.com/codeflow?start=24&length=4)

no need

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -124,6 +142,7 @@ public IReadOnlyList<Token> Tokenize(string expressionText)
         /// <returns></returns>
         public ParseResult Parse(string expressionText, ParserOptions options = null)
```
LucGenetier, line 143: >Parse [](http://example.com/codeflow?start=27&length=5)

we should test
- expressionText is not null
- expressionText.Length

LucGenetier, line 143: >Parse [](http://example.com/codeflow?start=27&length=5)

Same tests here (not null expression, length)

LucGenetier:
>         public static ParseResult Parse(string expressionText, ParserOptions options = null, CultureInfo cultureInfo = null)

Same tests here (not null expression, length)

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs:156 in 90e97e1. [](commit_id = 90e97e1db5fc1772e235f9d13df961dd29e3d180, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -144,13 +163,18 @@ public static ParseResult Parse(string expressionText, ParserOptions options = n
         /// </summary>
         public static ParseResult Parse(string expressionText, Features features, ParserOptions options = null, CultureInfo cultureInfo = null)
```
LucGenetier, line 164: >Parse [](http://example.com/codeflow?start=34&length=5)

test expression length

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -144,13 +163,18 @@ public static ParseResult Parse(string expressionText, ParserOptions options = n
...
+            }
+
             options ??= new ParserOptions();
 
             // If culture isn't explicitly set, use the one from PowerFx Config
             options.Culture ??= cultureInfo;
 
             var result = options.Parse(expressionText, features);
-            return result;
+            return result;
```
LucGenetier, line 177: >             [](http://example.com/codeflow?start=26&length=12)

CTRL-E, D

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -162,8 +186,12 @@ public static ParseResult Parse(string expressionText, Features features, Parser
         /// <returns></returns>
         public CheckResult Check(string expressionText, RecordType parameterType, ParserOptions options = null)
```
LucGenetier, line 187: >expressionText [](http://example.com/codeflow?start=40&length=14)

Public API, we should test expressionText is not null, below max length

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -173,92 +201,115 @@ public CheckResult Check(string expressionText, RecordType parameterType, Parser
...
+            return Check(parse, parameterType);
+        }
+
+        [Obsolete("Use other check overload. Shouldn't need both ParserOptions and ParseResult.")]
+        internal CheckResult Check(ParseResult parse, ParserOptions options)
+        {
+            parse.Options = options;
+            return Check(parse, null, options);
+        }
+
```
LucGenetier, line 216: We should remove these APIs now

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -173,92 +201,115 @@ public CheckResult Check(string expressionText, RecordType parameterType, Parser
...
+        public CheckResult Check(ParseResult parse, RecordType parameterType = null)
```
LucGenetier, line 217: >Check [](http://example.com/codeflow?start=27&length=5)

We should test parse is not null (or add a comment it's allowed)

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -173,92 +201,115 @@ public CheckResult Check(string expressionText, RecordType parameterType, Parser
...
+            var check = new CheckResult(this)
+               .SetText(parse)
+               .SetBindingInfo(parameterType);
 
-            return CheckInternal(parse, bindingConfig, symbols);
+            CheckWorker(check);
+            return check;
         }
 
         public CheckResult Check(
```
LucGenetier, line 227: >Check [](http://example.com/codeflow?start=27&length=5)

We need to enable CA1062 for GA

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -173,92 +201,115 @@ public CheckResult Check(string expressionText, RecordType parameterType, Parser
...
+            check.ApplyBindingInternal();
+            check.ApplyErrors();
+            check.ApplyDependencyAnalysis();
+        }
 
-            return CheckInternal(parse, bindingConfig, symbolTable);
+        // Called after check result, can inject additional errors or constraints. 
+        protected virtual IEnumerable<ExpressionError> PostCheck(CheckResult check)
+        {
+            return new ExpressionError[0];
```
LucGenetier, line 252: >new ExpressionError[0] [](http://example.com/codeflow?start=19&length=22)

Use Enumerable.Empty<ExpressionError>()

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -173,92 +201,115 @@ public CheckResult Check(string expressionText, RecordType parameterType, Parser
...
+        // Prefer to avoid this hook and use SymbolTables instead. 
+        private protected virtual RecordType GetRuleScope()
+        {
+            return null;
+        }
+
+        private BindingConfig GetDefaultBindingConfig()
+        {
+            var ruleScope = this.GetRuleScope();
+            bool useThisRecordForRuleScope = ruleScope != null;
```
LucGenetier, line 270: >useThisRecordForRuleScope [](http://example.com/codeflow?start=17&length=25)

I don't see a reason for creating a variable which is only used once.
Logically, I'm wondering if BindingConfig.Default going to modify ruleScope?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -173,92 +201,115 @@ public CheckResult Check(string expressionText, RecordType parameterType, Parser
...
+            return bindingConfig;
+        }
+
+        // Called by CheckResult.ApplyBinding to compute the binding. 
+        internal (TexlBinding, ReadOnlySymbolTable) ComputeBinding(CheckResult result)
         {
-            var startHash = symbolTable?.VersionHash;
+            var parse = result.ApplyParse();
+
+            ReadOnlySymbolTable symbolTable = result.Parameters;
```
LucGenetier, line 287: >symbolTable [](http://example.com/codeflow?start=32&length=11)

should be named parameters

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -173,92 +201,115 @@ public CheckResult Check(string expressionText, RecordType parameterType, Parser
...
             var resolver = CreateResolverInternal(out var combinedSymbols, symbolTable);
-            
+
             var glue = CreateBinderGlue();
 
+            var ruleScope = this.GetRuleScope();
+
+            // Canvas apps uses rule scope for lots of cases. 
+            // But in general, we should only use rule scope for 'ThisRecord' binding. 
+            // Anything else should be accomplished with SymbolTables.
```
LucGenetier, line 299: Shall we work with them to get this changed?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -281,11 +332,20 @@ public IIntellisenseResult Suggest(string expression, RecordType parameterType,
         /// </summary>
         public IIntellisenseResult Suggest(CheckResult checkResult, int cursorPosition)
```
LucGenetier, line 333: >checkResult [](http://example.com/codeflow?start=55&length=11)

We should check this is not null

LucGenetier: and have a test

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -294,7 +354,7 @@ public IIntellisenseResult Suggest(CheckResult checkResult, int cursorPosition)
         [Obsolete("Use overload without expression")]
         public IIntellisenseResult Suggest(string expression, CheckResult checkResult, int cursorPosition)
```
LucGenetier, line 355: >expression [](http://example.com/codeflow?start=50&length=10)

test (null + length) + checkResult + cursorPosition in range

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -342,8 +402,11 @@ public RenameDriver CreateFieldRenamer(RecordType parameters, DPath pathToRename
         /// as part of the RecordType.</param>
         /// <returns>The formula, with all identifiers converted to invariant form.</returns>
         public string GetInvariantExpression(string expressionText, RecordType parameters)
```
LucGenetier, line 404: >expressionText [](http://example.com/codeflow?start=52&length=14)

test input parameters

LucGenetier:
>         public bool ValidateInvocation(string functionName, IReadOnlyList<TexlNode> args, out FormulaType retType)

should be tested

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/IntellisenseOperations.cs:41 in 90e97e1. [](commit_id = 90e97e1db5fc1772e235f9d13df961dd29e3d180, deletion_comment = False)

LucGenetier:
>         public bool ValidateInvocation(string functionName, IReadOnlyList<TexlNode> args, out FormulaType retType)

and we should have a better name

---
In reply to: [1383725432](https://github.com/microsoft/Power-Fx/pull/993#issuecomment-1383725432) [](http://example.com/codeflow?ancestors=1383725432)

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/IntellisenseOperations.cs:41 in 90e97e1. [](commit_id = 90e97e1db5fc1772e235f9d13df961dd29e3d180, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Public/IntellisenseOperations.cs
```diff
@@ -47,20 +47,22 @@ public bool ValidateInvocation(string functionName, IReadOnlyList<TexlNode> args
                 throw new ArgumentNullException(nameof(args));
             }
 
+            var binding = _checkResult.Binding;
+
             foreach ((var arg, var index) in args.Select((value, index) => (value, index)))
             {
                 if (arg == null)
                 {
                     throw new ArgumentNullException(nameof(args), $"Argument {index} is null");
```
LucGenetier, line 56: >$"Argument {index} is null" [](http://example.com/codeflow?start=66&length=27)

in resources

src/libraries/Microsoft.PowerFx.Core/Public/IntellisenseOperations.cs
```diff
@@ -47,20 +47,22 @@ public bool ValidateInvocation(string functionName, IReadOnlyList<TexlNode> args
...
 
-                if (!_checkResult._binding.IsNodeValid(arg))
+                if (!binding.IsNodeValid(arg))
                 {
                     throw new ArgumentException($"Argument {index} does not belong to this result");
                 }
             }
 
-            var types = args.Select(node => _checkResult._binding.GetType(node)).ToArray();
+            var types = args.Select(node => binding.GetType(node)).ToArray();
```
LucGenetier, line 65: >var [](http://example.com/codeflow?start=12&length=3)

I'd specify the type

LucGenetier:
>         public static void Dump(CheckResult check, TextWriter tw, string indent = "")

Should add an exclusion for CA1303

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/DebugDump.cs:139 in 90e97e1. [](commit_id = 90e97e1db5fc1772e235f9d13df961dd29e3d180, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/ParsedExpression.cs
```diff
@@ -69,24 +69,19 @@ public static IExpressionEvaluator GetEvaluator(this CheckResult result)
 
         internal static IExpressionEvaluator GetEvaluator(this CheckResult result, StackDepthCounter stackMarker)
```
LucGenetier, line 70: >result [](http://example.com/codeflow?start=75&length=6)

Should check this isn't null + stackMarker

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -182,7 +163,7 @@ public async Task<FormulaValue> EvalAsync(string expressionText, CancellationTok
             var check = Check(expressionText, options, symbolsAll);
             check.ThrowOnErrors();
 
-            check.Parameters = parameterSymbols;
+            // check.Parameters = parameterSymbols;
             var stackMarker = new StackDepthCounter(Config.MaxCallDepth);
```
LucGenetier, line 167: remove

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -59,36 +59,17 @@ internal INameResolver TestCreateResolver()
...
         /// <param name = "result" >A successful binding from a previous call to.<see cref="Engine.Check(string, RecordType, ParserOptions)"/>. </param>        
         /// <returns></returns>
+        [Obsolete("Call CheckResult.GetEvaluator()")]
         public static IExpression CreateEvaluatorDirect(CheckResult result)
         {
-            return CreateEvaluatorDirect(result, new StackDepthCounter(PowerFxConfig.DefaultMaxCallDepth));
+            var eval = result.GetEvaluator();
+            var eval2 = (ParsedExpression)eval;
+            return eval2;
         }
```
LucGenetier, line 73: Can we remove it?

LucGenetier:
>                 throw new System.Exception("Expression is not a ParsedExpression");

We shouldn't throw just an Exception type + use resources

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/UDF/UserDefinedTexlFunction.cs:56 in 90e97e1. [](commit_id = 90e97e1db5fc1772e235f9d13df961dd29e3d180, deletion_comment = False)

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Public/EditorContextScope.cs
```diff
@@ -72,12 +72,29 @@ string IPowerFxScope.ConvertToDisplay(string expression)
...
+            var check = new CheckResult(this._engine)
+                .SetText(expression, _parserOptions)
+                .SetBindingInfo(_symbols);
+
             return _engine.Suggest(check, cursorPosition);
         }
 
         #endregion
 
+        public void AddQuickFixHandlers(params CodeFixHandler[] codeFixHandlers)
```
LucGenetier, line 85: >codeFixHandlers [](http://example.com/codeflow?start=64&length=15)

we should check none of the params is null

src/tests/Microsoft.PowerFx.Core.Tests/CheckResultTests.cs
```diff
@@ -0,0 +1,277 @@
...
+using Microsoft.PowerFx.Syntax;
+using Microsoft.PowerFx.Types;
+using Xunit;
+
+namespace Microsoft.PowerFx.Core.Tests
+{
+    public class CheckResultTests
+    {
+        [Fact]
+        public void Ctors()
```
LucGenetier, line 16: >Ctors [](http://example.com/codeflow?start=20&length=5)

We should prefix all tests with CheckResultTests or any equivalent or we end up with dozens of "Ctors" in test explorer

src/tests/Microsoft.PowerFx.Core.Tests/CheckResultTests.cs
```diff
@@ -0,0 +1,277 @@
...
+        {
+            // CheckResult must have an engine so that it can invoke operations.
+            Assert.Throws<ArgumentNullException>(() => new CheckResult((Engine)null));
+        }
+
+        [Fact]
+        public void Errors()
+        {
+            Assert.Throws<ArgumentNullException>(() => new CheckResult((IEnumerable<ExpressionError>)null));
+
```
LucGenetier, line 26: nit: shouldn't this one be in Ctors() ?

src/tests/Microsoft.PowerFx.Core.Tests/CheckResultTests.cs
```diff
@@ -0,0 +1,277 @@
...
+        }
+
+        // Ensure we can pass in ParserOptions. 
+        [Fact]
+        public void ParserOptions()
+        {
+            var check = new CheckResult(new Engine());
+            var opts = new ParserOptions
+            {
+                Culture = new System.Globalization.CultureInfo("fr-FR")
```
LucGenetier, line 167: >"fr-FR" [](http://example.com/codeflow?start=63&length=7)

Merci :)

src/tests/Microsoft.PowerFx.Core.Tests/CheckResultTests.cs
```diff
@@ -0,0 +1,277 @@
...
+            };
+            check.SetText("1,234", opts); // , is decimal separate for fr-FR.
```
LucGenetier, line 169: >separate [](http://example.com/codeflow?start=58&length=8)

separator

src/tests/Microsoft.PowerFx.Core.Tests/CheckResultTests.cs
```diff
@@ -0,0 +1,277 @@
...
+        public void TestIR()
+        {
+            var check = new CheckResult(new Engine());
+
+            Assert.Throws<InvalidOperationException>(() => check.ApplyIR());
+            check.SetText("1+2");
+            check.SetBindingInfo();
+
+            var ir = check.ApplyIR();
+            Assert.NotNull(ir);
```
LucGenetier, line 258: >ir [](http://example.com/codeflow?start=27&length=2)

Can we test ir content?

src/tests/Microsoft.PowerFx.Core.Tests/CheckResultTests.cs
```diff
@@ -0,0 +1,277 @@
...
+            var check = new CheckResult(new Engine());
+
+            Assert.Throws<InvalidOperationException>(() => check.ApplyIR());
+            check.SetText("1+x");
+            check.SetBindingInfo();
+
+            check.ApplyBinding();
+            Assert.False(check.IsSuccess);
+
+            Assert.Throws<InvalidOperationException>(() => check.ApplyIR());
```
LucGenetier, line 274: >Throws [](http://example.com/codeflow?start=19&length=6)

Can we check the message?

src/tests/Microsoft.PowerFx.Core.Tests/CheckResultTests.cs
```diff
@@ -0,0 +1,277 @@
...
+
+namespace Microsoft.PowerFx.Core.Tests
+{
+    public class CheckResultTests
+    {
+        [Fact]
+        public void Ctors()
+        {
+            // CheckResult must have an engine so that it can invoke operations.
+            Assert.Throws<ArgumentNullException>(() => new CheckResult((Engine)null));
```
LucGenetier, line 19: >Throws [](http://example.com/codeflow?start=19&length=6)

Can we check the message as well? (applies to all Assert.Throws)

src/tests/Microsoft.PowerFx.Core.Tests/DerivedEngineTests.cs
```diff
@@ -0,0 +1,192 @@
...
+using Xunit;
+using Xunit.Sdk;
+
+namespace Microsoft.PowerFx.Tests
+{
+    // Test various hooks deriving from Engine
+    public class DerivedEngineTests
+    {
+        [Fact]
+        public void Test()
```
LucGenetier, line 27: >Test [](http://example.com/codeflow?start=20&length=4)

Not allowed :)

LucGenetier: Prefix all tests with DerivedEngine_...

src/tests/Microsoft.PowerFx.Core.Tests/DerivedEngineTests.cs
```diff
@@ -0,0 +1,192 @@
...
+        {
+            Engine engineNormal = new Engine(new PowerFxConfig());
+            Engine engine = new MyEngine();
+
+            var result = engine.Check("1+2");
+            Assert.True(result.IsSuccess);
+
+            // PostCheck hook 
+            var expr = "1+\"abc\"";
```
LucGenetier, line 36: >"1+\"abc\"" [](http://example.com/codeflow?start=23&length=11)

(nit) Personally, I prefer @"1+""abc"""

src/tests/Microsoft.PowerFx.Core.Tests/DerivedEngineTests.cs
```diff
@@ -0,0 +1,192 @@
...
+            result = engineNormal.Check(expr);
+            Assert.True(result.IsSuccess);
+
+            result = engine.Check(expr);
+            Assert.False(result.IsSuccess);
+
+            var errors = result.Errors.ToArray();
+            Assert.Single(errors);
+            Assert.Contains("Test: can't have string litera", errors[0].Message);
```
LucGenetier, line 45: >litera [](http://example.com/codeflow?start=53&length=6)

literal ?

src/strings/PowerFxResources.en-US.resx
```diff
@@ -1837,6 +1837,10 @@
     <value>Numeric value is too large.</value>
     <comment>Error Message.</comment>
   </data>
+  <data name="ErrTextTooLarge" xml:space="preserve">
+    <value>Expression can't be more than {0} characters. The expression is {1} characters.</value>
```
LucGenetier, line 1841: >is [](http://example.com/codeflow?start=72&length=2)

has?
or "is {1} character long." ?

src/libraries/Microsoft.PowerFx.Core/Parser/ParserOptions.cs
LucGenetier, line 34: >MaxExpressionLength [](http://example.com/codeflow?start=19&length=19)

Add a default to 64k
Yes, this is breaking but we want it.

src/tests/Microsoft.PowerFx.Core.Tests/DerivedEngineTests.cs
```diff
@@ -0,0 +1,192 @@
...
+                public CustomResolver(ReadOnlySymbolTable functions)
+                    : base(functions)
+                {
+                }
+
+                public override bool Lookup(DName name, out NameLookupInfo nameInfo, NameLookupPreferences preferences = NameLookupPreferences.None)
+                {
+                    if (name.Value == "extra")
+                    {
+                        nameInfo = new NameLookupInfo(BindKind.PowerFxResolvedObject,   DType.Number, DPath.Root, 0);
```
LucGenetier, line 183: >    [](http://example.com/codeflow?start=85&length=3)

nit, too many spaces

src/libraries/Microsoft.PowerFx.Core/Parser/ParserOptions.cs
LucGenetier: Also add a test that this value is not 0 by default

src/libraries/Microsoft.PowerFx.Core/Parser/ParseResult.cs
MikeStall: ?

src/libraries/Microsoft.PowerFx.Core/Parser/ParserOptions.cs
MikeStall: This shouldn't have any behavior change to existing behavior.   (Although we will want to tackle that separately for our outstanding locale bugs like #990)

src/libraries/Microsoft.PowerFx.Core/Parser/ParserOptions.cs
MikeStall: >0 is just the "is enabled" check. 
It's whatever the host sets. If a host is running on a large dedicated VM   and wants to allow 100k strings, then they can configure to do that.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
MikeStall: Added for PAClient compat to avoid an internal breaking change. Kept internal to mitigate use. Can remove after PAClient sweeps to remove Obsoletes. .

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
MikeStall: it's initialized by the field initializer. (that runs as part of the ctor)

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
MikeStall: We don't need to. 
It just uses the Engine do do other operations. This is creating an "failed" check which can't do any further operations.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
MikeStall: It means 'field on this class' as opposed to something else (like static). This is #966.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
MikeStall: Uses should never see this string - None of our internal exceptions are from resources yet - so I'm just following that existing pattern.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
MikeStall: I wish - also needed for PAClient.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
MikeStall: This is a style thing. 
I prefer having the explicit locals because it's easier to spot things under a debugger and I can set breakpoints in the middle.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
MikeStall: I'm not sure - this is just existing code being moved; and I'm trying to preserve behavior with the refactor.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
MikeStall: The naming pattern here is Apply*() for all these operations. (ApplyParse,ApplyBinding,ApplyDependency,etc)
It's a method , not a property, because it does significant explicit work (and can be side-effecting).

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
MikeStall: This is just one path through the parser. We check it at a lower level in the  stack (in CheckWorker) to gaurantee we catch all paths.

src/libraries/Microsoft.PowerFx.Core/Parser/ParseResult.cs
MikeStall: only for 'out'. 
ref needs to be defined before usage.

src/libraries/Microsoft.PowerFx.Core/Parser/ParserOptions.cs
MikeStall: I have to check it against the text field. So I can't do in an individual setter - must be done at apply operation.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
MikeStall: Yes.  Because some things take just the ParseResults and may want to look back at the options it was parsed against.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
MikeStall: what's the memory issue? holding a reference to an existing (already allocated) object.  And Options is tiny (not the expression)

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -17,115 +19,413 @@
...
+        internal TexlBinding ApplyBindingInternal()
         {
+            if (_binding == null)
+            {
+                if (!this._setBindingCalled)
+                {
+                    throw new InvalidOperationException($"Must call {nameof(SetBindingInfo)} before calling {nameof(ApplyBinding)}.");
+                }
+
+                (var binding, var combinedSymbols) = Source.ComputeBinding(this);
```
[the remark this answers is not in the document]
MikeStall: alternative?

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
MikeStall: See naming convention in PR. 
I wanted to be clear this method is doing significant work and can be side-effecting - where Get/property accessor don't.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
MikeStall: It's a breaking change. 
I want to be able to find-all-ref to see which ctor was being used, ie, make sure this is a compile-time decision and not a runtime. 
Likewise, it's an error to pass a null config.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
MikeStall: If (options == null) { 
  options = this.GetDefaultParserOptionsCopy();
}
We don't want to pass null into line 146.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
MikeStall: There are many parse overloads - we test them at the common point below to make sure we catch all cases.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
MikeStall: Eventually - but keeping them since this is "just" a refactor and I want to avoid PAClient breaking changes.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
MikeStall: SetText will catch it.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
MikeStall: It's a style thing - it makes the code more self-documenting.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
MikeStall: Ideally, but it will be very challenging to get them to prioritize it. 
It's well outside of scope for this PR.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
MikeStall: We have 100s of methods missing these checks. Will handle separately.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
MikeStall: existing code. out of scope.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
MikeStall: PAClient

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Public/EditorContextScope.cs
```diff
@@ -72,12 +72,29 @@ string IPowerFxScope.ConvertToDisplay(string expression)
 
         IIntellisenseResult IPowerFxScope.Suggest(string expression, int cursorPosition)
```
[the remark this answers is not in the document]
MikeStall: Explicit interface Method doesn't allow private. 
[https://stackoverflow.com/questions/1392252/c-sharp-what-is-the-reason-for-explicitly-implemented-interface-members-are-a](https://stackoverflow.com/questions/1392252/c-sharp-what-is-the-reason-for-explicitly-implemented-interface-members-are-a)

src/tests/Microsoft.PowerFx.Core.Tests/CheckResultTests.cs
MikeStall: our other Throws test don't do this.

src/tests/Microsoft.PowerFx.Core.Tests/DerivedEngineTests.cs
MikeStall: I thought style cop would have caught

MikeStall:
>         public static void Dump(CheckResult check, TextWriter tw, string indent = "")

I'm not seeing the warning. 

---
In reply to: [1383742181](https://github.com/microsoft/Power-Fx/pull/993#issuecomment-1383742181) [](http://example.com/codeflow?ancestors=1383742181,1383742181)

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/DebugDump.cs:139 in 90e97e1. [](commit_id = 90e97e1db5fc1772e235f9d13df961dd29e3d180, deletion_comment = False)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.CheckResult.#ctor<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.get_Expression<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.get_Symbols<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.set_Expression(Microsoft.PowerFx.IExpression)<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.set_Parse(Microsoft.PowerFx.ParseResult)<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.set_Symbols(Microsoft.PowerFx.ReadOnlySymbolTable)<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.set_TopLevelIdentifiers(System.Collections.Generic.HashSet{System.String})<br />❌CP0002 M:Microsoft.PowerFx.Engine.Check(Microsoft.PowerFx.ParseResult,Microsoft.PowerFx.ParserOptions,Microsoft.PowerFx.ReadOnlySymbolTable)<br />❌CP0002 M:Microsoft.PowerFx.Engine.CreateEvaluator(Microsoft.PowerFx.CheckResult)

src/libraries/Microsoft.PowerFx.Core/Parser/ParseResult.cs
LucGenetier: We call Parse with ParseOptions and this returns a ParseResult.
Why should we include this input param in the output/result ?

LucGenetier (approved):
:shipit:

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.CheckResult.#ctor<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.get_Expression<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.get_Symbols<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.set_Expression(Microsoft.PowerFx.IExpression)<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.set_Parse(Microsoft.PowerFx.ParseResult)<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.set_Symbols(Microsoft.PowerFx.ReadOnlySymbolTable)<br />❌CP0002 M:Microsoft.PowerFx.CheckResult.set_TopLevelIdentifiers(System.Collections.Generic.HashSet{System.String})<br />❌CP0002 M:Microsoft.PowerFx.Engine.Check(Microsoft.PowerFx.ParseResult,Microsoft.PowerFx.ParserOptions,Microsoft.PowerFx.ReadOnlySymbolTable)<br />❌CP0002 M:Microsoft.PowerFx.Engine.CreateEvaluator(Microsoft.PowerFx.CheckResult)

PR #998, opened 2023-01-18
Added error for named formulas with no value

VamsiModem:
Named formulas without value was not showing any errors, this is fixed in this PR


LucGenetier:
✅ No public API change.

lesaltzm (approved):

PR #1008, opened 2023-01-20
Enforce parenthesis on reference to function property

anhtruong-MSFT:
Currently when a function property has an optional parameter and that parameter is omitted at reference, there is no error because we allow it. 
This change is to make sure that we throw an invalid reference error when this happens because we want function calls to end with parenthesis with or without parameter going forward.

Note: the added feature gate must be removed once ECS flag is completely rolled out.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/App/IExternalEnabledFeatures.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/App/IExternalEnabledFeatures.cs
@@ -20,6 +20,8 @@ namespace Microsoft.PowerFx.Core.App
         bool IsUseDisplayNameMetadataEnabled { get; }
 
         bool IsDynamicSchemaEnabled { get; }
+
+        bool IsEnhancedComponentFunctionProperty { get; }
     }
 
     internal sealed class DefaultEnabledFeatures : IExternalEnabledFeatures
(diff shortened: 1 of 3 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/App/IExternalEnabledFeatures.cs
```diff
@@ -20,6 +20,8 @@ internal interface IExternalEnabledFeatures
         bool IsUseDisplayNameMetadataEnabled { get; }
 
         bool IsDynamicSchemaEnabled { get; }
+
+        bool IsEnhancedComponentFunctionProperty { get; }
```
lesaltzm, line 24: naming pattern: This should be `Is ... Enabled`

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Connectors.OpenApiExtensions.IsInternal(Microsoft.OpenApi.Models.OpenApiParameter)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.OpenApiExtensions.ToFormulaType(Microsoft.OpenApi.Models.OpenApiSchema)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.OpenApiExtensions.TryGetDefaultValue(Microsoft.OpenApi.Models.OpenApiSchema,System.String@)

lesaltzm (approved):

bryancroteau-MSFT (approved):

LucGenetier:
✅ No public API change.

PR #1011, opened 2023-01-20
Add benchmarks for PVA team

LucGenetier:


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Performance.Tests/PvaPerformance.cs
```diff
@@ -0,0 +1,184 @@
...
+
+            if (!checkResult.IsSuccess)
+            {
+                throw new Exception($"{expression}\r\n{string.Join("\r\n", checkResult.Errors.Select(ee => $"{ee.Message}"))}");
+            }
+
+            return checkResult;
+        }
+
+        private static string GetRandomExpression()
```
Robulane, line 97: Tbh, I think that this method will take more time than Check itself that we want to benchmark.
I think you can just hardcode a couple expressions

Robulane (approved):

LucGenetier:
✅ No public API change.

MikeStall (dismissed):

MikeStall (approved):

MikeStall (approved):

PR #1016, opened 2023-01-23
Allows to compute return type with Deferred args.

jas-valgotar:
Current Behavior is we don't attach return type to public surface and keep `checkResult..ReturnType` null.
This would allow to attach Return Type even when `Deferred` type arg is present.


LucGenetier (approved):

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

Issue #1001, opened 2023-01-18
Breaking behavior when UntypedObject is used as a return type of a custom function


susosuso:
Related to https://github.com/microsoft/Power-Fx/issues/999

We have a custom function that returns FormulaType.UntypedObject

```csharp
public class MyCustom2Function : ReflectionFunction
    {
        public MyCustom2Function() : base("MyCustom2", FormulaType.UntypedObject, FormulaType.String, FormulaType.String) { ... }

        public virtual FormulaValue Execute(StringValue arg1, StringValue arg2) { ... }
    }
```

For the untyped object, we wrote our own custom class like the following:

```csharp

    public class WrapperUntypedObject : IUntypedObject
    {
        public FormulaType ActualType { get; private set; }

        public FormulaType Type => ActualType;

        public WrapperUntypedObject(FormulaValue realFormulaValue)
        {
            ActualType = realFormulaValue.Type;
        }
       ...
    }
```

Now, with the addition of the new error handling, 
https://github.com/microsoft/Power-Fx/blob/85a2baf0a4cf732da5a9d08d933c15c505b41808/src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs#L416-L421

Our scenarios are failing because our `WrapperUntypedObject.Type` isn't `FormulaType.UntypedObject`.

Is this new behavior intentional? We'd like the original behavior if possible :)

jas-valgotar:
Hi @susosuso,

Can you elaborate on what is the execute method returning? I can't see that in the snippet.

I assume it would be returning the `UntypedObjectValue`.
And while creating that `UntypedObjectValue` you would pass in an `IRContext` instance as a 1st argument, I believe that IRContext should have `ResultType` as `UntypedObject`. If that's the case I would have to investigate it more in detail.

So, `Execute` method could be returning something similar to below snippet,
`return new UntypedObjectValue(IRContext.NotInSource(FormulaType.UntypedObject), new WrapperUntypedObject(formulaValue));`

susosuso:
We don't have access to IRContext. Instead we call `FormulaValue.New(IUntypedObject)`
https://github.com/microsoft/Power-Fx/blob/cc8c619d248e6aebb86bbee65e15bc6bee7f8e88/src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNew.cs#L125

Our `Execute` does return UntypedObjectValue as you assumed. Here's a simplified version of our function:
```csharp
[...]

Issue #999, opened 2023-01-18
Breaking error handling for custom function


susosuso:
The new version of Power FX seems to wrap error value from custom functions to another error object -- 

https://github.com/microsoft/Power-Fx/commit/85a2baf0a4cf732da5a9d08d933c15c505b41808#diff-82a475d4e4e4571d947efb9787846da7381d8cbc1446ba284b94a02307b82384

https://github.com/microsoft/Power-Fx/blob/85a2baf0a4cf732da5a9d08d933c15c505b41808/src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs#L416-L421

This breaks our own custom error handling. Take the function below as example:

```csharp
internal class MyCustomFunction : ReflectionFunction
    {
        public MyCustomFunction () : base("MyCustom", FormulaType.Number, FormulaType.String)
        { }

        public static FormulaValue Execute(StringValue stringInput)
        {
            try
            {
                // do some other validation on stringInput (like Regex or other format validation).
            }
            catch
            {
                return FormulaValue.NewError(new ExpressionError { Kind = ErrorKind.InvalidArgument, Message = "Invalid string input..." });
            }

           ...
        }
    }
```

Before the NuGet update, our custom error would return to us.
`{ Kind = ErrorKind.InvalidArgument, Message = "Invalid string input..." `

After the NuGet update, the wrapper error is returned instead. 
`{ Kind = ErrorKind.Custom, Message = "Return type should have been n, found N" }`

Can we revert this behavior, or is there a recommended way for us to properly propagate our own error handling?

jas-valgotar:
Hi @susosuso ,

Since now we are strictly checking the return type to be Identical to one provided to function (Which in NumberType here) you are seeing this error.

And it's good to see it did work as purposed and caught that Error was not created with proper type.

Can you try and add return type while creating the error to see if it resolves your issue?

I think changing Error to below would make it same as previous behavior.
`return FormulaValue.NewError(new ExpressionError { Kind = ErrorKind.InvalidArgument, Message = "Invalid string input..." }, FormulaType.Number);`

key part here is now we added the type for error! ;)

susosuso:
Thank you @jas-valgotar for the explanation! The addition makes sense, and your suggestion fixed the errors 👍

PR #1018, opened 2023-01-24
Adds error handling via Exception for Custom Function.

jas-valgotar:
Relates #1001  #999 
Before users had to loosen the return type to FormulaValue from Custom Function just to return an error, and due to that they were losing compile time strict type checking and the implementations were prone to errors of returning incorrect types.

Now instead of creating their own error, users can just throw `CustomFunctionErrorException` so they can keep the return type of `Execute` method strict and we populate the error with the correct return type.

Before they would do:
```csharp
        public class CustomFunctionError : ReflectionFunction
        {
            public CustomFunctionError()
                : base("CustomFunctionError", FormulaType.Number, FormulaType.Number) 
            {
            }

            public static FormulaValue Execute(NumberValue arg) 
            {
                if (arg.Value < 0)
                {
                    return CommonErrors.CustomError("arg should be greater than 0", FormulaType.Number);
                }
                else if (arg.Value == 0)
                {
                    throw new NotSupportedException();
                }

                return arg;
            }
        }
```
Now insted they can keep return type stricter.

```csharp
        public class CustomFunctionError : ReflectionFunction
        {
            public CustomFunctionError()
                : base("CustomFunctionError", FormulaType.Number, FormulaType.Number) 
            {
            }

            public static NumberValue Execute(NumberValue arg) 
            {
                if (arg.Value < 0)
                {
                    throw new CustomFunctionErrorException("arg should be greater than 0");
                }
                else if (arg.Value == 0)
                {
                    throw new NotSupportedException();
                }

                return arg;
            }
        }
```

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Linq.Expressions;
 using System.Numerics;
 using System.Reflection;
 using System.Threading;
@@ -17,6 +18,7 @@ using Microsoft.PowerFx.Core.Localization;
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
 using Microsoft.PowerFx.Functions;
+using Microsoft.PowerFx.Interpreter;
 using Microsoft.PowerFx.Syntax;
 using Microsoft.PowerFx.Types;
 using static Microsoft.PowerFx.Core.Localization.TexlStrings;
@@ -403,7 +405,20 @@ namespace Microsoft.PowerFx
                 args2.Add(cancellationToken);
             }
 
-            var result = _info.Method.Invoke(this, args2.ToArray());
+            object result = default;
+            try
+            {
+                result = _info.Method.Invoke(this, args2.ToArray());
+            }
+            catch (TargetInvocationException e)
+            {
+                if (e.InnerException is CustomFunctionErrorException customFunctionErrorException)
+                {
+                    return CommonErrors.CustomError(IRContext.NotInSource(_info.RetType), customFunctionErrorException.Message);
+                }
+
+                throw e;
+            }
 
             if (_info.IsAsync)
             {
@@ -421,6 +436,20 @@ namespace Microsoft.PowerFx
             }
 
             return formulaResult;
+        }
+
+        private static bool IsCompatibleType(FormulaType potentialReturnType, FormulaType returnType)
+        {
+            //if (potentialReturnType is IUntypedObject && returnType is IUntypedObject)
+            //{
+            //    return true;
+            //}
+            //else if (potentialReturnType is IUntypedObject || returnType is IUntypedObject)
+            //{
+            //    return false;
+            //}
+
+            return potentialReturnType._type == returnType._type;
         }
 
         private static async Task<FormulaValue> Unwrap(object obj, Type resultType)
--- a/src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
@@ -423,12 +423,24 @@ namespace Microsoft.PowerFx
             if (_info.IsAsync)
             {
                 var resultType = result.GetType().GenericTypeArguments[0];
-                result = await Unwrap(result, resultType);
+                try
+                {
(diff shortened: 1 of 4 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
```diff
@@ -423,6 +438,20 @@ public async Task<FormulaValue> InvokeAsync(IServiceProvider serviceProvider, Fo
             return formulaResult;
         }
 
+        private static bool IsCompatibleType(FormulaType potentialReturnType, FormulaType returnType)
```
MikeStall, line 441: >IsCompatibleType [](http://example.com/codeflow?start=28&length=16)

remove?

src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
```diff
@@ -403,7 +405,20 @@ public async Task<FormulaValue> InvokeAsync(IServiceProvider serviceProvider, Fo
                 args2.Add(cancellationToken);
             }
 
-            var result = _info.Method.Invoke(this, args2.ToArray());
+            object result = default;
+            try
+            {
+                result = _info.Method.Invoke(this, args2.ToArray());
+            }
+            catch (TargetInvocationException e)
```
MikeStall, line 413: >TargetInvocationException [](http://example.com/codeflow?start=19&length=25)

Can we double check with an async function?
(ie, there could be some AggregateException in there that needs to get unwrapped)

src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
```diff
@@ -403,7 +405,20 @@ public async Task<FormulaValue> InvokeAsync(IServiceProvider serviceProvider, Fo
                 args2.Add(cancellationToken);
             }
 
-            var result = _info.Method.Invoke(this, args2.ToArray());
+            object result = default;
+            try
+            {
+                result = _info.Method.Invoke(this, args2.ToArray());
```
MikeStall, line 411: >result [](http://example.com/codeflow?start=16&length=6)

if result is null, can we convert to Blank(type)?

jas-valgotar: updated.

src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
jas-valgotar: You are correct, async would directly throw the `CustomFunctionErrorException` while unwrapping. Updated the code for the same and added the test.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/CustomFunctions.cs
```diff
@@ -3,17 +3,132 @@
...
+            var engine = new RecalcEngine(config);
+
+            // Shows up in enumeration
+            var func = engine.GetAllFunctionNames().First(name => name == "NullFunction");
+            Assert.NotNull(func);
+
+            // Can be invoked. 
+            var result = await engine.EvalAsync("NullFunction()", CancellationToken.None);
+            Assert.IsType<BlankValue>(result);
+            Assert.IsType<NumberType>(result.IRContext.ResultType);
```
MikeStall, line 129: >result.IRContext.ResultType [](http://example.com/codeflow?start=38&length=27)

nit: result.Type
(avoid IRContext)

src/tests/Microsoft.PowerFx.Interpreter.Tests/CustomFunctions.cs
```diff
@@ -3,17 +3,132 @@
...
+            public CustomFunctionError()
+                : base("CustomFunctionError", FormulaType.Number, FormulaType.Number) 
+            {
+            }
+
+            public static NumberValue Execute(NumberValue arg) 
+            {
+                if (arg.Value < 0)
+                {
+                    throw new CustomFunctionErrorException("arg should be greater than 0");
```
MikeStall, line 31: >CustomFunctionErrorException [](http://example.com/codeflow?start=30&length=28)

Great that you're testing both paths. 
nit - we should add some comments since people will use these tests as documentation. 

Line 31:
// CustomFunctionErrorException is caught and converted to error value 

Line 35:
// All other exceptions are uncaught and will abort the eval.

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1013, opened 2023-01-20
Clean up unused Enum localization code

lesaltzm:
This is a clean up of some enum code that has not been used since PowerApps' Beta. We get a bunch of benefits besides removing debt from this, it significantly reduces the number of resources we need to load on startup, as well as shrinks the resource package size a bit

PR #1019, opened 2023-01-24
Optimize function lookup

LucGenetier:
Store functions in optimized TexlFunctionSet<TexlFunction> object instead of IEnumerable<TextFunction>

Benchmark results
- -61% for Check N=1
- -77% for Check N=5
- -79% for Check N=10
- -42% for Eval N=1
- -65% for Eval N=5
- -67% for Eval N=10

Negligible improvement for Parse , Tokenize or PVA* benchmarks


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Binding/INameResolver.cs
```diff
@@ -34,7 +34,7 @@ internal interface INameResolver
 
         DPath CurrentEntityPath { get; }
 
-        IEnumerable<TexlFunction> Functions { get; }
+        TexlFunctionSet<TexlFunction> Functions { get; }
```
MikeStall, line 37: >TexlFunctionSet [](http://example.com/codeflow?start=8&length=15)

This is the key change

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,358 @@
...
+            _enums = function.GetRequiredEnumNames().ToList();
+            _count = 1;
+        }
+
+        internal TexlFunctionSet(IEnumerable<T> functions)
+            : this()
+        {
+            if (functions == null)
+            {
+                throw new ArgumentNullException($"{nameof(functions)} cannot be null", nameof(functions));
```
MikeStall, line 49: >$"{nameof(functions)} cannot be null", [](http://example.com/codeflow?start=48&length=38)

super nit - this arg isn't necessary. ArgumentNullException will make the message from just a single arg

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,358 @@
...
+            }
+
+            foreach (var functionSet in functionSets)
+            {
+                Add(functionSet);
+            }
+        }
+
+        // Slow API, only use for backward compatibility
+        [Obsolete("Slow API. Prefer using With* functions for idenfitying functions you need.")]
```
MikeStall, line 91: >idenfitying [](http://example.com/codeflow?start=62&length=11)

nit - typo. 

+1 to making this [Obsolete]

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
LucGenetier: this is for making sure we'll get a localized message when we'll have to deal with this later

src/libraries/Microsoft.PowerFx.Core/Functions/ITexlFunction.cs
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.PowerFx.Core.Utils;
+
+namespace Microsoft.PowerFx.Core.Functions
+{
+    internal interface ITexlFunction
```
MikeStall, line 11: >ITexlFunction [](http://example.com/codeflow?start=23&length=13)

Why do we need this ?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
MikeStall: note that we still need some way to "list all functions", such as for REPL help or IDE.

src/libraries/Microsoft.PowerFx.Interpreter/DebugDump.cs
```diff
@@ -58,13 +55,15 @@ public static void Dump(ReadOnlySymbolTable symbolTable, TextWriter tw, string i
                 if (symbolTable.Functions.Any())
                 {
                     tw.WriteLine();
-                    tw.WriteLine($"{indent} Functions ({symbolTable.Functions.Count()}) total:"); 
-                    foreach (var func in symbolTable.Functions)
+                    tw.WriteLine($"{indent} Functions ({symbolTable.Functions.Count()}) total:");
+#pragma warning disable CS0618 // Type or member is obsolete
```
MikeStall, line 59: per other comment - there should be some non-obsolete way to still enumerate all functions. 
Maybe a more explicit method.

src/libraries/Microsoft.PowerFx.Core/Utils/DName.cs
```diff
@@ -143,6 +143,8 @@ public static bool IsValidDName(string strName)
             return false;
         }
 
+        // $$$ Needs optimization $$$
```
MikeStall, line 146: >// $$$ Needs optimization $$$ [](http://example.com/codeflow?start=8&length=29)

is this showing up on a profiler?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,358 @@
...
+            }
+        }
+
+        internal Dictionary<string, List<T>>.KeyCollection Keys => _functions.Keys;
+
+        internal Dictionary<string, List<T>>.KeyCollection InvariantKeys => _functionsInvariant.Keys;
+
+        internal Dictionary<DPath, List<T>>.KeyCollection NamespaceKeys => _namespaces.Keys;
+
+        internal List<T> WithName(string name) => _functions.ContainsKey(name) ? _functions[name] : new List<T>();
```
MikeStall, line 254: >List [](http://example.com/codeflow?start=17&length=4)

IEnumerable?

List is mutable.

src/tests/Microsoft.PowerFx.Interpreter.Tests/ThreadingTests.cs
```diff
@@ -16,7 +16,10 @@ public void CheckInterpreter()
         {
             var asm = typeof(RecalcEngine).Assembly;
             var bugsFieldType = new HashSet<Type>();
-            var bugNames = new HashSet<string>();
+            var bugNames = new HashSet<string>()
+            {
+                "Library.TexlFunctionImplementations"
+            };
```
MikeStall, line 22: what was the error we got?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -44,27 +43,17 @@ private static T GetService<T>(this IServiceProvider services, T defaultService)
 
         public static readonly IReadOnlyDictionary<TexlFunction, AsyncFunctionPtr> FunctionImplementations;
 
+        internal static readonly TexlFunctionSet<TexlFunctionImplementation> TexlFunctionImplementations = new TexlFunctionSet<TexlFunctionImplementation>();
```
MikeStall, line 46: >static readonly TexlFunctionSet [](http://example.com/codeflow?start=17&length=31)

threading warning because this is static mutable -

MikeStall: Get rid of TexlFunctionImplementation 
- just used locally within library.cs cctor. 


Revert Library.cs

- fixes threading warning 

that removes all usage of TexlFunctionImplementation 

which means we don't need to add interface ITexlFunction 
Which means TexlFunctionSet isn't generic

src/libraries/Microsoft.PowerFx.Core/Functions/ITexlFunction.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        DPath Namespace { get; }
+
+        string Name { get; }
+
+        string LocaleInvariantName { get; }
```
lesaltzm, line 17: Can we avoid perpetuating the "LocaleInvariantName" for functions? Just as in #1013 , the function localization code is also unused.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
lesaltzm: Do we intend to localize exception messages? That can make reviewing telemetry more complicated than needed. generally, the exception messages should be dev-facing, not user-facing.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,358 @@
...
+
+        internal List<T> WithInvariantName(string name) => _functionsInvariant.ContainsKey(name) ? _functionsInvariant[name] : new List<T>();
+
+        internal List<T> WithInvariantName(string name, DPath ns) => _functionsInvariant.ContainsKey(name) ? new List<T>(_functionsInvariant[name].Where(f => f.Namespace == ns)) : new List<T>();
+
+        internal List<T> WithNamespace(DPath ns) => _namespaces.ContainsKey(ns) ? _namespaces[ns] : new List<T>();
+
+        internal List<string> Enums => _enums;
+
+        internal TexlFunctionSet<TexlFunction> ToTexlFunctions()
```
MikeStall, line 266: >ToTexlFunctions [](http://example.com/codeflow?start=47&length=15)

This is another thing we may be able to remove if we remove the generic parameter

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,358 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.PowerFx.Core.Utils;
+
+namespace Microsoft.PowerFx.Core.Functions
+{
+    internal class TexlFunctionSet<T>
+        where T : ITexlFunction
+    {
+        private Dictionary<string, List<T>> _functions;
```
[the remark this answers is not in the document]
LucGenetier: as it's changed with Add functions, we cannot

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
LucGenetier: For telemetry we should have an ErrorId

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
LucGenetier: I'll add a FunctionNames getter and we'll see if we need more

src/libraries/Microsoft.PowerFx.Core/Functions/ITexlFunction.cs
LucGenetier: Removed

src/libraries/Microsoft.PowerFx.Core/Functions/ITexlFunction.cs
LucGenetier: Removed interface

src/libraries/Microsoft.PowerFx.Interpreter/Functions/TexlFunctionImplementation.cs
```diff
@@ -0,0 +1,38 @@
...
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+using Microsoft.PowerFx.Core.Functions;
+using Microsoft.PowerFx.Core.Utils;
+using static Microsoft.PowerFx.Functions.Library;
+
+namespace Microsoft.PowerFx.Interpreter.Functions
+{
+    internal class TexlFunctionImplementation : ITexlFunction
```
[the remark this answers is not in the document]
LucGenetier: Removed

src/libraries/Microsoft.PowerFx.Interpreter/DebugDump.cs
LucGenetier: I added a functionName getter

src/libraries/Microsoft.PowerFx.Core/Utils/DName.cs
LucGenetier: Oh yes!

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,333 @@
...
+            }
+
+            if (_count == 0)
+            {
+                _functions = new Dictionary<string, List<TexlFunction>>(functionSet._functions);
+                _functionsInvariant = new Dictionary<string, List<TexlFunction>>(functionSet._functionsInvariant);
+                _namespaces = new Dictionary<DPath, List<TexlFunction>>(functionSet._namespaces);
+                _enums = new List<string>(functionSet._enums);
+                _count = functionSet._count;
+            }
```
MikeStall, line 208: Can we just omit this block, and let the case below handle it?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,333 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.PowerFx.Core.Texl.Builtins;
+using Microsoft.PowerFx.Core.Utils;
+
+namespace Microsoft.PowerFx.Core.Functions
+{
+    internal class TexlFunctionSet
+    {
+        private Dictionary<string, List<TexlFunction>> _functions;
```
MikeStall, line 14: >string [](http://example.com/codeflow?start=27&length=6)

nit: can we comment what the key is here? 
This looks like function name to a ordered list of functions with that name.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,333 @@
...
+            else
+            {
+                _namespaces.Add(function.Namespace, new List<TexlFunction>() { function });
+            }
+
+            _enums.AddRange(function.GetRequiredEnumNames());
+
+            _count++;
+
+            return function;
```
MikeStall, line 191: >return [](http://example.com/codeflow?start=12&length=6)

could we add a Contract.Assert that count == _enums.Count == _namespaces.Count , etc

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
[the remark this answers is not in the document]
LucGenetier: if we do that: error CS0191: A readonly field cannot be assigned to (except in a constructor or init-only setter of the type in which the field is defined or a variable initializer)

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
LucGenetier: this is a valuable perf improvement vs. several loops

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
LucGenetier: Not possible due to overloads or it's going to be an expensive check

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
LucGenetier: fixed it

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -57,21 +58,29 @@ public Engine(PowerFxConfig powerFxConfig)
...
         // By default, we pull the core functions. 
         // These can be overridden. 
-        internal IEnumerable<TexlFunction> Functions => CreateResolverInternal().Functions;
+        internal TexlFunctionSet Functions => CreateResolverInternal().Functions;
 
         /// <summary>
         /// Get all functions from the config and symbol tables. 
         /// </summary>
-        public IEnumerable<FunctionInfo> FunctionInfos => Functions.Select(f => new FunctionInfo(f));
+        [Obsolete("Slow API. Use With* APIs to select functions.")]
```
MikeStall, line 66: > [Obsolete("Slow API. Use With* APIs to select functions.")] [](http://example.com/codeflow?start=7&length=60)

These can't be Obsolete. 

These are public surface functions. (With* is not). We can't obsolete them until we have a suitable replacement. 

Ability to allow a host to enumerate all functions is still a valid scenario; we just don't want them to be doing it accidentally when they don't need to.

src/libraries/Microsoft.PowerFx.Core/Public/Config/ComposedReadOnlySymbolTable.cs
```diff
@@ -60,18 +59,25 @@ public override FormulaType GetTypeFromSlot(ISymbolSlot slot)
             return slot.Owner.GetTypeFromSlot(slot);
         }
 
+        private TexlFunctionSet _nameResolverFunctions = null;
+        private HashSet<VersionHash> _nameResolverFunctionsVersions = null;
```
MikeStall, line 63: >private HashSet<VersionHash> _nameResolverFunctionsVersions = null; [](http://example.com/codeflow?start=8&length=67)

no. 
VersionHash shouldn't be used as a hash key.

MikeStall, line 63: >private HashSet<VersionHash> _nameResolverFunctionsVersions = null; [](http://example.com/codeflow?start=8&length=67)

No - can we avoid this field?  And the SetEquals on 72. 

VersionHash (line 35) already provides a checksum

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -212,14 +211,23 @@ public void RemoveFunction(string name)
...
+        internal void AddFunctions(TexlFunctionSet functions)
+        {
+            Inc();
+            _functions.Add(functions);
+
+            // Add any associated enums 
+            EnumStoreBuilder?.WithRequiredEnums(functions);
         }
 
         internal void AddFunction(TexlFunction function)
```
MikeStall, line 233: >AddFunction [](http://example.com/codeflow?start=22&length=11)

Can this just call the above? 

AddFunctions(new TexlFunctionSet(function))

src/libraries/Microsoft.PowerFx.Core/Utils/DName.cs
MikeStall: Can you file an issue to track it?  
And then include that issue in the comments (and we don't need a $$$ for it then)

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -51,7 +50,7 @@ public RecalcEngine(PowerFxConfig powerFxConfig)
         }
 
         // Set of default functions supported by the interpreter. 
-        private static readonly ReadOnlySymbolTable _interpreterSupportedFunctions = ReadOnlySymbolTable.NewDefault(Library.FunctionList);
+        private static readonly ReadOnlySymbolTable _interpreterSupportedFunctions = ReadOnlySymbolTable.NewDefault(Library.FunctionList.ToTexlFunctionSet());
```
MikeStall, line 53: >NewDefault [](http://example.com/codeflow?start=105&length=10)

Add overload:
ReadOnlySymbolTable.NewDefault<TValue>(IReadOnyDictionary<TexlFunction, TValue> keys) 

And then you can completely remove Library.FunctionList!

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Extensions.cs
```diff
@@ -28,5 +30,17 @@ internal static bool IsValid(this DateTime dateTime, EvalVisitor runner)
             // ambiguous times (like 5 Nov 2023 01:10:00 is ambiguous in PST timezone) will be considered valid
             return true;
         }
+
+        internal static TexlFunctionSet ToTexlFunctionSet(this IEnumerable<TexlFunction> functions)
```
MikeStall, line 34: >ToTexlFunctionSet [](http://example.com/codeflow?start=40&length=17)

w/ update to RecalcEngine,, then we could remove this too

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -57,21 +58,29 @@ public Engine(PowerFxConfig powerFxConfig)
...
+        public IEnumerable<FunctionInfo> FunctionInfos => Functions.Functions.Select(f => new FunctionInfo(f));
 
         /// <summary>
         /// List all functions (both builtin and custom) registered with this evaluator. 
         /// </summary>
+#pragma warning disable CA1024 // Use properties where appropriate
+        [Obsolete("Slow API. Use With* APIs to select functions.")]
```
MikeStall, line 73: >[Obsolete("Slow API. Use With* APIs to select functions.")] [](http://example.com/codeflow?start=8&length=59)

not obsolete. 
(FunctionNames isn't)

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
LucGenetier: They look very similar but are really executing very different code under the hood

LucGenetier: These two methods are very different

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Utils/DName.cs
LucGenetier: Created issue #1025

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,342 @@
...
+        // Dictionary key: function.LocaleInvariantName
+        private Dictionary<string, List<TexlFunction>> _functionsInvariant;
+
+        // Dictionary key: function.Namespace
+        private Dictionary<DPath, List<TexlFunction>> _namespaces;
+
+        // List of all function.GetRequiredEnumNames()
+        private List<string> _enums;
+
+        // Cont of functions
```
MikeStall, line 26: >Cont [](http://example.com/codeflow?start=11&length=4)

nit: typo

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,342 @@
...
+            _enums = new List<string>();
+            _count = 0;
+        }
+
+        internal TexlFunctionSet(TexlFunction function)
+            : this()
+        {
+            if (function == null)
+            {
+                throw new ArgumentNullException($"{nameof(function)} cannot be null", nameof(function));
```
MikeStall, line 67: >$"{nameof(function)} cannot be null", [](http://example.com/codeflow?start=48&length=37)

 nit: remove these, just use ArgumentNullException(nameof( ))

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,344 @@
...
+                throw new ArgumentNullException(nameof(function));
+            }
+
+            var fList = WithNameInternal(function.Name);
+
+            if (fList.Any())
+            {
+                if (fList.Contains(function))
+                {
+                    throw new ArgumentException($"Function {function.Name} is already part of core or extra functions");
```
MikeStall, line 157: >ArgumentException($"Function  [](http://example.com/codeflow?start=30&length=29)

nit - this one

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,344 @@
...
+                _functions.Add(function.Name, new List<TexlFunction>() { function });
+            }
+
+            var fInvariantList = WithInvariantNameInternal(function.LocaleInvariantName);
+
+            if (fInvariantList.Any())
+            {
+                if (fInvariantList.Contains(function))
+                {
+                    throw new ArgumentException($"Function {function.Name} is already part of core or extra functions (invariant)");
```
MikeStall, line 173: >ArgumentException($"Functio [](http://example.com/codeflow?start=30&length=27)

nit - This one

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,344 @@
...
+                _functionsInvariant.Add(function.LocaleInvariantName, new List<TexlFunction>() { function });
+            }
+
+            var fnsList = WithNamespaceInternal(function.Namespace);
+
+            if (fnsList.Any())
+            {
+                if (fnsList.Contains(function))
+                {
+                    throw new ArgumentException($"Function {function.Name} is already part of core or extra functions (namespace)");
```
MikeStall, line 189: > ArgumentException($"Function  [](http://example.com/codeflow?start=29&length=30)

nit - this one

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,344 @@
...
+                throw new ArgumentNullException(nameof(functions));
+            }
+
+            foreach (var func in functions)
+            {
+                Add(func);
+            }
+        }
+
+        internal TexlFunctionSet(TexlFunctionSet other)
```
MikeStall, line 100: >TexlFunctionSet [](http://example.com/codeflow?start=17&length=15)

Can this be private/ Only used by Clone()?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,344 @@
...
+
+        internal IEnumerable<TexlFunction> WithName(string name) => WithNameInternal(name);
+
+        private List<TexlFunction> WithNameInternal(string name) => _functions.ContainsKey(name) ? _functions[name] : new List<TexlFunction>();
+
+        internal IEnumerable<TexlFunction> WithName(string name, DPath ns) => _functions.ContainsKey(name) ? new List<TexlFunction>(_functions[name].Where(f => f.Namespace == ns)) : new List<TexlFunction>();
+
+        internal IEnumerable<TexlFunction> WithInvariantName(string name) => WithInvariantNameInternal(name);
+
+        private List<TexlFunction> WithInvariantNameInternal(string name) => _functionsInvariant.ContainsKey(name) ? _functionsInvariant[name] : new List<TexlFunction>();
```
MikeStall, line 43: >_functionsInvariant.ContainsKey(name) ? _functionsInvariant[name] [](http://example.com/codeflow?start=77&length=65)

nit; use TryGetValue  - single lookup. rather than the double lookup

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,344 @@
...
+                    List<TexlFunction> fnsList = WithNamespaceInternal(f.Namespace);
+
+                    if (fnsList.Count == 1)
+                    {
+                        _namespaces.Remove(f.Namespace);
+                    }
+                    else
+                    {
+                        fnsList.Remove(f);
+                        _namespaces[f.Namespace] = fnsList;
```
MikeStall, line 324: >_namespaces[f.Namespace] = fnsList; [](http://example.com/codeflow?start=24&length=35)

Why do we need this?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,344 @@
...
+                    }
+
+                    f.GetRequiredEnumNames().ToList().ForEach(removedEnum => _enums.Remove(removedEnum));
```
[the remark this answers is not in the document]
MikeStall: I see - _enums would have multiple copies (1 per each Func) and this only removes 1 copy. 

And the callers are tolerant to multiple copies as well.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
```diff
@@ -14,13 +12,17 @@ namespace Microsoft.PowerFx.Core.Texl
...
     internal class BuiltinFunctionsCore
     {
-        public static IEnumerable<TexlFunction> BuiltinFunctionsLibrary => _library;
+        // Slow API, only use for backward compatibility
+#pragma warning disable CS0618 // Type or member is obsolete        
+        public static IEnumerable<TexlFunction> BuiltinFunctionsLibrary => _library.Functions;
 
-        internal static IEnumerable<TexlFunction> TestOnly_AllBuiltinFunctions => _library.Concat(_featureGateFunctions);
+        // Slow API, only use for backward compatibility
+        internal static IEnumerable<TexlFunction> TestOnly_AllBuiltinFunctions => _library.Add(_featureGateFunctions).Functions;
```
MikeStall, line 20: > _library.Add [](http://example.com/codeflow?start=81&length=13)

Does this mutates _library?

src/ReadBenchmarkData.ps1
```diff
@@ -21,7 +21,7 @@ function ConvertToMs([string]$str)
             ## Convert to milliseconds
             if     ($parts[1] -eq "s")  { $val *= 1000    }
             elseif ($parts[1] -eq "ms") {                 } ## Do nothing
-            elseif ($parts[1] -eq "µs") { $val /= 1000    }
+            elseif ($parts[1] -eq "μs") { $val /= 1000    }
```
MikeStall, line 24: >"μs") [](http://example.com/codeflow?start=34&length=5)

?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,322 @@
...
+                _functionsInvariant.Add(function.LocaleInvariantName, new List<TexlFunction>() { function });
+            }
+
+            var fnsList = WithNamespaceInternal(function.Namespace);
+
+            if (fnsList.Any())
+            {
+                if (fnsList.Contains(function))
+                {
+                    throw new ArgumentException($"Function {function.Name} is already part of core or extra functions (namespace)");
```
jas-valgotar, line 172: >throw new ArgumentException($"Function {function.Name} is already part of core or extra functions (namespace)"); [](http://example.com/codeflow?start=20&length=112)

Should we worry about side effect, if this throws we don't remove from _functionsInvariantand & _functions? similarly for above throw?

jas-valgotar (approved):

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,344 @@
...
+            if (functionSet == null)
+            {
+                throw new ArgumentNullException(nameof(functionSet));
+            }
+
+            if (_count == 0)
+            {
+                _functions = new Dictionary<string, List<TexlFunction>>(functionSet._functions);
+                _functionsInvariant = new Dictionary<string, List<TexlFunction>>(functionSet._functionsInvariant);
+                _namespaces = new Dictionary<DPath, List<TexlFunction>>(functionSet._namespaces);
```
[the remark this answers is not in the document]
LucGenetier: This seems to be a safe construction here

void Main()
{
 Dictionary<string, List<TexlFunction>> d = new Dictionary<string, List<TexlFunction>>() { { "a", new List<TexlFunction>() { new TexlFunction("a") } } };
 Dictionary<string, List<TexlFunction>> d2 = d.ToDictionary(kvp => kvp.Key, kvp => new List<TexlFunction>(kvp.Value));
 d2["a"].Add(new TexlFunction("b"));
 d.Dump();
 
 Dictionary<string, List<TexlFunction>> d3 = new Dictionary<string, List<TexlFunction>>(d);
 d3["a"].Add(new TexlFunction("c"));
 d.Dump();
}

class TexlFunction
{
 public string Name { get; set; }
 public TexlFunction(string x) { Name = x; }
}

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -1438,5 +1435,10 @@ internal ArgPreprocessor GetGenericArgPreprocessor(int index)
 
             return ArgPreprocessor.None;
         }
+
+        public TexlFunction ToTexlFunctions()
+        {
+            return this;
+        }
```
MikeStall, line 1442: remove?

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -24,12 +24,7 @@ namespace Microsoft.PowerFx
     public class SymbolTable : ReadOnlySymbolTable, IGlobalSymbolNameResolver
     {
         private readonly SlotMap<NameLookupInfo?> _slots = new SlotMap<NameLookupInfo?>();
-
-        // $$$ NEeded by config... make private?
-        internal readonly Dictionary<string, NameLookupInfo> _variables = new Dictionary<string, NameLookupInfo>();
-
-        private DisplayNameProvider _environmentSymbolDisplayNameProvider = new SingleSourceDisplayNameProvider();
```
MikeStall, line 31: Merge conflict?  Did you meant to revert?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,362 @@
...
+
+        internal IEnumerable<TexlFunction> WithName(string name) => WithNameInternal(name);
+
+        private List<TexlFunction> WithNameInternal(string name) => _functions.TryGetValue(name, out List<TexlFunction> result) ? result : new List<TexlFunction>();                
+
+        internal IEnumerable<TexlFunction> WithName(string name, DPath ns) => _functions.TryGetValue(name, out List<TexlFunction> result) ? result.Where(f => f.Namespace == ns) : new List<TexlFunction>();
+
+        internal IEnumerable<TexlFunction> WithInvariantName(string name) => WithInvariantNameInternal(name);
+
+        private List<TexlFunction> WithInvariantNameInternal(string name) => _functionsInvariant.TryGetValue(name, out List<TexlFunction> result) ? result : new List<TexlFunction>();
```
MikeStall, line 44: Returning an orphaned mutable collection is risky - it means we have to double check every callsite for improper usage. 

Can we just remove WithInvariantNameInternal() and have callers invoke _functionsInvariant.TryGetValue directly instead?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
LucGenetier: Method has been removed

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
LucGenetier: Thinking of it ... I fixed the code ;)

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
LucGenetier: Nope + I removed Clone method

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
LucGenetier: yes

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
LucGenetier: Yep, fixed

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
LucGenetier: We should not expose our internal structures (_functions and others), so we cannot have the callers use TryGetValue, hence this logic.
Now, if you think returning an empty enumerable is risky, I'll update the code to return null ... but this moves the risk to the caller which now needs to check the IEnumerable isn't null. I think this is worse but as you have a different opinion and already mentioned this "problem", I'll update the code accordingly.

LucGenetier: After looking at consequences, this change is far too heavy and risky

LucGenetier: The methods that return a List<T> are private.
All other methods are internal and shouldn't be called by anyone.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,362 @@
...
+
+        internal IEnumerable<TexlFunction> WithNamespace(DPath ns) => WithNamespaceInternal(ns);
+
+        private List<TexlFunction> WithNamespaceInternal(DPath ns) => _namespaces.TryGetValue(ns, out List<TexlFunction> result) ? result : new List<TexlFunction>();
+
+        internal IEnumerable<string> Enums => _enums;
+
+        internal TexlFunctionSet()
+        {
+            _functions = new Dictionary<string, List<TexlFunction>>();
```
lesaltzm, line 54: >new Dictionary<string, List<TexlFunction>>(); [](http://example.com/codeflow?start=25&length=45)

nit: needs string comparator

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
lesaltzm: this would be worth adding as a comment

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,362 @@
...
+                    f.GetRequiredEnumNames().ToList().ForEach(removedEnum => _enums.Remove(removedEnum));
+                }
+            }
+
+            // If functions with the given name aren't found, this is ok.
+        }
+
+        internal void RemoveAll(TexlFunction function)
+        {
+            if (_functions.TryGetValue(function.Name, out List<TexlFunction> funcs))
```
lesaltzm, line 320: >List<TexlFunction> [](http://example.com/codeflow?start=58&length=18)

I'd suggest using `var` more for `out` params, the type `List<TexlFunction> is repeated a bunch here and seems pretty clear to me. 

But, that's a personal preference, feel free to ignore.

src/libraries/Microsoft.PowerFx.Core/Functions/Utils.cs
```diff
@@ -16,4 +16,4 @@ public static bool TestBit(this BigInteger value, int bitIndex)
             return !(value & (BigInteger.One << bitIndex)).IsZero;
         }
     }
-}
\ No newline at end of file
+}
```
lesaltzm, line 19: nit: no change to this file?

lesaltzm:
>     internal partial class Intellisense

also an unchanged file?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/SuggestionHandlers/DottedNameNodeSuggestionHandler.cs:16 in 9438c15. [](commit_id = 9438c15d1b3ce7e686b86f9834e537c0c0f4641e, deletion_comment = False)

lesaltzm (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
LucGenetier: I wouldn't say "nit". You're right, this is missing and should probably be made mandatory via a .Net analyzer if it exists

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
LucGenetier: I personally hate 'var' while I admit this is convenient at times.
Here, I would say specifying the type is 'mandatory' as it makes the code much clearer. When you look at the code, if you don't remember the exact type for _functions, you can't know the type of funcs. With the 's' at the end, you could imagine it's an IEnumerable can can't know if that's a List, Array, Dictionary and what objects are stored there.
Now, I agree this is repetitive with lines 331 and 341 but I find it strange if I'd have used one form for line 320 and another for the rest (= makes it suspicious!)
As you said, it's a personal perference...

src/libraries/Microsoft.PowerFx.Core/Functions/Utils.cs
LucGenetier: This file has been changed and changed back...
Now, do we need a CR/LF at the end, do I need to change the file, it's hard to find an argument :)

LucGenetier:
>     internal partial class Intellisense

Same thing
I wish VS would have an easy way to revert such a useless change for a single file

---
In reply to: [1435058876](https://github.com/microsoft/Power-Fx/pull/1019#issuecomment-1435058876) [](http://example.com/codeflow?ancestors=1435058876,1435058876)

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/SuggestionHandlers/DottedNameNodeSuggestionHandler.cs:16 in 9438c15. [](commit_id = 9438c15d1b3ce7e686b86f9834e537c0c0f4641e, deletion_comment = False)

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -0,0 +1,363 @@
...
+                {
+                    foreach (var func in kvp.Value)
+                    {
+                        yield return func;
+                    }
+                }
+            }
+        }
+
+        internal TexlFunction Append(TexlFunction function) => Add(function);
```
CarlosFigueiraMSFT, line 123: >Append [](http://example.com/codeflow?start=30&length=6)

Do we need Append and Add? Can we have a single one?

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

PR #1020, opened 2023-01-24
Support coercions in the DateAdd function

CarlosFigueiraMSFT:
Functions should support implicit coercions from strings/numbers to Date/Time values; currently the DateAdd functions doesn't support it on its first argument. This fixes that.


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DateTime.cs
```diff
@@ -488,6 +493,11 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
                 {
                     var inputColumn = type0.GetNames(DPath.Root).Single();
                     var resultColumnType = inputColumn.Type;
+                    if (nodeToCoercedTypeMap != null && nodeToCoercedTypeMap.TryGetValue(args[0], out var coercedType))
```
anhtruong-MSFT, line 496: is there a reason to null check here and not on line 425?

CarlosFigueiraMSFT: Here if the first argument is a table of date/time/datetime then there will be no coercions, and nodeToCoercedTypeMap will not be set on line 490. In the other case, if the first argument is a date/time/datetime, then it will be handled by the first If in line 420. Having said that, it doesn't hurt to add that check there, so I'll add it as well.

LucGenetier:
✅ No public API change.

anhtruong-MSFT (approved):

LucGenetier:
✅ No public API change.

PR #1021, opened 2023-01-24
Fix IsBlank for empty strings from untyped objects

CarlosFigueiraMSFT:
Currently IsBlank returns false for an empty string from an untyped object; it should return true to be compatible with the "regular" IsBlank function.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #1024, opened 2023-01-25
Fix 767 - invariant translation with symbols 

### LINKED ISSUE #767: Engine.GetInvariantExpression should work wth symbols from config
It needs to be symmetric with Check() - so also accept a SymbolTable rather than just a record. 
Ideally also just accept a CheckResult, since that's already done the binding.
But then that will make it more difficult to share code with GetDisplayExpression (which is similar, but runs in the opposite direction direction)

MikeStall:
Fix #767  - needed for dataverse. 

Add to CheckResult: 
ApplyGetInvariant
ApplyGetLogging

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests/CheckResultTests.cs
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/CheckResultTests.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using Microsoft.PowerFx.Syntax;
 using Microsoft.PowerFx.Types;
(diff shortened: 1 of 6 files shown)


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/CheckResultTests.cs
```diff
@@ -280,5 +282,47 @@ public void TestIRFail()
...
+
+            Assert.Equal("new_field + 2.3", invariant);
+        }
+
+        // CheckResult properly wired up to Apply logging. 
+        [Theory]
+        [InlineData("123+abc", "#$number$# + #$firstname$#", true)] // display names
+        [InlineData("123+", "#$number$# + #$error$#", false)] // error 
+        [InlineData("123,456", "#$number$#", true)] // locales 
+        [InlineData("Power(2,3)", "Power(#$number$#)", true)] // functions aren't Pii
```
anderson-joyle, line 314: >[InlineData("Power(2,3)", "Power(#$number$#)", true)] [](http://example.com/codeflow?start=8&length=53)

Not within the scope of this change but shouldn't the expected result be "Power(#$number$#,#$number$#)"?

MikeStall: yeah, I'd expect that too - had to update the test.

LucGenetier:
✅ No public API change.

anderson-joyle (approved):
:shipit:

PR #1026, opened 2023-01-25
Enables Column suggestion for Patch and Collect function.

### LINKED ISSUE #1012: Collect function does not provide intellisense for the record being added
When using Collect to add a record to a table, when creating the record object you get no intellisense suggestions with the record field names.

image

jas-valgotar:
Fixes #1012


LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

PR #1029, opened 2023-01-25
Fixes the issue with TZ having minutes in utc offset.

### LINKED ISSUE #1015: DateDiff in hours doesn't work properly when in GMT+XX:YY Time zone, where YY is not 00
Below Test should pass in TZ that also has minutes in it (e.g. IST is UTC+5:30):

DateDiff(DateTime(2005,12,31,23,59,59,999), DateTime(2007,1,1,0,0,0), TimeUnit.Hours) = 8761
https://github.com/microsoft/Power-Fx/blob/main/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DateDiff.txt#L130

Issue should be around the rounding at https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs#L333

jas-valgotar:
Fixes #1015 
Now instead of converting to UTC TZ, we subtract 1 hour from the time to accommodate DST.

This helps in cases where TZ has UTC offset.
e.g.
`DateDiff(DateTime(2005,12,31,23,59,59,999), DateTime(2006,1,1,0,0,0), TimeUnit.Hours)`

When converted to UTC from GMT+5:30 (IST), the expression would become
`DateDiff(DateTime(2005,12,31,18,59,59,999), DateTime(2005,12,31,18,30,0), TimeUnit,Hours)`
This would render the Hour the same due to offset having minutes in it.

Instead, Now for example,
`start = DateTime(2023, 3, 12, 0, 0, 0) is in UTC-8`
`end = DateTime(2023, 3, 12, 3, 0, 0) is in UTC-7`
`startUTCOffset - endUTCOffset = -1`
so adding that UTC offset difference to the end(instead of converting both to UTC) will adjust the subtraction for DST
while preserving hours, since cases having minutes offset can potentially change the hour.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
@@ -266,24 +266,15 @@ namespace Microsoft.PowerFx.Functions
                     return new NumberValue(irContext, years);
             }
 
-            var utcOffset = (timeZoneInfo.GetUtcOffset(start) - timeZoneInfo.GetUtcOffset(end)).Duration();
-
-            // If diff of utcOffset is less than hour different and time unit is hour, there is no need to convert.
-            // Help with cases were UTC offset which also has minute component
-            // Ie, in IST (GMT+5:30): DateDiff(DateTime(2022,12,31,23,59,59,999), DateTime(2023,1,1,0,0,0), TimeUnit.Hours) = 1
-            if (!(utcOffset.Hours < 1 && timeUnit is "hours"))
-            {
-                // Convert to UTC to be accurate (apply DST if needed)
-                if (NeedToConvertToUtc(runner, start, timeUnit))
-                {
-                    start = TimeZoneInfo.ConvertTimeToUtc(start, timeZoneInfo);
-                }
-
-                if (NeedToConvertToUtc(runner, end, timeUnit))
-                {
-                    end = TimeZoneInfo.ConvertTimeToUtc(end, timeZoneInfo);
-                }
-            }
+            // This takes care of DST differences
+            // e.g. https://www.timeanddate.com/time/change/usa/seattle?year=2023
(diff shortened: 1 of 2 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
```diff
@@ -301,17 +301,18 @@ public static FormulaValue DateDiff(EvalVisitor runner, EvalVisitorContext conte
...
 
-            // Convert to UTC to be accurate (apply DST if needed)
             var timeZoneInfo = runner.GetService<TimeZoneInfo>() ?? LocalTimeZone;
 
-            if (NeedToConvertToUtc(runner, start, timeUnit))
+            // Subtract an hour if time falls in DST
+            if (timeZoneInfo.IsDaylightSavingTime(start))
             {
-                start = TimeZoneInfo.ConvertTimeToUtc(start, timeZoneInfo);
+                start = start.AddHours(-1);
```
MikeStall, line 309: WE should avoid -1 for DST, and instead rely on ConvertTimeToUtc

MikeStall (changes requested):
see comment about dst

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
jas-valgotar: Agree, Revisiting the logic for it.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1030, opened 2023-01-25
Fix null reference when table has blank() input.

nguhoa:
Add null check for the row and update logic to not only check first row but check the first null row for column def.

### DIFF
--- a/src/tools/Repl/Program.cs
+++ b/src/tools/Repl/Program.cs
@@ -368,6 +368,7 @@ namespace PowerFxHostSamples
                         if (row.Value != null)
                         {
                             columnCount = Math.Max(columnCount, row.Value.Fields.Count());
+                            break;
                         }
                     }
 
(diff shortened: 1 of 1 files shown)


LucGenetier:
✅ No public API change.

src/tools/Repl/Program.cs
```diff
@@ -361,21 +361,38 @@ private static string PrintResult(FormulaValue value, bool minimal = false)
...
+                    var columnCount = 0;
+                    foreach (var row in table.Rows)
+                    {
+                        if (row.Value != null)
+                        {
+                            columnCount = Math.Max(columnCount, row.Value.Fields.Count());
+                        }
+                    }
+
+                    if (columnCount == 0)
```
MikeStall, line 374: >if (columnCount == 0) [](http://example.com/codeflow?start=20&length=21)

Instead of checking columnCount (possibly record with 0 fields), we should check DValue.IsBlank.

src/tools/Repl/Program.cs
```diff
@@ -409,12 +436,15 @@ private static string PrintResult(FormulaValue value, bool minimal = false)
 
                         foreach (var row in table.Rows)
                         {
-                            column = 0;
-                            resultString += "\n ";
-                            foreach (var field in row.Value.Fields)
+                            if (row.Value != null)
```
MikeStall, line 439: >if (row.Value != null) [](http://example.com/codeflow?start=28&length=22)

We should have an Else here for row.IsError, row.Isblank.

nguhoa: Updated the code. thanks

LucGenetier:
✅ No public API change.

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1031, opened 2023-01-26
CheckResult.Symbol needs to stay public 

MikeStall:
Regressed with CheckResult refactor in #993 . It was made private and renamed to AllSymbols.

We have clients that require this property to stay public.


jas-valgotar (approved):

LucGenetier:
✅ No public API change.

PR #1032, opened 2023-01-26
Crash - If type check

### LINKED ISSUE #953: Crash - If type check
This crashes (may need to be run in repl in behavior context) 

`With({y:1}, If(false, {x:1}, {z:2})  )`

https://github.com/microsoft/Power-Fx/blob/2c3ee1bae5f48e7196076877e2fc584825b05b47/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs#L260

                    if (IfFunction.CanCheckIfReturn(func))
                    {
                        Contract.Assert(result.IRContext.ResultType == node.IRContext.ResultType || result is ErrorValue || result.IRContext.ResultType is BlankType);
                    }

This is a combination of several issues:
- we're wrapped in a With(), which breaks some hacky logic in If() 
- The branches have different types. 
- But they're both records - so we don't immediately fail on record/record mismatch.  If it was just record/primitive, then we'd fail to bind. 

Also related to #843.

jas-valgotar:
Fixes #953


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

PR #1035, opened 2023-01-26
Fix scope access bug in IfError function

CarlosFigueiraMSFT:
Currently an expression such as `With({x:1}, IfError(x,x,x))` - which should return 1 - fails to execute. The problem is with the scope access within the IfError function, caused by a bug in the IR translator. This change fixes it, and adds many tests for the IfError function.


LucGenetier:
✅ No public API change.

lesaltzm (approved):

LucGenetier:
✅ No public API change.

PR #1037, opened 2023-01-27
Handle blank values properly in operators

CarlosFigueiraMSFT:
Blank values should be handled as primitive types in functions / operator contexts, but this was not the case for all operators. This change addresses that, and adds tests to "spec" this behavior.


MikeStall (dismissed):

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Drawing;
```
Shpakh-MSFT, line 6: Is this needed?

Shpakh-MSFT: Never mind. I see it now.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1397,7 +1400,8 @@ static Library()
                 StandardErrorHandling<FormulaValue>(
                     BuiltinFunctionsCore.TimeZoneOffset.Name,
                     expandArguments: NoArgExpansion,
-                    replaceBlankValues: DoNotReplaceBlank,
+                    replaceBlankValues: ReplaceBlankWith(
+                        new DateTimeValue(IRContext.NotInSource(FormulaType.DateTime), _epoch)),
```
Shpakh-MSFT, line 1404: Does this need to be update on Power Apps side as well? So I am assuming that this will be under excel compatibility change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/inScalar.txt
```diff
@@ -34,21 +34,19 @@ true
 >> "" in ""
 true
 
-
-// If Blank is on either side, operator is false. 
-// Use If() to coerce Blank to string. 
->> Blank() in Blank()
-false
+// Blank values are coerced to strings
```
ian-legler, line 37: I noticed that "" in Blank() and Blank() in "" both produce binding errors that seem related to the tabular in.  Will this fix affect those? We should have explicit tests for those either way.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/arithmetic.txt
```diff
@@ -119,4 +119,29 @@ Error({Kind:ErrorKind.Div0})
 Error({Kind:ErrorKind.Div0})
 
 >> 5 <> 1/0
-Error({Kind:ErrorKind.Div0})
\ No newline at end of file
+Error({Kind:ErrorKind.Div0})
+
+// Blank coercions
```
ian-legler, line 124: >Blank coercions [](http://example.com/codeflow?start=3&length=15)

Should these have explicit tests for Blank()?

ian-legler:
> // SQL Null comparisons are always false

Nit: This comment is no longer needed

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Date.txt:163 in c816c72. [](commit_id = c816c72c958d2dd1ed084a0bc0a9cda7cc4e96b2, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
CarlosFigueiraMSFT: Yes. Some of those issues are already fixed in Power Apps, others have issues. We'll update them when we are bringing the latest Power Fx bits there.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/inScalar.txt
CarlosFigueiraMSFT: The operator `in` is overloaded - it can be used both for "string in string" or "record in table". Since Blank() doesn't have any type information, the binder doesn't know exactly which overload to take, and it will pick one - probably the incorrect one. This is not a real scenario - blank values would typically come from unitialized variables, or from a typed expression (such as `If(1<0,"hello")`, which is why I'm using the `If(1<0,<typed expression>)` syntax to specify a "typed blank value"

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/inScalar.txt
```diff
@@ -34,21 +34,19 @@ true
 >> "" in ""
 true
 
-
-// If Blank is on either side, operator is false. 
-// Use If() to coerce Blank to string. 
->> Blank() in Blank()
```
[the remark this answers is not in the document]
CarlosFigueiraMSFT: From the other comment - `Blank() in Blank()` is ambiguous, since the `in` operator is overloaded. This is also not something that one would write in a real app. I'm replacing this test with others where we have "typed blank values"

CarlosFigueiraMSFT:
> // SQL Null comparisons are always false

Removing...

---
In reply to: [1406913638](https://github.com/microsoft/Power-Fx/pull/1037#issuecomment-1406913638) [](http://example.com/codeflow?ancestors=1406913638,1406913638)

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Date.txt:163 in c816c72. [](commit_id = c816c72c958d2dd1ed084a0bc0a9cda7cc4e96b2, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/arithmetic.txt
CarlosFigueiraMSFT: I don't think it's necessary. Blank() by itself isn't very interesting - we would typically use it in an expression that has explicit type. But I added those here as well, txt tests are pretty fast to run.

LucGenetier:
✅ No public API change.

Shpakh-MSFT (approved):
:shipit:

ian-legler (approved):
:shipit:

Issue #1039, opened 2023-01-28
Culture: check culture on regex


MikeStall:
Our regex should have this flag: 
- RegexOptions.Compiled - should always be compile-time constants
- RegexOptions.CultureInvariant  - we should never be checking the current thread's culture. 

And not any other flags, especially any case insensitive compares. 

https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-options#compare-using-the-invariant-culture

PR #1040, opened 2023-01-29
CultureInfo support for errors

### LINKED ISSUE #918: Texl errors are not localized
Constructor of BaseError has the following line:

(var shortMessage, var longMessage) = ErrorUtils.GetLocalizedErrorContent(errKey, null, out var errorResource);

Null being passed is the CultureInfo. As we always pass null, errors are not localized. Here is the unit test that demonstrates the issue. (to copy paste in LanguageServerTests file):

        [Fact]
        public void ErrorIsLocalized()
        {
            Thread.CurrentThread.CurrentCulture = new System.Globalization.CultureInfo("fr-fr");
            config.AddFunction(new BehaviorFunction());
            var engine = new Engine(config);

            _sendToClientData = new List<string>();
            _scopeFactory = new TestPowerFxScopeFactory(
                (string documentUri) => engine.CreateEditorScope(new ParserOptions() { Culture = CultureInfo.InvariantCulture }, GetFromUri(documentUri)),
                GetParserOptions(false));
            _testServer = new TestLanguageServer(_sendToClientData.Add, _scopeFactory);

            _testServer.OnDataReceived(
                JsonSerializer.Serialize(new
                {
                    jsonrpc = "2.0",
                    method = "textDocument/didOpen",
                    @params = new DidOpenTextDocumentParams()
                    {
                        TextDocument = new TextDocumentItem()
                        {
                            Uri = "powerfx://app",
                            LanguageId = "powerfx",
                            Version = 1,
                            Text = "Bla."
                        }
                    }
                }));

            CheckBehaviorError(_sendToClientData[0], false, out var diags);

            Assert.DoesNotContain("Unexpected characters.", diags.First().Message); // the value should be localized. Resx files have this localized.
        }

MikeStall:
Allow Errors to be in separate culture info that parser. 
Fix #918, and also #1039. Sets up a fix for #990.  

Make ExpressionError.Message localize lazily.  
- Callers can get the localized message in any culture after the parse/bind
- just checking IsSuccess can then avoid localization since that doesn't read the message property 

Add test to ensure we're not using Thread.CurrentCulture when a culture is provided. 
- Update a bunch of usages.  StartsWith(), EndsWith() ,etc require StringComparison.Ordinal, else they will check current culture. 
- RegEx required CultureInvariant 

Add more tests
- add ExpressionErrorTests to get 100% code coverage on ExpressionError. 
- add MultiCulture* tests to demonstrate how we can set Parse/ErrorMessage/Runtime  cultures independently 
- remove parserOptions usage in LSP that wasn't being used.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/ExpressionError.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/ExpressionError.cs
@@ -18,10 +18,32 @@ namespace Microsoft.PowerFx
     /// </summary>
     public class ExpressionError
     {
+        public ExpressionError()
+        {
+        }
+
         /// <summary>
         /// A description of the error message. 
         /// </summary>
-        public string Message { get; set; }
+        public string Message
+        {
+            get
+            {
+                if (_message == null && this.MessageKey != null)
+                {
+                    (var shortMessage, var _) = ErrorUtils.GetLocalizedErrorContent(new ErrorResourceKey(this.MessageKey), _messageLocale, out _);
+
+                    var msg = ErrorUtils.FormatMessage(shortMessage, _messageLocale, _messageArgs);
+
+                    _message = msg;
+                }
+
+                return _message;
+            }
+
+            // If this is set directly, it will skip localization. 
+            set => _message = value;
+        }
 
         /// <summary>
         /// Source location for this error.
@@ -42,6 +64,43 @@ namespace Microsoft.PowerFx
         /// </summary>
         public bool IsWarning => Severity < ErrorSeverity.Severe;
 
+        // localize message lazily 
+        private string _message; 
+        private object[] _messageArgs;
+        private CultureInfo _messageLocale;
+
+        internal CultureInfo MessageLocale => _messageLocale;
+
+        /// <summary>
+        /// Get a copy of this error message for the given locale. 
+        /// <see cref="Message"/> will get lazily localized using <see cref="MessageKey"/>.
+        /// </summary>
+        /// <param name="culture"></param>
+        /// <returns></returns>
+        public ExpressionError GetInLocale(CultureInfo culture)
+        {
+            // In order to localize, we need a message key
+            if (this.MessageKey != null)
+            {
+                var error = new ExpressionError
+                {
+                    Span = this.Span,
+                    Kind = this.Kind,
+                    Severity = this.Severity,
+                    MessageKey = this.MessageKey
+                };
+
+                // New message can be localized
+                error._message = null; // will be lazily computed in new locale 
+                error._messageArgs = this._messageArgs;
+                error._messageLocale = culture;
+
+                return error;
+            }
+           
+            return this;            
+        }   
+
         public override string ToString()
         {
             var prefix = IsWarning ? "Warning" : "Error";
@@ -51,7 +110,7 @@ namespace Microsoft.PowerFx
             }
             else
             {
-                return $"{prefix} {Message}";
+                return $"{prefix}: {Message}";
             }    
         }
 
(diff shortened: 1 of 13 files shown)


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/Helpers/BombCulture.cs
```diff
@@ -0,0 +1,80 @@
...
+
+namespace Microsoft.PowerFx
+{
+#pragma warning disable CA1065 // Do not raise exceptions in unexpected locations
+
+    // Used to verify that that we're not accessing this culture. 
+    // For example, if we want to verify we don't access Thread.CurrentCulture. 
+    // then just set it to BombCulture and any access will throw. 
+    [DebuggerDisplay("Bomb culture")]
+    public class BombCulture : CultureInfo
```
MikeStall, line 18: I like this technique - adding a comment here to draw attention to it.  It made some surprising discoveries  (see changes in \Functions\Library*.cs)

src/tests/Microsoft.PowerFx.Interpreter.Tests/ConfigTests.cs
```diff
@@ -370,98 +371,184 @@ public void RecalcEngine_Symbol_CultureInfo4()
...
-        public void RecalcEngine_Symbol_CultureInfo6()
+        private void MultiCulture1()
         {
-            Exception exception = null;
+            RunOnIsolatedThread(_bombCulture, MultiCulture1ThreadProc);
+        }
 
-            var t = new Thread(() =>
+        // Test Parse and errors 
+        private void MultiCulture1ThreadProc()
```
MikeStall, line 464: These new MultiCulture* tests are the key addition here. 
The rest of the changes are noisy diffs, due to refactoring to use RunOnIsolatedThread helper

src/libraries/Microsoft.PowerFx.Core/Public/ExpressionError.cs
```diff
@@ -42,6 +64,43 @@ public class ExpressionError
...
+
+        internal CultureInfo MessageLocale => _messageLocale;
+
+        /// <summary>
+        /// Get a copy of this error message for the given locale. 
+        /// <see cref="Message"/> will get lazily localized using <see cref="MessageKey"/>.
+        /// </summary>
+        /// <param name="culture"></param>
+        /// <returns></returns>
+        public ExpressionError GetInLocale(CultureInfo culture)
```
LucGenetier, line 80: >GetInLocale [](http://example.com/codeflow?start=31&length=11)

Not sure what this means

src/libraries/Microsoft.PowerFx.Core/Types/DTypeSpecParser.cs
```diff
@@ -46,7 +47,7 @@ public static bool TryParse(DTypeSpecLexer lexer, out DType type)
                 return true;
             }
 
-            var typeIdx = _typeEncodings.IndexOf(token);
+            var typeIdx = _typeEncodings.IndexOf(token, StringComparison.Ordinal);
```
LucGenetier, line 50: > StringComparison.Ordinal); [](http://example.com/codeflow?start=55&length=27)

This is excellent but we should enable globalization .Net analyzers at the same time

src/tests/Microsoft.PowerFx.Core.Tests/Helpers/BombCulture.cs
LucGenetier: I don't like "bomb" term or any war reference
I think InvalidCulture is sufficient

LucGenetier (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Types/DTypeSpecParser.cs
MikeStall: +1 for enabling globalization .Net analyzers - it would probably have caught these and others.  But out of scope for this PR.

anderson-joyle (approved):
:shipit:

PR #1041, opened 2023-01-29
Allows Date and DateTime utc and fixes bugs around the same.

### LINKED ISSUE #718: Add support for UTCNow, UTCToday, IsUTCToday
`Now()`, `Today()`, `IsToday()` are taking the time zone of the "current user", which is not ideal for server-side product to consume.

Can we support `UTCNow`, `UTCToday`, `IsUTCToday`, like how Microsoft Dataverse has it, as mentioned in the [documentation](https://learn.microsoft.com/en-us/power-platform/power-fx/reference/function-now-today-istoday#description)?

jas-valgotar:
Fixes #718 
- Now `Date()` `DateValue()` accepts UTC time.
- Fixes `IsToday()` to take time zone in the account.


src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNew.cs
```diff
@@ -81,21 +81,11 @@ public static DateValue NewDateOnly(DateTime value)
                 throw new ArgumentException("Invalid DateValue, the provided DateTime contains a non-zero TimeOfDay");
             }
 
-            if (value.Kind == DateTimeKind.Utc)
```
MikeStall, line 84: Yes! removing these here is great.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
```diff
@@ -611,6 +636,33 @@ private static async ValueTask<FormulaValue> Now(EvalVisitor runner, EvalVisitor
...
         }
 
+        /// <summary>
+        /// This function is used to convert Local time if Time Zone is set to UTC
+        /// and given time is in Local time.
+        /// </summary>
+        /// <param name="dateTime">Date Time to be converted.</param>
+        /// <param name="timeZoneInfo"> Current Engine's Timezone.</param>
+        /// <returns></returns>
+        internal static DateTime PotenntiallyConvertToUtc(DateTime dateTime, TimeZoneInfo timeZoneInfo)
```
MikeStall, line 646: >PotenntiallyConvertToUtc [](http://example.com/codeflow?start=33&length=24)

nit: typo 
"potentially"

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -624,6 +624,22 @@ private FormulaValue GetVariableOrFail(ResolvedObjectNode node, ISymbolSlot slot
                 var value = _symbolValues.Get(slot);
                 if (value != null)
                 {
+                    DateTime dateTime = default;
```
MikeStall, line 627: >DateTime dateTime = default; [](http://example.com/codeflow?start=20&length=28)

This is the right idea, but we need to do this check in a more common place. The problem is that there are other paths to get a DateValue beside just symbol lookup (say field access). 

Maybe a helper directly on DateValue,DateTimeValue like 
"GetTimeInUTC(Engine engine)"

- Note it will need to take some extra param (like Engine) to get the TZI. 
- The benefit of putting it directly on Date*Value, especially next to existing Value, is that it's easier to see it and make sure it's getting called everywhere.

src/tests/Microsoft.PowerFx.Interpreter.Tests/DateTimeTests.cs
```diff
@@ -0,0 +1,97 @@
...
+using Xunit;
+
+namespace Microsoft.PowerFx.Interpreter.Tests
+{
+    public class DateTimeTests
+    {
+        private static readonly bool IsLocalUTC = TimeZoneInfo.Local.BaseUtcOffset == TimeSpan.Zero;
+
+        [Fact]
+        public async void DateTimeUTCTests()
```
MikeStall, line 17: >DateTimeUTCTests [](http://example.com/codeflow?start=26&length=16)

Nice!

Can we add a test that does a DateDiff between 2 dates, one in locale, one in UTC?

src/tests/Microsoft.PowerFx.Interpreter.Tests/DateTimeTests.cs
```diff
@@ -0,0 +1,97 @@
...
+            var symbolTable = new SymbolTable();
+            var symbolValues = new SymbolValues(symbolTable);
+            var myDateTimeSlot = symbolTable.AddVariable("myDateTime", FormulaType.DateTime);
+            symbolValues.Set(myDateTimeSlot, FormulaValue.New(localNow));
+            
+            // Creates Today's expression such that IsToday() can be true for local
+            // and false for UTC
+            var todayLocalDateTimeSlot = symbolTable.AddVariable("todayLocalDateTime", FormulaType.DateTime);
+            var localToday = DateTime.Today;
+            if (TimeZoneInfo.Local.BaseUtcOffset > TimeSpan.Zero)
```
MikeStall, line 34: >TimeZoneInfo.Local.BaseUtcOffset > TimeSpan.Zero [](http://example.com/codeflow?start=16&length=48)

This expression worries me because it will behave differently in different timezones, may pass in some cases, fail in others. - and appear non-deterministic.  For example, Seattle, Paris, IST.

MikeStall: Since we can set the TimeZone on the engine  (see SymbolExtensions.SetTimeZone) , can we set the timezone to a specific value and make this determinsitic? 

We should never be calling TimeZoneInfo.Local.

src/tests/Microsoft.PowerFx.Interpreter.Tests/DateTimeTests.cs
jas-valgotar: Added

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
jas-valgotar: Added a new Helper.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
jas-valgotar: corrected.

src/tests/Microsoft.PowerFx.Interpreter.Tests/DateTimeTests.cs
jas-valgotar: Now setting local timezone to IST.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateValue.cs
```diff
@@ -14,6 +14,28 @@ namespace Microsoft.PowerFx.Types
     /// </summary>
     public class DateValue : PrimitiveValue<DateTime>
     {
+        /// <summary>
+        /// Converts To UTC time if value is not utc and <paramref name="timeZoneInfo"/> is utc.
+        /// Converts from UTC time if  value is utc and <paramref name="timeZoneInfo"/> is non utc.
+        /// else returns the  value/>/>.
+        /// NOTE: if <paramref name="timeZoneInfo"/> is null, Local time zone is used.
+        /// </summary>
+        public DateTime GetConvertedDateValue(TimeZoneInfo timeZoneInfo)
```
MikeStall, line 23: Can we share this with DateTimeValue class?
(eg, a common static helper)

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateTimeValue.cs
```diff
@@ -15,6 +15,28 @@ namespace Microsoft.PowerFx.Types
...
+        /// Converts To UTC time if value is not utc and <paramref name="timeZoneInfo"/> is utc.
+        /// Converts from UTC time if value is utc and <paramref name="timeZoneInfo"/> is non utc.
+        /// else returns the value.
+        /// NOTE: if <paramref name="timeZoneInfo"/> is null, Local time zone is used.
+        /// </summary>
+        public DateTime GetConvertedDateTimeValue(TimeZoneInfo timeZoneInfo)
+        {
+            timeZoneInfo ??= TimeZoneInfo.Local;
+
+            if (_value.Kind != DateTimeKind.Utc && timeZoneInfo.BaseUtcOffset == TimeSpan.Zero)
```
MikeStall, line 28: >timeZoneInfo.BaseUtcOffset == TimeSpan.Zero [](http://example.com/codeflow?start=51&length=43)

do we need this check?  (is it already handled by ConvertTimeToUtc)

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -624,11 +627,30 @@ private FormulaValue GetVariableOrFail(ResolvedObjectNode node, ISymbolSlot slot
                 var value = _symbolValues.Get(slot);
                 if (value != null)
                 {
+                    value = AdjustDateorDateTimeValue(value);
```
MikeStall, line 630: >AdjustDateorDateTimeValue [](http://example.com/codeflow?start=28&length=25)

Do we need this yet - or can we defer until point of use?

There are too many ways we might get a FormulaValue

MikeStall:
>         private static TimeZoneInfo LocalTimeZone => TimeZoneInfo.Local;

Can we get rid of this and instead just set it once at runner.GetService<TimeZoneInfo> ? That' s what we do for CultureInfo.

That avoids having to check at every usage, like:

var timeZoneInfo = runner.GetService<TimeZoneInfo>() ?? LocalTimeZone;

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs:57 in 0e4cbb7. [](commit_id = 0e4cbb7af54796dff0606ed0d9f6e9223e7d24ef, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -168,8 +168,9 @@ private class MinDateAndDateTimeAgg : IAggregator
             protected DateTime _minValueDT = DateTime.MaxValue;
             protected int _count = 0;
 
-            public void Apply(FormulaValue value)
+            public void Apply(FormulaValue value, EvalVisitor runner)
```
MikeStall, line 171: >EvalVisitor runner [](http://example.com/codeflow?start=50&length=18)

Can we avoid passing EvalVisitor and instead pass a more restrictive object?  "Least privilege" principle - EV is too powerful 

Like runtimeConfig or IServiceProvider?

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateTimeValue.cs
```diff
@@ -15,6 +15,28 @@ namespace Microsoft.PowerFx.Types
     /// </summary>
     public class DateTimeValue : PrimitiveValue<DateTime>
```
MikeStall, line 16: Is there something we can do like 'new' keyword on Value to block usage of that and force via the safe helper

src/tests/Microsoft.PowerFx.Interpreter.Tests/DateTimeTests.cs
```diff
@@ -0,0 +1,131 @@
...
+using System.Text;
+using System.Threading;
+using Microsoft.PowerFx.Types;
+using Xunit;
+
+namespace Microsoft.PowerFx.Interpreter.Tests
+{
+    public class DateTimeTests
+    {
+        private static readonly bool IsLocalUTC = TimeZoneInfo.Local.BaseUtcOffset == TimeSpan.Zero;
```
MikeStall, line 15: nit - can we remove now?

src/tests/Microsoft.PowerFx.Interpreter.Tests/DateTimeTests.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        [Fact]
+        public async void DateTimeUTCTests()
```
MikeStall, line 18: >DateTimeUTCTests [](http://example.com/codeflow?start=26&length=16)

nit: can we split this up into a few tests, just for maintainability. 
Doesn't matter exactly how, maybe at least 3. 

Generally, I like ordering the tests by complexity - having a basic one and getting increasingly complex.

src/tests/Microsoft.PowerFx.Interpreter.Tests/DateTimeTests.cs
```diff
@@ -0,0 +1,131 @@
...
+using Microsoft.PowerFx.Types;
+using Xunit;
+
+namespace Microsoft.PowerFx.Interpreter.Tests
+{
+    public class DateTimeTests
+    {
+        private static readonly bool IsLocalUTC = TimeZoneInfo.Local.BaseUtcOffset == TimeSpan.Zero;
+
+        [Fact]
```
susosuso, line 17: Can we also test `DateTimeValue()`, `Today()`, and `IsToday()` functions?

- the function `DateTimeValue` would be internally evaluated to c# DateTime object of the correct time zone. Like `DateTimeValue("2022-10-01")` should be evaluated as 2022-10-01 in UTC when `Symbol.SetTimeZone(TimeZoneInfo.Utc)`
- `IsToday()` and `Today()` should honor the setting by SetTimeZone()

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateTimeValue.cs
jas-valgotar: We don't wanna attempt to convert to UTC if setTimeZone isn't UTC, otherwise ConvertTimeToUtc would throw

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
jas-valgotar: Agree, Changing it to IServiceProvide

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateValue.cs
jas-valgotar: done.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
jas-valgotar: Sure, updated.

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateTimeValue.cs
jas-valgotar: Added new keyword, to block usage of the field directly.

src/tests/Microsoft.PowerFx.Interpreter.Tests/DateTimeTests.cs
jas-valgotar: done.

src/tests/Microsoft.PowerFx.Interpreter.Tests/DateTimeTests.cs
```diff
@@ -0,0 +1,131 @@
...
+            var utcSymbol = new RuntimeConfig(symbolValues);
+            utcSymbol.SetTimeZone(TimeZoneInfo.Utc);
+
+            var localSymbol = new RuntimeConfig(symbolValues);
+            localSymbol.SetTimeZone(istTimeZone);
+
+            // Test UTC now as a variable
+            var evaluator = engine.Check("myDateTime", null, symbolTable).GetEvaluator();           
+            var result = await evaluator.EvalAsync(default, localSymbol);
+            Assert.Equal(localISTNow.ToString("g"), ((DateTimeValue)result).Value.ToString("g"));
```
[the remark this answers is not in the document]
jas-valgotar: I have updated the test with `("o")`, except for the `Now()` test due to the millisecond precision difference.

src/tests/Microsoft.PowerFx.Interpreter.Tests/DateTimeTests.cs
jas-valgotar: Added the new tests. Feel free to point out any specifics on those tests!

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateTimeValue.cs
```diff
@@ -15,6 +15,46 @@ namespace Microsoft.PowerFx.Types
...
     {
+        /// <summary>
+        /// Direct use of <see cref="Value"/> is prohibited in favor of <see cref="GetConvertedValue(TimeZoneInfo)"/> method.
+        /// </summary>
+        public new DateTime Value => throw new NotSupportedException(string.Format("Use Method {0} instead, with proper timezone information.", nameof(GetConvertedValue)));
```
jas-valgotar, line 21: >public new DateTime Value => throw new NotSupportedException(string.Format("Use Method {0} instead, with proper timezone information.", nameof(GetConvertedValue))); [](http://example.com/codeflow?start=8&length=164)

Note: This is a breaking change.

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateValue.cs
```diff
@@ -14,6 +14,22 @@ namespace Microsoft.PowerFx.Types
     /// </summary>
     public class DateValue : PrimitiveValue<DateTime>
     {
+        /// <summary>
+        /// Direct use of <see cref="Value"/> is prohibited in favor of <see cref="GetConvertedValue(TimeZoneInfo)"/> method.
+        /// </summary>
+        public new DateTime Value => throw new NotSupportedException(string.Format("Use Method {0} instead, with proper timezone information.", nameof(GetConvertedValue)));
```
jas-valgotar, line 20: >public new DateTime Value => throw new NotSupportedException(string.Format("Use Method {0} instead, with proper timezone information.", nameof(GetConvertedValue))); [](http://example.com/codeflow?start=8&length=164)

Note: This is a breaking change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateTimeValue.cs
MikeStall: We could use [Obsolete] attribute on this and still make it work. 
[Obsolete(...)]
public new DateTime Value => base.Value
 
We'll get a warning in our own code, which is probably sufficient. .
But then determined callers can still access the Value they passed in.

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateTimeValue.cs
```diff
@@ -15,6 +15,46 @@ namespace Microsoft.PowerFx.Types
...
+
+        /// <summary>
+        /// Converts To UTC time if value is not utc and <paramref name="timeZoneInfo"/> is utc.
+        /// Converts from UTC time if  value is utc and <paramref name="timeZoneInfo"/> is non utc.
+        /// else returns the  value/>/>.
+        /// NOTE: if <paramref name="timeZoneInfo"/> is null, Local time zone is used.
+        /// </summary>
+        public DateTime GetConvertedValue(TimeZoneInfo timeZoneInfo)
```
susosuso, line 29: Just so i understand, so if someone set `symbolValues.SetTimeZone()` , it isn't really honored here, and they still have to call GetConvertedValue() everytime they read a DateTime value?

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateTimeValue.cs
```diff
@@ -15,6 +15,46 @@ namespace Microsoft.PowerFx.Types
...
+        {
+            return GetConvertedDateTimeValue(_value, timeZoneInfo);
+        }
+
+        internal static DateTime GetConvertedDateTimeValue(DateTime value, TimeZoneInfo timeZoneInfo)
```
MikeStall, line 34: >GetConvertedDateTimeValue [](http://example.com/codeflow?start=33&length=25)

Can we adda  unit test for this that covers the various cases? 

I think there's at least 4? 
(value is utc / notUtc) * (timeZoneInfo is utc / notUtc)

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateTimeValue.cs
```diff
@@ -15,6 +15,46 @@ namespace Microsoft.PowerFx.Types
...
+        {
+            // If timeZoneInfo is null and stored value is in UTC, we don't want to convert.
+            if (timeZoneInfo == null && value.Kind == DateTimeKind.Utc)
```
MikeStall, line 37: >timeZoneInfo == null [](http://example.com/codeflow?start=16&length=20)

Can we ever have null?

MikeStall:
>             var timeUnit = ((StringValue)args[2]).Value.ToLowerInvariant();

this is great - we normalize to same timezone.
And ven better - we avoid conversions if it's not necessary (since conversions can be potentially lossly) 

Can we add this assert:

Contract.Assert(start.Kind== end.Kind) 



---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs:290 in b03cf5d. [](commit_id = b03cf5d1cdb1e3e7aed09f165349235694eb04f8, deletion_comment = False)

MikeStall:
>             }

This pattern repeats so many times ... 
Could we put behind a wrapper, like:

DateTime arg0 = GetNormalizedDateTime(irContext, args[0]);

For error handling, maybe throw a CustomFunctionErrorException? and 
- allow it to accept a ErrorValue
- Call visitor will have to catch it. 




---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs:421 in b03cf5d. [](commit_id = b03cf5d1cdb1e3e7aed09f165349235694eb04f8, deletion_comment = False)

MikeStall:
>             }

And this helper could be GetNormalizedTimeSpan 

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs:471 in b03cf5d. [](commit_id = b03cf5d1cdb1e3e7aed09f165349235694eb04f8, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
```diff
@@ -532,11 +542,17 @@ private static FormulaValue DateImpl(EvalVisitor runner, EvalVisitorContext cont
             // such as: Date(2000, 25, 69) -> 3/10/2002
             try
             {
-                var datetime = new DateTime(year, 1, 1)
+                var timeZoneInfo = runner.TimeZoneInfo;
+
+                var dateTimeKind = timeZoneInfo.BaseUtcOffset == TimeSpan.Zero ?
```
MikeStall, line 547: > timeZoneInfo.BaseUtcOffset == TimeSpan.Zero [](http://example.com/codeflow?start=34&length=44)

This expression is tricky -  can we factor behind a property? 

var dateTimeKind  = runner.GetTimeZoneKind;

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -14,12 +14,21 @@ namespace Microsoft.PowerFx.Functions
     // Direct ports from JScript. 
     internal static partial class Library
     {
+        /// <summary>
+        /// Get a service from the <paramref name="serviceProvider"/>,
+        /// Returns null if not present.
+        /// </summary>
+        public static T GetService<T>(IServiceProvider serviceProvider)
```
MikeStall, line 21: >GetService [](http://example.com/codeflow?start=24&length=10)

Can we put on more common place? like Utility.cs?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -14,12 +14,21 @@ namespace Microsoft.PowerFx.Functions
...
+            return (T)serviceProvider.GetService(typeof(T));
+        }
+
         private static readonly IRandomService _defaultRandService = new DefaultRandomService();
 
         // Support for aggregators. Helpers to ensure that Scalar and Tabular behave the same.
         private interface IAggregator
         {
-            void Apply(FormulaValue value);
+            void Apply(FormulaValue value, IServiceProvider serviceProvider);
```
MikeStall, line 31: >, IServiceProvider serviceProvider [](http://example.com/codeflow?start=41&length=34)

The parameter here means every call to the aggregator will have to repeat the lookup. 

Aggregators are short lived and allocated here .
So instead, can we pass the parameter to the ctor and save as a field?

And then IAggregator is unchanged.

jas-valgotar: That's a great idea! Thanks!

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateTimeValue.cs
jas-valgotar: added.

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateTimeValue.cs
jas-valgotar: That is correct, But For now, you can still use Value property directly but we are marking it Obsolete.

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateTimeValue.cs
jas-valgotar: done.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -167,9 +167,16 @@ private class MinDateAndDateTimeAgg : IAggregator
...
+            private readonly IServiceProvider _serviceProvider;
+
+            public MinDateAndDateTimeAgg(IServiceProvider serviceProvider)
+            {
+                _serviceProvider = serviceProvider;
+            }
 
             public void Apply(FormulaValue value)
             {
+                var timeZoneInfo = Utility.GetService<TimeZoneInfo>(_serviceProvider);
```
MikeStall, line 179: > var timeZoneInfo = Utility.GetService<TimeZoneInfo>(_serviceProvider); [](http://example.com/codeflow?start=15&length=71)

Apply() is a hot path. 

Just do this once in the ctor. 
In fact... you could chain ctors and just save the tzi:

            public MinDateAndDateTimeAgg(IServiceProvider serviceProvider) 
  : this(serviceProvider.GetService<TimeZoneInfo>)
            {
            }

            public MinDateAndDateTimeAgg(TZI)
            {
                _tzi = tzi;
            }

src/libraries/Microsoft.PowerFx.Interpreter/Utility.cs
```diff
@@ -25,5 +25,14 @@ public static bool TryGetElementType(Type type, Type genericDef, out Type elemen
                 return false;
             }
         }
+
+        /// <summary>
+        /// Get a service from the <paramref name="serviceProvider"/>,
+        /// Returns null if not present.
+        /// </summary>
+        public static T GetService<T>(IServiceProvider serviceProvider)
```
MikeStall, line 33: >IServiceProvider [](http://example.com/codeflow?start=38&length=16)

Make extension method. 

then callers can just go
serviceProvider.GetService<T>()

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryOperators.cs
```diff
@@ -609,16 +614,17 @@ private static FormulaValue AddDateTimeAndDay(EvalVisitor runner, EvalVisitorCon
...
-        private static FormulaValue DateDifference(IRContext irContext, FormulaValue[] args)
+        private static FormulaValue DateDifference(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
         {
+            var timeZoneInfo = runner.TimeZoneInfo;
             DateTime arg0;
             switch (args[0])
             {
                 case DateTimeValue dtv:
-                    arg0 = dtv.Value;
+                    arg0 = dtv.GetConvertedValue(timeZoneInfo);
```
MikeStall, line 624: >GetConvertedValue [](http://example.com/codeflow?start=31&length=17)

Can we apply the helper to cases of GetConvertedValue in this file too?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
jas-valgotar: updated.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -282,9 +289,16 @@ private class MaxDateAndDateTimeAgg : IAggregator
         {
             protected DateTime _maxValueDT = DateTime.MinValue;
             protected int _count = 0;
+            private readonly IServiceProvider _serviceProvider;
+
+            public MaxDateAndDateTimeAgg(IServiceProvider serviceProvider)
```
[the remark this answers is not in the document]
jas-valgotar: updated

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -257,7 +264,14 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
             {
                 if (FunctionImplementations.TryGetValue(func, out var ptr))
                 {
-                    result = await ptr(this, context.IncrementStackDepthCounter(childContext), node.IRContext, args);
+                    try
+                    {
+                        result = await ptr(this, context.IncrementStackDepthCounter(childContext), node.IRContext, args);
```
MikeStall, line 269: >node.IRContext [](http://example.com/codeflow?start=99&length=14)

Catch knows the irContext, so we can apply here.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
jas-valgotar: update

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -507,24 +530,6 @@ public override async ValueTask<FormulaValue> Visit(AggregateCoercionNode node,
...
-
-                foreach (var coercion in node.FieldCoercions)
-                {
-                    CheckCancel();
-
-                    var newValue = await coercion.Value.Accept(this, context.NewScope(newScope));
-                    var name = coercion.Key;
-
-                    fields.Add(new NamedValue(name.Value, newValue));
-                }
```
MikeStall: >GetNormalizedDateTime2 [](http://example.com/codeflow?start=24&length=22)

nit: rename - shouldn't have 2 suffix.

GetNormalizedDateTimeAllowTimeValue

MikeStall (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -256,12 +257,12 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
                     if (formatString != null && hasNumberFmt)
                     {
                         // It's a datetime, formatted as number. Let's convert it to a number value first
-                        var newNumber = Library.DateTimeToNumber(IRContext.NotInSource(FormulaType.Number), new DateTimeValue[] { dateTimeValue });
+                        var newNumber = Library.DateTimeToNumber(runner, context, IRContext.NotInSource(FormulaType.Number), new DateTimeValue[] { dateTimeValue });
                         resultString = newNumber.Value.ToString(formatString ?? "g", culture);
```
xieofxie, line 261: Should we change this to 'G'? Because ToString() default is 'G'
https://learn.microsoft.com/en-us/dotnet/api/system.datetime.tostring?view=net-7.0

xieofxie: Also for the `format == null` below, we should also check if empty

MikeStall (approved):

PR #1042, opened 2023-01-30
Fix bug, add test case for check.GetEvaluator()

MikeStall:
Fix bug, add test case for check.GetEvaluator()

Demote AllSymbols  from internal property to private field to protect against misuse.


src/libraries/Microsoft.PowerFx.Interpreter/ParsedExpression.cs
```diff
@@ -84,7 +83,7 @@ internal static IExpressionEvaluator GetEvaluator(this CheckResult result, Stack
             var expr = new ParsedExpression(irResult.TopNode, irResult.RuleScopeSymbol, stackMarker, result.CultureInfo)
             {
                 _globals = globals,
-                _allSymbols = allSymbols,
+                _allSymbols = result.Symbols,
```
MikeStall, line 86: This was the bug. Calling Symbols instead of AllSymbols would have caught it.  Hence demoting allSymbols to a private field to avoid misuse.

LucGenetier:
✅ No public API change.

PR #1044, opened 2023-01-31
Handling recordtorecord coercion

anderson-joyle:
Called when patching a record with different types of fields.

Example:
`With({t1:Table({a:stringVar})},Patch(t1,First(t1),{a:5}))`

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
@@ -508,6 +508,24 @@ namespace Microsoft.PowerFx
 
                 return new InMemoryTableValue(node.IRContext, resultRows);
             }
+            else if (node.Op == UnaryOpKind.RecordToRecord)
+            {
+                var fields = new List<NamedValue>();
+                var scopeContext = context.SymbolContext.WithScope(node.Scope);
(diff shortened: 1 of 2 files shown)


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -507,6 +507,24 @@ public override async ValueTask<FormulaValue> Visit(AggregateCoercionNode node,
...
+            {
+                var fields = new List<NamedValue>();
+                var scopeContext = context.SymbolContext.WithScope(node.Scope);
+                var newScope = scopeContext.WithScopeValues((RecordValue)arg1);
+
+                foreach (var coercion in node.FieldCoercions)
+                {
+                    CheckCancel();
+
+                    var newValue = await coercion.Value.Accept(this, context.NewScope(newScope));
```
MikeStall, line 521: Do we need a new scope? Can this just be context?

anderson-joyle: Context doesn't contain all necessary record type scopes.

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1046, opened 2023-01-31
Remove task.whenall and change to serial logic. Disable related async…

nguhoa:
Remove task.whenall and change to serial logic.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
@@ -1981,14 +1981,19 @@ namespace Microsoft.PowerFx.Functions
 
         // ForAll([1,2,3,4,5], Value * Value)
         public static async ValueTask<FormulaValue> ForAll(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
-        {// Streaming 
+        {
+            // Streaming 
             var arg0 = (TableValue)args[0];
             var arg1 = (LambdaFormulaValue)args[1];
 
             var rowsAsync = LazyForAll(runner, context, arg0.Rows, arg1);
 
-            // TODO: verify semantics in the case of heterogeneous record lists
-            var rows = await Task.WhenAll(rowsAsync);
+            var rows = new List<FormulaValue>();
+
+            foreach (var row in rowsAsync)
+            {
+                rows.Add(await row);
+            }
 
             var errorRows = rows.OfType<ErrorValue>();
             if (errorRows.Any())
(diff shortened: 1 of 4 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1981,22 +1981,27 @@ public static async ValueTask<FormulaValue> Switch(EvalVisitor runner, EvalVisit
...
+            }
 
             var errorRows = rows.OfType<ErrorValue>();
             if (errorRows.Any())
             {
                 return ErrorValue.Combine(irContext, errorRows);
             }
 
-            return new InMemoryTableValue(irContext, StandardTableNodeRecords(irContext, rows, forceSingleColumn: false));
+            return new InMemoryTableValue(irContext, StandardTableNodeRecords(irContext, rows.ToArray(), forceSingleColumn: false));
```
MikeStall, line 2004: >ToArray() [](http://example.com/codeflow?start=94&length=9)

Why ToArray?

nguhoa: because input of StandardTableNodeRecords required array instead of list that I have.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1981,22 +1981,27 @@ public static async ValueTask<FormulaValue> Switch(EvalVisitor runner, EvalVisit
 
         // ForAll([1,2,3,4,5], Value * Value)
         public static async ValueTask<FormulaValue> ForAll(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
```
LucGenetier, line 1983: This API is missing a CancellationToken
Shouldn't we start from there and ensure all inner tasks use it ?

MikeStall: The token is on runner.CheckCancel(); 
But we should be checking that inside the for-loop (see other uses in that file). 
Ideally, we'd have some form of static analysis to verify this,

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -415,8 +415,13 @@ public static async ValueTask<FormulaValue> DistinctTable(EvalVisitor runner, Ev
             var arg1 = (LambdaFormulaValue)args[1];
 
             var values = arg0.Rows.Select(row => ApplyLambda(runner, context, row, arg1));
-            var pairResults = await Task.WhenAll(values);
-            var pairs = pairResults.ToList();
+            
+            var pairs = new List<(DValue<RecordValue> row, FormulaValue distinctValue)>();
+
+            foreach (var pair in values)
```
MikeStall, line 421: >foreach [](http://example.com/codeflow?start=12&length=7)

check cancel token

nguhoa: Updated.

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1034, opened 2023-01-26
Add flag to treat enums as strongly-typed option sets

lesaltzm:
This draft adds support for strong-typing of built in enums. In the process, it refactors Option Sets to declare what the backing value type is. 

I'll be splitting this up into a Refactor PR and then a Feature PR, since this change wound up a little larger than intended.

PR #1047, opened 2023-01-31
All CheckTypes take a CheckTypesContext

lesaltzm:
Extracted from #1034. 

We want the base CheckTypes impl to be able to access Feature flags, and so we're standardizing all CheckTypes to include this argument.


CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1051, opened 2023-02-01
make OpenApiHelper, OpenApiParser public

xieofxie:


src/libraries/Microsoft.PowerFx.Connectors/Execution/OpenApiJsonSerializer.cs
```diff
@@ -130,8 +130,8 @@ protected virtual void Dispose(bool disposing)
             {
                 if (disposing)
                 {
+                    _writer?.Dispose();
```
xieofxie, line 133: dispose writer first then stream because writer dispose will flush stream which will cause already disposed exception.

src/libraries/Microsoft.PowerFx.Connectors/OpenApiHelperFunctions.cs
```diff
@@ -8,12 +8,12 @@
...
 namespace Microsoft.PowerFx.Connectors
 {
-    internal class OpenApiHelperFunctions
+    public class OpenApiHelperFunctions
     {
         private static readonly Regex NotAnXsdNCNameCharRegex = new Regex(@"[^a-zA-Z0-9_-]+", RegexOptions.Compiled);
         private static readonly Regex XsdNCNameStartCharRegex = new Regex(@"^[a-zA-Z_]", RegexOptions.Compiled);
 
-        internal static string NormalizeOperationId(string operationId)
+        public static string NormalizeOperationId(string operationId)
```
xieofxie, line 16: If it is required, we need to make it public so we also call it to match the function.name.

MikeStall (dismissed):
:shipit:

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1033, opened 2023-01-26
Support data types from Dataverse columns.

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx-Dataverse/issues/56

PR #1055, opened 2023-02-02
HyperlinkValue implementation to handle DType.Hypelink type

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx-Dataverse/issues/59
To compare against https://github.com/microsoft/Power-Fx/pull/1033

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
@@ -1916,7 +1916,10 @@ namespace Microsoft.PowerFx.Core.Types
                 case DKind.String:
                     accepts =
                         type.Kind == Kind ||
-                        type.Kind == DKind.Hyperlink ||
+
+                        // !!!???
+                        //type.Kind == DKind.Hyperlink || 
+
                         type.Kind == DKind.Image ||
                         type.Kind == DKind.PenImage ||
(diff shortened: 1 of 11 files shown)


src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1916,7 +1916,10 @@ public virtual bool Accepts(DType type, out KeyValuePair<string, DType> schemaDi
                 case DKind.String:
                     accepts =
                         type.Kind == Kind ||
-                        type.Kind == DKind.Hyperlink ||
+
+                        // !!!???
+                        //type.Kind == DKind.Hyperlink ||
```
MikeStall, line 1921: This will break PA because they don't go through IR. 
Can we have IR inject a coercion regardless of Accepts?

MikeStall:
>         public static CoercionKind GetCoercionKind(DType fromType, DType toType)

wondering:
instead of changing DType (risky to PAClient), can we just add the hyperink coercion check here?  (in the IR level, which is not used by PA yet). 


---
Refers to: src/libraries/Microsoft.PowerFx.Core/IR/CoercionMatrix.cs:11 in a95a91e. [](commit_id = a95a91eca8765eff70fd8b35d4fe02a619e1826e, deletion_comment = False)

PR #957, opened 2023-01-04
Removes STJ from interpreter and core.

jas-valgotar:

PR #1056, opened 2023-02-02
We can remove Sytem.Memory reference

MikeStall:
This was only used by System.Text.Json, which was removed in #957 .


lesaltzm (approved):

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #1058, opened 2023-02-03
Lazy loading of types for symbol tables 

### LINKED ISSUE #1017: SymbolTables lazy loading symbol definitions
For dataverse, we can have 1000 tables. 
We can quickly load their names; but we can't load all their type information. 

Allow SymbolTable to be lazy so that we can populate with names, and then lazily fetch additional type info as needed.

MikeStall:
Fix #1017 
Useful when names are known, but getting the list types is expensive -  common for dataverse tables. We can quickly get all tables in the org, but getting the table types requires EntityMetadata which is expensive.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/DeferredSymbolTable.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/DeferredSymbolTable.cs
@@ -55,8 +55,7 @@ namespace Microsoft.PowerFx
 
         private readonly SlotMap<NameLookupInfo?> _slots = new SlotMap<NameLookupInfo?>();
 
-        // Called by 
-        // $$$ OVerride full
+        // Override lookup.
         bool INameResolver.Lookup(DName name, out NameLookupInfo nameInfo, NameLookupPreferences preferences)
         {
             if (_displayNameLookup.TryGetLogicalOrDisplayName(name, out var logical, out var display))
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/DeferredSymbolTable.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/DeferredSymbolTable.cs
@@ -1,0 +1,75 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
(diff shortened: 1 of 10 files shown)


src/libraries/Microsoft.PowerFx.Core/Public/Config/DeferredSymbolTable.cs
```diff
@@ -0,0 +1,166 @@
...
+
+namespace Microsoft.PowerFx
+{
+    /// <summary>
+    /// Provides symbol table with deferred semantics. 
+    /// This is semantically readonly - there is a fixes set of known symbols, just the type information is populated lazily.
+    /// Good when we can quickly load the names (to populate the intellisense completion list), but then load remaining symbol details on-demand. 
+    /// </summary>
+    [DebuggerDisplay("{DebugName}")]
+    internal class DeferredSymbolTable : ReadOnlySymbolTable, IGlobalSymbolNameResolver, INameResolver
```
MikeStall, line 25: This is the key new class. 
It's internal - it's created via static helpers on ReadOnlySymbolTable, like the other derived symbols.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Config/DeferredSymbolTable.cs
```diff
@@ -0,0 +1,165 @@
...
+                    {
+                        return true;
+                    }
+                }
+
+                // Callback to host - do not hold lock while invoking this. 
+                // Host may do arbitrary operations, including network calls to fetch metadata.
+                var type = _fetchTypeInfo(logical.Value, display.Value);
+
+                lock (_lock)
```
jas-valgotar, line 76: >lock (_lock) [](http://example.com/codeflow?start=16&length=12)

Why not put lock inside method `AddUnderLock()`?

jas-valgotar (approved):

src/libraries/Microsoft.PowerFx.Core/Public/Config/DeferredSymbolTable.cs
MikeStall: It's a style thing - I wanted all the lock/unlock to be at the same level so that it was clear when it was being taken and released. It's especially important that we don't hold the lock while invoking callbacks (and virtuals implemented by the host are also callbacks).

Issue #1045, opened 2023-01-31
Type coercion in host


babrekel:
Power FX is able to convert various values between types. Examples:

```
Value("5") -> 5
Value("true") -> Error
Boolean(1) -> true
Boolean(0) -> false
Boolean(100) -> true
Boolean("yes") -> Error
Boolean("true") -> true
// And so forth...
``` 

Working with external data is a common paradigm for users of PVA. This includes parsing API responses from HTTP calls, parsing custom events / messages sent from the client, etc. Users of PVA can leverage the "Parse Value" node to select a type from Sample Data and convert an `UntypedObject` or `String` value to the target type.

To make this scenario work, a host API is requested to do appropriate conversions to the desired type. For example, if I have a `FormulaValue` that represents the number `1`, and the desired `FormulaType` is `Boolean`, the output would be `true`. Potential method signature:

```
FormulaValue TryCoerceTo(FormulaValue input, FormulaType targetType);
```

This can be used in data connectors to process things like `DateTime` from API responses.

MikeStall:
Some more notes:
1. This is just a hosting API and has no impact on language semantics
2. This should work recursively with records.  Errors are granular per-field. 
3.  This should just call existing ctor functions. So BooleanType calls Boolean() function, Numbertype calls Value() function
4.  Can we make this consistent with the coercion rules for untyped object.

nguhoa:
Issue is fixed in https://github.com/microsoft/Power-Fx/pull/1062

PR #1062, opened 2023-02-06
Add TryCoerceTo to convert FormulaValue to target type

nguhoa:
Fix the issue: https://github.com/microsoft/Power-Fx/issues/1045
Add TryCoerceTo to convert FormulaValue to target type.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Extensions.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Extensions.cs
@@ -9,8 +9,7 @@ namespace Microsoft.PowerFx.Functions
     {
         internal static bool IsValid(this DateTime dateTime, EvalVisitor runner)
         {
-            var tzi = runner.GetService<TimeZoneInfo>() ?? TimeZoneInfo.Local;
-            return IsValid(dateTime, tzi);
+            return IsValid(dateTime, runner.TimeZoneInfo);
         }
 
         internal static bool IsValid(this DateTime dateTime, TimeZoneInfo tzi)
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Extensions.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Extensions.cs
@@ -10,7 +10,11 @@ namespace Microsoft.PowerFx.Functions
         internal static bool IsValid(this DateTime dateTime, EvalVisitor runner)
         {
             var tzi = runner.GetService<TimeZoneInfo>() ?? TimeZoneInfo.Local;
+            return IsValid(dateTime, tzi);
+        }
 
+        internal static bool IsValid(this DateTime dateTime, TimeZoneInfo tzi)
+        {
             // If DateTime is UTC, the time is always valid
             if (dateTime.Kind == DateTimeKind.Utc)
             {
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
@@ -45,6 +45,16 @@ namespace Microsoft.PowerFx.Functions
 
         public static readonly IReadOnlyDictionary<TexlFunction, AsyncFunctionPtr> FunctionImplementations;
 
+        public static FormattingInfo CreateFormattingInfo(EvalVisitor runner)
+        {
+            return new FormattingInfo()
+            {
+                CultureInfo = runner.CultureInfo,
+                CancellationToken = runner.CancellationToken,
+                TimeZoneInfo = runner.GetService<TimeZoneInfo>() ?? TimeZoneInfo.Local
+            };
+        }
+
         static Library()
         {
             var allFunctions = new Dictionary<TexlFunction, AsyncFunctionPtr>();
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
@@ -900,8 +900,7 @@ namespace Microsoft.PowerFx.Functions
                     BuiltinFunctionsCore.Left.Name,
                     expandArguments: NoArgExpansion,
                     replaceBlankValues: ReplaceBlankWith(
-                        new StringValue(IRContext.NotInSource(FormulaType.String), string.Empty),
-                        new NumberValue(IRContext.NotInSource(FormulaType.Number), 0)),
+                        new StringValue(IRContext.NotInSource(FormulaType.String), string.Empty)),
                     checkRuntimeTypes: ExactSequence(
                         ExactValueType<StringValue>,
                         ExactValueType<NumberValue>),
@@ -1172,8 +1171,7 @@ namespace Microsoft.PowerFx.Functions
                     BuiltinFunctionsCore.Right.Name,
                     expandArguments: NoArgExpansion,
                     replaceBlankValues: ReplaceBlankWith(
-                        new StringValue(IRContext.NotInSource(FormulaType.String), string.Empty),
-                        new NumberValue(IRContext.NotInSource(FormulaType.Number), 0)),
+                        new StringValue(IRContext.NotInSource(FormulaType.String), string.Empty)),
                     checkRuntimeTypes: ExactSequence(
                         ExactValueType<StringValue>,
                         ExactValueType<NumberValue>),
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
@@ -631,6 +631,32 @@ namespace Microsoft.PowerFx.Functions
                     targetFunction: FirstN)
             },
             {
+                BuiltinFunctionsCore.First_UO,
+                StandardErrorHandling<UntypedObjectValue>(
+                    BuiltinFunctionsCore.First_UO.Name,
+                    expandArguments: NoArgExpansion,
+                    replaceBlankValues: DoNotReplaceBlank,
+                    checkRuntimeTypes: ExactValueTypeOrBlank<UntypedObjectValue>,
+                    checkRuntimeValues: UntypedObjectArrayChecker,
+                    returnBehavior: ReturnBehavior.ReturnBlankIfAnyArgIsBlank,
+                    targetFunction: First_UO)
+            },
+            {
+                BuiltinFunctionsCore.FirstN_UO,
+                StandardErrorHandling<FormulaValue>(
+                    BuiltinFunctionsCore.FirstN.Name,
+                    expandArguments: NoArgExpansion,
+                    replaceBlankValues: ReplaceBlankWithZeroForSpecificIndices(1),
+                    checkRuntimeTypes: ExactSequence(
+                        ExactValueTypeOrBlank<UntypedObjectValue>,
+                        ExactValueTypeOrBlank<NumberValue>),
+                    checkRuntimeValues: ExactSequence(
+                        UntypedObjectArrayChecker,
+                        DeferRuntimeValueChecking),
+                    returnBehavior: ReturnBehavior.ReturnBlankIfAnyArgIsBlank,
+                    targetFunction: FirstN_UO)
+            },
+            {
                 BuiltinFunctionsCore.ForAll,
                 StandardErrorHandlingAsync<FormulaValue>(
                     BuiltinFunctionsCore.ForAll.Name,
@@ -831,6 +857,32 @@ namespace Microsoft.PowerFx.Functions
                     checkRuntimeValues: DeferRuntimeValueChecking,
                     returnBehavior: ReturnBehavior.AlwaysEvaluateAndReturnResult,
                     targetFunction: LastN)
+            },
+            {
+                BuiltinFunctionsCore.Last_UO,
+                StandardErrorHandling<UntypedObjectValue>(
+                    BuiltinFunctionsCore.Last_UO.Name,
+                    expandArguments: NoArgExpansion,
+                    replaceBlankValues: DoNotReplaceBlank,
+                    checkRuntimeTypes: ExactValueTypeOrBlank<UntypedObjectValue>,
+                    checkRuntimeValues: UntypedObjectArrayChecker,
+                    returnBehavior: ReturnBehavior.ReturnBlankIfAnyArgIsBlank,
+                    targetFunction: Last_UO)
+            },
+            {
+                BuiltinFunctionsCore.LastN_UO,
+                StandardErrorHandling<FormulaValue>(
+                    BuiltinFunctionsCore.LastN.Name,
+                    expandArguments: NoArgExpansion,
+                    replaceBlankValues: ReplaceBlankWithZeroForSpecificIndices(1),
+                    checkRuntimeTypes: ExactSequence(
+                        ExactValueTypeOrBlank<UntypedObjectValue>,
+                        ExactValueTypeOrBlank<NumberValue>),
+                    checkRuntimeValues: ExactSequence(
+                        UntypedObjectArrayChecker,
+                        DeferRuntimeValueChecking),
+                    returnBehavior: ReturnBehavior.ReturnBlankIfAnyArgIsBlank,
+                    targetFunction: LastN_UO)
             },
             {
                 BuiltinFunctionsCore.Left,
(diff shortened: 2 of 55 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs
```diff
@@ -0,0 +1,88 @@
...
+                result = Value(formatInfo, irContext, new FormulaValue[] { value });
+            }
+            else if (targetType == FormulaType.DateTime)
+            {
+                result = GetDateTime(formatInfo, irContext, value);
+            }
+            else
+            {
+                result = CommonErrors.NotYetImplementedError(irContext, $"Do not support {targetType.GetType().Name} Type.");
+                return false;
```
MikeStall, line 77: For TryGet* methods, if we return false, we can also return result = null

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -45,6 +47,15 @@ private static T GetService<T>(this IServiceProvider services, T defaultService)
 
         public static readonly IReadOnlyDictionary<TexlFunction, AsyncFunctionPtr> FunctionImplementations;
 
+        internal class FormattingInfo
```
MikeStall, line 50: >internal class FormattingInfo [](http://example.com/codeflow?start=8&length=29)

nit - we don't need to nest this class.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
```diff
@@ -3,11 +3,17 @@
...
 using System.Threading.Tasks;
+using System.Xml;
 using Microsoft.PowerFx.Core.IR;
 using Microsoft.PowerFx.Core.IR.Nodes;
+using Microsoft.PowerFx.Core.Texl;
+using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Interpreter;
+using Microsoft.PowerFx.Interpreter.Exceptions;
 using Microsoft.PowerFx.Types;
+using static Microsoft.PowerFx.Syntax.PrettyPrintVisitor;
```
MikeStall, line 16: Why do we need this?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
```diff
@@ -335,6 +341,30 @@ public static FormulaValue TextToBoolean(IRContext irContext, StringValue[] args
...
+            switch (value)
+            {
+                case StringValue sv:
+                    result = TextToBoolean(irContext, new StringValue[] { sv });
+                    break;
+                case NumberValue num:
+                    result = NumberToBoolean(irContext, new NumberValue[] { num });
+                    break;
+                case BooleanValue boolVal:
+                    result = new BooleanValue(irContext, boolVal.Value);
```
MikeStall, line 356: >new BooleanValue(irContext, boolVal.Value); [](http://example.com/codeflow?start=29&length=43)

nit - can this jus be:
result = boolVal;

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
```diff
@@ -335,6 +341,30 @@ public static FormulaValue TextToBoolean(IRContext irContext, StringValue[] args
             return CommonErrors.InvalidBooleanFormatError(irContext);
         }
 
+        public static FormulaValue GetBoolean(IRContext irContext, FormulaValue value)
```
MikeStall, line 344: >FormulaValue [](http://example.com/codeflow?start=22&length=12)

Nit - can this be stronger typed, BooleanValue

MikeStall: I see this was following the existing pattern - the difference is that:
- for returning errors back to *power fx expressions*, we use ErrorValue (which CommonErrors produces). These can be caught and handled within the expression. 
- but for returning errors back to the host, we either throw exceptions (if it's a host bug) , or return false in a TryGet pattern where host can handle it. 

Since this is taking a runtime function and splitting it into 2 layers (runtime and a host ) , there's a bit of a change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/TypeCoercionTest.cs
```diff
@@ -0,0 +1,218 @@
...
+    {
+        [Fact]
+        public void TryCoerce()
+        {
+            TryCoerceToSuccess(null, FormulaType.String, FormulaValue.NewBlank(FormulaType.String));
+            TryCoerceToFailure(new StringValue(IRContext.NotInSource(FormulaType.String), "TestLink"), FormulaType.Hyperlink);
+        }
+
+        [Fact]
+        public void TryCoerceToBoolean()
```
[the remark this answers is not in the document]
MikeStall: Actuallly, maybe there's a way to combine tests - we're ultimately testing against a matrix. 

What if we had a single row that tested multiple targets, like: 

void TestCoerce(string expr, string exprBool, string exprNumber, string exprStr, string exprDate)

and null expr value indicates an error. 

[InlineData("1", "true", "1" "\"1\"", null)] , etc

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
nguhoa: you right, i go around here. updated.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
nguhoa: i dont need this. it is added when i did something previously. removed this.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
nguhoa: what is your suggestion here to create object for formattinginfo?

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs
nguhoa: updated to set result is null here.

src/tests/Microsoft.PowerFx.Interpreter.Tests/TypeCoercionTest.cs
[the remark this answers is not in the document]
nguhoa: Updated the test.

src/tests/Microsoft.PowerFx.Interpreter.Tests/TypeCoercionTest.cs
```diff
@@ -0,0 +1,101 @@
...
+        [InlineData(1, "true", "1", "1", "12/31/1899 12:00 AM")]
+        [InlineData(0, "false", "0", "0", "12/30/1899 12:00 AM")]
+        [InlineData(44962, "true", "44962", "44962", "2/5/2023 12:00 AM")]
+        public void TryCoerceFromNumberTest(double value, string exprBool, string exprNumber, string exprStr, string exprDateTime)
+        {
+            TryCoerceToTargetTypes(FormulaValue.New(value), exprBool, exprNumber, exprStr, exprDateTime);
+        }
+
+        // From string to other types
+        [Theory]
```
MikeStall, line 26: Does "True"  (capital t) coerce to true? (we should do whatever Boolean() does)

src/tests/Microsoft.PowerFx.Interpreter.Tests/TypeCoercionTest.cs
```diff
@@ -0,0 +1,101 @@
...
+        private void TryCoerceToTargetTypes(FormulaValue inputValue, string exprBool, string exprNumber, string exprStr, string exprDateTime)
+        {
+            bool isSucceeded = TypeCoercionProvider.TryCoerceTo(inputValue, out BooleanValue resultBoolean);
+            if (exprBool != null)
+            {
+                Assert.True(isSucceeded);
+                Assert.Equal(bool.Parse(exprBool), resultBoolean.Value);
+            }
+            else
+            {
```
MikeStall, line 63: nit - on false case, also assert resultBoolean is null

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
```diff
@@ -332,24 +354,74 @@ public static FormulaValue TextToBoolean(IRContext irContext, StringValue[] args
                 return new BooleanValue(irContext, false);
             }
 
-            return CommonErrors.InvalidBooleanFormatError(irContext);
+            throw new Exception("Invalid Boolean Format.");
```
MikeStall, line 357: >throw new Exception("Invalid Boolean Format."); [](http://example.com/codeflow?start=12&length=47)

we should avoid exceptions for control flow like this. 
Instead, use a `bool Try*(out value)` pattern and return false.

nguhoa: yes, will update.

src/tests/Microsoft.PowerFx.Interpreter.Tests/TypeCoercionTest.cs
nguhoa: will add that test.

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs
```diff
@@ -0,0 +1,111 @@
...
+using Microsoft.PowerFx.Functions;
+using Microsoft.PowerFx.Types;
+using static Microsoft.PowerFx.Functions.Library;
+
+namespace Microsoft.PowerFx.Interpreter.Marshal
+{
+    /// <summary>
+    /// Convert value to target format.
+    /// </summary>
+    internal static class TypeCoercionProvider
```
MikeStall, line 17: >TypeCoercionProvider [](http://example.com/codeflow?start=26&length=20)

We can expose this publicly on FormulaValue via extensions method:

        public static bool TryCoerceTo(this FormulaValue fv, FormulaType targetType, out FormulaValue newValue)
        {
            // calls your impl 
        }

nguhoa: updated this.

MikeStall: The class needs to be public for this to be publicly callable. 

And it should be in the root namespace (ignore the file path)
 Microsoft.PowerFx

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -305,7 +342,7 @@ internal static FormulaValue ExpandDateTimeExcelFormatSpecifiers(IRContext irCon
                     }
                     catch (FormatException)
                     {
-                        return CommonErrors.GenericInvalidArgument(irContext, StringResources.Get(TexlStrings.ErrTextInvalidFormat, culture.Name));
+                        return null;
```
MikeStall, line 345: >return null; [](http://example.com/codeflow?start=24&length=12)

Can we follow TryGet pattern , and avoid returning null. 
That makes it very explicit.

nguhoa: Updated this.

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs
nguhoa: Updated

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Extensions.cs
```diff
@@ -10,7 +10,11 @@ internal static class Extensions
         internal static bool IsValid(this DateTime dateTime, EvalVisitor runner)
         {
             var tzi = runner.GetService<TimeZoneInfo>() ?? TimeZoneInfo.Local;
```
jas-valgotar, line 12: >runner.GetService<TimeZoneInfo>() ?? TimeZoneInfo.Local [](http://example.com/codeflow?start=22&length=55)

I think we can use runner.TimeZoneInfo now here.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -45,6 +45,16 @@ private static T GetService<T>(this IServiceProvider services, T defaultService)
 
         public static readonly IReadOnlyDictionary<TexlFunction, AsyncFunctionPtr> FunctionImplementations;
 
+        public static FormattingInfo CreateFormattingInfo(EvalVisitor runner)
+        {
+            return new FormattingInfo()
+            {
+                CultureInfo = runner.CultureInfo,
+                CancellationToken = runner.CancellationToken,
+                TimeZoneInfo = runner.GetService<TimeZoneInfo>() ?? TimeZoneInfo.Local
```
jas-valgotar, line 54: >runner.GetService<TimeZoneInfo>() ?? TimeZoneInfo.Local [](http://example.com/codeflow?start=31&length=55)

Can we use, runner.TimeZoneInfo?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
```diff
@@ -561,12 +566,32 @@ private static bool TryGetCulture(string name, out CultureInfo value)
             }
         }
 
+        public static bool TryDateTimeParse(FormattingInfo formatInfo, IRContext irContext, StringValue value, out DateTimeValue result)
```
jas-valgotar, line 569: >TryDateTimeParse [](http://example.com/codeflow?start=27&length=16)

this would change as per latest fetch...

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs
```diff
@@ -0,0 +1,109 @@
...
+using static Microsoft.PowerFx.Functions.Library;
+
+namespace Microsoft.PowerFx
+{
+    /// <summary>
+    /// Convert value to target format.
+    /// </summary>
+    public static class TypeCoercionProvider
+    {
+        public static FormattingInfo CreateFormattingInfo()
```
MikeStall, line 19: >public [](http://example.com/codeflow?start=8&length=6)

We can make this internal.
Then that lets us leave FormattingInfo class as internal.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Extensions.cs
nguhoa: updated this.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
nguhoa: update this as well.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
```diff
@@ -309,47 +310,99 @@ private static string PowerFxBooleanToString(bool b)
...
+                case BooleanValue boolVal:
+                    result = boolVal;
+                    break;
             }
+            
+            return result != null;
+        }
 
-            return CommonErrors.InvalidBooleanFormatError(irContext);
+        public static DateTime GetNormalizedDateTime(FormulaValue value, TimeZoneInfo timeZoneInfo)
```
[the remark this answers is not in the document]
nguhoa: yes, update evalvisitor.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
nguhoa: fetch main brn.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs
nguhoa: udpated.

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

PR #1063, opened 2023-02-06
Excel Compat: Remove ConsistentOneColumnTableResult flag

Ananya-Gupta24:
As part of the excel compatibility work, all tabular overloads of functions will return "Value" as the output column name instead of the varying names returned currently ("Result" or input column name). This PR removes the flag and enables this behavior by default. Test updates included. 
This change will go in as part of a PowerFx version update to be added in this PAClient [PR](https://msazure.visualstudio.com/OneAgile/_git/PowerApps-Client/pullrequest/7341719) to ensure that there is no breakage.


Ananya-Gupta24:
Won't be checked in till [PAClient PR](https://msazure.visualstudio.com/OneAgile/_git/PowerApps-Client/pullrequest/7341719) is ready

LucGenetier:
✅ No public API change.

lesaltzm:
We can't remove flags that have back compat impact. Discussed offline.

lesaltzm (changes requested):
See above, marking as request change to  ensure this is never checked in.

PR #1064, opened 2023-02-06
Updated multiarg tabular function handler to return first error or blank tabular arg

CarlosFigueiraMSFT:
Currently the implementation of the tabular function handler needs to make two passes over the function arguments to determine whether a blank or error value will be returned (if passed in a tabular context). This is inefficient, so we are updating the implementation to return the first of those values.

This change also simplifies the implementation of the tabular handler, to avoid copying the arguments multiple times.

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/TabularOverloadsBlanksAndErrors.txt
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/TabularOverloadsBlanksAndErrors.txt
@@ -241,7 +241,7 @@ Error({Kind:ErrorKind.Numeric})
 Error({Kind:ErrorKind.Div0})
 
 >> Concatenate( If(Sqrt(-1)<0,["Hi", "hello"]), If(1/0<2, [" world", "people"]) )
-Error(Table({Kind:ErrorKind.Numeric}))
+Error({Kind:ErrorKind.Numeric})
 
 >> Concatenate(Filter([1,2], Value<>Value), If(1/0<0, "test"))
 Table()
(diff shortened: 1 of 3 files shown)


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/TabularOverloadsBlanksAndErrors.txt
```diff
@@ -241,7 +241,7 @@ Error({Kind:ErrorKind.Numeric})
 Error({Kind:ErrorKind.Div0})
```
Shpakh-MSFT, line 241: >ind:ErrorKind.Div0 [](http://example.com/codeflow?start=8&length=18)

Why is this not {Kind:ErrorKind.Numeric}?

CarlosFigueiraMSFT: The first error is on a scalar argument - so it is supposed to be distributed across the tabular arguments. The second one is a tabular argument, and since this is an error, this is what it is returned.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/TabularOverloadsBlanksAndErrors.txt
Shpakh-MSFT: This works differently in Power Apps though. Is it something we haven't fixed in PA yet?

Shpakh-MSFT (approved):
:shipit:

PR #1067, opened 2023-02-07
Fix #958 - allow LSP to enforce expected types

### LINKED ISSUE #958: Specify expected type to Check/LSP
If I have the formula that is used on a property that should only accept **text** values, but the formula has **Record** or **Table** values, it does not show any error indication. 

For example: `ForAll([1,2,3], ThisRecord.Value)` formula does not show error in FXEditor

image

MikeStall:
Fix #958 - see for details. 

Also creates better symmetry between LSP's EditorContextScope and CheckResult.
This makes new CheckResult features (such as this one) available to LSP.


LucGenetier:
✅ No public API change.

anderson-joyle (approved):
:shipit:

PR #1036, opened 2023-01-26
Require column names as identifiers for SortByColumns()

VamsiModem:
Remove use of string literals for column names arguments in functions: SortByColumns()

PR #1069, opened 2023-02-08
Revert SortByColumns changes to use identifiers as column names

CarlosFigueiraMSFT:
The [PR 1036](https://github.com/microsoft/Power-Fx/pull/1036) caused a regression in the intellisense for the SortByColumns function. This PR reverts that change, and adds a test that validates the previous behavior.


LucGenetier:
✅ No public API change.

VamsiModem (approved):

Issue #984, opened 2023-01-09
Don't allow Sqrt(Now())


MikeStall:
This is a non-sensical expression and should be blocked. 

It works today because we automatically coerce times to numbers. 

`Today() + 1` is fine.   As is `Date1-Date2`. And the thing is for these, we already have overloading. We have DateAdd and NumericalAdd, so intent is clear, and we can even skip the date-to-number coercion.

  But what does `SQRT(Today())` mean? Or `Now() * 2` Or `Now() > 10` Or  `IF(NOW(),1,2)`

PR #1071, opened 2023-02-08
Allow parameter coercion in the FirstN/LastN functions

CarlosFigueiraMSFT:
Functions with scalar arguments should support coercions. Currently FirstN/LastN don't, so this PR changes it.

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/First.txt
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/First.txt
@@ -82,9 +82,3 @@ Table({Value:1})
 
 >> LastN([1,2,3,4,5], true)
 Table({Value:5})
-
->> FirstN([1,2,3,4,5], Date(1900,1,1))
-Table({Value:1},{Value:2})
-
->> LastN([1,2,3,4,5], Date(1900,1,1))
-Table({Value:4},{Value:5})
--- a/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/First.txt
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/First.txt
@@ -69,3 +69,22 @@ Blank()
 
 >> FirstN(Filter([1,2,3],Value=4),2)
 Table()
+
+// ********** Coercions ************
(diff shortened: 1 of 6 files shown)


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/First.txt
```diff
@@ -69,3 +69,22 @@ Blank()
...
+>> LastN([1,2,3,4,5], "3")
+Table({Value:3},{Value:4},{Value:5})
+
+>> FirstN([1,2,3,4,5], true)
+Table({Value:1})
+
+>> LastN([1,2,3,4,5], true)
+Table({Value:5})
+
+>> FirstN([1,2,3,4,5], Date(1900,1,1))
```
MikeStall, line 86: I applaud the test , but this is so painful to see - I can't imagine a case where the customer ever intends to write this. Is there some way we can error or at least warn here? 
.  It's like #984.

CarlosFigueiraMSFT: Good point, this is indeed bizarre (albeit possible). I'll remove those tests, since we don't want to promote this usage. And we should have an internal discussion to decide what to do with it. On one hand, we want to have parity with Excel, and it does support that. On the other hand, it really doesn't make sense, so a warning would be a good compromise.

LucGenetier:
✅ No public API change.

gregli-msft:
OMG, I was literally about to ask you about this.  Not supporting coercions means needing to add overloads specifically for Decimal.  And I don't see a good reason why this should be different from functions like Mid and Left that do support coercion.

One question: can we do the Index function too?  Coercion is similarly blocked.

CarlosFigueiraMSFT:
Done. Updated both Index functions as well

---
In reply to: [1424650703](https://github.com/microsoft/Power-Fx/pull/1071#issuecomment-1424650703) [](http://example.com/codeflow?ancestors=1424650703,1424650703)

bryancroteau-MSFT (approved):

LucGenetier:
✅ No public API change.

PR #1059, opened 2023-02-03
Moves the blank value handling of function args to the IR

jas-valgotar:
Fixes #1006 
- This moves the Trivial BlankToZero argument handler from interpreter to IR Phase so other backends don't have to implement it.
- BlankToZero is wrapping arg in IR as `arg` => `Coalesce(arg, 0)`
- To Improve efficiency if arg was already wrapped in that fashion we don't wrap it again.
- This PR also improves Concatenate Operator as below
Before: `"1"&"2"&"3"...` => `Concatenate(Concatenate("1", "2"), "3"))...`
Now: `"1"&"2"&"3"...` => `Concatenate("1","2","3",...)`

NOTE:Other Preprocessing will be moved to IRPhase in followup PRs.

PR #1072, opened 2023-02-09
Adds truncate arg preprocessing in IR.

### LINKED ISSUE #808: More parameter coercions should be in IR
Right() function truncates its 2nd arg 

https://github.com/microsoft/Power-Fx/blob/d01d9b7baf81c7a64fc356329f83a38c55a497a6/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Right.txt#L6

Expected: "World". actual: "oWorld"

Currently, this is captured in the function's implementation - so each backend has to get it right.  
Instead, this should be captured in the IR. 

Ie, instead of an IR like:
Call(Right,  arg0: "HelloWorld", arg1: 5.8)
and then the impl of Right knows arg1 should be truncated.  (and some impls just get it wrong)

It should be an IR like:
Call(Right,  arg0: "HelloWorld", Truncate(arg1: 5.8))

_And then the impl of Right doesn't need to worry about this corner case._  So all backends will get it right. 
(TBD if Truncate is an existing operator or if we need to add a new one )

**Places where we can leverage this:**
- string functions like Right() that convert numeric argument to an interger. 
- Blank() to 0 
- Blank() to empty string .

This will require some sort of hook on the TexlFunction for translating the call node that the IR translator can use.

jas-valgotar:
Fixes #808 
Follow-up to #1059 
This moves arg truncating for `Right()` and `Left()` which truncates the 2nd arg.

Now. in IR `arg => Truc(Coalesce(arg, 0))`

NOTE: Truncate Preprocessor also means it would add BlankToZero preprocessing as well.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/ArgPreprocessor.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/ArgPreprocessor.cs
@@ -13,6 +13,6 @@ namespace Microsoft.PowerFx.Core.Functions
         ReplaceWithZero = 1,
         
         // This also includes ReplaceWithZero
-        Trurncate = 2,
+        Truncate = 2,
     }
 }
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/ArgPreprocessor.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/ArgPreprocessor.cs
@@ -11,5 +11,8 @@ namespace Microsoft.PowerFx.Core.Functions
     {
         None = 0,
         ReplaceWithZero = 1,
+        
+        // This also includes ReplaceWithZero
+        Trurncate = 2,
     }
 }
--- a/src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
@@ -361,7 +361,7 @@ namespace Microsoft.PowerFx.Core.IR
                         case ArgPreprocessor.ReplaceWithZero:
                             convertedNode = ReplaceBlankWithZero(args[i]);
                             break;
-                        case ArgPreprocessor.Trurncate:
+                        case ArgPreprocessor.Truncate:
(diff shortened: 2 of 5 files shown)


src/libraries/Microsoft.PowerFx.Core/Functions/ArgPreprocessor.cs
```diff
@@ -11,5 +11,8 @@ internal enum ArgPreprocessor
     {
         None = 0,
         ReplaceWithZero = 1,
+        
+        // This also includes ReplaceWithZero
+        Trurncate = 2,
```
MikeStall, line 16: >Trurncate [](http://example.com/codeflow?start=8&length=9)

nit: typo
Truncate

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -389,6 +392,16 @@ private static IntermediateNode ReplaceBlankWithZero(IntermediateNode arg)
                 return convertedNode;
             }
 
+            /// <summary>
+            /// Wraps node arg => Truc(Coalesce(arg , 0)).
+            /// </summary>
+            private static IntermediateNode TruncatePreProcessor(IntermediateNode arg)
+            {
+                var blankToZeroNode = ReplaceBlankWithZero(arg);
+                var truncateNode = new CallNode(blankToZeroNode.IRContext, BuiltinFunctionsCore.Trunc, blankToZeroNode);
```
MikeStall, line 401: >Trunc [](http://example.com/codeflow?start=96&length=5)

fyi - double checked that Sql does indeed include this one, so we should be good.

jas-valgotar: I checked, it is there.

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #1073, opened 2023-02-09
Add rounding to trig table tests

CarlosFigueiraMSFT:
The trigonometric tabular tests use functions that don't return whole numbers (such as Pi() and other trig functions). This can make the results not match exactly depending on floating point behavior, so we're adding a Round call to those tests to make sure that the results are the expected ones in all platforms.


LucGenetier:
✅ No public API change.

Ananya-Gupta24 (approved):
Could you also add the exact same tests for ConsistentOneColumnTableResult? If not, then I'll add them in a subsequent PR that I'm working on to refactor some of the tabular tests

PR #1074, opened 2023-02-09
Add trig tabular function implementations, enable tests

CarlosFigueiraMSFT:
We had a test for trig tabular functions that was never enabled, because those functions were not implemented. Added the implementation for the tabular functions, fixed the test and enabled it.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/NumberValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/NumberValue.cs
@@ -22,7 +22,7 @@ namespace Microsoft.PowerFx.Types
 
         public override void ToExpression(StringBuilder sb, FormulaValueSerializerSettings settings)
         {
-            sb.Append(Value.ToString());
+            sb.Append((Value == 0) ? "0" : Value.ToString());
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/libraries/Microsoft.PowerFx.Core/Public/Values/NumberValue.cs
```diff
@@ -22,7 +22,7 @@ public override void Visit(IValueVisitor visitor)
 
         public override void ToExpression(StringBuilder sb, FormulaValueSerializerSettings settings)
         {
-            sb.Append(Value.ToString());
+            sb.Append((Value == 0) ? "0" : Value.ToString());
```
CarlosFigueiraMSFT, line 25: >(Value == 0) ? "0" : Value.ToString() [](http://example.com/codeflow?start=22&length=37)

We were getting a "-0" result in a few scenarios of a very small value. This fixes it.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

Ananya-Gupta24 (approved):

LucGenetier:
✅ No public API change.

PR #1076, opened 2023-02-10
Add missing case to TryGetDataSourceNode

lesaltzm:
As nodes aren't handled when trying to find the Data Source TexlNode from a function call. This breaks legacy analysis ECS in some scenarios.


LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

PR #1078, opened 2023-02-10
Handling BooleanOptionSetToBoolean coercion

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx-Dataverse/issues/61

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
@@ -4,8 +4,10 @@
 using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using Microsoft.PowerFx.Core.Errors;
 using Microsoft.PowerFx.Core.IR;
 using Microsoft.PowerFx.Core.IR.Nodes;
+using Microsoft.PowerFx.Core.Localization;
 using Microsoft.PowerFx.Interpreter;
 using Microsoft.PowerFx.Interpreter.Exceptions;
 using Microsoft.PowerFx.Types;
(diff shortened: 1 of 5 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
```diff
@@ -449,6 +460,12 @@ public static FormulaValue OptionSetValueToString(IRContext irContext, OptionSet
             var displayName = optionSet.DisplayName;
             return new StringValue(irContext, displayName);
         }
+
+        public static FormulaValue BooleanOptionSetToBoolean(IRContext irContext, OptionSetValue[] args)
+        {
+            var arg0 = args[0];
+            return arg0.Option == "1" ? FormulaValue.New(true) : FormulaValue.New(false);
```
MikeStall, line 467: I wonder if we should throw / error if it's not "1" or "0".

LucGenetier:
✅ No public API change.

ian-legler:
Should there be some unit tests to cover these changes?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
anderson-joyle: Good question. From DV we know we can expect "1" or "0". I'm wondering if we should check for "true"/"false".

ian-legler (approved):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

PR #1080, opened 2023-02-13
Fixes #1077 SymbolTable not giving back DisplayNames

### LINKED ISSUE #1077: SymbolTable  not giving back DisplayNames
ReadOnlySymbolTable can enumerate NamedFormulaType, but that only gives logical name and leaves .DisplayName empty. 

https://github.com/microsoft/Power-Fx/blob/91833f01001ec79a8a9190c8e046cd1d96be7709/src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs#L254

Consider both variables added with display name but also rowscope over Records with display names. 

We use this for tooling and intellisense features that suggest display names. 

This should also let us enable this test:
https://github.com/microsoft/Power-Fx-Dataverse/blob/fb6d11c22f1b1491ab2887f3886d7b340eb5d544/src/PowerFx.Dataverse.Tests/PluginExecutionTests/PluginExecutionTests.cs#L59

jas-valgotar:
Fixes #1077

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
@@ -261,7 +261,8 @@ namespace Microsoft.PowerFx
                 foreach (var kv in globals.GlobalSymbols)
                 {
                     var type = FormulaType.Build(kv.Value.Type);
-                    yield return new NamedFormulaType(kv.Key, type);
+                    var displayName = kv.Value.DisplayName != default ? kv.Value.DisplayName.Value : null;
+                    yield return new NamedFormulaType(kv.Key, type, displayName);
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
```diff
@@ -260,8 +260,9 @@ public IEnumerable<NamedFormulaType> SymbolNames
                 // GlobalSymbols are virtual, so we get derived behavior via that.
                 foreach (var kv in globals.GlobalSymbols)
                 {
-                    var type = FormulaType.Build(kv.Value.Type);
-                    yield return new NamedFormulaType(kv.Key, type);
+                    var type = FormulaType.Build(kv.Value.Type);
+                    var displayName = kv.Value.DisplayName != default ? kv.Value.DisplayName.Value : null;
```
MikeStall, line 264: >kv.Value.DisplayName != default ? kv.Value.DisplayName.Value : null; [](http://example.com/codeflow?start=38&length=68)

Can we use ?. operator?

 kv.Value.DisplayName?.Value

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
jas-valgotar: Since DName is struct, this syntax wouldn't work.

Issue #1006, opened 2023-01-20
Consider moving the blank value handling of function parameters to the IR


CarlosFigueiraMSFT:
Power Fx has clear rules of how blank values should be handled: if passed to functions in a parameter where a string value is expected, it should be interpreted as an empty string; for numeric parameters, it should be interpreted as 0; false for Boolean values, and the "date 0" for date, time or date/time arguments. Since this is a common rule, we should implement this logic in the IR so that different platforms don't have to implement it themselves.

MikeStall:
See #820 - hat starts this work. 

See 'replaceBlankValues' in  [Library.cs ](https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs)  for inspiration. .
Right idea, but that's all Interpreter-specific, so not part of IR.  

This is most interesting for the subset of functions implemented by SQL backend. 

That has several "trivial" modes:
- DoNotReplaceBlank 
- ReplaceBlankWithZero - 40 usages
- ReplaceBlankWithEmptyString - 13 usages 
as well as some more complex ones 

1. Add a virtual onto TexlFunction to capture BlankHandling semantics
2. IR can read and handle this. 
3. Remove all "trivial blank" handling from Library.cs and leverage IR.  (effectively, a replace their blankHandler with "NopAlreadyHandledByIR" handler that is just a nop)

PR #1081, opened 2023-02-13
Moves BlankToEmptyString arg preprocessing to IR Phase.

jas-valgotar:
-Follow up to #1006 #1059 
-This introduces a new UNaryOpKind `BlankToEmptyString`, since no current function satisfies the BlankToEmptyString behavior.
-Note: Other backends must implement this new UNaryOpKind `BlankToEmptyString` for these functions to work.

NOTE: In this PR 3 function are left out purposefully. 
`Concatenate()` due to its frequency of use with `&` operator and because of that it may generate very long IR
 `StartsWith()` and `EndsWith()` due to complexity with support of these function in delegation. 

Corrsponding Dataverse PR: https://github.com/microsoft/Power-Fx-Dataverse/pull/81

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/ArgPreprocessor.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/ArgPreprocessor.cs
@@ -12,5 +12,6 @@ namespace Microsoft.PowerFx.Core.Functions
         None = 0,
         ReplaceBlankWithZero = 1,
         ReplaceBlankWithZeroAndTruncate = 2,
+        ReplaceBlankWithEmptyString = 3,
     }
 }
--- a/src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Runtime.CompilerServices;
 using Microsoft.PowerFx.Core.Binding;
 using Microsoft.PowerFx.Core.Functions;
 using Microsoft.PowerFx.Core.IR.Nodes;
@@ -15,7 +14,6 @@ using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
 using Microsoft.PowerFx.Syntax;
 using Microsoft.PowerFx.Types;
-using static Microsoft.PowerFx.Syntax.PrettyPrintVisitor;
 using BinaryOpNode = Microsoft.PowerFx.Core.IR.Nodes.BinaryOpNode;
 using CallNode = Microsoft.PowerFx.Core.IR.Nodes.CallNode;
 using ErrorNode = Microsoft.PowerFx.Core.IR.Nodes.ErrorNode;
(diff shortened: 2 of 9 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/ArgPreprocessor.cs
```diff
@@ -10,9 +10,11 @@ namespace Microsoft.PowerFx.Core.Functions
     internal enum ArgPreprocessor
     {
         None = 0,
-        ReplaceWithZero = 1,
+        ReplaceBlankWithZero = 1,
         
         // This also includes ReplaceWithZero
         Truncate = 2,
+
+        ReplaceBlankWithEmptyString = 3,
```
MikeStall, line 18: >ReplaceBlankWithEmptyString [](http://example.com/codeflow?start=8&length=27)

it would be good to comment why this can't just be coalesce.

jas-valgotar: Added the comment in IRTranslator

MikeStall:
Adding Breaking Tag because we will need to update Dataverse compiler to handle the new IR opcode. 
No breaks to PAClient.

MikeStall (dismissed):

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -385,7 +386,7 @@ private static IntermediateNode ReplaceBlankWithZero(IntermediateNode arg)
                     return arg;
                 }
 
-                // need a new context since when arg is Blank IRContext.Returntypee is not a Number but a Blank.
+                // need a new context since when arg is Blank IRContext.Returntype is not a Number but a Blank.
```
CarlosFigueiraMSFT, line 389: >Returntype [](http://example.com/codeflow?start=72&length=10)

ResultType? Same with BlankToEmptyString

CarlosFigueiraMSFT:
>                 BuiltinFunctionsCore.Concatenate,

Can we handle this function (in addition to [Starts|Ends]With) as well?

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs:307 in 9ad76ad. [](commit_id = 9ad76ad47950557c3f3f3be4ea1eba5a8eeddf5a, deletion_comment = False)

ian-legler (approved):
:shipit:

jas-valgotar:
>                 BuiltinFunctionsCore.Concatenate,

As discussed offline, we are skipping this functions for this PR.

---
In reply to: [1440609573](https://github.com/microsoft/Power-Fx/pull/1081#issuecomment-1440609573) [](http://example.com/codeflow?ancestors=1440609573,1440609573)

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs:307 in 166eb5f. [](commit_id = 166eb5ff8ab5dc95f3bfd14fee3a317cb231a154, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
jas-valgotar: updated.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

PR #1082, opened 2023-02-14
Include entire error response from the connector

Robulane:
In case of errors, we only get the error code. Having details is important to be able to understand and solve the issue.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #1087, opened 2023-02-14
Proper validation of RGBA arguments

CarlosFigueiraMSFT:
The first three arguments to the RGBA function are integer values, and as such are truncated when used - `RGBA(123.45, 0, 0, 1)` is the same as `RGBA(123, 0, 0, 1)`. However, the validation is done prior to the truncation, which makes some inputs to be considered invalid that should be valid, such as `RGBA(255.1, 0, 0, 1)`.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryColor.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryColor.cs
@@ -62,28 +62,29 @@ namespace Microsoft.PowerFx.Functions
 
         public static FormulaValue RGBA(IRContext irContext, NumberValue[] args)
         {
+            // Truncate (strip decimals)
+            var red = (int)args[0].Value;
+            var green = (int)args[1].Value;
+            var blue = (int)args[2].Value;
+            var alpha = args[3].Value;
+
             // Ensure rgb numbers are in range (0-255)
-            if (args[0].Value < 0.0d || args[0].Value > 255.0d
-                || args[1].Value < 0.0d || args[1].Value > 255.0d
-                || args[2].Value < 0.0d || args[2].Value > 255.0d)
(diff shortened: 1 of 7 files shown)


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/RGBA.txt
```diff
@@ -22,6 +22,9 @@ Error({Kind:ErrorKind.InvalidArgument})
 >> RGBA(255, 1.5, 255, 1.0)
 RGBA(255,1,255,1)
 
+>> RGBA(255.5, 1.5, 255.9, 1.0)
```
MikeStall, line 25: Great test - Can you add a comment on this to explain it.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryColor.cs
```diff
@@ -62,28 +62,29 @@ private static byte ParseColor(Match match, string color)
 
         public static FormulaValue RGBA(IRContext irContext, NumberValue[] args)
         {
+            // Truncate (strip decimals)
```
MikeStall, line 65: With #1072, can we now do the truncation at the IR level? 

We'd just need to implement TexlFunction.GetArgPreprocessor - in theory, LibraryColor.cs wouldn't even need any changes.

MikeStall (approved):

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryColor.cs
CarlosFigueiraMSFT: Updated function definition to use truncation at IR

PR #1088, opened 2023-02-14
Wrap Log/Exp tabular tests with Round to avoid rounding errors

CarlosFigueiraMSFT:
The value of an expression such as `Ln(Exp(3))` or `Exp(Ln(3))` is 3, but due to rounding / floating point errors in different platforms it can be a little different. Wrapping the expression with the Round function to make sure that we get the same result everywhere.


LucGenetier:
✅ No public API change.

PR #1094, opened 2023-02-15
Fixes #1038 If record type mismatch.

### LINKED ISSUE #1038: Crash from type mismatch
If() takes an intersection. 
Table() takes a union. 

Consider:
`If(false, { x : 1, y:1}, {x:1, z:2}).z`  is an error because If() takes the intersection. 

But 
`If(false, { x : 1, y:1}, {x:1, z:2})`

Still returns a runtime type showing {x:1, z:2} 
but compile time type  is {x}, so there's a mismatch between compile time and runtime type. 

PA Json() function correctly strips this out; but C# interpreter does not. 

Here are 2 more expressions that exploit this: 
```
First(Table(
  If(false, { x : 1, y:1}, {x:1, z:2}), // {x}, because intersection 
  {x:9,z:9})  // {x,z}
).z // Should be Blank()! actually showing 2
```
Z in first row should get stripped out .

If z in first row and z in second row have different types, leads to a crash in C# interpreter. 

```
Table(
If(false, { x : 1, y:1}, {x:1, z:"str"}), // {x}
{x:9,z:9}) // Blank(),  actually crashes 

```

jas-valgotar:
-Fixes #1038 
-This may wrap the result of the If statement in CompileTimeTypeWrapperRecordValue.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/CompileTimeTypeWrapperRecordValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/CompileTimeTypeWrapperRecordValue.cs
@@ -1,6 +1,9 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.PowerFx.Core.IR;
 
 namespace Microsoft.PowerFx.Types
@@ -23,6 +26,45 @@ namespace Microsoft.PowerFx.Types
         private CompileTimeTypeWrapperRecordValue(RecordType type, RecordValue inner)
             : base(IRContext.NotInSource(type), inner.Fields)
         {
+        }
+
+        protected override bool TryGetField(FormulaType fieldType, string fieldName, out FormulaValue result)
+        {
+            if (Type.TryGetFieldType(fieldName, out _))
+            {
+                // Only return field which were specified via the expectedType (IE RecordType),
+                // because inner record value may have more fields than the expected type.
+                return _fields.TryGetValue(fieldName, out result);
+            }
+
+            result = default;
+            return false;
+        }
+
+        public override async Task<DValue<RecordValue>> UpdateFieldsAsync(RecordValue changeRecord, CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            if (_mutableFields == null)
+            {
+                return await base.UpdateFieldsAsync(changeRecord, cancellationToken);
+            }
+
+            await UpdateExistingFields(changeRecord, cancellationToken);
+
+            var fields = new List<NamedValue>();
+
+            foreach (var kvp in _fields)
+            {
+                // Only add fields which were specified via the expectedType (IE RecordType),
+                // because inner record value may have more fields than the expected type.
+                if (Type.TryGetFieldType(kvp.Key, out _))
+                {
+                    fields.Add(new NamedValue(kvp.Key, kvp.Value));
+                }
+            }
+
+            return DValue<RecordValue>.Of(NewRecordFromFields(fields));
         }
     }
 }
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/CompileTimeTypeWrapperRecordValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/CompileTimeTypeWrapperRecordValue.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.PowerFx.Core.IR;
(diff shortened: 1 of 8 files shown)


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/First.txt
```diff
@@ -89,3 +89,8 @@ Table({Value:5})
 >> FirstN(If(false, Table({a:1}), Table({b:2})), 1)
 Table({})
 
+>> First(Table(If(false, {x:1, y:1}, {x:1, z:2}), {x:9,z:9})).z
```
MikeStall, line 92: Can you add a comment to help explain why. 
Actually, it'd even be fine to include this as a test case right before to help explain it. 

If(false, {x:1, y:1}, {x:1, z:2})

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/With.txt
```diff
@@ -14,4 +14,4 @@ Blank()
 Error(Table({Kind:ErrorKind.Div0},{Kind:ErrorKind.Div0}))
 
 >> With({y:1}, If(false, {x:1}, {z:2}))
```
MikeStall, line 16: >>> [](http://example.com/codeflow?start=0&length=2)

nice - also worth commenting here.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -15,6 +15,7 @@
 using Microsoft.PowerFx.Core.Utils;
 using Microsoft.PowerFx.Interpreter;
 using Microsoft.PowerFx.Types;
+using static Microsoft.PowerFx.Syntax.PrettyPrintVisitor;
```
MikeStall, line 18: >PrettyPrintVisitor [](http://example.com/codeflow?start=38&length=18)

Do we need this?
PrettyPrinting seems suspicious  here

src/libraries/Microsoft.PowerFx.Core/Public/Values/CompileTimeTypeWrapperRecordValue.cs
```diff
@@ -24,5 +27,44 @@ private CompileTimeTypeWrapperRecordValue(RecordType type, RecordValue inner)
             : base(IRContext.NotInSource(type), inner.Fields)
         {
         }
+
+        protected override bool TryGetField(FormulaType fieldType, string fieldName, out FormulaValue result)
```
MikeStall, line 31: >TryGetField [](http://example.com/codeflow?start=32&length=11)

Are these extra virtuals only required in the If() case where we *reduce* the type? (eg, have an intersection where the resulting type is smaller, and hence TryGetField needs to check if it should "block" a field?)

Ie, in Table() where we do a union, would we ever need these? My concern is the potential cost of the extra lookup. 
Either way - if it is a problem, I guess we'll find it in a profiler.

src/libraries/Microsoft.PowerFx.Core/Public/Values/CompileTimeTypeWrapperRecordValue.cs
```diff
@@ -24,5 +27,44 @@ private CompileTimeTypeWrapperRecordValue(RecordType type, RecordValue inner)
...
+                // Only return field which were specified via the expectedType (IE RecordType),
+                // because inner record value may have more fields than the expected type.
+                return _fields.TryGetValue(fieldName, out result);
+            }
+
+            result = default;
+            return false;
+        }
+
+        public override async Task<DValue<RecordValue>> UpdateFieldsAsync(RecordValue changeRecord, CancellationToken cancellationToken)
```
MikeStall, line 44: >UpdateFieldsAsync [](http://example.com/codeflow?start=56&length=17)

Can this be merged with base class more?

src/libraries/Microsoft.PowerFx.Core/Public/Values/CompileTimeTypeWrapperRecordValue.cs
jas-valgotar: No this is only used for intersection.

src/libraries/Microsoft.PowerFx.Core/Public/Values/CompileTimeTypeWrapperRecordValue.cs
jas-valgotar: done.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/First.txt
jas-valgotar: added

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/With.txt
jas-valgotar: added

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

Issue #1098, opened 2023-02-17
ArgumentOutOfRangeException in System.DateTime.Add


MikeStall:
Text(4492800,DateTimeFormat.LongDate)

System.ArgumentOutOfRangeException: Value to add was out of range. (Parameter 'value')
at System.DateTime.Add(Double value, Int32 scale)
at System.DateTime.AddDays(Double value)
at Microsoft.PowerFx.Functions.Library.NumberToDateTime(FormattingInfo formatInfo, IRContext irContext, NumberValue value)
at Microsoft.PowerFx.Functions.Library.TryText(FormattingInfo formatInfo, IRContext irContext, FormulaValue value, String formatString, StringValue& result)
at Microsoft.PowerFx.Functions.Library.Text(FormattingInfo formatInfo, IRContext irContext, FormulaValue[] args)
at Microsoft.PowerFx.Functions.Library.Text(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
at Microsoft.PowerFx.Functions.Library.<>c__DisplayClass343_0`1.<StandardErrorHandling>b__0(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, T[] args)
at Microsoft.PowerFx.Functions.Library.<>c__DisplayClass340_0`1.<<StandardErrorHandlingAsync>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
at Microsoft.PowerFx.EvalVisitor.Visit(CallNode node, EvalVisitorContext context)
at Microsoft.PowerFx.ParsedExpression.EvalAsync(CancellationToken cancellationToken, IRuntimeConfig runtimeConfig)
at FxPlayground.Core.Session.TryEvalAsync(String expression, UserInfo user, CancellationToken cancel) in D:\dev\pa2\PowerFxPlayground\src\FxPlaygroundCore\Session.cs:line 416

Repro:
http://aka.ms/fxplayground?session=anderdson1#expr_VGV4dCg0NDkyODAwLERhdGVUaW1lRm9ybWF0LkxvbmdEYXRlKQ==

PR #1100, opened 2023-02-17
Out of range number to date returns an error

anderson-joyle:
Issue #1098 
Out of range number to date returns an error instead of throwing an exception.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
@@ -442,7 +442,7 @@ namespace Microsoft.PowerFx.Functions
             }
             catch (ArgumentOutOfRangeException)
             {
-                (var shortMessage, _) = ErrorUtils.GetLocalizedErrorContent(TexlStrings.ErrTextOutOfRange, formatInfo.CultureInfo, out _);
+                (var shortMessage, _) = ErrorUtils.GetLocalizedErrorContent(TexlStrings.ErrTextInvalidArgDateTime, formatInfo.CultureInfo, out _);
 
                 throw new CustomFunctionErrorException(shortMessage, ErrorKind.InvalidArgument);
             }          
@@ -471,7 +471,7 @@ namespace Microsoft.PowerFx.Functions
                 case StringValue st:
                     return TryDateTimeParse(formatInfo, irContext, st, out result);
                 case NumberValue num:
-                    result = NumberToDateTime(formatInfo, irContext, num) as DateTimeValue;
+                    result = NumberToDateTime(formatInfo, irContext, num);
                     break;
                 case DateValue dv:
                     result = new DateTimeValue(irContext, dv.GetConvertedValue(formatInfo.TimeZoneInfo));
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Microsoft.PowerFx.Core.Errors;
 using Microsoft.PowerFx.Core.IR;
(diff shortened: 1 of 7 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Text.RegularExpressions;
```
MikeStall, line 6: >RegularExpressions [](http://example.com/codeflow?start=18&length=18)

needed?

src/tests/Microsoft.PowerFx.Interpreter.Tests/TypeCoercionTest.cs
```diff
@@ -79,16 +80,17 @@ private void TryCoerceToTargetTypes(FormulaValue inputValue, string exprBool, st
...
             }
 
-            isSucceeded = inputValue.TryCoerceTo(out StringValue resultString);
+            isSucceeded = inputValue.TryCoerceTo(out StringValue resultValue);
             if (exprStr != null)
             {
                 Assert.True(isSucceeded);
-                Assert.Equal(exprStr, resultString.Value);
+
+                Assert.Equal(exprStr, ((StringValue)resultValue).Value);
```
MikeStall, line 88: >StringValue [](http://example.com/codeflow?start=40&length=11)

is this cast needed? declared as string on line 83

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
```diff
@@ -425,19 +426,26 @@ public static DateValue NumberToDate(IRContext irContext, NumberValue[] args)
...
+        public static FormulaValue NumberToDateTime(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, NumberValue[] args)
         {
             return NumberToDateTime(CreateFormattingInfo(runner), irContext, args[0]);
         }
 
-        public static DateTimeValue NumberToDateTime(FormattingInfo formatInfo, IRContext irContext, NumberValue value)
+        public static FormulaValue NumberToDateTime(FormattingInfo formatInfo, IRContext irContext, NumberValue value)
         {
             var n = value.Value;
-            var date = _epoch.AddDays(n);
```
MikeStall, line 436: >       var date = _epoch.AddDays(n); [](http://example.com/codeflow?start=5&length=36)

This is where the failure is coming from (throws a ArgumentOutOfRangeException) 

We could just catch that and rethrow as CustomFunctionErrorException. That will get catch in EvalVisitor. 

That lets us keep the signature type safe (DateTimeValue instead of downgrading to FormulaValue); and means callers don't have to all check for failure.

MikeStall: else if we wanted  DateTimeValue or ErrroValue, we could use DValue<DateTimeValue>

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
```diff
@@ -463,7 +471,7 @@ public static bool TryGetDateTime(FormattingInfo formatInfo, IRContext irContext
                 case StringValue st:
                     return TryDateTimeParse(formatInfo, irContext, st, out result);
                 case NumberValue num:
-                    result = NumberToDateTime(formatInfo, irContext, num);
+                    result = NumberToDateTime(formatInfo, irContext, num) as DateTimeValue;
```
MikeStall, line 474: >DateTimeValue [](http://example.com/codeflow?start=77&length=13)

needed?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_Format.txt
```diff
@@ -636,6 +636,9 @@ Error({Kind:ErrorKind.InvalidArgument})
 >> Text(30470.01953125, "YYYY-MM-DD HH:MM:SS.00000000") 
 Error({Kind:ErrorKind.InvalidArgument})
 
+>> Text(4492800,DateTimeFormat.LongDate)
```
MikeStall, line 639: >>> [](http://example.com/codeflow?start=0&length=2)

Can we add a comment that this is an overflow?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
```diff
@@ -432,12 +431,21 @@ public static DateTimeValue NumberToDateTime(EvalVisitor runner, EvalVisitorCont
...
+                var n = value.Value;
+                var date = _epoch.AddDays(n);
 
-            date = MakeValidDateTime(formatInfo.TimeZoneInfo, date);
+                date = MakeValidDateTime(formatInfo.TimeZoneInfo, date);
 
-            return new DateTimeValue(irContext, date);
+                return new DateTimeValue(irContext, date);
+            }
+            catch (ArgumentOutOfRangeException)
```
MikeStall, line 443: >catch [](http://example.com/codeflow?start=12&length=5)

what error does PA give in this scenario?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
```diff
@@ -432,12 +431,21 @@ public static DateTimeValue NumberToDateTime(EvalVisitor runner, EvalVisitorCont
...
+            {
+                (var shortMessage, _) = ErrorUtils.GetLocalizedErrorContent(TexlStrings.ErrTextOutOfRange, formatInfo.CultureInfo, out _);
```
MikeStall, line 445: >ErrTextOutOfRange [](http://example.com/codeflow?start=88&length=17)

what error does PA give in this scenario?

anderson-joyle: PA has a larger limit. 
`The arguments to the Text function do not represent a valid date or time value`

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

PR #1101, opened 2023-02-17
Adds OptionSetInfo function, which returns the logical name of the option set value

### LINKED ISSUE #1091: Ability to access id from optionsets
We have an ask to support exposing the id from optionsets for multi-language scenarios where a customer is calling an api with the optionset as a parameter. Having access to the id will render the api consistant across all locales.

jas-valgotar:
-Fixes #1091 
-Adds OptionSetInfo function, which returns the logical name of the option set value

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/OptionSetInfo.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/OptionSetInfo.cs
@@ -1,0 +1,27 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+using Microsoft.PowerFx.Core.Functions;
+using Microsoft.PowerFx.Core.Localization;
+using Microsoft.PowerFx.Core.Types;
+
+namespace Microsoft.PowerFx.Core.Texl.Builtins
+{
+    internal sealed class OptionSetInfoFunction : BuiltinFunction
+    {
+        public override bool IsSelfContained => true;
+
+        public override bool SupportsParamCoercion => false;
+
+        public OptionSetInfoFunction()
+            : base("OptionSetInfo", TexlStrings.AboutOptionSetInfo, FunctionCategories.Text, DType.String, 0, 1, 1, DType.OptionSetValue)
+        {
+        }
+
+        public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
+        {
+            yield return new[] { TexlStrings.AboutOptionSetInfoArg1 };
+        }
+    }
+}
--- a/src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
+++ b/src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
@@ -11,7 +11,6 @@ using System.Threading.Tasks;
 using Microsoft.PowerFx.Core;
 using Microsoft.PowerFx.Core.Tests;
 using Microsoft.PowerFx.Core.Texl;
-using Microsoft.PowerFx.Core.Texl.Builtins;
 using Microsoft.PowerFx.Core.Types.Enums;
 using Microsoft.PowerFx.Core.Utils;
 using Microsoft.PowerFx.Functions;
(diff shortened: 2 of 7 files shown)


src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -732,6 +731,50 @@ public void OptionSetChecks()
             Assert.True(checkResult.IsSuccess);
         }
 
+        [Theory]
+        [InlineData("OptionSetInfo(OptionSet.option_1)", "option_1")]
+        [InlineData("OptionSetInfo(OptionSet.Option1)", "option_1")]
+        [InlineData("OptionSetInfo(Option1)", "option_1")]
+        [InlineData("OptionSetInfo(If(1<0, Option1))", null)]
+        public async void OptionSetInfoTests(string expression, string expected)
```
MikeStall, line 739: >OptionSetInfoTests [](http://example.com/codeflow?start=26&length=18)

Can we include test cases with Text() .

Helpful to see OptionSetInfo and Text() together to better understand their relationship.

LucGenetier:
✅ No public API change.

MikeStall (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/OptionSetInfo.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Microsoft.PowerFx.Core.Texl.Builtins
+{
+    internal sealed class OptionSetInfoFunction : BuiltinFunction
+    {
+        public override bool IsSelfContained => true;
+
+        public override bool SupportsParamCoercion => false;
+
+        public OptionSetInfoFunction()
+            : base("OptionSetInfo", TexlStrings.AboutOptionSetInfo, FunctionCategories.Text, DType.String, 0, 1, 1, DType.OptionSetValue)
```
lesaltzm, line 18: Likely this needs to have a CheckTypes, and instead of setting the return type to DType.String, set it based on the backing type of the option set (currently, use `isBooleanValued` to determine that, later update to what's in #1034)

lesaltzm: actually, without that PR we may not have enough info.

MikeStall: I'm fine initially limited this to strings (and failing for non-strings), since that's the biggest thing to unblock PVA.
But we should still fail gracefully for non-strings (not crash).

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -732,6 +731,50 @@ public void OptionSetChecks()
...
+            var config = new PowerFxConfig() { SymbolTable = symbol };
+            config.AddOptionSet(optionSet);
+            var recalcEngine = new RecalcEngine(config);
+            
+            var result = await recalcEngine.EvalAsync(expression, CancellationToken.None, symValues);
+            Assert.Equal(expected, result.ToObject());
+        }
+
+        [Fact]
+        public async Task OptionSetInfoNegativeTest()
```
MikeStall, line 771: >OptionSetInfoNegativeTest [](http://example.com/codeflow?start=26&length=25)

Would also be good to add some negative cases for what we *can't* pass to OptionSetInfo.
Ie, no strings, no enums, no numbers. 
OptionSetInfo(Color.Red) should fail.

LucGenetier:
✅ No public API change.

lesaltzm (dismissed):
🚢

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -732,6 +731,65 @@ public void OptionSetChecks()
...
+            {
+                    { "option_1", "Option1" },
+                    { "option_2", "Option2" }
+            }));
+
+            var config = new PowerFxConfig();
+            config.AddOptionSet(optionSet);
+            var recalcEngine = new RecalcEngine(config);
+
+            await Assert.ThrowsAsync<InvalidOperationException>(() => recalcEngine.EvalAsync(expression, CancellationToken.None));
```
MikeStall, line 790: nit - we can call ,Check() and verify checkResult.IsSuccess == false.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
jas-valgotar: updated

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
jas-valgotar: updated

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
jas-valgotar: updated

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1104, opened 2023-02-20
Fix version hash -shouldn't use timer.

MikeStall:
VersionHashTests used UtcNow which was non-deterministic and were randomly failing in CI. 

Add test to help verify uniqueness.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/VersionHash.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/VersionHash.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Threading;
 using Microsoft.PowerFx.Core;
 using Microsoft.PowerFx.Core.App;
 using Microsoft.PowerFx.Core.Binding;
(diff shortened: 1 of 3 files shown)


src/tests/Microsoft.PowerFx.Core.Tests/VersionHashTests.cs
```diff
@@ -65,5 +65,31 @@ public void Combine()
 
             Assert.NotEqual(v1.Combine(vx), v1.Combine(vy));
         }
+
+        // Hash doesn't *need* to be unique, but it is highly desirably to help catch checks. 
+        [Fact]
+        public void Unique()
```
MikeStall, line 71: This test consistently fails without the fix.

src/libraries/Microsoft.PowerFx.Core/Public/Config/VersionHash.cs
```diff
@@ -33,11 +34,15 @@ private VersionHash(int value)
             _value = value;
         }
 
+        private static int _hashStarter;
+
         public static VersionHash New()
         {
             // Give psuedo-random seed.
-            // This can help protect against 2 objects accidentally appearing the same. 
-            var value = (int)DateTime.UtcNow.Ticks;
```
MikeStall, line 40: Remove use of the timer here in favor of something more deterministic. 

2 calls can reasonably happen at the same tick - new test below catches and fails.

LucGenetier (dismissed):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier (approved):
:shipit:

Issue #530, opened 2022-08-09
Connectors -  don't call FromJson(string) since it loses swagger type information


MikeStall:
Basically, connectors should never call `FormulaValue.FromJson(json)` because that doesn't leverage swagger. 
For example, a string, guid, and date are all the same in json. 

The json+swagger serializer solve this problem. But direct calls to FromJson(json) avoid that. 

One usage is in DecodeResponseAsync(). 
What this means is that if a connector returns a Date (and specifies that in its swagger), we'll call FromJson and deserialize as a string!

AlaJaber-Tahaluf:
I support this issues, I was face this issue and I solved it using this modification:
`
public RecalcEngineScope GetScope(string contextJson)
        {
            var engine = GetEngine();
            var Context = JsonConvert.DeserializeObject<Dictionary<string, object>>(contextJson);
            foreach (var item in Context)
            {
                var value_eval = item.Value?.ToString().Trim('\"');
                var formula_value = EngineContext.GetFormulaValue(engine, value_eval);
                engine.Config.SymbolTable.RemoveVariable(item.Key);
                engine.UpdateVariable(item.Key, formula_value);
            }

            return RecalcEngineScope.FromJson(engine, "{}");
        }
`
where GetFormulaValue function is a custom function that I used to convert a string, object, or list value to a best formula available in RecalcEngine for example when {value_eval} is date e.g. {"20/09/2022"} I convert it to {DateTimeValue("20/09/2022")} and then pass it to `engine.Eval("DateTimeValue(\"20/09/2022\")")` and return value to store it in engine using `engine.UpdateVariable(item.Key, formula_value)`

but still, for this moment I still face some issues with other types, So I wish to solve this issue as soon as

PR #1112, opened 2023-02-24
Issue 1109

LucGenetier:
Identify UntypedObjects in connector function return type


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -170,7 +170,7 @@ public async Task<FormulaValue> DecodeResponseAsync(HttpResponseMessage response
             {                
                 return string.IsNullOrWhiteSpace(text) 
                     ? FormulaValue.NewBlank(_returnType) 
-                    : FormulaValueJSON.FromJson(text); // $$$ Do we need to check response media type to confirm that the content is indeed json?
+                    : FormulaValueJSON.FromJson(text, _returnType); // $$$ Do we need to check response media type to confirm that the content is indeed json?
```
MikeStall, line 173: This looks related to #530

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall:
>                     return new BlankValue(IRContext.NotInSource(FormulaType.Blank));

should this use formulaType?

---
Refers to: src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs:48 in 03aa316. [](commit_id = 03aa316e4c46e7b0bea7196244bc026bd204d310, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -57,18 +60,29 @@ public static FormulaValue FromJson(JsonElement element)
                     return new BooleanValue(IRContext.NotInSource(FormulaType.Boolean), true);
```
MikeStall, line 60: Should these verify formulaType?
Ie, if FormulaType doesn't match the element.ValueKind?

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -57,18 +60,29 @@ public static FormulaValue FromJson(JsonElement element)
...
                 case JsonValueKind.Array:
-                    return TableFromJsonArray(element);
+                    if (formulaType == null || formulaType is TableType)
+                    {
+                        return TableFromJsonArray(element, formulaType as TableType);
+                    }
+                    else if (formulaType is UntypedObjectType uot)
+                    {
+                        return new UntypedObjectValue(IRContext.NotInSource(FormulaType.UntypedObject), new JsonUntypedObject(element));
+                    }
```
MikeStall, line 73: Should this case be top-level?
IE, regardless of the element kind, if they ask for an UntypedObject, we could provide it.

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -92,16 +113,19 @@ private static RecordValue RecordFromJsonObject(JsonElement element)
...
         // [1,2,3]  is a single column table, actually equivalent to: 
         // [{Value : 1, Value: 2, Value :3 }]
-        internal static TableValue TableFromJsonArray(JsonElement array)
+        internal static TableValue TableFromJsonArray(JsonElement array, TableType tableType)
         {
             Contract.Assert(array.ValueKind == JsonValueKind.Array);
 
             var records = new List<RecordValue>();
+            var fields = tableType?.ToRecord().FieldNames;
+            bool isArray = fields?.Count() == 1 && fields?.First() == "Value";
```
MikeStall, line 122: >fields?.Count() == 1 && fields?.First() == "Value" [](http://example.com/codeflow?start=27&length=50)

tableType.IsColumn?

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -92,16 +113,19 @@ private static RecordValue RecordFromJsonObject(JsonElement element)
...
 
             for (var i = 0; i < array.GetArrayLength(); ++i)
             {
                 var element = array[i];
-                var val = GuaranteeRecord(FromJson(element));
+                FormulaType ft = isArray ? tableType.ToRecord().GetFieldType("Value") : tableType?.ToRecord();
```
MikeStall, line 127: Can this be lifted outside of the loop?

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -16,15 +18,15 @@ public class FormulaValueJSON
         /// <summary>
         /// Convenience method to create a value from a json representation. 
         /// </summary>
-        public static FormulaValue FromJson(string jsonString)
+        public static FormulaValue FromJson(string jsonString, FormulaType formulaType = null)
```
MikeStall, line 21: >formulaType [](http://example.com/codeflow?start=75&length=11)

IF this is non-null, does that mean that 

FromJson(json, type).Type == type?

MikeStall:
Direction looks good. 

We should add some new unit tests.  (code coverage will probably catch most of these)
- what if incoming formulaType doesn't match json schema?
- untyped object scenario 
- table: single column vs. records
- Records: when formulaType is a subset, superset of the json object.

MikeStall:
>                 case JsonValueKind.String:

If formulaType is Date, Guid  are serialized as string. Do we need to handle? 

This could be related  to the new helpers in:
[https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs](https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs)

---
Refers to: src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs:53 in 03aa316. [](commit_id = 03aa316e4c46e7b0bea7196244bc026bd204d310, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
MikeStall: This could be related  to the new helpers in:
[[https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs](https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs)](https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs)

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -79,8 +93,15 @@ private static RecordValue RecordFromJsonObject(JsonElement element)
             {
                 var name = pair.Name;
                 var value = pair.Value;
+                FormulaType fieldType = null;
 
-                var paValue = FromJson(value);
+                // if TryGetFieldType fails, fieldType is set to Blank
+                if (!recordType?.TryGetFieldType(name, out fieldType) != true)
```
xieofxie, line 99: I suppose this should be `if (recordType?.TryGetFieldType(name, out fieldType) != true)`?

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -57,18 +60,29 @@ public static FormulaValue FromJson(JsonElement element)
...
-                    return RecordFromJsonObject(element);
+                    return RecordFromJsonObject(element, formulaType as RecordType);
 
                 case JsonValueKind.Array:
-                    return TableFromJsonArray(element);
+                    if (formulaType == null || formulaType is TableType)
+                    {
+                        return TableFromJsonArray(element, formulaType as TableType);
+                    }
+                    else if (formulaType is UntypedObjectType uot)
```
xieofxie, line 70: We should return array of UntypedObjectType

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
xieofxie: Interestingly, if we set fieldtype to null, it will parse everything to the end instead of return a JsonUntypedObject which however, fits our code better

src/tests/Microsoft.PowerFx.Connectors.Tests/OpenApiParserTests.cs
```diff
@@ -170,9 +170,10 @@ public async Task ACSL_InvokeFunction()
 
             FormulaValue result = await function.InvokeAync(client, new FormulaValue[] { analysisInputParam, parametersParam }, CancellationToken.None);
 
-            string input = testConnector._log.ToString();
-            string output = System.Text.Json.JsonSerializer.Serialize(result.ToObject());
+            Assert.NotNull(result);
+            Assert.True(result is RecordValue);
```
xieofxie, line 174: check entities[0].resolutions[0] is untyped

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
LucGenetier: Yes

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
LucGenetier: It should be at top level, you're right

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
LucGenetier: element is an array and it will be serialized properly

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
LucGenetier: This is best effort here
If we can check the type, let's do it
If the property name is unknown, let's not do it and accept it as it comes

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
LucGenetier: Some returned values are optional so this isn't true

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -16,15 +18,15 @@ public class FormulaValueJSON
         /// <summary>
         /// Convenience method to create a value from a json representation. 
         /// </summary>
-        public static FormulaValue FromJson(string jsonString)
+        public static FormulaValue FromJson(string jsonString, FormulaType formulaType = null)
         {
             try
             {
                 using var document = JsonDocument.Parse(jsonString);
```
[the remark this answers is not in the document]
LucGenetier: [https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/using](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/using)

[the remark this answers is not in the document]
LucGenetier: This is OK as long as FromJson(...) doesn't keep a reference from propBag.
It was keeping a JsonElement, hence UntypedObjectValue.Impl being disposed.
This is now fixed

LucGenetier:
>                 case JsonValueKind.String:

We will probably fail in NotImplementException and this is fine for now as we don't have a need. We will fix it later.

---
In reply to: [1444653148](https://github.com/microsoft/Power-Fx/pull/1112#issuecomment-1444653148) [](http://example.com/codeflow?ancestors=1444653148,1444653148)

---
Refers to: src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs:53 in 03aa316. [](commit_id = 03aa316e4c46e7b0bea7196244bc026bd204d310, deletion_comment = False)

LucGenetier:
>                     return new BlankValue(IRContext.NotInSource(FormulaType.Blank));

Don't think so as any type can be "null" (or Blank)
[https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonvaluekind?view=net-7.0](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonvaluekind?view=net-7.0)


---
In reply to: [1444616665](https://github.com/microsoft/Power-Fx/pull/1112#issuecomment-1444616665) [](http://example.com/codeflow?ancestors=1444616665,1444616665)

---
Refers to: src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs:48 in 03aa316. [](commit_id = 03aa316e4c46e7b0bea7196244bc026bd204d310, deletion_comment = False)

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Json/Functions/JsonUntypedObject.cs
```diff
@@ -21,7 +21,7 @@ internal class JsonUntypedObject : IUntypedObject
 
         public JsonUntypedObject(JsonElement element)
         {
-            _element = element;
+            _element = element.ValueKind == JsonValueKind.Undefined ? element : element.Clone();
```
MikeStall, line 24: >element.Clone() [](http://example.com/codeflow?start=80&length=15)

We should double check whether we call this here, or have specific callers do it. 

Here impacts all scenarios, and there might be scenarios that *don't* want to clone.

LucGenetier: The problem is that we receive a JsonElement which is part of a disposable JsonDocument
If we don't clone it, we might lose the data

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Json.Tests/ParseJSONTests.cs
```diff
@@ -24,5 +21,285 @@ public void BasicParseJson()
...
+            RecordValue rv = (RecordValue)fv;
+            Assert.Empty(rv.Fields);
+            Assert.Equal("![]", rv.Type.ToStringWithDisplayNames());            
+
+            FormulaValue fv2 = FormulaValueJSON.FromJson(expr, rv.Type);
+            Assert.NotNull(fv2);
+            Assert.True(fv2 is RecordValue);
+
+            Assert.Throws<NotImplementedException>(() => FormulaValueJSON.FromJson(expr, TableType.Empty()));
+            Assert.Throws<NotImplementedException>(() => FormulaValueJSON.FromJson(expr, new NumberType()));
```
MikeStall, line 185: >new NumberType() [](http://example.com/codeflow?start=89&length=16)

nit: we have static fields for these:
FormulaType.Number

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -92,16 +151,19 @@ private static RecordValue RecordFromJsonObject(JsonElement element)
         // Parse json. 
         // [1,2,3]  is a single column table, actually equivalent to: 
         // [{Value : 1, Value: 2, Value :3 }]
-        internal static TableValue TableFromJsonArray(JsonElement array)
+        internal static TableValue TableFromJsonArray(JsonElement array, TableType tableType)
         {
             Contract.Assert(array.ValueKind == JsonValueKind.Array);
 
             var records = new List<RecordValue>();
+            var fields = tableType?.ToRecord().FieldNames;
```
MikeStall, line 159: do we use this new var fields?

MikeStall (approved):
:shipit:

Issue #843, opened 2022-11-14
With Forces Matching If statement return types


daryllabar:
This code works perfectly fine:
```
    If(IsBlank(note),
        Patch(SiteHistory, Defaults(SiteHistory), { TaskDetails: "Value" });
        Notify("New Note created", NotificationType.Information);
    ,
        Patch(SiteHistory, note, { TaskDetails: "Value" });
    );
```

But this code gives a very unhelpful error message of "The function If has some invalid arguments":
```
With({
    toUpdate: {
        TaskDetails: "Value"
    }
},
    If(IsBlank(note),
        Patch(SiteHistory, Defaults(SiteHistory), toUpdate);
        Notify("New Note created", NotificationType.Information);
    ,
        Patch(SiteHistory, note, toUpdate);
    );
);
```

But this works just fine:

```
With({
    toUpdate: {
        TaskDetails: "Value"
    }
},
    If(IsBlank(note),
        Notify("New Note created", NotificationType.Information);
        Patch(SiteHistory, Defaults(SiteHistory), toUpdate);
    ,
        Patch(SiteHistory, note, toUpdate);
    );
);
```

As well as this:
```
With({
    toUpdate: {
        TaskDetails: "Value"
    }
},
    If(IsBlank(note),
        Patch(SiteHistory, Defaults(SiteHistory), toUpdate);
        Notify("New Note created", NotificationType.Information);
    ,
        Patch(SiteHistory, note, toUpdate);
        Notify("Note updated", NotificationType.Information);
    );
);
```

Which leads me to believe that the If must return the same type of value within the With.  Is this expected?  If so, is this documented?  

I would expect that as long as I don't use the output of an if, it shouldn't care if different types of values are returned.

Session Details if helpful
> Session ID: e18c08cf-45bc-4662-87d1-2e5d548de70a
> PowerApps 3.22105.21
> Object ID: dc40ed33-a73f-4ffe-9c10-bf35b3cd21e8
> Tenant ID: d843cb99-8386-4d4c-a924-e7147b4e184a
> Embedded Correlation ID: 52450bb0-6463-11ed-a220-0fbb07ff730b
> App ID: c7eac419-9005-42bc-b262-ba75625182e6
> Telemetry location: global

MikeStall:
"I would expect that as long as I don't use the output of an if, it shouldn't care if different types of values are returned." 
[...]

PR #1113, opened 2023-02-24
Introduces a new Void type so that we can remove IsArgTypeInconsequential

### LINKED ISSUE #1092: Introduce Void type, Remove IsArgTypeInconsequential
Consider cases like If where true/false branch have incompatible types, such as returning 2 different records:
`If(test, Patch(Accounts,...), Patch(Contacts,...))`

This is not allowed in an expression, but very useful for behavior properties. 
Since there's a conflict , what's the return type for If here? 

Today, we have special casing to allow the mismatch in a behavior property, and look up the parents searching to see if the type is used.  And if not, we skip the type check.  This breaks the type system and is super hacky. For example, it means we can't assert that the value of a node matches the binder's computed type. This has led to crashes like #953, came up in #1032. Coverage gaps in #531. 

Instead, If() should just return Void in this case (where left & right can't match), and then we get a failure at the point that somebody tries to use void.  

We'll also need to be sure we still handle Blank and Error, and errors get propagated.

jas-valgotar:
- Fixes https://github.com/microsoft/Power-Fx/issues/1092 and #843 
- Corresponding PA PR: https://msazure.visualstudio.com/OneAgile/_git/PowerApps-Client/pullrequest/7705536
- Introduces new `void` DType and corresponding `VoidValue` which currently is only produced by the `If` function when argument type mismatches (useful for behavior function) and removes IsArgTypeInconsequential hackery.

Notes:
- `Void` type is denoted by `-`. 
and it has a corresponding `VoidValue` even though it may sound counterintuitive. This keeps things straightforward.
- `Void` type can't be consumed. Ie `void + 1` is invalid.
- `Void` type isn't allowed in the aggregate type. Ie `[void]` or `{field: void}` is invalid.
- `Void` type can't be accepted by any type or coerced to any type.
- Currently, this PR is limited only to the `If()` function, the `Void` type would be leveraged with other functions like `IfError()` & `Switch()` in separate follow-up PRs.

Breaking Changes:

**Things that used to be illegal, that are now legal:**
- These kinds of previously failing cases would pass now,
e.g. `If(true, {x:3}, 1)`. 
But it is safe since the eval of above would be `VoidValue` (practically null). And which can't also be consumed.
Language breaking changes 

**Things that used to be legal and are now illegal:**  None

**Things that used to be legal still is legal, but behave differently:** 
- The expected type of expression would change to `Void` Type.
e.g.
https://github.com/microsoft/Power-Fx/pull/1113/files#diff-9b837be0ccd7fa3b97c4cbdd5e2e022d26543e3b2ae811b0f7fd0fbd906c8dcbR506
Would have returned the type `number` earlier, which is now `Void`.
- Error may change in certain circumstances. (Since now `If` itself doesn't produce the error, rather the consumers do)
https://github.com/microsoft/Power-Fx/pull/1113/files#diff-93bb6d92ab7000b8d30b8fa252ce068cdd0f8b1319cddf68d6a30e8dd5613ae6L25

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -5164,7 +5164,7 @@ namespace Microsoft.PowerFx.Core.Binding
                         {
                             _txb.ErrorContainer.EnsureError(DocumentErrorSeverity.Severe, node.Children[i], TexlStrings.ErrMultipleValuesForField_Name, displayName);
                         }
-                        else if (_txb.GetType(node.Children[i]).IsDeferred)
+                        else if (_txb.GetType(node.Children[i]).IsDeferred || _txb.GetType(node.Children[i]).IsVoid)
                         {
                             _txb.ErrorContainer.EnsureError(DocumentErrorSeverity.Severe, node.Children[i], TexlStrings.ErrRecordDoesNotAcceptThisType);
                         }
@@ -5192,16 +5192,16 @@ namespace Microsoft.PowerFx.Core.Binding
                 {
                     var childType = _txb.GetType(child);
                     isSelfContainedConstant &= _txb.IsSelfContainedConstant(child);
-
-                    if (!childType.IsDeferred && !exprType.IsValid)
+                    var isNotDeferredOrVoid = !childType.IsDeferred && !childType.IsVoid;
+                    if (isNotDeferredOrVoid && !exprType.IsValid)
                     {
                         exprType = childType;
                     }
-                    else if (!childType.IsDeferred && exprType.CanUnionWith(childType))
+                    else if (isNotDeferredOrVoid && exprType.CanUnionWith(childType))
                     {
                         exprType = DType.Union(exprType, childType);
                     }
-                    else if (!childType.IsDeferred && childType.CoercesTo(exprType))
+                    else if (isNotDeferredOrVoid && childType.CoercesTo(exprType))
                     {
                         _txb.SetCoercedType(child, exprType);
                     }
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -5192,16 +5192,18 @@ namespace Microsoft.PowerFx.Core.Binding
                 {
                     var childType = _txb.GetType(child);
                     isSelfContainedConstant &= _txb.IsSelfContainedConstant(child);
-                    var isNotDeferredOrVoid = !childType.IsDeferred && !childType.IsVoid;
-                    if (isNotDeferredOrVoid && !exprType.IsValid)
+
+                    // Deferred and void types are not allowed in tables.
+                    var isChildTypeAllowedInTable = !childType.IsDeferred && !childType.IsVoid;
+                    if (isChildTypeAllowedInTable && !exprType.IsValid)
                     {
                         exprType = childType;
                     }
-                    else if (isNotDeferredOrVoid && exprType.CanUnionWith(childType))
+                    else if (isChildTypeAllowedInTable && exprType.CanUnionWith(childType))
                     {
                         exprType = DType.Union(exprType, childType);
                     }
-                    else if (isNotDeferredOrVoid && childType.CoercesTo(exprType))
+                    else if (isChildTypeAllowedInTable && childType.CoercesTo(exprType))
                     {
                         _txb.SetCoercedType(child, exprType);
                     }
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
@@ -24,6 +24,8 @@ using Microsoft.PowerFx.Core.IR.Nodes;
 using Microsoft.PowerFx.Core.IR.Symbols;
 using Microsoft.PowerFx.Core.Localization;
 using Microsoft.PowerFx.Core.Logging.Trackers;
+using Microsoft.PowerFx.Core.Texl;
+using Microsoft.PowerFx.Core.Texl.Builtins;
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Types.Enums;
 using Microsoft.PowerFx.Core.Utils;
@@ -496,7 +498,8 @@ namespace Microsoft.PowerFx.Core.Functions
                 }
 
                 var type = argTypes[i];
-                if (type.IsError)
+                if (type.IsError || 
+                    type.IsVoid)
                 {
                     errors.EnsureError(args[i], TexlStrings.ErrBadType);
                     fValid = false;
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
@@ -87,7 +87,7 @@ namespace Microsoft.PowerFx
             DName displayDName = default;
             DName varDName = ValidateName(name);
 
-            if (type is VoidType)
+            if (type is Types.Void)
             {
                 throw new NotSupportedException();
             }
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
@@ -87,6 +87,11 @@ namespace Microsoft.PowerFx
             DName displayDName = default;
             DName varDName = ValidateName(name);
 
+            if (type is VoidType)
+            {
+                throw new NotSupportedException();
+            }
+
             if (displayName != null)
             {
                 displayDName = ValidateName(displayName);
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Types/VoidType.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Types/VoidType.cs
@@ -1,0 +1,19 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Types
+{
+    public sealed class VoidType : FormulaType
+    {
+        internal VoidType()
+            : base(DType.Void)
+        {
+        }
+
+        /// <inheritdoc />
+        public override void Visit(ITypeVisitor visitor) => visitor.Visit(this);
+    }
+}
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Types/VoidType.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Types/VoidType.cs
@@ -6,6 +6,9 @@ using Microsoft.PowerFx.Types;
 
 namespace Microsoft.PowerFx.Types
 {
+    /// <summary>
+    /// Special Type which can't be coerced or accepted by any types.
+    /// </summary>
     public sealed class VoidType : FormulaType
     {
         internal VoidType()
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/VoidValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/VoidValue.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.Contracts;
 using System.Text;
 using Microsoft.PowerFx.Core.IR;
 using Microsoft.PowerFx.Types;
@@ -14,11 +15,12 @@ namespace Microsoft.PowerFx.Types
         internal VoidValue(IRContext irContext) 
             : base(irContext)
         {
+            Contract.Assert(irContext.ResultType == FormulaType.Void);
         }
 
         public override void ToExpression(StringBuilder sb, FormulaValueSerializerSettings settings)
         {
-            sb.Append($"\"Void\"");
+            sb.Append($"If(true, {{test:1}}, \"Mismatched args\")");
         }
 
         public override object ToObject()
@@ -28,7 +30,7 @@ namespace Microsoft.PowerFx.Types
 
         public override string ToString()
         {
-            return "Void";
+            return "-";
         }
 
         public override void Visit(IValueVisitor visitor)
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/VoidValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/VoidValue.cs
@@ -16,7 +16,6 @@ namespace Microsoft.PowerFx.Types
         /// Initializes a new instance of the <see cref="VoidValue"/> class.
         /// The void value should be used when the result of an expression is not used or is null.
         /// </summary>
-        /// <param name="irContext"></param>
         internal VoidValue(IRContext irContext) 
             : base(irContext)
         {
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/VoidValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/VoidValue.cs
@@ -20,7 +20,7 @@ namespace Microsoft.PowerFx.Types
 
         public override void ToExpression(StringBuilder sb, FormulaValueSerializerSettings settings)
         {
-            sb.Append($"If(true, {{test:1}}, \"Mismatched args\")");
+            sb.Append($"If(true, {{test:1}}, \"Mismatched args (result of the expression can't be used).\")");
         }
 
         public override object ToObject()
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/VoidValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/VoidValue.cs
@@ -12,6 +12,11 @@ namespace Microsoft.PowerFx.Types
 {
     public class VoidValue : ValidFormulaValue
     {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="VoidValue"/> class.
+        /// The void value should be used when the result of an expression is not used or is null.
+        /// </summary>
+        /// <param name="irContext"></param>
         internal VoidValue(IRContext irContext) 
             : base(irContext)
         {
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/VoidValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/VoidValue.cs
@@ -29,7 +29,7 @@ namespace Microsoft.PowerFx.Types
 
         public override object ToObject()
         {
-            return null;
+            throw new InvalidOperationException();
         }
 
         public override string ToString()
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/VoidValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/VoidValue.cs
@@ -1,0 +1,39 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Types
+{
+    public class VoidValue : ValidFormulaValue
+    {
+        internal VoidValue(IRContext irContext) 
+            : base(irContext)
+        {
+        }
+
+        public override void ToExpression(StringBuilder sb, FormulaValueSerializerSettings settings)
+        {
+            sb.Append($"\"Void\"");
+        }
+
+        public override object ToObject()
+        {
+            return null;
+        }
+
+        public override string ToString()
+        {
+            return "Void";
+        }
+
+        public override void Visit(IValueVisitor visitor)
+        {
+            throw new NotSupportedException();
+        }
+    }
+}
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/If.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/If.cs
@@ -109,21 +109,17 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
                     type = typeSuper;
                     fArgsValid = false;
                 }
-                else if (!type.IsError)
+                else if (!type.IsError && !type.IsVoid)
                 {
                     if (typeArg.CoercesTo(type))
                     {
                         CollectionUtils.Add(ref nodeToCoercedTypeMap, nodeArg, type);
                     }
-                    else if (!isBehavior || !IsArgTypeInconsequential(nodeArg))
+                    else
                     {
-                        errors.EnsureError(
-                            DocumentErrorSeverity.Severe,
-                            nodeArg,
-                            TexlStrings.ErrBadType_ExpectedType_ProvidedType,
-                            type.GetKindString(),
-                            typeArg.GetKindString());
-                        fArgsValid = false;
+                        // If the types are incompatible, the result type is void.
+                        type = DType.Void;
+                        break;
                     }
                 }
                 else if (typeArg.Kind != DKind.Unknown)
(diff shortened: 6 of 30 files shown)


LucGenetier:
✅ No public API change.

MikeStall:
>         private bool IsArgTypeInconsequential(TexlNode arg)

Can we remove this method entirely?
That's the major win for this change. 

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/If.cs:163 in b6b83e5. [](commit_id = b6b83e5f0e65d640e2a7918bd155bf5ffda275ac, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/If.cs
```diff
@@ -111,21 +111,16 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
...
                     fArgsValid = false;
                 }
-                else if (!type.IsError)
+                else if (!type.IsError && !type.IsVoid)
                 {
                     if (typeArg.CoercesTo(type))
                     {
                         CollectionUtils.Add(ref nodeToCoercedTypeMap, nodeArg, type);
                     }
-                    else if (!isBehavior || !IsArgTypeInconsequential(nodeArg))
```
MikeStall, line 120: >e if (!isBehavior || !IsArgTypeInconsequential(nodeArg)) [](http://example.com/codeflow?start=23&length=56)

yes!!

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -281,11 +281,8 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
                                 Kind = ex.ErrorKind
                             });
                     }
-
-                    if (IfFunction.CanCheckIfReturn(func))
-                    {
-                        Contract.Assert(result.IRContext.ResultType == node.IRContext.ResultType || result is ErrorValue || result.IRContext.ResultType is BlankType);
-                    }
+                    
+                    Contract.Assert(result.IRContext.ResultType == node.IRContext.ResultType || result is ErrorValue || result.IRContext.ResultType is BlankType || node.IRContext.ResultType._type.IsVoid);
```
MikeStall, line 285: > || node.IRContext.ResultType._type.IsVoid [](http://example.com/codeflow?start=160&length=42)

Can we avoid this?
Maybe we do need an empty  FormulaValue  that corresponds to Void? That will enforce the void type isn't being used. 
Else we maybe returning the "Real" object and using it even if the binder type is void.

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
```diff
@@ -41,13 +41,12 @@ public void InterpreterTestCase(ExpressionTestCase testCase)
             }
         }
 
-#if false
```
MikeStall, line 44: nit - don't forget to revert and disable this

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -3183,7 +3189,7 @@ public virtual bool CoercesTo(DType typeDest, out bool isSafe, out DType coercio
                                  Boolean.Accepts(this);
                     break;
                 case DKind.String:
-                    doesCoerce = Kind != DKind.Color && Kind != DKind.Control && Kind != DKind.DataEntity && Kind != DKind.OptionSet && Kind != DKind.View && Kind != DKind.Polymorphic && Kind != DKind.File && Kind != DKind.LargeImage;
+                    doesCoerce = Kind != DKind.Color && Kind != DKind.Control && Kind != DKind.DataEntity && Kind != DKind.OptionSet && Kind != DKind.View && Kind != DKind.Polymorphic && Kind != DKind.File && Kind != DKind.LargeImage && Kind != DKind.Void;
```
CarlosFigueiraMSFT, line 3192: >Kind != DKind.Void [](http://example.com/codeflow?start=237&length=18)

Similar to Accepts: Void is special enough that it should have its own treatment at the beginning of this method.

src/libraries/Microsoft.PowerFx.Core/Public/Values/VoidValue.cs
```diff
@@ -0,0 +1,39 @@
...
+    public class VoidValue : ValidFormulaValue
+    {
+        internal VoidValue(IRContext irContext) 
+            : base(irContext)
+        {
+        }
+
+        public override void ToExpression(StringBuilder sb, FormulaValueSerializerSettings settings)
+        {
+            sb.Append($"\"Void\"");
```
CarlosFigueiraMSFT, line 21: >sb.Append($"\"Void\"") [](http://example.com/codeflow?start=12&length=22)

Will this cause one of the round-trip tests to fail? `"Void"` is of type string, not of type Void.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1872,6 +1872,28 @@ private static FormulaValue MaybeWrapRecordValue(FormulaValue result, IRContext
...
             }
+            else if (result is TableValue tableValue && irContext.ResultType is TableType compileTimeTableType)
+            {
+                return CompileTimeTypeWrapperTableValue.AdjustType(compileTimeTableType, tableValue);
+            }
+
+            return result;
+        }
+
+        private static FormulaValue MaybeAdjustToCompieTimeType(FormulaValue result, IRContext irContext)
```
CarlosFigueiraMSFT, line 1883: >Compie [](http://example.com/codeflow?start=49&length=6)

[typo] Compile

src/libraries/Microsoft.PowerFx.Core/Public/Values/VoidValue.cs
jas-valgotar: I am just keeping this for now as a place holder, thinking of tackling it at last.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -5169,16 +5169,16 @@ public override void PostVisit(TableNode node)
                 {
                     var childType = _txb.GetType(child);
                     isSelfContainedConstant &= _txb.IsSelfContainedConstant(child);
-
-                    if (!childType.IsDeferred && !exprType.IsValid)
+                    var isNotDeferredOrVoid = !childType.IsDeferred && !childType.IsVoid;
```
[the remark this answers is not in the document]
jas-valgotar: The pattern is limited to this scope.

jas-valgotar:
>         private bool IsArgTypeInconsequential(TexlNode arg)

Sure thing!

---
In reply to: [1444059623](https://github.com/microsoft/Power-Fx/pull/1113#issuecomment-1444059623) [](http://example.com/codeflow?ancestors=1444059623,1444059623)

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/If.cs:163 in 124e349. [](commit_id = 124e34945f612b810b653b35baec0c675dd5e0aa, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1899,7 +1902,7 @@ public virtual bool Accepts(DType type, out KeyValuePair<string, DType> schemaDi
                         return TreeAccepts(this, TypeTree, type.TypeTree, out schemaDifference, out schemaDifferenceType, exact, useLegacyDateTimeAccepts);
                     }
 
-                    accepts = (IsMultiSelectOptionSet() && TypeTree.GetPairs().First().Value.OptionSetInfo == type.OptionSetInfo) || type.Kind == DKind.Unknown || type.Kind == DKind.Deferred;
+                    accepts = (IsMultiSelectOptionSet() && TypeTree.GetPairs().First().Value.OptionSetInfo == type.OptionSetInfo && type.Kind != DKind.Void) || type.Kind == DKind.Unknown || type.Kind == DKind.Deferred;
```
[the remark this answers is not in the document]
jas-valgotar: updated.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
jas-valgotar: updated.

src/libraries/Microsoft.PowerFx.Core/Public/Values/VoidValue.cs
MikeStall: yes  - ideally, this produces an expression that evaluates to void. 
However, we do have a flag in settings, because sometimes that expression can be very strange. So we have a "Friendly" flag to produce more human readable forms. 

Blank() is an example - technically is typed, so should be produced as If(false, 0); // blank  of number type.

src/libraries/Microsoft.PowerFx.Core/Public/Values/VoidValue.cs
```diff
@@ -0,0 +1,39 @@
...
+using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Types
+{
+    public class VoidValue : ValidFormulaValue
+    {
+        internal VoidValue(IRContext irContext) 
+            : base(irContext)
+        {
```
MikeStall, line 16: >{ [](http://example.com/codeflow?start=8&length=1)

Contract.Assert(irContext.Type == VoidType)

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -281,11 +281,8 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
                                 Kind = ex.ErrorKind
                             });
                     }
-
-                    if (IfFunction.CanCheckIfReturn(func))
-                    {
```
MikeStall, line 286: Also a great win!

src/libraries/Microsoft.PowerFx.Core/Public/Values/VoidValue.cs
jas-valgotar: updated

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.VoidType)

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.VoidType)

src/libraries/Microsoft.PowerFx.Core/Public/Values/VoidValue.cs
jas-valgotar: this is updated now.

CarlosFigueiraMSFT:
> // These test cases are part of the code coverage and are meant to reach IfFunction.IsArgTypeInconsequential

More tests that we should have:

- Validate that errors are propagated through void expressions:
    - `If(Sqrt(-1)<0, { a: 1 }, 2)` --> should result in Error({Kind:ErrorKind.Numeric})
    - `If(1<0, { a: 1 }, Sqrt(-1))` --> should result in Error({Kind:ErrorKind.Numeric})
    - `IfError( If(1<0, { a: 1 }, Sqrt(-1)), $"Error {FirstError.Kind}", "No errors"` - should result in "Error 25" (only possible result types are string)
    - `IsError( If(false, { a: 1 }, 2 ) )` --> should result false
- Integration test, possibly in C#, to validate the blank values (type Void) are propagated:
    - `IsBlank( If(1<0, If(false, 1, {a:2}) )` --> should return true
- Tests with other functions that can take any argument
    - IsBlankOrError

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/If_AllowsSideEffects.txt:4 in 124e349. [](commit_id = 124e34945f612b810b653b35baec0c675dd5e0aa, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
```diff
@@ -498,38 +499,41 @@ public void TexlFunctionTypeSemanticsIf()
...
+                                    symbol);
+            }
+            else
+            {
+                TestBindingErrors(
+                    expression,
+                    TestUtils.DT(expectedType),
+                    symbol);
+            }
         }
```
CarlosFigueiraMSFT, line 538: We should have more negative tests, validating that a Void expression V will fail in:

- Functions that take number: `Abs(V)`
- Functions that take string: `Len(V)`
- Functions that take boolean: `If(V,1,2)`
- Functions that take date/time: `Hour(V)`
    - In the tests above, bonus points if the actual runtime value of is of the expected type, such as `Abs( If(false,{a:1},1 )`
- Records: `{a:V}`
- Tables: `[V]`
- ForAll expressions: `ForAll([1,2,3], V)`

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/If_AllowsSideEffects.txt
```diff
@@ -45,3 +45,9 @@ Blank()
 
 >> If(Value("1")>Value("0"), Blank(); 1/0; Blank(), 3)
 Error({Kind:ErrorKind.Div0})
```
CarlosFigueiraMSFT, line 47: We should have similar tests where Switch or IfError has incompatible results, but things still work. For example:
- `With({a:1},Switch(a, 1, { prop:"complex" }, "scalar"))`
- `IfError({a:1}, false)`

And more complex expressions with those

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.VoidType)

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.VoidType)

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.VoidType)

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.VoidType)

src/libraries/Microsoft.PowerFx.Core/Public/Types/VoidType.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Types
+{
+    /// <summary>
+    /// Special Type which can't be coerced or accepted by any types.
+    /// </summary>
+    public sealed class VoidType : FormulaType
```
bryancroteau-MSFT, line 12: Are we actually constructing these values at runtime and returning them? If so, then technically this type is "Unit", the set of one element.

src/libraries/Microsoft.PowerFx.Core/Public/Types/VoidType.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        internal VoidType()
+            : base(DType.Void)
+        {
+        }
+
+        /// <inheritdoc />
+        public override void Visit(ITypeVisitor visitor) => visitor.Visit(this);
```
bryancroteau-MSFT, line 20: In what instances would we ever want the IR context from a void value? Shouldn't this class be a singleton with a single canonical instance (the Unit) with a NotInSource IRContext?

src/libraries/Microsoft.PowerFx.Core/Public/Types/VoidType.cs
MikeStall: For naming, I prefer "Void" since that is very familiar to C# consumers (who will be using this API).  

We do have a placeholder value that we allocate - since we're focused to have a Value for every return type.

src/libraries/Microsoft.PowerFx.Core/Public/Types/VoidType.cs
bryancroteau-MSFT: Can we not include an IRContext in the constructor then, and pass a NotInSource into the base class constructor?

src/libraries/Microsoft.PowerFx.Core/Public/Types/VoidType.cs
bryancroteau-MSFT: Agreed, Void is a better name for our customers and users.

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.VoidType)

src/libraries/Microsoft.PowerFx.Core/Public/Types/VoidType.cs
jas-valgotar: Updated name to be `Void`

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.Void)

src/libraries/Microsoft.PowerFx.Core/Public/Types/VoidType.cs
bryancroteau-MSFT: Sorry this comment is in the wrong file. It should be down in the VoidValue file.

src/libraries/Microsoft.PowerFx.Core/Public/Values/VoidValue.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Microsoft.PowerFx.Types
+{
+    public class VoidValue : ValidFormulaValue
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="VoidValue"/> class.
+        /// The void value should be used when the result of an expression is not used or is null.
+        /// </summary>
+        internal VoidValue(IRContext irContext)
```
bryancroteau-MSFT, line 19: Would we ever want to be able to recover the IRContext from a Void return value? Should this class be a singleton?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -487,7 +489,9 @@ private bool CheckTypesCore(CheckTypesContext context, TexlNode[] args, DType[]
                 }
 
                 var type = argTypes[i];
-                if (type.IsError)
+                if (type.IsError || 
+                    (type.IsVoid && 
+                    !(Name == BuiltinFunctionsCore.With.Name)))
```
CarlosFigueiraMSFT, line 494: The base class should not know about the derived class. This should probably be handled in the With function itself, saying that it accepts Void values (in its second argument, not the first)

src/libraries/Microsoft.PowerFx.Core/Public/Types/VoidType.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Types
+{
+    /// <summary>
+    /// Special Type which can't be coerced or accepted by any types.
+    /// </summary>
+    public sealed class Void : FormulaType
```
CarlosFigueiraMSFT, line 12: Consider renaming the file to match the class name

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/If.cs
```diff
@@ -104,6 +104,11 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
                 {
                     type = typeSuper;
                 }
+                else if (typeArg.IsVoid)
+                {
+                    type = DType.Void;
+                    break;
+                }
```
CarlosFigueiraMSFT, line 111: This if can be done right after line 96

src/libraries/Microsoft.PowerFx.Core/Public/Values/VoidValue.cs
jas-valgotar: This class needs to extend `ValidFormulaValue` class, it's singleton version can be accessed via `FormulaValue.NewVoid()`

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/If.cs
jas-valgotar: updated!

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
jas-valgotar: updated!

MikeStall (dismissed):

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -286,11 +286,8 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
                                 Kind = ex.ErrorKind
                             });
                     }
-
-                    if (IfFunction.CanCheckIfReturn(func))
```
anderson-joyle, line 290: >if (IfFunction.CanCheckIfReturn(func)) [](http://example.com/codeflow?start=20&length=38)

PR #1179 is introducing Table function to also skip this Assert. No action needed. Just a comment to link one PR to the other.

MikeStall (approved):

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.Void)

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -87,6 +87,11 @@ public ISymbolSlot AddVariable(string name, FormulaType type, bool mutable = fal
             DName displayDName = default;
             DName varDName = ValidateName(name);
 
+            if (type is Types.Void)
```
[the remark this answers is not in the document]
jas-valgotar: yes, void is the only type where this is invalid.

src/libraries/Microsoft.PowerFx.Core/Public/Values/VoidValue.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        public override void ToExpression(StringBuilder sb, FormulaValueSerializerSettings settings)
+        {
+            sb.Append($"If(true, {{test:1}}, \"Mismatched args (result of the expression can't be used).\")");
+        }
+
+        public override object ToObject()
+        {
+            return null;
```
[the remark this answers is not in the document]
jas-valgotar: I think null would be appropriate for void, what else is your recommendation?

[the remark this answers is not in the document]
bryancroteau-MSFT: Is this code path even possible to reach? Should this be an exception?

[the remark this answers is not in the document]
jas-valgotar: If hosting API does ToObject() on result of Eval, then yes! But I agree, the exception would be better choice here for that case!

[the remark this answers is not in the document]
jas-valgotar: updated.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/With.cs
```diff
@@ -44,8 +44,8 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
             Contracts.Assert(args.Length == argTypes.Length);
             Contracts.AssertValue(errors);
 
-            // Base call yields unknown return type, so we set it accordingly below
-            var fArgsValid = base.CheckTypes(context, args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
+            var fArgsValid = base.CheckType(args[0], argTypes[0], ParamTypes[0], errors, SupportCoercionForArg(0), out DType coercionType);
```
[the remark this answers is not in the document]
jas-valgotar: updated to discard.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -5192,16 +5192,16 @@ public override void PostVisit(TableNode node)
...
```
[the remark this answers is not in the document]
jas-valgotar: updated to generic name.

bryancroteau-MSFT (dismissed):

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.Void)

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.Void)

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.Void)

MikeStall (dismissed):

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.Void)

MikeStall (approved):

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.Void)

PR #1118, opened 2023-02-27
example of using table type in custom reflection function

jorisdg:


jas-valgotar (approved):

MikeStall (approved):

LucGenetier:
✅ No public API change.

Issue #1105, opened 2023-02-23
Culture: Text formatting of dates isn't using localized names


MikeStall:
Consider 
`Text(Today(), DateTimeFormat.LongDate, "fr-FR")`

In RecalcEngine, we return with English names:
`"Thursday, 23 February 2023"`

In PowerApps, we return with french names:
![image](https://user-images.githubusercontent.com/1538900/220799690-0c071dfc-49d8-4425-83d6-286d2e2f4dae.png)

MikeStall:
We should also verify this parses. 
DateValue(str, "fr-FR"), DateTimeValue(str, "fr-FR")

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/1119

PR #1119, opened 2023-02-27
Fix culture format of date for Text function and add tests. 

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1105

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
@@ -344,7 +344,7 @@ namespace Microsoft.PowerFx.Functions
                 case "'longtime24'":
                 case "'longtime'":
                 case "'longdate'":
-                    result = new StringValue(irContext, dateTime.ToString(ExpandDateTimeFormatSpecifiers(format, culture)));
+                    result = new StringValue(irContext, dateTime.ToString(ExpandDateTimeFormatSpecifiers(format, culture), culture));
                     break;
                 default:
                     try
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -344,7 +344,7 @@ internal static bool TryExpandDateTimeExcelFormatSpecifiersToStringValue(IRConte
                 case "'longtime24'":
                 case "'longtime'":
                 case "'longdate'":
-                    result = new StringValue(irContext, dateTime.ToString(ExpandDateTimeFormatSpecifiers(format, culture)));
+                    result = new StringValue(irContext, dateTime.ToString(ExpandDateTimeFormatSpecifiers(format, culture), culture));
```
MikeStall, line 347: nit:  can we extra this part to a local (just breaking up the long expression). 
var formatStr = ExpandDateTimeFormatSpecifiers(format, culture);

That also makes it clearer that the ToString() was missing a culture parameter.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text.txt
```diff
@@ -234,4 +234,19 @@ Error({Kind:ErrorKind.Div0})
...
+"5:02:03 CH"
+
+>> Text(DateValue("02/27/2023"), DateTimeFormat.LongDate, "vi-VN")
+"27 Tháng Hai 2023"
+
+>> Text(DateTimeValue("02/27/2023"), DateTimeFormat.LongDateTime, "vi-VN")
+"27 Tháng Hai 2023 12:00:00 SA"
+
+>> Text(Boolean("True"), "", "vi-Vn")
+"true"
```
MikeStall, line 249: is there ever a case where we don't return the literal string "true"? 
And can we double check that PowerApps returns the same values for the new cases.

nguhoa: as offline chat: only different between our and PowerApps are https://github.com/microsoft/Power-Fx/issues/1120.

PR #1121, opened 2023-02-28
Fix delegation for boolean fields and boolean option sets

sagarbakliwal:
Fix delegation problems with -
1) Filter(SharepointDS, BooleanCol And <some other predicate>) which is not delegated as it is looking for And operation in column metadata and it is not present there for sharepoint columns and only at table level.
2) Filter(DS, TextColumn) is marked as delegable which it shouldn't be, so restricting the boolean checks to only boolean optionset and boolean fields (and views too for dotted node)

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/FilterDelegationBase.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/FilterDelegationBase.cs
@@ -104,6 +104,14 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
 
                     case NodeKind.FirstName:
                         {
+                            // Only boolean option sets and boolean fields are allowed to delegate
+                            var nodeDType = binding.GetType(dsNode);
+                            if (!((nodeDType.IsOptionSet && nodeDType.OptionSetInfo != null && nodeDType.OptionSetInfo.IsBooleanValued) || (nodeDType == DType.Boolean && dsNode.Kind != NodeKind.BoolLit))) 
+                            {
+                                SuggestDelegationHint(dsNode, binding);
+                                return false;
+                            }
+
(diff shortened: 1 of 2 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/FilterDelegationBase.cs
```diff
@@ -114,6 +122,16 @@ protected bool IsValidDelegatableFilterPredicateNode(TexlNode dsNode, TexlBindin
 
                     case NodeKind.DottedName:
                         {
+                            // Only boolean option set, boolean fields and views are allowed to delegate
+                            var nodeDType = binding.GetType(dsNode);
+                            if (!((nodeDType.IsOptionSet && nodeDType.OptionSetInfo != null && nodeDType.OptionSetInfo.IsBooleanValued) 
+                                || (nodeDType == DType.Boolean && dsNode.Kind != NodeKind.BoolLit)
```
lesaltzm, line 128: replace the first two lines of this condition with `DType.Boolean.Accepts(nodeDType)`

Same goes for the condition on 109.

lesaltzm: and the one in opdelstrategy

lesaltzm: Whoops, I was thinking of CoercesTo, which is more than you're looking for.

lesaltzm (dismissed):

LucGenetier:
✅ No public API change.

lesaltzm (approved):

PR #1129, opened 2023-02-28
Optimized substitute implementation to stop high memory or system hang issue

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/971
Remove while loop in SubstituteWorker and avoid using string concatenate/substring to optimize memory. Using StringBuilder and replace function.
System now will not hang for high memory query of substitute but will throw exception of (Insufficient memory to continue the execution of the program).


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -827,56 +826,34 @@ private static StringValue SubstituteWorker(EvalVisitor eval, IRContext irContex
...
-                        idx = idx2;
-                    }
-                    else
-                    {
-                        idx = temp.Length + idx2;
-                    }
-
-                    sourceValue = temp + sourceValue;
-                }
+                strBuilder.Replace(match.Value, replacement.Value);
```
MikeStall, line 832: >Replace [](http://example.com/codeflow?start=27&length=7)

This is a big win - .Net's Replace() will handle for us.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Substitute.txt
```diff
@@ -192,4 +192,19 @@ Error({Kind:ErrorKind.InvalidArgument})
 
 //Excel returns "JaneSmith67890TRUE12.00028117FALSE"
 >> Substitute("Jane" & "Smith" & 67890 & true & Blank() & Text(12, "#.000") & Date(1976, 12,23) & true,true,false,2)
-"JaneSmith67890true12.00012/23/1976false"
\ No newline at end of file
+"JaneSmith67890true12.00012/23/1976false"
+
+>> Substitute( "Quarter 1, 2011", "1", "2", 2 )
```
MikeStall, line 197: >>> [](http://example.com/codeflow?start=0&length=2)

Can you add a test case for when instanceNum = 0
(should that be an error?)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -827,56 +826,34 @@ private static StringValue SubstituteWorker(EvalVisitor eval, IRContext irContex
...
             }
             else
             {
```
MikeStall, line 835: >{ [](http://example.com/codeflow?start=12&length=1)

Can you add:

// 0 is an error. This was already enforced by the IR
Contract.Assert(instanceNum > 0);

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -827,56 +826,34 @@ private static StringValue SubstituteWorker(EvalVisitor eval, IRContext irContex
...
                     {
-                        idx = idx2;
+                        break;
                     }
-                    else
+                    
+                    if (--instanceNum == 0)
                     {
-                        idx += match.Value.Length + idx2;
+                        strBuilder.Replace(match.Value, replacement.Value, idx, match.Value.Length);
```
MikeStall, line 848: >match.Value.Length) [](http://example.com/codeflow?start=80&length=19)

what does this parameter do?

Is there a test case that can catch bugs if we pass the wrong value here?

MikeStall:
> >> Substitute("HelloHelloHello", "He", "Je", 2)

one thing I notice is that many of these existing tests have same length for before/after. 
eg, ay be good to add a test like 

>> Substitute("HelloHelloHello", "He", "abc", 2)

- multiple replaces
- lengths are different


Here's another I'd add:
>> Substitute("ababb", "ab", "aba")
"abaabab"





---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Substitute.txt:6 in 4441ba8. [](commit_id = 4441ba8c6d949c395c931fc1c241895c74410d87, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: I used the replace function 

        public StringBuilder Replace(string oldValue, string newValue, int startIndex, int count);

which returns:
        //     A reference to this instance with all instances of oldValue replaced by newValue
        //     in the range from startIndex to startIndex + count - 1.
So count is the length of match string that we want to replace.
The regular test should cover this.

nguhoa:
> >> Substitute("HelloHelloHello", "He", "Je", 2)

added this test as well.

---
In reply to: [1449117751](https://github.com/microsoft/Power-Fx/pull/1129#issuecomment-1449117751) [](http://example.com/codeflow?ancestors=1449117751,1449117751)

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Substitute.txt:6 in 4441ba8. [](commit_id = 4441ba8c6d949c395c931fc1c241895c74410d87, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: yes it handles all for us.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: added this.

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

PR #1131, opened 2023-03-01
Restrict IsEmpty arguments to tables only

CarlosFigueiraMSFT:
IsEmpty is supposed to check if a table is empty or not, but currently it accepts different types of arguments. Since it doesn't make sense to use it for other types, we will restrict it.

Also adding the implementation to the C# interpreter.


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IsEmpty.cs
```diff
@@ -19,7 +19,7 @@ internal sealed class IsEmptyFunction : BuiltinFunction
         public override bool SupportsParamCoercion => false;
 
         public IsEmptyFunction()
-            : base("IsEmpty", TexlStrings.AboutIsEmpty, FunctionCategories.Table | FunctionCategories.Information, DType.Boolean, 0, 1, 1)
+            : base("IsEmpty", TexlStrings.AboutIsEmpty, FunctionCategories.Table | FunctionCategories.Information, DType.Boolean, 0, 1, 1, DType.EmptyTable)
```
Shpakh-MSFT, line 22: >DType.EmptyTable [](http://example.com/codeflow?start=139&length=16)

Will you be writing a converter on paclient side for this change?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1783,6 +1794,23 @@ private static bool IsBlank(FormulaValue arg)
...
+            }
+
+            if (args[0] is TableValue tv)
+            {
+                return new BooleanValue(irContext, tv.Count() == 0);
+            }
+            else
+            {
+                return CommonErrors.RuntimeTypeMismatch(irContext);
+            }
```
Shpakh-MSFT, line 1811: Ideally this should never happen, right?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IsEmpty.cs
CarlosFigueiraMSFT: Yes

lesaltzm (approved):

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1783,6 +1794,23 @@ private static bool IsBlank(FormulaValue arg)
...
+        private static FormulaValue IsEmpty(IRContext irContext, FormulaValue[] args)
+        {
+            if (args[0] is BlankValue)
+            {
+                return new BooleanValue(irContext, false);
+            }
+
+            if (args[0] is TableValue tv)
+            {
+                return new BooleanValue(irContext, tv.Count() == 0);
```
anderson-joyle, line 1806: >tv.Count() [](http://example.com/codeflow?start=51&length=10)

We could have an TableValue.IsEmpty property that calls Row.Any() in case the table has millions of records.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1783,6 +1794,23 @@ private static bool IsBlank(FormulaValue arg)
             return false;
         }
 
+        private static FormulaValue IsEmpty(IRContext irContext, FormulaValue[] args)
+        {
+            if (args[0] is BlankValue)
+            {
+                return new BooleanValue(irContext, false);
```
anderson-joyle, line 1801: >false [](http://example.com/codeflow?start=51&length=5)

If arg[0] is Blank(), shouldn't be TRUE? Usually Blank() means a negative result i.e CountRows(Blank()) is 0. 
 I'm aware this is PA behavior. Just curious on why we decided that Blank() equals to a table with records.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -816,6 +816,17 @@ static Library()
...
             },
+            {
+                BuiltinFunctionsCore.IsEmpty,
+                StandardErrorHandling<FormulaValue>(
+                    BuiltinFunctionsCore.IsEmpty.Name,
+                    expandArguments: NoArgExpansion,
+                    replaceBlankValues: DoNotReplaceBlank,
+                    checkRuntimeTypes: DeferRuntimeTypeChecking,
+                    checkRuntimeValues: DeferRuntimeValueChecking,
+                    returnBehavior: ReturnBehavior.AlwaysEvaluateAndReturnResult,
```
anderson-joyle, line 827: >AlwaysEvaluateAndReturnResult [](http://example.com/codeflow?start=51&length=29)

Since we are returning FALSE when arg[0] is Blank(), does it make sense change this to ReturnFalseIfAnyArgIsBlank?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
CarlosFigueiraMSFT: Good question - I'll raise it up with the team later

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
CarlosFigueiraMSFT: Good point, changed to `!tv.Rows.Any()` so that we don't need to count the number of rows.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
CarlosFigueiraMSFT: Will update it after we decide with the team what should be the value of something like `IsEmpty(If(1<0,[1,2,3]))`

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
CarlosFigueiraMSFT: Updated it to return true in this case.

LucGenetier:
✅ No public API change.

anderson-joyle (approved):
:shipit:

PR #1133, opened 2023-03-01
Removing coercion date to number within Collect.

anderson-joyle:
To support https://github.com/microsoft/Power-Fx-Dataverse/issues/79

Date to number is a special kind of coercion that is inconsistent with other conversions.
These changes are meant to prevent this coercion from within a record.


MikeStall (approved):

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #1142, opened 2023-03-03
Quick fix on disposable object not disposed properly

LucGenetier:


MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1146, opened 2023-03-03
Allow IfError to accept UntypedObjects in the first argument

bryancroteau-MSFT:
In general, this changes Binder to allow normal type checking to occur in the case where:

1. Lambdas exist on one of the overrides to a function
2. The first argument is UntypedObject
3. There is only a single overload, i.e. no scalar overloads at all

The 3rd item in this list was not true before this change. We would give record scope errors in the binder. Normal type checking will fail for all functions where this error would have been seen before, with a single exception: the IfError function, where type checking will succeed.

This is not a breaking change.


LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
🚢

LucGenetier:
✅ No public API change.

PR #1149, opened 2023-03-07
Remove GetUniqueTexlRuntimeName virtual from TexlFunction

CarlosFigueiraMSFT:
This is a method that is only used in Power Apps. Removing it and moving the logic to the Power Apps repository.


lesaltzm (approved):

MikeStall (approved):

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #1150, opened 2023-03-07
Fix Text function handling of GUID values

CarlosFigueiraMSFT:
GUID-to-text conversion shouldn't be done based on .NET rules. This change updates the Text implementation to ignore the second argument for GUID values.


jas-valgotar (approved):

LucGenetier:
✅ No public API change.

PR #1151, opened 2023-03-07
OptionSet serialization rely on Dictionary ordering no longer

anderson-joyle:
Issue #731.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Types/OptionSetValueType.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Types/OptionSetValueType.cs
@@ -2,7 +2,10 @@
 // Licensed under the MIT license.
 
 using System.Collections.Generic;
+using System.Linq;
+using System.Text;
 using Microsoft.PowerFx.Core.Entities;
+using Microsoft.PowerFx.Core.IR;
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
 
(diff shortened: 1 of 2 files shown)


MikeStall (approved):

MikeStall (approved):

src/libraries/Microsoft.PowerFx.Core/Public/Types/OptionSetValueType.cs
```diff
@@ -72,9 +72,11 @@ internal override void DefaultExpressionValue(StringBuilder sb)
 
             if (info != null && info.DisplayNameProvider.LogicalToDisplayPairs.Any())
             {
+                var firstOrderedValue = info.DisplayNameProvider.LogicalToDisplayPairs.OrderBy(x => x.Key.Value).First();
```
lesaltzm, line 75: probably should need to specify StringComparer.Ordinal as the 2nd arg to OrderBy, just to be clear

src/tests/Microsoft.PowerFx.Interpreter.Tests/SerializationTests.cs
```diff
@@ -106,5 +104,13 @@ public BooleanOptionSet(string name, DisplayNameProvider displayNameProvider)
 
             bool IExternalOptionSet.IsBooleanValued => true;
         }
+
+        private class MyOptionSet : OptionSet, IExternalOptionSet
```
lesaltzm, line 108: why do we need this? Can't the test just use `new OptionSet(name, displayNameProvider)`

src/libraries/Microsoft.PowerFx.Core/Public/Types/OptionSetValueType.cs
CarlosFigueiraMSFT, line 75: >var firstOrderedValue = info.DisplayNameProvider.LogicalToDisplayPairs [](http://example.com/codeflow?start=16&length=70)

Suggestion to avoid getting info.DisplayNameProvider.LogicalToDisplayPairs twice: move this to before the if:

    var firstOrdered = info
        ?.DisplayNameProvider
        .LogicalToDisplayPairs
        .OrderBy(x => x.Key.Value)
        .FirstOrDefault();

Then check if firstOrdered is not null (follow the if true branch) or not (follow the else branch)

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/SerializationTests.cs
anderson-joyle: Good point. Removed.

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

Issue #1138, opened 2023-03-02
Crash: Color.Red throws InvalidCastException in IR


MikeStall:
`Color.Red` expression causes crash:

System.InvalidCastException: Unable to cast object of type 'System.Double' to type 'System.UInt32'.
   at Microsoft.PowerFx.Core.IR.IRTranslator.IRTranslatorVisitor.Visit(DottedNameNode node, IRTranslatorContext context)
   at Microsoft.PowerFx.Syntax.DottedNameNode.Accept[TResult,TContext](TexlFunctionalVisitor`2 visitor, TContext context)
   at Microsoft.PowerFx.Core.IR.IRTranslator.Translate(TexlBinding binding)
   at Microsoft.PowerFx.CheckResult.ApplyIR()
   at Microsoft.PowerFx.CheckResultExtensions.GetEvaluator(CheckResult result, StackDepthCounter stackMarker)
   at FxPlayground.Core.Session.TryEvalAsync(String expression, UserInfo user, CancellationToken cancel) in D:\dev\pa2\PowerFxPlayground\src\FxPlaygroundCore\Session.cs:line 391
   at PowerFxService.Controllers.EvaluationController.HandleRequestAsync(String session, RequestBody body) in D:\dev\pa2\PowerFxPlayground\src\WebDemo\Controllers\EvaluationController.cs:line 194

http://aka.ms/fxplayground?session=mike1#expr_Q29sb3IuUmVk

MikeStall:
Found this while looking at #723, may be related.

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/1152

PR #1152, opened 2023-03-07
Add support to use Color.Red/other and add test.

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1138


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Color.txt
```diff
@@ -60,3 +60,12 @@ Blank()
...
 Blank()
+
+>> Color.Red
+RGBA(255,0,0,1)
+
+>> Color.Black
+RGBA(0,0,0,1)
+
+>> Color.ForestGreen
+RGBA(34,139,34,1)
```
anderson-joyle, line 71: Should we add test cases to ColorFade.txt and ColorFadeT.txt? Something like:
`ColorFade(Color.Red, 100%)`
`ColorFade(Color.Black, 100%)`
`ColorFade(Color.ForestGreen, 100%)`

Also:
`ColorFade([Color.Red, Color.Black, Color.ForestGreen], 100%)`

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -525,9 +525,8 @@ public override IntermediateNode Visit(DottedNameNode node, IRTranslatorContext
 
                     if (DType.Color.Accepts(resultType))
                     {
-                        Contracts.Assert(value is uint);
-                        result = new ColorLiteralNode(context.GetIRContext(node), (uint)value);
-                    }
+                        result = new ColorLiteralNode(context.GetIRContext(node), System.Drawing.Color.FromName(node.Right.Name.ToString()));
```
lesaltzm, line 528: The keys are already mapped to specific colors in the enum definition, which is the uint value that was used here previously. I'd suggest using that value rather than relying on c#'s mapping

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Color.txt
nguhoa: Added these tests. thanks.

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
nguhoa: our colorvalue function required input is 'Color' type, not uint. we still need to convert from uint to color, how we do that? system.drawing only convert from int to color.

lesaltzm: See LibraryUnary::OptionSetValueToColor.

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -534,9 +534,9 @@ public override IntermediateNode Visit(DottedNameNode node, IRTranslatorContext
 
                     if (DType.Color.Accepts(resultType))
                     {
-                        Contracts.Assert(value is double);
-                        result = new ColorLiteralNode(context.GetIRContext(node), Convert.ToUInt32((double)value));
-                    }
+                        var uintValue = Convert.ToUInt32(value);
+                        result = new ColorLiteralNode(context.GetIRContext(node), System.Drawing.Color.FromArgb((byte)((uintValue >> 24) & 0xFF), (byte)((uintValue >> 16) & 0xFF), (byte)((uintValue >> 8) & 0xFF), (byte)(uintValue & 0xFF)));
```
MikeStall, line 538: nit - can we put this System.Drawing.Color.FromArgb(...) in its own helper method

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -939,6 +938,16 @@ private IntermediateNode InjectCoercion(IntermediateNode child, IRTranslatorCont
 
                 return new UnaryOpNode(IRContext.NotInSource(FormulaType.Build(toType)), unaryOpKind, child);
             }
+
+            private System.Drawing.Color ConvertToColor(object inputValue)
```
MikeStall, line 942: super nit: static

MikeStall (dismissed):

MikeStall (dismissed):

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -939,6 +939,16 @@ private IntermediateNode InjectCoercion(IntermediateNode child, IRTranslatorCont
 
                 return new UnaryOpNode(IRContext.NotInSource(FormulaType.Build(toType)), unaryOpKind, child);
             }
+
+            private static System.Drawing.Color ConvertToColor(object inputValue)
```
MikeStall, line 943: actually - can this be double? (not object)

PR #1160, opened 2023-03-08
Remove 'accepts' relationship between Guid and string

CarlosFigueiraMSFT:
We need a coercion node between Guid and string values. This removes the 'Accepts' relationship between the two types. It also adds the parameter-less GUID function.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
@@ -1923,7 +1923,6 @@ namespace Microsoft.PowerFx.Core.Types
                         type.Kind == DKind.Blob ||
                         type.Kind == DKind.Unknown ||
                         type.Kind == DKind.Deferred ||
-                        type.Kind == DKind.Guid ||
                         (type.Kind == DKind.Enum && Accepts(type.GetEnumSupertype()));
                     break;
 
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
@@ -902,7 +902,13 @@ namespace Microsoft.PowerFx.Functions
             return new StringValue(irContext, result);
         }
 
-        public static FormulaValue Guid(IRContext irContext, StringValue[] args)
+        public static FormulaValue GuidNoArg(IRContext irContext, FormulaValue[] args)
+        {
+            var guid = System.Guid.NewGuid();
+            return new GuidValue(irContext, guid);
+        }
+
+        public static FormulaValue GuidPure(IRContext irContext, StringValue[] args)
         {
             var text = args[0].Value;
             try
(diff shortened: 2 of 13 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1923,7 +1923,6 @@ public virtual bool Accepts(DType type, out KeyValuePair<string, DType> schemaDi
                         type.Kind == DKind.Blob ||
                         type.Kind == DKind.Unknown ||
                         type.Kind == DKind.Deferred ||
-                        type.Kind == DKind.Guid ||
```
Shpakh-MSFT, line 1926: >  type.Kind == DKind.Guid || [](http://example.com/codeflow?start=22&length=28)

From PA customer POV, this would be a transparent change. Is that accurate?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -902,7 +902,13 @@ public static FormulaValue TrimEnds(IRContext irContext, StringValue[] args)
             return new StringValue(irContext, result);
         }
 
-        public static FormulaValue Guid(IRContext irContext, StringValue[] args)
+        public static FormulaValue GuidNoArg(IRContext irContext, FormulaValue[] args)
+        {
+            var guid = System.Guid.NewGuid();
```
Shpakh-MSFT, line 907: >System.Guid.NewGuid(); [](http://example.com/codeflow?start=23&length=22)

Is there a native support in JS to do so?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/GUID.txt
```diff
@@ -30,4 +30,24 @@ Error({Kind:ErrorKind.InvalidArgument})
 Error({Kind:ErrorKind.InvalidArgument})
 
 >> GUID("thisisastringthatismuchlongerthanaguidshouldbeitisquitealongstringandwillfinderrorswithlongstringsinthisfunction")
-Error({Kind:ErrorKind.InvalidArgument})
\ No newline at end of file
+Error({Kind:ErrorKind.InvalidArgument})
+
+// Coercion to text
+>> If(1<0,"",GUID("C203B79B-B985-42F0-B523-C10EB64387C6"))
```
Shpakh-MSFT, line 36: >If(1<0,"",GUID("C203B79B-B985-42F0-B523-C10EB64387C6")) [](http://example.com/codeflow?start=3&length=55)

Also add
If(1<0,GUID("C203B79B-B985-42F0-B523-C10EB64387C6"),"C203B79B-B985-42F0-B523-C10EB64387C6")

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
CarlosFigueiraMSFT: Correct. But we'll need to update this PR to prevent analysis breaking changes.

CarlosFigueiraMSFT:
Closing this one, more thorough work done at https://github.com/microsoft/Power-Fx/pull/1281

Issue #942, opened 2022-12-31
UDF Cleanup


MikeStall:
Writing UDFs needs to be more robust in face of user errors. 
- Don't crash  
- Report all user errors so that they're actionable. 

Things to cleanup in UDFs:

1. We immediately crash on unrecognized types (whenever GetFromStringOrNull returns null) 
2. We double parse the function body (UDFDefinition has a string body, it should store the TexlNode or even ParseResult). We're going from TexlNode back to String via TexlPretty ... and that injects `"<error>"` into the string contents on parse errors, which then get re-parsed!  This also means that error spans don't align up with the user's original input text. 
4. CheckWrapper - don't do deferred binding. Bind immediately. This may require 2 passes for recursion (first pass to populate signatures, second pass to handle bodies) 
5. Better error handling - ensure we return all errors

PR #1165, opened 2023-03-09
Create a common entry point for named formulas and UDFs

VamsiModem:
In power apps, named formulas and UDFs will coexist in App.Formulas and we need a common entry point to parse both in one go.

Added a new API to ErrorContainer that ensures token has at most a single error.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Parser/ParseNamedFormulasAndUDFResult.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.PowerFx.Core.Errors;
+
+namespace Microsoft.PowerFx.Core.Parser
+{
+    internal class ParseNamedFormulasAndUDFResult
```
bryancroteau-MSFT, line 10: This name is a bit long, even if it's on-the-nose. How about ParseUserDefinitionResult?

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -72,6 +73,17 @@ public static ParseUDFsResult ParseUDFsScript(string script, CultureInfo loc = n
             return parser.ParseUDFs(script);
         }
 
+        public static ParseNamedFormulasAndUDFResult ParseNamedFormulasAndUDFsScript(string script, CultureInfo loc = null)
```
bryancroteau-MSFT, line 76: How about ParseUserDefinitionScript?

src/libraries/Microsoft.PowerFx.Core/Parser/NamedFormula.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Syntax;
+
+namespace Microsoft.PowerFx.Core.Parser
+{
+    internal class NamedFormula
```
bryancroteau-MSFT, line 9: How (and where) was this concept being represented before? Or is the first time we had a class representing named formulas?

src/libraries/Microsoft.PowerFx.Core/Parser/ParseNamedFormulasAndUDFResult.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        internal IEnumerable<UDF> UDFs { get; }
+
+        internal IEnumerable<NamedFormula> NamedFormulas { get; }
+
+        internal IEnumerable<TexlError> Errors { get; }
+
+        internal bool HasErrors { get; }
+
+        public ParseNamedFormulasAndUDFResult(IEnumerable<NamedFormula> namedFormulas, IEnumerable<UDF> uDFs, IEnumerable<TexlError> errors)
```
bryancroteau-MSFT, line 20: Now that we have this concept, do we need the old ParseUDFs result in isolation?

bryancroteau-MSFT (changes requested):

src/libraries/Microsoft.PowerFx.Core/Parser/NamedFormula.cs
VamsiModem: We were using a IEnumerable<KVP<IdentToken, Formula>>

src/libraries/Microsoft.PowerFx.Core/Parser/ParseNamedFormulasAndUDFResult.cs
VamsiModem: We don't for now, but in future if we need a way to parse UDFs separately, we might need that.

LucGenetier:
✅ No public API change.

MikeStall:
UDFs support doesn't work yet and is not ready for consumption in PA. for example, it doesn't give errors properly and it's easy to crash the parser   - See #942

src/libraries/Microsoft.PowerFx.Core/Parser/ParseNamedFormulasAndUDFResult.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Microsoft.PowerFx.Core.Parser
+{
+    internal class ParseUserDefinitionResult
+    {
+        internal IEnumerable<UDF> UDFs { get; }
+
+        internal IEnumerable<NamedFormula> NamedFormulas { get; }
+
+        internal IEnumerable<TexlError> Errors { get; }
```
MikeStall, line 16: >internal IEnumerable<TexlError> Errors { get; } [](http://example.com/codeflow?start=8&length=47)

Can we tell which UDF/Named Formula each error is for?

src/libraries/Microsoft.PowerFx.Core/Parser/ParseNamedFormulasAndUDFResult.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.PowerFx.Core.Errors;
+
+namespace Microsoft.PowerFx.Core.Parser
+{
+    internal class ParseUserDefinitionResult
```
MikeStall, line 10: >internal [](http://example.com/codeflow?start=4&length=8)

We should expect that this class will soon become Public  (like ParseResult). 
It's ok for a public class to have internal methods.

src/libraries/Microsoft.PowerFx.Core/Parser/ParseNamedFormulasAndUDFResult.cs
MikeStall: nit - rename file to match new name

src/libraries/Microsoft.PowerFx.Core/Parser/ParseNamedFormulasAndUDFResult.cs
VamsiModem: No we can't at the moment, we show them based of  spans.

Shpakh-MSFT:
We will need to clean this up. It won't be enabled for consumption yet until it's ready.

---
In reply to: [1465719137](https://github.com/microsoft/Power-Fx/pull/1165#issuecomment-1465719137) [](http://example.com/codeflow?ancestors=1465719137,1465719137)

src/libraries/Microsoft.PowerFx.Core/Parser/ParseNamedFormulasAndUDFResult.cs
Shpakh-MSFT, line 10: >class [](http://example.com/codeflow?start=13&length=5)

sealed

src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormulasAndUDFs.cs
```diff
@@ -0,0 +1,45 @@
...
+using Microsoft.PowerFx.Core.Errors;
+using Microsoft.PowerFx.Core.Parser;
+using Microsoft.PowerFx.Core.Utils;
+
+namespace Microsoft.PowerFx.Syntax
+{
+    /// <summary>
+    /// This encapsulates a named formula and user defined functions: its original script, the parsed result, and any parse errors.
+    /// </summary>
+    internal class NamedFormulasAndUDFs
```
Shpakh-MSFT, line 18: >class [](http://example.com/codeflow?start=13&length=5)

sealed

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -403,6 +403,23 @@ public virtual bool CheckTypes(CheckTypesContext context, TexlNode[] args, DType
             return CheckTypesCore(context, args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
         }
 
+        /// <summary>
+        /// Perform sub-expression type checking and produce a return type for the function declaration, this is only applicable for UDFs.
+        /// </summary>
+        public virtual bool CheckTypesOnDeclaration(CheckTypesContext context, TexlNode[] args, DType[] argTypes, DType returnType, IErrorContainer errors, out DType actualReturnType)
```
bryancroteau-MSFT, line 409: These virtuals should not be here. They need to be in the UesrDefinedTexlFunction file.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -403,6 +403,23 @@ public virtual bool CheckTypes(CheckTypesContext context, TexlNode[] args, DType
...
+        {
+            actualReturnType = DType.Unknown;
```
bryancroteau-MSFT, line 411: Why would we need to set the actual return type here? Isn't the body return type calculated elsewhere?

src/libraries/Microsoft.PowerFx.Core/Syntax/NamedFormulasAndUDFs.cs
```diff
@@ -0,0 +1,45 @@
...
+using Microsoft.PowerFx.Core.Errors;
+using Microsoft.PowerFx.Core.Parser;
+using Microsoft.PowerFx.Core.Utils;
+
+namespace Microsoft.PowerFx.Syntax
+{
+    /// <summary>
+    /// This encapsulates a named formula and user defined functions: its original script, the parsed result, and any parse errors.
+    /// </summary>
+    internal sealed class NamedFormulasAndUDFs
```
bryancroteau-MSFT, line 18: Rename this to UserDefinitions

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -403,6 +403,23 @@ public virtual bool CheckTypes(CheckTypesContext context, TexlNode[] args, DType
...
+
+            return false;
+        }
+
+        /// <summary>
+        /// Perform expression-level semantics checks which require a binding, this is only applicable for UDFs.
+        /// </summary>
+        public virtual void CheckSemanticsOnDeclaration(TexlBinding binding, TexlNode[] args, DType[] argTypes, IErrorContainer errors)
```
bryancroteau-MSFT, line 419: Where are these new functions being called from?

bryancroteau-MSFT (changes requested):

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
VamsiModem: Yeah right, since its an out parm, we either assign null or Unknown

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
VamsiModem: Moved them to a new like we discussed offline

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
bryancroteau-MSFT: This question was never answered. Something needs to be calling these functions from within PowerFx to perform validation.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
bryancroteau-MSFT: Why does this out param even exist? What is it doing?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
VamsiModem: Sorry, thought it's not applicable anymore. These function will be called from PAClient

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
VamsiModem: Once the body is bound, we will know the actual return type and we are returning it if the call wants to do anything with it.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -0,0 +1,234 @@
...
+        private readonly bool _isImperative;
+        private readonly IdentToken _returnTypeToken;
+        private readonly ISet<UDFArg> _args;
+
+        public TexlNode UdfBody { get; }
+
+        public override bool IsSelfContained => !_isImperative;
+
+        public UserDefinedFunction(string name, FormulaType returnType, TexlNode udfBody, bool isImperative, params FormulaType[] paramTypes)
+            : base(DPath.Root, name, name, SG("Custom func " + name), FunctionCategories.MathAndStat, returnType._type, 0, paramTypes.Length, paramTypes.Length, Array.ConvertAll(paramTypes, x => x._type))
```
bryancroteau-MSFT, line 38: What does MathAndStat mean here?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -0,0 +1,234 @@
...
+        {
+            UdfBody = udfBody;
+            _isImperative = isImperative;
+        }
+
+        public UserDefinedFunction(DName udfName, IdentToken returnType, TexlNode body, bool isImperative, ISet<UDFArg> args)
```
bryancroteau-MSFT, line 44: Why are there 2 constructors? What do these 2 constructors mean?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
VamsiModem: This is something I have mentioned in the design doc, MathAndStat was used when UDFs were implement initially, but we may need a new category for UDFs.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
VamsiModem: oops not need, cleaned it up.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
VamsiModem: Added a new category 'UserDefined'

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -0,0 +1,229 @@
...
+    {
+        private readonly bool _isImperative;
+        private readonly IdentToken _returnTypeToken;
+        private readonly ISet<UDFArg> _args;
+
+        public TexlNode UdfBody { get; }
+
+        public override bool IsSelfContained => !_isImperative;
+
+        public UserDefinedFunction(string name, IdentToken returnType, TexlNode body, bool isImperative, ISet<UDFArg> args)
```
bryancroteau-MSFT, line 38: nit: returnTypeToken

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -0,0 +1,229 @@
...
+
+        private RecordType CheckParameters(out List<TexlError> errors)
+        {
+            var record = RecordType.Empty();
+            var argsAlreadySeen = new HashSet<string>();
+            errors = new List<TexlError>();
+
+            foreach (var arg in _args)
+            {
+                if (!argsAlreadySeen.Add(arg.VarIdent.Name))
```
bryancroteau-MSFT, line 77: This requires knowledge of how the Add function works and what the return value means. It would be better to call Contains on the hash set before trying to add anything.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -0,0 +1,229 @@
...
+        }
+
+        private static void CheckParameters(IEnumerable<UDFArg> args, IErrorContainer errorContainer)
+        {
+            foreach (var arg in args)
+            {
+                var argNameToken = arg.VarIdent;
+                var argTypeToken = arg.VarType;
+
+                if (argTypeToken.GetFormulaType()._type.Kind.Equals(DType.Unknown.Kind))
```
bryancroteau-MSFT, line 131: We also need to guarantee that the parameter type is supported right? For example, if the parameter type is Void or some other unsupported type, we should check here. I know that the parser will also do a similar check, however we should be able to use this API in contexts that don't involve the Parser.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -0,0 +1,229 @@
...
+
+        /// <summary>
+        /// NameResolver that combines global named resolver and params for user defined function.
+        /// </summary>
+        private class UserDefinitionsNameResolver : INameResolver
+        {
+            private readonly INameResolver _globalNameResolver;
+            private readonly INameResolver _parameterNamedResolver;
+
+            public static INameResolver Merge(INameResolver globalNameResolver, RecordType parametersRecordType)
```
bryancroteau-MSFT, line 156: Where is the list of all visible UDFs? The function body may reference UDFs that were defined before or after this one.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -0,0 +1,229 @@
...
+            }
+
+            if (documentBinderGlue is null)
+            {
+                throw new ArgumentNullException(nameof(documentBinderGlue));
+            }
+
+            var parametersRecordType = CheckParameters(out var errors);
+            var bindingConfig = new BindingConfig(this._isImperative);
+            var binding = TexlBinding.Run(documentBinderGlue, UdfBody, UserDefinitionsNameResolver.Merge(nameResolver, parametersRecordType), bindingConfig);
```
bryancroteau-MSFT, line 60: How is it possible to perform type checking on the body unless we already know the type of all the UDFs in this batch? Shouldn't the args and return type checks happen first, then a new name resolver is created with the correct signatures, then that new name resolver would be used to type check the body?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -0,0 +1,229 @@
...
+
+            CheckTypesOnDeclaration(binding.CheckTypesContext, _args, _returnTypeToken, binding.ResultType, binding.ErrorContainer);
+            userDefinitionSemanticsHandler?.CheckSemanticsOnDeclaration(binding, _args, binding.ErrorContainer);
```
bryancroteau-MSFT, line 63: Should this be a list of handlers instead of a single handler?

bryancroteau-MSFT (changes requested):

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -0,0 +1,229 @@
...
+
+        public UserDefinedFunction(string name, IdentToken returnType, TexlNode body, bool isImperative, ISet<UDFArg> args)
+        : base(DPath.Root, name, name, SG("Custom func " + name), FunctionCategories.UserDefined, returnType.GetFormulaType()._type, 0, args.Count, args.Count, args.Select(a => a.VarType.GetFormulaType()._type).ToArray())
+        {
+            this._returnTypeToken = returnType;
+            this._args = args;
+            this._isImperative = isImperative;
+        }
+
+        public TexlBinding Bind(INameResolver nameResolver, IBinderGlue documentBinderGlue, IUserDefinitionSemanticsHandler userDefinitionSemanticsHandler = null)
```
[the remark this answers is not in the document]
VamsiModem: Yes!

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
VamsiModem: GlobalNameResolver will be able to resolve other UDFs, by the time bind is called, all the UDFs were already registered by the host.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -0,0 +1,229 @@
...
+                var argTypeToken = arg.VarType;
+
+                if (argTypeToken.GetFormulaType()._type.Kind.Equals(DType.Unknown.Kind))
+                {
+                    errorContainer.EnsureError(DocumentErrorSeverity.Severe, argTypeToken, TexlStrings.ErrUDF_UnknownType, argTypeToken.Name);
+                }
+            }
+        }
+
+        public static StringGetter SG(string text)
```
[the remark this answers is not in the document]
VamsiModem: Not really sure, I got this from other functions.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
bryancroteau-MSFT: The host should not need to do any work to use this new API.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
VamsiModem: Yup, this is done in CheckTypesOnDeclaration

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
VamsiModem: We already know the type of each UDF but that type may not match with the body or it may be a unknown type. There may be cases where a() depends b() and type of b() is unknown, in this  case dataflow on the host must re-bind all the dependencies of b().

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
VamsiModem: I don't see a need for this at the moment, can you think of any other use case other than sematic checks?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
VamsiModem: Host is not doing any work to use the new API, Host calls the new API, that will parse all the named formulas and UDFs. It's now the host's responsibility to register the functions with it's own name resolver. During analysis we bind each UDF.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
bryancroteau-MSFT: No, this is unacceptable. We will bind all UDFs together as a batch and the host will never need to rebind or care about the order in which the UDFs are bound. We have all the UDF signatures already after we parse. We need to type check the signatures independently from the body, make a new name resolver with just the signatures, then bind the bodies using that name resolver.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
VamsiModem: Discussed offline with Bryan and made some changes as per the discussion.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -0,0 +1,199 @@
...
+                    return functionsFromGlobalNameResolver;
+                }
+                else
+                {
+                    var functions = _functionNameResolver.LookupFunctions(theNamespace, name, localeInvariant);
+                    return functions.Any() ? functions : functionsFromGlobalNameResolver;
+                }
+            }
+
+            public IEnumerable<TexlFunction> LookupFunctionsInNamespace(DPath nameSpace)
```
bryancroteau-MSFT, line 159: Nit: is this code identical to the code in the previous function? If so can they be shared?

VamsiModem: No they are not identical.

LucGenetier:
✅ No public API change.

bryancroteau-MSFT (approved):

LucGenetier:
✅ No public API change.

PR #1168, opened 2023-03-09
Fix IR context return type for blank values in standard error handling

CarlosFigueiraMSFT:
The StandardErrorHandling class handling of blank values is overriding the return value from the IR context, causing an error in scenarios where the actual type is important. This fixes it.


LucGenetier:
✅ No public API change.

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1169, opened 2023-03-09
Fix table coercion and Blank()/0 places for Dec2Hex

anhtruong-MSFT:
This fixes the table coercion when only the first argument is given and fixes inconsistent results when places is Blank() and 0


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Dec2Hex.txt
```diff
@@ -43,8 +43,11 @@
 >> Dec2Hex(45780)
 "B2D4"
 
+>> Dec2Hex(45780, Blank())
+Error({Kind:ErrorKind.Numeric})
+
 >> Dec2Hex(45780, 0)
```
Shpakh-MSFT, line 49: Dec2Hex(Blank(), 2)

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Dec2Hex.txt
```diff
@@ -177,12 +180,21 @@ Table({Value:"1"},{Value:"0"},{Value:Error({Kind:ErrorKind.Numeric})},{Value:"2"
 >> Dec2Hex([10,11,"12"], [2,5,10])
 Table({Value:"0A"},{Value:"0000B"},{Value:"000000000C"})
 
+>> Dec2Hex(["10","11","12"])
+Table({Value:"A"},{Value:"B"},{Value:"C"})
+
+>> Dec2Hex(["10","11","12"], [2,5,10])
```
Shpakh-MSFT, line 186: >["10","11","12"] [](http://example.com/codeflow?start=11&length=16)

Any testcase where first arg is not a valid numeric table type?

Dec2Hex([{A:10}], 3)

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Dec2Hex.txt
anhtruong-MSFT: This kind of test can be found in Dec2HexT_ConsistentOneColumnTableResult.txt

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Hex2Dec.cs
```diff
@@ -55,5 +59,33 @@ public override string GetUniqueTexlRuntimeName(bool isPrefetching = false)
...
+            Contracts.Assert(args.Length == 1);
+            Contracts.AssertValue(errors);
+            Contracts.Assert(MinArity <= args.Length && args.Length <= MaxArity);
+
+            var fValid = base.CheckTypes(context, args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
+            Contracts.Assert(returnType.IsTable);
+            Contracts.Assert(!fValid || returnType.IsColumn);
+            var type = argTypes[0];
+            var arg = args[0];
+            if (type.IsTable)
```
jas-valgotar, line 78: >if (type.IsTable) [](http://example.com/codeflow?start=12&length=17)

I think you don't need this If else, `base.CheckTypes` would have already validated this.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Dec2HexT_ConsistentOneColumnTableResultDisabled.txt
```diff
@@ -0,0 +1,8 @@
+﻿>> Dec2Hex([{A:2},{A:3}])
+Errors: Error 8-21: Invalid argument type (Table). Expecting a Number value instead.
+
+>> Dec2Hex([{A: 12}])
+Errors: Error 8-17: Invalid argument type (Table). Expecting a Number value instead.
+
+>> Dec2Hex([{A: 12}], 3)
+Errors: Error 8-17: Invalid argument type (Table). Expecting a Number value instead.
```
CarlosFigueiraMSFT, line 8: Instead of splitting the test in two files, you can merge them into a single one, and use the Table function instead of the `[...]` notation.

CarlosFigueiraMSFT:
> #SETUP: ConsistentOneColumnTableResult, TableSyntaxDoesntWrapRecords

The Dec2HexT/Hex2DecT functions don't change based on this feature, so there's no need to have two test files.

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Dec2HexT_ConsistentOneColumnTableResult.txt:1 in 482132a. [](commit_id = 482132aeae14fe7169b85e117a1a521ccd31c215, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Dec2Hex.txt
```diff
@@ -177,16 +183,31 @@ Table({Value:"1"},{Value:"0"},{Value:Error({Kind:ErrorKind.Numeric})},{Value:"2"
...
 Table({Value:"A"},{Value:"FF"},{Value:"FFFFFFFFF0"})
 
 >> Dec2Hex(Table({a:10},{a:11},{a:12}), Table({b:2},{b:5},{b:10}))
 Table({Value:"0A"},{Value:"0000B"},{Value:"000000000C"})
+
+>> Dec2Hex(Table({A: 12}), 3)
+Table({Value:"00C"})
+
+>> Dec2Hex(Table({A: 12, B:13}))
+Errors: Error 8-28: Invalid argument type (Table). Expecting a Number value instead.
```
CarlosFigueiraMSFT, line 213: >Invalid argument type (Table). Expecting a Number value instead. [](http://example.com/codeflow?start=20&length=64)

This error message is misleading. The problem is that we require a one-column table. See the tests for Hex2Dec - they have a good error message

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Dec2Hex.txt
CarlosFigueiraMSFT, line 213: >instead [](http://example.com/codeflow?start=76&length=7)

It would be good to have a negative test case with other invalid types as well:

- Dec2Hex(Table({a:[1]},{a:[2]}))
- Dec2Hex(Table({a:{b:1}},{a:{b:2}}))

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Hex2Dec.txt
```diff
@@ -81,3 +81,14 @@ Error({Kind:ErrorKind.Numeric})
 >> Hex2Dec(["3","E","I","B"])
 Table({Value:3},{Value:14},{Value:Error({Kind:ErrorKind.Numeric})},{Value:11})
 
+>> Hex2Dec(Table({a:"A"}))
```
CarlosFigueiraMSFT, line 84: More tests with different types:

    > Hex2Dec(Table({a:1},{a:10},{a:100}))
    Table({Value:1},{Value:16},{Value:256})

    > Hex2Dec(["a", 10, "b"]))
    Table({Value:10},{Value:16},{Value:11})

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Hex2Dec.txt
```diff
@@ -81,3 +81,14 @@ Error({Kind:ErrorKind.Numeric})
...
+Table({Value:10})
+
+>> Hex2Dec(Table({a:"A"},{a:"12"},{a:"FFFF"}))
+Table({Value:10},{Value:18},{Value:65535})
+
+>> Hex2Dec(Table({a:"A",b:"B"}))
+Errors: Error 0-29: The function 'Hex2Dec' has some invalid arguments.|Error 8-28: Invalid schema, expected a one-column table.
+
+>> Hex2Dec(Table({A:"dead", B: "beef"},{A:"beef",B:"dead"}))
+Errors: Error 0-57: The function 'Hex2Dec' has some invalid arguments.|Error 8-56: Invalid schema, expected a one-column table.
```
CarlosFigueiraMSFT, line 94: This is the same scenario as the previous test, can be removed.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Dec2Hex.txt
anhtruong-MSFT: per offline discussion, we need to keep this as of now we don't have a way to pick the right overload function when the function doesn't specify default argument types like Dec2Hex, Power, Log, etc

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Dec2Hex.txt
```diff
@@ -177,16 +183,37 @@ Table({Value:"1"},{Value:"0"},{Value:Error({Kind:ErrorKind.Numeric})},{Value:"2"
...
+
+>> Dec2Hex(Table({a:[1]},{a:[2]}))
+Errors: Error 8-30: Invalid argument type (Table). Expecting a Number value instead.
+
+>> Dec2Hex(Table({a:{b:1}},{a:{b:2}}))
+Errors: Error 8-34: Invalid argument type (Table). Expecting a Number value instead.
```
jas-valgotar, line 219: Do we have a test case where either of 2 arg is Blank Table and Blank Scaler?

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -940,15 +940,19 @@ private static FormulaValue Dec2Hex(IRContext irContext, NumberValue[] args)
...
+            {
+                places = (int)Math.Floor(args[1].Value);
+            }
 
             if (number < minNumber || number > maxNumber)
             {
                 return CommonErrors.OverflowError(irContext);
             }
 
             // places need to be non-negative and 10 or less
```
CarlosFigueiraMSFT, line 954: >non-negative [](http://example.com/codeflow?start=33&length=12)

positive - 0 is wrong

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #1170, opened 2023-03-09
Add features to enable Async and Impure node delegation

jelopezf:
This feature looks to enable async and impure node delegation as long as Features are enabled for it.

Additionally, removes the ICustomDelegationFunction interface in favor of overriding the delegation validation strategy in some functions to always delegate async calls of the function for appropriate functions (i.e. current behavior of AsType or User function in Power Apps).


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -95,6 +95,12 @@ internal abstract class TexlFunction : IFunction
         // Return true if the function is asynchronous, false otherwise.
         public virtual bool IsAsync => false;
 
+        // Return true if the function is a delegatable asynchronous, false otherwise.
+        // This means that the function should be a valid node for delegation regardless
+        // of the AllowAsyncDelegation feature, and only means that it'll be a valid node when used
+        // as the left hand side of a dotted name node (i.e. the User function in Power Apps)
+        public virtual bool IsDelegatableAsync => false;
```
Shpakh-MSFT, line 102: >IsDelegatableAsync [](http://example.com/codeflow?start=28&length=18)

Why do we need separate flag? Isn't it a combination of IsAsync and IsDelegatable?

Shpakh-MSFT:
>             }

Suggestion: This logic should probably be present in 
DelegationValidationStrategy.IsValidDottedNameNode

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/DelegationValidationStrategy.cs:372 in 9f8b53f. [](commit_id = 9f8b53fe3588cf0cc997324ddafdfa8476130a96, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/DelegationValidationStrategy.cs
```diff
@@ -412,12 +413,24 @@ protected bool IsValidNode(TexlNode node, TexlBinding binding)
...
 
-        private bool IsUserCallNodeDelegable(TexlNode node, TexlBinding binding)
+        protected bool IsValidDelegatableAsyncNode(TexlNode node, TexlBinding binding)
         {
-            if ((node is DottedNameNode)
-                && (node.AsDottedName().Left is CallNode)
-                && (binding.GetInfo(node.AsDottedName().Left.AsCall()).Function is ICustomDelegationFunction customDelFunc)
-                && customDelFunc.IsUserCallNodeDelegable())
+            // Dotted Name node with left hand delegatable async call
+            if ((node is DottedNameNode) && (node.AsDottedName().Left is CallNode))
```
Shpakh-MSFT, line 419: >(node is DottedNameNode) && (node.AsDottedName().Left is CallNode) [](http://example.com/codeflow?start=16&length=66)

This should be handled in IsValidDottedNameNode in DottedNamedValidationStrategy.

If LHS is call node then you just need to call IsValidCallNode on dottednamenode.Left and handle this there.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
jelopezf: What do you think a good name would be for an async function that we always allow delegation for?

Shpakh-MSFT: We already have a virtual for it.
IsServerDelegatable

Why not just use that?

jelopezf: Because IsServerDelegatable indicates the function itself can be delegated in the query, what this is meant to indicate is that async nodes that are made async specifically because of this function should not break delegation (as we presently do). The idea is that this will be standard behavior at some point but currently we only allow for delegating calls where predicate depends on User() function response. We want to keep this functionality while we test and slowly roll out full async delegation for other functions.

jelopezf: So basically not so much that it's delegable or not it's more so should further async nodes no longer be delegable because this is an async function.

src/libraries/Microsoft.PowerFx.Core/Public/Config/Features.cs
```diff
@@ -45,6 +45,18 @@ public enum Features : int
         /// </summary>
         StronglyTypedBuiltinEnums = 0x10,
 
+        /// <summary>
+        /// Allow delegation for async calls (delegate using awaited call result).
+        /// Added on 1st March 2023.
+        /// </summary>
+        AllowAsyncDelegation = 0x10,
```
ian-legler, line 52: Looks like you will need to increase this

anderson-joyle:
Are we planning to add test cases for these changes?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
jelopezf: Decided on overwriting the CallNodeValidationStrategy for the function instead.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AsType.cs
```diff
@@ -112,6 +113,26 @@ protected override bool RequiresPagedDataForParamCore(TexlNode[] args, int param
             // For the second argument, we need only metadata. No actual data from datasource is required.
             return paramIndex != 1;
         }
+
+        public override ICallNodeDelegatableNodeValidationStrategy GetCallNodeDelegationStrategy()
+        {
+            return base.GetCallNodeDelegationStrategy();
```
Shpakh-MSFT, line 119: >base.GetCallNodeDelegationStrategy() [](http://example.com/codeflow?start=19&length=36)

This needs to be AsTypeCallNodeDelegationStrategy, right?

src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/DelegationValidationStrategy.cs
```diff
@@ -104,18 +104,31 @@ private bool IsValidRowScopedDottedNameNode(DottedNameNode node, TexlBinding bin
...
-            if (node.Left.Kind == NodeKind.Call && binding.GetInfo(node.Left as CallNode).Function is AsTypeFunction)
+            return node.Left.Kind == NodeKind.FirstName;
+        }
+
+        private bool IsValidOptionSetOrViewDottedNameNode(DottedNameNode node, TexlBinding binding)
+        {
+            var leftType = binding.GetType(node.Left);
+            var nodeType = binding.GetType(node);
+
+            if ((leftType?.Kind == DKind.OptionSet && nodeType?.Kind == DKind.OptionSetValue) || (leftType?.Kind == DKind.View && nodeType?.Kind == DKind.ViewValue))
```
Shpakh-MSFT, line 125: >if ((leftTyp [](http://example.com/codeflow?start=12&length=12)

Instead add early return null check

if(leftType == null || nodeType == null)
   return false;

return (leftType.Kind == DKind.OptionSet && nodeType.Kind == DKind.OptionSetValue) || (leftType.Kind == DKind.View && nodeType.Kind == DKind.ViewValue));

src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/DelegationValidationStrategy.cs
```diff
@@ -344,84 +371,42 @@ public bool IsValidCallNode(CallNode node, TexlBinding binding, OperationCapabil
...
-        {
-            if ((node is DottedNameNode)
-                && (node.AsDottedName().Left is CallNode)
-                && (binding.GetInfo(node.AsDottedName().Left.AsCall()).Function is ICustomDelegationFunction customDelFunc)
-                && customDelFunc.IsUserCallNodeDelegable())
+            if (isAsync && !binding.Features.HasFlag(Features.AllowAsyncDelegation))
             {
-                return true;
+                TrackingProvider.Instance.SetDelegationTrackerStatus(DelegationStatus.AsyncPredicate, node, binding, Function);
             }
```
Shpakh-MSFT, line 405: I would simplify this even

if (binding.Features.HasFlag(Features.AllowAsyncDelegation))
{
    if(((node is CallNode) || 
         (node is FirstNameNode) || 
         (node is DottedNameNode)))
      { 
           return true;
       }
}
else
{
      if(isAsync)               TrackingProvider.Instance.SetDelegationTrackerStatus(DelegationStatus.AsyncPredicate, node, binding, Function);
}

src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/DelegationValidationStrategy.cs
```diff
@@ -223,7 +237,7 @@ public bool IsValidFirstNameNode(FirstNameNode node, TexlBinding binding, IOpDel
             Contracts.AssertValueOrNull(opDelStrategy);
 
             var isRowScoped = binding.IsRowScope(node);
-            var isValid = IsValidNode(node, binding);
+            var isValid = IsValidAsyncOrImpureNode(node, binding);
```
Shpakh-MSFT, line 240: >IsValidAsyncOrImpureNode [](http://example.com/codeflow?start=26&length=24)

IsValid would return true for Non-Async FirstNameNode before but that's no longer the case for IsValidAsyncOrImpureNode due to feature flag check AllowAsyncDelegation. You should fix that.

jelopezf: yea, I already fixed it in my local build, but I've been testing against the unit tests in PowerApps-Client so I haven't pushed the latest changes

src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/DelegationValidationStrategy.cs
```diff
@@ -313,18 +327,31 @@ protected bool IsDelegatableColumnNode(FirstNameNode node, TexlBinding binding,
             return true;
         }
 
-        public bool IsValidCallNode(CallNode node, TexlBinding binding, OperationCapabilityMetadata metadata)
+        public virtual bool IsValidCallNode(CallNode node, TexlBinding binding, OperationCapabilityMetadata metadata)
+        {
+            // Functions may have their specific CallNodeDelegationStrategies (i.e. AsType, User)
+            // so, if available, we need to ensure we use their specific delegation strategy.
+            var function = binding.GetInfo(node).Function;
+            return function?.GetCallNodeDelegationStrategy().IsValidCallNode(node, binding, metadata) ?? IsValidCallNodeInternal(node, binding, metadata, false);
```
Shpakh-MSFT, line 335: >function? [](http://example.com/codeflow?start=19&length=9)

Function can't be null here.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AsType.cs
```diff
@@ -112,6 +113,26 @@ protected override bool RequiresPagedDataForParamCore(TexlNode[] args, int param
...
+        public AsTypeCallNodeDelegationStrategy(TexlFunction function)
+            : base(function)
+        {
+        }
+
+        public override bool IsValidCallNode(CallNode node, TexlBinding binding, OperationCapabilityMetadata metadata)
+        {
+            // AsType CallNodeDelegationStrategy should match default IsValidCallNode from
+            // DelegationValidationStrategy but skips only the !IsBlockScopedConstant check.
+            return IsValidCallNodeInternal(node, binding, metadata, true);
```
Shpakh-MSFT, line 134: >IsValidCallNodeInternal [](http://example.com/codeflow?start=19&length=23)

I don't think this is correct. You want to skip IsValidAsyncOrImpureNode check for As function.

src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/DelegationValidationStrategy.cs
```diff
@@ -313,18 +327,31 @@ protected bool IsDelegatableColumnNode(FirstNameNode node, TexlBinding binding,
...
             {
                 SuggestDelegationHint(node, binding);
                 return false;
             }
 
+            if (!allowNonBlockScopedConstant && !binding.IsBlockScopedConstant(node))
+            {
+                return false;
+            }
+
```
Shpakh-MSFT, line 354: Previously for !binding.IsBlockScopedConstant(node) non-async/pure function, IsValid returned true. I am not sure why that was the case.

src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/DelegationValidationStrategy.cs
Shpakh-MSFT: Why it would be the sane logic unless I am missing something.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/DelegationValidationStrategy.cs
```diff
@@ -344,84 +403,51 @@ public bool IsValidCallNode(CallNode node, TexlBinding binding, OperationCapabil
...
-                SuggestDelegationHintAndAddTelemetryMessage(node, binding, telemetryMessage);
-
-                if (isAsync)
+                // If the feature is enabled, enable delegation for
+                // async call, first name and dotted name nodes.
+                if (binding.Features.HasFlag(Features.AllowAsyncDelegation))
                 {
-                    TrackingProvider.Instance.SetDelegationTrackerStatus(DelegationStatus.AsyncPredicate, node, binding, Function);
+                    return (node is CallNode) || (node is FirstNameNode) || (node is DottedNameNode);
                 }
```
Shpakh-MSFT, line 436: Do you need to add delegation hit for following scenario

isAync && !isPure 

In this case, currently you return true without checking the value of Features.AllowImpureNodeDelegation). Just want to confirm if this is the behavior that you want.

src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/DelegationValidationStrategy.cs
```diff
@@ -313,16 +352,36 @@ protected bool IsDelegatableColumnNode(FirstNameNode node, TexlBinding binding,
...
-        public bool IsValidCallNode(CallNode node, TexlBinding binding, OperationCapabilityMetadata metadata)
+        public virtual bool IsValidCallNode(CallNode node, TexlBinding binding, OperationCapabilityMetadata metadata)
+        {
+            // Functions may have their specific CallNodeDelegationStrategies (i.e. AsType, User)
+            // so, if available, we need to ensure we use their specific delegation strategy.
+            var function = binding.GetInfo(node)?.Function;
+
+            // We need to have this check in case an override does not properly overwrite this method to prevent
+            // getting lost in recursion.
+            if (function != null && function.QualifiedName != Function.QualifiedName)
```
Shpakh-MSFT, line 363: > function.QualifiedName != Function.QualifiedName [](http://example.com/codeflow?start=35&length=49)

Do referential equality.

function != Function

src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/OpDelegationStrategy.cs
```diff
@@ -87,31 +87,28 @@ private bool IsSupportedNode(TexlNode node, OperationCapabilityMetadata metadata
...
+            // Filter(Accounts, 'Account Name' in ["Foo", Bar"]) - Direct table use
+            // Set(Names, ["Foo", Bar"]); Filter(Accounts, 'Account Name' in Names) - Using variable of type table
+            // ClearCollect(Names, Accounts); Filter(Accounts, 'Account Name' in Names.'Account Name') - using column from collection.
+            // This won't be delegated if the AllowAsyncDelegation- Filter(Accounts, 'Account Name' in Accounts.'Account Name') as Accounts.'Account Name' is async.
+            if (isRHSNode && binding.Document.Properties.EnabledFeatures.IsEnhancedDelegationEnabled
+                && opDelStrategy is BinaryOpDelegationStrategy { Op: BinaryOp.In }
+                && !binding.IsRowScope(node)
+                && binding.GetType(node).IsTable 
+                && binding.GetType(node).IsColumn
+                && (binding.Features.HasFlag(Features.AllowAsyncDelegation) || !binding.IsAsync(node))
```
Shpakh-MSFT, line 108: How about impure node check?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AsType.cs
```diff
@@ -112,6 +113,25 @@ protected override bool RequiresPagedDataForParamCore(TexlNode[] args, int param
...
             return paramIndex != 1;
         }
+
+        public override ICallNodeDelegatableNodeValidationStrategy GetCallNodeDelegationStrategy()
+        {
+            return new AsTypeCallNodeDelegationStrategy(this);
+        }
+    }
+
+    internal class AsTypeCallNodeDelegationStrategy : DelegationValidationStrategy
```
Shpakh-MSFT, line 123: >class [](http://example.com/codeflow?start=13&length=5)

sealed?

LucGenetier:
✅ No public API change.

jelopezf:
Closing in favor of https://github.com/microsoft/Power-Fx/pull/1213

PR #1171, opened 2023-03-09
Ensure custom enum symbol names in intellisense suggestions are escaped

lesaltzm:
As none of our builtin enum names are escaped, we have a gap around suggestions, where if a custom enum is created, it isn't properly escaped in the intellisense suggestions. This only applies to canvas, for PCF custom enums. Every other use case is on Option Sets.

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests/IntellisenseTests/SuggestTest.cs
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/IntellisenseTests/SuggestTest.cs
@@ -3,10 +3,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using Microsoft.PowerFx.Core;
 using Microsoft.PowerFx.Core.Functions;
 using Microsoft.PowerFx.Core.Texl;
+using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Types.Enums;
 using Microsoft.PowerFx.Core.Utils;
 using Microsoft.PowerFx.Types;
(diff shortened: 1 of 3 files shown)


Alex-MSFT (approved):

Ananya-Gupta24 (dismissed):

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (dismissed):

Shpakh-MSFT:
>                 var enumName = enumInfo.EntityName.Value;

var enumName = TexlLexer.EscapeName(enumInfo.EntityName.Value) instead?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseHelper.cs:758 in 5780456. [](commit_id = 5780456b4d2d7e12a96a6ccc1b5cc0e40fd7a9d6, deletion_comment = False)

Shpakh-MSFT:
>                 AddSuggestion(intellisenseData, unqualifiedEnum, SuggestionKind.Enum, SuggestionIconKind.Other, enumType, requiresSuggestionEscaping: false);

We are not going to have unqualified enums any more and we will need to clean this code up. So this doesn't matter but for consistency, should we be escaping unqualifiedenum as well?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseHelper.cs:798 in 5780456. [](commit_id = 5780456b4d2d7e12a96a6ccc1b5cc0e40fd7a9d6, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/IntellisenseTests/SuggestTest.cs
```diff
@@ -235,6 +237,26 @@ public void TestSuggestEmptyAll(string expression, params string[] expectedSugge
             Assert.Equal(expectedSuggestions, actualSuggestions);
         }
 
+        [Fact]
+        public void TestSuggestEscapedEnumName()
+        {
+            var enumStoreBuilder = new EnumStoreBuilder();
+            enumStoreBuilder.TestOnly_WithCustomEnum(new EnumSymbol(
+                new DName("Name That.Requires!escaping"),
```
Shpakh-MSFT, line 245: >! [](http://example.com/codeflow?start=45&length=1)

'.' please

"Name That.Requires.escaping"

lesaltzm: No, the ! is also a character that requires escaping. In this case it's not used as the ! operator, just as a character in an identifier.

Shpakh-MSFT: Oh sorry my bad. I didn't see that.

Shpakh-MSFT (dismissed):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

Ananya-Gupta24 (approved):

Shpakh-MSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1178, opened 2023-03-13
Enabling table versions of RoundDown, RoundUp, and Trunc

gregli-msft:
The single column table versions of RoundDown, RoundUp, and Trunc were implemented, but never added to the interpreter's Library.cs.  They work in Canvas and I see no reason not to have them here too.


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/RoundDownT_ConsistentOneColumnTableResultDisabled.txt
```diff
@@ -0,0 +1,37 @@
...
+
+>> RoundDown([-9.8765, -8.7654, -7.6543, -6.5432, -5.4321], 2)
+Table({Value:-9.87},{Value:-8.76},{Value:-7.65},{Value:-6.54},{Value:-5.43})
+
+>> RoundDown(Filter([1,2,3], Value > 10), 0) // Empty table
+Table()
+
+// Scalar and table
+>> RoundDown(1234.5678, [-2, -1, 0, 1, 2])
+Table({Result:1200},{Result:1230},{Result:1234},{Result:1234.5},{Result:1234.56})
```
Shpakh-MSFT, line 19: Given that we already have shipped ConsistentOneColumnTable support in PA now (It's always ON), do we still need to add tests for both enable and disable support?

Shpakh-MSFT:
Thanks Greg for adding this support.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -714,11 +714,12 @@ public static FormulaValue RoundUp(IRContext irContext, NumberValue[] args)
             var x = Round(numberArg, digitsArg, RoundType.Up);
             return new NumberValue(irContext, x);
         }
-
+
+        // RoundDown is also used for Trunc which has an optional digits argument
```
Shpakh-MSFT, line 718: >// RoundDown is also used for Trunc which has an optional digits argument [](http://example.com/codeflow?start=8&length=73)

Can you please move this comment above line 722?

Shpakh-MSFT:
> #SETUP: ConsistentOneColumnTableResult

In few more weeks, this will be default in all PA studio versions available to customers. At that point, we should be able to clean this up and have only one set of tests and no longer need this setup.

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/RoundDownT_ConsistentOneColumnTableResult.txt:1 in f592fce. [](commit_id = f592fce0163932117b860e001bb7bbb805ee0338, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/RoundDownT_ConsistentOneColumnTableResult.txt
```diff
@@ -0,0 +1,52 @@
...
+
+// Blank values
+>> RoundDown( If(1<0,[1]), 1 )
+Blank()
+
+// Errors
+>> RoundDown( If(1/0<2, [1]), 1 )
+Error({Kind:ErrorKind.Div0})
+
+>> RoundDown( [1.2, 2.3, 3.4], 1/0 )
```
Shpakh-MSFT, line 35: >RoundDown( [1.2, 2.3, 3.4], 1/0 ) [](http://example.com/codeflow?start=3&length=33)

RoundDown( [1.2, 2.3, 3.4], [1/0] )

Shpakh-MSFT (dismissed):
:shipit:

CarlosFigueiraMSFT:
> #SETUP: ConsistentOneColumnTableResult

We may still need the tests without it for backcompat (conversion) scenarios, but we can greatly reduce the number of tests that don't use it.

---
In reply to: [1468090859](https://github.com/microsoft/Power-Fx/pull/1178#issuecomment-1468090859) [](http://example.com/codeflow?ancestors=1468090859,1468090859)

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/RoundDownT_ConsistentOneColumnTableResult.txt:1 in f592fce. [](commit_id = f592fce0163932117b860e001bb7bbb805ee0338, deletion_comment = False)

LucGenetier:
✅ No public API change.

Shpakh-MSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

Issue #1162, opened 2023-03-08
Columns dropped when coercing columns in record-to-record coercion


gregli-msft:
`[{a:1}, {a:true, b:2}]` results in `Table({a:1},{a:1})`.  We are correctly coercing the *true* to 1 in the second record, but in the process dropping the `b` column completely.  

The IR is `Call(Table, Record({a,Decimal(1)}, AggregateCoercion(RecordToRecord, a <- UnaryOp(BooleanToDecimal, ScopeAccess(Scope 1, a))))` which has no mention of `b`, likely there is an issue with AggregteCoercion of RecordToRecord.

anderson-joyle:
PA Client also drops `b` column:
`With({t:[{a:1}, {a:true,b:2}]}, First(t).b)` => Invalid. `b` isn't recognized.

Does this change apply to both PFx and PA Client?

gregli-msft:
It appears to, and that is indeed strange.  It may have been a bug in PA Client that was replicated in C#.  @lesaltzm took a quick look, he may be able to help with some context.  We talked earlier today, and with Carlos, and believe it the wrong behavior.  I don't yet know what it means for PA Client back compat.

anderson-joyle:
Fixed by https://github.com/microsoft/Power-Fx/pull/1271

PR #1179, opened 2023-03-13
Record-to-record coercion is dropping columns.

anderson-joyle:
Issue #1162.
A few challenges to fix this issue:

- How to change the table type in runtime and make it valid when asserting the result type. https://github.com/microsoft/Power-Fx/blob/8850cb940a0546839a0d74c11ac6cb26e1d5f0b7/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs#L292
- This draft fix aims `TableSyntaxDoesntWrapRecords` only.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
@@ -287,7 +287,7 @@ namespace Microsoft.PowerFx
                             });
                     }
 
-                    if (RuntimeHelpers.CanAssertReturn(func))
+                    if (IfFunction.CanCheckIfReturn(func))
                     {
                         Contract.Assert(result.IRContext.ResultType == node.IRContext.ResultType || result is ErrorValue || result.IRContext.ResultType is BlankType);
                     }
--- a/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
@@ -287,7 +287,7 @@ namespace Microsoft.PowerFx
                             });
                     }
 
-                    if (IfFunction.CanCheckIfReturn(func))
+                    if (RuntimeHelpers.CanAssertReturn(func))
                     {
                         Contract.Assert(result.IRContext.ResultType == node.IRContext.ResultType || result is ErrorValue || result.IRContext.ResultType is BlankType);
                     }
--- a/src/tests/Microsoft.PowerFx.Interpreter.Tests/IRTests.cs
+++ b/src/tests/Microsoft.PowerFx.Interpreter.Tests/IRTests.cs
@@ -66,6 +66,7 @@ namespace Microsoft.PowerFx.Interpreter.Tests
 
         [Theory]
         [InlineData("[{a:0},{a:\"3\",b:2}]", "Table({a:0,b:If(false,0)},{a:3,b:2})")]
+        [InlineData("[{a:1}, {a:true, b:2}]", "Table({a:1,b:If(false,0)},{a:1,b:2})")]
         public void RecordToRecordAggregateCoercionDontDropFieldsTest(string expr, string expected)
         {
             var iSetup = InternalSetup.Parse($"TableSyntaxDoesntWrapRecords");
--- a/src/tests/Microsoft.PowerFx.Interpreter.Tests/IRTests.cs
+++ b/src/tests/Microsoft.PowerFx.Interpreter.Tests/IRTests.cs
@@ -3,12 +3,12 @@
 
 using System;
 using System.Collections.Generic;
-using System.Collections.Immutable;
 using System.Linq;
 using System.Threading;
 using Microsoft.PowerFx.Core;
 using Microsoft.PowerFx.Core.Entities;
 using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Core.Tests;
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
 using Microsoft.PowerFx.Types;
@@ -62,6 +62,18 @@ namespace Microsoft.PowerFx.Interpreter.Tests
             var result = engine.Eval(expr, options: new ParserOptions() { AllowsSideEffects = true });
 
             Assert.IsNotType<ErrorValue>(result);
+        }
+
+        [Theory]
+        [InlineData("[{a:0},{a:\"3\",b:2}]", "Table({a:0,b:If(false,0)},{a:3,b:2})")]
+        public void RecordToRecordAggregateCoercionDontDropFieldsTest(string expr, string expected)
+        {
+            var iSetup = InternalSetup.Parse($"TableSyntaxDoesntWrapRecords");
+            var engine = new RecalcEngine(new PowerFxConfig(Features.TableSyntaxDoesntWrapRecords));
+
+            var result = engine.Eval(expr) as TableValue;
+
+            Assert.Equal(expected, result.ToExpression());
         }
 
         private class BooleanOptionSet : OptionSet, IExternalOptionSet
--- a/src/tests/Microsoft.PowerFx.Interpreter.Tests/IRTests.cs
+++ b/src/tests/Microsoft.PowerFx.Interpreter.Tests/IRTests.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text;
 using System.Threading;
 using Microsoft.PowerFx.Core;
 using Microsoft.PowerFx.Core.Entities;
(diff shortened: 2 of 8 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -287,7 +287,7 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
                             });
                     }
 
-                    if (IfFunction.CanCheckIfReturn(func))
+                    if (RuntimeHelpers.CanAssertReturn(func))
```
MikeStall, line 290: note this will go away with #1113

src/tests/Microsoft.PowerFx.Interpreter.Tests/MarshalTests.cs
```diff
@@ -405,6 +405,7 @@ public void MixRecordValueImplementationsIf()
             Assert.Equal(999.0, result1.ToObject());
         }
 
+        // !!!### This test case is failing!
```
MikeStall, line 408: still?

anderson-joyle: No. Thanks for pointing that out.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
anderson-joyle: Then it might be a good idea until #1113 is merged. I don't see this change working without this skip.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/IRTests.cs
```diff
@@ -64,6 +64,20 @@ public void RecordToRecordAggregateCoercionTest(string expr)
             Assert.IsNotType<ErrorValue>(result);
         }
 
+        [Theory]
+        
+        [InlineData("[{a:0},{a:\"3\",b:2}]", "Table({a:0,b:If(false,0)},{a:3,b:2})")]
+        [InlineData("[{a:1}, {a:true, b:2}]", "Table({a:1,b:If(false,0)},{a:1,b:2})")]
```
jas-valgotar, line 70: >[InlineData("[{a:1}, {a:true, b:2}]", "Table({a:1,b:If(false,0)},{a:1,b:2})")] [](http://example.com/codeflow?start=8&length=78)

Can we add negative tests? Ie what happens when field can't be coerced to certain type.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
anderson-joyle: Thats a good suggestion but I don't think it fits for this situation, unless we "unwrap" the table value after EvalVisitor.
Expression: `[{a:1}, {a:true, b:2}]`
The binder produces a `*[a:n]` table type, which should be `*[a:n, b:n]`. I can wrap the result back to `*[a:n]` but then we we cant call `recordvalue.b`.
If I am missing something, please let me know.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/IRTests.cs
```diff
@@ -64,6 +64,20 @@ public void RecordToRecordAggregateCoercionTest(string expr)
...
+        public void RecordToRecordAggregateCoercionDontDropFieldsTest(string expr, string expected)
+        {
+            var iSetup = InternalSetup.Parse($"TableSyntaxDoesntWrapRecords");
+            var engine = new RecalcEngine(new PowerFxConfig(Features.TableSyntaxDoesntWrapRecords));
+
+            var result = engine.Eval(expr) as TableValue;
```
MikeStall, line 76: >as TableValue; [](http://example.com/codeflow?start=43&length=14)

nit:
(TableValue) engine.Eval(expr)

will give us an exception immediately at the cast. 
Whereas 'as TableValue' will return null and we'll get a nullref later - which is harder to diagnose.

src/tests/Microsoft.PowerFx.Interpreter.Tests/IRTests.cs
```diff
@@ -64,6 +64,20 @@ public void RecordToRecordAggregateCoercionTest(string expr)
             Assert.IsNotType<ErrorValue>(result);
         }
 
+        [Theory]
+        
+        [InlineData("[{a:0},{a:\"3\",b:2}]", "Table({a:0,b:If(false,0)},{a:3,b:2})")]
+        [InlineData("[{a:1}, {a:true, b:2}]", "Table({a:1,b:If(false,0)},{a:1,b:2})")]
+        public void RecordToRecordAggregateCoercionDontDropFieldsTest(string expr, string expected)
+        {
+            var iSetup = InternalSetup.Parse($"TableSyntaxDoesntWrapRecords");
```
MikeStall, line 73: > var iSetup = InternalSetup.Parse($"TableSyntaxDoesntWrapRecords"); [](http://example.com/codeflow?start=11&length=67)

nit - not used?

src/tests/Microsoft.PowerFx.Interpreter.Tests/IRTests.cs
MikeStall, line 70: >If(false,0)} [](http://example.com/codeflow?start=60&length=12)

nit - we can use ToExpression(settings = minimal) to simplify this.

src/tests/Microsoft.PowerFx.Interpreter.Tests/IRTests.cs
```diff
@@ -64,6 +64,20 @@ public void RecordToRecordAggregateCoercionTest(string expr)
             Assert.IsNotType<ErrorValue>(result);
         }
 
+        [Theory]
+        
+        [InlineData("[{a:0},{a:\"3\",b:2}]", "Table({a:0,b:If(false,0)},{a:3,b:2})")]
```
MikeStall, line 69: >[{a:0},{a:\"3\",b:2}] [](http://example.com/codeflow?start=21&length=21)

Is [] special?
Could we add a test where Expr uses Table()? 
like: 
Table({a:0},{a:\"3\",b:2})

src/tests/Microsoft.PowerFx.Interpreter.Tests/IRTests.cs
```diff
@@ -64,6 +64,20 @@ public void RecordToRecordAggregateCoercionTest(string expr)
...
+        [InlineData("[{a:1}, {a:true, b:2}]", "Table({a:1,b:If(false,0)},{a:1,b:2})")]
+        public void RecordToRecordAggregateCoercionDontDropFieldsTest(string expr, string expected)
```
MikeStall, line 71: another test - is this recursive? (not sure the answer, but we should test either way)

[{b: {x:1}},
 {b: {x:"2", y:2}}]

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/IRTests.cs
anderson-joyle: Removed the cast since we are working with the serialized version of it

src/tests/Microsoft.PowerFx.Interpreter.Tests/IRTests.cs
anderson-joyle: Good catch.

src/tests/Microsoft.PowerFx.Interpreter.Tests/IRTests.cs
anderson-joyle: [] is an special case. Why? I need to understand this better with the PMs.
`Table({a:0},{a:\"3\",b:2})` gives a compilation error due to the conflict of types.

src/tests/Microsoft.PowerFx.Interpreter.Tests/IRTests.cs
anderson-joyle: This fails. Which brings the question: in case this is allowed, what would be the deepest recursion level?

src/tests/Microsoft.PowerFx.Interpreter.Tests/IRTests.cs
anderson-joyle: Test case added. Results in a compilation error.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
jas-valgotar: I am wondering why we don't compute the correct type in the binder. I think it has all the information to compute it. Shouldn't we? @MikeStall

PR #1184, opened 2023-03-13
Fix error message to use correct type for option set config errors

lesaltzm:


MikeStall (approved):

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1191, opened 2023-03-14
Restrict IsEmpty arguments to tables only

CarlosFigueiraMSFT:
IsEmpty is supposed to check if a table is empty or not, but currently it accepts different types of arguments. Since it doesn't make sense to use it for other types, we will restrict it. But to avoid breaking Power Apps, we will make the change behind a new feature.

Also adding the implementation to the C# interpreter.


LucGenetier:
✅ No public API change.

lesaltzm (approved):

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT:
Closing this one, will create another one with a branch in the microsoft account.

PR #1192, opened 2023-03-15
Avoid double parsing UDF body

VamsiModem:
Avoid double parsing UDF body. 
Added TexlNode for the body in the UDFDefinition that can be used later on.


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -277,6 +279,13 @@ public IEnumerable<ExpressionError> Errors
             set => _errors.AddRange(value);
         }
 
+        public CheckResult SetTexlNode(TexlNode texlNode)
```
lesaltzm, line 282: This seems odd to allow especially publicly... What if the caller passed it a TexlNode from a different expression?

VamsiModem: Yeah, I realized that can happen here

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall:
This is one of the issues called out in #942

src/libraries/Microsoft.PowerFx.Interpreter/UDF/UDFDefinition.cs
```diff
@@ -11,15 +12,15 @@ internal class UDFDefinition
     {
         internal readonly string Name;
 
-        internal readonly string Body;
+        internal readonly TexlNode Body;
```
MikeStall, line 15: >TexlNode [](http://example.com/codeflow?start=26&length=8)

This is the key change. String --> TexlNode

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -176,10 +176,15 @@ public DefineFunctionsResult DefineFunctions(string script)
...
                 udf.Ident.ToString(),
-                udf.Body.ToString(),
-                FormulaType.GetFromStringOrNull(udf.ReturnType.ToString()),
+                udf.Body,
+                FormulaType.GetFromStringOrNull(udf.ReturnType.ToString()) ?? FormulaType.Unknown,
                 udf.IsImperative,
-                udf.Args.Select(arg => new NamedFormulaType(arg.VarIdent.ToString(), FormulaType.GetFromStringOrNull(arg.VarType.ToString()))).ToArray())).ToArray();
+                udf.Args.Select(arg =>
+                {
+                    var formulaType = FormulaType.GetFromStringOrNull(arg.VarType.ToString()) ?? FormulaType.Unknown;
```
MikeStall, line 184: >FormulaType.GetFromStringOrNull(arg.VarType.ToString()) ?? FormulaType.Unknown [](http://example.com/codeflow?start=38&length=78)

nit: Can we put this in a wrapper since we repeat this.
Maybe an extension method on IdentToken?

Then we can  just write:
arg.GetFormulaType();

src/libraries/Microsoft.PowerFx.Core/Public/Types/NamedFormulaType.cs
```diff
@@ -16,6 +16,8 @@ public sealed class NamedFormulaType
 
         public NamedFormulaType(string name, FormulaType type, string displayName = null)
         {
+            Contracts.AssertValue(type);
```
MikeStall, line 19: >Contracts [](http://example.com/codeflow?start=12&length=9)

nit - this is public method, so we should actually throw ArgumentNullException instead of assert.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -291,7 +307,12 @@ private BindingConfig GetDefaultBindingConfig()
         // Called by CheckResult.ApplyBinding to compute the binding. 
         internal (TexlBinding, ReadOnlySymbolTable) ComputeBinding(CheckResult result)
         {
-            var parse = result.ApplyParse();
+            ParseResult parseResult = null;
+
+            if (result.RootNode == null)
+            {
+                parseResult = result.ApplyParse();
+            }
```
MikeStall, line 315: can we avoid this extra check. 
If result.RootNode is null, then ApplyParse() should nop. 

Essentially, setting a ParseResult on the CheckREsult "short circuits"  Parsing. 
But everything downstream doesn't care and we avoid extra checks like 312 and 336.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -237,6 +251,8 @@ public HashSet<string> TopLevelIdentifiers
             }
         }
 
+        public ParserOptions ParserOptions => _parserOptions;
```
MikeStall, line 254: >ParserOptions [](http://example.com/codeflow?start=29&length=13)

Why do we need to expose this? 
Is it encapsulated by the results of ParseResult?

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -584,4 +603,15 @@ internal interface IOperationStatus
 
         public bool IsSuccess { get; }
     }
+
+    /// <summary>
+    /// Used specifically for UDF, so that we can set the texlnode for UDF body that was already parsed.
+    /// </summary>
+    internal class UDFCheckResult : CheckResult
```
MikeStall, line 610: >UDFCheckResult [](http://example.com/codeflow?start=19&length=14)

Why a new derived class? It's not adding anything.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
VamsiModem: When we use TexlNode we don't have ParseResult

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
VamsiModem: Setting TexlNode is very specific to UDFs and we don't want to allow setting that from anywhere else other then from UDFCheckResult.

MikeStall:
>         public CheckResult SetText(ParseResult parse)

Can we use this entry?
Create a ParseREsult around the (TexlNode +Errors)




---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs:113 in f4368a5. [](commit_id = f4368a5edd9d80cbe45cc8f789d38daf1ea03804, deletion_comment = False)

MikeStall:
>         public void DefFunc()

This should be a huge step towards fixing source spans on error reporting (since round tripping loses all that information). 

Can we add a test that verifies the source span? 

Take in 2 named formulas, error in the 2nd one, and verify the source span of the error is reasonable. 
Because when we round trip to string, we'd lose that info. 


---
Refers to: src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs:315 in f4368a5. [](commit_id = f4368a5edd9d80cbe45cc8f789d38daf1ea03804, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
lesaltzm: Why can't that be in the constructor of CheckResult?

MikeStall: wrap in a ParseResult and call the existing SetText() method. 

Else every caller of CheckResult has to now special case TexlNode vs. ParseResult.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #1193, opened 2023-03-15
Connectors, add support for SQL server with dynamic intellisense

LucGenetier:
Add support for x-ms-dynamic-values swagger extension


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/Microsoft.PowerFx.Connectors.csproj
```diff
@@ -21,7 +21,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.OpenApi.Readers" Version="1.3.1" />
+    <PackageReference Include="Microsoft.OpenApi.Readers" Version="1.6.3" />
```
MikeStall, line 24: >"1.6.3 [](http://example.com/codeflow?start=66&length=6)

note - this is important because there is a bug in 1.3.1 on response body type.

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -199,13 +199,17 @@ public static (FormulaType, RecordType) ToFormulaType(this OpenApiSchema schema,
...
                         case "uuid":
+                        case "mquery": // For SQL Server Power Query language
                             return (FormulaType.String, null);
 
                         case "date-time":
                             // Consider this as a string for now
                             // $$$ Should be DateTime type
                             return (FormulaType.String, null);
 
+                        case "binary":
```
MikeStall, line 210: >binary [](http://example.com/codeflow?start=30&length=6)

What does this mean?
Binary/String seems strange combo

src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
```diff
@@ -72,6 +73,29 @@ private void WriteObject(string objectName, OpenApiSchema schema, IEnumerable<Na
...
+            {
+                foreach (NamedValue nv in fields)
+                {
+                    WriteProperty(
+                        nv.Name,
+                        new OpenApiSchema()
+                        {
+                            Type = nv.Value.Type._type.Kind switch
+                            {
+                                DKind.Number => "number",
```
MikeStall, line 86: >"number" [](http://example.com/codeflow?start=48&length=8)

nit - (future PR is fine) - we repeat these string literals alot. 
Can we encapsulate somewhere.
public const string NumberType = "number";

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
```diff
@@ -285,6 +287,78 @@ public override bool CheckForDynamicReturnType(TexlBinding binding, TexlNode[] a
...
+                    arguments.Add(arg);
+                }
+
+                cts.ThrowIfCancellationRequested();
+                FormulaValue result = await cdv.ServiceFunction.InvokeAsync(arguments.ToArray(), cts);
+
+                List<string> suggestions = new List<string>();
+                if (result is RecordValue rv)
+                {
+                    if (!string.IsNullOrEmpty(cdv.ValueCollection) && !string.IsNullOrEmpty(cdv.ValuePath))
```
MikeStall, line 333: nit - can we check this earlier? Before the network call?

Or even at cdv creation time?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -208,6 +210,13 @@ internal abstract class TexlFunction : IFunction
         /// </summary>
         public virtual bool ModifiesValues => false;
 
+        // This method is used for managing "x-ms-dynamic-values" OpenApi extensions in connectors
```
MikeStall, line 213: FYI - this is the key hook from intellisense to dynamic intellisense.

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseHelper.cs
```diff
@@ -324,35 +326,45 @@ public static void AddTopLevelSuggestionsForCursorType(IntellisenseData.Intellis
...
+
+                    if (!info.RequiresScopeIdentifier)
+                    {
+                        AddTopLevelSuggestions(intellisenseData, type);
+                    }
                 }
 
-                if (!info.RequiresScopeIdentifier)
+                // If connector function has some suggestions, let's add them here
+                List<string> suggestions = info.Function.GetConnectorSuggestionsAsync(callNode, argPosition, CancellationToken.None).ConfigureAwait(false).GetAwaiter().GetResult();
```
MikeStall, line 360: >CancellationToken.None [](http://example.com/codeflow?start=109&length=22)

note - if intellisense ever gets a cancellation token, we would pass it in here.

src/tests/Microsoft.PowerFx.Connectors.Tests/OpenApiParserTests.cs
```diff
@@ -425,6 +425,15 @@ public void LQA_Load()
             List<ServiceFunction> functionList = OpenApiParser.Parse("LQA", doc);
             Assert.Contains(functionList, sf => sf.GetUniqueTexlRuntimeName() == "lQA__GetAnswersFromText");
         }
+
+        [Fact]
```
MikeStall, line 429: nit - empty line on 430?

src/tests/xunit.runner.json
```diff
@@ -0,0 +1,4 @@
+{
+  "$schema": "https://xunit.net/schema/current/xunit.runner.schema.json",
+  "methodDisplay": "method"
```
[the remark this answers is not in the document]
LucGenetier: [https://xunit.net/docs/configuration-files](https://xunit.net/docs/configuration-files) 

Set this to override the default display name for test cases. If you set this to method, the display name will be just the method (without the class name); if this set this value to classAndMethod, the default display name will be the fully qualified class name and method name.

src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
LucGenetier: Can do that in another PR but we would need a prefix
Ex: OpenApiNumberType (maybe there is an existing one..)

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
LucGenetier: cdv creation time -> those are optional params so that's not the right place
before the call -> we _could_ but I imagine we could add support for more combinations later - the the cases that not supported today & that would duplicate efforts

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseHelper.cs
LucGenetier: yes intellisense is synchronous for now
there is no SuggestAsync API...
probably needs to be added

src/tests/Microsoft.PowerFx.Connectors.Tests/OpenApiParserTests.cs
LucGenetier: really nit.... I can updated that

src/tests/xunit.runner.json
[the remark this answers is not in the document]
LucGenetier: This is much cleaner that way and you don't need to resize text explorer to discover what test you're runinng

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
MikeStall: Just put them in a static class - that's your prefx.

PR #1196, opened 2023-03-15
Fixes 1187 Adds ability to add host object.

### LINKED ISSUE #1187: Defining Host Objects
We can define builtin functions, but we want to allow defining host _objects_, like User (see #1114), Environment, etc . 

These objects can be instantiated from the ServiceProvider instead of SymbolValues.  

Add a method like this To SymbolTable:
        public void AddHostObject(string name, FormulaType type, Func<IServiceProvider,FormulaValue> getValue)

So 'getValue' means they won't need a slot.  getValue could be null if this is just intellisense only, since it's only used at eval time. Likely make an update here in interpreter to consume it:

https://github.com/microsoft/Power-Fx/blob/f11ba50c39c018302e088ed6a34f33a131f996fb/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs#L661

Since IServiceProvider is a builtin interface, we can define this in Core without a layering violation.

jas-valgotar:
Fixes #1187

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
@@ -277,5 +277,35 @@ namespace Microsoft.PowerFx
 
             _variables.Add(entity.EntityName, nameInfo);
         }
+
+        /// <summary>
+        /// Adds a host object schema, that can be referenced in the formula.
+        /// Actual object is added in Runtime config service provider.
+        /// </summary>
+        /// <param name="name">Name of the object.</param>
+        /// <param name="type">Type of the object.</param>
+        /// <param name="getValue">Call back that will retrieve object from the service provider.
+        /// It can throw CustomFunctionErrorException, that fx will convert to an error.</param>
+        public void AddHostObject(string name, FormulaType type, Func<IServiceProvider, FormulaValue> getValue)
+        {
+            var hostDName = ValidateName(name);
+
+            // Attempt to update display name provider before symbol table,
+            // since it can throw on collision and we want to leave the config in a good state.
+            if (_environmentSymbolDisplayNameProvider is SingleSourceDisplayNameProvider ssDnp)
+            {
+                _environmentSymbolDisplayNameProvider = ssDnp.AddField(hostDName, default);
+            }
+
+            var info = new NameLookupInfo(
+                BindKind.PowerFxResolvedObject,
+                type._type,
+                DPath.Root,
+                0,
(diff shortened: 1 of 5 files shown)


src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -276,6 +276,28 @@ internal void AddEntity(IExternalEntity entity, DName displayName = default)
             }
 
             _variables.Add(entity.EntityName, nameInfo);
+        }
+
+        public void AddHostObject(string name, FormulaType type, Func<IServiceProvider, FormulaValue> getValue)
```
MikeStall, line 281: We should document what getValue does on failure. Return null? Return an ErrorValue? Throw a CustomFunctionException (that we wrap and return an error)

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
jas-valgotar: updated.

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -276,6 +276,36 @@ internal void AddEntity(IExternalEntity entity, DName displayName = default)
...
+        }
+
+        /// <summary>
+        /// Adds a host object schema, that can be referenced in the formula.
+        /// Actual object is added in Runtime config service provider.
+        /// </summary>
+        /// <param name="name">Name of the object.</param>
+        /// <param name="type">Type of the object.</param>
+        /// <param name="getValue">Call back that will retrieve object from the service provider.
+        /// It can throw CustomFunctionErrorException, that fx will convert to an error.</param>
```
MikeStall, line 288: >CustomFunctionErrorException [](http://example.com/codeflow?start=25&length=28)

nit - you can use <see= > to link to class name within comments. 
This also means if the class gets rename, the comment is updated

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
jas-valgotar: As discussed offline, we can't do it since the Exception is in the Interpreter.

LucGenetier:
✅ No public API change.

PR #1199, opened 2023-03-16
Fix intellisense suggestion ordering

CarlosFigueiraMSFT:
Revert behavior change from a previous PR. A proper fix (without the exclusion) will come later.


LucGenetier:
✅ No public API change.

Issue #1203, opened 2023-03-16
Make Error( {Kind:”2” } ) work 


MikeStall:
This works:
`Error( {Kind: 2 } )`

But this does not:
`Error( {Kind:"2" } )`

and gets compile time error because it won't allow string to number coercion:
Error 7-19: Invalid type for field 'Kind'. Expected field type of 'ErrorKind'.

We want to allow the string-to-number coercion and for 2nd one to work. 

Fix is likely in:
https://github.com/microsoft/Power-Fx/blob/2cace9a65b3a98e792ba0072cd2a2a4c9d27d191/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Error.cs#L44

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/1206

PR #1206, opened 2023-03-16
Allow the string-to-number coercion in Error function

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1203

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Error.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Error.cs
@@ -99,7 +99,7 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
                 argumentKindType = argumentKindType.GetEnumSupertype();
             }
 
-            if (argumentKindType.Kind != requiredKindField.Type.Kind)
+            if (!(requiredKindField.Type.Kind == DKind.Number) && (argumentKindType.Kind != requiredKindField.Type.Kind))
             {
                 errors.EnsureError(
                     argument,
@@ -143,7 +143,7 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
                     acceptedFields.Where(field =>
 
                         // Kind has already been handled before
-                        field.Name != "Kind" && names.Any(name => name.Name == field.Name)));
+                        ((requiredKindField.Type.Kind == DKind.Number) ? true : field.Name != "Kind") && names.Any(name => name.Name == field.Name)));
 
                 typeValid = CheckType(argument, argumentType, expectedOptionalFieldsRecord, errors, true, out matchedWithCoercion);
             }
--- a/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Error.txt
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Error.txt
@@ -97,10 +97,10 @@ true
 >> IsError(Error({Kind: "2" }))
 true
 
->> IfError( If( false, 1, Error({Kind:2}) ), FirstError.Kind )
+>> IfError( If( false, 1, Error({Kind: 2}) ), FirstError.Kind )
 2
 
->> IfError( If( false, 1, Error({Kind:"2"}) ), FirstError.Kind )
+>> IfError( If( false, 1, Error({Kind: "2"}) ), FirstError.Kind )
 2
 
 >> IfError( If( false, 1, Error({Kind:true}) ), FirstError.Kind )
--- a/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Error.txt
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Error.txt
@@ -98,10 +98,10 @@ true
 true
 
 >> IfError( If( false, 1, Error({Kind:2}) ), FirstError.Kind )
-2 
+2
 
 >> IfError( If( false, 1, Error({Kind:"2"}) ), FirstError.Kind )
-2 
+2
 
 >> IfError( If( false, 1, Error({Kind:true}) ), FirstError.Kind )
 1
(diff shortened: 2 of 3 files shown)


LucGenetier:
✅ No public API change.

MikeStall:
>             var argumentKindType = names.First(tn => tn.Name == requiredKindField.Name).Type;

From maintenance perspective, I wondered here "what if requiredKindField isn't in the list" ... and then I see we handled that on line 81. 

Can we use FirstOrDefault:
var x = names.FirstOrDefault(tn => tn.Name == requiredKindField.Name)

and then apply in 81 and 96. 
(also has bonus that we'll stop scanning the list twice. )




---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Error.cs:96 in a8b015b. [](commit_id = a8b015b2fc8801db3a07a433ef8e72f8ae404154, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Error.txt
```diff
@@ -86,4 +86,28 @@ Error({Kind:ErrorKind.Custom})
 "lorem ipsum"
 
 >> Date(2000, Sqrt(-1), Error("lorem ipsum"))
-Error(Table({Kind:ErrorKind.Numeric},{Kind:ErrorKind.Custom}))
\ No newline at end of file
+Error(Table({Kind:ErrorKind.Numeric},{Kind:ErrorKind.Custom}))
+
+>> Error( {Kind:"test" } )
```
MikeStall, line 91: > Error( {Kind:"test" } ) [](http://example.com/codeflow?start=2&length=24)

curious - what is the IR in this case?

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Error.txt
nguhoa: IR in this case has resulttype is N.

nguhoa:
>             var argumentKindType = names.First(tn => tn.Name == requiredKindField.Name).Type;

Updated here.

---
In reply to: [1474276566](https://github.com/microsoft/Power-Fx/pull/1206#issuecomment-1474276566) [](http://example.com/codeflow?ancestors=1474276566,1474276566)

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Error.cs:96 in a8b015b. [](commit_id = a8b015b2fc8801db3a07a433ef8e72f8ae404154, deletion_comment = False)

nguhoa:
>             var optionalFields = acceptedFields.Where(tn => tn.Name != "Kind");

just noticed that this variable never been used so I removed it.

---
In reply to: [1474280259](https://github.com/microsoft/Power-Fx/pull/1206#issuecomment-1474280259) [](http://example.com/codeflow?ancestors=1474280259,1474280259)

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Error.cs:57 in a8b015b. [](commit_id = a8b015b2fc8801db3a07a433ef8e72f8ae404154, deletion_comment = False)

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Error.cs
```diff
@@ -93,13 +92,13 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
                 return false;
             }
 
-            var argumentKindType = names.First(tn => tn.Name == requiredKindField.Name).Type;
+            var argumentKindType = names.FirstOrDefault(tn => tn.Name == requiredKindField.Name).Type;
```
CarlosFigueiraMSFT, line 95: >FirstOrDefault [](http://example.com/codeflow?start=41&length=14)

Calling `FirstOrDefault` implies that `null` can be returned, but the code is dereferencing it right after it (`.Type`). We know that this will not return null, since this case is handled by the previous `if` statement, so leaving it as `First` is fine.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Error.cs
```diff
@@ -93,13 +92,13 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
...
             if (argumentKindType.IsEnum)
             {
                 argumentKindType = argumentKindType.GetEnumSupertype();
             }
 
-            if (argumentKindType.Kind != requiredKindField.Type.Kind)
+            if (!(requiredKindField.Type.Kind == DKind.Number) && (argumentKindType.Kind != requiredKindField.Type.Kind))
```
CarlosFigueiraMSFT, line 101: >!(requiredKindField.Type.Kind == DKind.Number) [](http://example.com/codeflow?start=16&length=46)

This will always be false, so this `if` will never be executed:

- requiredKindField is defined as the 'Kind' field of acceptedFields (line 55)
- acceptedFields is defined as the fields from reifiedError (line 54)
- reifiedError is defined from ErrorType, and the schema defines Kind as a numeric field (ErrorType.cs, line 27)

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Error.cs
nguhoa: updated code. thanks

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Error.cs
nguhoa: you are right, removed this. thanks

CarlosFigueiraMSFT (approved):
:shipit:

PR #1198, opened 2023-03-16
Introduce 5 icons to be used for status of canvas apps offline sync.

manuelmillot:

PR #1210, opened 2023-03-17
Revert "Introduce 5 icons to be used for status of canvas apps offline sync."

MikeStall:
Reverts microsoft/Power-Fx#1198

Causing some failures in PAClient:
Server debug failure: All enum options should exist in the categorization map. Callstack: at System.Environment.get_StackTrace()

Turns out original change is breaking internal afterall, so we need a PAClient PR for it.


jas-valgotar (approved):

LucGenetier:
✅ No public API change.

Issue #1114, opened 2023-02-24
Expose User host object


MikeStall:
Power Apps has existing user function:  https://learn.microsoft.com/en-us/power-platform/power-fx/reference/function-user

The schema here includes: Email, FullName, and Image. 

This should be in Power Fx Core so that other hosts can access in a consistent way.  

There's an existing UserFunction definition in PowerApps (USer.cs) that should just be moved to core.  

**For interpreter**, 
- hosts can provide the values via a ServiceProvider. 
- We have tests that demonstrate this: https://github.com/microsoft/Power-Fx/blob/76508caefa6f2c4deece9d3eded53ced7071583d/src/tests/Microsoft.PowerFx.Interpreter.Tests/ConfigTests.cs#L662 . But note that the test example derives from CustomFunction, whereas we should just use the definition from PAclient in User.cs.
- if the User value is not set, then function can fail. 
- we may need to allow it to be a **lazy lookup** to get email/name. Sometimes the host just has a user guid, and getting name,email requires a network call. 

TBD on how to handle Image.

MikeStall:
fyi,  PowerApps impl of Email has some issues: https://github.com/MicrosoftDocs/powerapps-docs/issues/3152

MikeStall:
Adding Needs Discussion tag, cc @jorisdg

MikeStall:
Upon discussion with @gregli-msft, @jorisdg , @Shpakh-MSFT , we decided on:

1. User object, not User() function. This is more consistent with other host-level objects.  This is a known divergence from PowerApps, but we will update Power Apps.
2. But still same fields (full name, email, image)
3. Still add some hosting API to allow hosts to specify the fields

In case of conflict between a local symbol and the builtin global, not that Power Fx allows [@identifier] to refer to global namespace.  We can double check that Intellisense shows both [@User] and User if there is a conflict.

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/1214

PR #1214, opened 2023-03-17
Add user object

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1114


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Environment/SymbolExtensions.cs
```diff
@@ -31,6 +32,17 @@ public static void SetCulture(this RuntimeConfig symbols, CultureInfo culture)
             symbols.AddService(culture ?? throw new ArgumentNullException(nameof(culture)));
         }
 
+        /// <summary>
+        /// Set UserInfo.
+        /// </summary>
+        /// <param name="symbols">SymbolValues where to set the UserInfo.</param>
+        /// <param name="userInfo">UserInfo to set.</param>
+        /// <exception cref="ArgumentNullException">When userInfo is null.</exception>
+        public static void SetUserInfo(this RuntimeConfig symbols, UserInfo userInfo)
```
MikeStall, line 41: This should take the interface, IUserInfo.

src/tests/Microsoft.PowerFx.Interpreter.Tests/UserInfoTests.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            var result = await check.GetEvaluator().EvalAsync(CancellationToken.None, rc);
+            Assert.IsType<ErrorValue>(result);
+            Assert.NotNull(((ErrorValue)result).Errors.Where((error) => error.Message.Contains("UserInfo object was not added to service")));
+        }
+
+        public FormulaValue GetUserInfoObject(IServiceProvider serviceProvider)
+        {
+            var cache = new TypeMarshallerCache();
+            var userInfo = (UserInfo)serviceProvider.GetService(typeof(UserInfo)) ?? throw new CustomFunctionErrorException("UserInfo object was not added to service");
```
MikeStall, line 86: >UserInfo [](http://example.com/codeflow?start=71&length=8)

get as IUserInfo to support other derived implementations.

src/tests/Microsoft.PowerFx.Interpreter.Tests/UserInfoTests.cs
```diff
@@ -0,0 +1,90 @@
...
+                FullName = fullName,
+                Email = email
+            };
+
+            var userInfoType = RecordType.Empty()
+                .Add("FullName", FormulaType.String)
+                .Add("Email", FormulaType.String);
+
+            SymbolTable symbol = new SymbolTable();
+            symbol.AddHostObject("UserInfo", userInfoType, GetUserInfoObject);
```
MikeStall, line 37: >symbol.AddHostObject("UserInfo", userInfoType, GetUserInfoObject); [](http://example.com/codeflow?start=12&length=66)

We should wrap this as a convenience method in Core. 

Extension like:
symbol.AddUserObject();

or 
config.SymbolTable.AddUserObject();

which adds to the symbol table, and then at runtime, they call the RuntimeConfig extension on line 41

So GetUserInfoObject gets pulled into Core as a default implementation, but note that 
- uses IUserInfo
- does not need to use TypeMarshallerCache (since that's only in interpreter). We can just manually create the FormulaVAlue

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/UserInfoTests.cs
nguhoa: update

src/tests/Microsoft.PowerFx.Interpreter.Tests/UserInfoTests.cs
nguhoa: update this. thanks

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -306,6 +307,30 @@ public void AddHostObject(string name, FormulaType type, Func<IServiceProvider,
...
+            var userInfoType = RecordType.Empty()
+                .Add("FullName", FormulaType.String)
+                .Add("Email", FormulaType.String);
+
+            AddHostObject("UserInfo", userInfoType, GetUserInfoObject);
+        }
+
+        private FormulaValue GetUserInfoObject(IServiceProvider serviceProvider)
+        {
+            var userInfo = (IUserInfo)serviceProvider.GetService(typeof(IUserInfo)) ?? throw new Exception("UserInfo object was not added to service");
```
MikeStall, line 327: >Exception [](http://example.com/codeflow?start=97&length=9)

C# tip - we should never throw "Exception" directly - instead throw an appropriate derived type.  InvalidOperationException is a good catch all. 


Although in this case, we could just return null and update the caller to convert that into an error. 

see [https://github.com/microsoft/Power-Fx/pull/1196](https://github.com/microsoft/Power-Fx/pull/1196)

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -306,6 +307,30 @@ public void AddHostObject(string name, FormulaType type, Func<IServiceProvider,
...
+
+            RecordValue userRecord = FormulaValue.NewRecordFromFields(
+                new NamedValue("FullName", FormulaValue.New(userInfo.FullName ?? string.Empty)),
+                new NamedValue("Email", FormulaValue.New(userInfo.Email ?? string.Empty)));
+
+            return userRecord;
         }
```
MikeStall, line 334: Can we move these to an extension method, and not directly on SymbolTable.

src/tests/Microsoft.PowerFx.Interpreter.Tests/UserInfoTests.cs
```diff
@@ -0,0 +1,80 @@
...
+            symbol.AddUserInfoObject();
+
+            var engine = new RecalcEngine(new PowerFxConfig() { SymbolTable = symbol });
+            var rc = new RuntimeConfig();
+
+            var check = engine.Check("UserInfo.FullName");
+            Assert.True(check.IsSuccess);
+
+            try
+            {
```
MikeStall, line 71: nit: for expected exceptions, see Assert.ThrowsAsync<>

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
MikeStall: Actually, thinking more ... exception (instead of ErrorValue) is the right failure here since this scenario is a host bug. The host told us to enable User object, but then didn't provide the value for it. 

So just make it a more specific exception, like InvalidOperationException

nguhoa: updated to InvalidOperationException

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
nguhoa: yes, moved this to extension method,

src/libraries/Microsoft.PowerFx.Interpreter/Environment/SymbolExtensions.cs
nguhoa: update

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTableExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// Adds a UserInfo object schema.
+        /// Actual object is added in Runtime config service provider.
+        /// </summary>
+        public static void AddUserInfoObject(this SymbolTable symbolTable)
+        {
+            var userInfoType = RecordType.Empty()
+                .Add("FullName", FormulaType.String)
+                .Add("Email", FormulaType.String);
+
+            symbolTable.AddHostObject("UserInfo", userInfoType, GetUserInfoObject);
```
MikeStall, line 21: This should be "User", not "UserInfo"

PR #1215, opened 2023-03-20
Moving EncodeUrl function to Builtin folder

anderson-joyle:
Moving EncodeUrl function file to Texl/Builtins folder, where other builtin functions live.


LucGenetier:
✅ No public API change.

jas-valgotar (dismissed):

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/EncodeUrl.cs
```diff
@@ -1,10 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-using System;
 using System.Collections.Generic;
-using System.Text;
```
lesaltzm, line 6: This function should be added to BuiltinFunctionsCore and removed from PAClient's builtinfunctions, seems like it was missed in the initial migration

anderson-joyle: I have [this PR out](https://msazure.visualstudio.com/OneAgile/_build/results?buildId=70286885&view=results), but due to some broken enum changes, I get to change PFx version. working on it.

CarlosFigueiraMSFT (approved):

LucGenetier:
✅ No public API change.

PR #1218, opened 2023-03-20
Tabular Dec2Hex doesn't work with coercion

LucGenetier:
Dec2Hex(["10","255","-16"]) works correctly in Canvas, returning ["A", "FF", "FFFFFFFFF0" ] as it should


LucGenetier:
✅ No public API change.

lesaltzm (approved):

PR #1221, opened 2023-03-21
Add Decimal

gregli-msft:


src/libraries/Microsoft.PowerFx.Core/Public/Values/DecimalValue.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        public decimal Normalize(decimal value)
+        {
+            // remove trailing 0's (significant digits)
+            return Value / 1.000000000000000000000000000000m;
+        }
+
+        public override void ToExpression(StringBuilder sb, FormulaValueSerializerSettings settings)
+        {
+            decimal normalized = Normalize(Value);
```
MikeStall, line 38: > decimal normalized = Normalize(Value); [](http://example.com/codeflow?start=11&length=39)

The key concern here is round tripping. 
We want to produce an string that when parsed, gives us the value back. 

It's ok to be invariant culture, but I wonder if parse flags could trip us up. 
Is there an unambiguous way to ensure this is a decimal, regardless of parse flag?

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -637,39 +637,56 @@ BinderCheckTypeResult ReportInvalidOperation()
...
                             // Regular Addition
-                            var leftResAdd = CheckTypeCore(errorContainer, node.Left, leftType, DType.Number, /* coerced: */ DType.String, DType.Boolean, DType.UntypedObject);
-                            var rightResAdd = CheckTypeCore(errorContainer, node.Right, rightType, DType.Number, /* coerced: */ DType.String, DType.Boolean, DType.UntypedObject);
-
-                            // Deferred + number or number + Deferred
-                            if (leftKind == DKind.Deferred || rightKind == DKind.Deferred)
-                            {
-                                return new BinderCheckTypeResult() { Node = node, NodeType = DType.Deferred, Coercions = leftResAdd.Coercions.Concat(rightResAdd.Coercions).ToList() };
+                            // Only operations where both operands are Decimal-compatible result in Decimal
+                            if (DType.DecimalBinaryOp(leftType, rightType, numberIsFloat))
```
lesaltzm, line 662: >DType.DecimalBinaryOp [](http://example.com/codeflow?start=32&length=21)

This helper method should probably be in BinderUtils, not DType

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -721,16 +738,28 @@ private static BinderCheckTypeResult CheckComparisonArgTypesCore(IErrorContainer
                 else if (typeLeft.Kind == DKind.OptionSetValue)
                 {
                     // The other value (right in this case) needs to be a number or coerced to a number in order to compare
-                    var rightResult = CheckComparisonTypeOneOfCore(errorContainer, right, typeRight, DType.Number, DType.Date, DType.Time, DType.DateTime);
-                    coercions.AddRange(rightResult.Coercions);
+                    var rightResult = CheckComparisonTypeOneOfCore(errorContainer, right, typeRight, DType.Number, DType.Decimal, DType.Date, DType.Time, DType.DateTime);
+                    coercions.AddRange(rightResult.Coercions);
+
+                    // CheckComparisonTypeOneOfCore above will validate that Decimal is OK, but will not coerce to a number
+                    if (!DType.Number.Accepts(typeRight))
```
lesaltzm, line 745: good catch

src/libraries/Microsoft.PowerFx.Core/Binding/BindingConfig.cs
```diff
@@ -17,12 +17,15 @@ internal class BindingConfig
 
         public bool AllowsSideEffects { get; }
 
-        public bool UseThisRecordForRuleScope { get; }
+        public bool UseThisRecordForRuleScope { get; }
+
+        public bool NumberIsFloat { get; }
```
lesaltzm, line 23: Should this be on binding config, or should it be on Features?

lesaltzm: Binding Config is currently just used for permanent flags, not 1.0 features.

src/libraries/Microsoft.PowerFx.Core/Functions/CheckTypesContext.cs
```diff
@@ -20,16 +20,19 @@ internal class CheckTypesContext
 
         public bool IsEnhancedDelegationEnabled { get; }
 
-        public bool AllowsSideEffects { get; }
+        public bool AllowsSideEffects { get; }
+
+        public bool NumberIsFloat { get; }
```
lesaltzm, line 25: >NumberIsFloat [](http://example.com/codeflow?start=20&length=13)

same here wrt features

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -1080,6 +1086,26 @@ protected bool CheckBooleanColumnType(DType type, TexlNode arg, IErrorContainer
...
+        protected DType NumDecReturnType(CheckTypesContext context, DType argType)
+        {
+            IEnumerable<TypedName> columns;
+
+            if (argType.IsTable && (columns = argType.GetNames(DPath.Root)).Count() == 1)
+            {
+                argType = columns.Single().Type;
+            }
+
+            // Decimal TODO: Is this the right thing to do for deferred and unknown?
```
lesaltzm, line 1098: >Decimal TODO [](http://example.com/codeflow?start=15&length=12)

definitely for deferred, but I'm not entirely sure what this is trying to do otherwise.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -1080,6 +1086,26 @@ protected bool CheckBooleanColumnType(DType type, TexlNode arg, IErrorContainer
             return CheckColumnType(type, arg, DType.Boolean, errors, TexlStrings.ErrInvalidSchemaNeedColorCol_Col, ref nodeToCoercedTypeMap);
         }
 
+        protected DType NumDecReturnType(CheckTypesContext context, DType argType)
+        {
+            IEnumerable<TypedName> columns;
+
+            if (argType.IsTable && (columns = argType.GetNames(DPath.Root)).Count() == 1)
```
lesaltzm, line 1093: >columns = argType.GetNames(DPath.Root) [](http://example.com/codeflow?start=36&length=38)

is the assignment necessary? the added complexity makes the condition a bit hard to read, either pulling the assignment before the condition or just repeating the GetNames call would be better

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -93,6 +93,16 @@ public override IntermediateNode Visit(NumLitNode node, IRTranslatorContext cont
                 return MaybeInjectCoercion(node, new NumberLiteralNode(context.GetIRContext(node), value), context);
             }
 
+            public override IntermediateNode Visit(DecLitNode node, IRTranslatorContext context)
+            {
+                Contracts.AssertValue(node);
+                Contracts.AssertValue(context);
+
+                // I think node.NumValue might be dead code, this could be cleaned up
```
lesaltzm, line 101: >// I think node.NumValue might be dead code, this could be cleaned up [](http://example.com/codeflow?start=16&length=69)

copy/paste?

lesaltzm:
> // Licensed under the MIT license.

whitespace only changes?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/IR/Nodes/CallNode.cs:2 in c550917. [](commit_id = c550917917ae9dcbefbfc833b75f77f9ceca30f1, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Lexer/Tokens/NumLitToken.cs
```diff
@@ -14,7 +14,8 @@ public class NumLitToken : Token
     {
         internal NumLitToken(double value, Span span)
             : base(TokKind.NumLit, span)
-        {
+        {
+            // Decimal TODO: Why isn't this <= double.MaxValue?
```
lesaltzm, line 18: >Decimal TODO: Why isn't this <= double.MaxValue [](http://example.com/codeflow?start=15&length=47)

no idea, should be fine to change, I think?

src/libraries/Microsoft.PowerFx.Core/Localization/Strings.cs
```diff
@@ -76,7 +76,9 @@ internal static class TexlStrings
         public static StringGetter FirstLastArg1 = (b) => StringResources.Get("FirstLastArg1", b);
 
         public static StringGetter AboutFirstN = (b) => StringResources.Get("AboutFirstN", b);
-        public static StringGetter AboutLastN = (b) => StringResources.Get("AboutLastN", b);
+        public static StringGetter AboutLastN = (b) => StringResources.Get("AboutLastN", b);
+        public static StringGetter AboutFirstNW = (b) => StringResources.Get("AboutFirstNW", b);
```
lesaltzm, line 80: >AboutFirstNW [](http://example.com/codeflow?start=78&length=12)

unused?

src/libraries/Microsoft.PowerFx.Core/Localization/Strings.cs
```diff
@@ -97,7 +101,12 @@ internal static class TexlStrings
         public static StringGetter AboutBooleanN = (b) => StringResources.Get("AboutBooleanN", b);
         public static StringGetter BooleanNArg1 = (b) => StringResources.Get("BooleanNArg1", b);
         public static StringGetter AboutBooleanNT = (b) => StringResources.Get("AboutBooleanNT", b);
-        public static StringGetter BooleanNTArg1 = (b) => StringResources.Get("BooleanNTArg1", b);
+        public static StringGetter BooleanNTArg1 = (b) => StringResources.Get("BooleanNTArg1", b);
+
+        public static StringGetter AboutBooleanW = (b) => StringResources.Get("AboutBooleanW", b);
+        public static StringGetter BooleanWArg1 = (b) => StringResources.Get("BooleanWArg1", b);
+        public static StringGetter AboutBooleanWT = (b) => StringResources.Get("AboutBooleanWT", b);
+        public static StringGetter BooleanWTArg1 = (b) => StringResources.Get("BooleanWTArg1", b);
```
lesaltzm, line 110: Does this function need it's own strings? Is there a reason it shouldn't reuse Boolean's?

I see it was done for BooleanN and BooleanB, but that seems kind of silly as well.

src/libraries/Microsoft.PowerFx.Core/IR/Nodes/UnaryOpKind.cs
```diff
@@ -8,14 +8,22 @@ internal enum UnaryOpKind
         // Value operations
         Negate,
         Percent,
+        NegateDecimal,
+        PercentDecimal,
 
         // Coercion operations
         BooleanToNumber,
-        OptionSetToNumber,
+        OptionSetToNumber,  // Decimal TODO: this, and others, don't appear in the interpreter's LibaryUnary.cs?
```
lesaltzm, line 16: >// Decimal TODO: this, and others, don't appear in the interpreter's LibaryUnary.cs? [](http://example.com/codeflow?start=28&length=84)

out of date comment, yes it does

src/libraries/Microsoft.PowerFx.Core/Parser/ParseFormulasResult.cs
```diff
@@ -66,15 +66,18 @@ internal class UDF
 
         internal ISet<UDFArg> Args { get; }
 
-        internal bool IsImperative { get; }
+        internal bool IsImperative { get; }
+
+        internal bool NumberIsFloat { get; }
```
lesaltzm, line 71: >internal bool NumberIsFloat { get; } [](http://example.com/codeflow?start=8&length=36)

why does this need to go on the UDF classes?

src/libraries/Microsoft.PowerFx.Core/Public/ErrorKind.cs
```diff
@@ -36,7 +36,8 @@ public enum ErrorKind
         Numeric = 24,
         InvalidArgument = 25,
         Internal = 26,
-        NotApplicable = 27,
+        NotApplicable = 27,
+        Overflow = 28,          // Decimal TODO: fold into Numeric?  Odd error message in canvas for IfError(1e308*10,FirstError.Message) about non-finite number?  Where is Excel's #Value?
```
lesaltzm, line 40: this seems reasonable to me. 
You'll need to update the EnumStore as well with this.

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
```diff
@@ -233,7 +236,12 @@ public override int GetHashCode()
         internal virtual void DefaultExpressionValue(StringBuilder sb)
         {
             throw new NotSupportedException($"{GetType().FullName} doesn't implement DefaultExpressionValue.");
-        }
+        }
+
+        internal virtual void DefaultUniqueExpressionValue(StringBuilder sb)
+        {
+            DefaultExpressionValue(sb);
+        }
```
lesaltzm, line 244: When is this necessary? Would we ever expect serializers to not use unique expression value?

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/DecLitNode.cs
```diff
@@ -0,0 +1,73 @@
...
+    /// <summary>
+    /// Decimal literal parse node. Example:
+    /// 
+    /// <code>12.34</code>
+    /// </summary>
+    public sealed class DecLitNode : TexlNode
+    {
+        // If Value is non-null, then the token represents its value.
+        // Otherwise, the value is in DecValue.
+        internal readonly decimal DecValue;
```
lesaltzm, line 19: this seems confusing. Why shouldn't it always be in DecValue?

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/DecLitNode.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        /// <summary>
+        /// The numeric value of the node.
+        /// </summary>
+        public decimal ActualDecValue => Value?.Value ?? DecValue;
+
+        internal DecLitNode(ref int idNext, DecLitToken tok)
+            : base(ref idNext, tok, new SourceList(tok))
+        {
+            DecValue = 0;
```
lesaltzm, line 29: >0 [](http://example.com/codeflow?start=23&length=1)

could this be assigned to tok.Value?

lesaltzm:
> // Copyright (c) Microsoft Corporation.

whitespace only

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/NumLitNode.cs:1 in c550917. [](commit_id = c550917917ae9dcbefbfc833b75f77f9ceca30f1, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
```diff
@@ -16,7 +16,8 @@ internal class BuiltinFunctionsCore
         // Functions in this list are shared and may show up in other hosts by default.
         internal static readonly TexlFunctionSet _library = new TexlFunctionSet();
         private static readonly TexlFunctionSet _featureGateFunctions = new TexlFunctionSet();
-
+
+        // Decimal TODO: Best way to handle limiting exposure of Float and Decimal functions, W overloads, especially for Canvas
```
lesaltzm, line 20: >Decimal TODO: Best way to handle limiting exposure of Float and Decimal functions, W overloads, especially for Canvas [](http://example.com/codeflow?start=11&length=117)

see _featureGateFunctions as below

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Sequence.cs
```diff
@@ -24,6 +24,28 @@ public SequenceFunction()
...
+    {
+        public override ArgPreprocessor GetArgPreprocessor(int index)
+        {
+            return base.GetGenericArgPreprocessor(index);
+        }
+
+        public override bool IsSelfContained => true;
+
+        public SequenceWFunction()
+            : base("Sequence", TexlStrings.AboutSequence, FunctionCategories.MathAndStat, DType.CreateTable(new TypedName(DType.Decimal, new DName("Value"))), 0, 1, 3, DType.Decimal, DType.Decimal, DType.Decimal)
```
lesaltzm, line 45: this doesn't seem right. Do we really need a separate overload of the sequence function?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MinMaxTable.cs
```diff
@@ -16,52 +16,23 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
     // Min(source:*, projection:n)
     // Max(source:*, projection:n)
     // Corresponding DAX functions: Min, MinA, MinX, Max, MaxA, MaxX
-    internal sealed class MinMaxTableFunction : StatisticalTableFunction
+    internal sealed class MinTableFunction : StatisticalTableFunction
```
lesaltzm, line 19: Why is this being split up and could it be extracted to a separate PR?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/StatisticalTableFunction.cs
```diff
@@ -19,12 +20,18 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
     // a table as the first argument, and a value function as the second argument.
     internal abstract class StatisticalTableFunction : FunctionWithTableInput
     {
-        public override bool IsSelfContained => true;
-
-        public StatisticalTableFunction(string name, TexlStrings.StringGetter description, FunctionCategories fc)
-            : base(name, description, fc, DType.Number, 0x02, 2, 2, DType.EmptyTable, DType.Number)
+        public override bool IsSelfContained => true;
+
+        public bool _nativeDecimal;
```
lesaltzm, line 25: >public [](http://example.com/codeflow?start=8&length=6)

public?

lesaltzm, line 25: >_nativeDecimal [](http://example.com/codeflow?start=20&length=14)

comments explaining what these are for

lesaltzm: also, readonly

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -3535,5 +3554,28 @@ public bool TryGetSimilarName(DName name, FieldNameKind fieldNameKind, out strin
             return similar != null &&
                    comparer.Distance(similar) < (name.Value.Length / 3) + 3;
         }
+
+        // Determins if the result of a numeric binary operation should be a Decimal or a Number.
+        // If !numberIsFloat then Strings and Booleans are allowed to coerce to Decimal for this test, as if they were passed through Value
+        public static bool DecimalBinaryOp(DType leftType, DType rightType, bool numberIsFloat)
```
lesaltzm, line 3560: probably move to BindingHelper/BinderUtils,

Also, could be named something like: `IsBinaryOpResultDecimal`

src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
```diff
@@ -218,8 +218,10 @@ private FunctionDescr Scan()
             if (_info == null)
             {
                 var t = GetType();
-                var suffix = "Function";
-                var name = t.Name.Substring(0, t.Name.Length - suffix.Length);
+                var suffix = "Function";
+
+                //                 var name = t.Name.Substring(0, t.Name.Length - suffix.Length);  and replace Program.cs, Decimal TODO before PR
```
lesaltzm, line 223: >Decimal TODO before PR [](http://example.com/codeflow?start=123&length=22)

Before PR?

src/libraries/Microsoft.PowerFx.Core/Public/Types/DecimalType.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Text;
+using Microsoft.PowerFx.Core.Types;
+
+namespace Microsoft.PowerFx.Types
+{
+    public class DecimalType : FormulaType
```
MikeStall, line 9: >DecimalType [](http://example.com/codeflow?start=17&length=11)

Can we leverage this in SQL compiler and remove this:
[https://github.com/microsoft/Power-Fx-Dataverse/blob/main/src/PowerFx.Dataverse/Types/SqlDecimalType.cs](https://github.com/microsoft/Power-Fx-Dataverse/blob/main/src/PowerFx.Dataverse/Types/SqlDecimalType.cs)

src/libraries/Microsoft.PowerFx.Core/Public/Values/DecimalValue.cs
gregli-msft: Funny you should bring this up, I was just looking at the impact of this in src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs as it is has broken some tests.  If the primary mission of ToExpression is to be unambiguous, then yes, for now we can wrap all floating point values in Float( "number" ).  This would be a reason to add a literal suffix, like the `10.2m` or `10.2d` in C#, something I'm not completely opposed to, but was trying to avoid adding that concept, not that I think it is too copmlicated, just more concepts.  :)  Open to your thoughts on this.

src/libraries/Microsoft.PowerFx.Core/Binding/BindingConfig.cs
gregli-msft: I see NumberIsFloat as a permanent feature.  Folks like Azure IOT will never want the performance hit of working with decimal.  For this reason, I moved it out of features and into the config here, in the parser, and in the lexer.  Totally open to having this discussion.

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
gregli-msft: Yes, it comes from line 91 above it for NumLitNode.  I figured if it was true there, it is probably true here too.  Not sure if and how it should be cleaned up.

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/DecLitNode.cs
gregli-msft: I took this from NumLitNode which has the same comments and logic.  The only real difference between the two is that NumLitNode uses NaN and DecLitNode uses 0 as the default in the constructor, since decimal dos not have a NaN.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Sequence.cs
gregli-msft: Let's chat.  If we don't and rely on floating point, and someone uses the output of Sequence in a calculation, then the whole calculation is tainted with floating point.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MinMaxTable.cs
gregli-msft: It could be a separate PR.  It is being split up because after other changes to the class hierarchy and the removal of its own CheckTypes, there is little if any advantage to having them together.

src/libraries/Microsoft.PowerFx.Interpreter/ReflectionFunction.cs
gregli-msft: Oops, this is destined for a different PR.  I'll remove.

src/libraries/Microsoft.PowerFx.Core/Public/Types/DecimalType.cs
gregli-msft: Yes.  We should rework the SQL backend to use Decimal and then it can also support Float.  But I wanted to do that change as a separate PR, in coordination with the Dataverse team.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MathFunction.cs
```diff
@@ -1,51 +1,175 @@
...
+            for (int i = 0; i < argTypes.Length; i++)
+            {
+                if (CheckType(args[i], argTypes[i], returnType, DefaultErrorContainer, out var matchedWithCoercion))
+                {
+                    if (matchedWithCoercion)
+                    {
+                        if (nodeToCoercedTypeMap == null)
+                        {
+                            nodeToCoercedTypeMap = new Dictionary<TexlNode, DType>();
+                        }
```
lesaltzm, line 111: nit: CollectionUtils initializes this, no need to do it separately

src/libraries/Microsoft.PowerFx.Core/Public/Types/DecimalType.cs
MikeStall: Ack it's separate PR. 
I filed issues to track:
[https://github.com/microsoft/Power-Fx-Dataverse/issues/99](https://github.com/microsoft/Power-Fx-Dataverse/issues/99)
[https://github.com/microsoft/Power-Fx-Dataverse/issues/100](https://github.com/microsoft/Power-Fx-Dataverse/issues/100)

src/libraries/Microsoft.PowerFx.Core/Public/Values/DecimalValue.cs
gregli-msft: Actually, I think we should stick to number type invariant.  I'm running all of the expression tests right now in both Decimal and Float modes and with only a few exceptions everything passes.  I think this is a good thing - by default if someone adds a .txt file, it will automatically be tested in both modes.  The exceptions are due to inconsistencies in how C# serializes decimal and float numbers and in numbers that are beyond the decimal range.   I think we should either clean up the serialization inconsistencies and/or walk the tree to do type aware comparisons instead of string comparisons.  We should chat about the approach.

src/tests/Microsoft.PowerFx.Core.Tests/CheckResultTests.cs
```diff
@@ -182,7 +182,7 @@ public void BasicParseErrors2()
             // Still assign some types
             var node = ((BinaryOpNode)parse.Root).Left;
             var type = check.GetNodeType(node);
-            Assert.Equal(FormulaType.Number, type);
+            Assert.True(type == FormulaType.Number || type == FormulaType.Decimal);
```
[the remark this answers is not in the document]
gregli-msft: Good catch.  I made this change in the early days before I had an approach to testing established.  Tests have been restored to former glory, and some additional floating point tests added for good measure.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Value.cs
```diff
@@ -9,27 +10,25 @@
...
 using Microsoft.PowerFx.Syntax;
 
 namespace Microsoft.PowerFx.Core.Texl.Builtins
 {
     // Value(arg:s|n [,language:s])
     // Corresponding Excel and DAX function: Value
-    internal sealed class ValueFunction : BuiltinFunction
+    internal class ValueBaseFunction : BuiltinFunction
     {
-        public const string ValueInvariantFunctionName = "Value";
```
lesaltzm, line 20: this constant is used in Canvas code in a few places, can you bring it back?

lesaltzm:
For safety in getting this PR integrated in PA Client, can you split the PR up into 2?

Ideally, extracting the Function changes to a separate pr should work, and would allow us to reduce risk. In particular, changing several of those return types to `DType.Unknown` could be a potential issue for variable analysis, and would be good to review and integrate separately.

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -202,1408 +205,1411 @@ private bool ParseUDF(List<UDF> udfs)
...
-                    if (thisEq != null)
-                    {
+            var parser = new TexlParser(tokens, flags, features);
+            List<TexlError> errors = null;
+            var parsetree = parser.Parse(ref errors);
+
+            return new ParseResult(parsetree, errors, errors?.Any() ?? false, parser._comments, parser._before, parser._after, script, loc);
+        }
+
+        public static ParseFormulasResult ParseFormulasScript(string script, CultureInfo loc = null)
```
lesaltzm, line 238: This needs a param to enable parsing number as float.

lesaltzm: Also, pass through EnsureParsed in NamedFormulas.cs

gregli-msft:
Broken into smaller PRs.  Infrastructure is in https://github.com/microsoft/Power-Fx/pull/1237

PR #1222, opened 2023-03-21
Add connector x-ms-dynamic-values management

LucGenetier:
This is used in intellisense, for proposing, for example, stored procedure paramater names in SQL connector


LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

PR #1224, opened 2023-03-21
Change UserInfo to User for user object

nguhoa:
Follow https://github.com/microsoft/Power-Fx/pull/1214


src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Error.cs
```diff
@@ -78,7 +78,7 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
 
             // First handle required fields (currently only 'Kind')
             var argumentKind = names.FirstOrDefault(tn => tn.Name == requiredKindField.Name);
-            if (argumentKind == null)
+            if (argumentKind.Type == null)
```
MikeStall, line 81: different change?

nguhoa: moved this back to old code.

MikeStall (approved):

MikeStall (approved):

PR #1228, opened 2023-03-22
Update .txt tests to use the features enabled in Power Apps by default

CarlosFigueiraMSFT:
We currently have tests that run in Power Fx but are being ignored in Power Apps, because the JS runner doesn't understand some of the #setup directives. This updates the tests to use the same defaults as Power Apps, so that the tests won't have any setup directives, and will run properly in the Power Apps repository.


MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/TestRunnerTests/InternalSetupTests.cs
```diff
@@ -100,13 +100,13 @@ public void InternalSetup_Parse_HandlerAndTwoFlags2()
         }
 
         [Fact]
-        public void InternalSetup_Parse_TableSyntaxDoesntWrapRecordsFlags()
+        public void InternalSetup_Parse_DisableTableSyntaxDoesntWrapRecordsFlags()
```
Ananya-Gupta24, line 103: Could you also add a test for both flags disabled in case we need to disable multiple default features in the future?

Ananya-Gupta24 (dismissed):

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/TestRunnerTests/InternalSetupTests.cs
CarlosFigueiraMSFT: Added that and a few more tests

LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #1229, opened 2023-03-22
adds void type support to Switch()

### LINKED ISSUE #1201: Review Switch()'s checkType and add void type support.
-Currently in Switch()'s check type we allow usage of the result of the switch statement, we shouldn't be doing that.
e.g. `Set(x, Switch(20, 20, 1, 21, {x:1}))` should not be allowed.

-Return void when the switch() has the arg type mismatch.

jas-valgotar:
-Fixes #1201 
-Currently in Switch()'s check type we allow usage of the result of the switch statement even when arg is mismatched, we shouldn't be doing that.
e.g. `Set(x, Switch(20, 20, 1, 21, {x:1}))` should not be allowed.

-Return void when the switch() has the arg type mismatch.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Switch.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Switch.cs
@@ -112,19 +112,12 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
             var type = ReturnType;
             nodeToCoercedTypeMap = null;
 
-            // Are we on a behavior property?
-            var isBehavior = context.AllowsSideEffects;
-
             // Compute the result type by joining the types of all non-predicate args.
             Contracts.Assert(type == DType.Unknown);
             for (var i = 2; i < count;)
             {
                 var nodeArg = args[i];
                 var typeArg = argTypes[i];
-                if (typeArg.IsError)
-                {
-                    errors.EnsureError(args[i], TexlStrings.ErrTypeError);
-                }
 
                 var typeSuper = DType.Supertype(type, typeArg);
 
@@ -132,9 +125,13 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
                 {
                     type = typeSuper;
                 }
-                else if (type.Kind == DKind.Unknown)
+                else if (typeArg.IsVoid)
                 {
-                    type = typeSuper;
+                    type = DType.Void;
+                }
+                else if (typeArg.IsError)
+                {
+                    errors.EnsureError(args[i], TexlStrings.ErrTypeError);
                     fArgsValid = false;
                 }
                 else if (!type.IsError)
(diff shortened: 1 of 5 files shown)


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/switch.txt
```diff
@@ -42,4 +42,21 @@ true
...
+// Result of switch can't be consumed.
+>> Switch(1, 1, 2, {x:4}) + 1
+Errors: Error 0-22: Invalid argument type. Expecting one of the following: Number, Text, Boolean, UntypedObject.
+
+// Expression generating void value can be used inside of the switch. If(1<0, 1, {x:1}) => void value
+>> Switch(1, 1, 2, If(1<0, 1, {x:4}))
+If(true, {test:1}, "Mismatched args (result of the expression can't be used).")
+
+>> Switch(1, 1, 1/0, 10, {x:4})
+Error({Kind:ErrorKind.Div0})
```
CarlosFigueiraMSFT, line 56: Nice test. it would be good to have a similar one where an error is not returned

    >> Switch(10, 1, 1/0, 10, {x:4})
    If(true, {test:1}, "Mismatched args...")

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Switch.cs
```diff
@@ -112,29 +112,26 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
...
             // Compute the result type by joining the types of all non-predicate args.
             Contracts.Assert(type == DType.Unknown);
             for (var i = 2; i < count;)
             {
                 var nodeArg = args[i];
                 var typeArg = argTypes[i];
-                if (typeArg.IsError)
-                {
-                    errors.EnsureError(args[i], TexlStrings.ErrTypeError);
-                }
```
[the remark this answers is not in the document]
jas-valgotar: This check is just moved down to an else if brnach. Since we should be marking fArgsValid to false anytime we encounter error arg, previously it would only mark fArgsValid false when the first arg is error.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/switch.txt
jas-valgotar: added.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/switch.txt
```diff
@@ -309,3 +309,24 @@ Error({Kind:ErrorKind.InvalidArgument})
...
+
+>> Switch("a", "a", 2, {x:4}, 3)
+Errors: Error 20-25: Invalid argument type (Record). Expecting a Text value instead.|Error 0-29: The function 'Switch' has some invalid arguments.
+
+>> Switch(1, 2, 2, 3, b)
+Errors: Error 19-20: Name isn't valid. 'b' isn't recognized.|Error 0-21: The function 'Switch' has some invalid arguments.
+
+>> Switch(10, 1, 1/0, 10, {x:4})
+If(true, {test:1}, "Mismatched args (result of the expression can't be used).")
+
```
CarlosFigueiraMSFT, line 332: Can you move this test together with the one on line 321, and add a comment that compares the two? For example: "// Void expressions can become error" in the first, and "// Void expressions that are not errors have a void value"

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/switch.txt
jas-valgotar: Updated.

CarlosFigueiraMSFT (approved):
🚢

LucGenetier:
✅ No public API change.

PR #1230, opened 2023-03-22
adds void type to IfError()

### LINKED ISSUE #1173: Review IfError()'s CheckType() and add void type support.
`IfError()` has buggy implementation of CheckType(),

e.g. 
below expression should be an error:
`IfError(1, {x:1}, {x:1}, {y:1}, 1)` Since it can potentially return a record or a number type

In both cases, it should have been consistent, since the possible result type can be either Number or Time type for both the expression and should have been super typed to the same thing...
`IfError(1, 1, Time(12,0,0)) => Time Type` 
`IfError(1, Time(12,0,0)) => Number Type`

-Also add void type support for mismatched args.

jas-valgotar:
-Fixes #1173 

`IfError()` has buggy implementation of CheckType(),

e.g. 
below expression should be an error (now error in the PR):
`IfError(1, {x:1}, {x:1}, {y:1}, 1)` Since it can potentially return a record or a number type

In both cases, it should have been consistent, since the possible result type can be either Number or Time type for both the expression and should have been super typed to the same thing...
`IfError(1, 1, Time(12,0,0)) => Time Type` 
`IfError(1, Time(12,0,0)) => Number Type`
Now Above both are Number Type.

-Also add void type support for mismatched args.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IfError.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IfError.cs
@@ -121,6 +121,9 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
             return result;
 
             In this case, we prefer the type of the last v_i as the return type.
+            Possible return values are:
+            Even case: f1, f2, ..., vn, fn
+            Odd case: f1, f2, ..., vn
             */
 
             var preferredTypeIndex = (count % 2) == 0 ? count - 2 : count - 1;
@@ -136,7 +139,7 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
                 var nodeArg = args[i];
                 var typeArg = argTypes[i];
 
-                var typeSuper = DType.Supertype(type, typeArg);
+                var typeSuper = type.IsError ? typeArg : DType.Supertype(type, typeArg);
 
                 if (!typeSuper.IsError)
                 {
@@ -175,73 +178,6 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
 
             returnType = type;
             return fArgsValid;
-        }
-
-        // In behavior properties, the arg type is irrelevant if nothing actually depends
-        // on the output type of IfError (see If.cs, Switch.cs)
-        private bool IsArgTypeInconsequential(TexlNode arg)
-        {
-            Contracts.AssertValue(arg);
-            Contracts.Assert(arg.Parent is ListNode);
-            Contracts.Assert(arg.Parent.Parent is CallNode);
-            Contracts.Assert(arg.Parent.Parent.AsCall().Head.Name == Name);
-
-            var call = arg.Parent.Parent.AsCall().VerifyValue();
-
-            // Pattern: OnSelect = IfError(arg1, arg2, ... argK)
-            // Pattern: OnSelect = IfError(arg1, IfError(arg1, arg2,...), ... argK)
-            // ...etc.
-            var ancestor = call;
-            while (ancestor.Head.Name == Name)
-            {
-                if (ancestor.Parent == null && ancestor.Args.Children.Length > 0)
-                {
-                    return true;
-                }
-
-                // Deal with the possibility that the ancestor may be contributing to a chain.
-                // This also lets us cover the following patterns:
-                // Pattern: OnSelect = X; IfError(arg1, arg2); Y; Z
-                // Pattern: OnSelect = X; IfError(arg1;arg11;...;arg1k, arg2;arg21;...;arg2k); Y; Z
-                // ...etc.
-                VariadicOpNode chainNode;
-                if ((chainNode = ancestor.Parent.AsVariadicOp()) != null && chainNode.Op == VariadicOp.Chain)
-                {
-                    // Top-level chain in a behavior rule.
-                    if (chainNode.Parent == null)
-                    {
-                        return true;
-                    }
-
-                    // A chain nested within a larger non-call structure.
-                    if (!(chainNode.Parent is ListNode) || !(chainNode.Parent.Parent is CallNode))
-                    {
-                        return false;
-                    }
-
-                    // Only the last chain segment is consequential.
-                    var numSegments = chainNode.Children.Length;
-                    if (numSegments > 0 && !arg.InTree(chainNode.Children[numSegments - 1]))
-                    {
-                        return true;
-                    }
-
-                    // The node is in the last segment of a chain nested within a larger invocation.
-                    ancestor = chainNode.Parent.Parent.AsCall();
-                    continue;
-                }
-
-                // Walk up the parent chain to the outer invocation.
-                if (!(ancestor.Parent is ListNode) || !(ancestor.Parent.Parent is CallNode))
-                {
-                    return false;
-                }
-
-                ancestor = ancestor.Parent.Parent.AsCall();
-            }
-
-            // Exhausted all supported patterns.
-            return false;
         }
 
         public override bool IsLambdaParam(int index)
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IfError.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IfError.cs
@@ -96,75 +96,60 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
             var count = args.Length;
             nodeToCoercedTypeMap = null;
 
-            // Check the predicates.
             var fArgsValid = true;
-            var type = ReturnType;
 
-            var isBehavior = context.AllowsSideEffects;
+            // if count is 2, then first argument can be returned.
+            var type = count == 2 ? argTypes[0] : ReturnType;
 
-            Contracts.Assert(type == DType.Unknown);
-            for (var i = 0; i < count;)
+            for (var i = 1; i < count;)
             {
                 var nodeArg = args[i];
                 var typeArg = argTypes[i];
 
-                if (typeArg.IsError)
+                var typeSuper = DType.Supertype(type, typeArg);
+                if (!typeSuper.IsError)
+                {
+                    type = typeSuper;
+                }
+                else if (typeArg.IsVoid)
+                {
+                    type = DType.Void;
+                }
+                else if (typeArg.IsError)
                 {
                     errors.EnsureError(args[i], TexlStrings.ErrTypeError);
+                    fArgsValid = false;
                 }
+                else if (!type.IsError)
+                {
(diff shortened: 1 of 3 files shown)


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/IfError.txt
```diff
@@ -195,3 +198,24 @@ Time(11,30,59,0)
...
+If(true, {test:1}, "Mismatched args (result of the expression can't be used).")
+
+// IfError accepts the void value. But then result of the expression can't be used. If(true, 1, {x:4}) => void value.
+>> IfError(1/0, 1, If(true, 1, {x:4}), 1)
+If(true, {test:1}, "Mismatched args (result of the expression can't be used).")
+
+// Errors are still returned.
+>> IfError(1, 2, 3/0, If(true, 1/0, {x:4}))
+Error({Kind:ErrorKind.Div0})
+
```
CarlosFigueiraMSFT, line 216: It would be good to have a similar test where the error is not returned:

    >> IfError(1, 2, 3/0, If(false, 1/0, {x:4}))
    If(true, {test:1}, "Mismatched...")

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/IfError.txt
```diff
@@ -195,3 +198,24 @@ Time(11,30,59,0)
...
+>> IfError(1, a)
```
CarlosFigueiraMSFT, line 217: Another interesting test would be if we have a Void expression inside IfError that cannot be returned - in this case the result should be non-Void. Something along the lines of

    >> IfError( If(true, 1, {a:1}), 1, 2 )
    2

    >> IfError( If(true, 1/0, {a:1}), 1, 2 )
    1

    >> IfError( 1, "hello", If(true,1/0,{a:1}), "great", "world")
    "great"

    >> IfError( 1, "one", If(true,1,{a:1}), "two", "three")
    "three"

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IfError.cs
```diff
@@ -96,75 +96,60 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
...
-                            fArgsValid = false;
-                        }
-                    }
-                    else if (typeArg.Kind != DKind.Unknown)
-                    {
-                        type = typeArg;
-                        fArgsValid = false;
+                        // If the types are incompatible, the result type is void.
+                        type = DType.Void;
+                        break;
```
CarlosFigueiraMSFT, line 134: >break [](http://example.com/codeflow?start=24&length=5)

Should we continue (instead of breaking) to see if there are any arguments with error types?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1967,18 +1967,20 @@ public static async ValueTask<FormulaValue> IfError(EvalVisitor runner, EvalVisi
...
                 {
-                    return res.ToFormulaValue();
+                    return MaybeAdjustToCompileTimeType(res.ToFormulaValue(), irContext);
                 }
 
                 if (i + 2 == args.Length - 1)
                 {
                     var falseBranch = args[i + 2];
-                    return (await runner.EvalArgAsync<ValidFormulaValue>(falseBranch, context, falseBranch.IRContext)).ToFormulaValue();
+                    var falseBranchResult = (await runner.EvalArgAsync<ValidFormulaValue>(falseBranch, context, falseBranch.IRContext)).ToFormulaValue();
```
CarlosFigueiraMSFT, line 1982: >falseBranchResult [](http://example.com/codeflow?start=24&length=17)

suggestion: 'elseBranchResult' or 'otherwiseBranchResult' - this is not an if (true/false)

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/IfError.txt
jas-valgotar: Added!

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/IfError.txt
jas-valgotar: Added! Thanks for the tests. :)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
jas-valgotar: updated

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IfError.cs
jas-valgotar: you are correct, removed the `break;`

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/IfError.txt
```diff
@@ -149,6 +149,9 @@ Error({Kind:ErrorKind.InvalidArgument})
 >> IfError(1/0, Time(12,0,0))
 0.5
 
+>> IfError(1, 1/0, Time(12,0,0))
+0.5
```
CarlosFigueiraMSFT, line 153: This currently returns Time(12,0,0,0) - a time value - not a number value.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/IfError.txt
jas-valgotar: updated as discussed offline

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IfError.cs
```diff
@@ -96,77 +96,80 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
...
+                return fn();  // omit this last catch if there is an odd number of arguments
+            }
+            return result;
 
-            Contracts.Assert(type == DType.Unknown);
-            for (var i = 0; i < count;)
+            In this case, we prefer the type of the last v_i as the return type.
+            */
+
+            var prefereedTypeIndex = (count % 2) == 0 ? count - 2 : count - 1;
```
CarlosFigueiraMSFT, line 126: >prefereed [](http://example.com/codeflow?start=16&length=9)

[typo] prefered (or preferred), no 'ee'

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IfError.cs
```diff
@@ -96,77 +96,80 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
...
-                        fArgsValid = false;
+                        type = DType.Void;
                     }
                 }
 
-                // If there are an odd number of args, the last arg also participates.
-                i += 2;
-                if (i == count)
+                if ((count % 2) != 0 && i == count - 1)
                 {
```
CarlosFigueiraMSFT, line 171: Since you already handled the odd argument case (preferredTypeIndex on line 126-127), do you still need this?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IfError.cs
```diff
@@ -96,77 +96,80 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
...
+
+            var preferredTypeIndex = (count % 2) == 0 ? count - 2 : count - 1;
+            var type = argTypes[preferredTypeIndex];
+            if (type.IsError)
+            {
+                errors.EnsureError(args[preferredTypeIndex], TexlStrings.ErrTypeError);
+                fArgsValid = false;
+            }
+
+            for (var i = count - 1; i >= 0; i -= 2)
```
CarlosFigueiraMSFT, line 134: >for (var i = count - 1; i >= 0; i -= 2) [](http://example.com/codeflow?start=12&length=39)

Why going backwards, when most functions go forwards? Not a problem with this PR, just curiosity :-)

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IfError.cs
jas-valgotar: Yes, because I need to start at the end of arg list (for both even and odd).
e.g. even arg case `IfError(v1, f1, v2, f2)` 
preferred return type is v2. But f2 can also be returned and need to type check it.

When it comes to odd arg
e.g. odd arg case `IfError(v1, f1, v2, f2, v3)`
preferred return type is v3. but f1 or f2 could also possibly be returned. so this line adjusts loop to scan f1, f2 and typecheck them.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IfError.cs
jas-valgotar: It is so,
In Case of last arg is Blank() or error, I can have a return type from last possible arg that can be returned.
e.g. 
even arg `If(v1, f1, Blank(), f2)` return type would be same as f2. and if f2 is blank then of f1 and so on...
odd arg `If(v1, f1, Blank())` return type would be same as f1.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IfError.cs
jas-valgotar: updated

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #1231, opened 2023-03-22
Add IR printing with type and Import improvements to Console REPL

gregli-msft:
Adds a pseudo-function IR() to the REPL which is very handy to see the IR for an expression.  Added type information which is also very handy to have.  Note that the expression does not need to be quoted.

```
> ForAll([1,2,3], Mid("World", Value, 1))
["W", "o", "r"]

> IR( ForAll([1,2,3], Mid("World", Value, 1)) )
ForAll:*[Value:s], Scope 1(Table:*[Value:n]({Value: 1:n}, {Value: 2:n}, {Value: 3:n}), Lazy(Mid:s("World":s, ScopeAccess(Scope 1, Value), 1:n)))
```

Totally open to feedback on the new format.

All changes here are not Power Fx maker facing.  This change opens up a small amount of public surface area CheckResult.PrintIR() which returns the IR pretty printing.  I don't think this would be an issue, but wanted to call it out.

Also adds a second argument to Import to write out the results of the run.  Can be used to feed REPL with a raw list of tests, and it will emit the correct output for making it into a test .txt file.

ReflectionFunction was changed to support more than one function with the same name, providing the ability to do parameter overloading, used here for a 1 arg and 2 arg version of Import.


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/IR/Nodes/BinaryOpNode.cs
```diff
@@ -29,7 +29,7 @@ public BinaryOpNode(IRContext irContext, BinaryOpKind op, IntermediateNode left,
 
         public override string ToString()
         {
-            return $"BinaryOp({Op}, {Left}, {Right})";
+            return $"{Op}:{IRContext.ResultType._type}({Left}, {Right})";
```
LucGenetier, line 32: >{IRContext.ResultType._type} [](http://example.com/codeflow?start=26&length=28)

I like the format for booleans/numbers where you have ":b" or ":n" suffix.
Here it's a prefix.
Not sure which one is best but it could be nice to be consistent

src/libraries/Microsoft.PowerFx.Core/IR/Nodes/CallNode.cs
```diff
@@ -75,23 +75,19 @@ public bool IsLambdaArg(int i)
         }
 
         public override string ToString()
```
LucGenetier, line 77: I'd keep the use of parenthesis for calls

src/libraries/Microsoft.PowerFx.Core/IR/Nodes/TextLiteralNode.cs
```diff
@@ -20,7 +20,7 @@ public TextLiteralNode(IRContext irContext, string value)
 
         public override string ToString()
         {
-            return $"Text({LiteralValue})";
+            return $"\"{LiteralValue}\":s";
```
LucGenetier, line 23: >LiteralValue [](http://example.com/codeflow?start=24&length=12)

Is the result ok if that string contains double quotes?

src/tests/Microsoft.PowerFx.Core.Tests/CheckResultTests.cs
```diff
@@ -394,7 +394,7 @@ public void TestIR()
 
             var ir = check.ApplyIR();
             Assert.NotNull(ir);
-            Assert.Equal("BinaryOp(AddNumbers, Number(1), Number(2))", ir.TopNode.ToString());
+            Assert.Equal("AddNumbers:n(1:n, 2:n)", ir.TopNode.ToString());
```
LucGenetier, line 397: >"AddNumbers:n(1:n, 2:n) [](http://example.com/codeflow?start=25&length=23)

I'd prefer AddNumbers(1:n, 2:n):n
Feel free to disagree

src/tools/Repl/Program.cs
```diff
@@ -138,11 +164,23 @@ public static void REPL(TextReader input, bool echo)
...
+                        var opts = new ParserOptions() { AllowsSideEffects = true };
+                        var cr = _engine.Check(match.Groups["expr"].Value, options: opts);
+                        var ir = cr.PrintIR();
+                        Console.WriteLine(ir);
+                        output?.WriteLine(ir);
                     }
 
                     // formula definition: <ident> = <formula>
-                    else if ((match = Regex.Match(expr, @"^\s*(?<ident>\w+)\s*=(?<formula>.*)$", RegexOptions.Singleline)).Success)
+                    else if ((match = Regex.Match(expr, @"^\s*(?<ident>[a-zA-Z]\w+)\s*=(?<formula>.*)$", RegexOptions.Singleline)).Success &&
```
LucGenetier, line 182: >a-zA-Z [](http://example.com/codeflow?start=72&length=6)

numbers, underscore ?

src/libraries/Microsoft.PowerFx.Core/IR/Nodes/BinaryOpNode.cs
gregli-msft: This is postfix notation too.  `1+2` results in `AddNumbers:n(1:n, 2:n)`

src/tests/Microsoft.PowerFx.Core.Tests/CheckResultTests.cs
gregli-msft: I hear you, it would overall be more consistent.  My concern is that for lots of nested calls in a complex epxression it will become difficult to match up the return type with the function. 

For example: `IR(Max(Max(Max(Today(),DateAdd(Today(),30)),Now()),Sqrt(2)))` would result in 
```
Max:n(DateTimeToNumber:n(Max:d(DateToDateTime:d(Max:D(Today:D(), DateAdd:D(Today:D(), 30:n))), Now:d())), Sqrt:n(2:n))
```

If the end type was after the closing paren, then it would be:
```
Max(DateTimeToNumber(Max(DateToDateTime(Max(Today():D, DateAdd:D(Today():D, 30:n):D):d):d, Now():d):n), Sqrt:n(2:n)):n
```
What's the return type of the second Max?  It's not as clear, one has to do a lot of paren matching.

So, I'd prefer with the function name for now.  We can try it this way and change it later if desired.

src/tools/Repl/Program.cs
```diff
@@ -138,11 +164,23 @@ public static void REPL(TextReader input, bool echo)
                     // variable assignment: Set( <ident>, <expr> )
                     if (TryMatchSet(expr, out var varName, out var varValue))
                     {
-                        Console.WriteLine(varName + ": " + PrintResult(varValue));
+                        Console.WriteLine(varName + ": " + PrintResult(varValue));
+                        output?.WriteLine(varName + ": " + PrintResult(varValue));
+                    }
+
+                    // IR pretty printer: IR( <expr> )
+                    else if ((match = Regex.Match(expr, @"^\s*IR\((?<expr>.*)\)\s*$", RegexOptions.Singleline)).Success)
```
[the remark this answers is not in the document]
gregli-msft: Nope, don't need to.  It's a pseudo function that always returns true, it can be restricted to just one at a time.  We don't support the chaining operator with the Set function either, for example `Set(x,3); Set(y,5)` is not supported.

src/tools/Repl/Program.cs
gregli-msft: Good catch.  There are a few bugs here:
* It should support underscore, but not numbers.  This is why I had changed it to begin with.  Numbers are actually supported with single quotes.
* Single quotes aren't supported
* As currently written, it would require at least two characters, when one should suffice.

Updates made.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -535,6 +535,14 @@ internal IRResult ApplyIR()
             return _irresult;
         }
 
+        // pretty print IR for debugging purposes, used by the Console REPL
+        public string PrintIR()
+        {
+            var x = this.ApplyIR().TopNode;
```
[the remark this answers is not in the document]
gregli-msft: Good point.  Fixed.

src/libraries/Microsoft.PowerFx.Core/IR/Nodes/TextLiteralNode.cs
gregli-msft: Good call.  I changed it to escape the string, as is done for ToExpression string values.

src/libraries/Microsoft.PowerFx.Core/IR/Nodes/CallNode.cs
gregli-msft: Discussed in another comment.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/IR/Nodes/BinaryOpNode.cs
LucGenetier: Why not `AddNumbers(1:n, 2:n):n` ?

LucGenetier (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1234, opened 2023-03-22
Add TexlTest for Error function

CarlosFigueiraMSFT:
We recently had a bug added to the Error function (in #1206), and it was fixed in the next commit (#1224). This adds some additional tests to prevent the error from happening in the first place.


jas-valgotar (approved):

LucGenetier:
✅ No public API change.

Issue #1205, opened 2023-03-16
Patch, Collect, and ClearCollect should support coercion for record fields


gregli-msft:
Today, this gives a compile time error in the C# Repl:

```
> Set( y, [1,2,3] )
y: [1, 2, 3]

> Patch( y, First(y), { Value:"1" } )
One or more errors occurred. (Errors: Error 20-33: The type of this argument 'Value' does not match the expected type 'Number'. Found type 'Text'.
Error 0-35: The function 'Patch' has some invalid arguments.)

> Collect( y, { Value:"1" } )
One or more errors occurred. (Errors: Error 12-25: The type of this argument 'Value' does not match the expected type 'Number'. Found type 'Text'.
Error 0-27: The function 'Collect' has some invalid arguments.)
```

Likewise it fails in Canvas.  However, this works because the operator is doing the coercion for the maker.  This is inconsistent and inconsistent with the coercion already supported with the [ ] operator and planned for the Table function.  In general, we have decided to support coercion of field values in records.

```
> Patch( y, First(y), { Value:"1"+0 } )
{Value:1}

> Collect( y, { Value:"1"+0 } )
{Value:1}
```
We should change the Patch, Collect, and ClearCollect functions to support field coercion.

gregli-msft:
As discussed in Chat, we should add field coercion to the Update and Remove functions too.  If a field is using a Number as its type, it won't work with a Decimal constant without this coercion.  

That said, this can come after the original PR.

Issue #1204, opened 2023-03-16
Table function should support coercion like [ ] does


gregli-msft:
The Table function and [ ] notation differ in how they handle coercion.  

`[ {a:1}, {a:"1"}, {a:true} ]` returns a table with three records, each record with the value {a:1}, as the string and Boolean are coerced to the first instance of a which is a number.

`Table( {a:1}, {a:"1"}, {a:true} )` returns compile time errors:
One or more errors occurred. (Errors: Error 0-33: The function 'Table' has some invalid arguments.
Error 14-21: Cannot use this record. It may contain colliding fields of incompatible types.
Error 23-31: Cannot use this record. It may contain colliding fields of incompatible types.)

Table should be changed to support the same coercion rules that [ ] follows.

PR #1235, opened 2023-03-22
Draft. These changes fix 1162, 1204, 1205 (wip).

### LINKED ISSUE #1162: Columns dropped when coercing columns in record-to-record coercion

anderson-joyle:
These changes fix #1162, #1204, #1205 (wip).

- #1162: Two problems were happening. 
1) When building the aggregate types coercion list, IR was not including fields that don't need coercion.
https://github.com/microsoft/Power-Fx/blob/72b2496154a662741b9e9ddd69d6fc5c8e57f58f/src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs#L748
2) Due to how `DType.Union` currently works, field types needing coercion were dropped from the result type. So even if the aggregate types coercion list included all fields since the beginning, PFx would still 
"drop" the columns since it's no longer part of the result type e.g. `[{a:1}, {a:true, b:2}]` result type would be '![a:n]'.
https://github.com/microsoft/Power-Fx/blob/72b2496154a662741b9e9ddd69d6fc5c8e57f58f/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs#L5216

- #1204: `Table` tried to `DType.Union` all the aggregate arguments without trying to coerce the field types, like `[]` does. Now `Table` shares `[]` behavior. This issue is deeply linked with #1162 since `Table` would also drop columns if we tried to union and coerce aggregate types only without fixing the IRTranslator first.

- #1205: Work in progress. 

Some comments:
- This is a draft PR, and the code needs to get cleaned up. I'm looking for feedback about the UnionWithCoercion idea.
- There is no test case in this PR yet. Manual tests have been performed and will be added once I move this out of draft.


src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -5205,6 +5205,7 @@ public override void PostVisit(TableNode node)
                     }
                     else if (isChildTypeAllowedInTable && childType.CoercesTo(exprType))
                     {
+                        exprType = DType.UnionWithCoercion(exprType, childType);
```
CarlosFigueiraMSFT, line 5208: This should be done based on a feature, so that we don't update Power Apps until it's ready to take this change.

src/libraries/Microsoft.PowerFx.Core/Public/Config/Features.cs
```diff
@@ -45,6 +45,12 @@ public enum Features : int
         /// </summary>
         StronglyTypedBuiltinEnums = 0x10,
 
+        /// <summary>
+        /// Updates the IsEmpty function to only allow table arguments, since it
+        /// does not work properly with other types of arguments.
+        /// </summary>
+        RestrictedIsEmptyArguments = 0x20,
```
CarlosFigueiraMSFT, line 52: This change is already in main, it should not be shown in the diff for your PR; maybe you need to rebase / merge from the latest main branch?

PR #1238, opened 2023-03-23
don't fail on passing a polymorphic type through.

MikeStall:


LucGenetier:
✅ No public API change.

anderson-joyle (approved):
:shipit:

PR #1245, opened 2023-03-23
Support DataEntity type within mutation functions

anderson-joyle:
To support https://github.com/microsoft/Power-Fx-Dataverse/issues/102.

Problem: Power Fx Dataverse would crash when we tried to Collect/Patch a record with a non-existing column. A new validation was placed to prevent this from happening. This validation checked if a column with the same name and type what present. If false, the mutation function returned an error. This new validation could not compare a LookUp field (Dtype.DataEntity) and a record type.

Solution: Expand the DataEntity type into a Record type and then compare the types.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

ian-legler (approved):

PR #1254, opened 2023-03-24
Remove enums that are not used in Power Fx

CarlosFigueiraMSFT:
The Power Fx project has lots of enums that are only used in Power Apps; this removes them from this repository, and will be moved to the next one.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

lesaltzm (approved):

PR #1258, opened 2023-03-27
Draft. Fixes for 1162, 1204, 1205. WIP

### LINKED ISSUE #1162: Columns dropped when coercing columns in record-to-record coercion

anderson-joyle:
These changes fix #1162, #1204, #1205 (wip).

- #1162: Two problems were happening. 
1) When building the aggregate types coercion list, IR was not including fields that don't need coercion.
https://github.com/microsoft/Power-Fx/blob/72b2496154a662741b9e9ddd69d6fc5c8e57f58f/src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs#L748
2) Due to how `DType.Union` currently works, field types needing coercion were dropped from the result type. So even if the aggregate types coercion list included all fields since the beginning, PFx would still 
"drop" the columns since it's no longer part of the result type e.g. `[{a:1}, {a:true, b:2}]` result type would be '![a:n]'.
https://github.com/microsoft/Power-Fx/blob/72b2496154a662741b9e9ddd69d6fc5c8e57f58f/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs#L5216

- #1204: `Table` tried to `DType.Union` all the aggregate arguments without trying to coerce the field types, like `[]` does. Now `Table` shares `[]` behavior. This issue is deeply linked with #1162 since `Table` would also drop columns if we tried to union and coerce aggregate types only without fixing the IRTranslator first.

- #1205: Work in progress. 

Some comments:
- This is a draft PR, and the code needs to get cleaned up. I'm looking for feedback about the UnionWithCoercion idea.
- Some test cases have been added but more will be added.


src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/PatchFunction.cs
```diff
@@ -118,6 +118,8 @@ internal class PatchFunction : PatchAndValidateRecordFunctionBase, IAsyncTexlFun
     {
         public override bool IsSelfContained => false;
 
+        public override bool SupportsParamCoercion => true;
```
CarlosFigueiraMSFT, line 121: >public override bool SupportsParamCoercion => true; [](http://example.com/codeflow?start=8&length=51)

This is the default for TexlFunction, no need to override it

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -5201,10 +5203,15 @@ public override void PostVisit(TableNode node)
                     }
                     else if (isChildTypeAllowedInTable && exprType.CanUnionWith(childType))
                     {
-                        exprType = DType.Union(exprType, childType);
+                        exprType = DType.Union(exprType, childType, powerfxV1: powerfx1);
                     }
                     else if (isChildTypeAllowedInTable && childType.CoercesTo(exprType))
                     {
+                        if (powerfx1)
```
anderson-joyle, line 5210: No need for the `if`

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -5201,10 +5203,15 @@ public override void PostVisit(TableNode node)
...
+                        {
+                            exprType = DType.Union(exprType, childType, powerfxV1: powerfx1);
```
anderson-joyle, line 5212: Changes to Dtype will prevent field types from dropping.

PR #1260, opened 2023-03-27
Connector Wizard

LucGenetier:
Implement connector GetParameter APIs containing dynamic Intellisence capabilities

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/ConfigExtensions.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/ConfigExtensions.cs
@@ -52,7 +52,7 @@ namespace Microsoft.PowerFx
             return functionInfos;
         }
 
-        public static void AddService(this PowerFxConfig config, string functionNamespace, ConnectorFunction function, HttpMessageInvoker httpClient = null, ICachingHttpClient cache = null)
+        public static void AddService(this PowerFxConfig config, string functionNamespace, ConnectorFunction function, HttpClient httpClient = null, ICachingHttpClient cache = null)
         {
             if (config == null) 
             { 
@@ -68,8 +68,8 @@ namespace Microsoft.PowerFx
             { 
                 throw new ArgumentNullException(nameof(function)); 
             }
-
-            config.AddFunction(function.GetServiceFunction());
+            
+            config.AddFunction(function.GetServiceFunction(functionNamespace, httpClient, cache));
         }
 
         public static void Add(this Dictionary<string, FormulaValue> map, string fieldName, FormulaValue value)
--- a/src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
@@ -23,6 +23,8 @@ namespace Microsoft.PowerFx.Connectors
     {
         public string Name { get; }
 
+        public string Namespace { get; private set; }
+
         public string OriginalName => Operation.OperationId;
 
         public string Description => Operation.Description ?? $"Invoke {Name}";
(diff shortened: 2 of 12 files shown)


src/libraries/Microsoft.PowerFx.Connectors/ConfigExtensions.cs
```diff
@@ -52,7 +52,7 @@ public static IReadOnlyList<FunctionInfo> AddService(this PowerFxConfig config,
             return functionInfos;
         }
 
-        public static void AddService(this PowerFxConfig config, string functionNamespace, ConnectorFunction function, HttpMessageInvoker httpClient = null, ICachingHttpClient cache = null)
+        public static void AddService(this PowerFxConfig config, string functionNamespace, ConnectorFunction function, HttpClient httpClient = null, ICachingHttpClient cache = null)
```
MikeStall, line 55: >HttpClient [](http://example.com/codeflow?start=119&length=10)

why this change?

src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
```diff
@@ -39,58 +41,119 @@ public class ConnectorFunction
...
+            {
+                return null;
+            }
+
+            string expr = $@"{@namespace}.{Name}({string.Join(", ", parameters.Parameters.Take(Math.Min(parameters.Parameters.Length, ArityMax - 1)).Select(p => p.Value.ToExpression()))}";
+
+            if (parameters.Parameters.Length > ArityMax - 1)
+            {
+                ConnectorParameterWithSuggestions lastParam = parameters.Parameters.Last();
+                expr += $@", {{ {string.Join(", ", lastParam.ParameterNames.Zip(lastParam.Values, (name, fv) => $"{name}: {fv.ToExpression()}"))} }}";
```
MikeStall, line 116: >fv.ToExpression() [](http://example.com/codeflow?start=123&length=17)

we should use StringBuilder here, and then we can pass it into fv.ToExpression as well.

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
```diff
@@ -287,17 +287,17 @@ public override bool CheckForDynamicReturnType(TexlBinding binding, TexlNode[] a
         }
 #endif
 
-        public override async Task<List<string>> GetConnectorSuggestionsAsync(CallNode callNode, int argPosition, CancellationToken cts)
+        public override async Task<List<RecordValue>> GetConnectorSuggestionsAsync(FormulaValue[] knownParameters, int argPosition, CancellationToken cts)
```
MikeStall, line 290: >RecordValue [](http://example.com/codeflow?start=40&length=11)

RecordValue doesn't seem like right return type. 
Ie, these are not values that will flow into evaluator.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -214,7 +215,7 @@ internal abstract class TexlFunction : IFunction
 
         // This method is used for managing "x-ms-dynamic-values" and "x-ms-dynamic-schema" OpenApi extensions in connectors
         // https://learn.microsoft.com/en-us/connectors/custom-connectors/openapi-extensions#use-dynamic-values
-        public virtual async Task<List<string>> GetConnectorSuggestionsAsync(CallNode callNode, int argPosition, CancellationToken cts)
+        public virtual async Task<List<RecordValue>> GetConnectorSuggestionsAsync(FormulaValue[] knownParameters, int argPosition, CancellationToken cts)
```
MikeStall, line 218: >RecordValue [](http://example.com/codeflow?start=39&length=11)

RecordValue doesn't seem like right return type. 
Ie, these are not values that will flow into evaluator.

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseHelper.cs
```diff
@@ -356,14 +357,66 @@ public static void AddTopLevelSuggestionsForCursorType(IntellisenseData.Intellis
...
+                List<RecordValue> suggestions = info.Function.GetConnectorSuggestionsAsync(parameters, argPosition, CancellationToken.None).ConfigureAwait(false).GetAwaiter().GetResult();
 
                 if (suggestions != null)
-                {
-                    foreach (string suggestion in suggestions)
+                {                    
+                    foreach (RecordValue suggestion in suggestions)
                     {
-                        AddSuggestion(intellisenseData, suggestion, SuggestionKind.Global, SuggestionIconKind.Other, DType.String, false);
+                        if (argPosition < info.Function.MaxArity - 1)
```
MikeStall, line 374: >if (argPosition < info.Function.MaxArity - 1) [](http://example.com/codeflow?start=24&length=45)

What does this check mean? 
If we wrap the return result in a class, can it abstract this check as well?

PR #1261, opened 2023-03-27
Update First/LastN functions to require second argument

CarlosFigueiraMSFT:
It doesn't make sense to request the first/last "N" elements of a table without specifying that "N" is. Also, the default value for those functions is currently set at 1, which is inconsistent with all other defaults. We are going to update the function to require both arguments to be passed (but this change will be done under a feature to maintain back compat with Power Apps.


LucGenetier:
✅ No public API change.

MikeStall (approved):

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):

lesaltzm (approved):

PR #1271, opened 2023-03-29
AggregateCoercionNode  bug.

MikeStall:
We only add fields to AggregateCoercionNode that require a coercion. But eval only evals the fields listed in AggregateCoercionNode. So a field that doesn't require coercion gets skipped in eval!

This fixes it via eval - but should probalby be fixed at IR level. Since eval processes every time, whereas we know this statically.

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ArgCoercion.txt
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ArgCoercion.txt
@@ -41,3 +41,15 @@ Error({Kind:ErrorKind.InvalidArgument})
 
 >> Hour(If(false, "hi", Blank()))
 0
+
+// Ensure record coercion doesn't prevent us from accessing non-coerced fields. 
+>> Last([{a:1,b:2}, {a:"1", b: 3}]).b
+3
+
+// extra nesting, a.x coerces. 
+>> Last([{a: { x:1},b:2}, {a:{x:"1"}, b: 3}]).b
+3
+
+// Coerce to 1st element in list; so middle element shouldn't impact  
+>> Last([{a:1,b:2}, {a:2, b: "2"}, {a:"1", b: 3}]).b
+3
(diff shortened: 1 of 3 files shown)


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ArgCoercion.txt
```diff
@@ -40,4 +40,16 @@ Error({Kind:ErrorKind.InvalidArgument})
...
+// Ensure record coercion doesn't prevent us from accessing non-coerced fields. 
+>> Last([{a:1,b:2}, {a:"1", b: 3}]).b
+3
+
+// extra nesting, a.x coerces. 
+>> Last([{a: { x:1},b:2}, {a:{x:"1"}, b: 3}]).b
+3
+
+// Coerce to 1st element in list; so middle element shouldn't impact  
+>> Last([{a:1,b:2}, {a:2, b: "2"}, {a:"1", b: 3}]).b
```
Shpakh-MSFT, line 54: Last([ {a:2}, {a:true, b:true, c:20}, {a:2, b:2, c:true} ]).c

This should return true.

Shpakh-MSFT (approved):
:shipit:

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ArgCoercion.txt
anderson-joyle: The first record `{a:1,b:2}` "initializes" field `b` as number.

MikeStall: added

LucGenetier:
✅ No public API change.

anderson-joyle (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1280, opened 2023-03-30
Better error message for "unknown" functions

LucGenetier:
When a function is valid in Power Apps products, we don't want to report "unknown or unsupported function".
If those known functions are not implemented in other products we'll report "recognized but not yet supported function".


LucGenetier:
✅ No public API change.

MikeStall (dismissed):

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1281, opened 2023-03-30
Update Accepts/CoercesTo to take a flag to control legacy behavior and breaks accepts relationships with the flag enabled

CarlosFigueiraMSFT:
Many primitive types have "Accepts" relationships as a byproduct of the JavaScript legacy coming from Power Apps: a string "accepts" a Guid because they are both implemented internally as strings; a number "accepts" a date/time value because internally both are numbers. But this doesn't work in other hosts where we don't have this "duck typing" semantics. This change removes the Accepts relationships between primitive types, moving them to coercions, making such conversions explicit.


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1936,31 +1940,60 @@ public virtual bool Accepts(DType type, out KeyValuePair<string, DType> schemaDi
...
-                            type.Kind == DKind.Time ||
-                            type.Kind == DKind.DateTimeNoTimeZone)) ||
-                        (type.Kind == DKind.Enum && Accepts(type.GetEnumSupertype()));
+                    if (usePowerFxV1CompatibilityRules)
+                    {
+                        accepts =
+                            type.Kind == Kind ||
+                            type.Kind == DKind.Unknown ||
+                            type.Kind == DKind.Deferred ||
+                            (useLegacyDateTimeAccepts &&
```
Shpakh-MSFT, line 1975: >useLegacyDateTimeAccepts [](http://example.com/codeflow?start=29&length=24)

Even with V1 compat, are we going to have this legacy mode?

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -3183,8 +3316,9 @@ public virtual bool CoercesTo(DType typeDest, out bool isSafe, out DType coercio
                 case DKind.Boolean:
                     isSafe = Kind != DKind.String;
                     doesCoerce = Kind == DKind.String ||
-                                 Number.Accepts(this) ||
-                                 Decimal.Accepts(this) ||
+                                 (usePowerFxV1CompatibilityRules && Currency.Accepts(this, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: true)) ||
```
Shpakh-MSFT, line 3319: >Currency.Accepts(this, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: true) [](http://example.com/codeflow?start=68&length=106)

Why are we adding this?

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -3196,56 +3330,142 @@ public virtual bool CoercesTo(DType typeDest, out bool isSafe, out DType coercio
...
                     isSafe = Kind != DKind.String;
                     doesCoerce = Kind == DKind.String ||
-                                 Number.Accepts(this) ||
-                                 Decimal.Accepts(this) ||
-                                 DateTime.Accepts(this);
+                                 Number.Accepts(this, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: usePowerFxV1CompatibilityRules) ||
+                                 Decimal.Accepts(this, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: usePowerFxV1CompatibilityRules) ||
+                                 DateTime.Accepts(this, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: usePowerFxV1CompatibilityRules) ||
+                                 Time.Accepts(this, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: usePowerFxV1CompatibilityRules) ||
+                                 Date.Accepts(this, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: usePowerFxV1CompatibilityRules);
```
Shpakh-MSFT, line 3337: Are there just gaps in existing implementation?

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -3196,56 +3330,142 @@ public virtual bool CoercesTo(DType typeDest, out bool isSafe, out DType coercio
...
+                    break;
+                case DKind.Guid:
+                    if (usePowerFxV1CompatibilityRules)
+                    {
+                        doesCoerce = String.Accepts(this, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: true);
+                    }
+                    else
+                    {
+                        doesCoerce = Kind == DKind.Guid;
+                    }
```
Shpakh-MSFT, line 3383: How did it work for Guid type previously? doesCoerce was always false for DKind.Guid.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -3196,56 +3330,142 @@ public virtual bool CoercesTo(DType typeDest, out bool isSafe, out DType coercio
...
                     doesCoerce = Kind != DKind.Color && Kind != DKind.Control && Kind != DKind.DataEntity && Kind != DKind.OptionSet && Kind != DKind.View && Kind != DKind.Polymorphic && Kind != DKind.File && Kind != DKind.LargeImage;
+                    break;
+                case DKind.Guid:
+                    if (usePowerFxV1CompatibilityRules)
+                    {
+                        doesCoerce = String.Accepts(this, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: true);
+                    }
+                    else
+                    {
+                        doesCoerce = Kind == DKind.Guid;
```
Shpakh-MSFT, line 3382: >Kind == DKind.Guid; [](http://example.com/codeflow?start=37&length=19)

For destType.Kind = Guid, we shouldn't need to do coercion check.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -3196,56 +3330,142 @@ public virtual bool CoercesTo(DType typeDest, out bool isSafe, out DType coercio
...
+                                 Date.Accepts(this, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: usePowerFxV1CompatibilityRules) ||
+                                 Time.Accepts(this, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: usePowerFxV1CompatibilityRules);
                     break;
                 case DKind.String:
                     doesCoerce = Kind != DKind.Color && Kind != DKind.Control && Kind != DKind.DataEntity && Kind != DKind.OptionSet && Kind != DKind.View && Kind != DKind.Polymorphic && Kind != DKind.File && Kind != DKind.LargeImage;
+                    break;
+                case DKind.Guid:
+                    if (usePowerFxV1CompatibilityRules)
+                    {
+                        doesCoerce = String.Accepts(this, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: true);
```
Shpakh-MSFT, line 3378: >String.Accepts(this, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: true); [](http://example.com/codeflow?start=37&length=105)

Is string to GUID coercion supported?

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1854,10 +1856,12 @@ public bool Accepts(DType type, bool exact = true, bool useLegacyDateTimeAccepts
         /// should affect acceptance.
         /// </param>
         /// <param name="useLegacyDateTimeAccepts"></param>
```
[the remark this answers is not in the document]
CarlosFigueiraMSFT: Added

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
CarlosFigueiraMSFT: Currency was accepted by number, so it could be converted to boolean. Now we don't have a currency/number accepts relationship anymore, so we need to add a new coercion node between the two types

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
CarlosFigueiraMSFT: Before DateTime would accept Time and Date, so those would be covered by that. Now they aren't accepted anymore, so we need to explicitly add allow coercion here.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
CarlosFigueiraMSFT: Yes, we need to have a coercion relationship now. Before it was an accepts relationship, now we need to add a coercion node.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
CarlosFigueiraMSFT: It worked (in Power Apps, not in C#) because there was an Accepts relationship between string and Guid.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
CarlosFigueiraMSFT: Good point, removed.

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Intellisense.cs
```diff
@@ -204,10 +204,10 @@ protected static void TypeMatchPriority(DType type, IList<IntellisenseSuggestion
                 if (!suggestion.Type.IsUnknown && 
 
                     // Most type acceptance is straightforward
-                    (type.Accepts(suggestion.Type) ||
+                    (type.Accepts(suggestion.Type, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: true) ||
```
Shpakh-MSFT, line 207: >true [](http://example.com/codeflow?start=129&length=4)

Is this always going to be true? Aren't we using any config to drive this value?

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Intellisense.cs
```diff
@@ -204,10 +204,10 @@ protected static void TypeMatchPriority(DType type, IList<IntellisenseSuggestion
...
 
                     // Option Set expected types should also include the option set base as a reccomendation.
-                    (suggestion.Type.IsOptionSet && type.Accepts(DType.CreateOptionSetValueType(suggestion.Type.OptionSetInfo)))))
+                    (suggestion.Type.IsOptionSet && type.Accepts(DType.CreateOptionSetValueType(suggestion.Type.OptionSetInfo), exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: true))))
```
Shpakh-MSFT, line 210: >exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: true [](http://example.com/codeflow?start=128&length=82)

I understand the reason to make it explicit, to ensure there is no strange ordering issue, but most of the time they are going to be the same values for these three params. Is there a way to simplify this?

exact should have a default value as true.
useLegacyDateTimeAccepts is only applicable to certain types, so it's strange this to be a required parameter.

Not part of this change but after this change, may be we should try to simplify this.

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Intellisense.cs
CarlosFigueiraMSFT: Since this is intellisense (only for new or modified apps) we should be using the new values. But I'll pass the existing parameter along to minimize the number of changes for now.

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Intellisense.cs
CarlosFigueiraMSFT: Makes sense, we should clean it up later.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1936,31 +1940,55 @@ public virtual bool Accepts(DType type, out KeyValuePair<string, DType> schemaDi
...
-                        type.Kind == DKind.Deferred ||
-                        type.Kind == DKind.Guid ||
-                        (type.Kind == DKind.Enum && Accepts(type.GetEnumSupertype()));
+                    if (usePowerFxV1CompatibilityRules)
+                    {
+                        accepts =
+                            type.Kind == Kind ||
+                            type.Kind == DKind.Unknown ||
+                            type.Kind == DKind.Deferred ||
+                            (type.Kind == DKind.Enum && Accepts(type.GetEnumSupertype(), exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: true));
```
Shpakh-MSFT, line 1949: >true [](http://example.com/codeflow?start=167&length=4)

usePowerFxV1CompatibilityRules

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1936,31 +1940,55 @@ public virtual bool Accepts(DType type, out KeyValuePair<string, DType> schemaDi
...
+                            type.Kind == Kind ||
+                            type.Kind == DKind.Hyperlink ||
+                            type.Kind == DKind.Image ||
+                            type.Kind == DKind.PenImage ||
+                            type.Kind == DKind.Media ||
+                            type.Kind == DKind.Blob ||
+                            type.Kind == DKind.Unknown ||
+                            type.Kind == DKind.Deferred ||
+                            type.Kind == DKind.Guid ||
+                            (type.Kind == DKind.Enum && Accepts(type.GetEnumSupertype(), exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: false));
```
Shpakh-MSFT, line 1963: >false [](http://example.com/codeflow?start=167&length=5)

usePowerFxV1CompatibilityRules

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/ColorFadeT.cs
```diff
@@ -111,7 +111,7 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
             return fValid;
         }
 
-        private bool CheckOtherType(DType otherType, TexlNode otherArg, DType expectedType, IErrorContainer errors, ref Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
+        private bool CheckOtherType(CheckTypesContext context, DType otherType, TexlNode otherArg, DType expectedType, IErrorContainer errors, ref Dictionary<TexlNode, DType> nodeToCoercedTypeMap, Features features)
```
Shpakh-MSFT, line 114: >Features features [](http://example.com/codeflow?start=197&length=17)

This param is not needed if you are already passing context to it.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Trunc.cs
```diff
@@ -79,7 +79,7 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
                 if (type0.IsTable)
                 {
                     // Ensure we have a one-column table of numerics
-                    fValid &= CheckNumericColumnType(type0, args[0], errors, ref nodeToCoercedTypeMap);
+                    fValid &= CheckNumericColumnType(type0, args[0], context.Features, errors, ref nodeToCoercedTypeMap);
```
Shpakh-MSFT, line 82: With Greg's PR,  you are going to run into a conflict for signature of bunch of Check*Type functions.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/BasicCoercion.txt
```diff
@@ -49,4 +49,28 @@ true
...
+Time(0,0,0,0)
+
+>> If(1<0, Time(1,2,3), DateTime(2000,1,1,12,34,56,789))
+Time(12,34,56,789)
+
+>> If(1<0, Date(2000,1,1), Time(12,0,0))
+DateTime(1899,12,30,12,0,0,0)
+
+>> If(1<0, Date(2000,1,1), DateTime(2134,5,6,12,0,0))
+DateTime(2134,5,6,12,0,0,0)
```
Shpakh-MSFT, line 70: Is this correct? I thought this would be Date.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/ColorFadeT.cs
CarlosFigueiraMSFT: Good point, removed

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
CarlosFigueiraMSFT: This is inside `if(usePowerFxV1CompatibilityRules)` so I know that the value is true... But I can change to use the variable to be consistent with other places.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Trunc.cs
CarlosFigueiraMSFT: Yep, I've been resolving lots of conflicts with this PR :-(

Shpakh-MSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

Shpakh-MSFT (dismissed):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

Shpakh-MSFT (dismissed):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

Shpakh-MSFT (dismissed):
:shipit:

LucGenetier:
✅ No public API change.

Shpakh-MSFT (approved):
:shipit:

Shpakh-MSFT (changes requested):
:clock1:

Shpakh-MSFT (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -79,7 +79,7 @@ public void CheckTypesOnDeclaration(CheckTypesContext context, DType actualBodyR
 
             var returnTypeFormulaType = _returnTypeToken.GetFormulaType()._type;
 
-            if (!returnTypeFormulaType.Kind.Equals(actualBodyReturnType.Kind) || !returnTypeFormulaType.CoercesTo(returnTypeFormulaType))
+            if (!returnTypeFormulaType.Kind.Equals(actualBodyReturnType.Kind) || !returnTypeFormulaType.CoercesTo(returnTypeFormulaType, true, true, context.Features.PowerFxV1CompatibilityRules))
```
Shpakh-MSFT, line 82: >true [](http://example.com/codeflow?start=143&length=4)

This probably should be false as default value for isTopLevelCoercion is false. Also use named param syntaxso it's easy to read.

Shpakh-MSFT (approved):
:shipit:

PR #1284, opened 2023-03-30
Add function to check if call node has an async argument

jelopezf:
We already have a function that checks for async arguments with no side effects, however, since we intend to allow for impure node delegation, we should have a function that simply checks for async arguments regardless of node purity.


Shpakh-MSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1293, opened 2023-03-31
Updates to Decimal support

gregli-msft:
Added more tests, updates to testing infrastructure, minor bug fixes and tweaks, and change to Date-Date and Time-Time to return Decimal when NumberIsFloat is disabled.  Rules for when a decimal vs number is returned, when involving a decimal, have been simplified.  Over half of the file changes are adding tests.

I later discovered that there is a breaking change here for NumberIsFloat.  In main currently, Today()*Today() and other math operations involving dat/time is a compile time error, a regression.  Only for NIF.  That has been corrected with this PR.

Only for the interpreter.  When doing Blank()-Today(), the error return used to be `Invalid Argument - Runtime type mismatch` which is really a missed internal check.  With this change, it now properly reports `NotSupported - The operator is invalid for these types`.  We had missed a place where we should have been replacing blank values.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryOperators.cs
```diff
@@ -809,6 +839,26 @@ private static FormulaValue TimeDifference(IRContext irContext, FormulaValue[] a
 
             var result = arg0.Value.Subtract(arg1.Value);
             return new NumberValue(irContext, result.TotalDays);
+        }
+
+        private static FormulaValue DateDifferenceDecimal(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
```
CarlosFigueiraMSFT, line 844: >DateDifferenceDecimal [](http://example.com/codeflow?start=36&length=21)

You can merge those two new methods with the ones above, and check the irContext's return type (irContext.ResultType._type.Kind) to determine whether to create a NumberValue or a DecimalValue

src/libraries/Microsoft.PowerFx.Core/IR/Nodes/BinaryOpKind.cs
```diff
@@ -23,7 +23,9 @@ internal enum BinaryOpKind
         AddDecimals,
 
         DateDifference,
-        TimeDifference,
+        TimeDifference,
+        DateDifferenceDecimal,
+        TimeDifferenceDecimal,
```
CarlosFigueiraMSFT, line 28: We shouldn't need this - depending on the flags the result will always be the same type

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
```diff
@@ -47,6 +47,12 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
   </ItemGroup>
+
+  <ItemGroup>
+    <None Remove="ExpressionTestCases\OpMatrix_Decimal.txt" />
+    <None Remove="ExpressionTestCases\OpMatrix_Eq_Decimal.txt" />
+    <None Remove="ExpressionTestCases\OpMatrix_Float.txt" />
+  </ItemGroup>
```
CarlosFigueiraMSFT, line 55: these can be removed

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryOperators.cs
gregli-msft: Done.

src/libraries/Microsoft.PowerFx.Core/IR/Nodes/BinaryOpKind.cs
gregli-msft: Done

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
gregli-msft: Done

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/FloatLarge.txt
```diff
@@ -117,6 +117,4 @@ Table({Value:0},{Value:1.5708},{Value:-1.5708},{Value:0.7854},{Value:-0.7854},{V
 >> Round(Acot([0, 1e100, -1e100, 1, -1, Blank()]), 4)
 Table({Value:1.5708},{Value:0},{Value:3.1416},{Value:0.7854},{Value:2.3562},{Value:1.5708})
 
->> Power(Exp(42.5),2)
-8223012714622878000000000000000000000
```
LucGenetier, line 120: Why removing this test?

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -336,10 +336,29 @@ protected override async Task<RunResult> RunAsyncInternal(string expr, string se
...
+                    // This is for tests that return 1e100 as a result, verifying proper floating point operation
+                    if (!NumberIsFloat && e.Message.Contains("value is too large"))
+                    {
+                        // Serialization test. Serialized expression must produce an identical result.
+                        ParserOptions options = new ParserOptions() { NumberIsFloat = true };
+                        newValueDeserialized = await engine.EvalAsync(newValue.ToExpression(), CancellationToken.None, options, runtimeConfig: runtimeConfig);
+                    }
+                    else
+                    {
+                        throw e;
```
LucGenetier, line 359: >e [](http://example.com/codeflow?start=30&length=1)

just "throw;"
adding "e" modifies the callstack - not sure if intended

gregli-msft: Done.

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -336,10 +336,29 @@ protected override async Task<RunResult> RunAsyncInternal(string expr, string se
...
-                // Serialization test. Serialized expression must produce an identical result.
-                ParserOptions options = new ParserOptions() { NumberIsFloat = NumberIsFloat };
-                var newValueDeserialized = await engine.EvalAsync(newValue.ToExpression(), CancellationToken.None, options, runtimeConfig: runtimeConfig);
+                FormulaValue newValueDeserialized;
+
+                try
+                {
+                    // Serialization test. Serialized expression must produce an identical result.
+                    ParserOptions options = new ParserOptions() { NumberIsFloat = NumberIsFloat };
+                    newValueDeserialized = await engine.EvalAsync(newValue.ToExpression(), CancellationToken.None, options, runtimeConfig: runtimeConfig);
```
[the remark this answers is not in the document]
gregli-msft: Done.  Although I'm not exactly sure what that does, I trust you.  :)  I just added it in this file.

gregli-msft:
@LucGenetier , tests appear in Text Explorer with InterpreterTestCase_NumberIsFloat so it is pretty easy to see.  It's important to have it at that level as most test files are run through both with and without NIF.  NIF specific .txt files are also postfixed with _Decimal or _Float to indicate that they are special.

LucGenetier (approved):

PR #1302, opened 2023-04-03
Move error test for recognized but unsupported function to interpreter list

CarlosFigueiraMSFT:
The txt-based tests are run in both the Power Fx repository and in Power Apps for compatibility checks. A new test was recently added that checks for an error in the C# interpreter for a function that doesn't exist in that interpreter, but is defined in Power Apps - so it will fail on that host. This change moves this test to the interpreter-only list.


LucGenetier:
✅ No public API change.

LucGenetier (approved):

LucGenetier:
✅ No public API change.

PR #1303, opened 2023-04-03
Ensure we can get a DPath when left hand side of dotted name node is a call node

jelopezf:
Recent changes in IsValidDottedNameNode causes some calls (i.e. LookUp(CDS1, Name = LookUp(CDS, Name = Text).Value)) to fail when the AsyncDelegation and ImpureDelegation features are on because it fails when calling CharacterUtils.MakeSafeForFormatString (DelegationValidationStrategy.cs:213) due to the columnPath being null because TryConvertNodeToDPath just returns the root DPath when left node is a call node.


LucGenetier:
✅ No public API change.

LucGenetier (dismissed):

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -67,6 +67,17 @@ internal static bool TryConvertNodeToDPath(TexlBinding binding, DottedNameNode n
...
                 return true;
             }
+            else if (node.Left is CallNode call)
+            {
+                var rightNodeName = node.Right.Name;
+                if (binding.TryGetReplacedIdentName(node.Right, out var rename))
+                {
+                    rightNodeName = new DName(rename);
+                }
+
```
Shpakh-MSFT, line 77: You can use function name here to add to DPath as we discussed offline.

jelopezf: done

Shpakh-MSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #1306, opened 2023-04-04
Enable CA2007 rule

LucGenetier:
Enable .Net Analyzer CA2007 rule: Do not directly await a Task
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2007

When an asynchronous method awaits a [Task](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task) directly, continuation usually occurs in the same thread that created the task, depending on the async context. This behavior can be costly in terms of performance and can result in a deadlock on the UI thread. Consider calling [Task.ConfigureAwait(Boolean)](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.configureawait#system-threading-tasks-task-configureawait(system-boolean)) to signal your intention for continuation.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #1282, opened 2023-03-30
Table function now supports coercion

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/1204

PR #1309, opened 2023-04-04
Table function now supports coercion

anderson-joyle:
Issue #1204.
Re-pushing due to merge/rebase conflicts. https://github.com/microsoft/Power-Fx/pull/1282

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ArgCoercion.txt
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ArgCoercion.txt
@@ -55,4 +55,4 @@ Error({Kind:ErrorKind.InvalidArgument})
 3
 
 >> Last([ {a:2}, {a:true, b:true, c:20}, {a:2, b:2, c:true} ]).c
-true
+1
--- a/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ArgCoercion.txt
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ArgCoercion.txt
@@ -55,4 +55,4 @@ Error({Kind:ErrorKind.InvalidArgument})
 3
 
 >> Last([ {a:2}, {a:true, b:true, c:20}, {a:2, b:2, c:true} ]).c
-1
+true
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -3254,7 +3254,7 @@ namespace Microsoft.PowerFx.Core.Binding
                         SetDottedNameError(node, TexlStrings.ErrInvalidIdentifier);
                         return;
                     }
-                    
+
                     // Validate that the name exists in the enum type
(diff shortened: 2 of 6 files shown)


LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT:
>                     else if (DType.TryUnionWithCoerce(exprType, childType, out var returnType, out var needCoercion))

Should this change be gated based on the new PFxV1Compat flag? Today this is valid: `AddColumns( [{a:1,b:2},{a:true,b:"2",c:3}], "c", b+100 )`; if the type of the table node in the expression starts including "c", then this expression won't be valid anymore.

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs:5229 in 54e5030. [](commit_id = 54e50307ae799439b6189995c0b386dab2e47701, deletion_comment = False)

CarlosFigueiraMSFT:
>         /// <param name="coercionNeeded">True if union with coercion was called.</param>

successful

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Types/DType.cs:3603 in 54e5030. [](commit_id = 54e50307ae799439b6189995c0b386dab2e47701, deletion_comment = False)

CarlosFigueiraMSFT:
>         public static bool TryUnionWithCoerce(DType argType1, DType argType2, out DType returnType, out bool coercionNeeded)

Does this need to be public?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Types/DType.cs:3605 in 54e5030. [](commit_id = 54e50307ae799439b6189995c0b386dab2e47701, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ArgCoercion.txt
```diff
@@ -55,4 +55,4 @@ Error({Kind:ErrorKind.InvalidArgument})
 3
 
 >> Last([ {a:2}, {a:true, b:true, c:20}, {a:2, b:2, c:true} ]).c
```
CarlosFigueiraMSFT, line 57: >Last([ {a:2}, {a:true, b:true, c:20}, {a:2, b:2, c:true} ]).c [](http://example.com/codeflow?start=3&length=61)

This is a breaking change, so this should only be done with the new PowerFxV1CompatibilityRules feature enabled.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

anderson-joyle:
>         public static bool TryUnionWithCoerce(DType argType1, DType argType2, out DType returnType, out bool coercionNeeded)

Yes. This is called by Table and Binder.

---
In reply to: [1496350985](https://github.com/microsoft/Power-Fx/pull/1309#issuecomment-1496350985) [](http://example.com/codeflow?ancestors=1496350985,1496350985)

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Types/DType.cs:3605 in 7233ea1. [](commit_id = 7233ea1c37dec7ab6138ca573d9b7d1b38a8a30a, deletion_comment = False)

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1310, opened 2023-04-04
Add Visibility to ConnectorFunction

LucGenetier:


LucGenetier:
✅ No public API change.

Issue #1252, opened 2023-03-24
Table versions of Round* and Trunc return the wrong type when input table is coerced


gregli-msft:
```
> IR(Trunc([1,2,3]))
Trunc:*[Value:n](Table:*[Value:n]({Value: 1:n}, {Value: 2:n}, {Value: 3:n}))

> IR(Trunc(["1","2","3"]))
Trunc:*[Value:n](AggregateCoercion(TableToTable, Value <- Value:n(ScopeAccess(Scope 1, Value))))

> Option( Options.ConsistentOneColumnTableResult, false )
false

> IR(Trunc(["1","2","3"]))
Trunc:*[Value:s](AggregateCoercion(TableToTable, Value <- Value:n(ScopeAccess(Scope 1, Value))))

> Trunc(["1","2","3"])
[1, 2, 3]
```

Notice the `*[Value:s]` as the return type from Trunc.  It should be `*[Value:n]`.  That is because the code that establishes the type is in Trunc.cs/CheckTypes copies the input type name for the column, but also copies the type information too:

```
                    returnType = context.Features.HasFlag(Features.ConsistentOneColumnTableResult)
                        ? DType.CreateTable(new TypedName(DType.Number, new DName(ColumnName_ValueStr)))
                        : type0;
```

Notice that it still works, kind of.  The interpreter hits an assert in the NumberValue constructor when it checks the IRContext result type, which causes the test suite to fail:

```
        internal NumberValue(IRContext irContext, double value)
            : base(irContext, value)
        {
            Contract.Assert(IRContext.ResultType == FormulaType.Number);
        }
```

Roughly, the fix is to replace the trunc.cs/round*.cs code with:

```
returnType = context.Features.HasFlag(Features.ConsistentOneColumnTableResult)
   ? DType.CreateTable(new TypedName(DType.Number, new DName(ColumnName_ValueStr)))
   : DType.CreateTable(new TypedName(DType.Number, new DName(type0.GetNames(DPath.Root).First().Name)));
```

We need to make sure that takes into account display names properly.

This fails for Round, RoundUp, RoundDown, and Trunc functions.  

This blocks decimal progress as it also looks like a non-number type that isn't handled correctly.

PR #1311, opened 2023-04-04
Update CheckColumnType to incorporate return type for function result

gregli-msft:
This PR fixes https://github.com/microsoft/Power-Fx/issues/1252 which has a description of the issue.

Turns out most of functions that take a single column table had this bug.  MathFunction.cs had a good solution to the problem, which I had initially propagated to the other functions.  After feedback from the team, I generalized the code to move this solution into the CheckColumnType functions, significantly simplifying the CheckTypes in each of the functions.

This fix is needed for Round* and Trunc in particular, because with decimal single column tables these functions would hit an assert and stop the test suite from proceeding. Something as simple as [1,2,3] would be a problem.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Concatenate.cs
```diff
@@ -132,9 +126,16 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
 
             // Type check the args
             for (var i = 0; i < count; i++)
-            {
-                fArgsValid &= CheckParamIsTypeOrSingleColumnTable(DType.String, args[i], argTypes[i], errors, out var isTable, ref nodeToCoercedTypeMap);
-                hasTableArg |= isTable;
+            {
+                if (argTypes[i].IsTable && argTypes[i] != DType.ObjNull)
```
CarlosFigueiraMSFT, line 130: > argTypes[i] != DType.ObjNull [](http://example.com/codeflow?start=42&length=29)

Should this check be done in the Check<Type>ColumnType (i.e., return true, with no coercions, if arg is DType.ObjNull)?

gregli-msft: As discussed, we are inconsistent on how untyped null tables are handled.  I added a comment and some tests to illustrate the point.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MathFunction.cs
```diff
@@ -67,23 +68,167 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
...
+            yield return new[] { TexlStrings.MathTFuncArg1, TexlStrings.MathTFuncArg2 };
+        }
+    }
+
+    internal abstract class MathTwoArgTableFunction : BuiltinFunction
+    {
+        public override bool IsSelfContained => true;
+
+        // Before ConsistentOneColumnTableResult, this function would always return a fixed name "Result" (Mod)
+        public virtual bool NonConsistentFixedName => false;
```
CarlosFigueiraMSFT, line 113: >NonConsistent [](http://example.com/codeflow?start=28&length=13)

Suggestion: InconsistentTableResultFixedName / InconsistentTableResultUseSecondArg

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -1011,76 +1025,115 @@ protected bool CheckType(TexlNode node, DType nodeType, DType expectedType, IErr
...
-                        errors.EnsureError(DocumentErrorSeverity.Severe, arg, errKey, column.Name.Value);
-                        return false;
-                    }
+                    errors.EnsureError(DocumentErrorSeverity.Severe, arg, errKey, column.Name.Value);
+                    return false;
                 }
             }
-
+
+            if (context != null && context.Features.ConsistentOneColumnTableResult)
```
CarlosFigueiraMSFT, line 1070: >context != null && context.Features.ConsistentOneColumnTableResult [](http://example.com/codeflow?start=16&length=66)

Suggestion: `if (context?.Features.ConsistentOneColumnTableResult == true)`

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -1011,76 +1025,115 @@ protected bool CheckType(TexlNode node, DType nodeType, DType expectedType, IErr
...
+            {
+                // Note that DateTime retains the subtype and does not snap to expectedType
```
CarlosFigueiraMSFT, line 1072: >Note that DateTime retains the subtype and does not snap to expectedType [](http://example.com/codeflow?start=19&length=72)

What's an example of this special case?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Log.cs
```diff
@@ -39,37 +32,19 @@ public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
...
+        }
+
         public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
         {
-            yield return new[] { TexlStrings.MathFuncArg1 };
-        }
-
-        public override string GetUniqueTexlRuntimeName(bool isPrefetching = false)
-        {
-            return GetUniqueTexlRuntimeName(suffix: "_T");
```
CarlosFigueiraMSFT, line 58: >return GetUniqueTexlRuntimeName(suffix: "_T"); [](http://example.com/codeflow?start=12&length=46)

We still need to keep this override. We'll be removing it soon, but there's some work in Power Apps required to do so

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MathFunction.cs
gregli-msft: Done.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
gregli-msft: Done.  I had tried this without the == true on the end and it hadn't worked.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
gregli-msft: Added this comment:
`// For example, DateAdd([Date(2000,1,1)],3) should return *[Value:D] and not *[Value:d]`

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Log.cs
gregli-msft: I put it back in the base class.  Is that sufficient?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -1011,76 +1025,116 @@ protected bool CheckType(TexlNode node, DType nodeType, DType expectedType, IErr
...
+                    errors.EnsureError(DocumentErrorSeverity.Severe, arg, errKey, column.Name.Value);
+                    return false;
                 }
             }
-
+
+            if (context?.Features.ConsistentOneColumnTableResult == true)
+            {
+                // Note that DateTime retains the subtype and does not snap to expectedType.
+                // For example, DateAdd([Date(2000,1,1)],3) should return *[Value:D] and not *[Value:d]
```
Shpakh-MSFT, line 1073: Can this be handled in DateTime specific case instead of here? For example,
CheckDateColumnType

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -1011,76 +1025,116 @@ protected bool CheckType(TexlNode node, DType nodeType, DType expectedType, IErr
...
-            else
+
+            var column = columns.Single();
+            var colType = column.Type;
+
+            if (!expectedType.Accepts(colType))
             {
-                var column = columns.Single();
-                if (!expectedType.Accepts(column.Type))
+                if (SupportsParamCoercion && column.Type.CoercesTo(expectedType))
```
Shpakh-MSFT, line 1058: >column.Type [](http://example.com/codeflow?start=45&length=11)

colType

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -962,6 +964,18 @@ protected bool CheckType(TexlNode node, DType nodeType, DType expectedType, IErr
             var typeChecks = CheckType(node, nodeType, expectedType, errors, coerceIfSupported, out DType coercionType);
             matchedWithCoercion = typeChecks && coercionType != null;
             return typeChecks;
+        }
+
+        protected bool CheckType(TexlNode node, DType nodeType, DType expectedType, IErrorContainer errors, ref Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
+        {
+            var typeChecks = CheckType(node, nodeType, expectedType, errors, true, out DType coercionType);
```
Shpakh-MSFT, line 971: >true [](http://example.com/codeflow?start=77&length=4)

coerceIfSupported: true

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
gregli-msft: It would be difficult to hoist this out.  ColType is only down in here, and it would need to be either an additional out parameter or recomputed.  And then this code to create the table would need to be repeated in all of the Check*ColumnType functions that take a context.  Seemed to me that the best solution with the least code duplication (and opportunity to get out of sync) was to put it here and comment it well.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
gregli-msft: Done.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
gregli-msft: Done.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MathFunction.cs
```diff
@@ -67,23 +68,172 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
...
+            yield return new[] { TexlStrings.MathTFuncArg1, TexlStrings.MathTFuncArg2 };
+        }
+    }
+
+    internal abstract class MathTwoArgTableFunction : BuiltinFunction
+    {
+        public override bool IsSelfContained => true;
+
+        // Before ConsistentOneColumnTableResult, this function would always return a fixed name "Result" (Mod)
+        public virtual bool InConsistentTableResultFixedName => false;
```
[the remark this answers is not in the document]
gregli-msft: Done.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -7,6 +7,7 @@
 using System.Globalization;
 using System.Linq;
 using System.Numerics;
+using System.Runtime.InteropServices.ComTypes;
```
CarlosFigueiraMSFT, line 10: Do we need this?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MathFunction.cs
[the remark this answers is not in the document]
gregli-msft: Done.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
gregli-msft: Done.

LucGenetier:
✅ No public API change.

gregli-msft:
@Shpakh-MSFT , no this does not need to be in pa client immediately.  It is needed by consumers of Decimal.  And I need it as a base for the next PR of decimal functions.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -31,6 +31,7 @@
 using Microsoft.PowerFx.Syntax;
 using Microsoft.PowerFx.Types;
 using static Microsoft.PowerFx.Core.IR.IRTranslator;
+using static Microsoft.PowerFx.Syntax.PrettyPrintVisitor;
 using CallNode = Microsoft.PowerFx.Syntax.CallNode;
```
lesaltzm, line 35: what was this needed for?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -1011,76 +1023,116 @@ protected bool CheckType(TexlNode node, DType nodeType, DType expectedType, IErr
...
-        // accordingly. Returns true if the types align, false otherwise. matchedWithCoercion is set
-        // to true if the types align only with coercion.
+        }
+
+        // Check that the type of a specified node is a number column type, and possibly emit errors
+        // accordingly. Returns true if the types align, false otherwise.
         public bool CheckNumericColumnType(DType type, TexlNode arg, IErrorContainer errors, ref Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
         {
-            return CheckColumnType(type, arg, DType.Number, errors, TexlStrings.ErrInvalidSchemaNeedNumCol_Col, ref nodeToCoercedTypeMap);
+            return CheckColumnType(type, arg, DType.Number, errors, TexlStrings.ErrInvalidSchemaNeedNumCol_Col, ref nodeToCoercedTypeMap, null, out _);
```
bryancroteau-MSFT, line 1082: Why are we passing null into the CheckTypesContext?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -1011,76 +1023,116 @@ protected bool CheckType(TexlNode node, DType nodeType, DType expectedType, IErr
...
+            return CheckColumnType(type, arg, DType.Number, errors, TexlStrings.ErrInvalidSchemaNeedNumCol_Col, ref nodeToCoercedTypeMap, context, out returnType);
         }
 
         // Check that the type of a specified node is a color column type, and possibly emit errors
-        // accordingly. Returns true if the types align, false otherwise.
+        // accordingly. Returns true if the types align, false otherwise.
         protected bool CheckColorColumnType(DType type, TexlNode arg, IErrorContainer errors, ref Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
         {
-            return CheckColumnType(type, arg, DType.Color, errors, TexlStrings.ErrInvalidSchemaNeedColorCol_Col, ref nodeToCoercedTypeMap);
+            return CheckColumnType(type, arg, DType.Color, errors, TexlStrings.ErrInvalidSchemaNeedColorCol_Col, ref nodeToCoercedTypeMap, null, out _);
```
bryancroteau-MSFT, line 1095: Same question as above

bryancroteau-MSFT (changes requested):

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
gregli-msft: Because I'm old school.  :)  I've had enough questions on this today so I refactored it to make it clearer.  No longer a null check.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
gregli-msft: Because I'm old school.  :)  I've had enough questions on this today so I refactored it to make it clearer.  No longer a null check.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MathFunction.cs
```diff
@@ -67,23 +68,172 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
...
+    {
+        public override bool IsSelfContained => true;
+
+        // Before ConsistentOneColumnTableResult, this function would always return a fixed name "Result" (Mod)
+        protected virtual bool InConsistentTableResultFixedName => false;
+
+        // Before ConsistentOneColumnTableResult, this function would use the second argument name if a table (Log, Power)
+        protected virtual bool InConsistentTableResultUseSecondArg => false;
+
+        public MathTwoArgTableFunction(string name, TexlStrings.StringGetter description, int minArity)
```
bryancroteau-MSFT, line 118: nit: Can you use the explicit parameter name syntax: minArity: 1 etc, in all the places where this constructor is called?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MathFunction.cs
```diff
@@ -67,23 +68,172 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
...
+    {
+        public override ArgPreprocessor GetArgPreprocessor(int index)
+        {
+            return base.GetGenericArgPreprocessor(index);
+        }
 
-            if (!fValid)
+        public override bool IsSelfContained => true;
+
+        public MathTwoArgFunction(string name, TexlStrings.StringGetter description, int minArity)
```
bryancroteau-MSFT, line 92: Same here, explicit syntax where this is being called, minArity: 1 etc

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MathFunction.cs
gregli-msft: Done.  Good call.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MathFunction.cs
gregli-msft: Done.

LucGenetier:
✅ No public API change.

bryancroteau-MSFT (approved):

PR #1315, opened 2023-04-05
Initial start on 1314

### LINKED ISSUE #1314: Decimal: FormulaValue.ToExpression should show float vs. decimal
NumberValue should serialize as `Float(123)`, whereas DecimalValue should serialize as `Decimal(123)`

Compact mode can still show them both as just `123`, which will mitigate .txt updates

MikeStall:
Fix #1314. 

This is still failing tests - needs test updates.


MikeStall:
Fixed in #1366

PR #1323, opened 2023-04-07
Fix table nodes IR translation

CarlosFigueiraMSFT:
Currently the `[...]` notation doesn't work properly (it is incorrectly wrapping record nodes in `{Value:X}`) if there are expressions in the nodes. This change fixes it by updating the IR translator to look at the binding type instead of the parser node.


LucGenetier:
✅ No public API change.

anderson-joyle (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1324, opened 2023-04-07
Fix check of FirstN/LastN functions with new feature flag

CarlosFigueiraMSFT:
Check for validity of calls to the FirstN/LastN function with the new feature flag that requires two arguments was broken. This fixes it.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/FirstLastN.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/FirstLastN.cs
@@ -50,7 +50,7 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
 
             var fArgsValid = base.CheckTypes(context, args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
 
-            if (context.Features.FirstLastNRequiresSecondArguments)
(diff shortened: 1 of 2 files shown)


Shpakh-MSFT:
>                   1,

Shouldn't be MinArity be 2 as well if context.Features.FirstLastNRequiresSecondArguments true?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/FirstLastN.cs:30 in 7d403ec. [](commit_id = 7d403ecf5795e4c95ea9ab167ea6fb858fde494a, deletion_comment = False)

CarlosFigueiraMSFT:
>                   1,

We don't have the context at this point

---
In reply to: [1500329853](https://github.com/microsoft/Power-Fx/pull/1324#issuecomment-1500329853) [](http://example.com/codeflow?ancestors=1500329853,1500329853)

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/FirstLastN.cs:30 in 7d403ec. [](commit_id = 7d403ecf5795e4c95ea9ab167ea6fb858fde494a, deletion_comment = False)

Shpakh-MSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1325, opened 2023-04-07
implement async CollectionTableValue

LucGenetier:


src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
```diff
@@ -19,23 +20,51 @@ namespace Microsoft.PowerFx.Types
...
+        private IEnumerable<T> _enumerator; // required. supports enumeration;
+
+        // Async implementation
+        private IAsyncEnumerable<T> _asyncEnumerable;        
 
         // Additional capabilities. 
-        private readonly IReadOnlyList<T> _sourceIndex; // maybe null. supports index. 
-        private readonly IReadOnlyCollection<T> _sourceCount; // maybe null. supports count;
-        private readonly ICollection<T> _sourceList; // maybe null. supports mutation;
+        private IReadOnlyList<T> _sourceIndex; // maybe null. supports index.
```
MikeStall, line 29: >IReadOnlyList [](http://example.com/codeflow?start=16&length=13)

Can we keep these readonly. 

I see the Init() method, but there are other patterns (like chaining ctors).

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
```diff
@@ -19,23 +20,51 @@ namespace Microsoft.PowerFx.Types
...
+        private IReadOnlyCollection<T> _sourceCount; // maybe null. supports count;
+        private ICollection<T> _sourceList; // maybe null. supports mutation;
 
         public CollectionTableValue(RecordType recordType, IEnumerable<T> source)
           : this(IRContext.NotInSource(recordType.ToTable()), source)
         {
             RecordType = recordType;
         }
 
+        public CollectionTableValue(RecordType recordType, bool inMemory, IAsyncEnumerable<T> source)
```
MikeStall, line 39: >bool inMemory, [](http://example.com/codeflow?start=59&length=14)

why do we need this? We need to be very clear about the invariants around this new flag.

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
```diff
@@ -60,10 +89,30 @@ public override IEnumerable<DValue<RecordValue>> Rows
...
+                }
+                else
                 {
-                    var record = Marshal(item);
-                    yield return record;
+                    IAsyncEnumerator<T> asyncEnumerator = _asyncEnumerable.GetAsyncEnumerator();
+
+                    try
+                    {                        
+                        while (asyncEnumerator.MoveNextAsync().GetAwaiter().GetResult())
```
MikeStall, line 106: Can we use foreach?

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
```diff
@@ -34,6 +34,8 @@ public abstract class TableValue : ValidFormulaValue
 
         public abstract IEnumerable<DValue<RecordValue>> Rows { get; }
```
MikeStall, line 35: >IEnumerable<DValue<RecordValue>> Rows  [](http://example.com/codeflow?start=24&length=38)

With the change, do we want to add IAsyncEnumerable RowsAsync?

src/libraries/Microsoft.PowerFx.Interpreter/Values/QueryableTableValue.cs
```diff
@@ -1,23 +1,23 @@
...
+using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 using Microsoft.PowerFx.Core.IR;
 
 namespace Microsoft.PowerFx.Types
 {
     public abstract class QueryableTableValue : TableValue
     {
-        private Lazy<Task<List<DValue<RecordValue>>>> _lazyTaskRows;
+        private IAsyncEnumerable<DValue<RecordValue>> _cachedRows;
```
MikeStall, line 13: >_cachedRows [](http://example.com/codeflow?start=54&length=11)

Removing Lazy<List> will remove the cache. 
The enum is not actually caching.

MikeStall: The test would be to verify we only enumerate once even when we call foreach many times: 

foreach(var row in Rows) { }
foreach(var row in Rows) { }
foreach(var row in Rows) { }

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -613,71 +613,53 @@ private static FormulaValue SortOptionSet(List<(DValue<RecordValue> row, Formula
...
-            EvalVisitor runner,
-            EvalVisitorContext context,
-            IEnumerable<DValue<RecordValue>> sources,
-            LambdaFormulaValue filter,
-            int topN = int.MaxValue)
-        {
-            var results = new List<DValue<RecordValue>>();
-
-            // Filter needs to allow running in parallel. 
+        private static async IAsyncEnumerable<DValue<RecordValue>> FilterAsync(EvalVisitor runner, EvalVisitorContext context, IEnumerable<DValue<RecordValue>> sources, LambdaFormulaValue filter, int topN = int.MaxValue)
```
MikeStall, line 649: > IAsyncEnumerable<DValue<RecordValue>>  [](http://example.com/codeflow?start=28&length=39)

This is significant:
before, we return T[], which is mutable, (but not Lazy!)
Now we return IEnum which is lazy, but not mutable. 

Mutable determines Patch, Collect, etc.

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
```diff
@@ -34,6 +34,8 @@ public abstract class TableValue : ValidFormulaValue
...
 
+        public abstract bool InMemory { get; }
```
MikeStall, line 37: >public abstract bool InMemory { get; } [](http://example.com/codeflow?start=8&length=38)

Adding a new property to base class is big - we need to be extremely clear on what exactly these semantics  mean. 
Ie, given an expression with table t, how does it behave differently if t.InMemory=true vs. false.

LucGenetier: After our discussion I think we should consider capabilities via a set of interfaces to determine what's possible for the rows

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
LucGenetier: Except that we can't use different base() calls here, hence this non-optimal logic, I agree

Can be considered for later as this is not the most important [private prop] (in this PR though)

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
LucGenetier: indirectly, to consider capabilities

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
LucGenetier: we cant' use 'await foreach' here as Rows isn't async

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
LucGenetier: Maybe

src/libraries/Microsoft.PowerFx.Interpreter/Values/QueryableTableValue.cs
LucGenetier: I need to add a UT To validate this important point and ensure we call GetRowsAsync() only once in this scenario

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
LucGenetier: This is the main reason for this change
Capabilities (or interfaces) will have to be managed by the caller

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
MikeStall: But can we factor it into a generic helper:

IEnum<T> Sync(IAsyncEnum<T> inner);

We're calling GetResult() either way and doing sync-over-async.

That would quarantine the code. 
And then this method can just use Foreach on the Sync() helper.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -613,71 +613,53 @@ private static FormulaValue SortOptionSet(List<(DValue<RecordValue> row, Formula
             return new InMemoryTableValue(irContext, pairs.Select(pair => pair.row));
         }
 
-        private static async Task<DValue<RecordValue>> LazyFilterRowAsync(
-           EvalVisitor runner,
-           EvalVisitorContext context,
-           DValue<RecordValue> row,
-           LambdaFormulaValue filter)
+        private static async IAsyncEnumerable<DValue<RecordValue>> FilterRowAsync(EvalVisitor runner, EvalVisitorContext context, DValue<RecordValue> row, LambdaFormulaValue filter, [EnumeratorCancellation] CancellationToken cts = default)
```
MikeStall, line 616: before, this returned a single item; now it returns an enumerable.  Intentional?

PR #1330, opened 2023-04-07
Enables AsType function for Dataverse 

jas-valgotar:
Fixes https://github.com/microsoft/Power-Fx-Dataverse/issues/113

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
@@ -12,6 +12,7 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Core.Localization;
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
 
@@ -37,6 +38,29 @@ namespace Microsoft.PowerFx.Types
         public bool IsColumn => IRContext.ResultType._type.IsColumn;
 
         public new TableType Type => (TableType)base.Type;
+
+        /// <summary>
+        /// Casts <paramref name="record"/> to the table's record type.
+        /// </summary>
+        /// <param name="record"> Record to cast.</param>
+        /// <param name="cancellationToken"></param>
+        public virtual DValue<RecordValue> CastRecord(RecordValue record, CancellationToken cancellationToken)
+        {
+            if (record.Type == Type.ToRecord())
+            {
+                return DValue<RecordValue>.Of(record);
+            }
+
+            var error = new ErrorValue(IRContext, new ExpressionError()
+            {
+                MessageKey = TexlStrings.InvalidCast.Key,
+                Span = IRContext.SourceContext,
+                Kind = ErrorKind.InvalidArgument,
+                MessageArgs = new object[] { record.Type, Type.ToRecord() }
+            });
+
+            return DValue<RecordValue>.Of(error);
+        }
 
         public TableValue(RecordType recordType)
             : this(IRContext.NotInSource(recordType.ToTable()))
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AsType.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AsType.cs
@@ -88,7 +88,7 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
 
             // Check if table arg referrs to a connected data source.
             var tableArg = args[1];
-            var ads = argTypes[1].AssociatedDataSources.FirstOrDefault();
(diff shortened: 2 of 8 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
```diff
@@ -38,6 +38,16 @@ public abstract class TableValue : ValidFormulaValue
 
         public new TableType Type => (TableType)base.Type;
 
+        /// <summary>
+        /// Retrieves record from the table by a primary key.
+        /// </summary>
+        /// <param name="key">Primary key that can be retrieved via <see cref="RecordValue.TryGetPrimaryKey(out string)"/>.</param>
+        /// <param name="cancellationToken"></param>
+        public virtual async Task<RecordValue> GetRecordAsync(string key, CancellationToken cancellationToken)
```
MikeStall, line 46: GetRecordByPrimaryKey

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AsType.cs
```diff
@@ -90,8 +90,9 @@ public override void CheckSemantics(TexlBinding binding, TexlNode[] args, DType[
             // Check if table arg referrs to a connected data source.
             var tableArg = args[1];
             if (!binding.TryGetFirstNameInfo(tableArg.Id, out var tableInfo) ||
-                tableInfo.Data is not IExternalDataSource tableDsInfo ||
-                !(tableDsInfo is IExternalTabularDataSource))
+                (tableInfo.Data is not NameSymbol &&
```
MikeStall, line 93: are there additional compile time checks we should do? Or do we punt to runtime?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -529,6 +529,27 @@ public static async ValueTask<FormulaValue> SortTable(EvalVisitor runner, EvalVi
             {
                 return CommonErrors.RuntimeTypeMismatch(irContext);
             }
+        }
+
+        public static async ValueTask<FormulaValue> AsType(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
+        {
+            var arg0 = (RecordValue)args[0];
```
MikeStall, line 536: did pipeline confirm they really are these types, or could they be Blank/Error?

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
MikeStall, line 46: nit to consider:
- Who generates the error message? There's probably a common message we can generate ("record X can't be cast to type Y").  
-  if it is a common message, then this just needs to signal failure (bool TryGetRecord ), and we can fill in the message. 
- else we could return DValue<REcordValue> instead, which would let the caller pass an error message.  Although if we use common one, that's not necessary, 
- Does this need to be async? Can we tell statically whether it will work?


-

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
```diff
@@ -38,6 +38,21 @@ public abstract class TableValue : ValidFormulaValue
...
+        /// <param name="record"> Record to cast.</param>
+        /// <param name="cancellationToken"></param>
+        public virtual RecordValue CastRecord(RecordValue record, CancellationToken cancellationToken)
+        {
+            if (record.Type == Type.ToRecord())
+            {
+                return record;
+            }
+
+            return RecordValue.Empty();
```
MikeStall, line 53: error?

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
jas-valgotar: -Added a common error msg
-Now returns DValue<RecordValue> instead of RecordValue
-it sync operation now.

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
jas-valgotar: updated

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
jas-valgotar: Yes, add a test in DV repo for the same.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

Issue #1331, opened 2023-04-08
Interpreter Round* Float functions disagree with Excel on large numbers and/or large number of digits


gregli-msft:
```
> Option( Options.NumberIsFloat, true )
true

> Round( 1.23456789e200, [-190,-196,-198,-200,-201] )
[1.23456789E+200, 1.23456789E+200, 1.23456789E+200, 1.23456789E+200, 1.23456789E+200]
// Excel returns 1.2346E+200, 1.2346E+200, 1.23E+200, 1E+200, 0

> Round( 1.23456789E-200, [196,198,200,201,202,203] )
[1.23456789E-200, 1.23456789E-200, 1.23456789E-200, 1.23456789E-200, 1.23456789E-200, 1.23456789E-200]
// Excel returns 0, 0, 1E-200, 1.2E-200, 1.23E-200, 1.235E-200
```

These lines are in RoundFloat in LibraryMath.cs and I think it either needs to be eliminated or adjusted.  
```
            if (dg < -15 || dg > 15 || number < -1e20d || number > 1e20d)
            {
                return num;
            }
```

The other option is to document and enforce a limit (throw an exception).  C# Math.Round apparently has a limit at 15 digits, while decimal.Round does not (but is naturally limited to 28 digits).
```
Console.WriteLine(Math.Round(1.2345E-100, 16));
Unhandled exception. System.ArgumentOutOfRangeException: Rounding digits must be between 0 and 15, inclusive. (Parameter 'digits')
   at System.Math.Round(Double value, Int32 digits, MidpointRounding mode)
   at Program.<Main>$(String[] args) in d:\repos\ConsoleApp9\Program.cs:line 53
```

gregli-msft:
After this is fixed, we shuld add more tests with really big and small numbers to RoundingOps.txt.

PR #1333, opened 2023-04-10
Add Decimal function support

gregli-msft:
Currently all the functions operate on and return Float values.  Which means functions like Abs( Decimal ) will return a Float with a possible loss of precision when it should instead return a Decimal.  This PR updates core math functions Sum, Average, Max, Min, Abs, Mod, Round*, Int, and Trunc to return the same type that they are passed.


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -675,49 +851,199 @@ private static IEnumerable<double> LazySequence(double records, double start, do
...
+
+            return CommonErrors.UnreachableCodeError(irContext);
+        }
+
+        // The algorithm for Decimal is different from that of Float because with less range we are going out of our way to avoid overflow
+        // At the time of this writing, the version of .NET being targeted only supports two varieties of MidpointRounding
+        // In the future, some of this can be replaced with built in support in decimal.Round()
+        internal static FormulaValue RoundDecimal(IRContext irContext, DecimalValue dec, int digits, RoundType roundType = RoundType.Default)
+        {
+            // Decimal TODO: This should be moved out as a static member, but tests complain
```
gregli-msft, line 958: @MikeStall , I still need a solution to this problem.  The test suite still complains if I try to make this a static in some manner.

src/libraries/Microsoft.PowerFx.Core/Functions/ArgPreprocessor.cs
```diff
@@ -9,9 +9,16 @@ namespace Microsoft.PowerFx.Core.Functions
...
     {
-        None = 0,
-        ReplaceBlankWithZero = 1,
-        ReplaceBlankWithZeroAndTruncate = 2,
-        ReplaceBlankWithEmptyString = 3,
+        None = 0,
+
+        ReplaceBlankWithFloatZero = 1,
+        ReplaceBlankWithDecimalZero = 2,
+        ReplaceBlankWithFloatZeroAndTruncate = 3,
```
lesaltzm, line 16: does this one need a decimal equivalent?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -32,6 +32,7 @@
 using Microsoft.PowerFx.Syntax;
 using Microsoft.PowerFx.Types;
 using static Microsoft.PowerFx.Core.IR.IRTranslator;
+using static Microsoft.PowerFx.Syntax.PrettyPrintVisitor;
 using CallNode = Microsoft.PowerFx.Syntax.CallNode;
```
lesaltzm, line 36: unused?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -951,6 +952,23 @@ public virtual bool IsMetadataTypeArg(int index)
         public virtual bool HasSuggestionsForParam(int index)
         {
             return false;
+        }
+
+        // Return the data type, either Decimal or Number, that should be the return type for a function
+        // if the provided argument was the first type encountered.
+        protected DType DetermineNumericFunctionReturnType(bool nativeDecimal, bool numberIsFloat, DType argType)
```
lesaltzm, line 959: >protected [](http://example.com/codeflow?start=8&length=9)

static?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -1026,14 +1044,30 @@ protected bool CheckType(TexlNode node, DType nodeType, DType expectedType, IErr
             return false;
         }
 
+        protected bool TryGetSingleColumn(DType type, TexlNode arg, IErrorContainer errors, out TypedName column)
+        {
+            IEnumerable<TypedName> columns;
+
+            column = new TypedName();
```
lesaltzm, line 1051: >new TypedName(); [](http://example.com/codeflow?start=21&length=16)

likely should be null/default, not new TypedName()

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -383,30 +398,131 @@ private List<IntermediateNode> AttachArgPreprocessor(List<IntermediateNode> args
...
+            }
+
+            /// <summary>
+            /// Wraps node arg => Coalesce(arg , 0) when arg is not Number Literal.
+            /// </summary>
+            private static IntermediateNode ReplaceBlankWithDecimalZero(IntermediateNode arg)
+            {
+                IntermediateNode zeroLitNode;
+                IRContext convertedIRContext;
+
```
lesaltzm, line 449: nit: no need to declare these before their assignment a few lines later.

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -383,30 +398,131 @@ private List<IntermediateNode> AttachArgPreprocessor(List<IntermediateNode> args
...
+                    convertedIRContext = new IRContext(arg.IRContext.SourceContext, FormulaType.Decimal);
+                    zeroLitNode = new DecimalLiteralNode(convertedIRContext, 0m);
+                }
+
+                var convertedNode = new CallNode(convertedIRContext, BuiltinFunctionsCore.Coalesce, arg, zeroLitNode);
+                return convertedNode;
+            }
 
             /// <summary>
             /// Wraps node arg => Truc(Coalesce(arg , 0)).
```
lesaltzm, line 521: >Truc [](http://example.com/codeflow?start=34&length=4)

sp: trunc

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MathFunction.cs
```diff
@@ -67,9 +105,15 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
             Contracts.Assert(returnType.IsTable);
 
             var arg = args[0];
-            var argType = argTypes[0];
-            fValid &= CheckNumericColumnType(argType, arg, errors, ref nodeToCoercedTypeMap, context, out returnType);
-
+            var argType = argTypes[0];
+
+            if (argType.IsTable)
```
lesaltzm, line 110: >argType.IsTable [](http://example.com/codeflow?start=16&length=15)

looks like previously, it always did the checkcolumntype operation, even if `!argtype.istable`

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -6,6 +6,7 @@
 using System.Globalization;
 using System.Linq;
 using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.VisualBasic.Syntax;
 using Microsoft.PowerFx.Core.IR;
```
lesaltzm, line 10: ?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -580,7 +727,7 @@ private static IAggregator GetMinMaxAggType(IServiceProvider serviceProvider, IR
         // Average(1,2,3)
         public static FormulaValue Average(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
         {
-            return RunAggregator(new AverageAgg(), irContext, args);
+            return RunAggregator(irContext.ResultType == FormulaType.Decimal ? new AverageDecimalAgg() : new AverageAgg(), irContext, args);
```
lesaltzm, line 730: >irContext.ResultType == FormulaType.Decimal [](http://example.com/codeflow?start=33&length=43)

Design-wise, this kind of seems like a bit of a smell. We should likely have resolved this at IRTranslation time, making decisions this late based on type seems wrong.

+ @MikeStall

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
MikeStall: Fixed, #1341

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
gregli-msft: Thank you!

src/libraries/Microsoft.PowerFx.Core/Functions/ArgPreprocessor.cs
gregli-msft: Added.  It isn't used today, but probably wise to include it so someone has it when they need it.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -951,6 +952,23 @@ public virtual bool IsMetadataTypeArg(int index)
...
+        }
+
+        // Return the data type, either Decimal or Number, that should be the return type for a function
+        // if the provided argument was the first type encountered.
+        protected DType DetermineNumericFunctionReturnType(bool nativeDecimal, bool numberIsFloat, DType argType)
+        {
+            // when numberIsFloat, favor Number, return type is always Number except when operand is Decimal
+            // when !numberIsFloat, favor Decimal, return type is only Number operand is Number
+            // should match the logic in CheckDecimalBinaryOp in BinderUtils.cs
+#if false
```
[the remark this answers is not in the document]
gregli-msft: Oops, thanks!  Removed.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
gregli-msft: I have no idea why VS keeps inserting that.  Removed.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
gregli-msft: Done.

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -364,14 +368,25 @@ private List<IntermediateNode> AttachArgPreprocessor(List<IntermediateNode> args
...
                         case ArgPreprocessor.ReplaceBlankWithEmptyString:
                             convertedNode = BlankToEmptyString(args[i]);
+                            break;
+                        case ArgPreprocessor.ReplaceBlankWithCallZero_Scalar:
+                            var callIRContext_Scalar = context.GetIRContext(node);
+                            convertedNode = ReplaceBlankWithCallTypedZero_Scalar(args[i], callIRContext_Scalar.ResultType);
+                            break;
+                        case ArgPreprocessor.ReplaceBlankWithCallZero_SingleColumnTable:
+                            var callIRContext_SCT = context.GetIRContext(node);
+                            convertedNode = ReplaceBlankWithCallTypedZero_Scalar(args[i], ((TableType)callIRContext_SCT.ResultType).SingleColumnFieldType);
```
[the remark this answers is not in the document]
gregli-msft: It is as must be a single column table for this ArgPreprocessor to be invoked, so TableType should always work.  Should I add an assert of some kind, would that help?

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
gregli-msft: Fixed.

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -383,30 +398,131 @@ private List<IntermediateNode> AttachArgPreprocessor(List<IntermediateNode> args
...
+                    return arg;
+                }
+
+                // need a new context since when arg is Blank IRContext.ResultType is not a Number but a Blank.
+                if (returnType == FormulaType.Number)
+                {
+                    convertedIRContext = new IRContext(arg.IRContext.SourceContext, FormulaType.Number);
+                    zeroLitNode = new NumberLiteralNode(convertedIRContext, 0d);
+                }
+                else
```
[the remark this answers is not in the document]
gregli-msft: Done.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MathFunction.cs
gregli-msft: Good call.  The error message was different as a result of the change.  I added an else clause consistent with the two arg version.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
gregli-msft: Your guess is as good as mine.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Int.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
     internal sealed class IntFunction : MathOneArgFunction
     {
         public IntFunction()
-            : base("Int", TexlStrings.AboutInt, FunctionCategories.MathAndStat)
+            : base("Int", TexlStrings.AboutInt, FunctionCategories.MathAndStat, nativeDecimal: true)
```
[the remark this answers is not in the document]
gregli-msft: It doesn't change for PA Client.  This just says the function is capable of working with decimals natively.  On PA Client, NumberIsFloat is in place and so no decimals will enter the system, and the decimal equivalent will not be referenced.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
gregli-msft: Done.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
gregli-msft: McCall and I chatted and believe that we should move forward with the approach in the PR.  

We both agree that likely long term we should revise our function overload resolution logic, to better accommodate both Decimal and UDFs, but we don't have to do that now.  For Decimal, we can use the CheckTypes approach as currently coded, and UDFs can have their own rules as a new feature.  We have a discussion about UDFs scheduled for tomorrow.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -689,49 +850,198 @@ private static IEnumerable<double> LazySequence(double records, double start, do
...
         }
 
-        public static FormulaValue Round(IRContext irContext, NumberValue[] args)
+        public static FormulaValue Abs(IRContext irContext, FormulaValue[] args)
         {
-            var numberArg = args[0].Value;
-            var digitsArg = args[1].Value;
+            var arg0 = args[0];
+
+            if (arg0 is NumberValue num)
```
MikeStall, line 867: > if (arg0 is NumberValue num) [](http://example.com/codeflow?start=11&length=29)

This kind of checking motivates me to use overloads.

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
```diff
@@ -47,6 +47,10 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
   </ItemGroup>
+
+  <ItemGroup>
+    <None Remove="expressiontestcases\RoundingOps.txt" />
+  </ItemGroup>
```
MikeStall, line 53: nit remove? Probably VS being silly

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Average.txt
```diff
@@ -35,9 +35,13 @@
 2.34567
 
 //Excel returns -0.781666667
->> Average(-1.234, 2.345, -3.456)
--0.78166667
+// Rounding here so we get consistent results between float and decimal
+>> Round(Average(-1.234, 2.345, -3.456),7)
```
MikeStall, line 39: >Round [](http://example.com/codeflow?start=3&length=5)

Can  BaseRunner.NumberCompare  overload let us handle this?

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Average.txt
gregli-msft: Unfortunately no, decimal comparisons must be exact.  The other option is just to remove this test and rely on the one after it that returns an exact answer rather than a repeating decimal.  But I hate removing tests.  :)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
gregli-msft: Discussed at length today.  Seeing if you come up with some type system magic to help with validating the return value.  If not, we can check in this way, and look for a better option for that part later.  I have cleaned up the separation between decimal and float values and enhanced the checking of the integer arguments for Round*, Trunc, and Sequence.

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
gregli-msft: Fixed.  Love you VS.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
MikeStall: AbsFloat/AbsDecimal is a good mitigation

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
MikeStall: Still there...

MikeStall (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Functions/ArgPreprocessor.cs
```diff
@@ -9,9 +9,17 @@ namespace Microsoft.PowerFx.Core.Functions
 {
     internal enum ArgPreprocessor
```
LucGenetier, line 10: >ArgPreprocessor [](http://example.com/codeflow?start=18&length=15)

I'd add a comment on what this enum is

lesaltzm (approved):

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -383,30 +400,90 @@ private List<IntermediateNode> AttachArgPreprocessor(List<IntermediateNode> args
...
+                    zeroLitNode = new NumberLiteralNode(convertedIRContext, 0d);
+                }
+                else if (returnType == FormulaType.Decimal)
+                {
+                    convertedIRContext = new IRContext(arg.IRContext.SourceContext, FormulaType.Decimal);
+                    zeroLitNode = new DecimalLiteralNode(convertedIRContext, 0m);
+                }
+                else
+                {
+                    throw new NotImplementedException("Unexpcted type");
```
LucGenetier, line 464: >Unexpcted [](http://example.com/codeflow?start=55&length=9)

nit: typo

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -3,8 +3,10 @@
 
 using System;
 using System.Collections.Generic;
+using System.Drawing;
```
LucGenetier, line 6: do we need this one?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -3,8 +3,10 @@
 
 using System;
 using System.Collections.Generic;
+using System.Drawing;
 using System.Globalization;
 using System.Linq;
+using System.Security.Cryptography;
```
LucGenetier, line 9: >Cryptography [](http://example.com/codeflow?start=22&length=12)

this one too?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -607,50 +741,44 @@ public static async ValueTask<FormulaValue> AverageTable(EvalVisitor runner, Eva
                 return CommonErrors.DivByZeroError(irContext);
             }
 
-            return await RunAggregatorAsync("Average", new AverageAgg(), runner, context, irContext, args).ConfigureAwait(false);
```
LucGenetier, line 610: >ConfigureAwait [](http://example.com/codeflow?start=107&length=14)

keep ConfigureAwait

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -662,24 +790,70 @@ private static FormulaValue ModInternal(double arg0, double arg1, IRContext irCo
             return new NumberValue(irContext, result);
         }
 
+        public static FormulaValue ModDecimal(IRContext irContext, DecimalValue arg0dv, DecimalValue arg1dv)
+        {
+            decimal arg0 = arg0dv.Value;
+            decimal arg1 = arg1dv.Value;
+            decimal q;
+
+            if (arg1 == 0m)
```
LucGenetier, line 799: >0m [](http://example.com/codeflow?start=24&length=2)

you could use decimal.Zero

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -662,24 +790,70 @@ private static FormulaValue ModInternal(double arg0, double arg1, IRContext irCo
...
+
+            if (arg1 == 0m)
+            {
+                return CommonErrors.DivByZeroError(irContext);
+            }
+
+            // r = a – N × floor(a/b)
+            try
+            {
+                q = decimal.Floor(arg0 / arg1);
```
LucGenetier, line 807: >q = decimal.Floor(arg0 / arg1); [](http://example.com/codeflow?start=16&length=31)

decimal.MaxValue % 0.7m works in C#

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -689,49 +863,202 @@ private static IEnumerable<double> LazySequence(double records, double start, do
...
         }
 
-        public static FormulaValue Abs(IRContext irContext, NumberValue[] args)
+        private static IEnumerable<decimal> LazySequenceDecimal(int records, decimal start, decimal step)
+        {
+            var x = start;
+            for (var i = 1; i <= records; i++)
+            {
+                yield return x;
+                x += step;
```
LucGenetier, line 872: >x += step [](http://example.com/codeflow?start=16&length=9)

could this overflow?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -689,49 +863,202 @@ private static IEnumerable<double> LazySequence(double records, double start, do
...
+                return AbsFloat(irContext, num);
+            }
+            else if (arg0 is DecimalValue dec)
+            {
+                return AbsDecimal(irContext, dec);
+            }
+            else
+            {
+                return CommonErrors.UnreachableCodeError(irContext);
+            }
```
LucGenetier, line 891: I'd write

return arg[0] switch
{
  NumberValue num => AbsFloat(irContext, num),
  DecimalValue dec => AbsDecimal(irContext, dec),
  _ => CommonError....
}

See [https://learn.microsoft.com/en-US/dotnet/csharp/language-reference/operators/switch-expression](https://learn.microsoft.com/en-US/dotnet/csharp/language-reference/operators/switch-expression)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -689,49 +863,202 @@ private static IEnumerable<double> LazySequence(double records, double start, do
...
+            double x = arg.Value;
+            double val = Math.Abs(x);
             return new NumberValue(irContext, val);
         }
 
-        public static FormulaValue Round(IRContext irContext, NumberValue[] args)
+        public static FormulaValue AbsDecimal(IRContext irContext, DecimalValue arg)
+        {
+            decimal x = arg.Value;
+            decimal val = x < 0m ? -x : x;
```
LucGenetier, line 904: >x < 0m ? -x : x; [](http://example.com/codeflow?start=26&length=16)

why not using decimal.Abs(x)?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -689,49 +863,202 @@ private static IEnumerable<double> LazySequence(double records, double start, do
...
+            return CommonErrors.UnreachableCodeError(irContext);
         }
 
-        internal static double Round(double number, double digits, RoundType rt = RoundType.Default)
+        internal static FormulaValue RoundFloat(IRContext irContext, NumberValue num, double doubleDigs, RoundType rt = RoundType.Default)
         {
-            var s = number < 0 ? -1d : 1d;
+            var number = num.Value;
+            var s = number < 0 ? -1d : 1d;
             var n = number * s;
```
LucGenetier, line 937: >number * s [](http://example.com/codeflow?start=20&length=10)

double.Abs(number) ?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -689,49 +863,202 @@ private static IEnumerable<double> LazySequence(double records, double start, do
...
+            1e-20m, 1e-21m, 1e-22m, 1e-23m, 1e-24m, 1e-25m, 1e-26m, 1e-27m, 1e-28m
+        };
+
+        // The algorithm for Decimal is different from that of Float because with less range we are going out of our way to avoid overflow
+        // At the time of this writing, the version of .NET being targeted only supports two varieties of MidpointRounding
+        // In the future, some of this can be replaced with built in support in decimal.Round()
+        internal static FormulaValue RoundDecimal(IRContext irContext, DecimalValue dec, double doubleDigs, RoundType roundType = RoundType.Default)
+        {
+            var signedNumber = dec.Value;
+            var sign = signedNumber < 0 ? -1m : 1m;
```
LucGenetier, line 985: >signedNumber < 0 ? -1m : 1m; [](http://example.com/codeflow?start=23&length=28)

decimal.Sign ?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -689,49 +863,202 @@ private static IEnumerable<double> LazySequence(double records, double start, do
...
+            var unsignedNumber = signedNumber < 0 ? -signedNumber : signedNumber;
```
LucGenetier, line 986: >signedNumber < 0 ? -signedNumber : signedNumber; [](http://example.com/codeflow?start=33&length=48)

decimal.Abs ?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -741,35 +1068,91 @@ public enum RoundType
...
+            {
+                digits = numberDigs.Value;
+            }
+            else if (args.Length == 1)
+            {
+                digits = 0;
+            }
+            else
+            {
+                return CommonErrors.UnreachableCodeError(irContext);
```
LucGenetier, line 1113: >eturn CommonErrors.UnreachableCodeError(irContext); [](http://example.com/codeflow?start=17&length=51)

would we reach that point if args.Length == 2 && args[1] is DecimalValue ?

LucGenetier:
>         private static FormulaValue Sqrt(IRContext irContext, NumberValue[] args)

no change :cry:  (kidding)

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs:598 in 50f6672. [](commit_id = 50f6672d3e2f52066f5d809b37666e1e1f27320d, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -607,50 +741,44 @@ public static async ValueTask<FormulaValue> AverageTable(EvalVisitor runner, Eva
...
             }
         }
 
-        private static FormulaValue ModInternal(double arg0, double arg1, IRContext irContext)
+        public static FormulaValue ModFloat(IRContext irContext, NumberValue arg0nv, NumberValue arg1nv)
         {
+            double arg0 = arg0nv.Value;
+            double arg1 = arg1nv.Value;
+
             if (arg1 == 0)
```
LucGenetier, line 769: >if (arg1 == 0) [](http://example.com/codeflow?start=12&length=14)

even if absolutely not important, i'd add a comment that double.NegativeZero is also considered here

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DecimalMathFuncs.txt
```diff
@@ -0,0 +1,453 @@
...
+
+>> RoundUp( [Float(300000.1234), 1000000000000000000000001.1, 2000000000000000000000002.2], 2 )
+Table({Value:300000.13},{Value:1000000000000000100000000},{Value:2000000000000000200000000})
+
+>> RoundDown( [Float(30000.1234), 1000000000000000000000001.1, 2000000000000000000000002.2], 2 )
+Table({Value:30000.12},{Value:1000000000000000100000000},{Value:2000000000000000200000000})
+
+// non-decimal aware numeric functions
+
+>> Sqrt( 1524157899404237160815376 )         // exact square of 1234567899876, should be done in float and match next test
```
LucGenetier, line 182: >1524157899404237160815376 [](http://example.com/codeflow?start=9&length=25)

you should try 1524157899404237160815377
(result = 1234567899876.000000000000405000000445678200490442)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DecimalMathFuncs.txt
```diff
@@ -0,0 +1,453 @@
...
+>> Log( 10000000000000000000000, 10 )         // above tests should match these answers
+22
+
+>> Log( 10000000000000000000000, Float(10) )  
+22
+
+>> Log( Float("1e100"), 10 )
+100
+
+>> Ln( 2.71828182845904523536028747135266249775724709369995 )
```
LucGenetier, line 212: >Ln( 2.71828182845904523536028747135266249775724709369995 ) [](http://example.com/codeflow?start=3&length=58)

I'd try Ln( 2.71828182845904523536028747135266249775724709369995 ) - 1

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DecimalMathFuncs.txt
```diff
@@ -0,0 +1,453 @@
...
+1
+
+>> Ln( 100000000 )
+18.420680743952367
+
+>> Exp( 18.420680743952367 )
+100000000.00000007
+
+>> Pi()
```
LucGenetier, line 221: >Pi() [](http://example.com/codeflow?start=3&length=4)

We could return 3.1415926535897932384626433833 (or ...34 as next digit is 8)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DecimalMathFuncs.txt
```diff
@@ -0,0 +1,453 @@
...
+Table({Value:-1})
+
+>> Sin( 0 )
+0
+
+>> Sin( [0] )
+Table({Value:0})
+
+>> Sin( Pi() )
+1.2246467991473532E-16
```
LucGenetier, line 243: Oh!

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DecimalMathFuncs.txt
```diff
@@ -0,0 +1,453 @@
...
+Table({Value:Error({Kind:ErrorKind.Div0})})
+
+>> Cot( 3.14159265358979323846264338327950288419716939937510 )
+-8165619676597685
+
+>> Cot( [3.14159265358979323846264338327950288419716939937510] )
+Table({Value:-8165619676597685})
+
+>> Cot( Pi() )
+-8165619676597685
```
LucGenetier, line 258: What's the value of this test which is wrong?
It should return an overflow error

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DecimalMathFuncs.txt
```diff
@@ -0,0 +1,453 @@
...
+>> Asin( [1] )
+Table({Value:1.5707963267948966})
+
+>> Atan( 0 )
+0
+
+>> Atan( [0] )
+Table({Value:0})
+
+>> Atan( 1 )
```
LucGenetier, line 314: >Atan( 1 ) [](http://example.com/codeflow?start=3&length=9)

you should try 4 * Atan(1) - Pi()

LucGenetier (approved):

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
gregli-msft: Thanks!  Done.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
gregli-msft: I'm taking a new approach to math... :)  VS inserts these things on me.  Is there a way to turn that off?  May be worth me doing if there is.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
gregli-msft: Ditto.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
gregli-msft: That line is old enough that it predates the addition.  Added back in.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
gregli-msft: I could.  I think I prefer the 0m as it looks more like what we do with integer and double.  I like the consistency.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
gregli-msft: I like that!  Thanks!

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
gregli-msft: Interestingly, decimal does not have an Abs method.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -662,24 +790,70 @@ private static FormulaValue ModInternal(double arg0, double arg1, IRContext irCo
...
+                return SequenceFloat(irContext, count, startN, stepN);
+            }
+            else if (args[1] is DecimalValue startW && args[2] is DecimalValue stepW)
+            {
+                return SequenceDecimal(irContext, count, startW, stepW);
+            }
+            else
+            {
+                return CommonErrors.UnreachableCodeError(irContext);
+            }
```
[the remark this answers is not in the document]
gregli-msft: Done!

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
gregli-msft: This is what was there before, I just effectively added a line.  But it makes the code clearer to read, I changed it.  Note that it is Math.Abs() and it has an overload for Decimal too, so I'll fix that one up as well.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
gregli-msft: My bad, it is actually on the Math object.  Updated.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
gregli-msft: There is a Math.Sign but it returns an int.  I really don't want it to have to promote on the fly.  I think I'll leave it as.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
gregli-msft: Correct.  CheckTypes ensures that the second argument is Number, as does Library.cs.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
gregli-msft: Done.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
gregli-msft: Good call.  I followed the float algorithm.  I'm going to put an issue in on this and we can revisit, both for decimal and float.  https://github.com/microsoft/Power-Fx/issues/1367

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
gregli-msft: Great call!  Yes, this could overflow, and I think the right thing to do would be to include an ErrorValue in this situation.  It's easier to hit with Decimal but has the same problem in Float.  I'm going to punt this for the moment and look for the best way to address both.  https://github.com/microsoft/Power-Fx/issues/1368

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
gregli-msft: Done.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DecimalMathFuncs.txt
gregli-msft: Good call.  I went even futher and used 1524157899404230000000000 which still rounds to the same number in float, but would be quite different in decimal.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DecimalMathFuncs.txt
gregli-msft: Comes back with 0.  :)  I'm not trying to identify tiny differences, instead looking for consistency between what decimal and float would return.  And consistency with SQL.  In this case it is all float so it shouldn't matter much.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DecimalMathFuncs.txt
gregli-msft: We could but then it would change the typing to decimal and would be inconsistent with the return type from Sin, Cos, etc.  If someone wants Pi to many digits, they can easily put it in a decimal themeslf.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DecimalMathFuncs.txt
gregli-msft: Good call!  Fixed.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DecimalMathFuncs.txt
MikeStall: How stable is this response?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/RoundingOps.txt
```diff
@@ -0,0 +1,92 @@
...
+
+>> Trunc(Sequence(41,-5,0.25),2)
+Table({Value:-5},{Value:-4.75},{Value:-4.5},{Value:-4.25},{Value:-4},{Value:-3.75},{Value:-3.5},{Value:-3.25},{Value:-3},{Value:-2.75},{Value:-2.5},{Value:-2.25},{Value:-2},{Value:-1.75},{Value:-1.5},{Value:-1.25},{Value:-1},{Value:-0.75},{Value:-0.5},{Value:-0.25},{Value:0},{Value:0.25},{Value:0.5},{Value:0.75},{Value:1},{Value:1.25},{Value:1.5},{Value:1.75},{Value:2},{Value:2.25},{Value:2.5},{Value:2.75},{Value:3},{Value:3.25},{Value:3.5},{Value:3.75},{Value:4},{Value:4.25},{Value:4.5},{Value:4.75},{Value:5})
+
+>> Sum(ForAll(Sequence(41,-5,0.25),Round(Value,2)=RoundDown(Value,2) && Round(Value,2)=RoundUp(Value,2) && Round(Value,2)=Trunc(Value,2)),Value)
+41
+
+// *********** Big numbers
```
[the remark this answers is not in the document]
gregli-msft: Agreed.  But since float is messed up for big numbers, we should wait for that to get fixed first.  https://github.com/microsoft/Power-Fx/issues/1331.  I made a note in that issue that we should add tests here too.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DecimalMathFuncs.txt
gregli-msft: Fixed.  :)

src/libraries/Microsoft.PowerFx.Core/Functions/ArgPreprocessor.cs
gregli-msft: Done.

LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #1334, opened 2023-04-10
New exception handler for missing SymbolValues

anderson-joyle:
This change supports LookUp function when retrieving from Dataverse.

The following expression would throw since the GUID doesn't exist:
`LookUp(t1, localid = GUID(\"00000000-0000-0000-9999-000000000001\")).Price`


LucGenetier:
✅ No public API change.

PR #1338, opened 2023-04-11
Updates to FromJson for NumberIsFloat

gregli-msft:
Adding flag to FormulaValueJSON.FromJson so that JSON can be interpreted with NumberIsFloat semantics.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
+++ b/src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
@@ -22,7 +22,7 @@ namespace Microsoft.PowerFx.Types
         /// <summary>
         /// Convenience method to create a value from a json representation. 
         /// </summary>
-        public static FormulaValue FromJson(string jsonString, FormulaType formulaType = null)
+        public static FormulaValue FromJson(string jsonString, FormulaType formulaType = null, bool numberIsFloat = false)
         {
             try
             {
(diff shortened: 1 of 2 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -59,11 +60,11 @@ public static FormulaValue FromJson(JsonElement element, FormulaType formulaType
                     return FormulaValue.NewBlank(formulaType);
 
                 case JsonValueKind.Number:
-                    if (formulaType is NumberType)
+                    if ((skipTypeValidation && numberIsFloat) || formulaType is NumberType)
```
MikeStall, line 63: numberIsFloat || (formulaType is NumberType)   ?

MikeStall (approved):

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -22,29 +22,30 @@ public class FormulaValueJSON
         /// <summary>
         /// Convenience method to create a value from a json representation. 
         /// </summary>
-        public static FormulaValue FromJson(string jsonString, FormulaType formulaType = null)
+        public static FormulaValue FromJson(string jsonString, FormulaType formulaType = null, bool numberIsFloat = false)
```
MikeStall, line 25: @LucGenetier  - this extra parameter here is a binary breaking change, but the api compat tool didn't flag it. 
Is it scanning this dll?

PR #1342, opened 2023-04-11
Supporting multiple types

### LINKED ISSUE #1332: Supporting Multiple Type fails when Check function is ran with EditorContextScope
Found an issue with supporting multiple types. Seems like when new `EditorContextScope` is added with `allowCoerceTo`, 

`var sameType = this._expectedReturnType == this.ReturnType` in `CheckResult.cs` returns the **sameType** of **false** and adds the error message.
```
        [Fact]
        public void CheckExpectedReturnValueNumber()
        {
            var check = new EditorContextScope((expr) => new CheckResult(new Engine()).SetText(expr).SetBindingInfo().SetExpectedReturnValue(FormulaType.String, true));
            var result = check.Check("123");

        }
```

anderson-joyle:
Supporting multiple types when Check function is ran with EditorContextScope.

Fix #1332

### DIFF
--- a/src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/EditorContextScopeTests.cs
+++ b/src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/EditorContextScopeTests.cs
@@ -199,6 +199,18 @@ namespace Microsoft.PowerFx.Tests.LanguageServiceProtocol
             Assert.Equal("Microsoft.PowerFx.Tests.LanguageServiceProtocol.EditorContextScopeTests+MyHandler", name);
         }
 
+        [Fact]
+        public void CheckExpectedReturnValueNumber()
+        {
+            var editorContextScope = new EditorContextScope(
+                            (expr) => new CheckResult(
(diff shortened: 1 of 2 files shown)


src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/EditorContextScopeTests.cs
```diff
@@ -199,6 +199,18 @@ public void HandlerName()
...
+        [Fact]
+        public void CheckExpectedReturnValueNumber()
+        {
+            var editorContextScope = new EditorContextScope(
+                            (expr) => new CheckResult(
+                                    new Engine()).SetText(expr).SetBindingInfo().SetExpectedReturnValue(FormulaType.String, true));
+
+            var check = editorContextScope.Check("123");
+
+            Assert.True(check.IsSuccess);
```
MikeStall, line 211: Should also assert:
ReturnType=number

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

PR #1343, opened 2023-04-11
Ensure we're using httpClient we're given and don't use a cached one …

LucGenetier:
…(possibly disposed)

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
@@ -153,29 +153,26 @@ namespace Microsoft.PowerFx.Connectors
 
         internal ServiceFunction GetServiceFunction(string ns = null, HttpMessageInvoker httpClient = null, ICachingHttpClient cache = null, bool throwOnError = false)
         {
-            if (_serviceFunction == null)
-            {
-                IAsyncTexlFunction invoker = null;
-                string func_ns = string.IsNullOrEmpty(ns) ? "Internal_Function" : ns;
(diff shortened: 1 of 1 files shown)


Robulane (dismissed):

src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
```diff
@@ -156,29 +156,26 @@ public string GetExpression(string @namespace, ConnectorParameters parameters)
 
         internal ServiceFunction GetServiceFunction(string ns = null, HttpMessageInvoker httpClient = null, ICachingHttpClient cache = null, bool throwOnError = false)
         {
-            if (_serviceFunction == null)
```
MikeStall, line 159: >_serviceFunction [](http://example.com/codeflow?start=16&length=16)

Discussed offline,  but there are correctness issues here. 

consider this sequence:

GetServiceFunction(client1)
// _serviceFunction now points to client1
binding1 on expression, now captures _serviceFunction

GetServiceFunction(client2)
binding1._serviceFunction now points to client2!

MikeStall: The key challenge here is _serviceFunction is *shared state* under contention, and 2 callers  want 2 separate HttpClients.  Before, we had a first-writer wins policy; previous iteration was just last-writer wins policy .. both were wrong. 

Removing the field is key because that removes the contention.

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1347, opened 2023-04-11
Revert "adds void type to IfError()"

MikeStall:
Reverts microsoft/Power-Fx#1230

Reverting because it was a breaking change - not ready to take in Power Apps yet. 
-Previously allowed now error:
expressions such as IfError(1, {x:1}, {x:1}, {y:1}, 1) Since it can potentially return a record or a number type.


Shpakh-MSFT (approved):
:shipit:

jas-valgotar (approved):

MikeStall:
@CarlosFigueiraMSFT , fyi.

LucGenetier:
✅ No public API change.

PR #1351, opened 2023-04-12
Add IR transform support

MikeStall:
Support for IR rewriting.  Since IR is internal, these changes are internal too and not directly exposed. 
This enables things like Delegation and constant folding. 

IR Rewriters can be registered with the engine and then run on ApplyErrors and ApplyIR.


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/IR/Visitors/RewritingIRVisitor.cs
```diff
@@ -0,0 +1,251 @@
...
+    // Common helper for IR Rewriting. 
+    // IR trees are immutable, so we need to walk and rewrite the tree to include any new nodes.
+    // Only return new nodes if needed, so if there are no changes, the original object instances are returned.
+    internal abstract class RewritingIRVisitor<TRet, TCtx> : IRNodeVisitor<TRet, TCtx>
+    {
+        protected abstract IntermediateNode Materialize(TRet ret);
+
+        protected abstract TRet Ret(IntermediateNode node);
+
+        public override TRet Visit(TextLiteralNode node, TCtx context)
```
lesaltzm, line 20: /// <summary> comments for these methods would be helpful for future implementers.

src/libraries/Microsoft.PowerFx.Core/IR/Visitors/RewritingIRVisitor.cs
```diff
@@ -0,0 +1,251 @@
...
+                }
+
+                newArgs?.Add(result);
+            }
+
+            return newArgs;
+        }
+
+        // Pass in arg0 is we already called Accept on it.
+        public TRet Visit(CallNode node, TCtx context, TRet arg0)
```
lesaltzm, line 148: >public [](http://example.com/codeflow?start=8&length=6)

should this one be private?

MikeStall: possibly protected? Derived classes need to call it.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -525,6 +539,19 @@ internal IRResult ApplyIR()
...
+                if (list != null)
+                {
+                    foreach (var transform in list)
+                    {
+                        irnode = transform.Transform(irnode, _errors);
+
+                        // Additional errors from phases. 
+                        // Stop any further processing if we have errors. 
+                        this.ThrowOnErrors();
+                    }
```
lesaltzm, line 552: Wouldn't users generally want all the errors?

src/tests/Microsoft.PowerFx.Core.Tests/IRTests/IRPrettyPrinter.cs
```diff
@@ -0,0 +1,215 @@
...
+using Microsoft.PowerFx.Core.IR.Symbols;
+
+namespace Microsoft.PowerFx.Tests
+{
+    // Pretty printer for tests. 
+    // Stable  - tests rely on this, so we need to fix the format 
+    // Compact / No newlines - ideally can fit on a single line in Theorys
+    // Deterministic ordering - when traversing dictionaries
+    // Unambiguous - additional parenthesis as needed to show (a+b)+c vs. a+(b+c). 
+    internal class PrettyPrintIRVisitor : IRNodeVisitor<PrettyPrintIRVisitor.RetVal, PrettyPrintIRVisitor.Context>
```
lesaltzm, line 18: Would be useful in Core, for CheckResult.PrintIR()

(which, also, should probably not be public on CheckResult??)

lesaltzm (changes requested):
:clock1:

lesaltzm (approved):
:shipit:

src/tests/Microsoft.PowerFx.Core.Tests/IRTests/IRPrettyPrinter.cs
MikeStall: starting in Test seemed safe. We could move in future. 
Note that this Compact form is much less invasive than the one in product.   More ideal for test baselines, less ideal for diagnostics.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
MikeStall: IR is assuming we have valid input (we don't even try translation if binding fails, see 539). 

So to preserve that invariant, IR is always assuming we're valid, and hence any translation aborts if we get errors.  (warnings are ok). 
It makes it much easier to reason about now - although we could try to expand in the future if we care to get more errors.

lesaltzm (approved):
🚢

LucGenetier:
✅ No public API change.

PR #1359, opened 2023-04-12
Fix #1337 - support [@ ] for symbols

### LINKED ISSUE #1337: Proper usage of [@ ] disambiguation
When in row scope of Accounts, 

`CountRows([@Tasks])`

Should refer to Task table, not  Account.Tasks field. 

This is the workaround for #1336.

See https://learn.microsoft.com/en-us/power-platform/power-fx/operators#disambiguation-operator
It's implemented for RowScope on lambdas/With, but not for row scope introduced by symbols.

MikeStall:
Fix #1337

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTableOverRecordType.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTableOverRecordType.cs
@@ -72,9 +72,17 @@ namespace Microsoft.PowerFx
         {
             return slot.SlotIndex == int.MaxValue;
         }
-
-        internal override bool TryLookup(DName name, out NameLookupInfo nameInfo)
+                
+        bool INameResolver.Lookup(DName name, out NameLookupInfo nameInfo, NameLookupPreferences preferences)
         {
+            if (preferences.HasFlag(NameLookupPreferences.GlobalsOnly))
+            {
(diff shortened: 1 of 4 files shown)


src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTableOverRecordType.cs
```diff
@@ -72,9 +72,17 @@ public bool IsThisRecord(ISymbolSlot slot)
         {
             return slot.SlotIndex == int.MaxValue;
         }
-
-        internal override bool TryLookup(DName name, out NameLookupInfo nameInfo)
+                
+        bool INameResolver.Lookup(DName name, out NameLookupInfo nameInfo, NameLookupPreferences preferences)
         {
+            if (preferences.HasFlag(NameLookupPreferences.GlobalsOnly))
```
MikeStall, line 78: This is the key fix. 

We conveniently know which symbols are row scope because they're in this class.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

anderson-joyle (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1370, opened 2023-04-14
Revert "Add Decimal function support"

lesaltzm:
Reverts microsoft/Power-Fx#1333


LucGenetier:
✅ No public API change.

PR #1371, opened 2023-04-14
Revert "Update Accepts/CoercesTo to take a flag to control legacy behavior and breaks accepts relationships with the flag enabled"

lesaltzm:
Reverts microsoft/Power-Fx#1281


gregli-msft (approved):

LucGenetier:
✅ No public API change.

Issue #1335, opened 2023-04-10
Take "NumberAsFloat" as an option in FormulaValueJSON.FromJson


coletaylor1:
When updating packages, we want to disable decimal as it is a breaking change which introduces risk to our release. However, we add values to our engine parsed using the following expression. This expression is returning `decimal` types causing the engine to return `decimal` types, despite the flag `NumberAsFloat = true` being passed in the parser options.

`FormulaValueJSON.FromJson(JsonConvert.SerializeObject(value))`

coletaylor1:
Found an example where logic is not working when values are converted to `Decimal`. 

`Collect(checktable, {Flavor: \"Strawberry\", Quantity: 300 })`

Here `checktable` is a table initialized to have a single record `{Flavor:"Raspberry",Quantity:200}`. Initialization uses `FormulaValueJSON.FromJson(JsonConvert.SerializeObject(value))` for the value for `Quantity` and it therefore has type `Decimal`. 

When running the `Collect` function above. The following error is thrown:

`{Error 20-58: The type of this argument 'Quantity' does not match the expected type 'Decimal'. Found type 'Number'.}`

While this particular issue can be fixed by letting us control how we parse values by default, customers can still explicitly provide decimals and likely hit this error that way as well. Will let feature owner evaluate additional impact and file additional issues.

MikeStall:
That issue may be fixed with #1309

MikeStall:
Fixed with 1338

MikeStall:
Mitigated with https://github.com/microsoft/Power-Fx/pull/1338, 
We should capture this scenario in a test.

PR #1373, opened 2023-04-14
Adding regression test to #1335

anderson-joyle:
Issue #1335.

https://github.com/microsoft/Power-Fx/pull/1338 has mitigated the issue but a regression test case was missing.


src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationFunctionsTests.cs
```diff
@@ -20,6 +20,7 @@
 using Microsoft.PowerFx.Interpreter;
 using Microsoft.PowerFx.Syntax;
 using Microsoft.PowerFx.Types;
+using Newtonsoft.Json;
```
MikeStall, line 23: can we avoid?

MikeStall (approved):

LucGenetier:
✅ No public API change.

Issue #1346, opened 2023-04-11
Use Decimal for return type of Len, Count*, etc. when NumberIsFloat is not enabled


gregli-msft:
Even with NumberIsFloat disables, these functions currently return a floating point value.  If these functions return a floating point value, then that taints any downstream calculations based on them.  In a non-NumberIsFloat world, that can lead to needless precision loss.  

The request from PVA is to have these functions return a Decimal value when NumberIsFloat is disabled.

PR #1378, opened 2023-04-14
Support decimal return values for integer returning functions

gregli-msft:
A variety of our functions return simple integers, such as Len, Count, Year, Hex2Dec, etc.  At present, they all return a Number, even if NumberIsFloat is disabled.  This means that calculations involving them will be promoted to Float and there could be a loss of precision.  This PR addresses this by returning a Decimal for these functions if NumberIsFloat is not in effect.

Addresses https://github.com/microsoft/Power-Fx/issues/1346.


Shpakh-MSFT (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DateTime.cs
```diff
@@ -264,6 +285,18 @@ public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
...
+        }
+
+        public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DType[] argTypes, IErrorContainer errors, out DType returnType, out Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
+        {
+            var fValid = base.CheckTypes(context, args, argTypes, errors, out _, out nodeToCoercedTypeMap);
+
+            // As this is an integer returning function, it can be either a number or a decimal depending on NumberIsFloat.
+            // We do this to preserve decimal precision if this function is used in a calculation
+            // since returning Float would promote everything to Float and precision could be lost
+            returnType = context.NumberIsFloat ? DType.Number : DType.Decimal;
```
LucGenetier, line 297: >context.NumberIsFloat ? DType.Number : DType.Decimal; [](http://example.com/codeflow?start=25&length=53)

As this is used in MANY locations, wondering if that couldn't be useful to have a method like
context.GetDefaultNumericType()

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -1237,6 +1248,28 @@ public static FormulaValue RandBetween(IServiceProvider services, IRContext irCo
...
+                    Message = $"Lower value cannot be greater than Upper value",
+                    Span = irContext.SourceContext,
+                    Kind = ErrorKind.Numeric
+                });
+            }
+
+            lower = Math.Ceiling(lower);
+            upper = Math.Floor(upper);
+
+            decimal value = (decimal)services.SafeNextDouble();
```
LucGenetier, line 1269: >(decimal)services.SafeNextDouble() [](http://example.com/codeflow?start=28&length=34)

We should really have a SafeNextDecimal() method here

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -1300,6 +1333,70 @@ private static ErrorValue GetDiv0Error(IRContext irContext)
...
+                return new NumberValue(irContext, (double)value);
+            }
+            else if (irContext.ResultType == FormulaType.Decimal)
+            {
+                return new DecimalValue(irContext, (decimal)value);
+            }
+            else
+            {
+                return CommonErrors.UnreachableCodeError(irContext);
+            }
```
LucGenetier, line 1356: (nit) I would make it more readable with this 

return irContext.ResultType._type.Kind switch
            {
                DKind.Number => new NumberValue(irContext, (double)value),
                DKind.Decimal => new DecimalValue(irContext, (decimal)value),
                _ => CommonErrors.UnreachableCodeError(irContext)
            };

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DecimalIntReturn.txt
```diff
@@ -0,0 +1,75 @@
...
+21.0000000000000000000001
+
+>> Second( DateTime(1990,3,2,5,20,54) ) + 1.0000000000000000000001
+55.0000000000000000000001
+
+>> RandBetween(11,11) + 1.00000000000000000000000001
+12.00000000000000000000000001
+
+>> Hex2Dec( "1234BEEF" ) + 1.00000000000000000000000001
+305446640
```
LucGenetier, line 69: Could be interesting to try 
>> Hex2Dec( "1234BEEF" ) + 0.99999999999999999999999999

LucGenetier (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DateTime.cs
gregli-msft: I hear you, but if the check is as simple as this one, I prefer to keep it exposed so that the underlying logic could be seen.  If it was anything more than a single ternary operator, I'd put it in a function.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Drawing;
```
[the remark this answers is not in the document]
gregli-msft: Thanks.  Damn you VS.  I think I've turned this feature off, but we'll see.  :)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -1237,6 +1248,28 @@ public static FormulaValue RandBetween(IServiceProvider services, IRContext irCo
...
+                    Span = irContext.SourceContext,
+                    Kind = ErrorKind.Numeric
+                });
+            }
+
+            lower = Math.Ceiling(lower);
+            upper = Math.Floor(upper);
+
+            decimal value = (decimal)services.SafeNextDouble();
+            return new DecimalValue(irContext, Math.Floor((value * (upper - lower + 1)) + lower));
```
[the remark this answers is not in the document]
gregli-msft: Added.  Ints coerce to decimal cleanly so there isn't technically a need, but it is clearer.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
gregli-msft: I think I'll leave it for now.  I don't like reaching into the _type if not needed, and that patterns isn't used in LibraryMath.cs.  It is used in LibraryData.cs however.  :)  Something we should apply consistently, but that can happen at another time.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DecimalIntReturn.txt
gregli-msft: Added!

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -1300,6 +1333,70 @@ private static ErrorValue GetDiv0Error(IRContext irContext)
...
+            {
+                return new NumberValue(irContext, value);
+            }
+            else if (irContext.ResultType == FormulaType.Decimal)
+            {
+                try
+                {
+                    return new DecimalValue(irContext, (decimal)value);
+                }
+                catch
```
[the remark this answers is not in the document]
gregli-msft: Done.

LucGenetier (dismissed):

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Count.cs
```diff
@@ -23,7 +23,7 @@ internal sealed class CountFunction : FunctionWithTableInput
         public override bool IsSelfContained => true;
 
         public CountFunction()
-            : base("Count", TexlStrings.AboutCount, FunctionCategories.Table | FunctionCategories.MathAndStat, DType.Number, 0, 1, 1, DType.EmptyTable)
+            : base("Count", TexlStrings.AboutCount, FunctionCategories.Table | FunctionCategories.MathAndStat, DType.Unknown, 0, 1, 1, DType.EmptyTable)
```
MikeStall, line 26: Just double checking impact of setting this to .Unknown.

gregli-msft: This is what we are thinking we should revert.  I changed these to .Unknown because it is unknown (could be either Decimal or Number) and CheckTypes will determine.  This change appears to break PA Client and is making it hard for Carlos to integrate.  The interpreter doesn't appear to care either way.  So, we are in the process of changing these back, if it works, for PA Client.  Stay tuned...

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Hex2Dec.cs
```diff
@@ -26,13 +26,25 @@ public override ArgPreprocessor GetArgPreprocessor(int index)
         public override bool IsStateless => true;
 
         public Hex2DecFunction()
-            : base("Hex2Dec", TexlStrings.AboutHex2Dec, FunctionCategories.MathAndStat, DType.Number, 0, 1, 1, DType.String)
+            : base("Hex2Dec", TexlStrings.AboutHex2Dec, FunctionCategories.MathAndStat, DType.Unknown, 0, 1, 1, DType.String)
```
CarlosFigueiraMSFT: >Unknown [](http://example.com/codeflow?start=94&length=7)

Number

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/RandBetween.cs
```diff
@@ -2,37 +2,23 @@
...
-                "RandBetween",
-                TexlStrings.AboutRandBetween,
-                FunctionCategories.MathAndStat,
-                returnType: DType.Number,
-                maskLambdas: 0,
-                arityMin: 2,
-                arityMax: 2,
-                DType.Number,
-                DType.Number)
+            : base("RandBetween", TexlStrings.AboutRandBetween, minArity: 2, nativeDecimal: true)
```
CarlosFigueiraMSFT, line 21: >nativeDecimal: true [](http://example.com/codeflow?start=77&length=19)

We should not rely on this being true

CarlosFigueiraMSFT (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Count.cs
gregli-msft: The change to DType.Number appears to have worked.  We are moving forward with having all native Decimal functions return DType.Number.  When we add Decimal to PA Client, we can revisit this approach.

LucGenetier:
✅ No public API change.

Issue #1375, opened 2023-04-14
ErrNumberExpected should give actual type too


MikeStall:
We replaced a bunch of cases of ErrBadType_ExpectedType_ProvidedType with ErrNumberExpected.

So before, we had this, which included **actual** type.
  [1]: {Error 8-15: Invalid argument type (**Table**). Expecting a **Number** value instead.}
  
  Now we get this, which doesn't include actual type. 
{Error 8-15: Expected number. We expect a **number** at this point in the formula.}

PR #1380, opened 2023-04-17
Updates to CheckType error handling in function CheckTypes

gregli-msft:
Addresses https://github.com/microsoft/Power-Fx/issues/1375

Today, in some places, we call CheckTypes with "DefaultErrorHandler" as the error collection to catch the errors.  This is a /dev/null like error container where errors are not reported.  In general, we shouldn't need to use this - we should be calling CheckType once for each argument.  

Either:
1. The function's CheckTypes should take full responsibility for calling CheckType/CheckColumnType itself and not rely on the base.  This is done when the base didn't have good information about the parameters because their data types were not fixed, such as calling a function that can take/return Decimal vs. Number arguments.
2. The function's CheckTypes should defer to the base for argument processing and only augment with additional processing that does not involve calling CheckType/CheckColumnType.

The problem now is that it is confusing.  DefaultErrorHandler seemed like it was doing something useful when it was not.  The class and the one legit usage of it have been renamed to NullErrorHandler to better reflect how and when it would be used and the instance has been removed from TexlFunction.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/FunctionScopeInfo.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/FunctionScopeInfo.cs
@@ -142,7 +142,10 @@ namespace Microsoft.PowerFx.Core.Functions
         // Same as the virtual overload, however all typechecks are done quietly, without posting document errors.
         public virtual bool CheckInput(Features features, TexlNode inputNode, DType inputSchema, out DType typeScope)
         {
-            return CheckInput(features, inputNode, inputSchema, TexlFunction.DefaultErrorContainer, out typeScope);
+            // A "no-op" or /dev/null error container that does not post document errors.
+            NullErrorContainer nullErrorContainer = new NullErrorContainer();
+
+            return CheckInput(features, inputNode, inputSchema, nullErrorContainer, out typeScope);
         }
 
         public void CheckLiteralPredicates(TexlNode[] args, IErrorContainer errors)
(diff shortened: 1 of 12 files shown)


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
```diff
@@ -34,7 +34,7 @@ public class FileExpressionEvaluationTests : PowerFxTest
         public void InterpreterTestCase(ExpressionTestCase testCase)
         {
             // This is running against embedded resources, so if you're updating the .txt files,
-            // make sure they build is actually copying them over.
+            // make sure they build is actually copying them over.ab
```
CarlosFigueiraMSFT, line 37: >ab [](http://example.com/codeflow?start=66&length=2)

typo?

gregli-msft: Oops, typo.  It's how I trigger the tests to rebuild after a .txt change without having to rebuild everything.

src/libraries/Microsoft.PowerFx.Core/Functions/FunctionScopeInfo.cs
```diff
@@ -141,8 +141,11 @@ public virtual bool CheckInput(Features features, TexlNode inputNode, DType inpu
 
         // Same as the virtual overload, however all typechecks are done quietly, without posting document errors.
         public virtual bool CheckInput(Features features, TexlNode inputNode, DType inputSchema, out DType typeScope)
-        {
-            return CheckInput(features, inputNode, inputSchema, TexlFunction.DefaultErrorContainer, out typeScope);
+        {
+            // A "no-op" or /dev/null error container that does not post document errors.
+            NullErrorContainer nullErrorContainer = new NullErrorContainer();
```
Shpakh-MSFT, line 146: >ullErrorContainer nullErrorContainer = new NullErrorContainer(); [](http://example.com/codeflow?start=13&length=64)

Why remove the static instance? Do we need this new allocation per call?

PR #1381, opened 2023-04-17
Allow equality comparison between date/time values

CarlosFigueiraMSFT:
With the new PowerFxV1 feature set, we are stricter with comparing values of different types. Something like `1 = "1"` has always returned false, which is probably not what the maker intended, so we are making them errors. Comparisons between dates and times, however, should be allowed, since for Excel compatibility they are equivalent to numbers. This change relaxes the restriction and makes the equality / inequality comparison between date/time values valid.


LucGenetier:
✅ No public API change.

gregli-msft (approved):

PR #1386, opened 2023-04-17
Fix Error function to work with strongly typed enums

CarlosFigueiraMSFT:
With the StronglyTypedBuiltinEnums on, the Error function wasn't working properly on its overload that takes a table. This change fixes it.


LucGenetier:
✅ No public API change.

lesaltzm (approved):

LucGenetier:
✅ No public API change.

PR #1387, opened 2023-04-17
Fixes IfError under V1 flag

### LINKED ISSUE #1173: Review IfError()'s CheckType() and add void type support.

jas-valgotar:
- Followup to https://github.com/microsoft/Power-Fx/pull/1230
- Fixes #1173


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

lesaltzm (approved):

LucGenetier:
✅ No public API change.

Shpakh-MSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #1389, opened 2023-04-18
Update float tests

gregli-msft:
Simple change to use the Value function instead of Float for PA Client compatibility.


LucGenetier:
✅ No public API change.

Issue #1360, opened 2023-04-13
Collect error message has wrong number of args 


MikeStall:
Reprod with:
`Collect(Tasks,{Subject:"test",Owner:First(Users)})`

On this line in CollectFunctions
errors.EnsureError(DocumentErrorSeverity.Severe, args[0], TexlStrings.ErrNeedValidVariableName_Arg, Name);

https://github.com/microsoft/Power-Fx/blob/f5ef31b5d61e45ee25cbc314477638bfb964747b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/CollectFunction.cs#L186

message
"The first argument of '{0}' should be a valid variable name, and cannot conflict with any existing control, screen, collection, or data source names. Found type '{1}'"
args
{object[1]}
    [0]: "Collect"

MikeStall:
Also hitting with:
`Collect(Faxes, { Subject : "First", Regarding : Index(Accounts, 2) } )`

anderson-joyle:
PR #1393 provided mitigation for the problem. This will stay open and will be moved to GA 1.1.

PR #1393, opened 2023-04-18
Collection to display better error message

anderson-joyle:
To mitigate issue #1360.

New error message:

> Incompatible type. The item you are trying to put into a table has a type that is not compatible with the table.

`Union` fails here:
https://github.com/microsoft/Power-Fx/blob/5b890ffe6bd07eff9a7671d01ca100316fcc3c2a/src/libraries/Microsoft.PowerFx.Core/Types/DType.cs#L2813

Then PFx tries to show which column has failed. The problem is that Union fails for a particular reason, but we use a different method to find the problematic column. PFx can accept Polymorphic/Record types. For some reason that I couldn't identify PA Client can't. This will result in `Owner:{e}` and PFx handling this scenario at
https://github.com/microsoft/Power-Fx/blob/5b890ffe6bd07eff9a7671d01ca100316fcc3c2a/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs#L888

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/CollectFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/CollectFunction.cs
@@ -188,7 +188,7 @@ namespace Microsoft.PowerFx.Interpreter
                     fValid = false;
                     if (!SetErrorForMismatchedColumns(collectionType, collectedType, args[1], errors, context.Features))
                     {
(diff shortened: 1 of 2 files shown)


MikeStall (approved):
:shipit:

MikeStall (approved):
Approved - but is there any way to get the faulting column / or expected type vs. actual type?

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #1400, opened 2023-04-19
Validate we can rename option sets and their values properly

LucGenetier:


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #1407, opened 2023-04-20
Revert DType.Unknown for native decimal function return type

gregli-msft:
Several functions can now return either DType.Decimal or DType.Number.  Reflecting this, these functions advertise their return type as DType.Unknown in the function info constructor.  This was causing problems for PA Client, and does not appear to actually impact the interpreter, so we are going to change it back to DType.Number which is all PA Client supports at the moment.  It isn't clear how this information would have been used anyway, as CheckTypes for each function ultimately determines the return type.


LucGenetier:
✅ No public API change.

MikeStall (approved):

CarlosFigueiraMSFT (approved):
🚢

PR #1408, opened 2023-04-20
Reversing the default for ReservedKeywords

### LINKED ISSUE #1385: Add reserved words

gregli-msft:
In general, new features shouldn't change default behavior, they should be an opt-in.  However in this case it makes sense to make it an opt-out, controlled by an internal flag that Cnavs can set.  The nature of this feature and that it is set in ParserOptions and not directly effected by engine.Features (as there may not be one) led us to this conclusion.  

Fixes #1385


MikeStall:
This makes sense because we are intentionally changing the default behavior.  So we want to get in for GA

src/tests/Microsoft.PowerFx.Core.Tests/TestRunnerTests/InternalSetup.cs
```diff
@@ -26,6 +26,12 @@ internal class InternalSetup
         /// </summary>
         internal bool DisableMemoryChecks { get; set; }
 
+        /// <summary>
+        /// By default, we run the result back through the serializer and deserialize again to se if we get the same result.
```
MikeStall, line 30: >se [](http://example.com/codeflow?start=95&length=2)

nit: typo

src/tests/Microsoft.PowerFx.Core.Tests/TestRunnerTests/InternalSetup.cs
```diff
@@ -26,6 +26,12 @@ internal class InternalSetup
         /// </summary>
         internal bool DisableMemoryChecks { get; set; }
 
+        /// <summary>
+        /// By default, we run the result back through the serializer and deserialize again to se if we get the same result.
+        /// For some tests, for example reserved words, the deserialization will not work properly.
+        /// </summary>
+        internal bool SkipDeserializeComparison { get; set; }
```
MikeStall, line 33: >SkipDeserializeComparison [](http://example.com/codeflow?start=22&length=25)

Do we need the new flag - can we just infer this if the test case has an error that we shouldn't try serializing?

MikeStall (approved):
:shipit:

src/tests/Microsoft.PowerFx.Core.Tests/TestRunnerTests/InternalSetup.cs
MikeStall: ?

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -362,12 +367,12 @@ protected override async Task<RunResult> RunAsyncInternal(string expr, string se
...
                     {
                         // Serialization test. Serialized expression must produce an identical result.
-                        ParserOptions options = new ParserOptions() { NumberIsFloat = true };
+                        options.NumberIsFloat = true;
                         newValueDeserialized = await engine.EvalAsync(sb.ToString(), CancellationToken.None, options, runtimeConfig: runtimeConfig).ConfigureAwait(false);
                     }
                     else
                     {
-                        throw;
+                        throw e;
```
MikeStall, line 375: > throw e; [](http://example.com/codeflow?start=23&length=9)

nit - changing to 'throw e' means we lose the callstack of the original exception. 
Can we keep as 'throw' ?

MikeStall (dismissed):
:shipit:

src/tests/Microsoft.PowerFx.Core.Tests/TestRunnerTests/InternalSetup.cs
MikeStall: Tracking #1409 to remove this flag.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/TestRunnerTests/InternalSetup.cs
gregli-msft: Flag removed and tests skipped for now.  I'll make a note in 1409, but the tests should be unskipped when that is done.

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
gregli-msft: Fixed.

MikeStall (approved):

LucGenetier:
✅ No public API change.

Issue #1409, opened 2023-04-20
RecordValue.ToExpression should escape identifiers as needed 


MikeStall:
You can use ticks to make an identifier that conflicts with a keyword.  Normally ticks are also used when identifiers contain spaces (such as for display names).
this is important since identifiers and keywords are different tokens, so it's handled at the lexer level. 

`{ As : 3}`  // not legal, As is a keyword
`{ 'As' : 3}` // legal, 'As' is an identifer. 

When RecordValue.ToExpression serializes, it needs to escape identifiers that may conflict with keywords.  Basically, IdentToken.MakeValidIdentifier, will escape for spaces, but we need to also check keywords.

MikeStall:
Note we can optimistically enclose things in ticks, so we could just escape any possible keyword. But not all identifiers.

PR #1411, opened 2023-04-20
RecordValue.ToExpression to escape identifiers

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/1409.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
@@ -248,11 +248,6 @@ namespace Microsoft.PowerFx.Syntax
             return _keywords.Keys.ToList();
         }
 
-        public static IReadOnlyList<string> GetReservedKeywords()
-        {
-            return _reservedKeywords.ToList();
-        }
-
         private TexlLexer(string preferredDecimalSeparator)
         {
             // List and decimal separators.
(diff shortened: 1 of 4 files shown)


src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -246,6 +246,11 @@ public static TexlLexer GetLocalizedInstance(CultureInfo culture)
         public static IReadOnlyList<string> GetKeywords()
         {
             return _keywords.Keys.ToList();
+        }
+
+        public static IReadOnlyList<string> GetReservedKeywords()
```
MikeStall, line 251: Ienumerable? then we can avoid ToList call.

src/tests/Microsoft.PowerFx.Interpreter.Tests/SerializationTests.cs
```diff
@@ -97,7 +100,44 @@ public void OptionSetDefaultExpressionValueErrorTests()
...
+        }
+
+        [Fact]
+        public void RecordValueKeyworkSerializationTest()
+        {
+            var engine = new RecalcEngine(new PowerFxConfig());
+            var symbol = new SymbolTable();
+
+            // Combining both keywords and reserved keywords
+            var keywords = TexlLexer.GetKeywords().Concat(TexlLexer.GetReservedKeywords()).ToList();
```
MikeStall, line 112: this list should be fixed. can we inline it?

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
anderson-joyle: This has been removed since we adopted the inline approach in the new test cases.

LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #1412, opened 2023-04-21
Allow renaming of option sets

LucGenetier:
New RenameOptionSet API on RenameDriver


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/DisplayNameTests.cs
```diff
@@ -411,6 +404,37 @@ public void RenameParameter(string expressionBase, string expectedExpression, st
...
+
+            DPath oldNameAsPath = DPath.Root;
+            foreach (var segment in oldName.Split('.'))
+            {
+                oldNameAsPath = oldNameAsPath.Append(new DName(segment));
+            }
+
+            Engine engine = new Engine(config);
+
+            RenameDriver renamer = engine.CreateFieldRenamer(r1, oldNameAsPath, new DName(newName), CultureInfo.InvariantCulture);
```
lesaltzm, line 434: Interesting calling pattern. What if there's a field matching oldName as well as an optionSet? A method on engine: `createOptionSetRenamer` may be a better approach.

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/RenameDriver.cs
```diff
@@ -59,6 +59,15 @@ public string ApplyRename(string expressionText)
             return ExpressionLocalizationHelper.ConvertExpression(converted, strippedRenameParameters, BindingConfig.Default, _resolver, _binderGlue, CultureInfo.InvariantCulture, Features.None, false);
         }
 
+        public string RenameOptionSet(string expressionText)
```
lesaltzm, line 62: Why do we need a separate RenameOptionSet method? If you've set the _renameResolver/_renameParameters right in the constructor, wouldn't ApplyRename work?

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/RenameDriver.cs
```diff
@@ -67,16 +76,52 @@ public string ApplyRename(string expressionText)
...
+            return updated ? new ComposedReadOnlySymbolTable(st, rost) : resolver;
+        }
+
         private FormulaType RenameFormulaTypeHelper(AggregateType nestedType, Queue<DName> segments, DName updatedName)
         {
             var field = segments.Dequeue();
             if (segments.Count == 0)
             {
-                // Create a display name provider with only the name in question
+                // Create a display name provider with only the name in question
```
lesaltzm, line 124: nit: extra whitespace

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/RenameDriver.cs
```diff
@@ -67,16 +76,52 @@ public string ApplyRename(string expressionText)
...
+
+            if (!rost.TryLookup(pathToRename[0], out NameLookupInfo nameInfo))
+            {
+                return resolver;
+            }
+
+            SymbolTable st = new SymbolTable();
+            bool updated = false;
+
+            if (nameInfo.Type.Kind == Types.DKind.OptionSet)
```
lesaltzm, line 97: Maybe: `if (nameInfo.Data is OptionSet os)`?

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/RenameDriver.cs
```diff
@@ -67,16 +76,52 @@ public string ApplyRename(string expressionText)
...
+                if (pathToRename.Length == 1)
+                {
+                    st.AddEntity(os, updatedName);
+                    updated = true;
+                }
+                else if (pathToRename.Length == 2)
+                {
+                    if (os.TryGetValue(pathToRename[1], out OptionSetValue osValue))
+                    {
+                        st.AddEntity(new OptionSet(os.EntityName, new SingleSourceDisplayNameProvider(new Dictionary<DName, DName> { [new DName(osValue.Option)] = updatedName })));
```
lesaltzm, line 110: Ah: this'll probably need to have the other logical->logical pairs in the existing option set, I think?

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/RenameDriver.cs
LucGenetier: Correct. Let me remove it

src/tests/Microsoft.PowerFx.Core.Tests/DisplayNameTests.cs
```diff
@@ -411,6 +404,37 @@ public void RenameParameter(string expressionBase, string expectedExpression, st
...
+            OptionSet optionSet1 = new OptionSet("firstos", DisplayNameUtility.MakeUnique(new Dictionary<string, string>()
+            {
+                    { "option_1", "Option1" },
+                    { "option_2", "Option2" }
+            }));
+
+            config.AddOptionSet(optionSet1, new DName("TestFirstOptionSet"));
+
+            RecordType r1 = RecordType.Empty()                
+                .Add(new NamedFormulaType("Os1Value", optionSet1.FormulaType, "DisplayOS1Value"));
```
lesaltzm, line 424: Another test: Add a member to this RecordType with logical name `firstos`, make sure that references to that don't get renamed at the same time as references to the global, which would need to be accessed as `[@firstos]` if that collision exists

src/libraries/Microsoft.PowerFx.Core/Public/Config/ComposedReadOnlySymbolTable.cs
```diff
@@ -136,6 +136,11 @@ public virtual bool Lookup(DName name, out NameLookupInfo nameInfo, NameLookupPr
             return false;
         }
 
+        internal override bool TryLookup(DName name, out NameLookupInfo nameInfo)
```
MikeStall, line 139: >TryLookup [](http://example.com/codeflow?start=31&length=9)

What's the relationship here to Lookup() on 125?

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/RenameDriver.cs
LucGenetier: Good call

src/libraries/Microsoft.PowerFx.Core/Public/Config/ComposedReadOnlySymbolTable.cs
LucGenetier: Lookup is doing a TryLookup but isn't wired to it in the base class which always return false

LucGenetier: I'll remove it

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using System.Text.RegularExpressions;
```
MikeStall, line 8: used?

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/RenameDriver.cs
```diff
@@ -19,13 +18,15 @@ public sealed class RenameDriver
         private readonly RecordType _baseParameters;
         private readonly RecordType _renameParameters;
         private readonly INameResolver _resolver;
+        private readonly INameResolver _renameResolver;
```
MikeStall, line 21: What the difference between these 2 INameResolvers?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/RenameDriver.cs
```diff
@@ -67,10 +70,46 @@ public string ApplyRename(string expressionText)
...
-
             return invariantExpression != ApplyRename(invariantExpression);
         }
 
+        private INameResolver RenameResolverHelper(INameResolver resolver, DPath pathToRename, DName updatedName)
+        {
+            if (resolver is not ReadOnlySymbolTable rost)
+            {
+                throw new Exception("Unexpected resolver");
+            }
```
MikeStall, line 81: Can we push this up to and have signature just take in a ReadOnlySymbolTable?

MikeStall: or actually ... do we need the cast? It looks like we go resolver --> ReadOnlySymbolTable --> resolver?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
LucGenetier: not a  lot, i remove it

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/RenameDriver.cs
LucGenetier: the second one is used for renaming option sets

src/tests/Microsoft.PowerFx.Core.Tests/DisplayNameTests.cs
MikeStall: can you add a test demonstrating this scenario works?

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/DisplayNameTests.cs
LucGenetier: done

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -350,7 +349,12 @@ public RenameDriver CreateFieldRenamer(RecordType parameters, DPath pathToRename
             ** but that we don't return any display names for them. Thus, we clone a PowerFxConfig but without 
             ** display name support and construct a resolver from that instead, which we use for the rewrite binding.
             */
-            return new RenameDriver(parameters, pathToRename, updatedName, this, CreateResolverInternal(), CreateBinderGlue(), culture);
+            return new RenameDriver(parameters, pathToRename, updatedName, this, CreateResolverInternal() as ReadOnlySymbolTable, CreateBinderGlue(), culture, false);
```
MikeStall, line 352: >as ReadOnlySymbolTable [](http://example.com/codeflow?start=106&length=22)

This could potentially break existing paths if CreateFieldRenamer does not return ROST. 

But 
1.  that can only happen in cases where Engine.CreateREsolver is derived - which is internal. Probably just a PA scenario, and PA doesn't call this method anyways. 
2. OptionSetRename would still be broken. 

So it's probably ok.

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

Issue #1294, opened 2023-03-31
Null object reference for negate of date/datetime/time


gregli-msft:
The interpreter's Negate function can't handle a non-numeric input.  And it shouldn't have to, as the compiler should block these operations from appearing in the IR.  
```
> Time(1,0,0)
Time(1,0,0,0)

> -Time(1,0,0)
One or more errors occurred. (Object reference not set to an instance of an object.)

> IR(-Time(1,0,0))
Negate:T(Time:T(Coalesce:n(Float:n(1:w), 0:n), Coalesce:n(Float:n(0:w), 0:n), Coalesce:n(Float:n(0:w), 0:n)))

> IR(-Date(2,0,0))
Negate:D(Date:D(Coalesce:n(Float:n(2:w), 0:n), Coalesce:n(Float:n(0:w), 0:n), Coalesce:n(Float:n(0:w), 0:n)))

> IR(-DateTime(2,0,0,0,0,0))
Negate:d(DateTime:d(Coalesce:n(Float:n(2:w), 0:n), Coalesce:n(Float:n(0:w), 0:n), Coalesce:n(Float:n(0:w), 0:n), Coalesce:n(Float:n(0:w), 0:n), Coalesce:n(Float:n(0:w), 0:n), Coalesce:n(Float:n(0:w), 0:n)))
```
They are permitted partially because `a-b` translates into `a+-b` and that is needed for `DateTime(1,0,0,0,0,0)-Time(1,0,0)`.  Before they get to Negate, the opcode is replaced with a binary op for doing subtraction.
```
> IR(Now()-Time(1,0,0))
SubtractDateAndTime:d(Now:d(), Time:T(Coalesce:n(Float:n(1:w), 0:n), Coalesce:n(Float:n(0:w), 0:n), Coalesce:n(Float:n(0:w), 0:n)))

> IR(Time(2,0,0)-Time(1,0,0))
TimeDifferenceDecimal:w(Time:T(Coalesce:n(Float:n(2:w), 0:n), Coalesce:n(Float:n(0:w), 0:n), Coalesce:n(Float:n(0:w), 0:n)), Time:T(Coalesce:n(Float:n(1:w), 0:n), Coalesce:n(Float:n(0:w), 0:n), Coalesce:n(Float:n(0:w), 0:n)))
```

MikeStall:
Callstack is:
System.NullReferenceException: Object reference not set to an instance of an object.
[...]

PR #1414, opened 2023-04-21
Handle null object for negate of datetime/date/time

nguhoa:
Fix issue https://github.com/microsoft/Power-Fx/issues/1294

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
@@ -474,8 +474,13 @@ namespace Microsoft.PowerFx.Functions
 #endregion
 
 #region Unary Operator Implementations
-        private static NumberValue NumericNegate(IRContext irContext, NumberValue[] args)
+        private static FormulaValue NumericNegate(IRContext irContext, NumberValue[] args)
         {
+            if (args.Length == 0 || args[0] == null)
+            {
+                return CommonErrors.GenericInvalidArgument(irContext);
+            }
+
             var result = -args[0].Value;
             return new NumberValue(irContext, result);
         }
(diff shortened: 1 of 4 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
```diff
@@ -474,11 +474,16 @@ internal static partial class Library
 #endregion
 
 #region Unary Operator Implementations
-        private static NumberValue NumericNegate(IRContext irContext, NumberValue[] args)
-        {
+        private static FormulaValue NumericNegate(IRContext irContext, NumberValue[] args)
+        {
+            if (args.Length == 0 || args[0] == null)
```
MikeStall, line 479: >if (args.Length == 0 || args[0] == null) [](http://example.com/codeflow?start=12&length=40)

How does this happen?

nguhoa: it does not. i saw it pass null value in to negate func.

MikeStall: That may be the bug ... do you know where those nulls are coming from?

CarlosFigueiraMSFT: This is just masking the symptom of the problem, not addressing the root cause. It should be valid to negate a time/date/datetime value. Something like `-dateValue` should be equivalent to `NumberToDate(Negate(DateToNumber(dateValue)))` - and similar for time/datetime.

CarlosFigueiraMSFT (changes requested):
:clock1:

PR #1415, opened 2023-04-21
Crash: Polymorphic field in a within With() expression 

jas-valgotar:
-Fixes https://github.com/microsoft/Power-Fx-Dataverse/issues/171

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Types/RecordType.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Types/RecordType.cs
@@ -99,6 +99,10 @@ namespace Microsoft.PowerFx.Types
             return new KnownRecordType();
         }
 
+        /// <summary>
+        /// Static builder method for constructing a Polymorphic record type.
+        /// Used for Dataverse Polymorphic types.
+        /// </summary>
         public static RecordType Polymorphic()
         {
             return new KnownRecordType(DType.Polymorphic);
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Types/RecordType.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Types/RecordType.cs
@@ -25,7 +25,7 @@ namespace Microsoft.PowerFx.Types
         internal RecordType(DType type)
             : base(type)
         {
-            Contracts.Assert(type.IsRecord);
+            Contracts.Assert(type.IsRecord || type.IsPolymorphic);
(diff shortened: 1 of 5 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -290,7 +290,7 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
                             });
                     }
                     
-                    if (!(result.IRContext.ResultType == node.IRContext.ResultType || result is ErrorValue || result.IRContext.ResultType is BlankType || func is IfErrorFunction))
+                    if (!(result.IRContext.ResultType._type == node.IRContext.ResultType._type || result is ErrorValue || result.IRContext.ResultType is BlankType || func is IfErrorFunction))
```
MikeStall, line 293: >_type [](http://example.com/codeflow?start=54&length=5)

Do we need this? Looks like it's to invoke Dtype's equality. 
FormulaType overrides == already and forward to Dtype?

MikeStall (dismissed):

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
jas-valgotar: Yes, we should only rely on DType inside Binder...

Think of 2 different implementation of RecordType, If they have same underlying dtype, should this check pass?

src/libraries/Microsoft.PowerFx.Core/Public/Types/RecordType.cs
```diff
@@ -99,6 +99,11 @@ public static RecordType Empty()
             return new KnownRecordType();
         }
 
+        public static RecordType Polymorphic()
+        {
+            return new KnownRecordType(DType.Polymorphic);
```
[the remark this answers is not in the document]
jas-valgotar: We should loose it to have compile and runtime type compat.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier (approved):
:shipit:

Issue #1297, opened 2023-04-01
"16^Blank()" and "Blank()^16" leads to a null exception


gregli-msft:
The ^ operator inserts a Call node to the Power function in IRTranslator.cs:
```
                switch (kind)
                {
                    // Call Node Replacements:
                    case BinaryOpKind.Power:
                        binaryOpResult = new CallNode(context.GetIRContext(node), BuiltinFunctionsCore.Power, left, right);
                        break;
```
And the Power entry in Library.cs says that handling blanks is handled by the IR:
```
            {
                BuiltinFunctionsCore.Power,
                StandardErrorHandling<NumberValue>(
                    BuiltinFunctionsCore.Power.Name,
                    expandArguments: NoArgExpansion,
                    replaceBlankValues: NoOpAlreadyHandledByIR,
                    checkRuntimeTypes: ExactValueTypeOrBlank<NumberValue>,
                    checkRuntimeValues: DeferRuntimeValueChecking,
                    returnBehavior: ReturnBehavior.AlwaysEvaluateAndReturnResult,
                    targetFunction: Power)
            },
```
And indeed, there is a pre-processor setup for Power in Power.cs:
```
        public override ArgPreprocessor GetArgPreprocessor(int index)
        {
            return base.GetGenericArgPreprocessor(index);
        }
```
However, since the Call node is inserted directly by the IR Translator, the standard function processing, like inserting blank checks into the IR, is skipped.

We should go through all the places we insert direct calls and ensure that the library replaces blanks rather than having the IR do it.  Or have the insert of Call nodes follow standard procedures for arg preprocessing.

PR #1416, opened 2023-04-21
Add null check for power

nguhoa:
Fix issue https://github.com/microsoft/Power-Fx/issues/1297


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -1158,6 +1158,11 @@ public static FormulaValue Exp(IRContext irContext, NumberValue[] args)
 
         public static FormulaValue Power(IRContext irContext, NumberValue[] args)
         {
+            if (args.Length < 2 || args[0] == null || args[1] == null)
+            {
```
MikeStall, line 1162: How is this possible? Does binder guarantee these are not-null?

nguhoa: As Greg's mentioned in description of issue: "However, since the Call node is inserted directly by the IR Translator, the standard function processing, like inserting blank checks into the IR, is skipped." so we need to check blank/null in direct calls.

LucGenetier:
✅ No public API change.

nguhoa:
Fixed in PR: https://github.com/microsoft/Power-Fx/pull/1462 so close this PR

PR #1417, opened 2023-04-21
Fix DateTime-to-Time conversion and time value serialization

CarlosFigueiraMSFT:
The DateTime-to-Time conversion is stripping out the days part of the value, which causes those two expressions to have different values (and they should be the same):

- Value(If(1>0, DateTime(1900,1,1,0,0,0), Time(12,0,0)))
- Value(If(1<0, Time(12,0,0), DateTime(1900,1,1,0,0,0)))

This change also fixes the serialization for TimeValue objects to prevent negative hours/minutes/seconds from showing up in case of expressions such as `Time(6,0,0) - 1`.


MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier (approved):

PR #1420, opened 2023-04-21
Extending Language Server to support Semantic Tokenization and Migrating Tokenization Logic from Canvas App Source to PowerFx

akshar2401:
This pr extends language server to handle two kinds of semantic tokens request (Full and Range Semantic Tokens). This pr also migrates the core tokenization logic from Canvas App backend to the PowerFx. In doing so, i had to update an interface which might break PowerFx.DataVerse but the Canvas App backend would not have any breaking changes after these changes are merged in and consumed there.  Due to little to no experience in working with PowerFx, i may have done few things against the established norms around PowerFx so i'd greatly appreciate any feedback and more than happy to reconsider things.  The semantic tokenization implemented in this PR follows the LSP specification around semantic tokenization and encoding as described here https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_semanticTokens. Language Server also supports multi line tokens such as multi line comments, multi line strings. 

Some results from the end to end prototype created based on these changes in PowerFx

Before:
image

After  (Ignore the Colors which would be refined later):
image

Before:
image

After  (Ignore the Colors which would be refined later):
image

Before:
image

After  (Ignore the Colors which would be refined later):
image


src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalEntityScope.cs
```diff
@@ -18,6 +18,15 @@ internal interface IExternalEntityScope
...
             where T : class, IExternalEntity;
+
+        /// <summary>
+        /// Checks if the given name is available or not.
+        /// <para>This is used in Tokenization to determine if namespace of the dotted name can be hidden or not.</para>
+        /// </summary>
+        /// <param name="name">Name.</param>
+        /// <param name="ignoreNamedFormulas">Flag indicating whether to ignore named formulas or not when checking for the availability of the given name.</param>
+        /// <returns>True if name is available and not used or false otherwise.</returns>
+        bool IsNameAvailable(string name, bool ignoreNamedFormulas = false);
```
jas-valgotar, line 29: >bool IsNameAvailable(string name, bool ignoreNamedFormulas = false); [](http://example.com/codeflow?start=8&length=68)

Do we also need it in case of Dataverese?
[https://github.com/microsoft/Power-Fx-Dataverse/blob/main/src/PowerFx.Dataverse/CdsEntityMetadataProvider.cs#L32](https://github.com/microsoft/Power-Fx-Dataverse/blob/main/src/PowerFx.Dataverse/CdsEntityMetadataProvider.cs#L32)

If not, Can we have a new Interface implementing current and not break the Dataverse?

akshar2401: That was my first thought but then i would need to change the definition of GlobalScope to implement the new interface instead of IExternalEntity in the backend of Canvas App. There are significant number of references to IExternalEntity in there and it would introduce a lot more compiler errors. I had to compromise somewhere so i thought of adding it to the IExternalEntity itself so whenever dataverse consumes the new power fx it could implement this one method. It won't break right away as dataverse would first need to consume new version after my changes are merged in. Anyways most of the methods of IExternalEntity in dataverse throw notsupportedexception so the implementation might be trivial.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -124,6 +127,11 @@ private protected INameResolver CreateResolverInternal(out ReadOnlySymbolTable s
             return symbols;
         }
 
+        private protected virtual IExternalRuleScopeResolver CreateExternalRuleScopeResolver()
```
akshar2401, line 130: This IExternalRuleScopeResolver is used in Canvas App backend when calling TexlBinding.Run(). There was no option to supply a custom implementation of this in Engine, so following the existing pattern in Engine, added this virtual function for the derived classes of Engine to override and supply custom implementation.

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Tokenization.cs
```diff
@@ -0,0 +1,319 @@
...
+                tokens.Add(new TokenTextSpan(dottedName.Node.Right.Name, span.Min + 1, span.Lim, dottedNameType, /* canHide */ false));
+            }
+
+            // String interpolations
+            ExtractTokensFromStringInterpolationNodesList(tokens, binding, complilerGeneratedCallNodes);
+
+            foreach (var call in binding.GetCalls())
+            {
+                // Skip the compiler generated nodes as they don't appear in the actual source code and should not up as a token
+                if (complilerGeneratedCallNodes.Contains(call.Node.Id))
```
akshar2401, line 90: Binder generates a transient Concatenate call node whose span overlaps with the span $" (start of interpolated string). This node was not skipped in the original version of the whole Tokenization logic and this resulted into two overlapping tokens. Ideally, Concatenate is a compiler generated node and should not appear in the set of tokens which are purely generated out of user typed expression. Since monaco-editor doesn't support overlapping tokens and Concatenate appears before $", $" was dropped and was highlighted as a function. Legacy formula bar suffers from the same problem but since it doesn't use monaco's in built tokenization, it got lucky.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -588,8 +589,17 @@ public FormulaType GetNodeType(TexlNode node)
 
         // Called by language server to get custom tokens.
         // If binding is available, returns context sensitive tokens.  $$$
+        // Keeping this temporarily for custom publish tokens notification
```
akshar2401, line 592: I do have another user story to revisit Publish Tokens and clean this up later on

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -277,14 +287,7 @@ private BindingConfig GetDefaultBindingConfig()
...
             var bindingConfig = new BindingConfig(result.Parse.Options.AllowsSideEffects, useThisRecordForRuleScope, result.Parse.Options.NumberIsFloat);
-
-            var binding = TexlBinding.Run(
-                glue,
-                parse.Root,
-                resolver,
-                bindingConfig,
-                ruleScope: ruleScope?._type,
-                features: Config.Features);
+            var binding = TexlBinding.Run(glue, externalRuleScopeResolver, new DataSourceToQueryOptionsMap(), parse.Root, resolver, bindingConfig, false, ruleScope?._type, false, null, Config.Features);
```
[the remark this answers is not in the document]
akshar2401: logically it is still the same as previous overload was passing "null" for IExternalRuleScopeResolver and the default implementation of CreateExternalRuleScopeResolver() in Engine.cs returns null.

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Tokenization.cs
```diff
@@ -0,0 +1,320 @@
...
+        /// <summary>
+        /// Returns an enumeration of token text spans in a expression rule with their start and end indices and token type.
+        /// </summary>
+        /// <param name="expression">Expression which would be tokenized.</param>
+        /// <param name="binding">Binding which would be used to tokenize the operation and determine the type of each token.</param>
+        /// <param name="comments">Colllection of comment tokens extracted from the given expression.</param>
+        /// <param name="comparer">optional comparer to sort tokens.</param>
+        /// <param name="computeHiddenTokens">Optional flag to indicate whether to compute whether token can be hidden or not.</param>
+        /// <returns>Ordered or undordered collection of tokens.</returns>
+        internal static ICollection<ITokenTextSpan> Tokenize(string expression, TexlBinding binding, IEnumerable<CommentToken> comments = null, IComparer<ITokenTextSpan> comparer = null, bool computeHiddenTokens = false)
```
akshar2401, line 41: We don't want to compute hiddenness of the tokens for the new formula bar as we are not going to support that so by default it is off. However, we want to keep this for old formula bar as long as it exists and once we consume the changes in Canvas App backend, we would update the tokenization logic behind old formula bar to use this one instead and computeHiddenTokens would be true for old formula bar.

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Tokenization.cs
```diff
@@ -0,0 +1,320 @@
...
+    internal static class Tokenization
+    {
+        /// <summary>
+        /// Returns an ordered enumeration of token text spans in a expression rule with their start and end indices and token type. Orders token by default comparer.
+        /// </summary>
+        /// <param name="expression">Expression which would be tokenized.</param>
+        /// <param name="binding">Binding which would be used to tokenize the operation and determine the type of each token.</param>
+        /// <param name="comments">Colllection of comment tokens extracted from the given expression.</param>
+        /// <returns>Ordered collection of tokens.</returns>
+        internal static ICollection<ITokenTextSpan> Tokenize(string expression, TexlBinding binding, IEnumerable<CommentToken> comments = null)
```
akshar2401, line 27: This overload would be mostly used by old formula bar once these changes are consumed

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -415,6 +422,152 @@ private void HandleCodeActionRequest(string id, string paramsJson)
...
+                // Empty tokens for the empty expression
+                SendEmptySemanticTokensResponse(id);
+                return;
+            }
+
+            // Monaco-Editor sometimes uses \r\n for the newline character. \n is not always the eol character so allowing clients to pass eol character
+            var eol = queryParams?.Get("eol");
+            eol = !string.IsNullOrEmpty(eol) ? eol : EOL.ToString();
+
+            var tokenTypesToSkip = ParseTokenTypesToSkipParam(queryParams.Get("tokenTypesToSkip"));
```
akshar2401, line 452: We are planning to combine top down tokenizer (client level) and semantic tokenizer (language server level) to deliver best and efficient user experience when it comes to tokenization and syntax highlighting. We are working on creating a top down tokenizer in monaco-editor that matches TexlLexer. With that, we can recognize many of the token types such as numbers, booleans, string literals, parts of string interpolations, both types of comments at the client level itself. This opens up an opportunity to skip these token types when doing semantic tokenizer thus decreasing the time it takes to compute and process these tokens at the server level and also greatly reduce the size of the response we have to send back to the client. For example, if there is a string that spans 15 lines and given that the encoding implementation creates 15 string tokens for each line to support multi line tokens and also given that each token is encoded to 5 integers according to LSP specification, we can reduce the size of the encoded tokens array by 15 * 5 integers if we could recognize it at the client level and skip it at the language server level. This ability allows us to skip arbitrary token types from the client level. Currently, this is done by passing a list of token types as a value of query param "tokenTypesToSkip" in document URI. Original solution was to use TokenType enum which contains all possible token types as Flags and pass in a bit mask in the document URI and check for the bit status to determine whether to skip token type or not. This would have required changing values in TokenType enum to be powers of 2 but didn't choose do it due to recent regression in this enum that broke old formula bar's syntax highlighting. It is very sensitive. however the list of token types is not huge so this alternative should be fine

jas-valgotar (commented):
/azp run

jas-valgotar (commented):
/azp run 'Power-Fx PR (PFX)'

jas-valgotar (commented):
/azp run "Power-Fx PR (PFX)"

LucGenetier (commented):
/AzurePipelines run

src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalEntityScope.cs
```diff
@@ -18,6 +18,15 @@ internal interface IExternalEntityScope
 
         bool TryGetEntity<T>(DName currentEntityEntityName, out T externalEntity)
             where T : class, IExternalEntity;
+
+        /// <summary>
+        /// Checks if the given name is available or not.
+        /// <para>This is used in Tokenization to determine if namespace of the dotted name can be hidden or not.</para>
```
lesaltzm, line 24: Note: not a namespace. Perhaps: "Determine if LHS of the dotted name ..."

src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalEntityScope.cs
lesaltzm: This is definitely the correct place to add this, fwiw. It's a method on the object that this was intended to be an interface for. We should look into getting DV off of this interface.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -588,8 +589,18 @@ public FormulaType GetNodeType(TexlNode node)
...
+        // Feature to auto fix casing of function name in the editor also depends on this function and custom publish tokens notification
         internal IReadOnlyDictionary<string, TokenResultType> GetTokens(GetTokensFlags flags) => GetTokensUtils.GetTokens(this._binding, flags);
 
+        /// <summary>
+        /// Returns an enumeration of token text spans in a expression rule with their start and end indices and token type.
+        /// </summary>
+        /// <param name="comparer">Optional comparer to sort tokens.</param>
+        /// <param name="computeHiddenTokens">Optional flag to indicate whether to compute whether token can be hidden or not.</param>
+        /// <returns>Ordered or undordered collection of tokens.</returns>
+        internal ICollection<ITokenTextSpan> GetTokens(IComparer<ITokenTextSpan> comparer = null, bool computeHiddenTokens = false) => Tokenization.Tokenize(_expression, _binding, Parse.Comments, comparer, computeHiddenTokens);
```
lesaltzm, line 602: >ICollection [](http://example.com/codeflow?start=17&length=11)

why ICollection vs IEnumerable?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -5,8 +5,11 @@
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using System.Xml.Linq;
 using Microsoft.PowerFx.Core;
```
lesaltzm, line 9: unused?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
lesaltzm: Add this as a comment in the code! Very useful context

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Tokenization.cs
```diff
@@ -0,0 +1,320 @@
...
+        /// <summary>
+        /// Returns an enumeration of token text spans in a expression rule with their start and end indices and token type.
+        /// </summary>
+        /// <param name="expression">Expression which would be tokenized.</param>
+        /// <param name="binding">Binding which would be used to tokenize the operation and determine the type of each token.</param>
+        /// <param name="comments">Colllection of comment tokens extracted from the given expression.</param>
+        /// <param name="comparer">optional comparer to sort tokens.</param>
+        /// <param name="computeHiddenTokens">Optional flag to indicate whether to compute whether token can be hidden or not.</param>
+        /// <returns>Ordered or unordered collection of tokens.</returns>
+        internal static ICollection<ITokenTextSpan> Tokenize(string expression, TexlBinding binding, IEnumerable<CommentToken> comments = null, IComparer<ITokenTextSpan> comparer = null, bool computeHiddenTokens = false)
```
lesaltzm, line 41: >computeHiddenTokens [](http://example.com/codeflow?start=192&length=19)

naming: allowTokenHiding?

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Tokenization.cs
lesaltzm: Consider also adding part of this as a comment for this arg, capture that we don't intend to enable it at any point.

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Tokenization.cs
```diff
@@ -0,0 +1,320 @@
...
+            // Numeric Literals
+            ExtractTokensFromNodesList(tokens, binding.GetNumericLiterals(), TokenizerConstants.NumericLiteral);
+
+            // Decimal Literal Nodes
+            ExtractTokensFromNodesList(tokens, binding.GetDecimalLiterals(), TokenizerConstants.DecimalLiteral);
+
+            // String Literals
+            ExtractTokensFromNodesList(tokens, binding.GetStringLiterals(), TokenizerConstants.StringLiteral);
+
+            // Unary Operators ie Currently just used for Percentages
```
lesaltzm, line 152: >Currently just used for Percentages [](http://example.com/codeflow?start=34&length=35)

Not unary negate or not?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -415,6 +422,152 @@ private void HandleCodeActionRequest(string id, string paramsJson)
...
+            }
+
+            var tokenTypesToSkip = ParseTokenTypesToSkipParam(queryParams.Get("tokenTypesToSkip"));
+            var scope = _scopeFactory.GetOrCreateInstance(semanticTokensParams.TextDocument.Uri);
+            var result = scope.Check(expression);
+
+            // Skip over the token types that clients don't want in the response
+            var tokens = result.GetTokens().Where(tok => !tokenTypesToSkip.Contains(tok.TokenType));
+
+            // Only consider overlapping tokens. end index is exlcusive
```
lesaltzm, line 511: A bunch of this looks shared with HandleFullDocumentSemanticTokens, can some of it be factored out to a common function?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -415,6 +422,152 @@ private void HandleCodeActionRequest(string id, string paramsJson)
...
+                    if (tokenType != TokenType.Lim)
+                    {
+                        tokenType = tokenType == TokenType.Min ? TokenType.Unknown : tokenType;
+                        tokenTypesToSkip.Add(tokenType);
+                    }
+                }
+
+                return tokenTypesToSkip;
+            }
+            catch
```
lesaltzm, line 541: >catch [](http://example.com/codeflow?start=12&length=5)

Should we be specific here about what exceptions we intend to catch?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -415,6 +422,152 @@ private void HandleCodeActionRequest(string id, string paramsJson)
...
+            var tokens = result.GetTokens().Where(tok => !tokenTypesToSkip.Contains(tok.TokenType));
+
+            // Only consider overlapping tokens. end index is exlcusive
+            var overlappingTokens = tokens.Where(token => !(token.EndIndex <= startIndex || token.StartIndex >= endIndex));
+            var encodedTokens = SemanticTokensEncoder.EncodeTokens(overlappingTokens, expression, eol);
+
+            _sendToClient(JsonRpcHelper.CreateSuccessResult(id, new SemanticTokensResponse() { Data = encodedTokens }));
+        }
+
+        private static ICollection<TokenType> ParseTokenTypesToSkipParam(string rawTokenTypesToSkipParam)
```
lesaltzm, line 518: >ICollection [](http://example.com/codeflow?start=23&length=11)

Again, why ICollection instead of IEnumerable? Just curious, might be something I don't know.

src/tests/Microsoft.PowerFx.Core.Tests/IntellisenseTests/TokenizationTests.cs
```diff
@@ -0,0 +1,203 @@
...
+using Microsoft.PowerFx.Syntax;
+using Xunit;
+
+namespace Microsoft.PowerFx.Core.Tests.IntellisenseTests
+{
+    public class TokenizationTests : PowerFxTest
+    {
+        [Theory]
+        [InlineData("gallery1", 1, 0, "gallery1", 0, 8, TokenType.Control, false)]
+        [InlineData("gallery1!Selected!nestedLabel1", 3, 0, "gallery1", 0, 8, TokenType.Control, false)]
```
lesaltzm, line 24: >! [](http://example.com/codeflow?start=29&length=1)

we're moving away from allowing `!` in any cases, consider just leaving out those tests.

src/tests/Microsoft.PowerFx.Core.Tests/IntellisenseTests/TokenizationTests.cs
```diff
@@ -0,0 +1,203 @@
...
+            }
+
+            public void AddControlAsControlType(string name)
+            {
+                var controlType = new DType(DKind.Control);
+                var controlInfo = new NameLookupInfo(BindKind.Control, controlType, DPath.Root, 0);
+                Add(name, controlInfo);
+            }
+
+            public void AddTable(string name, BindKind? type = null, params TypedName[] keyValues)
```
lesaltzm, line 172: Ah, I was wondering how you were getting control types to show up. Nice mock!

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/LanguageServerTests.cs
```diff
@@ -1376,6 +1378,373 @@ public void CustomCheckResult()
...
+
+        private static T AssertAndGetResponsePayload<T>(string response, string id)
+        {
+            Assert.NotNull(response);
+            var derializedResponse = JsonDocument.Parse(response);
+            var root = derializedResponse.RootElement;
+            root.TryGetProperty("id", out var responseId);
+            Assert.Equal(id, responseId.GetString());
+            root.TryGetProperty("result", out var resultElement);
+            var @params = JsonSerializer.Deserialize<T>(resultElement.GetRawText(), _jsonSerializerOptions);
```
lesaltzm, line 1711: >@params [](http://example.com/codeflow?start=16&length=7)

nit: can you name this something that doesn't require `@`?

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
akshar2401: I didn't pay attention to it really. I will change it to IEnumerable.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
akshar2401: I debated that but we really don't need to catch it and log it as failure to parse tokenTypesToSkip param is considered as not skipping any tokenTypes. But i agree the general good practice is to be specific. I will change this

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
akshar2401: I am using a Contains method in the places where this is used. I know there's where and other method but simple Contains does that trick. This is a hashset. I know there's also Contains on IEnumerable but not sure how it works and if it performs a linear search. 
image

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Tokenization.cs
akshar2401: Yes, i copied over this comment as well but i did look into this now that you pointed out and we only add/track UnaryOpNode whose Token kind is TokKind.PercentSign in the Binder.cs
image
Since we only track UnaryOpNode with PercentSign, binding.GetUnaryOperators() returns nodes with Token.Kind == TokKind.PercentSign. Also added a quick test to confirm this

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalEntityScope.cs
akshar2401: Resolving this since putting this function directly in this interface is not a problem

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
akshar2401: Added

src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalEntityScope.cs
akshar2401: Done

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
akshar2401: Changed To IEnumerable

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Tokenization.cs
akshar2401: Renamed

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Tokenization.cs
```diff
@@ -0,0 +1,320 @@
...
+        }
+
+        /// <summary>
+        /// Determines if namespaces can be hidden to reduce complexity when the right-hand side of a dotted name is unique.
+        /// When there is a conflict, we need to show the fully qualified name.
+        /// </summary>
+        /// <param name="nameInfo">NameInfo.</param>
+        /// <param name="binding">TexlBinding.</param>
+        /// <returns>True if the namespace can be hidden or false otherwise.</returns>
+        private static bool CanHideNamespaceOfDottedName(NameInfo nameInfo, TexlBinding binding)
```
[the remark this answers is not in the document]
akshar2401: Updated to left hand side

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
akshar2401: Refactored

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
akshar2401: Catching specific type of exceptions

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
akshar2401: Resolving this. Let me know if you have any other concerns

src/tests/Microsoft.PowerFx.Core.Tests/IntellisenseTests/TokenizationTests.cs
akshar2401: Removed those test cases

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -588,8 +589,18 @@ public FormulaType GetNodeType(TexlNode node)
...
         // If binding is available, returns context sensitive tokens.  $$$
+        // Keeping this temporarily for custom publish tokens notification
+        // Feature to auto fix casing of function name in the editor also depends on this function and custom publish tokens notification
         internal IReadOnlyDictionary<string, TokenResultType> GetTokens(GetTokensFlags flags) => GetTokensUtils.GetTokens(this._binding, flags);
 
+        /// <summary>
+        /// Returns an enumeration of token text spans in a expression rule with their start and end indices and token type.
+        /// </summary>
+        /// <param name="comparer">Optional comparer to sort tokens.</param>
+        /// <param name="allowTokenHiding">Optional flag to indicate whether to compute whether token can be hidden or not.</param>
```
MikeStall, line 600: > compute whether token can be hidden or not [](http://example.com/codeflow?start=79&length=43)

what does this mean?

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -588,8 +589,18 @@ public FormulaType GetNodeType(TexlNode node)
...
+        /// <returns>Ordered or undordered collection of tokens.</returns>
```
MikeStall, line 601: >Ordered or undordered  [](http://example.com/codeflow?start=21&length=22)

nit: phrasing?  just "unordered"  as is, not necessarily sorted by span.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
MikeStall: Ack on benefits of client-side tokenization. But:

1. How will we ensure this new JS implementation stays in sync with C# impl?
2. What happens when they diverge?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -415,6 +422,152 @@ private void HandleCodeActionRequest(string id, string paramsJson)
             _sendToClient(JsonRpcHelper.CreateSuccessResult(id, codeActions));
         }
 
+        /// <summary>
+        /// Handles requests to compute semantic requests for the full document or expression.
```
MikeStall, line 426: >semantic requests [](http://example.com/codeflow?start=40&length=17)

Can we clarify: what is "semantic request"?

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/LanguageServerTests.cs
```diff
@@ -1376,6 +1378,373 @@ public void CustomCheckResult()
...
+        [Theory]
+        [InlineData("Create", TokenType.Function, TokenType.BoolLit)]
+        [InlineData(null)]
+        [InlineData("")]
+        [InlineData("[2, 3")]
+        [InlineData("Create", TokenType.StrInterpStart, TokenType.BinaryOp, TokenType.NumLit, TokenType.DecLit, TokenType.Control)]
+        [InlineData("[]")]
+        [InlineData("1,2]")]
+        [InlineData("[98]")]
+        [InlineData("Create", (TokenType)78)]
```
MikeStall, line 1412: >(TokenType)78 [](http://example.com/codeflow?start=30&length=13)

where is 78 coming from?
ie, if the test starts returning 79, how do we know if that's a  product bug or a test bug?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
akshar2401: Folks from VSCode/Monaco recommend combining both Top Down Tokenizer and Semantic Tokenizer for the best tokenization experience. There are various reasons for this but most revolve around efficiency. We have a lexer in powerfx which tokenizes the expression but it doesn't know if an identifier token is a function name or the first name is referring to control or not until the binder. Binder adds semantics to each token. The idea is the same. Client Top down tokenizer acts as a PowerFx Lexer while Language Server adds semantics to each token. Monaco-Editor has two types of tokenization mechanisms (top-down tokenizers and semantic tokenizers). There are certainly many benefits of this combination. However, as you pointed out, there are challenges to this approach which manly revolve around keeping them sync
Both questions touch the same point. 

1) This is definitely a problem but even before these changes, we have been using top down tokenizer  in new formula bar and it has not been correctly set up and also  out of sync with power fx lexer hence there's another work going on to bring up it up to speed with power fx lexer. The cleanest solution for this would be powerfx's implementation in Javascript/Typescript but that is out of scope. Monaco exposes Api to tokenize the expression for all the languages we have registered to it. We are planning to use this to add tests around new top down tokenizer to make sure it is working correctly. The tests would include different type of expressions which would be tokenized using Monaco's API and asserted against expected token types. If the client tokenization goes out of sync then only syntax coloring is affected. The tokenization for any meaningful operations still happen at the server level. Overall usability of the editor is not affected with tokenizer going out of sync. Intellisense, signature help and other features are not affected by client level tokenization. The tests we plan to add would ensure correctness, but it still won't completely solve the problem as any changes in the C# implementation would require us to update the tests as well. How often PowerFx Lexer is updated? The client level tokenizer mostly focuses on simple tokens like numbers, boolean literals, string literals and anything else gets mostly identified as identifiers. How often do we change how basic tokens are recognized? I don't expect that too change that often (probably not all). Note that the list of token types that we want to skip during semantic tokenization would only involve these basic tokens which probably don't change at all.
 
2) This feature to skip certain token types is totally optional. You could always not choose to skip any token type and you will get all token types by default. If client level tokenizer starts to differ and we catch that, we would request language server to send all semantic token types and temporarily disable client level tokenizer until we fix it. That's y i allowed clients to control which tokens show up in the response of semantic tokens request from language server. 

1) & 2) I mentioned in #1 that the cleanest solution would be to implement JS version of powerfx but that it is out of scope. there's another viable better solution which would be less work compared to JS version of powerfx but it would still require large effort in both language server and client code of new formula bar. Language Server Protocol has a concept of exchanging capabilities between client and server when the communication between both is established. We could exchange a wide range of information with the client including the PowerFx lexer rules. The tokenizer in the client is declarative (bunch of regular expressions). This gives us an opportunity to create a declarative version of PowerFx Lexer (TexlLexer) and we can exchange that with the client when the communication is established and client then can setup the top down tokenizer with the declarative rules it got from the server. This basically solves both problems you pointed out above. But this would require a significant effort but something to consider.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
akshar2401: Yes. The idea is to not sort the tokens at all if comparer is not provided. The reason why i did it this way is language server doesn't need to spend probably nlogn time to sort the tokens as they will be encoded in the order based on the spans. The algorithm to encode goes from the smallest possible start index (0) to the biggest possible length (expression.Length - 1). Since we are iterating in the sorted order already, why spend time to sort it unnecessarily. If someone wants to use this function for other purposes, they can pass in a comparer that sorts based on start or end indexes. Assuming the number of tokens are <= expression.Length, assuming that C# sort is no better than nlogn, if we sort by spans and then encode the computational cost would be Sorting time = O(expression.Length * Log(expression.Length) + encoding time = O(expression.Length). If we just encode and given that we traverse expression from left to right, we already processing tokens in sorted order so computational cost would be O(expression.Length)

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
akshar2401: this is a typo. Thanks for catching it. It should be semantic tokenization request. I will elaborate more on this in this xml doc.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
akshar2401: Each token has a CanBeHidden property. CanBeHidden property is used in old formula bar to hide (when it is true) that token in the editor when editor is not in focus (user is doing something else). For example, Color.Blue becomes Blue when user leaves the editor and again becomes Color.Blue when user comes back to editor. This is mostly UI feature as logically the formula is still Color.Blue. CanBeHidden property is computed for tokens created from first name nodes.  This flag allowTokenHiding basically indicates whether we want to compute the value of CanBeHidden property or not. We don't plan to support hiding tokens like old formula bar in the new formula bar so we don't need to spend time computing the value of CanBeHidden property as we would not use this. However, since we plan to replace the similar tokenization logic that backs the old formula bar, i added this allowTokenHiding here so i could enable computing the value of CanBeHidden property for tokens extracted out of first name nodes. I will try to improve the phrasing

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/LanguageServerTests.cs
akshar2401: I wanted to test a scenario where client would pass in a token type that it wants to skip but the token type is not a valid one (a.k.a out of range). In this case, no tokens should be skipped as 78 is not one of the values of any token type (the max is 26). I can change this to TokenType.Lim which is also considered as not a valid token type. I compute a list of all possible token types in the range within a formula without skipping any token type and then i re compute the tokens in the same range this time with skipping the token types and then asserting against the list of possible token types in the range before skipping with the list of possible token types after skipping. For this case both list should be same as 78 doesn't represent a valid token type and therefore no tokens should be skipped.

akshar2401: felt like this test was a bit magic so made it a less magic by manually setting all possible expected token types in the range i am using for the test. but the expectation remains the same (78 or TokenType.Lim) are not considered valid token types and since there would be no token types with these two values, none of the expected token types should be skipped

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SemanticTokensEncoder.cs
```diff
@@ -0,0 +1,230 @@
...
+
+        /// <summary>
+        /// A constant to indicate how many slots each token would occupy in the encoded tokens collection.
+        /// </summary>
+        internal const int SlotsPerToken = 5;
+
+        /// <summary>
+        /// A mapping from token type to its unsigned interger encoded value. Each token type is encoded using its position in the TokenType enum.
+        /// </summary>
+        private static readonly IDictionary<TokenType, uint> EncodedTokenTypes = Enum.GetValues(typeof(TokenType))
```
lesaltzm, line 29: IReadOnlyDictionary?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
lesaltzm: Yes, in this case if you're using this as a Set, I'd suggest returning ISet (or even just HashSet, since it's a private method)

lesaltzm: Sadly, IReadOnlySet isn't in .netstandard2.0 :(

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SemanticTokensEncoder.cs
```diff
@@ -0,0 +1,230 @@
...
+        /// Usually this is the position of token in the list or enum of all token types. </para>
+        /// 
+        /// <para>5) Fifth integer is a bit mask representing what token modifiers are applied to each token. this is always zero as powerfx doesn't have token modifiers.</para>
+        /// </summary>
+        /// <param name="tokens">Collection of tokens that would be encoded.</param>
+        /// <param name="expression">Expression from which these tokens were extracted from.</param>
+        /// <param name="eol">End of line character indiciating the line breaks in the given expression.</param>
+        /// <returns>Encoded Tokens.</returns>
+        // This encoding is done according to LSP specification for semantic tokens methods https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_semanticTokens
+        public static ICollection<uint> EncodeTokens(IEnumerable<ITokenTextSpan> tokens, string expression, string eol)
```
lesaltzm, line 56: Why not IEnumerable?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SemanticTokensEncoder.cs
```diff
@@ -0,0 +1,230 @@
...
+                    tokensByStartIdxs.Add(token.StartIndex, token);
+                }
+            }
+
+            if (tokensByStartIdxs.Count == 0)
+            {
+                return new uint[0];
+            }
+
+            ICollection<uint> encodedTokens = new LinkedList<uint>();
```
lesaltzm, line 93: Why not just a List<uint>? (and no need for the interface in the declaration, it's a local. Just use var)

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SemanticTokensResponse.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using System.Collections.Generic;
+
+namespace Microsoft.PowerFx.LanguageServerProtocol.Protocol
+{
+    internal class SemanticTokensResponse
+    {
+        public string ResultId { get; set; } = null;
+
+        public ICollection<uint> Data { get; set; } = new uint[0];
```
lesaltzm, line 12: And again here. Why not uint[], IList<uint>, IEnumerable<uint> or any number of more common/idiomatic interfaces

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SemanticTokensEncoder.cs
akshar2401: makes sense! i was debating between list vs linkedlist. Each token gets converted to 5 values and since we need to add 5 uints for  each token to the end of the list, i thought linked list would be better to avoid internal resizing of List<T> when its gets full. Also given how List<T>s are implemented and we need to do more insertions towards the end, we could end up with about twice (or whatever the growth factor is) the extra space. i don't have a strong opinion. In the end we just serialize whatever sequential collection we return to json array and send it back to the client.  We are not doing any other operations with the returned collection of uints.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SemanticTokensResponse.cs
akshar2401: Some of my thoughts process behind this.
1) If i knew how many tokens i am going to skip or how many multiline tokens i am going to break down then i would have created a uint[] array in EncodeTokens() method and write it using a pointer index.
2) But i need to run one more loop to compute this size so i had to use a dynamic list (LinkedList/List). I went with LinkedList for faster appends towards the end as each token is mapped to 5 tokens. We just serialize this uint collection in the end. 
3) I could convert the LinkedList or List (whichever dynamic list i use) to an uint[] array and return type of EncodeTokens in this case would be uint[]. This array would then be serialized to json array by the JsonSerializer. The question i asked myself was do i need to convert the dynamic list i used before to an array and then serialize that array. I could just directly serialize the dynamic list i used to build tokens. No need to do take an extra step of converting to array if in the end i am just going to iterate over it from left to right and serialize it.  
4) My bad on this that i didn't spend time thinking about whether to return dynamic list as ICollection or IEnumerable. IEnumerable makes sense for what i described above in the first 3 points. I can use that instead of ICollection.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SemanticTokensEncoder.cs
akshar2401: Wasn't aware of the existence of this interface

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
akshar2401: Guess i am used to returning everything as ICollection. But agreed i should be careful with my return types.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SemanticTokensResponse.cs
akshar2401: I changed this to IEnumerable. Let me know if you have any other feedback around this

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SemanticTokensEncoder.cs
akshar2401: Changed to IEnumerable

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SemanticTokensEncoder.cs
akshar2401: Thank you for pointing this out. IReadOnlyDictionary make sense

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SemanticTokensResponse.cs
akshar2401: Resolving this. Let me know if you have any concerns about this

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
akshar2401: Resolving this. Let me know if you have any feedback around this

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -588,8 +589,21 @@ public FormulaType GetNodeType(TexlNode node)
...
+        /// </summary>
+        /// <param name="comparer">Optional comparer to sort tokens.</param>
+        /// <param name="allowTokenHiding">Optional: This flag determines whether to compute the value of a CanBeHidden property for tokens generated out of first names present in the expression.</param>
+        /// <returns> Enumerable of tokens. Tokens are ordered only if comparer is provided.</returns>
+        // allowTokenHiding flag is off by default as we don't want to compute hiddenness of the tokens for the new formula bar as we are not going to support that.
+        // However, we want to keep this for old formula bar as long as it exists and once we consume the changes in Canvas App backend,
+        // we would update the tokenization logic behind old formula bar to use this one instead and allowTokenHiding would be true for old formula bar.
+        internal IEnumerable<ITokenTextSpan> GetTokens(IComparer<ITokenTextSpan> comparer = null, bool allowTokenHiding = false) => Tokenization.Tokenize(_expression, _binding, Parse.Comments, comparer, allowTokenHiding);
```
MikeStall, line 605: >IComparer<ITokenTextSpan> comparer = null, bool allowTokenHiding = false [](http://example.com/codeflow?start=55&length=72)

consider removing the args

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Helpers/PositionRangeHelper.cs
```diff
@@ -0,0 +1,75 @@
...
+    {
+        /// <summary>
+        /// Converts the given range into coressponding start and end indexes.
+        /// </summary>
+        /// <param name="range">Range to convert to flat positions.</param>
+        /// <param name="expression">Expression from which range was created.</param>
+        /// <param name="eol">End of line character.</param>
+        /// <param name="oneBasedRange">Optional: Indicates whether range is one based or zero based.</param>
+        /// <returns>Start and end indexes.</returns>
+        public static (int startIndex, int endIndex) ConvertRangeToPositions(Range range, string expression, string eol, bool oneBasedRange = true)
```
MikeStall, line 21: >Range [](http://example.com/codeflow?start=77&length=5)

nit: extension method?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -415,6 +422,145 @@ private void HandleCodeActionRequest(string id, string paramsJson)
...
+        {
+            var tokenTypesToSkip = new HashSet<TokenType>();    
+            if (string.IsNullOrWhiteSpace(rawTokenTypesToSkipParam))
+            {
+                return tokenTypesToSkip;
+            }
+
+            try
+            {
+                var tokenTypesToSkipParam = JsonSerializer.Deserialize<List<int>>(rawTokenTypesToSkipParam);
```
MikeStall, line 521: > JsonSerializer.Deserialize<List<i [](http://example.com/codeflow?start=43&length=34)

use TryParseParams? and then avoid try/catch

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SemanticTokensEncoder.cs
```diff
@@ -0,0 +1,230 @@
...
+        private static readonly IReadOnlyDictionary<TokenType, uint> EncodedTokenTypes = Enum.GetValues(typeof(TokenType))
+                                                                                             .OfType<TokenType>()
+                                                                                             .Where(tokenType => tokenType == TokenType.Unknown || (tokenType != TokenType.Min && tokenType != TokenType.Lim))
+                                                                                             .Distinct()
+                                                                                             .Select((tokenType, index) => (tokenType, index))
+                                                                                             .ToDictionary((tokenTypeAndIndex) => tokenTypeAndIndex.tokenType, (tokenTypeAndIndex) => (uint)tokenTypeAndIndex.index);
+
+        /// <summary>
+        /// Encodes the given collection of tokens into an unsigned integer array. 
+        /// <para>Eaxh token is encoded to 5 integer values and therefore the size of the encoded array is 5 * number of given tokens.</para>
```
MikeStall, line 38: >Eaxh [](http://example.com/codeflow?start=18&length=4)

typo

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/SemanticTokensEncoderTests.cs
```diff
@@ -0,0 +1,349 @@
...
+            var encodedTokens = SemanticTokensEncoder.EncodeTokens(tokens, expression, eol);
+
+            // Assert
+            AssertEncodedTokens(encodedTokens, tokens, expression, eol, true);
+        }
+
+        // Skipping because of a bug in TexlLexer.cs which causes the spans to first comment and second comment tokens in the expression below to overlap
+        // Enable this test after fixing it in a separate user story
+        // For some reason, Skip attribute on InlineData for the TestMultilineTokensAreEncodedCorrectly test doesn't skip the tests in Visual Studio so adding a separate test
+        [Fact(Skip = "\"There's a bug in TexlLexer which causes the first two comment tokens to overlap. The end index of first comment is one more than (16) than the start of second comment (15). Fix it and then enable this test\"")]
```
MikeStall, line 167: >Skip = "\"There's a bug in TexlLexer w [](http://example.com/codeflow?start=14&length=38)

you can include comment to github issue

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Tokenization.cs
```diff
@@ -0,0 +1,324 @@
...
+        /// <param name="comments">Colllection of comment tokens extracted from the given expression.</param>
+        /// <param name="comparer">optional comparer to sort tokens.</param>
+        /// <param name="allowTokenHiding">Optional: This flag determines whether to compute the value of a CanBeHidden property for tokens generated out of first names present in the expression.</param>
+        /// <returns> Enumerable of tokens. Tokens are ordered only if comparer is provided.</returns>
+        // allowTokenHiding flag is off by default as we don't want to compute hiddenness of the tokens for the new formula bar as we are not going to support that.
+        // However, we want to keep this for old formula bar as long as it exists and once we consume the changes in Canvas App backend,
+        // we would update the tokenization logic behind old formula bar to use this one instead and allowTokenHiding would be true for old formula bar.
+        internal static IEnumerable<ITokenTextSpan> Tokenize(string expression, TexlBinding binding, IEnumerable<CommentToken> comments = null, IComparer<ITokenTextSpan> comparer = null, bool allowTokenHiding = false)
```
MikeStall, line 45: This is the key function that's providing semantic coloring.

MikeStall (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Tokenization.cs
akshar2401: Not sure if you want me to do anything here. Resolving this. Let me know if i missed anything

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/SemanticTokensEncoderTests.cs
akshar2401: Added link to the issue

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
akshar2401: Refactored this to use TryParseParams

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Helpers/PositionRangeHelper.cs
akshar2401: Changed this to extension method

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
akshar2401: Removed the args

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
akshar2401: Resolving this @MikeStall. Let me know if you want to discuss this further. any work to keeping them in sync is out of scope of this pr but this could be solved like i mentioned

MikeStall (approved):

src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalEntityScope.cs
```diff
@@ -18,6 +18,15 @@ internal interface IExternalEntityScope
 
         bool TryGetEntity<T>(DName currentEntityEntityName, out T externalEntity)
             where T : class, IExternalEntity;
+
+        /// <summary>
```
akshar2401, line 22: Dropping a comment to prevent auto merging

MikeStall (approved):

jas-valgotar (approved):

LucGenetier:
✅ No public API change.

PR #1422, opened 2023-04-21
Update Time function to stop 'wrapping around' values beyond 24h

CarlosFigueiraMSFT:
Currently the Time function wraps around values beyond 24 hours: Time(36,0,0) is the same value as Time(12,0,0) (and Value(Time(36,0,0)) == Value(Time(12,0,0)) == 0.5). This doesn't happen on negative values, though: Time(-6,0,0) and Time(-30,0,0) are not the same value (their Values are -0.25 and -1.25, respectively). This change makes the function consistent - Time(36,0,0) will not wrap around and will be a different value than Time(12,0,0), with values of 1.5 and 0.5, respectively.


LucGenetier:
✅ No public API change.

MikeStall (approved):

LucGenetier:
✅ No public API change.

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1432, opened 2023-04-24
Add Format/Language arguments to Text_UO

bryancroteau-MSFT:
These arguments are currently not used in any way by the C# interpreter. When the changes are incorporated into PAClient, new tests will need to be added.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUntypedObject.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUntypedObject.cs
@@ -582,24 +582,13 @@ namespace Microsoft.PowerFx.Functions
         {
             var impl = args[0].Impl;
 
-            if (impl.Type == FormulaType.String || (impl.Type is ExternalType et && et.Kind == ExternalTypeKind.UntypedNumber))
+            if (impl.Type is ExternalType et && et.Kind == ExternalTypeKind.UntypedNumber)
             {
                 var valueRes = Value_UO(runner, context, IRContext.NotInSource(numberType), new UntypedObjectValue[] { args[0] });
 
                 if (valueRes is T nv)
                 {
                     return new UntypedObjectValue(IRContext.NotInSource(FormulaType.UntypedObject), constructUO(nv));
-                }
-
-                var dateTimeRes = DateTimeValue_UO(runner, context, IRContext.NotInSource(FormulaType.DateTime), new UntypedObjectValue[] { args[0] });
-
-                if (dateTimeRes is not ErrorValue)
-                {
-                    var valueRes2 = Value(runner, context, IRContext.NotInSource(numberType), new FormulaValue[] { dateTimeRes });
-                    if (valueRes2 is T nv2)
-                    {
-                        return new UntypedObjectValue(IRContext.NotInSource(FormulaType.UntypedObject), constructUO(nv2));
-                    }
                 }
             }
 
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUntypedObject.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUntypedObject.cs
@@ -306,40 +306,12 @@ namespace Microsoft.PowerFx.Functions
             if (impl.Type == FormulaType.String)
             {
                 var str = impl.GetString();
-                var stringArg0 = new StringValue(irContext, str);
-                newArg0 = stringArg0;
-
-                // The second argument to Text is a format, if it exists then we need to convert the string to a type that
-                // can be formatted. We try to convert to number first and if that fails we try dates.
-                if (args.Length > 1)
-                {
-                    var numArg0 = Value_UO(runner, context, IRContext.NotInSource(FormulaType.Number), new UntypedObjectValue[] { arg0 });
-                    if (numArg0 is not ErrorValue)
-                    {
-                        newArg0 = numArg0;
-                    }
-                    else
-                    {
-                        // Use DateTimeValue_UO to apply format restrictions
-                        var dateArg0 = DateTimeValue_UO(runner, context, IRContext.NotInSource(FormulaType.DateTime), new UntypedObjectValue[] { arg0 });
-                        if (dateArg0 is not ErrorValue)
-                        {
-                            newArg0 = dateArg0;
-                        }
-                    }
-                }
+                newArg0 = new StringValue(irContext, str);
             }
             else if (impl.Type is ExternalType et && et.Kind == ExternalTypeKind.UntypedNumber)
             {
                 var str = impl.GetUntypedNumber();
                 newArg0 = new StringValue(irContext, str);
-
-                // If a second argument is provided, then we cannot guarantee that the original precison will be
-                // preserved. We convert to a number, either Float or Decimal depending on the settings.
-                if (args.Length > 1)
-                {
-                    newArg0 = Value_UO(runner, context, IRContext.NotInSource(FormulaType.Number), new UntypedObjectValue[] { arg0 });
-                }
             }
             else if (impl.Type == FormulaType.Number)
             {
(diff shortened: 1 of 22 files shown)


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

gregli-msft:
I guess we never did the locale specific version of Value_UO?  Seems like if the third argument to Text is provided, that should be funneled down to Value and Value_UO for the first argument.   

We may want to block the third argument from being used if the first argument is a UO until we do.  

The maker can always call Value themselves, as shown here, to accomplish this and transition manually from a UO to a number.

For illustration, the difference between these two lines:

```
> Text(Value("1,234","fr-fr"),"","fr-fr")
"1,234"

> Text(Value("1,234"),"","fr-fr")
"1234"
```

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUntypedObject.cs
```diff
@@ -297,37 +297,71 @@ public static FormulaValue Float_UO(EvalVisitor runner, EvalVisitorContext conte
...
 
-        public static FormulaValue Text_UO(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, UntypedObjectValue[] args)
-        {
-            var impl = args[0].Impl;
-
+        public static FormulaValue Text_UO(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
+        {
+            var arg0 = (UntypedObjectValue)args[0];
+            var impl = arg0.Impl;
+
```
gregli-msft, line 304: What happens with the third arg to Text?

```
> Text(Value("1,234","fr-fr"),"","fr-fr")
"1,234"

> Text(Value("1,234"),"","fr-fr")
"1234"
```

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUntypedObject.cs
```diff
@@ -297,37 +297,71 @@ public static FormulaValue Float_UO(EvalVisitor runner, EvalVisitorContext conte
...
+                }
             }
             else if (impl.Type is ExternalType et && et.Kind == ExternalTypeKind.UntypedNumber)
             {
                 var str = impl.GetUntypedNumber();
-                return new StringValue(irContext, str);
+                newArg0 = new StringValue(irContext, str);
+
+                // If a second argument is provided, then we cannot guarantee that the original precison will be
+                // preserved. We convert to a number, either Float or Decimal depending on the settings.
```
gregli-msft, line 338: Unfortunately, it won't be Float or Decimal, as written it will always be Float as FormulaType.Number was used to call Value_UO on line 341.  To this point, the interpreter hasn't needed to know if NumberIsFloat was being used and so it isn't passed in.  That's in general a good thing.  Value_UO figures it out because CheckTypes for Value_UO provides the return type through the irContext.  But there is no return type in this case (besides String for the Text function).  Is it time for the intepreter to know more directly?  Include it as ambient state in the irContext?  Thiking about it...  Some of our hosts, like PVA, are heavily Decimal, I don't think we should favor Float just because of this information gap.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ParseJson.txt
```diff
@@ -157,7 +157,7 @@ Table({Value:Blank()},{Value:Blank()},{Value:Blank()})
 Table({Value:5},{Value:Blank()},{Value:7})
 
 >> ForAll(ParseJSON("[""hi"", null, ""hello""]"), Text(ThisRecord))
-Table({Value:"hi"},{Value:Blank()},{Value:"hello"})
+Table({Value:"hi"},{Value:""},{Value:"hello"})
```
CarlosFigueiraMSFT, line 160: Conversion functions (such as Text) should preserve blank/null values

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUntypedObject.cs
bryancroteau-MSFT: I removed support for the 3rd argument.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUntypedObject.cs
bryancroteau-MSFT: The argument is now pre-processed. Code is injected to either convert to float or decimal in IRTranslator.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ParseJson.txt
bryancroteau-MSFT: @gregli-msft is this true? In our chat, I remember that we talked about making this more consistent with the existing Text function which maps Blank() to ""

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ParseJson.txt
```diff
@@ -490,4 +490,10 @@ Blank()
...
 >> Boolean(IfError(ParseJSON("{""field1"":true}").field1, ParseJSON("false")))
-true
\ No newline at end of file
+true
+
+>> Text(ParseJSON("1.234"), "#.##")
+"1.23"
+
+>> Text(ParseJSON("1.234"), "#.##", "fr-fr")
+"1,23"
```
CarlosFigueiraMSFT, line 499: Add some tests with blank values and 2+ arguments:

    >> Text(ParseJSON("null"), "0.00")
    ""

    >> Text(ParseJSON("null"), "0.00", "de-DE")
    ""

bryancroteau-MSFT: Done

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUntypedObject.cs
```diff
@@ -297,37 +297,64 @@ public static FormulaValue Float_UO(EvalVisitor runner, EvalVisitorContext conte
...
-            var impl = args[0].Impl;
-
+        public static FormulaValue Text_UO(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
+        {
+            if ((args[0] is UntypedObjectValue uo && uo.Type == FormulaType.Blank) || args[0] is BlankValue)
+            {
+                if (args.Length == 1)
+                {
+                    // As a special case, blank propagates for the single argument only
+                    return args[0];
```
CarlosFigueiraMSFT, line 307: >return args[0]; [](http://example.com/codeflow?start=20&length=15)

We should return a blank value - which won't be the case on `Text(ParseJSON("null"))` - it will be returning an UntypedObjectValue

CarlosFigueiraMSFT, line 307: >return args[0]; [](http://example.com/codeflow?start=20&length=15)

If args[0] is an UntypedObjectValue, should we return it or just return a new BlankValue? I don't think Text_UO should return an untyped object.

bryancroteau-MSFT: This has been fixed.

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1433, opened 2023-04-25
Fix negation of Date/Time values - IR option

CarlosFigueiraMSFT:
This is a small fix to the issue of negating date/time values (which currently doesn't work - see #1294). The proper fix would be to introduce a new unary node (so that it can be reused in Power Apps), but this fixes the immediate problem.


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DateTime.txt
```diff
@@ -94,3 +94,14 @@ DateTime(2022,1,3,12,0,0,0)
...
+
+// Negation
+>> -DateTime(1900,1,1,3,0,0)
+DateTime(1899,12,27,21,0,0,0)
+
+>> Value(-DateTime(1900,1,1,3,0,0))
+-2.125
+
+// Negation may lose precision as DateTime values have smaller precision than numbers
+>> With({n:Now()}, Math.Abs(Value(n) + Value(-n)) < 0.000001)
```
CarlosFigueiraMSFT, line 106: >Math [](http://example.com/codeflow?start=19&length=4)

This should be removed, and the result should be `true`

CarlosFigueiraMSFT:
#1443 has a better fix for this issue

PR #1437, opened 2023-04-25
InternalsVisibleTo for new test projects

lesaltzm:
We're adding some (temporary) test projects on the PA Client side. These will be cleaned up in a few months, however, for now, they need InternalsVisibleTo


MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1439, opened 2023-04-25
Added argument index to UDF params

VamsiModem:
Added arg index to udf params, we need this during code gen on the host.
Created a new info type for UDF parameters.
Modified UserDefinitionsNameResolver to not use record type, since I didn't find a way to get the index of a pram.


src/libraries/Microsoft.PowerFx.Core/Binding/BindInfo/UDFParameterInfo.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.Utils;
+
+namespace Microsoft.PowerFx.Core.Binding.BindInfo
+{
+    internal sealed class UDFParameterInfo
```
lesaltzm, line 9: >UDFParameterInfo [](http://example.com/codeflow?start=26&length=16)

nit: Doc comment about what this is for

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -34,6 +34,26 @@ internal class UserDefinedFunction : TexlFunction
...
+            if (string.IsNullOrEmpty(argName))
+            {
+                return argIndex;
+            }
+
+            foreach (var arg in _args)
+            {
+                if (arg.VarIdent.Name.Value == argName)
+                {
+                    argIndex = arg.ArgIndex;
```
lesaltzm, line 50: >argIndex = arg.ArgIndex [](http://example.com/codeflow?start=20&length=23)

return?

Could there be multiple args with the same name?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -34,6 +34,26 @@ internal class UserDefinedFunction : TexlFunction
 
         public TexlNode UdfBody { get; }
 
+        public int GetArgIndex(string argName)
+        {
```
lesaltzm, line 38: organization nit: move after the constructor and probably after the main functionality (bind/checktypes)

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -101,18 +121,27 @@ public override IEnumerable<StringGetter[]> GetSignatures()
         private partial class UserDefinitionsNameResolver : INameResolver
         {
             private readonly INameResolver _globalNameResolver;
-            private readonly INameResolver _parameterNamedResolver;
+            private readonly Dictionary<string, UDFArg> _args;
```
lesaltzm, line 124: >Dictionary [](http://example.com/codeflow?start=29&length=10)

IReadOnlyDictionary?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
VamsiModem: There can be, but rule will be erroneous, in that case we won't even get here.

lesaltzm:
>                 }

When would this occur? Should this be an error?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs:139 in 64e7d6e. [](commit_id = 64e7d6e276de357a62c50204f83f96e29582f63b, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
lesaltzm: great then we can return early when the condition is true

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
lesaltzm: When do you need to Add to it? Looks like it's just in the constructor. 

Couldn't that be:
```
var argsByName = new Dictionary<string, UDFArg>();
foreach (UDFArg arg in args)
{
    if (!argsByName.ContainsKey(arg.VarIdent.Name.Value))
    {
        argsByName.Add(arg.VarIdent.Name.Value, arg);
    }
}

this._args = argsByName;
```

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -101,18 +121,27 @@ public override IEnumerable<StringGetter[]> GetSignatures()
...
-            private UserDefinitionsNameResolver(INameResolver globalNameResolver, INameResolver parameterNamedResolver, INameResolver functionNameResolver = null)
+            private UserDefinitionsNameResolver(INameResolver globalNameResolver, ISet<UDFArg> args, INameResolver functionNameResolver = null)
             {
                 this._globalNameResolver = globalNameResolver;
-                this._parameterNamedResolver = parameterNamedResolver;
+                this._args = new Dictionary<string, UDFArg>();
+
+                foreach (UDFArg arg in args)
+                {
+                    if (!this._args.ContainsKey(arg.VarIdent.Name.Value))
```
lesaltzm, line 139: Should this mark it as erroneous if there are colliding keys?

lesaltzm: or where should that be handled.

VamsiModem:
> > ```
> >             }
> > ```
> 
> When would this occur? Should this be an error?
> 
> Refers to: src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs:139 in [64e7d6e](https://github.com/microsoft/Power-Fx/commit/64e7d6e276de357a62c50204f83f96e29582f63b). [](commit_id = [64e7d6e](https://github.com/microsoft/Power-Fx/commit/64e7d6e276de357a62c50204f83f96e29582f63b), deletion_comment = False)

TokEat() will add an error if current token is not a colon

VamsiModem:
> > > ```
> > >             }
> > > ```
> > 
> > 
> > When would this occur? Should this be an error?
> > Refers to: src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs:139 in [64e7d6e](https://github.com/microsoft/Power-Fx/commit/64e7d6e276de357a62c50204f83f96e29582f63b). [](commit_id = [64e7d6e](https://github.com/microsoft/Power-Fx/commit/64e7d6e276de357a62c50204f83f96e29582f63b), deletion_comment = False)
> 
> This will occur if we have a UDF like Add(a, b:Number):Number = a + b;
 colon is missing after a, TokEat() will add an error if current token is not a colon

VamsiModem:
> > > ```
> > >             }
> > > ```
> > 
> > 
> > When would this occur? Should this be an error?
> > Refers to: src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs:139 in [64e7d6e](https://github.com/microsoft/Power-Fx/commit/64e7d6e276de357a62c50204f83f96e29582f63b). [](commit_id = [64e7d6e](https://github.com/microsoft/Power-Fx/commit/64e7d6e276de357a62c50204f83f96e29582f63b), deletion_comment = False)
> 
> This will occur if we have a UDF like Add(a: Number. b:Number) = a + b;
 cTokEat() will add an error if current token is not a colon

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
VamsiModem: Done!

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
lesaltzm: Then why check ContainsKey at all? 
Why not just
`this._args = args.ToDictionary(arg => arg.VarIdent.Name.Value, arg => arg)`

VamsiModem: By handling I meant, we see that there is a duplicate and add an error on the node but we don't remove that from the args array.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -47,6 +47,29 @@ public UserDefinedFunction(string name, IdentToken returnTypeToken, TexlNode bod
             this.UdfBody = body;
         }
 
+        /// <summary>
+        /// Gets the index of the given parameter.
+        /// </summary>
+        /// <param name="argName">Parameter name. </param>
+        /// <returns>Returns index >= 0 on a valid param, else -1. </returns>
+        public int GetArgIndex(string argName)
```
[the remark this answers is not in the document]
VamsiModem: I like this approach!

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
```diff
@@ -99,18 +99,20 @@ private IEnumerable<UserDefinedFunction> CreateUserDefinedFunctions(IEnumerable<
...
                 var udfName = udf.Ident.Name;
-                var func = new UserDefinedFunction(udfName.Value, udf.ReturnType, udf.Body, udf.IsImperative, udf.Args, paramsRecordType);
-
                 if (texlFunctionSet.AnyWithName(udfName) || BuiltinFunctionsCore._library.AnyWithName(udfName))
                 {
                     errors.Add(new TexlError(udf.Ident, DocumentErrorSeverity.Severe, TexlStrings.ErrUDF_FunctionAlreadyDefined, udfName));
                     continue;
                 }
 
+                if (!(CheckParameters(udf.Args, errors) & CheckReturnType(udf.ReturnType, errors)))
```
VamsiModem, line 109: '&' to evaluate both CheckParameters() and CheckReturnType() so we can add errors on the nodes

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

lesaltzm:
Add tests, and then I'll sign off!

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Parser/ParseFormulasResult.cs
```diff
@@ -94,12 +94,14 @@ public UDF(IdentToken ident, IdentToken returnType, HashSet<UDFArg> args, TexlNo
     internal class UDFArg
     {
         internal IdentToken VarIdent;
-        internal IdentToken VarType;
+        internal IdentToken VarType;
```
lesaltzm, line 97: naming nit: why start with Var?

Also, both are Idents. Maybe instead call these NameIdent and TypeIdent?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -60,12 +84,12 @@ public TexlBinding BindBody(INameResolver nameResolver, IBinderGlue documentBind
             }
 
             bindingConfig = bindingConfig ?? new BindingConfig(this._isImperative);
-            var binding = TexlBinding.Run(documentBinderGlue, UdfBody, UserDefinitionsNameResolver.Merge(nameResolver, _parametersRecordType, functionNameResolver), bindingConfig, features: features);
+            _binding = TexlBinding.Run(documentBinderGlue, UdfBody, UserDefinitionsNameResolver.Create(nameResolver, _args, functionNameResolver), bindingConfig, features: features, rule: rule);
```
lesaltzm, line 87: Is there a reason we're doing this with a NameResolver, rather than treating the args as a ruleScope? Either way works, just curious.

VamsiModem: No reason, didn't know about the rule scope, happy to look into that, if that's a better way to go

lesaltzm (approved):

LucGenetier:
✅ No public API change.

PR #1443, opened 2023-04-26
Fix negation of date/time values

### LINKED ISSUE #1294: Null object reference for negate of date/datetime/time

CarlosFigueiraMSFT:
Fixes #1294 - add support for negation of date/time/datetime values.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
```diff
@@ -32,14 +32,14 @@ internal static partial class Library
             },
             {
                 UnaryOpKind.NegateDecimal,
-                StandardErrorHandling<DecimalValue>(
+                StandardErrorHandling<FormulaValue>(
                     "-",
                     expandArguments: NoArgExpansion,
                     replaceBlankValues: ReplaceBlankWithDecimalZero,
-                    checkRuntimeTypes: ExactValueType<DecimalValue>,
+                    checkRuntimeTypes: DateNumberTimeOrDateTime,
```
lesaltzm, line 39: Any reason to do this at runtime rather than pick between them in the IRTranslator? We know statically that we're negating a number vs a date, vs a time. 
Adding UnaryOpKinds for each of those would likely be cleaner, and would allow backends like SQL to handle them more easily and check that they're handling everything they need to.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
CarlosFigueiraMSFT: Updated it to use new unary op kinds

LucGenetier:
✅ No public API change.

lesaltzm (approved):

PR #1445, opened 2023-04-26
Non row-scope pure sync nodes should always be allowed.

jelopezf:
Rules that use binary op nodes for non delegable ops are marked as invalid even if both sides of the op are literals (can be delegable since op can be handled by runtime before the delegated call is made)

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/OpDelegationStrategy.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/OpDelegationStrategy.cs
@@ -111,6 +111,12 @@ namespace Microsoft.PowerFx.Core.Functions.Delegation.DelegationStrategies
                 return true;
             }
 
+            // Non row-scope, non async, pure nodes should always be valid because we can calculate value in runtime before delegation.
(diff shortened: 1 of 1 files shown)


src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/OpDelegationStrategy.cs
```diff
@@ -111,6 +111,12 @@ private bool IsSupportedNode(TexlNode node, OperationCapabilityMetadata metadata
                 return true;
             }
 
+            // Non row-scope, non async, pure nodes should always be valid because we can calculate value in runtime before delegation.
+            if (!binding.IsRowScope(node) && !binding.IsAsync(node) && binding.IsPure(node))
+            {
+                return true;
+            }
```
Shpakh-MSFT, line 118: Please add test coverage for this when you bring this change to Power Apps.

jelopezf: will do

Shpakh-MSFT (approved):
:shipit:

PR #1446, opened 2023-04-26
Non row-scope pure sync nodes should always be allowed by jelopezf 

Shpakh-MSFT:
Helping Jesus to port his change.
https://github.com/microsoft/Power-Fx/pull/1445


MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1447, opened 2023-04-26
Renamed HasErrors to HasErrorsOrWarnings.

Shpakh-MSFT:
HasErrors property name is misleading as it also returns true if there are only warnings on the rule. Hence renaming it to increase clarity. Recently we have seen a bug which was caused due to this misunderstanding.


lesaltzm (approved):

LucGenetier:
✅ No public API change.

PR #1448, opened 2023-04-26
Avoid duplicated compilation error messages

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/1435.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
@@ -289,7 +289,7 @@ namespace Microsoft.PowerFx
         {
             culture ??= _defaultErrorCulture ?? ParserCultureInfo;
 
-            foreach (var error in this._errors)
+            foreach (var error in this._errors.Distinct(new ExpressionErrorComparer()))
             {
                 yield return error.GetInLocale(culture);
             }
--- a/src/libraries/Microsoft.PowerFx.Core/Public/ExpressionError.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/ExpressionError.cs
@@ -179,11 +179,21 @@ namespace Microsoft.PowerFx
         // We compare only Message
         public override bool Equals(ExpressionError error1, ExpressionError error2)
         {
+            if (error1 == null && error2 == null)
+            {
+                return true;
+            }
+
+            if (error1 == null || error2 == null)
+            {
+                return false;
+            }
+
             return error1.ToString() == error2.ToString();
         }
 
         public override int GetHashCode(ExpressionError error)
-        {
+        {            
             return error.ToString().GetHashCode();
         }
     }
--- a/src/libraries/Microsoft.PowerFx.Core/Public/ExpressionError.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/ExpressionError.cs
@@ -171,6 +171,9 @@ namespace Microsoft.PowerFx
         }
     }
 
+    /// <summary>
+    /// Used to compare CheckResult.Errors and avoid duplicates.
+    /// </summary>
     internal class ExpressionErrorComparer : EqualityComparer<ExpressionError>
     {
         // We compare only Message
--- a/src/libraries/Microsoft.PowerFx.Core/Public/ExpressionError.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/ExpressionError.cs
@@ -184,7 +184,7 @@ namespace Microsoft.PowerFx
 
         public override int GetHashCode(ExpressionError error)
         {
-            return 0;
+            return error.ToString().GetHashCode();
         }
     }
 }
--- a/src/libraries/Microsoft.PowerFx.Core/Public/ExpressionError.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/ExpressionError.cs
@@ -170,4 +170,18 @@ namespace Microsoft.PowerFx
             }
         }
     }
+
+    internal class ExpressionErrorComparer : EqualityComparer<ExpressionError>
+    {
+        // We compare only Message
+        public override bool Equals(ExpressionError error1, ExpressionError error2)
+        {
+            return error1.ToString() == error2.ToString();
+        }
+
+        public override int GetHashCode(ExpressionError error)
+        {
+            return 0;
+        }
+    }
 }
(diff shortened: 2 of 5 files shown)


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Error.txt
```diff
@@ -117,3 +117,7 @@ Errors: Error 0-10: 'Wyz' is an unknown or unsupported function.
 
 >> Error(Table({Kind:ErrorKind.Div0},{Kind:ErrorKind.Numeric}))
 Error(Table({Kind:ErrorKind.Div0},{Kind:ErrorKind.Numeric}))
+
+// No duplicated errors
+>> 1.2.3
```
jas-valgotar, line 122: > 1.2.3 [](http://example.com/codeflow?start=2&length=6)

Can we test expression which has same error message but at different locations?
e.g. `1.2.3 + 1.2.3`
This should saw 2 error messages!

src/libraries/Microsoft.PowerFx.Core/Public/ExpressionError.cs
```diff
@@ -170,4 +170,18 @@ internal static IEnumerable<ExpressionError> New(IEnumerable<IDocumentError> err
             }
         }
     }
+
+    internal class ExpressionErrorComparer : EqualityComparer<ExpressionError>
+    {
+        // We compare only Message
```
jas-valgotar, line 176: >// We compare only Message [](http://example.com/codeflow?start=8&length=26)

What would happen if there are 2 same messages but at different spans?

src/libraries/Microsoft.PowerFx.Core/Public/ExpressionError.cs
```diff
@@ -170,4 +170,18 @@ internal static IEnumerable<ExpressionError> New(IEnumerable<IDocumentError> err
...
+        public override bool Equals(ExpressionError error1, ExpressionError error2)
+        {
+            return error1.ToString() == error2.ToString();
+        }
+
+        public override int GetHashCode(ExpressionError error)
+        {
+            return 0;
```
jas-valgotar, line 184: > return 0; [](http://example.com/codeflow?start=11&length=10)

Should either thow or generate unique hash code.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -289,7 +289,7 @@ public IEnumerable<ExpressionError> GetErrorsInLocale(CultureInfo culture)
         {
             culture ??= _defaultErrorCulture ?? ParserCultureInfo;
 
-            foreach (var error in this._errors)
+            foreach (var error in this._errors.Distinct(new ExpressionErrorComparer()))
```
jas-valgotar, line 292: >this._errors.Distinct(new ExpressionErrorComparer())) [](http://example.com/codeflow?start=34&length=53)

What is the reason we get duplicate error? Is that fixable?

src/libraries/Microsoft.PowerFx.Core/Public/ExpressionError.cs
anderson-joyle: The messages are considered different since they happened at different locations.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
anderson-joyle: One from the parser and one from the binder. This is the most straightforward and less invasive fix for the issue.

src/libraries/Microsoft.PowerFx.Core/Public/ExpressionError.cs
anderson-joyle: This class is specific for comparing error messages. We could create a new one in case we need to work with `Dict` and substitute this one.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Error.txt
anderson-joyle: Added. It results in two equal messages, with different spans.

src/libraries/Microsoft.PowerFx.Core/Public/ExpressionError.cs
jas-valgotar: I feel we should probably return `error.ToString().GetHashCode()` or throw not implemented here...

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

jas-valgotar (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
CarlosFigueiraMSFT: Should we go to the binder if the parser fail?

src/libraries/Microsoft.PowerFx.Core/Public/ExpressionError.cs
```diff
@@ -170,4 +170,21 @@ internal static IEnumerable<ExpressionError> New(IEnumerable<IDocumentError> err
...
+
+    /// <summary>
+    /// Used to compare CheckResult.Errors and avoid duplicates.
+    /// </summary>
+    internal class ExpressionErrorComparer : EqualityComparer<ExpressionError>
+    {
+        // We compare only Message
+        public override bool Equals(ExpressionError error1, ExpressionError error2)
+        {
+            return error1.ToString() == error2.ToString();
```
LucGenetier, line 182: >error1.ToString() == error2.ToString(); [](http://example.com/codeflow?start=19&length=39)

if error1 and/or error2 is/are null, this will fail
consider adding 

if (error1 == null && error2 == null)
return true;

if (error1 ==null || error2 == null)
return false;

then return error1.ToString...

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionErrorTests.cs
```diff
@@ -114,6 +114,21 @@ public void Localize()
             Assert.Equal("Името не е валидно. „name“ не е разпознато.", errorBg.Message);
         }
 
+        [Fact]
+        public void NoDuplicatedErrorsTest()
+        {
+            var engine = new Engine();
+            var parseOpt = new ParserOptions() { Culture = CultureInfo.CreateSpecificCulture("fr-FR") };
```
LucGenetier, line 121: >fr-FR [](http://example.com/codeflow?start=94&length=5)

Merci :)

LucGenetier (approved):
:shipit:

LucGenetier (approved):
:shipit:

LucGenetier:
✅ No public API change.

Issue #1301, opened 2023-04-03
Blank comparison with Blank results in internal "specified method is not supported" error message.


gregli-msft:
```
> Blank()=Blank()
true

> Blank()<>Blank()
false

> Blank()<Blank()
One or more errors occurred. (Specified method is not supported.)

> Blank()<=Blank()
One or more errors occurred. (Specified method is not supported.)

> Blank()>Blank()
One or more errors occurred. (Specified method is not supported.)

> Blank()>=Blank()
One or more errors occurred. (Specified method is not supported.)

> IR(Blank()>Blank())
Specified method is not supported.
```

PR #1451, opened 2023-04-28
Comparison with blank crash

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/1301.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/IR/BinaryOpMatrix.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/IR/BinaryOpMatrix.cs
@@ -272,8 +272,12 @@ namespace Microsoft.PowerFx.Core.IR
                     switch (node.Op)
                     {
                         case BinaryOp.NotEqual:
+                        case BinaryOp.Less:
+                        case BinaryOp.Greater:
                             return BinaryOpKind.NeqNull;
                         case BinaryOp.Equal:
+                        case BinaryOp.LessEqual:
+                        case BinaryOp.GreaterEqual:
                             return BinaryOpKind.EqNull;
                         default:
                             throw new NotSupportedException();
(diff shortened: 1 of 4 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/IR/BinaryOpMatrix.cs
```diff
@@ -271,9 +271,13 @@ private static BinaryOpKind GetBooleanBinaryOp(PowerFx.Syntax.BinaryOpNode node,
                 case DKind.ObjNull:
                     switch (node.Op)
                     {
-                        case BinaryOp.NotEqual:
+                        case BinaryOp.NotEqual:
+                        case BinaryOp.Less:
+                        case BinaryOp.Greater:
```
CarlosFigueiraMSFT, line 276: It feels wrong to use NeqNull for `<` and `>` (and EqNull for `>=` and `<=`). Should we introduce new op kinds?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -349,6 +349,8 @@ private ValueTask<FormulaValue> VisitBinaryOpNode(BinaryOpNode node, EvalVisitor
                 case BinaryOpKind.EqDecimals:
                     return OperatorBinaryEq(this, context, node.IRContext, args);
                 case BinaryOpKind.EqNullUntyped:
+                case BinaryOpKind.GeqNull:
```
CarlosFigueiraMSFT, line 352: >case BinaryOpKind.GeqNull: [](http://example.com/codeflow?start=16&length=26)

OperatorBinaryGeq?

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

Issue #1356, opened 2023-04-12
Support only Decimal and Number types when specifying the return type for the field requiring Number


justinpark0916:
Cards has a property that requires Number/Decimal. Currently, it seems like allowing multiple types is only allowed in `FormulaType.String` case. 

We would need a feature for `FormulaType.Number` so that this supports both `FormulaType.Number` and `FormulaType.Decimal`

MikeStall:
Related to #1357, #1125. May need to get more details here.

MikeStall:
This applies to  CheckResult.SetExpectedReturnValue().

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/1453

nguhoa:
CARDS would want coercion to allow both decimal and number. 
in a number field, maybe Float(1.22) and Decimal(1.22) should allowed Also 1 and 1.2 should be allowed

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/1500

PR #1453, opened 2023-04-28
Add decimal/number return type to checkresult

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1356


src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -447,14 +447,30 @@ internal TexlBinding ApplyBindingInternal()
...
+                        {
+                            if (!DecimalValue.AllowedListConvertToDecimal.Contains(this.ReturnType))
+                            {
+                                notCoerceToType = true;
+                            }
+                        }
+                        else
+                        {
+                            throw new NotImplementedException();
                         }
```
jas-valgotar, line 474: I would write switch case for readability, something like

switch (this._expectedReturnType)
{
    case FormulaType.String:
        notCoerceToType = !StringValue.AllowedListConvertToString.Contains(this.ReturnType);
        break;
    case FormulaType.Number:
        notCoerceToType = !NumberValue.AllowedListConvertToNumber.Contains(this.ReturnType);
        break;
    case FormulaType.Decimal:
        notCoerceToType = !DecimalValue.AllowedListConvertToDecimal.Contains(this.ReturnType);
        break;
    default:
        throw new NotImplementedException();
}

nguhoa: Updated this. thanks

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/CheckResultTests.cs
```diff
@@ -262,70 +262,36 @@ public void BindingChangeSymbols()
...
+        public void CheckResultExpectedReturnValueString(string inputExpr, bool allowCoerceTo, bool isSuccess, string errorMsg)
+        {
+            CheckResultExpectedReturnValue(inputExpr, allowCoerceTo, isSuccess, errorMsg, FormulaType.String);
         }
 
         [Theory]
         [InlineData("12", false, true, "")]
         [InlineData("12", true, true, "")]
-        [InlineData("\"test string\"", true, false, "The method or operation is not implemented")]
-        [InlineData("\"test string\"", false, false, "The type of this expression does not match the expected type 'Number'")]
```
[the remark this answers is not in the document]
nguhoa: we now allow numeric return type so we allow string convert to number. Remove these test to have clear test for right string to number or non support type (record) to number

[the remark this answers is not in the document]
jas-valgotar: Can we keep the test and change the `isSuccess` value?
I wouldn't want to loose the test coverage from existing tests...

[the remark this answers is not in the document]
nguhoa: This test is wrong test anyway; [InlineData("\"test string\"", true, false, "The method or operation is not implemented")] where this string cannot coerce to number but I set allowtocoerceto is true. This test need to remove. It passed last time because we did not support it. But now we support number and it will show the failure. I don't think we should keep this test.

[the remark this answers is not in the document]
jas-valgotar: I think we should keep the test and change it to below, 
Here's why,
First 2 test checks for number, which would always return number type, since input is number type.
Last 2 test checks for record type, which can never be coerced to number type.
The below 2 test would check for something that isn't number type and can be coerced to number type.
`[InlineData("\"test string\"", true, true, "")]`
`[InlineData("\"test string\"", false, false, "The type of this expression does not match the expected type 'Number'")]`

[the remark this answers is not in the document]
nguhoa: i will keep it then. thanks

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -447,14 +447,19 @@ internal TexlBinding ApplyBindingInternal()
...
+                                notCoerceToType = !StringValue.AllowedListConvertToString.Contains(this.ReturnType);
+                                break;
+                            case NumberType:
+                                notCoerceToType = !NumberValue.AllowedListConvertToNumber.Contains(this.ReturnType);
+                                break;
+                            case DecimalType:
+                                notCoerceToType = !DecimalValue.AllowedListConvertToDecimal.Contains(this.ReturnType);
+                                break;
+                            default:
+                                throw new NotImplementedException();
```
MikeStall, line 462: >NotImplementedException [](http://example.com/codeflow?start=42&length=23)

nit: for debuggability, can we update this message to include the type? 
$"Setting ExpectedReturnType to {_expectedReturnType.GetType().FullName} is not implemented";

src/tests/Microsoft.PowerFx.Core.Tests/CheckResultTests.cs
```diff
@@ -479,6 +447,36 @@ public void TestApplyGetLogging(string expr, string execptedLog, bool success)
...
+            {
+                string exMsg = null;
+
+                try
+                {
+                    var errors = check.ApplyErrors();
+                    exMsg = errorMsg.ToString();
+                    Assert.False(check.IsSuccess);
+                }
+                catch (Exception ex)
```
MikeStall, line 473: >catch (Exception ex) [](http://example.com/codeflow?start=16&length=20)

Who is throwing? Is it check.ApplyErrors()?

nguhoa: Yes, check.Applyerrors will throw exception of not implementation method if expectedreturntype is not string or number

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
nguhoa: Updated here. thanks

LucGenetier:
✅ No public API change.

jas-valgotar (approved):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

Issue #1429, opened 2023-04-24
PlainText function


MikeStall:
Implment PlainText in interpreter. 
https://learn.microsoft.com/en-us/power-platform/power-fx/reference/function-encode-decode

This is very similar to EncodeUrl, see #1194, #1215.

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/1454

PR #1454, opened 2023-04-28
Add PlainText Function

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1429


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -1059,6 +1068,34 @@ public static FormulaValue OptionSetValueToLogicalName(IRContext irContext, Opti
             var optionSet = args[0];
             var logicalName = optionSet.Option;
             return new StringValue(irContext, logicalName);
+        }
+
+        public static FormulaValue PlainText(IRContext irContext, StringValue[] args)
```
anderson-joyle, line 1073: >public static FormulaValue PlainText(IRContext irContext, StringValue[] args) [](http://example.com/codeflow?start=8&length=77)

How `PlainText` function should handle a blank value? PA returns `Blank()` while this implementation is returning `Error({Kind:ErrorKind.InvalidArgument})`

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1705,10 +1705,21 @@ static Library()
             {
                 BuiltinFunctionsCore.Language,
                 NoErrorHandling(Language)
-            }
+            },
+            {
+                BuiltinFunctionsCore.PlainText,
```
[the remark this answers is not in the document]
nguhoa: ordered them in alphabet. thanks

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: Checked on PowerApps side and updated logic to allow formularvalue input to fix here. Thanks.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/PlainText.cs
```diff
@@ -0,0 +1,28 @@
...
+        public PlainTextFunction()
+            : base("PlainText", TexlStrings.AboutPlainText, FunctionCategories.Text, DType.String, 0, 1, 1, DType.String)
+        {
+        }
+
+        public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
+        {
+            yield return new[] { TexlStrings.PlainTextArg1 };
+        }
+    }
```
jas-valgotar, line 27: Can we move already existing Implementation from PA?
(Luc recently moved regex function from PA so may have pointers on how to do it w/o breaking PA)
[https://msazure.visualstudio.com/OneAgile/_git/PowerApps-Client?path=/src/Cloud/DocumentServer.Core/Document/Texl/Builtins/PlainText.cs&version=GBmaster](https://msazure.visualstudio.com/OneAgile/_git/PowerApps-Client?path=/src/Cloud/DocumentServer.Core/Document/Texl/Builtins/PlainText.cs&version=GBmaster)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -939,6 +939,10 @@ static Library()
                     returnBehavior: ReturnBehavior.ReturnFalseIfAnyArgIsBlank,
                     targetFunction: IsToday)
             },
+            {
+                BuiltinFunctionsCore.Language,
+                NoErrorHandling(Language)
```
jas-valgotar, line 944: >NoErrorHandling [](http://example.com/codeflow?start=16&length=15)

Why No Error Handling?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -1059,6 +1069,39 @@ public static FormulaValue OptionSetValueToLogicalName(IRContext irContext, Opti
             var optionSet = args[0];
             var logicalName = optionSet.Option;
             return new StringValue(irContext, logicalName);
+        }
+
+        public static FormulaValue PlainText(IRContext irContext, FormulaValue[] args)
+        {
+            if (args[0].Type == FormulaType.String)
```
jas-valgotar, line 1076: >if (args[0].Type == FormulaType.String) [](http://example.com/codeflow?start=12&length=39)

Can we avoid this check and use error handling pipeline maybe?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
jas-valgotar: It has `returnBehavior: ReturnBehavior.ReturnBlankIfAnyArgIsBlank` which may be useful ;)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/PlainText.txt
```diff
@@ -0,0 +1,62 @@
...
+"-12.34"
+
+>> PlainText(Date(2023,1,1))
+"1/1/2023"
+
+>> PlainText(DateTime(2023,1,1,12,34,56))
+"1/1/2023 12:34 PM"
+
+>> PlainText(Time(8,1,15))
+"8:01 AM"
```
jas-valgotar, line 62: Can you add negative tests if it applies? Ie. where the function would return an error.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
nguhoa: This is language function, not in scope of this PR. but this is what Calos suggested.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/PlainText.cs
nguhoa: not sure what this mean but this function is trying to implement the same way with PA.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: this is not error handling. If it is string format then I tried to decode it.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
jas-valgotar: Oops my bad thought it was PlainText :D

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1204,6 +1208,17 @@ static Library()
...
             },
+            {
+                BuiltinFunctionsCore.PlainText,
+                StandardErrorHandling<FormulaValue>(
+                    BuiltinFunctionsCore.PlainText.Name,
+                    expandArguments: NoArgExpansion,
+                    replaceBlankValues: NoOpAlreadyHandledByIR,
+                    checkRuntimeTypes: DeferRuntimeTypeChecking,
+                    checkRuntimeValues: DeferRuntimeValueChecking,
+                    returnBehavior: ReturnBehavior.AlwaysEvaluateAndReturnResult,
```
jas-valgotar, line 1219: >ReturnBehavior.AlwaysEvaluateAndReturnResult [](http://example.com/codeflow?start=36&length=44)

maybe `ReturnBehavior.ReturnBlankIfAnyArgIsBlank ` ?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1204,6 +1208,17 @@ static Library()
                 BuiltinFunctionsCore.Pi,
                 NoErrorHandling(Pi)
             },
+            {
+                BuiltinFunctionsCore.PlainText,
+                StandardErrorHandling<FormulaValue>(
+                    BuiltinFunctionsCore.PlainText.Name,
+                    expandArguments: NoArgExpansion,
+                    replaceBlankValues: NoOpAlreadyHandledByIR,
```
jas-valgotar, line 1216: >NoOpAlreadyHandledByIR [](http://example.com/codeflow?start=40&length=22)

Should this be `DoNotReplaceBlank` ?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1204,6 +1208,17 @@ static Library()
...
+                    checkRuntimeTypes: DeferRuntimeTypeChecking,
```
jas-valgotar, line 1217: >DeferRuntimeTypeChecking [](http://example.com/codeflow?start=39&length=24)

Should this be `ExactValueTypeOrBlank<StringValue>` ?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
jas-valgotar: When would this be not string type? (Wouldn't Checktype already catch it?)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
nguhoa: Looks like ReturnBlankIfAnyArgIsBlank has been used for functions will multiple args and make sure return blank if any of them blank. But this plaintext function has only 1 input. and logic of it return blank if input blank already.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/PlainText.cs
jas-valgotar: Ideally, we want PA and FX to share a TexlFunction file.
Currently this file already exists in PA repo, so Ideally we should remove it from PA repo and move it to here FX repo...

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
nguhoa: I think this is fine. input is blank then output blank. so default let IR handle

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
nguhoa: No. want the value is formulavalue.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/PlainText.txt
nguhoa: Yes, added more error test here.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/PlainText.txt
```diff
@@ -0,0 +1,62 @@
...
+>> PlainText("  <  xml><p>Input<tr>  </xml>")
+"Input"
+
+>> PlainText("http&#x3a;//microsoft.com")
+"http://microsoft.com"
+
+>> PlainText("http&#58;//microsoft.com")
+"http://microsoft.com"
+
+>> PlainText(Blank())
```
jas-valgotar, line 46: >PlainText(Blank()) [](http://example.com/codeflow?start=3&length=18)

Can you please also add typed blank in the test
e.g. `PlainText(If(1<0, "text"))`

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
jas-valgotar: I think we set this to `NoOpAlreadyHandledByIR` when the IR actually converts Blank to something. Otherwise `DoNotReplaceBlank` is preferred.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
jas-valgotar: I think Line 1213 Also needs to be `StringValue`

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
jas-valgotar: Correct, I think the same should work for 1 Arg as well.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/PlainText.txt
nguhoa: Added this to test.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/PlainText.txt
```diff
@@ -0,0 +1,73 @@
...
+Blank()
```
CarlosFigueiraMSFT, line 47: Blank values, when passed to arguments that expect strings, should be interpreted as empty strings. So this should return `""`

anderson-joyle: Thanks for correcting me on this. `IsBlank(PlainText(Blank()))` returns true in PA, which made me think `Blank()` would be the right result. But then I realized that `IsBlank("")` is also true in PA.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/PlainText.cs
nguhoa: it is the same content file. what do you want to move from PA to Fx?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: i just allow they input differnt format here: string, blank() or datetime or boolean. but the decoding only process for string.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5158,4 +5158,15 @@
     <value>Can't delegate {0}: contains a behavior function '{1}'.</value>
     <comment>Warning message.</comment>
   </data>
+  <data name="AboutPlainText" xml:space="preserve">
+    <value>The PlainText function removes HTML and XML tags.</value>
+    <comment>Description of 'PlainText' function.</comment>
+  </data>
+  <data name="PlainTextArg1" xml:space="preserve">
+    <value>PlainText</value>
```
CarlosFigueiraMSFT, line 5166: >PlainText [](http://example.com/codeflow?start=11&length=9)

`text` (this is the text that is going to be "plainified")

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5158,4 +5158,15 @@
...
+    <comment>Function_parameter - First parameter for the PlainText function</comment>
+  </data>
+  <data name="AboutPlainText_PlainText" xml:space="preserve">
```
CarlosFigueiraMSFT, line 5169: >PlainText [](http://example.com/codeflow?start=29&length=9)

`text`

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5158,4 +5158,15 @@
     <value>Can't delegate {0}: contains a behavior function '{1}'.</value>
     <comment>Warning message.</comment>
   </data>
+  <data name="AboutPlainText" xml:space="preserve">
+    <value>The PlainText function removes HTML and XML tags.</value>
```
CarlosFigueiraMSFT, line 5162: >The PlainText function removes HTML and XML tags. [](http://example.com/codeflow?start=11&length=49)

We should keep the existing description (unless it is buggy): "Removes all formatting and returns a plain text value."

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5158,4 +5158,15 @@
...
+    <comment>Description of 'PlainText' function.</comment>
+  </data>
+  <data name="PlainTextArg1" xml:space="preserve">
+    <value>PlainText</value>
+    <comment>Function_parameter - First parameter for the PlainText function</comment>
```
CarlosFigueiraMSFT, line 5167: >Function_parameter - First parameter for the PlainText function [](http://example.com/codeflow?start=13&length=63)

"function_parameter - First argument of the PlainText function - the text to have its HTML and XML tags removed."

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
CarlosFigueiraMSFT, line 1216: >NoOpAlreadyHandledByIR [](http://example.com/codeflow?start=40&length=22)

If we say that it's handled by the IR, we probably need to override GetArgPreprocessor in the function (or make it inherit from StringOneArgFunction instead of BuiltinFunction)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -41,6 +43,14 @@ internal static partial class Library
         private static readonly Regex _minutesDetokenizeRegex = new Regex("[\u000A][\u000A]+", RegExFlags);
         private static readonly Regex _secondsDetokenizeRegex = new Regex("[\u0008][\u0008]+", RegExFlags);
         private static readonly Regex _milisecondsDetokenizeRegex = new Regex("[\u000e]+", RegExFlags);
+        private static readonly Regex _tdTagRegex = new Regex("<\\s*([tT][dD])\\s*\\/{0,1}>", RegExFlags);
```
CarlosFigueiraMSFT, line 46: >RegExFlags [](http://example.com/codeflow?start=94&length=10)

Instead of having the casing done in the regex string we can use RegexOptions.IgnoreCase in those regexes

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
CarlosFigueiraMSFT: Since this function declares its argument as string, it will only receive a string - if we pass a different data type it will be first coerced to a string prior to being passed to the function. You can check on the REPL project to see what happens under the hood.

For example, if you enter `IR(Len(Today()))` in the repl console, you'll see the call to the Text function (which implements Date[Time]-to-text coercion) in the IR:

    Len:w(
        BlankToEmptyString:s(
            Text:s(Today:D(), "'shortdate'":s)))

nguhoa: I see. thanks for explaining. i changed this back to stringvalue set.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
nguhoa: Ah I see, I updated this to replace empty string for blank value,

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: I am trying to use the same pattern with other Regex where we want to use RegExFlags. should I use IgnoreCase here instead?

src/strings/PowerFxResources.en-US.resx
nguhoa: Update here

src/strings/PowerFxResources.en-US.resx
nguhoa: Updated. thanks

src/strings/PowerFxResources.en-US.resx
nguhoa: update

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/PlainText.txt
nguhoa: Updated logic and now it return ""

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
CarlosFigueiraMSFT: I believe we can move all of them to use IgnoreCase - it will make it much easier to read

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -1059,6 +1069,34 @@ public static FormulaValue OptionSetValueToLogicalName(IRContext irContext, Opti
...
+            text = _styleTagRegex.Replace(text, string.Empty);
+
+            // Remove all comments.
+            text = _commentTagRegex.Replace(text, string.Empty);
+
+            // Insert empty string in place of <td>
+            text = _tdTagRegex.Replace(text, string.Empty);
+
+            //Replace <br> or <li> with line break
+            text = _lineBreakTagRegex.Replace(text, Environment.NewLine);
```
CarlosFigueiraMSFT, line 1091: >text = _lineBreakTagRegex.Replace(text, Environment.NewLine); [](http://example.com/codeflow?start=12&length=61)

We should have tests for this case. And for the next one.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: updated here.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: added tests.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -20,18 +22,18 @@ namespace Microsoft.PowerFx.Functions
...
     internal static partial class Library
     {
         private static readonly RegexOptions RegExFlags = LibraryFlags.RegExFlags;
 
-        private static readonly Regex _ampmReplaceRegex = new Regex("[aA][mM]\\/[pP][mM]", RegExFlags);
-        private static readonly Regex _apReplaceRegex = new Regex("[aA]\\/[pP]", RegExFlags);
-        private static readonly Regex _minutesBeforeSecondsRegex = new Regex("[mM][^dDyYhH]+[sS]", RegExFlags);
-        private static readonly Regex _minutesAfterHoursRegex = new Regex("[hH][^dDyYmM]+[mM]", RegExFlags);
-        private static readonly Regex _minutesRegex = new Regex("[mM]", RegExFlags);
+        private static readonly Regex _ampmReplaceRegex = new Regex("am\\/pm", RegExFlags);
```
MikeStall, line 32: why the change in regexs?

src/tests/Microsoft.PowerFx.Interpreter.Tests/PlainTextTests.cs
```diff
@@ -0,0 +1,36 @@
...
+        [Theory]
+        [InlineData("PlainText(\"Test<br id = 'br1'>break line</  BR>\")", "Test\r\nbreak line")]
+        [InlineData("PlainText(\"1<br>2<li>3\")", "1\r\n2\r\n3")]
+        [InlineData("PlainText(\"1<BR/>2<li/>3\")", "1\r\n2\r\n3")]
+        [InlineData("PlainText(\"Test<p >double</  p><div >break line</ Div>\")", "Test\r\n\r\ndouble\r\n\r\nbreak line")]
+        [InlineData("PlainText(\"1<div>2<p>3<tr>4\")", "1\r\n\r\n2\r\n\r\n3\r\n\r\n4")]
+        [InlineData("PlainText(\"1<Div/>2<p/>3<Tr/>4\")", "1\r\n\r\n2\r\n\r\n3\r\n\r\n4")]
+        [InlineData("PlainText(\"1<  br  >2<  Li />3\")", "1\r\n2\r\n3")]
+        [InlineData("PlainText(\"1<div  class = 'div2'>2< P />3<  tr>4\")", "1\r\n\r\n2\r\n\r\n3\r\n\r\n4")]
+        public async Task PlainTextFunctionTest(string inputExp, string expectedResult)
```
MikeStall, line 23: Can these be merged into PlainText.txt?  EG, we may not need any dedicated C#  test.

CarlosFigueiraMSFT: I don't think the .txt tests support \r\n in the output. We may be able to work around it by using something like

    >> Substitute(Substitute(PlainText(\"1<  br  >2<  Li />3\")", "1\r\n2\r\n3"), Char(13), "\r"), Char(10), "\n")
    "1\r\n2\r\n3"

Or we may want to see if we can "fix" the test runner to account for multi-line strings

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: Carlos mentioned use ignorecase for easy reading so I updated existing regex as well. Is it ok? or you want me remove ignorecase tage from RegExFlags?

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/PlainTextTests.cs
nguhoa: Thanks for suggestion. Removed this file and add to test txt file.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: Updated RegExFlags for ignorecase here

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/PlainText.txt
```diff
@@ -0,0 +1,97 @@
+>> PlainText("")
```
MikeStall, line 1: >PlainText [](http://example.com/codeflow?start=3&length=9)

Just double checking: these are all correct for PowerApps ? This .txt will run against JS tests too.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: updated

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/PlainText.cs
nguhoa: will remove this from PA after add to Fx.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
nguhoa: Thanks

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
nguhoa: Updated here.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
nguhoa: updated.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
nguhoa: updated

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -1059,6 +1071,34 @@ public static FormulaValue OptionSetValueToLogicalName(IRContext irContext, Opti
             var optionSet = args[0];
             var logicalName = optionSet.Option;
             return new StringValue(irContext, logicalName);
+        }
+
+        public static FormulaValue PlainText(IRContext irContext, StringValue[] args)
+        {        
+            //Decode html specific characters
+            string text = WebUtility.HtmlDecode(args[0].Value.Trim());
```
CarlosFigueiraMSFT, line 1079: >string text = WebUtility.HtmlDecode(args[0].Value.Trim()); [](http://example.com/codeflow?start=12&length=58)

This may cause double decoding. For example, `PlainText("Hello&lt;br/&gt;world")` should return `"Hello<br/>world"` but it now adds a newline where it shouldn't.

nguhoa: Move this to the end of function to not double decoding.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/PlainText.txt
nguhoa: all these tests have the same results with PA.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (dismissed):
🚢

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/PlainText.cs
anderson-joyle: You can use the following PR as a guide:
[https://msazure.visualstudio.com/OneAgile/_git/PowerApps-Client/pullrequest/7757449?_a=files](https://msazure.visualstudio.com/OneAgile/_git/PowerApps-Client/pullrequest/7757449?_a=files)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
jas-valgotar: I agree with Carlos, we should inherit `StringOneArgFunction` and have this as `NoOpAlreadyHandledByIR` that way it can be handled at IR phase and the advantage is there other backend in future would not need to worry about the blank handling.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/PlainText.cs
nguhoa: Thanks, will use it.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
nguhoa: Updated to use StringOneArgFunction.

jas-valgotar (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1457, opened 2023-04-28
Fixes Issue for DV when collect field is Polyymorphic.

jas-valgotar:
Fixes https://github.com/microsoft/Power-Fx-Dataverse/issues/165


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -2813,6 +2813,16 @@ private static DType UnionCore(ref bool fError, DType type1, DType type2, bool u
                         isMatchingExpandType = IsMatchingExpandType(field2Type, field1Type);
                         expandType = field2Type;
                     }
+                    else if (field1Type.HasPolymorphicInfo && field2Type.IsAggregate)
```
MikeStall, line 2816: >HasPolymorphicInfo [](http://example.com/codeflow?start=40&length=18)

Do we need to do some additional comparison? 
Would this match Polymorphic Contacts with Aggregate Accounts?

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

Issue #1402, opened 2023-04-19
Decimal RandBetween should have more granularity


gregli-msft:
Using a double for random mumbers results in at most 15 decimal places of random numbers.  Decimal RandBetween should return a full 28 digits if RandBetween is called with -MaxDecimal to MaxDecimal.

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/1460

PR #1460, opened 2023-04-30
Decimal RandBetween return a full 28 digits

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1402


LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Functions.IRandomService.NextDecimal

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Functions.IRandomService.NextDecimal

src/libraries/Microsoft.PowerFx.Interpreter/Functions/IRandomService.cs
```diff
@@ -31,12 +39,25 @@ internal class DefaultRandomService : IRandomService
         [ThreadSafeProtectedByLock(nameof(_randomizerLock))]
         private static readonly Random _random = new Random();
 
+        private static readonly object _randomizerDecimalLock = new object();
+
+        [ThreadSafeProtectedByLock(nameof(_randomizerDecimalLock))]
+        private static readonly Random _randomDecimal = new Random();
```
CarlosFigueiraMSFT, line 45: Do we need another Random object? We should be able to use the same as the one used to return double values.

CarlosFigueiraMSFT: That should not be a problem - those methods are very small and should not have locking problems.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/IRandomService.cs
```diff
@@ -31,12 +39,25 @@ internal class DefaultRandomService : IRandomService
...
             {
                 return _random.NextDouble();
             }
         }
+
+        public virtual decimal NextDecimal()
+        {
+            lock (_randomizerDecimalLock)
+            {
+                return new decimal(_randomDecimal.Next(), _randomDecimal.Next(), _randomDecimal.Next(542101087), false, 28);
```
CarlosFigueiraMSFT, line 59: >return new decimal(_randomDecimal.Next(), _randomDecimal.Next(), _randomDecimal.Next(542101087), false, 28); [](http://example.com/codeflow?start=16&length=108)

It's possible that this will return a number greater than 1:

    var dec = new Decimal(
        int.MaxValue,
        int.MaxValue,
        542101086, false, 28);

returns

    1.0000000004745384014313947135

nguhoa: i run manytime and always get less than 1 so I thought it was right. But sounds it can be more than 1. 
I updated logic here to make sure we get value less than 1.

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Functions.IRandomService.NextDecimal

src/libraries/Microsoft.PowerFx.Interpreter/Functions/IRandomService.cs
CarlosFigueiraMSFT: >_randomDecimal.Next() [](http://example.com/codeflow?start=35&length=21)

This only creates numbers with the lower 31 bits possibly set (the higher bit is always 0), which makes this distribution not uniform.

nguhoa: I tried to update logic to make it more random. Please review my new change.

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Functions.IRandomService.NextDecimal

src/libraries/Microsoft.PowerFx.Interpreter/Functions/IRandomService.cs
nguhoa: Updated here

src/libraries/Microsoft.PowerFx.Interpreter/Functions/IRandomService.cs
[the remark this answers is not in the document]
nguhoa: Changed logic to create random decimal from 0 to 1 here.

nguhoa: Thanks for offline chat and suggestion. I updated logic here.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/IRandomService.cs
```diff
@@ -38,5 +47,28 @@ public virtual double NextDouble()
...
+                    randChars[i] = (char)_random.Next(48, 58);
+                }
+
+                var randStr = new string(randChars);
+                if (decimal.TryParse(randStr, out decimal decimalResult))
+                {
+                    return decimalResult;
+                }
+
+                return decimal.MaxValue;
```
CarlosFigueiraMSFT, line 70: >decimal.MaxValue [](http://example.com/codeflow?start=23&length=16)

We shouldn't return decimal.MaxValue - this function should always return something between [0,1). 0.5?

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Functions.IRandomService.NextDecimal

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Functions.IRandomService.NextDecimal

src/libraries/Microsoft.PowerFx.Interpreter/Functions/IRandomService.cs
```diff
@@ -38,5 +47,28 @@ public virtual double NextDouble()
...
+                randChars[0] = '0';
+                randChars[1] = '.';
+
+                for (int i = 2; i < 30; i++)
+                {
+                    randChars[i] = (char)_random.Next(48, 58);
+                }
+
+                var randStr = new string(randChars);
+                if (decimal.TryParse(randStr, out decimal decimalResult))
```
CarlosFigueiraMSFT, line 65: >randStr,  [](http://example.com/codeflow?start=37&length=9)

Pass invariant culture, since we are using `.` as the decimal separator

nguhoa: passed invariant culture here.

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Functions.IRandomService.NextDecimal

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Functions.IRandomService.NextDecimal

CarlosFigueiraMSFT (approved):
🚢

src/libraries/Microsoft.PowerFx.Interpreter/Functions/IRandomService.cs
MikeStall: Note this means it's not truly a linear distribution:  there is a higher chance of getting .5 than other values.

CarlosFigueiraMSFT: This shouldn't actually happen - the number created above is well-formed, so this is more of a "safeguard". We could, however, change it to `return (decimal)_random.NextDouble();` to be safe

MikeStall: There is an interesting discussion on SO here: 
[https://stackoverflow.com/questions/609501/generating-a-random-decimal-in-c-sharp](https://stackoverflow.com/questions/609501/generating-a-random-decimal-in-c-sharp)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -1297,7 +1316,7 @@ public static FormulaValue RandBetweenDecimal(IServiceProvider services, IRConte
             lower = Math.Ceiling(lower);
             upper = Math.Floor(upper);
 
-            decimal value = (decimal)services.SafeNextDouble();
+            decimal value = (decimal)services.SafeNextDecimal();
```
MikeStall, line 1319: >(decimal) [](http://example.com/codeflow?start=28&length=9)

remove cast?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/IRandomService.cs
CarlosFigueiraMSFT: We started from it :-). As it is stated in the discussion, the distribution in most cases is not uniform, so we iterated through some and came up with that algorithm which works for our scenario (we don't need a random decimal, we just need one between 0-1 that has as much precision as possible. Decimal can have up to 29 digits, but with 29 it would only go to a 7.9 mantissa, so limiting it to 28 significant digits gives us a uniform distribution that other distributions couldn't provide.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/IRandomService.cs
```diff
@@ -38,5 +47,28 @@ public virtual double NextDouble()
                 return _random.NextDouble();
             }
         }
+
+        public virtual decimal NextDecimal()
```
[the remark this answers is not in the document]
nguhoa: Updated to create the extension method for nextdecimal.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/IRandomService.cs
```diff
@@ -21,6 +22,14 @@ public interface IRandomService
...
         /// <returns></returns>
         double NextDouble();
+
+        /// <summary>
+        /// Returns a random decimal number that is greater than or equal to 0.0, and less than 1.0.
+        /// Should be from a linear distribution. 
+        /// Expected this call is thread-safe and may be called concurrently from multiple threads.
+        /// </summary>
+        /// <returns></returns>
+        decimal NextDecimal();
```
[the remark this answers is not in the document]
nguhoa: Updated.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
nguhoa: removed

CarlosFigueiraMSFT (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -1235,6 +1235,33 @@ private static double SafeNextDouble(this IServiceProvider services)
             return random.SafeNextDouble();
         }
 
+        private static decimal SafeNextDecimal(this IRandomService random)
```
MikeStall, line 1238: >SafeNextDecimal [](http://example.com/codeflow?start=31&length=15)

super nit:
can you move this to its own class 

internal static class RandomServiceExtensions 
{
public static decimal SafeNextDecimal(this IRandomService random)
}

That can live in same file as IRandomService, or a new file.

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
nguhoa: updated

LucGenetier:
✅ No public API change.

PR #1462, opened 2023-05-01
Adds null handling in Power operator. (^)

### LINKED ISSUE #1297: "16^Blank()" and "Blank()^16" leads to a null exception

jas-valgotar:
-Fixes #1297

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
@@ -1462,7 +1462,8 @@ namespace Microsoft.PowerFx.Core.Functions
         /// By default, function does not attach any pre-processing for arguments.
         /// </summary>
         /// <param name="index">0 based index of argument.</param>
-        /// <param name="argCount">The count of argument to call the function.</param>
+        /// <param name="argCount">The count of arguments to the called function. 
+        /// e.g. Text() function, argCount can be 2 or 3 depending on the expression.</param>
         /// <returns></returns>
         public virtual ArgPreprocessor GetArgPreprocessor(int index, int argCount)
         {
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
@@ -1462,8 +1462,8 @@ namespace Microsoft.PowerFx.Core.Functions
         /// By default, function does not attach any pre-processing for arguments.
         /// </summary>
         /// <param name="index">0 based index of argument.</param>
-        /// <param name="argCount">The count of arguments to the called function. 
-        /// e.g. Text() function, argCount can be 2 or 3 depending on the expression.</param>
+        /// <param name="argCount">The number of arguments passed in the function invocation. 
+        /// For example, the Text function has different blank handling behaviors depending on whether it receives 1 argument (blanks are propagated) or more (blanks are converted to empty strings).</param>
         /// <returns></returns>
         public virtual ArgPreprocessor GetArgPreprocessor(int index, int argCount)
         {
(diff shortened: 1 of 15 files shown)


LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
🚢

CarlosFigueiraMSFT (dismissed):
🚢

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -208,8 +208,10 @@ public override IntermediateNode Visit(TexlBinaryOpNode node, IRTranslatorContex
                 switch (kind)
                 {
                     // Call Node Replacements:
-                    case BinaryOpKind.Power:
-                        binaryOpResult = new CallNode(context.GetIRContext(node), BuiltinFunctionsCore.Power, left, right);
+                    case BinaryOpKind.Power:
+                        var args = new List<IntermediateNode> { left, right };
+                        args = AttachArgPreprocessor(args, BuiltinFunctionsCore.Power, node, 2, context);
```
ian-legler, line 213: There should probably be some code comments here explaining why the IR is doing explicit arg preprocessing here and not in other call or binary op nodes generated

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
jas-valgotar: Added. Thanks!

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -1462,9 +1462,9 @@ internal virtual IntermediateNode CreateIRCallNode(CallNode node, IRTranslatorCo
         /// By default, function does not attach any pre-processing for arguments.
         /// </summary>
         /// <param name="index">0 based index of argument.</param>
-        /// <param name="node">The original call node.</param>
+        /// <param name="argCount">The count of argument to call the function.</param>
```
CarlosFigueiraMSFT, line 1465: >The count of argument to call the function. [](http://example.com/codeflow?start=35&length=43)

Can you give an example of this parameter? It's not clear based on the description.

MikeStall (dismissed):

CarlosFigueiraMSFT (approved):
🚢

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
jas-valgotar: Added more description.

CarlosFigueiraMSFT (approved):
🚢

LucGenetier:
✅ No public API change.

PR #1469, opened 2023-05-02
Fix to delegation warning for named formula if it's non-delegable.

Shpakh-MSFT:
This change brings the named formula to the same consistency for delegation warnings and marks the expression as non-delegable if it's a part of non-delegable expression.


LucGenetier:
✅ No public API change.

ian-legler (approved):
:shipit:

LucGenetier:
✅ No public API change.

Issue #1425, opened 2023-04-24
Error when provided empty expression to the field with "Specify Return Type" feature


justinpark0916:
We found out that we are seeing this error("The type of this expression does not match the expected type `Type`. Found type `ObjNull`.) when an empty expression is provided to the field with `Specify Return Type` feature. Instead of this, we would like to avoid providing an error message when an empty expression is passed.

![image](https://user-images.githubusercontent.com/104940715/234017980-d690e292-39e7-4848-9ff0-48d2900f9e74.png)

Attached the unit test to reproduce:
```
        [Fact]
        public async Task test()
        {
            // Setup
            var scope = new EditorContextScope((expr) => new CheckResult(new Engine())
                    .SetText(expr)
                    .SetBindingInfo()
                    .SetExpectedReturnValue(FormulaType.String));

            // Act
            var result = scope.Check("");

            // Assert
            Assert.True(result.IsSuccess);
        }
```

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/1475

PR #1475, opened 2023-05-03
Handle empty expression for return type

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1425


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/LanguageServerTests.cs
```diff
@@ -1376,6 +1376,16 @@ public void CustomCheckResult()
...
         }
 
+        [Fact]
+        public async Task TestExpectedReturnValueForEmptyExpression()
+        {
+            var scope = TestCreateEditorScope(string.Empty);
+
+            var result = scope.Check(string.Empty);
+
+            Assert.True(result.IsSuccess);
```
MikeStall, line 1386: what happens if we call result.GetEvaluator()  and try to run it? 
Does it succeed and return Blank()?

LucGenetier:
✅ No public API change.

MikeStall (approved):

LucGenetier:
✅ No public API change.

Issue #1458, opened 2023-04-28
Text( Blank() ) should return Blank()


gregli-msft:
For the 1 argument version of Text, which is serving as a Power Fx type converter, we should preserve Blank() flowing through.  That means that Text( Blank() ), Text( If(1<0, 2) ), and Text( ParseJSON( "null" ) ) should all return Blank().

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/1478

PR #1478, opened 2023-05-03
Return Blank for blank input of Text

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1458


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text.txt
```diff
@@ -5,7 +5,7 @@
 "123.400"
 
 >> Text(Blank(), "#.000")
-""
+Blank()
```
CarlosFigueiraMSFT, line 8: In the latest discussions with PMs we agreed that if there is only one argument to the Text function (i.e., it's being used as a conversion function) then blank values are propagated. However, if we explicitly ask for a format, as in this example, then the result should be an empty string (`""`)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text.txt
```diff
@@ -136,32 +136,37 @@
...
 "pqr"
 
-
 >> Text(1/0, "#.000")
 Error({Kind:ErrorKind.Div0})
 
+>> Text(If(1<0,2))
+Blank()
+
+>> Text(ParseJSON("null"))
```
CarlosFigueiraMSFT, line 167: We already have this test in ParseJson.txt, no need to repeat it

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text.txt
nguhoa: Updated to handle this. It was not clear in the problem. Thanks for explaining here.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text.txt
nguhoa: remove this.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
```diff
@@ -18,7 +18,8 @@ private enum ReturnBehavior
             AlwaysEvaluateAndReturnResult,
             ReturnBlankIfAnyArgIsBlank,
             ReturnEmptyStringIfAnyArgIsBlank,
-            ReturnFalseIfAnyArgIsBlank
+            ReturnFalseIfAnyArgIsBlank,
+            ReturnBlankIfOneArgIsBlankAndEmptyStringForOtherArgIsBlank
```
CarlosFigueiraMSFT, line 22: >ReturnBlankIfOneArgIsBlankAndEmptyStringForOtherArgIsBlank [](http://example.com/codeflow?start=12&length=58)

This is way too specific - it's only used by the Text function. Do we need to have a new one, or should just use AlwaysEvaluateAndReturnResult and let the Text function handle it?

nguhoa: Updated here and handle in text.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -295,7 +295,20 @@ public static bool TryDecimal(FormattingInfo formatInfo, IRContext irContext, Fo
 
         // https://docs.microsoft.com/en-us/powerapps/maker/canvas-apps/functions/function-text
         public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
-        {
+        {
+            if (args.Length == 1 && (args[0] is BlankValue || (args[0] is UntypedObjectValue untypedObjectValue && untypedObjectValue.Impl.Type == FormulaType.Blank)))
+            {
+                return new BlankValue(irContext);
```
CarlosFigueiraMSFT, line 301: Can you add a comment explaining the logic? Something like "// When used as a pure conversion function (single argument, no format string), Text() propagates null values"

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -295,7 +295,20 @@ public static bool TryDecimal(FormattingInfo formatInfo, IRContext irContext, Fo
 
         // https://docs.microsoft.com/en-us/powerapps/maker/canvas-apps/functions/function-text
         public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
-        {
+        {
+            if (args.Length == 1 && (args[0] is BlankValue || (args[0] is UntypedObjectValue untypedObjectValue && untypedObjectValue.Impl.Type == FormulaType.Blank)))
```
MikeStall, line 299: >args[0] is BlankValue || (args[0] is UntypedObjectValue untypedObjectValue && untypedObjectValue.Impl.Type == FormulaType.Blank [](http://example.com/codeflow?start=37&length=127)

Do we have other places that need this check?

I wonder if we should encapsulate it in a new helper like:
FormulaValue.IsBlank()

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: Add comment. thanks

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: Yes, we use in few places so I updated Greg's isblank function to share.

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

Issue #1441, opened 2023-04-25
DateTimeValue() without a date should use the epoch instead of today's date


gregli-msft:
These are not consistent:

```
> DateTimeValue("12:00:00")
DateTime(2023,4,25,12,0,0,0)

> Text(DateTimeValue("12:00:45"),"yyyy-mm-dd hh:mm:ss")
"2023-04-25 12:00:45"

> Text(DateTimeValue("1/1/1999 12:00:45"),"yyyy-mm-dd hh:mm:ss")
"1999-01-01 12:00:45"

> Text(TimeValue("12:00:45"),"yyyy-mm-dd hh:mm:ss")
"1899-12-30 12:00:45"
```

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/1481

PR #1481, opened 2023-05-04
Use epoch for DateTimeValue and DateValue without date

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1441


MikeStall (approved):

LucGenetier:
✅ No public API change.

Issue #1482, opened 2023-05-04
Errors should have default message if one is not provided


CarlosFigueiraMSFT:
This already happens in Power Apps, and we should do the same in the C# interpreter:
![image](https://user-images.githubusercontent.com/44986255/236238910-0769bdad-6bf2-4b81-b4b6-9a4c1e1cb41f.png)

Issue #548, opened 2022-08-12
Localize interpreter error messages.


MikeStall:
Interpreter error messages are hard coded.  

>> IfError(Text(1/0), FirstError.Message)
"Invalid operation: division by zero."

But they should actually be localized. This requires some additional infrastructure we don't have yet. 

![image](https://user-images.githubusercontent.com/1538900/184449794-01e9cdfd-1716-499c-987f-db83c7643701.png)

PR #1483, opened 2023-05-04
Add default messages for errors

CarlosFigueiraMSFT:
Error values should have a default message if one is not provided. See #1482 for an example.


src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -2255,10 +2255,111 @@ public static FormulaValue Error(IRContext irContext, FormulaValue[] args)
...
             return result;
+        }
+
+        private static string GetDefaultErrorMessage(ErrorKind errorKind)
+        {
+            switch (errorKind)
+            {
+                case ErrorKind.None:
+                    // Default message that is shown to users when an error was produced, but with kind = 'none'
+                    return "System no error";
```
anderson-joyle, line 2271: >"System no error" [](http://example.com/codeflow?start=27&length=17)

Do we need to localize those messages? Maybe we could populate MessageKey field so they can get translated automatically.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
CarlosFigueiraMSFT: Good point, we definitely should... but I've seen many other messages from the interpreter hardcoded. Let me check with Mike to see what's our story for the interpreter.

CarlosFigueiraMSFT: Messages in the interpreter are not localized today, we have some work items to fix that. While that assembly doesn't have its own resources file, the messages need to stay hardcoded for now.

MikeStall (dismissed):

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -2255,10 +2255,111 @@ public static FormulaValue Error(IRContext irContext, FormulaValue[] args)
...
 
-                result.Add(new ExpressionError { Kind = errorKind, Message = messageField?.Value as string });
+                var message = messageField != null ? messageField.Value : GetDefaultErrorMessage(errorKind);
+                result.Add(new ExpressionError { Kind = errorKind, Message = message });
             }
 
             return result;
+        }
+
+        private static string GetDefaultErrorMessage(ErrorKind errorKind)
```
MikeStall, line 2265: >GetDefaultErrorMessage [](http://example.com/codeflow?start=30&length=22)

Where are these coming from?  (PA?)

If somebody adds to ErrorKind, how do they know to update this? 
Can we add a test that just uses refleection to get each value of ErrorKind, calls this method, and ensures valid (eg, doesn't hit default case at end)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
MikeStall: Yes, tracked with #548

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
CarlosFigueiraMSFT: Yes, the messages come from Power Apps. Good idea about the test, added one.

MikeStall (approved):

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #1484, opened 2023-05-04
[PowerApps] Localization hand-back [CTAS - RunID=20230504-201158-njr6…

toshio-msft:
…ff8ag3]


toshio-msft:
Create draft PR for automation test

PR #1486, opened 2023-05-04
Allow coercion from numeric option set to decimal

CarlosFigueiraMSFT:
We have an unary coercion from numeric option set to decimal, but this isn't allowed today. This change fixes it.


LucGenetier:
✅ No public API change.

MikeStall (approved):

LucGenetier:
✅ No public API change.

Issue #1440, opened 2023-04-25
Locale specification in Text format string not working in interpreter


gregli-msft:
Diverges from Excel and Power Apps
```
> Text(Value("1,234","en-us"),"[$-fr-FR]#,###","fr-fr")
"[$-fr-FR]1 234"  // should be "1234"

> Text(Value("1,234","en-us"),"[$-en-us]#,###","fr-fr")
"[$-en-us]1 234"  // should be "1 234"
```

PR #1488, opened 2023-05-05
Locale specification in Text format string

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1440


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -351,8 +351,24 @@ public static bool TryText(FormattingInfo formatInfo, IRContext irContext, Formu
...
+            if (!string.IsNullOrEmpty(formatString) && formatString.Contains("[$-"))
+            {
+                hasFormatCulture = true;
+                int startIdx = formatString.IndexOf("[$-", StringComparison.Ordinal) + 3;
+                int endIdx = formatString.IndexOf("]", StringComparison.Ordinal);
+                formatCulture = CultureInfo.GetCultureInfo(formatString.Substring(startIdx, endIdx - startIdx));
+                formatString = formatString.Substring(endIdx + 1);
+                revFormatStr = formatString.Replace(",", "!");
+                revFormatStr = revFormatStr.Replace(".", ",");
+                revFormatStr = revFormatStr.Replace("!", ".");
```
CarlosFigueiraMSFT, line 369: There are cases other than `.` and `,` (and vice-versa) for decimal separator and thousands grouping. For example, `Text(1234.5678, "[$-fr-FR]# ##0,00", "en-US")` returns "1,234.57" (the thousands grouping in fr-FR is a space)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text.txt
```diff
@@ -388,4 +388,40 @@ Error({Kind:ErrorKind.BadLanguageCode})
 "2023-12-21T12:34:56.789Z"
 
 >> Text(DateTimeValue("12:34:56"),"yyyy-mm-dd hh:mm:ss")
-"1899-12-30 12:34:56"
\ No newline at end of file
+"1899-12-30 12:34:56"
+
+>> Text(Value("1,234","en-US"),"[$-fr-FR]#,###","fr-FR")
```
CarlosFigueiraMSFT, line 393: >Text(Value("1,234","en-US"),"[$-fr-FR]#,###","fr-FR") [](http://example.com/codeflow?start=3&length=53)

Those tests don't need to use the Value function - we can use a constant number since we're testing the Text function here. A good number to test would be 1234.5678 - it will have both decimal separator and possible thousands separator as well.

Some good tests:
    >> Text(1234.5678, "[$-en-US]#,##0.00", "en-US")
    "1,234.57"
    
    >> Text(1234.5678, "[$-pt-BR]#.##0,00", "en-US")
    "1,234.57"
    
    >> Text(1234.5678, "[$-pt-BR]#.##0,00", "fr-FR")
    "1 234,57"
    
    >> Text(1234.5678, "[$-en-US]#,##0.00", "pt-BR")
    "1.234,57"
    
    // If not closed by a ']', then not interpreted as a format specifier
    >> Text(1234.5678, "[$-en-US#,##0.00", "en-US")
    "[$-en-US1,234.57"
    
    >> Text(1234.5678, "[$-fr]# ##0,00", "en-US")
    "1,234.57"
    
    >> Text(1234.5678, "[$-fr-FR]# ##0,00", "vi-VI")
    "1.234,57"
    
    >> Text(1234.5678, "[$-en-US]#,##0.00", "en-US")
    "1,234.57"
    
    >> Text(1234.5678, "[$-pt-BR]#.##0,00", "en-US")
    "1,234.57"
    
    >> Text(1234.5678, "[$-pt-BR]#.##0,00", "fr-FR")
    "1 234,57"
    
    >> Text(1234.5678, "[$-en-US]#,##0.00", "pt-BR")
    "1.234,57"
    
    // If not closed by a ']', then not interpreted as a format specifier
    >> Text(1234.5678, "[$-en-US#,##0.00", "en-US")
    "[$-en-US1,234.57"
    
    >> Text(1234.5678, "[$-fr]# ##0,00", "en-US")
    "1,234.57"
    
    // Unknown locales would default to en-US in C# (error in Power Apps)
    >> Text(1234.5678, "[$-xx-YY]0,0", "en-US")
    "1,235"

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -351,8 +351,24 @@ public static bool TryText(FormattingInfo formatInfo, IRContext irContext, Formu
...
 
-            Contract.Assert(StringValue.AllowedListConvertToString.Contains(value.Type));
-
+            Contract.Assert(StringValue.AllowedListConvertToString.Contains(value.Type));
+
+            bool hasFormatCulture = false;
+            string enUsCultureName = "en-US";
+            CultureInfo formatCulture = culture;
+            string revFormatStr = formatString;
+            if (!string.IsNullOrEmpty(formatString) && formatString.Contains("[$-"))
```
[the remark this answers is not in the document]
nguhoa: Updated. thanks

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: Add space handling here.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -351,8 +356,37 @@ public static bool TryText(FormattingInfo formatInfo, IRContext irContext, Formu
...
+                    {
+                        result = new StringValue(irContext, formatCultureName);
+                        return false;
+                    }
+
+                    formatString = formatString.Substring(endIdx + 1);
+                    revFormatStr = formatString.Replace(",", "!");
+                    revFormatStr = revFormatStr.Replace(".", ",");
+                    revFormatStr = revFormatStr.Replace(" ", ",");
+                    revFormatStr = revFormatStr.Replace("!", ".");
```
CarlosFigueiraMSFT, line 386: Can we pull the separators from the formatCulture itself?

CarlosFigueiraMSFT, line 386: `'!'` and `' '` are valid characters in the format string (are output "as-is"). Something like `Text(1234.5678,"0 # # # !!!")` is valid (and outputs "1 2 3 5 !!!"

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -25,7 +26,9 @@ public static bool IsValidFormatArg(string formatArg, out bool hasDateTimeFmt, o
                 }
             }
 
-            hasDateTimeFmt = formatArg.IndexOfAny(new char[] { 'm', 'M', 'd', 'D', 'y', 'Y', 'h', 'H', 's', 'S', 'a', 'A', 'p', 'P' }) >= 0;
+            hasDateTimeFmt = (formatArg.IndexOfAny(new char[] { 'm', 'M', 'd', 'D', 'y', 'Y', 'h', 'H', 's', 'S', 'a', 'A', 'p', 'P' }) >= 0) 
+                && (formatArg.ToLower(CultureInfo.InvariantCulture).IndexOf("us", StringComparison.Ordinal) < 0);
```
MikeStall, line 30: Can this be ignore case? and then we can avoid ToLower call?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -351,8 +356,37 @@ public static bool TryText(FormattingInfo formatInfo, IRContext irContext, Formu
...
+            Contract.Assert(StringValue.AllowedListConvertToString.Contains(value.Type));
+
+            bool hasFormatCulture = false;
+            string enUsCultureName = "en-US";
+            CultureInfo formatCulture = culture;
+            string revFormatStr = formatString;
+
+            if (!string.IsNullOrEmpty(formatString) && formatString.StartsWith("[$-", StringComparison.OrdinalIgnoreCase))
+            {
+                int startIdx = formatString.IndexOf("[$-", StringComparison.Ordinal) + 3;
```
MikeStall, line 368: >formatString.IndexOf("[$-" [](http://example.com/codeflow?start=31&length=26)

We do  this check twice. (StartsWith does it) 

Can we just do it once? 
Just call IndexOf, and then check StartIndex==0.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -351,8 +356,37 @@ public static bool TryText(FormattingInfo formatInfo, IRContext irContext, Formu
...
+                int endIdx = formatString.IndexOf("]", StringComparison.Ordinal);
```
MikeStall, line 369: >IndexOf [](http://example.com/codeflow?start=42&length=7)

You can pass the starting index in here; so you don't scan the start of the string twice.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
MikeStall, line 386: What is this doing?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text.txt
MikeStall: Can you include the test comments too?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -383,8 +434,26 @@ public static bool TryText(FormattingInfo formatInfo, IRContext irContext, Formu
                     }
                     else
                     {
-                        var normalized = dec.Normalize();
-                        result = new StringValue(irContext, normalized.ToString(formatString ?? "g", culture));
+                        var normalized = dec.Normalize();
+
+                        if (!string.IsNullOrEmpty(formatString))
```
MikeStall, line 439: >if [](http://example.com/codeflow?start=24&length=2)

Can this share with above?
Both decimal and double implement common 
IFormattable

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -367,8 +401,25 @@ public static bool TryText(FormattingInfo formatInfo, IRContext irContext, Formu
...
+                        {
+                            if (hasFormatCulture)
+                            {                                
+                                if (revFormatStr != formatString
+                                    && (num.Value.ToString(formatString, formatCulture) != num.Value.ToString(formatString, CultureInfo.GetCultureInfo(enUsCultureName))
+                                        || num.Value.ToString(revFormatStr, formatCulture) != num.Value.ToString(revFormatStr, CultureInfo.GetCultureInfo(enUsCultureName))))
+                                {
+                                    formatString = revFormatStr;
+                                }
```
MikeStall, line 414: what is this doing?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: In this one I tried to find the reverse version of input format string. So I cannot get separators by formatculture.

nguhoa: format string in c# only work for en-US format. They don't understand the , and . in different culture. 
As in https://learn.microsoft.com/en-us/dotnet/standard/base-types/custom-numeric-format-strings.
So my idea here is if we pass different culture format, we finding the reverse version of it (will be in en-US format if the input is non en-US and have different format like fr-fr or vi-vi). 
Then we use both format, the input string format and reverse input string format to compare the results of output string when we applied 2 different cultures we have here.
If any of comparison return different mean 2 cultures have different numeric format. Then we need to use reverse format (english format) first then applied that different culture (fr-fr for example).

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: I explained above:

format string in c# only work for en-US format. They don't understand the , and . in different culture.
As in https://learn.microsoft.com/en-us/dotnet/standard/base-types/custom-numeric-format-strings.
So my idea here is if we pass different culture format, we finding the reverse version of it (will be in en-US format if the input is non en-US and have different format like fr-fr or vi-vi).
Then we use both format, the input string format and reverse input string format to compare the results of output string when we applied 2 different cultures we have here.
If any of comparison return different mean 2 cultures have different numeric format. Then we need to use reverse format (english format) first then applied that different culture (fr-fr for example).

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: update here

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
nguhoa: yes, update

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: updated here

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: updated

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: updated

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text.txt
nguhoa: update

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -351,8 +356,38 @@ public static bool TryText(FormattingInfo formatInfo, IRContext irContext, Formu
...
+
+                if (startIdx > -1 && endIdx > 2)
+                {
+                    startIdx += 3;
+                    hasFormatCulture = true;
+                    string formatCultureName = formatString.Substring(startIdx, endIdx - startIdx);
+
+                    if (!TryGetCulture(formatCultureName, out formatCulture))
+                    {
+                        result = new StringValue(irContext, formatCultureName);
```
CarlosFigueiraMSFT, line 379: >new StringValue(irContext, formatCultureName); [](http://example.com/codeflow?start=33&length=46)

If the result of a TryXXX method is false, we can set the result to null.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
CarlosFigueiraMSFT: Can you do something like the code below?

    var nfi = formatCulture.NumberFormat;
    formatString = formatString.Substring(endIdx + 1);
    revFormatStr = formatString.Replace(nfi.NumberDecimalSeparator, "\uFEFF");
    revFormatStr = revFormatStr.Replace(nfi.NumberGroupSeparator, ",");
    revFormatStr = revFormatStr.Replace("\uFEFF", ".");

This way we can get the appropriate separator for the locale. Even languages that have "space" as the thousands grouping have different spaces: fr-FR has \u202F (narrow no-break space), hu-HU (and many others) has \u00A0 (no-break space)

nguhoa: Thanks for sharing. Update the code. It is ood that this revFormatStr = formatString.Replace(numberCultureFormat.NumberGroupSeparator, "\uFEFF"); does not work when NumberGroupSeparator is space. So I have to add if line to update for space case.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text.txt
```diff
@@ -390,4 +388,44 @@ Error({Kind:ErrorKind.BadLanguageCode})
...
+>> Text(1234.5678, "[$-en-US]#,##0.00", "en-US")
+"1,234.57"
+
+>> Text(1234.5678, "[$-pt-BR]#.##0,00", "en-US")
+"1,234.57"
+
+>> Text(1234.5678, "[$-pt-BR]#.##0,00", "fr-FR")
+"1 234,57"
+
+>> Text(1234.5678, "[$-en-US]#,##0.00", "pt-BR")
```
CarlosFigueiraMSFT, line 410: `ParseJSON("""\u0020""")` is the same thing as a single space, so there's no need to use the JSON function for that. It would be interesting to use it to create the French thousands separator, though (U+202F).

nguhoa: updated here.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (dismissed):
🚢

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text.txt
MikeStall: are these all added?
I don't see the last one

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text.txt
```diff
@@ -390,4 +390,46 @@ Error({Kind:ErrorKind.BadLanguageCode})
...
+"1.234,57"
+
+// If not closed by a ']', then not interpreted as a format specifier
+>> Text(1234.5678, "[$-en-US#,##0.00", "en-US")
+"[$-en-US1,234.57"
+
+>> Text(1234.5678, "[$-a]0,0", "en-US")
+Error({Kind:ErrorKind.InvalidArgument})
+
+>> Text(1234.5678,"[$-en-US]0 # # # !!!", "en-US")
```
MikeStall, line 420: > Text(1234.5678,"[$-en-US]0 # # # !!!", "en-US") [](http://example.com/codeflow?start=2&length=48)

Can you add this test case:


>> Text(1234.5678,"abc[$-en-US]0 # # # !!!", "en-US")

MikeStall: also:

Text(1234.5678,"[$-en-US]")

MikeStall (changes requested):
:clock1:

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text.txt
MikeStall: Text(1234.5678,"[$-]")

MikeStall: Text(1234.5678,"[$-a")

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -365,8 +365,43 @@ public static bool TryText(FormattingInfo formatInfo, IRContext irContext, Formu
...
+                        return false;
+                    }
+
+                    var numberCultureFormat = formatCulture.NumberFormat;
+                    formatString = formatString.Substring(endIdx + 1);                    
+                    revFormatStr = formatString.Replace(numberCultureFormat.NumberGroupSeparator, "\uFEFF");
+
+                    if (string.IsNullOrWhiteSpace(numberCultureFormat.NumberGroupSeparator))
+                    {
+                        revFormatStr = revFormatStr.Replace(" ", "\uFEFF").Replace("\u202F", "\uFEFF");
```
MikeStall, line 397: >"\uFEFF" [](http://example.com/codeflow?start=65&length=8)

nit: can we assign this to a const string field, and document what it means. That will make it easier to read code (and verify all of these literals are indeed the same unicode char)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -1111,6 +1171,20 @@ internal static bool TryGetInt(FormulaValue value, out int outputValue)
 
             outputValue = (int)inputValue;
             return true;
+        }
+
+        private static string GetFormatString(object value, string formatString, string revFormatStr, CultureInfo formatCulture, CultureInfo enUsCulture)
```
MikeStall, line 1176: >object [](http://example.com/codeflow?start=46&length=6)

can this be IFormattable directly?

MikeStall, line 1176: >string formatString, string revFormatStr [](http://example.com/codeflow?start=60&length=40)

what is the difference between these two?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -341,32 +345,57 @@ public static FormulaValue Text(FormattingInfo formatInfo, IRContext irContext,
                 return CommonErrors.GenericInvalidArgument(irContext, string.Format(CultureInfo.InvariantCulture, customErrorMessage, formatSize));
             }
 
-            if (formatString != null && !TextFormatUtils.IsValidFormatArg(formatString, out bool hasDateTimeFmt, out bool hasNumberFmt))
+            if (formatString != null && !TextFormatUtils.IsValidFormatArg(formatString, out hasDateTimeFmt, out hasNumberFmt, out startIdx, out endIdx))
```
MikeStall, line 348: >out startIdx, out endIdx [](http://example.com/codeflow?start=126&length=24)

combine to single cultureParameter?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -341,32 +345,57 @@ public static FormulaValue Text(FormattingInfo formatInfo, IRContext irContext,
...
             }
 
-            var isText = TryText(formatInfo, irContext, args[0], formatString, out StringValue result);
+            var isText = TryText(formatInfo, irContext, args[0], hasDateTimeFmt, hasNumberFmt, formatString, startIdx, endIdx, out StringValue result);
 
             return isText ? result : CommonErrors.GenericInvalidArgument(irContext, StringResources.Get(TexlStrings.ErrTextInvalidFormat, culture.Name));
         }
 
-        public static bool TryText(FormattingInfo formatInfo, IRContext irContext, FormulaValue value, string formatString, out StringValue result)
+        public static bool TryText(FormattingInfo formatInfo, IRContext irContext, FormulaValue value, bool hasDateTimeFmt, bool hasNumberFmt, string formatString, int startIdx, int endIdx, out StringValue result)
```
MikeStall, line 359: Are there any values of CCC and XXX where these are not the same?

>> Text(1234.5678, "[$-CCCCCCCC]XXXXXXXX")
  
>> Text(1234.5678, "XXXXXXXX", "CCCCCCCC")

MikeStall: If yes, then we should definitely have a test case for that. 
If no, then we should normalize them to a single (formatString, CultureInfo) in this function.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: updated

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: removed this.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: updated

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: updated to handle single format

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -35,7 +52,7 @@ public static bool IsValidFormatArg(string formatArg, out bool hasDateTimeFmt, o
                 hasNumericFmt = formatWithoutZeroSubseconds.IndexOfAny(new char[] { '0', '#' }) >= 0;
             }
 
-            if (hasDateTimeFmt && hasNumericFmt)
+            if ((hasDateTimeFmt && hasNumericFmt) || (startIdx > 0) || (startIdx == 0 && endIdx <= 0))
```
MikeStall, line 55: >(startIdx == 0 && endIdx <= 0) [](http://example.com/codeflow?start=71&length=30)

can this check be above, on line 40.

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -2,30 +2,47 @@
...
+                    if (string.IsNullOrEmpty(formatCultureName))
+                    {
+                        return false;
+                    }
                 }
             }
 
-            hasDateTimeFmt = formatArg.IndexOfAny(new char[] { 'm', 'M', 'd', 'D', 'y', 'Y', 'h', 'H', 's', 'S', 'a', 'A', 'p', 'P' }) >= 0;
+            hasDateTimeFmt = (formatArg.IndexOfAny(new char[] { 'm', 'M', 'd', 'D', 'y', 'Y', 'h', 'H', 's', 'S', 'a', 'A', 'p', 'P' }) >= 0) 
+                && (formatArg.IndexOf("us", StringComparison.OrdinalIgnoreCase) < 0);
```
MikeStall, line 44: > (formatArg.IndexOf("us", StringComparison.OrdinalIgnoreCase) < 0); [](http://example.com/codeflow?start=18&length=67)

still needed?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -341,32 +346,55 @@ public static FormulaValue Text(FormattingInfo formatInfo, IRContext irContext,
...
+                    return false;
+                }
+
+                formatString = formatArg;
+            }
+    
+            if (!string.IsNullOrEmpty(formatArg))
+            {
+                // Get en-US numeric format string
+                // \uFEFF is a the zero width no-break space codepoint. This will be used to swap with number group seperator character.
```
MikeStall, line 382: >a the [](http://example.com/codeflow?start=29&length=5)

nit: typo

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -341,32 +346,55 @@ public static FormulaValue Text(FormattingInfo formatInfo, IRContext irContext,
...
             }
 
-            var isText = TryText(formatInfo, irContext, args[0], formatString, out StringValue result);
+            var isText = TryText(formatInfo, irContext, args[0], hasDateTimeFmt, hasNumberFmt, formatString, formatCultureName, formatArg, out StringValue result);
 
             return isText ? result : CommonErrors.GenericInvalidArgument(irContext, StringResources.Get(TexlStrings.ErrTextInvalidFormat, culture.Name));
         }
 
-        public static bool TryText(FormattingInfo formatInfo, IRContext irContext, FormulaValue value, string formatString, out StringValue result)
+        public static bool TryText(FormattingInfo formatInfo, IRContext irContext, FormulaValue value, bool hasDateTimeFmt, bool hasNumberFmt, string formatString, string formatCultureName, string formatArg, out StringValue result)
```
MikeStall, line 360: > string formatString [](http://example.com/codeflow?start=142&length=20)

why do we have both formatString and formatArg?

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -2,30 +2,47 @@
...
+using System.Xml;
 
 namespace Microsoft.PowerFx.Core.Utils
 {
     internal sealed class TextFormatUtils
     {
         private static readonly Regex _formatWithoutZeroSubsecondsRegex = new Regex(@"[sS]\.?(0+)", RegexOptions.Compiled);
 
-        public static bool IsValidFormatArg(string formatArg, out bool hasDateTimeFmt, out bool hasNumericFmt)
+        public static bool IsValidFormatArg(string formatString, out string formatCultureName, out string formatArg, out bool hasDateTimeFmt, out bool hasNumericFmt)
```
MikeStall, line 16: > out string formatCultureName, out string formatArg, out bool hasDateTimeFmt, out bool hasNumericFmt [](http://example.com/codeflow?start=64&length=100)

With 4 out params, consider TryParseFormatInfo(formatString, out TextFormatArgs info);

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
nguhoa: updated.

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
nguhoa: removed

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
nguhoa: updated

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: updated. use one.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -2,15 +2,85 @@
...
+
+            if (textFormatArgs.HasDateTimeFmt && textFormatArgs.HasNumericFmt)
+            {
+                return false;
+            }
+
+            return true;
+        }
+
+        public static bool IsValidCompiledTimeFormatArg(string formatArg)
```
MikeStall, line 83: >IsValidCompiledTimeFormatArg [](http://example.com/codeflow?start=27&length=28)

add comment to explain difference with IsValidFormatArg. 

IsLegacyValidCompiledTimeFormatArg() 
eg: 
This is just called in pre-V1 scenarios by Text function's CheckTypes(). 
For V1, merge this check with runtime check.

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -2,15 +2,85 @@
...
 
 using System;
+using System.Collections.Immutable;
+using System.Globalization;
 using System.Text.RegularExpressions;
+using System.Xml;
 
 namespace Microsoft.PowerFx.Core.Utils
 {
+    internal class TextFormatArgs
```
MikeStall, line 12: >internal [](http://example.com/codeflow?start=4&length=8)

comment

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -353,39 +360,56 @@ public static FormulaValue Text(FormattingInfo formatInfo, IRContext irContext,
...
             {
                 return false;
             }
 
-            Contract.Assert(StringValue.AllowedListConvertToString.Contains(value.Type));
+            if (!string.IsNullOrEmpty(formatString) && textFormatArgs.HasNumericFmt)
+            {
+                // Get en-US numeric format string.
+                // \uFEFF is the zero width no-break space codepoint. This will be used to swap with number group seperator character.
+                string numberGroupSeperator = "\uFEFF";
```
MikeStall, line 393: >string [](http://example.com/codeflow?start=16&length=6)

'const; ?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -353,39 +360,56 @@ public static FormulaValue Text(FormattingInfo formatInfo, IRContext irContext,
...
+
+                formatString = formatString.Replace(numberCultureFormat.NumberGroupSeparator, numberGroupSeperator);
+                if (string.IsNullOrWhiteSpace(numberCultureFormat.NumberGroupSeparator))
+                {
+                    formatString = formatString.Replace(" ", numberGroupSeperator).Replace("\u202F", numberGroupSeperator);
+                }
+
+                formatString = formatString.Replace(numberCultureFormat.NumberDecimalSeparator, ".");
+                formatString = formatString.Replace(numberGroupSeperator, ",");
+            }
```
MikeStall, line 404: can this be moved into textFormatArgs?
It's ultimately parsing...

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_Format_Decimal.txt
```diff
@@ -0,0 +1,23 @@
+#SETUP: disable:NumberIsFloat
```
MikeStall, line 1: >disable:NumberIsFloat [](http://example.com/codeflow?start=8&length=21)

why does decimal vs. float matter?

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -2,15 +2,85 @@
...
+        internal string FormatCultureName { get; set; }
+
+        internal string FormatArg { get; set; }
+
+        internal bool HasDateTimeFmt { get; set; }
+
+        internal bool HasNumericFmt { get; set; }
+    }
+
     internal sealed class TextFormatUtils
```
MikeStall, line 23: >TextFormatUtils [](http://example.com/codeflow?start=26&length=15)

We should be sure our tests give 100% code coverage on this file.

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
nguhoa: updated

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
nguhoa: added comment.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: moved it.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: updated

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_Format_Decimal.txt
nguhoa: removed this and created based on PowerFxV1 tag instead.

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
nguhoa: run code coverage and got all lines cover

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -2,15 +2,147 @@
...
 {
+    /// <summary>
+    /// Definition for format string object ([$-FormatCultureName]FormatArg).
+    /// </summary>
+    internal class TextFormatArgs
+    {
+        /// <summary>
+        /// Culture name of the format string.
+        /// </summary>
+        internal string FormatCultureName { get; set; }
```
MikeStall, line 20: >internal [](http://example.com/codeflow?start=8&length=8)

super nit: these can be public. The class is already internal, we can have public members on internal class.

nguhoa: updated

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -42,5 +170,21 @@ public static bool IsValidFormatArg(string formatArg, out bool hasDateTimeFmt, o
 
             return true;
         }
+
+        private static bool TryGetCulture(string name, out CultureInfo value)
```
anderson-joyle, line 174: >private static bool TryGetCulture(string name, out CultureInfo value) [](http://example.com/codeflow?start=8&length=69)

There is a similar block at [https://github.com/microsoft/Power-Fx/blob/f660b11723af19cb94f3ea67f6b76257a7811c6f/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs#L557](https://github.com/microsoft/Power-Fx/blob/f660b11723af19cb94f3ea67f6b76257a7811c6f/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs#L557).
Is is possible you can remove it form there and call this nice helper?

nguhoa: Remove there and update all references.

LucGenetier:
✅ No public API change.

PR #1494, opened 2023-05-09
Fixing small null ref errors that i found from the changes in Pr #1420

akshar2401:
Forgot to handle few cases where null ref could be thrown in this pr https://github.com/microsoft/Power-Fx/pull/1420. This pr handles those null ref scenarios. It is a small pr.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
+++ b/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
@@ -493,7 +493,12 @@ namespace Microsoft.PowerFx.LanguageServerProtocol
 
             var tokenTypesToSkip = ParseTokenTypesToSkipParam(queryParams?.Get("tokenTypesToSkip"));
             var scope = _scopeFactory.GetOrCreateInstance(semanticTokensParams.TextDocument.Uri);
-            var result = scope.Check(expression);
+            var result = scope?.Check(expression);
+            if (scope == null)
+            {
+                SendEmptySemanticTokensResponse(id);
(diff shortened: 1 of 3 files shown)


src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -493,7 +493,12 @@ private void HandleSemanticTokens<T>(string id, T semanticTokensParams, string m
 
             var tokenTypesToSkip = ParseTokenTypesToSkipParam(queryParams?.Get("tokenTypesToSkip"));
             var scope = _scopeFactory.GetOrCreateInstance(semanticTokensParams.TextDocument.Uri);
-            var result = scope.Check(expression);
+            var result = scope?.Check(expression);
+            if (scope == null)
+            {
+                SendEmptySemanticTokensResponse(id);
+                return;
+            }
```
CarlosFigueiraMSFT, line 502: You can move this check to before the previous line, and after that you won't need to use the `?` null-propagating operator in the Check call.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
akshar2401: This is a nice catch. I meant to check for result == null and not just scope.

jas-valgotar (dismissed):

jas-valgotar (approved):

LucGenetier:
✅ No public API change.

PR #1496, opened 2023-05-10
Better warning message for delegation

anderson-joyle:
To support DV issue https://github.com/microsoft/Power-Fx-Dataverse/issues/183.

### DIFF
--- a/src/strings/PowerFxResources.en-US.resx
+++ b/src/strings/PowerFxResources.en-US.resx
@@ -5139,7 +5139,7 @@
     <comment>Error Message.</comment>
   </data>
   <data name="WrnDelagationTableNotSupported" xml:space="preserve">
-    <value>Delegation warning. Table '{0}' part might not work correctly on large tables (>{1} rows).</value>
+    <value>This operation on table '{0}' may not work if it has more than {1} rows.</value>
     <comment>Warning message.</comment>
   </data>
   <data name="WrnDelagationPredicate" xml:space="preserve">
--- a/src/strings/PowerFxResources.en-US.resx
+++ b/src/strings/PowerFxResources.en-US.resx
@@ -5139,7 +5139,7 @@
     <comment>Error Message.</comment>
   </data>
   <data name="WrnDelagationTableNotSupported" xml:space="preserve">
-    <value>This operation may not work as expected on large tables (>{0} rows). We are working on delegating formulas to support large data sets.</value>
+    <value>Delegation warning. Table '{0}' part might not work correctly on large data sets (>{1} rows).</value>
     <comment>Warning message.</comment>
   </data>
   <data name="WrnDelagationPredicate" xml:space="preserve">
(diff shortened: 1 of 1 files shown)


LucGenetier:
✅ No public API change.

MikeStall (dismissed):

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5139,7 +5139,7 @@
     <comment>Error Message.</comment>
   </data>
   <data name="WrnDelagationTableNotSupported" xml:space="preserve">
-    <value>Delegating this operation on table '{0}' is not supported.</value>
```
MikeStall, line 5142: We're losing the table name? That may be useful to show

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5139,7 +5139,7 @@
...
+    <value>This operation may not work as expected on large tables (>{0} rows). We are working on delegating formulas to support large data sets.</value>
```
MikeStall, line 5142: "We are working on delegating formulas to support large data sets" - wording here probably needs to be tweaked. 

What error does Power Apps show in this case?

LucGenetier:
✅ No public API change.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5139,7 +5139,7 @@
     <comment>Error Message.</comment>
   </data>
   <data name="WrnDelagationTableNotSupported" xml:space="preserve">
```
CarlosFigueiraMSFT, line 5141: >WrnDelagationTableNotSupported [](http://example.com/codeflow?start=14&length=30)

Where is this used? If we are changing the string from expecting one argument to expecting two, we need to pass the additional argument to it.

MikeStall (dismissed):

src/strings/PowerFxResources.en-US.resx
anderson-joyle: This is going to be used from Dataverse project side. Please check PR https://github.com/microsoft/Power-Fx-Dataverse/pull/187.

CarlosFigueiraMSFT: We shouldn't have a resource from this project that is not used here. This should be moved to the PowerFx-Dataverse project instead.

LucGenetier:
✅ No public API change.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5139,7 +5139,11 @@
...
-    <value>Delegating this operation on table '{0}' is not supported.</value>
+    <value>Delegation warning. Table '{0}' part might not work correctly.</value>
```
CarlosFigueiraMSFT, line 5142: >Delegation warning [](http://example.com/codeflow?start=11&length=18)

The error type for this error is already a warning - we probably don't need to say it again in the message.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #1499, opened 2023-05-11
Prevent mutation functions from updating immutable values

CarlosFigueiraMSFT:
Currently the Patch/Collect functions can update values that should be immutable (like scope variables defined in a With record). This change defines a "mutable" concept in the binder, which can be checked by mutation functions so that they won't update those values.


src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationFunctionsTests.cs
```diff
@@ -224,6 +225,24 @@ public void MutationCheckFailTests(string expression)
...
         }
 
+        // Immutable arguments
+        [Theory]
+        [InlineData("Patch([1,2,3], {Value:1}, {Value:9})")]
+        [InlineData("With({x:[1,2,3]},Patch(x, {Value:1}, {Value:9}))")]
+        [InlineData("Patch(namedFormula, {Value:1}, {Value:9})")]
+
+        [InlineData("Collect([1,2,3], {Value:9})")]
+        [InlineData("With({x:[1,2,3]},Collect(x, {Value:9}))")]
```
MikeStall, line 235: > [InlineData("With({x:[1,2,3]},Collect(x, {Value:9}))")] [](http://example.com/codeflow?start=7&length=56)

beware - this works in power apps today

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/PatchFunction.cs
```diff
@@ -52,6 +53,18 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
...
+        {
+            base.CheckSemantics(binding, args, argTypes, errors);
+            if (!binding.IsMutable(args[0]))
+            {
+                errors.EnsureError(
+                    args[0],
+                    new ErrorResourceKey("ErrorResource_MutationFunctionCannotBeUsedWithImmutableValue"),
+                    this.Name);
+            }
+        }
```
MikeStall, line 66: can we share with Collect? 


Does Set() need this check? Here's what it does today:
[https://github.com/microsoft/Power-Fx/blob/813e341e3996b12312500bbf31ce869549a3039e/src/libraries/Microsoft.PowerFx.Interpreter/Functions/SetFunction.cs#L72](https://github.com/microsoft/Power-Fx/blob/813e341e3996b12312500bbf31ce869549a3039e/src/libraries/Microsoft.PowerFx.Interpreter/Functions/SetFunction.cs#L72)

src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationFunctionsTests.cs
CarlosFigueiraMSFT: Yep, which is why the new mutation bit is only checked if we are using PowerFxV1CompatibilityRules

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/PatchFunction.cs
CarlosFigueiraMSFT: Set can also be used to define a variable, so the check would be different in that function.

LucGenetier:
✅ No public API change.

jas-valgotar (dismissed):

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -3561,6 +3586,9 @@ public override void PostVisit(DottedNameNode node)
                 _txb.SetStateful(node, _txb.IsStateful(node.Left));
                 _txb.SetContextual(node, _txb.IsContextual(node.Left));
 
+                // An `a.b` expression will be mutable if `a` is mutable
+                _txb.SetMutable(node, _txb.IsMutable(node.Left));
```
jas-valgotar, line 3590: >_txb.SetMutable(node, _txb.IsMutable(node.Left)); [](http://example.com/codeflow?start=16&length=49)

What happens for expression like,
`First(DataverseTable).OneToManyField`
Should the result be mutable?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/PatchFunction.cs
MikeStall: 2 issues here:
1. Sharing with Collect + Patch  
2. Set - in Fx, the variable is already defined by time we call this. For analysis in PA, I get we have an earlier phase that creates definitions. 


Can we do 1 here?

src/tests/Microsoft.PowerFx.Core.Tests/BinderTests.cs
```diff
@@ -0,0 +1,139 @@
...
+                TableType.Empty().Add("Value", FormulaType.Number),
+                mutable: true);
+            var engine = new Engine(config);
+            var checkResult = engine.Check(expression);
+            Assert.True(checkResult.IsSuccess);
+            var binding = checkResult.Binding;
+            Assert.False(binding.IsMutable(binding.Top));
+        }
+
+        private class FakePatchFunction : BuiltinFunction
```
MikeStall, line 95: >FakePatchFunction [](http://example.com/codeflow?start=22&length=17)

What is significant about this override as opposed to other functions?

src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationFunctionsTests.cs
MikeStall: What is the alternative expression we'd recommend they write? Eg, do they need a Set in there or something?

src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationFunctionsTests.cs
```diff
@@ -224,6 +225,24 @@ public void MutationCheckFailTests(string expression)
...
+        [InlineData("Collect(namedFormula, {Value:9})")]
+        public void MutationCheckFailImmutableNodesTests(string expression)
```
MikeStall, line 237: >MutationCheckFailImmutableNodesTests [](http://example.com/codeflow?start=20&length=36)

case for Remove?

src/tests/Microsoft.PowerFx.Core.Tests/BinderTests.cs
CarlosFigueiraMSFT: Nothing really, just to showcase exactly what we want to block in the test below, without referencing the Patch function from this repository (since it's defined in MS.PowerFx.Interpreter, not here in Core). Once we move the *definition* of Patch from the interpreter to the core, we should be able to use that.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
CarlosFigueiraMSFT: At this point it is mutable; is there a way to know whether the left node, if it's a FirstNameNode, refers to a local or connected data source?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/PatchFunction.cs
CarlosFigueiraMSFT: Updating...

MikeStall (approved):
:shipit:

anderson-joyle (dismissed):
:shipit:

gregli-msft (dismissed):

gregli-msft (approved):

LucGenetier:
✅ No public API change.

PR #1501, opened 2023-05-11
Fixes FromJson issue with single column table

jas-valgotar:


src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -163,14 +163,12 @@ internal static FormulaValue TableFromJsonArray(JsonElement array, TableType tab
         {
             Contract.Assert(array.ValueKind == JsonValueKind.Array);
 
-            var records = new List<RecordValue>();            
-            bool isArray = tableType?._type.IsColumn == true;
-            FormulaType ft = isArray ? tableType.ToRecord().GetFieldType("Value") : tableType?.ToRecord();
```
jas-valgotar, line 168: This check was incorrect, even if you get the type of single column table, `JsonElemnt.Kind` is still `Object` for each element and would conflict when calling `FromJson()` at line 171.

MikeStall (dismissed):

CarlosFigueiraMSFT:
>                 JsonElement propBag = document.RootElement;

Not related to your changes, but can you remove this unused line?

---
Refers to: src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs:31 in 1a27ea3. [](commit_id = 1a27ea313ff45266b12cf1e2ccb2714418e1b351, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -163,14 +163,12 @@ internal static FormulaValue TableFromJsonArray(JsonElement array, TableType tab
...
+            var records = new List<RecordValue>();
 
             for (var i = 0; i < array.GetArrayLength(); ++i)
             {
                 JsonElement element = array[i];
-                var val = GuaranteeRecord(FromJson(element, ft, numberIsFloat));
+                var val = GuaranteeRecord(FromJson(element, tableType?.ToRecord(), numberIsFloat));
```
CarlosFigueiraMSFT, line 171: >tableType?.ToRecord() [](http://example.com/codeflow?start=60&length=21)

Consider moving this call to outside the loop - the value will be the same for all elements.

jas-valgotar: Done

jas-valgotar:
>                 JsonElement propBag = document.RootElement;

removed.

---
In reply to: [1544788100](https://github.com/microsoft/Power-Fx/pull/1501#issuecomment-1544788100) [](http://example.com/codeflow?ancestors=1544788100,1544788100)

---
Refers to: src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs:31 in 1a27ea3. [](commit_id = 1a27ea313ff45266b12cf1e2ccb2714418e1b351, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
jas-valgotar: Turns out, Check was needed but with some extra checking.

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -163,8 +157,12 @@ internal static FormulaValue TableFromJsonArray(JsonElement array, TableType tab
         {
             Contract.Assert(array.ValueKind == JsonValueKind.Array);
 
-            var records = new List<RecordValue>();            
-            bool isArray = tableType?._type.IsColumn == true;
+            var records = new List<RecordValue>();
+
+            // Single Column table (e.g. [1,2,3]) Pattern for table is unique
+            // since in that case nested elements are not object and hence needs to be handled differently.
+            var nestedElementsAreObjects = array.EnumerateArray().Any(nestedElement => nestedElement.ValueKind == JsonValueKind.Object);
```
CarlosFigueiraMSFT, line 164: >.Any( [](http://example.com/codeflow?start=65&length=5)

It would be good to have tests with mixed-type tables, something like `[1,{"a":2}]`. It should fail, but not throw.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
jas-valgotar: `nestedElementsAreObjects` is only used when the formulatype is single column table, and since single column table needs to be homogenous in Fx, I am not sure if I can test that scenario.

MikeStall (approved):

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1489, opened 2023-05-08
Only allows coercion to first argument type on tables

CarlosFigueiraMSFT:
We have update the Table function (and the `[...]` syntax) to favor the first type. But the current union logic would allow coercion from the first value into subsequent ones, which would cause issues like in the example `[ErrorKind.Div0, 12]` - the resulting type is `![Value:ErrorKindOptionSet], but the IR would try to coerce the data to number (since numeric option set can be coerced to number, but not the other way around). This change forces, within those contexts, that the union should only happen to the left type, not the right one.

Fixes #1477

PR #1502, opened 2023-05-12
Use 'prefer left type' if PFx v1 is enabled as well

CarlosFigueiraMSFT:
Follow-up to #1489 - using the 'only coerce to left type' on the Table function either if StronglyTypedEnums is enabled (the root cause of the bug it fixed), or if we are using the Power Fx V1 compat rules (since it moved to a "first-type" rule)

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -5240,7 +5240,13 @@ namespace Microsoft.PowerFx.Core.Binding
 
                     if (usePFxV1CompatRules)
                     {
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -5240,7 +5240,13 @@ public override void PostVisit(TableNode node)
 
                     if (usePFxV1CompatRules)
                     {
-                        if (DType.TryUnionWithCoerce(exprType, childType, usePowerFxV1CompatibilityRules: true, coerceToLeftTypeOnly: true, out var returnType, out var needCoercion))
+                        if (DType.TryUnionWithCoerce(
```
CarlosFigueiraMSFT, line 5243: Formatting change only, no code change here

LucGenetier:
✅ No public API change.

Shpakh-MSFT (approved):
:shipit:

PR #1504, opened 2023-05-15
Add refresh function

LucGenetier:


src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Refresh.cs
```diff
@@ -0,0 +1,125 @@
...
+            Contracts.AssertValue(binding);
+            Contracts.AssertValue(args);
+            Contracts.AssertAllValues(args);
+            Contracts.AssertValue(argTypes);
+            Contracts.AssertAllValid(argTypes);
+            Contracts.Assert(args.Length == argTypes.Length);
+            Contracts.AssertValue(errors);
+            Contracts.Assert(MinArity <= args.Length && args.Length <= MaxArity);
+
+            if (!argTypes[0].IsAggregate)
```
anderson-joyle, line 79: >if (!argTypes[0].IsAggregate) [](http://example.com/codeflow?start=12&length=29)

Why checking types at `CheckSemantics` and not `CheckTypes`?

anderson-joyle:
>         public static void EnableMutationFunctions(this SymbolTable symbolTable)

Since `Refresh` is a mutation function, shouldn't it be added to this list so host can explicitly activate it?

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Environment/PowerFxConfigExtensions.cs:39 in f3a420c. [](commit_id = f3a420c10739ad05a0e1cdd34c50b1452ef2bd3c, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Entities/IRefreshable.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Core.Entities
+{
+    public interface IRefreshable
```
MikeStall, line 8: this needs some comment - at least to note who calls it?  Ie, who should implement it?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Refresh.cs
```diff
@@ -0,0 +1,125 @@
...
+using Microsoft.PowerFx.Core.Localization;
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Syntax;
+
+namespace Microsoft.PowerFx.Core.Texl.Builtins
+{
+    // Refresh(source:*[...]) : b
+    // Refresh(source:![...]) : b
+    internal sealed class RefreshFunction : BuiltinFunction
```
MikeStall, line 21: >RefreshFunction [](http://example.com/codeflow?start=26&length=15)

I'm assuming this is straight from PA. Any changes?

MikeStall:
>         public static void EnableMutationFunctions(this SymbolTable symbolTable)

Refresh is useful even in readonly scenarios - since it's about refreshing when external changes happen. 

It's safe to have Refresh () function always enabled. 

---
In reply to: [1548023899](https://github.com/microsoft/Power-Fx/pull/1504#issuecomment-1548023899) [](http://example.com/codeflow?ancestors=1548023899,1548023899)

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Environment/PowerFxConfigExtensions.cs:39 in f3a420c. [](commit_id = f3a420c10739ad05a0e1cdd34c50b1452ef2bd3c, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -639,6 +636,16 @@ private static FormulaValue SortOptionSet(List<(DValue<RecordValue> row, Formula
...
+        }        
+
+        private static async ValueTask<FormulaValue> Refresh(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
+        {
+            if (args[0] is IRefreshable r)
+            {
+                r.Refresh();
+            }
+
+            return FormulaValue.NewBlank();
```
MikeStall, line 648: >return FormulaValue.NewBlank(); [](http://example.com/codeflow?start=12&length=31)

To consider: should this be an error?

Ideally, we'd get compile time errors when passing a non-refreshable item, but I don't think we have the infrastructure for that.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RefreshTests.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        [Fact]
+        public void RefreshTest_InMemoryTable()
+        {
+            PowerFxConfig config = new PowerFxConfig(Features.PowerFxV1);                        
+            RecalcEngine engine = new RecalcEngine(config);
+
+            config.EnableSetFunction();
+            engine.UpdateVariable("t", TableValue.NewTable(RecordType.Empty()));            
+            FormulaValue result = engine.Eval("Set(t, Table()); Refresh(t)", null, new ParserOptions { AllowsSideEffects = true });
```
MikeStall, line 22: >Set(t, Table()); [](http://example.com/codeflow?start=47&length=16)

Is this necessary if set set it via hosting API on line 21?

src/tests/Microsoft.PowerFx.Interpreter.Tests/RefreshTests.cs
```diff
@@ -0,0 +1,60 @@
...
+        public class TestTableValue : TableValue, IRefreshable
+        {
+            public TestTableValue(RecordType recordType) 
+                : base(recordType)
+            {
+            }
+
+            public bool HasBeenRefreshed = false;
+
+            public override IEnumerable<DValue<RecordValue>> Rows => Enumerable.Empty<DValue<RecordValue>>();
```
MikeStall, line 52: this asserts Refresh() is called; but a stronger test would be to ensure the rows are actually updated (ie, there's not another intermediate cache involved, like we used to have before this PR).  
Ie, something like this:

With( { before : First(t).value},  Refresh(t); 
  before &  First(t).value; )

This would also be a good test on the DV side to ensure that scenario is working end-2-end

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Refresh.cs
LucGenetier: All this code is coming from Power App client.
I only changes namespace/usings and line 94 (new interface).
CheckSemantics comes earlier in the chain and seems more adapted for some types of errors (CheckTypes doesn't have the same list of parameters)

src/tests/Microsoft.PowerFx.Interpreter.Tests/RefreshTests.cs
```diff
@@ -0,0 +1,60 @@
...
+using Xunit;
+
+namespace Microsoft.PowerFx.Tests
+{
+    public class RefreshTests
+    {
+        [Fact]
+        public void RefreshTest_InMemoryTable()
+        {
+            PowerFxConfig config = new PowerFxConfig(Features.PowerFxV1);
```
CarlosFigueiraMSFT, line 17: >Features.PowerFxV1 [](http://example.com/codeflow?start=53&length=18)

No need - this is the default now. Same for the next test method

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Refresh.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        public override IEnumerable<Identifier> GetIdentifierOfModifiedValue(TexlNode[] args, out TexlNode identifierNode)
+        {
+            Contracts.AssertValue(args);
+
+            identifierNode = null;
+            if (args.Length == 0)
+            {
+                return null;
+            }
```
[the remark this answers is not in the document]
LucGenetier: All this code is coming from PA Client and this part is the same (aside the curly brackets)

Even if this part is 100% useless, I trust you on that, I always like defensive code as line 110 would throw in case of a 0-length array. 

I can remove it if you think that's really dead code and we'll never ever be in that situation.

For ref: [https://msazure.visualstudio.com/OneAgile/_git/PowerApps-Client?path=/src/Cloud/DocumentServer.Core/Document/Texl/Builtins/Refresh.cs&version=GBmaster&line=107&lineEnd=108&lineStartColumn=1&lineEndColumn=29&lineStyle=plain&_a=contents](https://msazure.visualstudio.com/OneAgile/_git/PowerApps-Client?path=/src/Cloud/DocumentServer.Core/Document/Texl/Builtins/Refresh.cs&version=GBmaster&line=107&lineEnd=108&lineStartColumn=1&lineEndColumn=29&lineStyle=plain&_a=contents)

src/libraries/Microsoft.PowerFx.Core/Entities/IRefreshable.cs
LucGenetier: Always a good idea to have comments

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Refresh.cs
LucGenetier: Correct.
Only line 94 with the new interface

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
CarlosFigueiraMSFT: We should have a runtime error - that would match the Power Apps behavior. We don't always get compile-time errors when we have a non-refreshable item, like in the example below:

    Set(cache, FirstNSharePointList, 10));
    Refresh(If(1<0, SharePointList, cache))

LucGenetier: The documentation doesn't meantion any possible error.
[https://learn.microsoft.com/en-us/power-platform/power-fx/reference/function-refresh](https://learn.microsoft.com/en-us/power-platform/power-fx/reference/function-refresh)

PA Client returns "only managed connections can be refreshed".

We can do the same to be consistent.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RefreshTests.cs
LucGenetier: Not really, I'll remove it.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

MikeStall:
Adding BreakingInternals since RefreshFunction needs to be removed from PA

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Refresh.cs
```diff
@@ -0,0 +1,122 @@
...
+            Contracts.AssertValue(binding);
+            Contracts.AssertValue(args);
+            Contracts.AssertAllValues(args);
+            Contracts.AssertValue(argTypes);
+            Contracts.AssertAllValid(argTypes);
+            Contracts.Assert(args.Length == argTypes.Length);
+            Contracts.AssertValue(errors);
+            Contracts.Assert(MinArity <= args.Length && args.Length <= MaxArity);
+
+            if (args[0] is not FirstNameNode)
```
CarlosFigueiraMSFT, line 79: >args[0] is not FirstNameNode [](http://example.com/codeflow?start=16&length=28)

This should only be done if binding.Features.PowerFxV1CompatibilityRules is true. Otherwise we should keep the same behavior.

MikeStall (dismissed):
:shipit:

CarlosFigueiraMSFT:
>             "RecordInfo", "Refresh", "Relate", "RemoveAll", "RemoveIf", "RenameColumns", "RequestHide", "Reset", "ResetForm", "Revert", "SaveData", "ScanBarcode", "Search", "Select", "SetFocus",

Can be removed from this list

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs:22 in 99745aa. [](commit_id = 99745aac7644350696c63c67e8098440d312e4fd, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Environment/PowerFxConfigExtensions.cs
```diff
@@ -3,7 +3,10 @@
 
 using System;
 using System.Collections.Generic;
+using System.Text;
+using Microsoft.PowerFx.Core;
```
CarlosFigueiraMSFT, line 7: Those new using statements are probably not needed

LucGenetier: Absolutely true :)

MikeStall (approved):

LucGenetier:
>             "RecordInfo", "Refresh", "Relate", "RemoveAll", "RemoveIf", "RenameColumns", "RequestHide", "Reset", "ResetForm", "Revert", "SaveData", "ScanBarcode", "Search", "Select", "SetFocus",

Removed the complete line.


---
In reply to: [1559403253](https://github.com/microsoft/Power-Fx/pull/1504#issuecomment-1559403253) [](http://example.com/codeflow?ancestors=1559403253,1559403253)

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs:22 in 99745aa. [](commit_id = 99745aac7644350696c63c67e8098440d312e4fd, deletion_comment = False)

LucGenetier:
✅ No public API change.

Issue #1503, opened 2023-05-12
Add more tests to With.txt


MikeStall:
https://github.com/microsoft/Power-Fx/blob/main/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/With.txt

// All these return 6.  (verified in Power Apps) 

 

With( { x:5}, With( ThisRecord, ThisRecord.x +1))
With( { x:5}, With( ThisRecord, x +1))
With( { x:5}, With( ThisRecord As T2, T2.x +1))
With( { x:5}, With( ThisRecord As T2, x +1))
With( { x:5} As T1, With( T1 As T2, T1.x +1))
With( { x:5} As T1, With( T1 As ThisRecord, ThisRecord.x +1))

PR #1505, opened 2023-05-15
Add tests for With function

nguhoa:
Fix https://github.com/microsoft/Power-Fx/issues/1503

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/With.txt
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/With.txt
@@ -66,5 +66,6 @@ Time(2,3,32,0)
 >> With( { year : 2023, month : 5, day : 15 }, DateTime( year, month, day, 0, 0, 0 ) )
 DateTime(2023,5,15,0,0,0,0)
 
+// Scope variables defined in With function are case sensitive
 >> With ( { x : 3 , X : 5 }, x ^ X )
 243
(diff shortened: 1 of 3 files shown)


CarlosFigueiraMSFT:
If we're adding more tests, we should also add something like the following, that shows that case sensitivity is important:

    >> With({x:3,X:5},x^X)
    243

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/With.txt
```diff
@@ -19,4 +19,52 @@ Error(Table({Kind:ErrorKind.Div0},{Kind:ErrorKind.Div0}))
...
+>> With( { x : 2, y : 3}, With( { x : x * 2, y : y + 2 }, With( { x : x * 2, y : y + 2 }, With( { x : x * 2, y : y + 2 }, With( { x : x * 2, y : y + 2}, With( { x : x * 2, y : y + 2 }, x * y ) ) ) ) ) )
+832
+
+>> With( { hours : 2, minutes : 3, seconds : 32 }, Time( hours, minutes, seconds ) )
+Time(2,3,32,0)
+
+>> With( { year : 2023, month : 5, day : 15 }, DateTime( year, month, day, 0, 0, 0 ) )
+DateTime(2023,5,15,0,0,0,0)
+
+>> With ( { x : 3 , X : 5 }, x ^ X )
```
CarlosFigueiraMSFT, line 69: Tests can be seen as a _de facto_ specification for the product, so in this case it would be good to add a comment explaining what is being done, something along the lines of `// scope variables defined in With are case sensitive`

CarlosFigueiraMSFT (dismissed):

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/With.txt
nguhoa: Thanks. Added this comment.

CarlosFigueiraMSFT (approved):
🚢

LucGenetier:
✅ No public API change.

Issue #1497, opened 2023-05-10
If function should follow first type rule


gregli-msft:
```
> If( false, {a:1}, {b:2} )
{}

> If( true, {a:1}, {b:2} )
{}
```

The first argument should determine the type, a change from the current intersection semantics in Canvas.  The `true` case should return {a:1}.  The `false` case should return an empty record, as it is today.

CarlosFigueiraMSFT:
Resolved by #1507

PR #1507, opened 2023-05-16
Make the If function follow the first-type rule

CarlosFigueiraMSFT:
Currently the If function uses an intersection rule for the arguments - `If(varBool, {a:1}, {b:2})` returns an empty record since there are no common properties in the two records. With the PowerFxV1 rules, we are changing it to follow the "first-type rule" that is used in other functions, so the type of that expression would be `![a:n]` with the new feature.

Addresses #1497

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/If.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/If.cs
@@ -83,49 +83,88 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
                 }
             }
 
-            var type = ReturnType;
+            var type = context.Features.PowerFxV1CompatibilityRules ? argTypes[1] : ReturnType;
 
-            // Compute the result type by joining the types of all non-predicate args.
-            Contracts.Assert(type == DType.Unknown);
+            // For pre-PowerFxV1, compute the result type by joining the types of all non-predicate args.
+            // For PowerFxV1 compat rules, validate that all non-predicate args can be coerced to the first one
             for (var i = 1; i < count;)
             {
                 var nodeArg = args[i];
                 var typeArg = argTypes[i];
 
-                var typeSuper = DType.Supertype(
-                    type, 
-                    typeArg,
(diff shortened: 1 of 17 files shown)


LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/If_V1Compat.txt
```diff
@@ -0,0 +1,13 @@
+#SETUP: PowerFxV1CompatibilityRules
+
+>> If(1>0, {a:1}, {b:2})
```
MikeStall, line 3: nit - for documentation purposes, can you add a comment about what V1 does differently?

LucGenetier:
✅ No public API change.

MikeStall (approved):

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/If.cs
```diff
@@ -83,49 +82,94 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
...
+                            CollectionUtils.Add(ref nodeToCoercedTypeMap, nodeArg, type);
+                        }
                     }
                     else
                     {
-                        // If the types are incompatible, the result type is void.
+                        // If types are incompatible, result is Void
                         type = DType.Void;
                     }
                 }
```
lesaltzm, line 133: Is this logic shareable with Switch/Coalesce/IfError/..? Should/could it be extracted to a helper?

CarlosFigueiraMSFT: Yes, next step will be to do the same for Switch/IfError. At that time we'll refactor it.

lesaltzm (approved):
:shipit:

PR #1509, opened 2023-05-17
Fxing expression formatting/pretty print bug with reserved keywords

akshar2401:
Earlier this week, Customer reached out to us about a bug when trying to format the expression like "Set(Children; true)" in the formula bar. "Children" is not a reserved keyword and unless the disabled reserved keyword flag is not set when invoking Lexer, "Children" gets identified as an ErrorToken. This causes errors in the AST and we don't format when there are errors and just return the same expression.

To solve this, we need to pass in a Flags.DisabledReservedKeyword flag to the parser and ultimately lexer when invoking them before the pretty printing. There was no ability to pass in custom flags so I extended the "Format" function (which is an entry point for pretty printing) with an optional argument to allow passing custom flags. These changes would be later consumed in Canvas Apps Backend (Would be passing in Flags.DisabledReservedKeyword | Flags.EnableExpressionChaining to Format in Canvas Apps Backend)

@gregli-msft Could you take a look at this and see if this is the right way to solve it

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
@@ -1766,12 +1766,17 @@ namespace Microsoft.PowerFx.Core.Parser
             }
         }
 
-        public static string Format(string text, Flags? flags = null)
+        /// <summary>
+        /// Formats/Pretty Print the given expression to more a readable form.
+        /// </summary>
+        /// <param name="text">Expression Text to format.</param>
+        /// <param name="flags">Optional flags to customize the behavior of underlying lexer and parser. By default, Expression Chaining is enabled.</param>
+        /// <returns>Formatted Expression.</returns>
+        public static string Format(string text, Flags flags = Flags.EnableExpressionChaining)
         {
-            flags = flags ?? Flags.EnableExpressionChaining;
             var result = ParseScript(
                 text,
-                flags: flags.Value);
+                flags: flags);
 
             // Can't pretty print a script with errors.
             if (result.HasError)
--- a/src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
@@ -1766,11 +1766,12 @@ namespace Microsoft.PowerFx.Core.Parser
             }
         }
 
-        public static string Format(string text)
+        public static string Format(string text, Flags? flags = null)
         {
+            flags = flags ?? Flags.EnableExpressionChaining;
             var result = ParseScript(
                 text,
-                flags: Flags.EnableExpressionChaining);
+                flags: flags.Value);
 
             // Can't pretty print a script with errors.
             if (result.HasError)
--- a/src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
@@ -1767,11 +1767,11 @@ namespace Microsoft.PowerFx.Core.Parser
         }
 
         /// <summary>
-        /// Formats/Pretty Print the given expression to more a readable form.
+        /// Formats/Pretty Print the given expression text to more a readable form.
         /// </summary>
-        /// <param name="text">Expression Text to format.</param>
-        /// <param name="flags">Optional flags to customize the behavior of underlying lexer and parser. By default, Expression Chaining is enabled.</param>
-        /// <returns>Formatted Expression.</returns>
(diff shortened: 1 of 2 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -1766,11 +1766,12 @@ internal static string GetTokString(TokKind kind)
             }
         }
 
-        public static string Format(string text)
+        public static string Format(string text, Flags? flags = null)
```
gregli-msft, line 1769: This should probably have a similar signature to ParseSecipt above in this file, which uses `Flags flags = Flags.None` above.  We don't generally use nullable types.

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -1766,11 +1766,12 @@ internal static string GetTokString(TokKind kind)
             }
         }
 
-        public static string Format(string text)
+        public static string Format(string text, Flags? flags = null)
         {
+            flags = flags ?? Flags.EnableExpressionChaining;
```
gregli-msft, line 1771: Matched with the comment on the signature, just or in expression chaining, as in `flags |= Flags.EnableExpressionChaining`.  Unless there is a case, which doesn't exist today, where we should be formatting without this flag.

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -1766,11 +1766,12 @@ internal static string GetTokString(TokKind kind)
...
         }
 
-        public static string Format(string text)
+        public static string Format(string text, Flags? flags = null)
         {
+            flags = flags ?? Flags.EnableExpressionChaining;
             var result = ParseScript(
                 text,
-                flags: Flags.EnableExpressionChaining);
+                flags: flags.Value);
```
gregli-msft, line 1774: Shouldn't this just be `flags` as the parameter?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -1766,11 +1766,17 @@ internal static string GetTokString(TokKind kind)
             }
         }
 
-        public static string Format(string text)
+        /// <summary>
+        /// Formats/Pretty Print the given expression to more a readable form.
+        /// </summary>
+        /// <param name="text">Expression Text to format.</param>
```
CarlosFigueiraMSFT, line 1772: >Text [](http://example.com/codeflow?start=42&length=4)

[minor] `text` (no need for capitalization)

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -1766,11 +1766,17 @@ internal static string GetTokString(TokKind kind)
             }
         }
 
-        public static string Format(string text)
+        /// <summary>
+        /// Formats/Pretty Print the given expression to more a readable form.
+        /// </summary>
+        /// <param name="text">Expression Text to format.</param>
+        /// <param name="flags">Optional flags to customize the behavior of underlying lexer and parser. By default, Expression Chaining is enabled.</param>
+        /// <returns>Formatted Expression.</returns>
```
CarlosFigueiraMSFT, line 1774: >Expression [](http://example.com/codeflow?start=31&length=10)

[minor] `expression` (no need for capitalization)

CarlosFigueiraMSFT (dismissed):
:shipit:

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -1766,11 +1766,17 @@ internal static string GetTokString(TokKind kind)
...
+        /// Formats/Pretty Print the given expression text to more a readable form.
+        /// </summary>
+        /// <param name="text">Expression text to format.</param>
+        /// <param name="flags">Optional flags to customize the behavior of underlying lexer and parser. By default, expression chaining is enabled.</param>
+        /// <returns>Formatted expression text.</returns>
+        public static string Format(string text, Flags flags = Flags.EnableExpressionChaining)
         {
             var result = ParseScript(
                 text,
-                flags: Flags.EnableExpressionChaining);
```
akshar2401, line 1773: Dropping the comment to prevent auto completion until i test this out with a private deployment

LucGenetier:
✅ No public API change.

PR #1512, opened 2023-05-17
Copy on write semantics for InMemoryTables and InMemoryRecords

gregli-msft:
The first argument to Patch, Remove, Collect, ClearCollect, Clear will copy their argument before making a mutation.  They will even do this for deep mutations, such as Patch( First( a ).Value, ... ).  The big changes required:

1. InMemoryTable and InMemoryRecord need the ability to make shallow copies, hence IMutationCopy.  Copying a record will result in the scalar class members being copied, and the enumeration for the fields being copied, but that is all - one level deep (ditto for tables and rows).
1. Field accessor dot and First/Last/Index need to copy as they go down through a data structure.  We don't always want to do this, we don't want to make copies if just reading, so there is a new flag in IRContext to let the interpreter know when we are in a mutation context.
1. Mutation functions now Lazy evaluate their first argument.  This allows mutations to happen in the other arguments before making the copy.
1. No change to Set, as it doesn't mutate anything, the variable just points to a new object.  Reference semantics are still in place until a mutation is made.


LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.CollectionTableValue`1.FindAsync(Microsoft.PowerFx.Types.RecordValue,System.Threading.CancellationToken)

MikeStall:
>             base.ValidateArgumentIsMutable(binding, args[0], errors);

This check was just added by Carlos. 

Should we rolls this into the default behavior and check if based on the new  MutatesArg0 flag?


---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/PatchFunction.cs:59 in b30e42e. [](commit_id = b30e42e3563a2d7c9f2997bf7d9d314aaca97b0b, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValue.cs
```diff
@@ -58,5 +75,20 @@ public string ToExpression()
...
+    /// <summary>
+    /// Indicates that a FormulaValue should be copied before being mutated, for Copy on Write semantics.
+    /// </summary>
+    internal interface IMutationCopy
+    {
+        /// <summary>
+        /// Returns a shallow copy of a FormulaValue. For potentially deep data structures such as a Table or Record,
+        /// this includes the head object and any first level collections for rows or fields respectively.
+        /// It stops there, for example even the records within the rows of a Table are not copied.
+        /// If a shallow copy is conditional or not desired, have ShallowCopy "return this;".
```
MikeStall, line 89: >/// If a shallow copy is conditional or not desired, have ShallowCopy "return this;". [](http://example.com/codeflow?start=8&length=85)

? Could this lead to accidental sharing?
Are callers specifically checked for this?
Would bool TryShallowCopy(out FormulaValue) be more appropriate?

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
```diff
@@ -271,6 +271,11 @@ public override void ToExpression(StringBuilder sb, FormulaValueSerializerSettin
             }
 
             sb.Append("}");
+        }
+
+        internal virtual void ShallowCopyFieldInPlace(string fieldName)
+        {
+            throw new NotImplementedException();
```
MikeStall, line 278: >throw new NotImplementedException(); [](http://example.com/codeflow?start=12&length=36)

should this method just be abstract?

src/libraries/Microsoft.PowerFx.Core/Public/Values/InMemoryRecordValue.cs
```diff
@@ -38,6 +38,24 @@ public InMemoryRecordValue(IRContext irContext, IReadOnlyDictionary<string, Form
...
+        }
+
+        public InMemoryRecordValue(InMemoryRecordValue orig)
+            : this(orig.IRContext, new Dictionary<string, FormulaValue>(orig._mutableFields))
+        {
+        }
+
+        FormulaValue IMutationCopy.ShallowCopy()
+        {
+            return (FormulaValue)new InMemoryRecordValue(this);
```
MikeStall, line 50: >(FormulaValue) [](http://example.com/codeflow?start=19&length=14)

nit - unnecessary cast?

src/tools/Repl/Program.cs
```diff
@@ -617,6 +645,12 @@ public FormulaValue Execute(StringValue option, BooleanValue value)
                     return value;
                 }
 
+                if (option.Value.ToLower(CultureInfo.InvariantCulture) == OptionHashCodes.ToLower(CultureInfo.InvariantCulture))
```
MikeStall, line 648: >ToLower(CultureInfo.InvariantCulture) [](http://example.com/codeflow?start=33&length=37)

nit: ToLowerInvariant

src/libraries/Microsoft.PowerFx.Core/Public/Values/InMemoryRecordValue.cs
```diff
@@ -11,7 +11,7 @@
 namespace Microsoft.PowerFx.Types
 {
     // Represent record backed by known list of values. 
-    internal class InMemoryRecordValue : RecordValue
+    internal class InMemoryRecordValue : RecordValue, IMutationCopy
```
MikeStall, line 14: >IMutationCopy [](http://example.com/codeflow?start=54&length=13)

Is there a case where  a RecordValue should not implement this?

gregli-msft: I am assuming that there are other kinds of records, such as marshaled objects. In those cases, the objects should not be copied and treated as a reference, or mutations should not be allowed, or it should implement IMutationCopy or sone other interface to facilitate copying when mutated. I figured we could figure this all out later, this PR bring surgical and leaving these other cases unchanged, but we can have this conversation now if you would prefer.

src/tools/Repl/Program.cs
gregli-msft: Done.

src/libraries/Microsoft.PowerFx.Core/Public/Values/InMemoryRecordValue.cs
gregli-msft: Done.

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValue.cs
gregli-msft: We could make this change.  But, it makes the logic for the caller more complicated - they have to check if the copy was made and replace the original only in those cases.  Callers shouldn't need to care if a copy was made.  Perhaps rename it to MaybeShallowCopy?

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.CollectionTableValue`1.FindAsync(Microsoft.PowerFx.Types.RecordValue,System.Threading.CancellationToken)

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
MikeStall: I'm assuming we can't get the exception at runtime; so hw do we know *not* to call it in cases like UserInfoRecord?

src/tools/Repl/Program.cs
MikeStall: sorry, realized there's an aequality here. You could even just say:
String.Equals(a,b,StringComparison.OrdinalIgnoreCase)

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -722,13 +728,19 @@ public override async ValueTask<FormulaValue> Visit(ResolvedObjectNode node, Eva
...
         {
             if (_symbolValues != null)
             {
                 var value = _symbolValues.Get(slot);
                 if (value != null)
                 {
+                    if (isMutation)
+                    {
+                        value = value.MaybeShallowCopy();
+                        _symbolValues.Set(slot, value);
```
MikeStall, line 741: This replaces in symbol table. 

But I think we need to modify the FormulaValue itself. 
If anybody else is holding onto the FormulaValue, they will have the stale result.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -641,13 +641,19 @@ public override async ValueTask<FormulaValue> Visit(RecordFieldAccessNode node,
...
             {
                 return left;
-            }
+            }
+
+            var record = (RecordValue)left;
+
+            if (node.IRContext.IsMutation)
+            {
+                record.ShallowCopyFieldInPlace(node.Field.Value);
```
MikeStall, line 650: >ShallowCopyFieldInPlace [](http://example.com/codeflow?start=23&length=23)

our default impl of ShallowCopyFieldInPlace throws NotImplemented. 
Could that mean runtime exceptions coming from here?

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
MikeStall: this may also let us avoid the extra LazyNode for arg0.

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.CollectionTableValue`1.FindAsync(Microsoft.PowerFx.Types.RecordValue,System.Threading.CancellationToken)

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
MikeStall: Consider this test: 

Set(x, { t : Table(User)} )
Patch(x.t, First(t), { Name : "...." })

MikeStall: see UserInfoTests for tips

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.CollectionTableValue`1.FindAsync(Microsoft.PowerFx.Types.RecordValue,System.Threading.CancellationToken)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.CollectionTableValue`1.FindAsync(Microsoft.PowerFx.Types.RecordValue,System.Threading.CancellationToken)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.CollectionTableValue`1.FindAsync(Microsoft.PowerFx.Types.RecordValue,System.Threading.CancellationToken)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.CollectionTableValue`1.FindAsync(Microsoft.PowerFx.Types.RecordValue,System.Threading.CancellationToken)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.CollectionTableValue`1.FindAsync(Microsoft.PowerFx.Types.RecordValue,System.Threading.CancellationToken)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.CollectionTableValue`1.FindAsync(Microsoft.PowerFx.Types.RecordValue,System.Threading.CancellationToken)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.CollectionTableValue`1.FindAsync(Microsoft.PowerFx.Types.RecordValue,System.Threading.CancellationToken)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.CollectionTableValue`1.FindAsync(Microsoft.PowerFx.Types.RecordValue,System.Threading.CancellationToken)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.CollectionTableValue`1.FindAsync(Microsoft.PowerFx.Types.RecordValue,System.Threading.CancellationToken)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.CollectionTableValue`1.FindAsync(Microsoft.PowerFx.Types.RecordValue,System.Threading.CancellationToken)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.CollectionTableValue`1.FindAsync(Microsoft.PowerFx.Types.RecordValue,System.Threading.CancellationToken)

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
gregli-msft: I created a separate interface for this, IMutationCopyField.  The binder should prevent mutation of a value which does not support this, but if not, the interface will cast and throw if not found in EvalVisitor.

src/libraries/Microsoft.PowerFx.Core/Public/Values/InMemoryRecordValue.cs
gregli-msft: I've moved this interface to RecordValue with an implementation that copies any record to an InMemoryRecord.  There are test scenarios where the table/record will override these interface implementations, as an extreme example of where a different implementation may be desired.   It is also possible that the DataverseRecordValue will have it's own implementation.  But, bottom line, no there is no record value that should not be copyable.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
gregli-msft: As discussed, yes this does replace the symbol table value, as designed.  This code will need to replace the outcome of the Patch in y, without impacting the existing value in x.
```
> Set( x, [1,2,3] )
x: [1, 2, 3]

> Set( y, x )
y: [1, 2, 3]

> Patch( y, {Value:2}, {Value:9} )
{Value:9}

> x
[1, 2, 3]

> y
[1, 9, 3]
```

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
gregli-msft: See notes above about the interface changes.  Specifically added tests for User scenarios such as this one.  In this case, a copy of User will be made in an InMemoryRecord.

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.CollectionTableValue`1.FindAsync(Microsoft.PowerFx.Types.RecordValue,System.Threading.CancellationToken)

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValue.cs
```diff
@@ -58,5 +75,19 @@ public string ToExpression()
 
             return sb.ToString();
         }
+    }
+
+    /// <summary>
+    /// Indicates that a FormulaValue should be copied before being mutated, for Copy on Write semantics.
+    /// </summary>
+    internal interface IMutationCopy
```
MikeStall, line 83: If DV needs to implement, this would need to be public.

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValue.cs
gregli-msft: Agreed.  We'll look at this again after we get through the Dataverse issues.

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.CollectionTableValue`1.FindAsync(Microsoft.PowerFx.Types.RecordValue,System.Threading.CancellationToken)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.CollectionTableValue`1.FindAsync(Microsoft.PowerFx.Types.RecordValue,System.Threading.CancellationToken)

MikeStall (approved):

PR #1513, opened 2023-05-18
Fxing expression formatting/pretty print bug with reserved keywords (…

akshar2401:
…#1509)

Earlier this week, Customer reached out to us about a bug when trying to format the expression like "Set(Children; true)" in the formula bar. "Children" is not a reserved keyword and unless the disabled reserved keyword flag is not set when invoking Lexer, "Children" gets identified as an ErrorToken. This causes errors in the AST and we don't format when there are errors and just return the same expression.

To solve this, we need to pass in a Flags.DisabledReservedKeyword flag to the parser and ultimately lexer when invoking them before the pretty printing. There was no ability to pass in custom flags so I extended the "Format" function (which is an entry point for pretty printing) with an optional argument to allow passing custom flags. These changes would be later consumed in Canvas Apps Backend (Would be passing in Flags.DisabledReservedKeyword | Flags.EnableExpressionChaining to Format in Canvas Apps Backend)

@gregli-msft Could you take a look at this and see if this is the right way to solve it

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
@@ -1766,11 +1766,17 @@ namespace Microsoft.PowerFx.Core.Parser
             }
         }
 
-        public static string Format(string text)
+        /// <summary>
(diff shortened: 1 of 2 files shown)


jas-valgotar (approved):
:shipit:

PR #1521, opened 2023-05-19
Make the Switch function follow the first-type rule

CarlosFigueiraMSFT:
Currently the Switch function uses an intersection rule for the arguments - Switch(var, 1, {a:1}, {b:2}) returns an empty record since there are no common properties in the two records. With the PowerFxV1 rules, we are changing it to follow the "union with first-type rule" that is used in other functions, so the type of that expression would be ![a:n] with the new feature.

This is similar to #1507 which did the change for the If function.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):

LucGenetier:
✅ No public API change.

Issue #1367, opened 2023-04-14
Use native double and decimal mod operator in the Mod function


gregli-msft:
Should we use the native implementation?  Any differences with Excel or Canvas behavior?  Better or worse operand range?

nguhoa:
Mod(230m, 3m)
Our - Result = 2; In tick time = 7994
C# - Result = 2; In tick time = 927
Mod(3456m, -1m)
Our - Result = 0; In tick time = 42
C# - Result = 0; In tick time = 14
Mod(14.123456m, 1m)
Our - Result = 0.123456; In tick time = 10
C# - Result = 0.123456; In tick time = 5
Mod(-3m,2m)
Our - Result = 1; In tick time = 14
C# - Result = -1; In tick time = 3
Mod(79228162514264337593543m, 3m)
Our - Result = 2; In tick time = 9
C# - Result = 2; In tick time = 2
Mod(79000000000000000000000m, 3m)
Our - Result = 1; In tick time = 7
C# - Result = 1; In tick time = 3
Mod(decimal.MaxValue, 3m)
Our - Result = 0; In tick time = 4
C# - Result = 0; In tick time = 2
Mod(230m, 3.2m)
Our - Result = 2.8; In tick time = 5
C# - Result = 2.8; In tick time = 5
Mod(79228162514264337593543.1234m, 3.5678m)
Our - Result = 1.3776; In tick time = 6
C# - Result = 1.3776; In tick time = 3

nguhoa:
Start from 1e+16 to 1e+20, current mod_float return 0 for most of the case (mostly wrong value). And overflow from 1e+21 or 1e+22.
Mod_decimal work fine.
************ mod 3 ******************

i = 10000000000; % in c# has mod result = 1
i = 10000000000; got mod result = 1
i = 100000000000; % in c# has mod result = 1
i = 100000000000; got mod result = 1
i = 1000000000000; % in c# has mod result = 1
i = 1000000000000; got mod result = 1
i = 10000000000000; % in c# has mod result = 1
i = 10000000000000; got mod result = 1
i = 100000000000000; % in c# has mod result = 1
i = 100000000000000; got mod result = 1
i = 1000000000000000; % in c# has mod result = 1  // 1e15
i = 1000000000000000; got mod result = 1
i = 10000000000000000; % in c# has mod result = 1 // 1e16
i = 10000000000000000; got mod result = 0
i = 1E+17; % in c# has mod result = 1
[...]

PR #1522, opened 2023-05-21
Fix overflow of Mod_Float for large number

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1367

Problem this PR fix:
Start from 1e+16 to 1e+20, current mod_float return 0 for most of the case (mostly wrong value). And overflow from 1e+21 or higher.
This PR use C# % function to do mod for the wrong/overflow issues.

Note on PA (Need to fix as well): mod result is always 0 from 1e18 (mod 123). So, it shared the same problem. mod result is wrong for > 1e17. It did not return overflow error for 1e306 but return 0 (wrong value).


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -769,11 +769,12 @@ public static FormulaValue ModFloat(IRContext irContext, NumberValue arg0nv, Num
...
-            // We validate the reminder is in a valid range.
-            // This is mainly to support very large numbers (like 1E+308) where the calculation could be incorrect
-            if (result < -Math.Abs(arg1) || result > Math.Abs(arg1))
+            // If result is overflow or (arg0 is more than 1e15 and result = 0 - highly wrong result), using C# mod function
+            if ((Math.Abs(arg0) > 1e15 && result == 0) || (result < -Math.Abs(arg1) || result > Math.Abs(arg1)))
             {
-                return CommonErrors.OverflowError(irContext);
+                // C# % result has the same sign with dividend.
+                // Using the following fomular to have the sign of mod result is the same as divisor as in Excel and PA Mod
+                result = (arg0 % arg1) * Math.Sign(arg0) * Math.Sign(arg1);
```
CarlosFigueiraMSFT, line 777: This is wrong - instead of flipping the signal, we should add the divisor (if the result is negative and should be positive) or subtract the divisor (if the result is positive and should be negative)

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Mod_Float.txt
```diff
@@ -0,0 +1,43 @@
...
+1
+
+>> Mod(1E+305, 123)
+112
+
+>> Mod(1.2345E+260, 67.89)
+41.28195726574958
+
+>> Mod(1.7976931348623157E+308, 25)
+18
```
CarlosFigueiraMSFT, line 25: This is also wrong - the result will be 0 (since the dividend is a multiple of 100). Like the previous one, we should not "spec" an incorrect result.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Mod_Float.txt
```diff
@@ -0,0 +1,43 @@
...
+
+>> Mod(1E+13, 123)
+16
+
+>> Mod(1E+18, 123)
+16
+
+>> Mod(1E+25, 123)
+4
```
CarlosFigueiraMSFT, line 34: The correct result is 1

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Mod_Float.txt
```diff
@@ -0,0 +1,43 @@
...
+
+>> Mod(1E+25, -123)
+-119
+
+>> Mod(10^200,-10^-200)
```
CarlosFigueiraMSFT, line 39: >Mod(10^200,-10^-200) [](http://example.com/codeflow?start=3&length=20)

For some of those tests, if we want to leave them here, it may be worth calling out that they are what we show here because of floating point semantics.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -760,20 +760,11 @@ public static FormulaValue ModFloat(IRContext irContext, NumberValue arg0nv, Num
...
-            }
-
-            double result = arg0 - (arg1 * ((long)q));
+            double result = arg0 % arg1;
 
-            // We validate the reminder is in a valid range.
-            // This is mainly to support very large numbers (like 1E+308) where the calculation could be incorrect
-            if (result < -Math.Abs(arg1) || result > Math.Abs(arg1))
+            if (Math.Sign(result) * Math.Sign(arg1) == -1)
             {
```
CarlosFigueiraMSFT, line 766: We should add a comment to explain what we are doing here

nguhoa: Added comment. thanks

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Mod_Float.txt
```diff
@@ -0,0 +1,43 @@
...
+2
+
+>> Mod(-1E+307, -3)
+-1
+
+>> Mod(1.7976931348623157E+300, 3)
+1
+
+>> Mod(1E+305, 123)
+112
```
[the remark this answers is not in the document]
nguhoa: Remove this test.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Mod_Float.txt
nguhoa: removed this as well.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Mod_Float.txt
nguhoa: removed this

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Mod_Float.txt
nguhoa: add comment.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #1523, opened 2023-05-22
add thread safety guard 

MikeStall:
add guard to catch bugs in host where it calls a single-threaded region on multiple threads.


lesaltzm (dismissed):

LucGenetier:
✅ No public API change.

lesaltzm (approved):

nguhoa (approved):
:shipit:

PR #1526, opened 2023-05-24
Fix Refresh function

LucGenetier:


LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):

PR #1527, opened 2023-05-24
Allow Void expressions to be used within ForAll

CarlosFigueiraMSFT:
If a ForAll expression is used within a context where its result is not needed, then it should be able to use void expressions, like in the example below. This change allows it, making the entire result of the ForAll expression a void result.

    ForAll(
        Sequence(4),
        If(
            Mod(Value, 2) = 1,
            Patch( table, Index(table, Value), { IsOdd: true } ),  // this returns a record, incompatible with tables
            Collect( table, { Value: Value, IsOdd: false }, { Value: Value + 1, IsOdd: true }   // this returns a table
        )
    )


jas-valgotar (approved):
:shipit:

LucGenetier:
✅ No public API change.

Issue #1474, opened 2023-05-02
EvalAsync takes CurrentCulture to get FormulaValue


justinpark0916:
We found out that `EvalAsync` function takes a Current Culture when getting a FormulaValue result even though we don't pass CultureInfo to the ParserOptions.

Attached unit tests to reproduce 

```
        [Fact]
        public async Task test()
        {
            // Setup
            CultureInfo culture = new CultureInfo("es-ES");
            CultureInfo.CurrentCulture = culture;
            var parserOptions = new ParserOptions() { AllowsSideEffects = true, NumberIsFloat = true};
            RecalcEngine recalcEngine = new RecalcEngine();

            var check = recalcEngine.Check("Float(\"4,99\")", parserOptions);
            var run = check.GetEvaluator();
            var result = await run.EvalAsync(CancellationToken.None);
            // Look at Result Value => Seems like 4.99
            Assert.Equal(4.99, result.ToObject());
        }

        [Fact]
        public async Task test2()
        {
            // Setup
            CultureInfo culture = new CultureInfo("en-US");
            CultureInfo.CurrentCulture = culture;
            var parserOptions = new ParserOptions() { AllowsSideEffects = true, NumberIsFloat = true };
            RecalcEngine recalcEngine = new RecalcEngine();

            var check = recalcEngine.Check("Float(\"4,99\")", parserOptions);
            var run = check.GetEvaluator();
            var result = await run.EvalAsync(CancellationToken.None);
            // Look at Result Value => Seems like 499
            Assert.Equal((double)499, result.ToObject());
        }
```

MikeStall:
We should be able to add these to our existing tests, and leverage the "DoNotUserCulture" object to verify it's not used.

justinpark0916:
Might be the same root cause but adding this test as well:

```
        [Fact]
        public async Task test()
        {
            CultureInfo culture = new CultureInfo("en-US");
            CultureInfo.CurrentCulture = new CultureInfo("es-ES");
            var parserOptions = new ParserOptions() { AllowsSideEffects = true, NumberIsFloat = true, Culture = culture };
            var expression = "Date(1996; 09; 16)";
            var check = _recalcEngine.Check(expression, parserOptions, null);

            Assert.True(check.IsSuccess);
        }
```

PR #1530, opened 2023-05-26
Replacing some CultureInfo.CurrentCulture occurrences

anderson-joyle:
Issue #1474.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Errors/ErrorUtils.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Errors/ErrorUtils.cs
@@ -30,7 +30,6 @@ namespace Microsoft.PowerFx.Core.Errors
             {
                 try
                 {
-                    // $$$ can't use current culture
                     sb.AppendFormat(locale ?? CultureInfo.InvariantCulture, message, args);
                 }
                 catch (FormatException)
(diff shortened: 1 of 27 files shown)


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Errors/ErrorUtils.cs
```diff
@@ -30,8 +30,7 @@ internal static string FormatMessage(string message, CultureInfo locale, params
             {
                 try
                 {
-                    // $$$ can't use current culture
-                    sb.AppendFormat(locale ?? CultureInfo.CurrentCulture, message, args);
```
MikeStall, line 34: >CurrentCulture [](http://example.com/codeflow?start=58&length=14)

Can we be confident this change isn't breaking power Apps?

MikeStall (changes requested):
:clock1:

MikeStall:
Can we be confident this change isn't breaking power Apps? PA runs in a dedicated process - does PA rely on setting CurrentCulture anywhere?

LucGenetier:
✅ No public API change.

PR #1531, opened 2023-05-26
Adding ability to supply custom binding config to Engine.ComputeBinding() & Fixing Dup String Interp Tokens bug

akshar2401:
This pr handles few things that are mentioned below:
 1) It adds a new virtual overload of `GetDefaultBindingConfig` in Engine.cs that takes in ParserOptions and optional rule scope and returns a binding config that is used by `Engine.ComputeBinding()`. This is done so that Power Apps can supply custom binding config for semantic tokenization. Power Apps never set `BindingConfig.UseThisRecordForRuleScope` to true but changes before this pr calculated the value of `BindingConfig.UseThisRecordForRuleScope` based on the value returned from `Engine.GetRuleScope()`. It would be true when GetRuleScope() returns a non null value. There are properties such as `OnErrors `on AppInfo control in PowerApps that returns a non null value for `GetRuleScope()`. In this case,` useThisRecordForRuleScope `would be true in` Engine.ComputeBinding() `but it should never be true when `Engine.ComputeBinding() `is run from PowerApps.

2) This pr fixes a bug where duplicate string interpolation and island tokens were added when expression has nested string interpolation to the list of tokens returned from Tokenization.Tokenize. This pr fixes that logic and adds tests to validate it. A good example of this bug is an expression like `$"{$"{$"{$"{.12e4}"}"}"}"`. The number of tokens should be 17 but it is 41. Many of them are duplicate with same spans and token types.

3) This pr also fixes one test that happened due to Visual Studio picking up a different line feed for C# multiline strings `@""` . Updates the test to avoid using multiline string. @gregli-msft is the only and first one so far to face this problem. Tests were failing for him. Confirmed with him that avoiding multiline string fixes the test.


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/IntellisenseTests/TokenizationTests.cs
```diff
@@ -15,7 +16,7 @@
 using Microsoft.PowerFx.Syntax;
 using Xunit;
 
-namespace Microsoft.PowerFx.Core.Tests.IntellisenseTests
```
jas-valgotar, line 18: >Microsoft.PowerFx.Core.Tests.IntellisenseTests [](http://example.com/codeflow?start=10&length=46)

why namespace change?

akshar2401: To be consistent with other tests in the same project. The previous namespace was not matching the tests around the ones in this file. The previous namespace was auto generated by the VS and i didn't notice that it was bit different than the other tests.

akshar2401: Feel free to reopen this if you want me to change it back

LucGenetier:
✅ No public API change.

jas-valgotar (approved):
:shipit:

src/tests/Microsoft.PowerFx.Core.Tests/IntellisenseTests/TokenizationTests/ExpectedToken.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        public int StartIndex { get; set; } = SpecialStartIdx;
+
+        public int Length { get; set; }
+
+        public TokenType TokenType { get; set; }
+
+        public int EndIndex => Length + StartIndex;
+
+        public bool CanBeHidden { get; set; } = false;
```
MikeStall, line 24: >set [](http://example.com/codeflow?start=39&length=3)

can these be readonly?  easier to verify in threading cases.

akshar2401: Yeah. I could convert them to readonly. I need setter for StartIndex but i would instead create a new instance of expected token and make start index readonly.

MikeStall (approved):
:shipit:

PR #1533, opened 2023-05-28
Update Abs.txt

NeonG4:
Added 2 tests to check numbers that cannot fit inside an integer-type variable in the Abs() function test cases

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Abs.txt
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Abs.txt
@@ -9,6 +9,9 @@
 >> Abs(2)
 2
 
+>> Abs(2147483649) // testing a number larger than the interger range 
+2147483649
+
 >> Abs(9007199254740991) // testing a number larger than the float range (-9007199254740991 to 9007199254740991, (+/- 2^53 - 1))
 9007199254740991
 
@@ -19,6 +22,9 @@
 
 >> Abs(-2)
 2
+
+>> Abs(-2147483649) // testing a number smaller than the interger range 
+2147483649
 
 >> Abs(-9007199254740991) // testing a number smaller than the float range (-9007199254740991 to 9007199254740991, (+/- 2^53 - 1))
 9007199254740991
--- a/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Abs.txt
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Abs.txt
@@ -9,8 +9,8 @@
 >> Abs(2)
 2
 
->> Abs(2147483648) // testing a number larger than the interger range
-2147483648
+>> Abs(9007199254740991) // testing a number larger than the float range (-9007199254740991 to 9007199254740991, (+/- 2^53 - 1))
+9007199254740991
 
 // Negative Integer Parameters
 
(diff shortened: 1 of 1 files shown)


NeonG4:
> Thank you for the tests; there are a couple of typos, and I suggested a couple of new tests that would go well with this change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Abs.txt
```diff
@@ -9,6 +9,12 @@
 >> Abs(2)
 2
 
+>> Abs(2147483649) // testing a number larger than the interger range 
+2147483649
+
+>> Abs(9007199254740991) // testing a number larger than the float range (-9007199254740991 to 9007199254740991, (+/- 2^53 - 1))
```
CarlosFigueiraMSFT, line 15: Suggestion: `// testing the largest consecutive integer that represented without loss in a floating point`. "larger than the float range" isn't correct, since numbers larger than that can be represented as well.

CarlosFigueiraMSFT: Same for the test with negative number

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Abs.txt
```diff
@@ -9,6 +9,12 @@
 >> Abs(2)
 2
 
+>> Abs(2147483649) // testing a number larger than the interger range
```
CarlosFigueiraMSFT, line 12: typo: interger --> integer

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Abs.txt
NeonG4: Okay, thank you!

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Abs.txt
```diff
@@ -17,6 +23,12 @@
 >> Abs(-2)
 2
 
+>> Abs(-2147483649) // testing a small number
```
CarlosFigueiraMSFT, line 26: This is still a very large number - a large _negative_ number. Typically "small numbers" represent those close to zero (e.g., 1e-23, 0.000000000321, and so on). A better comment would be something like you had before, which explains why you chose this number specifically. Tests can be seen as the _de facto_ specification for a product, and you made a conscious choice in choosing this value (+/- 2^31+2), so it's useful to let the reader know why this is the case.

CarlosFigueiraMSFT: Same for the other cases (2^53-1) - there is a reason why we chose 9007199254740991 and not 1234567890123456.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Abs.txt
```diff
@@ -9,6 +9,12 @@
 >> Abs(2)
 2
 
+>> Abs(2147483647) // testing a large postive number (+/- 2^31) the upper limit for an integer-type variable 
+2147483647
+
+>> Abs(9007199254740992) // testing a really large positive number (+/- 2^53-1) the upper limit for a long-type variable
```
CarlosFigueiraMSFT, line 15: A "long-type" variable (assuming 'long' has its traditional meaning, a 64-bit integer value) could go all the way to 2^63 - 1, so this is not the upper limit for that type. 2^53 is the limit for the largest consecutive integer that represented without loss in a floating point type - which is what Power Fx uses if ParserOptions.NumberIsFloat (https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Core/Parser/ParserOptions.cs#L26) is true.

Issue #1519, opened 2023-05-19
Robustness checks, thread safety 


MikeStall:
Robustness checks (#981) - Several potential races. 
_Writing is single threaded. Reading is multi-threaded._ 

We have some existing tests  (https://github.com/microsoft/Power-Fx/blob/main/src/tests/Microsoft.PowerFx.Core.Tests/ImmutabilityTests.cs) . Expanding that test shows a few suspicious things:

1. DType.AssociatedDataSources returns mutable value
2. DType.ExpandInfo returns mutable value  (IExpandInfo.UpdateEntityInfo)
3. DType.LazyTypeProvider returns mutable value (fixed in #1511)
4. WrappedDerivedRecordType._renameDriver returns mutable value

MikeStall:
See #1520 for new test to scan these failures. 

Currently, there are about 100 failures, mostly false-positives.  We should manually review determine if it is safe?
- Yes, because it's protected under a lock -->  adding ThreadSafeProtectedByLock? 
- Yes, because the field's type is immutable --> add ThreadSafeImmutable to the type, or add to known immutable list. 
- Yes, because the member type could use a safer type  --> update , eg Dict --> IReadOnlyDict
- Yes, manually confirmed --> can we refactor to make it easy to verify? Eg, make the field readonly?

	... if we _can't_ do any of the above ... is it actually safe?

PR #1534, opened 2023-05-29
Update readonly and use immutable type to have thread safe.

nguhoa:
Partly fix issue: https://github.com/microsoft/Power-Fx/issues/1519
- Update readonly and use immutable type to have thread safe.
- Add more [ThreadSafeImmutable] for immutable types in PFx
- Update immutability test to check not only readonly but const as well.
- Add ImmutableDictionary and ImmutableList to immutable type list.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/CustomTexlFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/CustomTexlFunction.cs
@@ -39,7 +39,7 @@ namespace Microsoft.PowerFx
         }
 
         public CustomTexlFunction(string name, FunctionCategories functionCategory, FormulaType returnType, Func<IServiceProvider, FormulaValue[], CancellationToken, Task<FormulaValue>> impl, BigInteger lamdaParamMask, params FormulaType[] paramTypes)
-            : this(name, returnType._type, Array.ConvertAll(paramTypes, x => x._type))
+            : this(name, functionCategory, returnType._type, Array.ConvertAll(paramTypes, x => x._type))
         {
             _impl = impl;
             LamdaParamMask = lamdaParamMask;


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Lexer/Tokens/CommentToken.cs
```diff
@@ -11,13 +11,14 @@ namespace Microsoft.PowerFx.Syntax
     /// </summary>
     public sealed class CommentToken : Token
     {
-        internal bool IsOpenBlock;
+        internal readonly bool IsOpenBlock;
 
-        internal CommentToken(string val, Span span)
+        internal CommentToken(string val, Span span, bool isOpenBlock = false)
```
MikeStall, line 16: >internal [](http://example.com/codeflow?start=8&length=8)

it's possible this might be internal breaking on PA; 
that's ok - trivial fix , just fyi.

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValue.cs
```diff
@@ -11,6 +11,7 @@ namespace Microsoft.PowerFx.Types
     /// Represent a value in the formula expression. 
     /// </summary>
     [DebuggerDisplay("{ToObject().ToString()} ({Type})")]
+    [ThreadSafeImmutable]
```
MikeStall, line 14: >   [ThreadSafeImmutable] [](http://example.com/codeflow?start=1&length=24)

I think there are cases where this (notably Table/Record) can get mutated. Is there a case forcing this?

nguhoa: I don't remember which one require this one be immutable. I remove this out of PR.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/MinMaxTable.cs
```diff
@@ -15,7 +15,8 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
 {
     // Min(source:*, projection:n)
     // Max(source:*, projection:n)
-    // Corresponding DAX functions: Min, MinA, MinX, Max, MaxA, MaxX
+    // Corresponding DAX functions: Min, MinA, MinX, Max, MaxA, MaxX
+    [ThreadSafeImmutable]
```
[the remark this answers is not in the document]
nguhoa: StatisticalTableFunction is not tagged as thread safe. So update in base class for this.

LucGenetier:
✅ No public API change.

Issue #1535, opened 2023-05-29
Fix not thread safe code


nguhoa:
Here are the list of the bugs that do not have thread safe code which can create race condition.
 
Bug 1: _symbolValues is modified in Set function, outside of constructor. 
	SymbolValues._symbolValues returns mutable value
	SymbolValues.OnUpdate is not readonly

Bug 2: ScopeSymbol (_topScopeSymbol) is not thread safe. List _fields is modifed in AddOrGetIndexForField, outside of constructor. 
	readonly CultureInfo is not thread safe.
		ParsedExpression._topScopeSymbol returns mutable value. 
		ParsedExpression._cultureInfo returns mutable value

  Bug: ScopeSymbol is not thread safe. List _fields is modifed in AddOrGetIndexForField, outside of constructor. Same bug with		ParsedExpression.
		AggregateCoercionNode.Scope returns mutable value

  Bug: ScopeSymbol is not thread safe. List _fields is modifed in AddOrGetIndexForField, outside of constructor. Same bug with		ParsedExpression.
		ScopeAccessNode.Value returns mutable value

Bug 3: _expr is modifed in outside of constructor.
	UserDefinedTexlFunction._expr is not readonly
	UserDefinedTexlFunction._check returns mutable value

Bug 4: not thread safe - modify out of constructor and not  readonly
	UserDefinedFunction._args returns mutable value
	UserDefinedFunction._binding is not readonly
	UserDefinedFunction._binding returns mutable value

Bug 5: - internal class TexlFunctionSet is not thread safe.
		ComposedReadOnlySymbolTable.Microsoft.PowerFx.Core.Binding.INameResolver.Functions returns mutable value
		ComposedReadOnlySymbolTable._nameResolverFunctions is not readonly
		ComposedReadOnlySymbolTable._nameResolverFunctions returns mutable value
		ComposedReadOnlySymbolTable._cachedVersionHash is not readonly
		ComposedReadOnlySymbolTable._cachedVersionHash returns mutable value
 
Bug 6: VersionHash (_constant) is not thread safe.
	DeferredSymbolTable._constant returns mutable value

Bug 7: TexlFunctionSet (ReadOnlySymbolTable._functions) is not thread safe. 
	VersionHash (_version) is not thread safe. 
	EnumStoreBuilder (ReadOnlySymbolTable._enumStoreBuilder) is not thread safe - _workingEnums is dictionary and is modified in non constructor function. 
	_variables is modified outside of constructor.

[...]

PR #1537, opened 2023-05-30
Bug 9: Update VariadicBase and related to have thread safe

nguhoa:
Fix bug 9 in issue: https://github.com/microsoft/Power-Fx/issues/1535.


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/VariadicBase.cs
```diff
@@ -20,7 +20,7 @@ public abstract class VariadicBase : TexlNode
         /// </summary>
         public IReadOnlyList<TexlNode> ChildNodes => Children;
 
-        internal readonly TexlNode[] Children;
+        internal readonly IReadOnlyList<TexlNode> Children;
```
nguhoa, line 23: >internal readonly IReadOnlyList<TexlNode> Children; [](http://example.com/codeflow?start=8&length=51)

This is key change in this PR. Other places are dependency updated.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4006,7 +4006,7 @@ private ScopeUseSet GetCallNodeScopeUseSet(CallNode node, CallInfo info)
                 }
                 else
                 {
-                    var args = node.Args.Children;
+                    var args = node.Args.Children.ToArray();
```
CarlosFigueiraMSFT, line 4009: >.ToArray() [](http://example.com/codeflow?start=49&length=10)

no need to `ToArray` here - can use .Count on line 4012

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4328,7 +4328,7 @@ public override bool PreVisit(CallNode node)
                 // The zeroth arg should not be a lambda. Instead it defines the context type for the lambdas.
                 Contracts.Assert(!maybeFunc.IsLambdaParam(0));
 
-                var args = node.Args.Children;
+                var args = node.Args.Children.ToArray();
```
CarlosFigueiraMSFT, line 4331: >.ToArray() [](http://example.com/codeflow?start=45&length=10)

No need to `ToArray` - can use args.Count where args.Length is used.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4801,7 +4801,7 @@ private void PreVisitMetadataArg(CallNode node, TexlFunction func)
                 Contracts.Assert(func.SupportsMetadataTypeArg);
                 Contracts.Assert(!func.HasLambdas);
 
-                var args = node.Args.Children;
+                var args = node.Args.Children.ToArray();
```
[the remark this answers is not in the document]
nguhoa: HandleCheckInvocation in 4855 need to pass in arrary. so either we have to toarray here or toarrary in line 4855.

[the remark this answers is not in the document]
CarlosFigueiraMSFT: Ah, you're right, missed that one.

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1542, opened 2023-05-31
Fix flaky threading test

### LINKED ISSUE #1541: fix Flaky threading test
Test added in #1523 can fail because it assumes Parallel.For will run on multiple threads. It can sometimes be single threaded.

MikeStall:
Fix #1541


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #1545, opened 2023-06-01
Add support for Office 365 Outlook connector, make sure we can load the connector without issue

LucGenetier:
Add support for Office 365 Outlook connector
Add support for optional parameter name conflict management
Ignore "builtInOperation" in x-ms-dynamic-values (= no support)
Support schema formats email, uri, byte (Base64), html as strings
Support schema formats date, date-no-tz as strings (like date-time)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/ArgumentMapper.cs
```diff
@@ -229,6 +229,28 @@ public ArgumentMapper(IEnumerable<OpenApiParameter> parameters, OpenApiOperation
             HiddenRequiredParamInfo = hiddenRequiredParams.ConvertAll(x => Convert(x, numberIsFloat: numberIsFloat)).Union(hiddenRequiredBodyParams.ConvertAll(x => Convert(x, numberIsFloat: numberIsFloat))).ToArray();
             OptionalParamInfo = optionalParams.ConvertAll(x => Convert(x, numberIsFloat: numberIsFloat)).Union(optionalBodyParams.ConvertAll(x => Convert(x, numberIsFloat: numberIsFloat))).ToArray();
 
+            // Validate we have no name conflict between required and optional parameters
+            // In case of conflict, we rename the optional parameter and add _1, _2, etc. until we have no conflict
+            List<string> requiredParamNames = RequiredParamInfo.Select(rpi => rpi.TypedName.Name.Value).ToList();
+            foreach (ServiceFunctionParameterTemplate opi in OptionalParamInfo)
```
lesaltzm, line 235: Is it possible for an optional param name to collide with a different optional param?

What about after rename, say we had:

Required: "Foo"
Optional: "Foo", "Foo_1"

src/tests/Microsoft.PowerFx.Connectors.Tests/OpenApiParserTests.cs
```diff
@@ -39,10 +41,50 @@ public void ACSL_GetFunctionNames()
             Assert.Equal("Conversations (CLU) (2022-05-01)", function.Summary);
             Assert.Equal("/apim/cognitiveservicestextanalytics/{connectionId}/language/:analyze-conversations", function.OperationPath);
             Assert.Equal(HttpMethod.Post, function.HttpMethod);
-        }
-
+        }
+
+        [Fact(Skip = "Need files from AAPT-connector and PowerPlatformConnectors projects")]
```
lesaltzm, line 46: This seems like a test that shouldn't live here? Or those files should be added to this project. Referring to absolute paths in a test seems like a bad idea, generally we should be using relative paths that are populated by the build.

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -205,10 +206,16 @@ public static ConnectorParameterType ToFormulaType(this OpenApiSchema schema, St
...
+                        case "html":
                             return new ConnectorParameterType(schema, FormulaType.String);
-
-                        case "date-time":
+
+                        case "date":
+                        case "date-time":
+                        case "date-no-tz":
                             // Consider this as a string for now
                             // $$$ Should be DateTime type
```
MikeStall, line 220: >// $$$ Should be DateTime type [](http://example.com/codeflow?start=28&length=30)

Will this be a compat issue with power apps?

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunctionParameterTemplate.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.AppMagic.Authoring
 {
     internal sealed class ServiceFunctionParameterTemplate
     {
-        private readonly TypedName _typedName;
+        private TypedName _typedName;
```
MikeStall, line 14: >TypedName [](http://example.com/codeflow?start=16&length=9)

Can we keep readonly?

src/libraries/Microsoft.PowerFx.Connectors/ArgumentMapper.cs
LucGenetier: Notice that it's only between required parameters & optional *body* parameters (non-body parameters cannot use the same name as required parameters, per OpenAPI specification)
This is only due to the fact that we flatten body parameters for Power Apps (in a record, at the end...)

That said, we could construct such a strange API but I'm not considering it here.
I don't believe this is a valid scenario and while I appreciate your (very clever) comment, I don't want to complicate the code here - I'll add a comment though for this non-trivial possibility.

In parallel, and to mitigate this risk, I'm adding extra tests for tests all our connectors

src/tests/Microsoft.PowerFx.Connectors.Tests/OpenApiParserTests.cs
LucGenetier: This is mostly 
- dirty, I know that ;)
- only for me, to tests all connectors
I could use env variables too, define a root localtion string, etc. but I don't see the value.

The code is only there to keep it and easily run it on my machine.
I could also add #ifdef CONNECTOR_TESTING if we really want to add some extra protection

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunctionParameterTemplate.cs
LucGenetier: With SetTypedName, no.

We could return a new ServiceFunctionParameterTemplate in this method to keep readonly but that's going to be far less efficient and I don't see the benefit.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
MikeStall: This is converting from Swagger (which has dates) to Power Fx (which has dates).

For example, can we test this:
Office365Users.MyProfileV2().birthday > Date(2001, 3,1)

src/libraries/Microsoft.PowerFx.Connectors/ArgumentMapper.cs
MikeStall: I think the real question here is: what does PowerApps do? 
We should match that behavior

src/tests/Microsoft.PowerFx.Connectors.Tests/OpenApiParserTests.cs
MikeStall: I'd be fine keeping as a test and moving this to a separate file, and adding a comment to make the intent clear.

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
LucGenetier: Swagger/OpenAPI has dates but they are stored as strings
See [https://swagger.io/specification/v2/](https://swagger.io/specification/v2/)

The only test is around loading the swagger which is our first goal here (what Cole and Justin have reported).

Supporting dates in APIs can be done in a separated PR.

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunctionParameterTemplate.cs
LucGenetier: There is no possible threading issue here but i can use the constructor to change that single value
Will be very inefficient but we could keep this field readonly

src/libraries/Microsoft.PowerFx.Connectors/ArgumentMapper.cs
LucGenetier: I'd need help to find where this is done. Would could help me here?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
MikeStall: This change will suppress exceptions during the parse phase, but the goal here is to be able to successfully use the O365 Users connector, ie, the methods are imported and work same as in PA  ...   

So if we have to add new cases for parser, that's a signal we should verify that new functionality also works during eval.

src/libraries/Microsoft.PowerFx.Connectors/ArgumentMapper.cs
LucGenetier: Finally found where this logic is implemented in PA Client after a long research
This is in src\Cloud\DocumentServer.Core\Document\Importers\ServiceConfig\RestFunctionDefinitionBuilder.cs at line 1176 - CreateUniqueImpliedParameterName

They search for conflicts between any parameter (mandatory, optional) and any location (query, body) and then append "_n" where needed (starting with 1, and checking each number till no conflict)

I follow the same logic here, in a more restrictive manner
As the only example we need to support here is O365 Outlook connector and this code is sufficient, I think we're good enough for now and this code can be improved later when we'll have more testing around connectors.

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
LucGenetier: While I agree we can warn them that 'date' and 'date-no-tz' might not work as expected as they'll be treated as strings, there is nothing new with 'date-time' which is supported as a string for a long time and without any partner complaint - and many connectors (as I remember) use 'date-time'.

I'm in favor in not delaying this PR anymore, allow Cards to load the O365 Outlook connector (not the Users one), and continue improving our code in future PRs.
(this PR isn't about supporting date formats)

LucGenetier: Have checked how this is managed in PA Client
In src\AppMagic\dll\AuthoringCore\Importers\DataDescription\DataDescriptionParser.cs, line 653-654 (in ParseDataType)

Their implementation is trickier as we will need to support the different formats (Iso8601ExtendedUtc, Iso8601Datetime) and different behaviors (UserLocal, NoTimeZone) - at minimum (we can easily imagine a 3rd behavior 'WithDefinedTimeZone'.

Clearly out of scope for this PR to my eyes.

LucGenetier: Will add comments with these findings

src/libraries/Microsoft.PowerFx.Connectors/ArgumentMapper.cs
LucGenetier: Added a comment in code

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
LucGenetier: Logged this issue for dates: [https://github.com/microsoft/Power-Fx/issues/1557](https://github.com/microsoft/Power-Fx/issues/1557)

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

Issue #1490, opened 2023-05-08
stronger enforcement for records  in custom functions 


MikeStall:
Today, with V1 semantics, any record coerces to any other record.  This is too loose for custom functions that take records. 

Goals:
1. No change to Patch/Collect
2. Ideally, a custom function can wrap Patch/Collect
3. We want to catch cases where a function expects Accounts, but gets passed a Contact.

So update to be this: Modify Custom function handler so that for a parameter that expects a record, passing in extra fields is an error.  Passing in missing fields is ok.

jas-valgotar:
This function maybe useful in matching fields. (https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Extensions.cs#L52)

MikeStall:
Suppose we have a function that takes:
`{ X : Number, Y:Number }`

There are several cases to try:

Example | Kind |  Change
--- | --- | --- 
`{ X : Number, Y: Number }`  | perfect | 
`{ X : Number, Y: Number, A : Number }`  | extra, A?  ... ignored | !!! Should be an error case (this bug) 
`{ X : Number }`  | missing Y  ... fill in blank  | 
`{ A : Number}`  | No relation!!!!  |  Today, this is allow if Extra+ missing are allowed.  But that's too flexible.   By making extra fields an error, that will make this an error. 
`{ X : String }`  | Coercable !!  | 
`{ X : { Nested: String }}`  | Error - not coercable  |

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/1555

PR #1555, opened 2023-06-04
Stronger enforcement for records in custom functions

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1490


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/CustomTexlFunction.cs
```diff
@@ -50,5 +54,32 @@ public override bool IsLazyEvalParam(int index)
...
+            Contracts.AssertValue(args);
+            Contracts.AssertAllValues(args);
+            Contracts.AssertValue(argTypes);
+            Contracts.Assert(args.Length == argTypes.Length);
+            Contracts.AssertValue(errors);
+
+            var isValid = base.CheckTypes(context, args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
+
+            // Check if all record names of args exist against return type and if its possible to coerce.
+            if (isValid && returnType.IsRecord)
```
jas-valgotar, line 69: >returnType.IsRecord [](http://example.com/codeflow?start=27&length=19)

Why do we check ReturnType? What if function takes in a record but doesn't return record?

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/CustomTexlFunction.cs
```diff
@@ -50,5 +54,32 @@ public override bool IsLazyEvalParam(int index)
...
+            {
+                for (var i = 0; i < args.Length; i++)
+                {
+                    DType curType = argTypes[i];
+
+                    if (curType.IsRecord && !curType.CheckAggregateNames(returnType, args[i], errors, SupportsParamCoercion, context.Features.PowerFxV1CompatibilityRules))
```
jas-valgotar, line 75: >returnType [](http://example.com/codeflow?start=73&length=10)

Should we check it against paramTypes? Instead of return type.

CarlosFigueiraMSFT:
>             : base(DPath.Root, name, name, CustomFunctionUtility.SG("Custom func " + name), FunctionCategories.MathAndStat, returnType, 0, paramTypes.Length, paramTypes.Length, paramTypes)

Not related to your changes, but this is not correct

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/CustomTexlFunction.cs:39 in 2a6bdc1. [](commit_id = 2a6bdc14f327c5a7f28513ae00bff5385d2a65fb, deletion_comment = False)

src/tests/Microsoft.PowerFx.Interpreter.Tests/CustomFunctions.cs
```diff
@@ -247,6 +250,210 @@ public static RecordValue Execute()
...
+            }
+            else
+            {
+                Assert.False(check.IsSuccess);
+                Assert.Contains("invalid arguments", errorMsg);
+            }
+        }
+
+        // Must have "Function" suffix. 
+        private class TestSuccessRecordCustomFunction : ReflectionFunction
```
CarlosFigueiraMSFT, line 332: >TestSuccessRecordCustomFunction [](http://example.com/codeflow?start=22&length=31)

Suggestion: instead of having multiple functions, have a single custom (reflection) function that takes multiple arguments - and in the test pass blank value to the arguments that don't apply to the test.

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/CustomTexlFunction.cs
nguhoa: Check returntype here because compare between argtypes and return type. paramtypes is argTypes so cannot use here.

nguhoa:
>             : base(DPath.Root, name, name, CustomFunctionUtility.SG("Custom func " + name), FunctionCategories.MathAndStat, returnType, 0, paramTypes.Length, paramTypes.Length, paramTypes)

Updated function category to pass in through constructor.

---
In reply to: [1576989104](https://github.com/microsoft/Power-Fx/pull/1555#issuecomment-1576989104) [](http://example.com/codeflow?ancestors=1576989104,1576989104)

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/CustomTexlFunction.cs:39 in 2a6bdc1. [](commit_id = 2a6bdc14f327c5a7f28513ae00bff5385d2a65fb, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/CustomTexlFunction.cs
nguhoa: Good point. remove this check. Add test for non record.

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/CustomTexlFunction.cs
```diff
@@ -24,6 +28,8 @@ internal class CustomTexlFunction : TexlFunction
 
         internal BigInteger LamdaParamMask;
 
+        public override bool IsSelfContained => true;
```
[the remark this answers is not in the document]
nguhoa: updated.

src/tests/Microsoft.PowerFx.Interpreter.Tests/CustomFunctions.cs
nguhoa: Good suggestion. Update the code to remove duplicated.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/CustomTexlFunction.cs
```diff
@@ -50,5 +57,32 @@ public override bool IsLazyEvalParam(int index)
...
+            Contracts.AssertValue(errors);
+
+            var isValid = base.CheckTypes(context, args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
+
+            // Check if all record names of args exist against return type and if its possible to coerce.
+            if (isValid)
+            {
+                for (var i = 0; i < args.Length; i++)
+                {
+                    DType curType = argTypes[i];
```
jas-valgotar, line 76: >argTypes [](http://example.com/codeflow?start=36&length=8)

Should this be paramTypes[i] ?

src/tests/Microsoft.PowerFx.Interpreter.Tests/CustomFunctions.cs
```diff
@@ -247,6 +250,120 @@ public static RecordValue Execute()
...
+            FormulaValue result = null; 
+            ConfiguredTaskAwaitable<FormulaValue> resultAsync;
+            string errorMsg = string.Empty;
+
+            try
+            {
+                result = engine.Eval(functionName + inputRecord);
+
+                using var cts = new CancellationTokenSource();
+                resultAsync = engine.EvalAsync(functionName + inputRecord, cts.Token).ConfigureAwait(false);
```
jas-valgotar, line 314: >EvalAsync [](http://example.com/codeflow?start=37&length=9)

I think we can skip EvalAsync part since if Check().IsSuccess() is false we always throw.

src/tests/Microsoft.PowerFx.Interpreter.Tests/CustomFunctions.cs
```diff
@@ -247,6 +250,120 @@ public static RecordValue Execute()
...
+
+            var check = engine.Check(functionName + inputRecord);
+
+            FormulaValue result = null; 
+            ConfiguredTaskAwaitable<FormulaValue> resultAsync;
+            string errorMsg = string.Empty;
+
+            try
+            {
+                result = engine.Eval(functionName + inputRecord);
```
jas-valgotar, line 311: >engine [](http://example.com/codeflow?start=25&length=6)

nit: you can also get evaluator from `check.getEvaluator()` and eval on that, that way the test doesn't do `check` twice

src/tests/Microsoft.PowerFx.Interpreter.Tests/CustomFunctions.cs
jas-valgotar: thinking back I think its fine, since you are also testing non-error scenario.

src/tests/Microsoft.PowerFx.Interpreter.Tests/CustomFunctions.cs
```diff
@@ -247,6 +250,120 @@ public static RecordValue Execute()
...
+            CustomFunctionRecord(new TestRecordsCustomFunction(record, coerceRecord), "RecordsTest", "({x: \"2\", y: \"5\"})", true);            
+            CustomFunctionRecord(new TestRecordsCustomFunction(record, extraFieldRecord), "RecordsTest", "({x: 2, y: 5, a: 8})", false);
+            CustomFunctionRecord(new TestRecordsCustomFunction(record, notExistRecord), "RecordsTest", "({a: 8})", false);            
+            CustomFunctionRecord(new TestRecordsCustomFunction(nestedRecord, nestedRecord), "RecordsTest", "({x: {x : 2}})", true);           
+            CustomFunctionRecord(new TestRecordsCustomFunction(record, nestedRecord), "RecordsTest", "({x: {x : 2}})", false);
+            
+            CustomFunctionRecord(new TestNonRecordCustomFunction(), "NonRecordTest", "({x: 2})", false);
+        }
+
+        private void CustomFunctionRecord(ReflectionFunction reflectionFunction, string functionName, string inputRecord,  bool isSuccess)
```
jas-valgotar, line 294: >functionName [](http://example.com/codeflow?start=88&length=12)

Can we get function name from the instance of `reflectionFunction` ?

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/CustomTexlFunction.cs
nguhoa: I see, i updated to use paramtypes here

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/CustomTexlFunction.cs
nguhoa: keep this and use paramtypes in line below.

src/tests/Microsoft.PowerFx.Interpreter.Tests/CustomFunctions.cs
nguhoa: update

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs
```diff
@@ -34,6 +36,8 @@ public abstract class ReflectionFunction
         // If non-null, specify what type of runtime config this function can accept. 
         protected Type ConfigType { get; init; }
 
+        internal readonly string FunctionName;
```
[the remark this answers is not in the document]
nguhoa: it only be null if caller pass null name in constructor. which is not ever.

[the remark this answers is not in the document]
jas-valgotar: What if another Constructor is used? Note that Reflection Function has 2 constructors...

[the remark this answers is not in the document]
nguhoa: i created new function to get name from _info to avoid create new variable like this.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs
```diff
@@ -270,9 +277,38 @@ public async Task<FormulaValue> InvokeAsync(IServiceProvider serviceProvider, Fo
 
             formulaResult ??= FormulaValue.NewBlank(_info.RetType);
 
-            if (!formulaResult.Type._type.Accepts(_info.RetType._type, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: true))
+            var dataSourceType = formulaResult.Type._type;
+            var retType = _info.RetType._type;
+
```
CarlosFigueiraMSFT, line 282: This method is already quite large (100+ lines); consider breaking it down into smaller parts

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/CustomTexlFunction.cs
```diff
@@ -50,5 +57,32 @@ public override bool IsLazyEvalParam(int index)
...
+            var isValid = base.CheckTypes(context, args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
+
+            // Check if all record names of args exist against arg types and if its possible to coerce.
+            if (isValid)
+            {
+                for (var i = 0; i < args.Length; i++)
+                {
+                    DType curType = argTypes[i];
+
+                    if (curType.IsRecord && !curType.CheckAggregateNames(this.ParamTypes[i], args[i], errors, SupportsParamCoercion, context.Features.PowerFxV1CompatibilityRules))
```
CarlosFigueiraMSFT, line 78: >IsRecord [](http://example.com/codeflow?start=32&length=8)

Do we need to do a similar check for table parameters? For example, this check validates that {a:1,b:2,c:3} can't be sent to an argument that expects ![a:n, b:n]. But if we have an argument of type *[a:n, b:n] we probably don't want to allow [{a:1,b:2,c:3}] either

CarlosFigueiraMSFT, line 78: >SupportsParamCoercion [](http://example.com/codeflow?start=110&length=21)

Some functions support coercion only for certain parameters, so we should call `SupportCoercionForArg(i)` instead

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs
```diff
@@ -270,9 +277,38 @@ public async Task<FormulaValue> InvokeAsync(IServiceProvider serviceProvider, Fo
...
             formulaResult ??= FormulaValue.NewBlank(_info.RetType);
 
-            if (!formulaResult.Type._type.Accepts(_info.RetType._type, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: true))
+            var dataSourceType = formulaResult.Type._type;
+            var retType = _info.RetType._type;
+            
+            if (!dataSourceType.Accepts(retType, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: true))
             {
-                return CommonErrors.CustomError(formulaResult.IRContext, string.Format(CultureInfo.InvariantCulture, "Return type should have been {0}, found {1}", _info.RetType._type, formulaResult.Type._type));
+                bool isVaid = false;
```
CarlosFigueiraMSFT, line 285: >isVaid [](http://example.com/codeflow?start=21&length=6)

[typo] isValid

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs
```diff
@@ -270,9 +277,38 @@ public async Task<FormulaValue> InvokeAsync(IServiceProvider serviceProvider, Fo
 
             formulaResult ??= FormulaValue.NewBlank(_info.RetType);
 
-            if (!formulaResult.Type._type.Accepts(_info.RetType._type, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: true))
+            var dataSourceType = formulaResult.Type._type;
```
CarlosFigueiraMSFT, line 280: >dataSourceType [](http://example.com/codeflow?start=16&length=14)

This doesn't have to be a data source, we can have any type here. Naming it as such can be misleading

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs
```diff
@@ -270,9 +277,38 @@ public async Task<FormulaValue> InvokeAsync(IServiceProvider serviceProvider, Fo
...
-            if (!formulaResult.Type._type.Accepts(_info.RetType._type, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: true))
+            var dataSourceType = formulaResult.Type._type;
+            var retType = _info.RetType._type;
+            
+            if (!dataSourceType.Accepts(retType, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: true))
             {
-                return CommonErrors.CustomError(formulaResult.IRContext, string.Format(CultureInfo.InvariantCulture, "Return type should have been {0}, found {1}", _info.RetType._type, formulaResult.Type._type));
+                bool isVaid = false;
+
+                if (retType.IsRecord)
```
CarlosFigueiraMSFT, line 287: >IsRecord [](http://example.com/codeflow?start=28&length=8)

Similar comment than the one in CustomTexlFunction - if we are checking for records we should probably check for tables as well

src/tests/Microsoft.PowerFx.Interpreter.Tests/CustomFunctions.cs
```diff
@@ -247,6 +250,119 @@ public static RecordValue Execute()
...
+        [Theory]
+        [InlineData("({x: 2})", true)]
+        [InlineData("({x: 2, y: 5})", true)]
+        [InlineData("({x: \"2\", y: \"5\"})", true)]
+        [InlineData("({x: 2, y: 5, a: 8})", false)]
+        [InlineData("({a: 8})", false)]
+        [InlineData("({x: {x : 2}})", false)]
+        public void CustomFunctionRecordTest(string inputRecord, bool isSuccess)
+        {
+            var record = RecordType.Empty().Add(new NamedFormulaType("x", FormulaType.Number)).Add(new NamedFormulaType("y", FormulaType.Number));
```
CarlosFigueiraMSFT, line 282: >record [](http://example.com/codeflow?start=16&length=6)

[suggestion] recordType. This is the type, not the record itself

src/tests/Microsoft.PowerFx.Interpreter.Tests/CustomFunctions.cs
```diff
@@ -247,6 +250,119 @@ public static RecordValue Execute()
...
+            public static RecordValue Execute(RecordValue recordValue)
+            {
+                return FormulaValue.NewRecordFromFields(recordValue.Type, recordValue.Fields);
+            }
+        }
+
+        // Must have "Function" suffix. 
+        private class TestNonRecordCustomFunction : ReflectionFunction
+        {
+            public TestNonRecordCustomFunction()
```
CarlosFigueiraMSFT, line 351: >TestNonRecordCustomFunction [](http://example.com/codeflow?start=19&length=27)

This change is checking the input types of functions; this function has an input type of type record, just like the previous one. Do we need two of them?

src/tests/Microsoft.PowerFx.Interpreter.Tests/CustomFunctions.cs
```diff
@@ -247,6 +250,119 @@ public static RecordValue Execute()
...
+        public void CustomFunctionRecordTest(string inputRecord, bool isSuccess)
+        {
+            var record = RecordType.Empty().Add(new NamedFormulaType("x", FormulaType.Number)).Add(new NamedFormulaType("y", FormulaType.Number));
+            CustomFunctionRecord(new TestRecordsCustomFunction(record), inputRecord, isSuccess);
+        }
+
+        [Theory]
+        [InlineData("({x: 2})", true)]
+        [InlineData("({x: 2, y: 5})", false)]
+        public void CustomFunctionNonRecordTest(string inputRecord, bool isSuccess)
```
CarlosFigueiraMSFT, line 289: >CustomFunctionNonRecordTest [](http://example.com/codeflow?start=20&length=27)

How is this test different than the previous one? The first case seems identical to the second case in the previous test (same schema expected / actual), and the second one is the same as the fourth case in the previous test (one extra property in the actual)

nguhoa: This test different custom function where resulttype is not record. same input record. Just want to test that this work for both result type is record or not.

src/tests/Microsoft.PowerFx.Interpreter.Tests/CustomFunctions.cs
nguhoa: remove this.

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/CustomTexlFunction.cs
nguhoa: updated

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs
nguhoa: updated

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs
nguhoa: update name

src/tests/Microsoft.PowerFx.Interpreter.Tests/CustomFunctions.cs
```diff
@@ -247,6 +250,119 @@ public static RecordValue Execute()
...
+        {
+            CustomFunctionRecord(new TestNonRecordCustomFunction(), inputRecord, isSuccess);
+        }
+
+        private void CustomFunctionRecord(ReflectionFunction reflectionFunction, string inputRecord,  bool isSuccess)
```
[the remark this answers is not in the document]
nguhoa: update

src/tests/Microsoft.PowerFx.Interpreter.Tests/CustomFunctions.cs
nguhoa: update

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/CustomTexlFunction.cs
nguhoa: updated to check table as well.

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs
nguhoa: update to check table as well.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs
```diff
@@ -156,13 +158,18 @@ internal TexlFunction GetTexlFunction()
...
+            return new CustomTexlFunction(info.Name, FunctionCategories.UserDefined, info.RetType, info.ParamTypes)
             {
                 _impl = (runtimeConfig, args, cancellationToken) => InvokeAsync(runtimeConfig, args, cancellationToken),
                 LamdaParamMask = info.LamdaParamMask,
             };
         }
 
+        internal string GetFunctionName()
+        {
+            return _info?.Name;
```
jas-valgotar, line 170: >_info [](http://example.com/codeflow?start=19&length=5)

nit: To avoid  _info being null, I would recommend making it readonly, and no arg constructor can just call _info = Scan();

nguhoa: update

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs
```diff
@@ -156,21 +158,27 @@ internal TexlFunction GetTexlFunction()
...
+            return new CustomTexlFunction(info.Name, FunctionCategories.UserDefined, info.RetType, info.ParamTypes)
             {
                 _impl = (runtimeConfig, args, cancellationToken) => InvokeAsync(runtimeConfig, args, cancellationToken),
                 LamdaParamMask = info.LamdaParamMask,
             };
         }
 
+        internal string GetFunctionName()
+        {
+            var info = Scan();
```
jas-valgotar, line 170: > var info = Scan(); [](http://example.com/codeflow?start=11&length=19)

Why do we need to call Scan? Do we have this info already in _info?

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs
```diff
@@ -156,21 +158,27 @@ internal TexlFunction GetTexlFunction()
...
+
         public FormulaValue Invoke(IServiceProvider serviceProvider, FormulaValue[] args)
         {
             return InvokeAsync(serviceProvider, args, CancellationToken.None).Result;
         }
 
         public async Task<FormulaValue> InvokeAsync(IServiceProvider serviceProvider, FormulaValue[] args, CancellationToken cancellationToken)
         {
-            Scan();
+            var info = Scan();
```
jas-valgotar, line 181: > var info = Scan(); [](http://example.com/codeflow?start=11&length=19)

Why do we again need to Scan() if we _info is already populated in ctor?

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs
nguhoa: if _info is null, scan will find the info.

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs
nguhoa: because this class have 2 construtors and one of it set _info is null.

jas-valgotar: What if we call Scan() there?

jas-valgotar:
>             _info = null;

We should call `Scan()` here

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs:45 in 822521c. [](commit_id = 822521cff7be8b4fdba99731bccddffb52ef7304, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs
nguhoa: the old code use just Scan() because scan will update _info. But now we change _info to readonly so it cannot be modified in scan. so my newest update will need pass scan to variable and use that variable.

jas-valgotar (approved):
:shipit:

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/CustomFunctions.cs
```diff
@@ -247,6 +250,111 @@ public static RecordValue Execute()
...
+        public void CustomFunctionRecordTest(string inputRecord, bool isSuccess)
+        {
+            var recordType = RecordType.Empty().Add(new NamedFormulaType("x", FormulaType.Number)).Add(new NamedFormulaType("y", FormulaType.Number));
+            var tableType = recordType.ToTable();
+            
+            CustomFunctionRecord(new TestRecordsCustomFunction(recordType), "(" + inputRecord + ")", isSuccess);
+            CustomFunctionRecord(new TestTableCustomFunction(tableType), "([" + inputRecord + "])", isSuccess);
+        }
+
+        private void CustomFunctionRecord(ReflectionFunction reflectionFunction, string inputRecord, bool isSuccess)
```
CarlosFigueiraMSFT, line 289: >CustomFunctionRecord [](http://example.com/codeflow?start=21&length=20)

Naming suggestion: TestCustomFunctionWithRecordArgument

src/tests/Microsoft.PowerFx.Interpreter.Tests/CustomFunctions.cs
```diff
@@ -247,6 +250,111 @@ public static RecordValue Execute()
...
+                : base("RecordsTest", tableType, tableType)
+            {
+            }
+
+            // Must have "Execute" method. 
+            public static TableValue Execute(TableValue tableValue)
+            {
+                return tableValue;
+            }
+        }
```
CarlosFigueiraMSFT, line 356: You can merge those two functions into a single one:

        // Must have "Function" suffix. 
        private class TestAggregateIdentityCustomFunction<TType, TValue> : ReflectionFunction
            where TType : AggregateType
            where TValue : FormulaValue
        {
            public TestAggregateIdentityCustomFunction(TType argType)
                : base("RecordsTest", argType, argType)
            {
            }

            // Must have "Execute" method. 
            public static TValue Execute(TValue recordValue)
            {
                return recordValue;
            }
        }

nguhoa: updated.

LucGenetier:
✅ No public API change.

PR #1558, opened 2023-06-05
Blocked using IsType, AsType, Type names for UDFs

VamsiModem:
Blocked naming UDFs with Type, IsType, AsType 
Added MaxAllowedOperations to PowerFx config, this will be used with recursive UDFs

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests/UserDefinedFunctionTests.cs
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/UserDefinedFunctionTests.cs
@@ -34,7 +34,7 @@ namespace Microsoft.PowerFx.Core.Tests
         [InlineData("Add(x: Number, y:Number): Number = x + y; Foo(x: Number): Number = Abs(x); y = 2;", 2, 1, false)]
         [InlineData("Foo(x: Number): Number = /*this is a test*/ Abs(x); y = 2;", 1, 1, false)]
         [InlineData("Add(x: Number, y:Number): Number = b + b; Foo(x: Number): Number = Abs(x); y = 2;", 2, 1, true)]
-        [InlineData("Add(x: Number, y:Number): Boolean = x + y;", 1, 0, true)]
+        [InlineData("Add(x: Number, y:Number): Boolean = x + y;", 1, 0, false)]
         [InlineData("Add(x: Number, y:Number): SomeType = x + y;", 0, 0, true)]
         [InlineData("Add(x: SomeType, y:Number): Number = x + y;", 0, 0, true)]
         [InlineData("Add(x: Number, y:Number): Number = x + y", 0, 0, true)]


lesaltzm (dismissed):

lesaltzm (approved):

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
```diff
@@ -83,7 +84,13 @@ private IEnumerable<UserDefinedFunction> CreateUserDefinedFunctions(IEnumerable<
             foreach (var udf in uDFs)
             {
                 var udfName = udf.Ident.Name;
-                if (texlFunctionSet.AnyWithName(udfName) || BuiltinFunctionsCore._library.AnyWithName(udfName))
+                if (_restrictedUDFNames.Contains(udfName))
+                {
+                    errors.Add(new TexlError(udf.Ident, DocumentErrorSeverity.Severe, TexlStrings.ErrUDF_FunctionAlreadyDefined, udfName));
+                    continue;
```
CarlosFigueiraMSFT, line 90: This is the same as line 95; can we merge the `if` statements?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -42,6 +43,11 @@ public sealed class PowerFxConfig
 
         public int MaxCallDepth { get; set; }
 
+        /// <summary>
+        /// Used with recursive user-defined functions to break long running functions, a good example of this is fibonacci. 
+        /// </summary>
+        public int MaxOperationsAllowed { get; set; }
```
MikeStall, line 49: >MaxOperationsAllowed [](http://example.com/codeflow?start=19&length=20)

how does this compare to MaxCallDepth?

VamsiModem: MaxCallDepth tells us how deep the recursive call is, in cases like Fibonacci recursive tree doesn't go that deep but it's wide and this can potentially block the app. To avoid this we use MaxOperationsAllowed .

MikeStall (changes requested):
:clock1:

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
MikeStall: What *exactly* does that mean?  

For example, in C# interpreter we have a timeout that signals on a timeout - that accepts a TimeValue is a well defined quota.  But was is an "operation"?

LucGenetier: Could you add a test where you 
1/ set MaxOps = 10, 
2/ use an UDF that will make Ops=30 
3/ return an error "Too many operations" of some sort, in the middle of the execution
4/ check we have that error & message

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

Issue #1553, opened 2023-06-02
Crash when CurrentUICulture is Invariant


jcoc611-microsoft:
We would like to run with `CultureInfo.CurrentUICulture = CultureInfo.InvariantCulture`. However this causes a crash because BaseError's constructor does not use the culture we set on `PowerFxConfig`. I think you should either fall back to `en` in this case, or use the culture set on RecalcEngine with PowerFxConfig.

```
System.ArgumentException: locale
   at Microsoft.PowerFx.Core.Utils.Contracts.CheckNonEmpty(String s, String paramName)
   at Microsoft.PowerFx.Core.Localization.StringResources.TryGetErrorResource(ErrorResourceKey resourceKey, ErrorResource& resourceValue, String locale)
   at Microsoft.PowerFx.Core.Errors.ErrorUtils.GetLocalizedErrorContent(ErrorResourceKey errKey, CultureInfo locale, ErrorResource& errorResource)
   at Microsoft.PowerFx.Core.Errors.BaseError..ctor(IDocumentError innerError, Exception internalException, DocumentErrorKind kind, DocumentErrorSeverity severity, ErrorResourceKey errKey, Span textSpan, IEnumerable`1 sinkTypeErrors, Object[] args)
   at Microsoft.PowerFx.Core.Errors.BaseError..ctor(IDocumentError innerError, Exception internalException, DocumentErrorKind kind, DocumentErrorSeverity severity, ErrorResourceKey errKey, Object[] args)
   at Microsoft.PowerFx.Core.Errors.TexlError..ctor(TexlNode node, DocumentErrorSeverity severity, ErrorResourceKey errKey, Object[] args)
   at Microsoft.PowerFx.Core.App.ErrorContainers.ErrorContainer.Error(DocumentErrorSeverity severity, TexlNode node, ErrorResourceKey errKey, Object[] args)
   at Microsoft.PowerFx.Core.App.ErrorContainers.ErrorContainer.Error(TexlNode node, ErrorResourceKey errKey, Object[] args)
   at Microsoft.PowerFx.Core.Binding.TexlBinding.Visitor.Visit(FirstNameNode node)
   at Microsoft.PowerFx.Syntax.FirstNameNode.Accept(TexlVisitor visitor)
   at Microsoft.PowerFx.Core.Binding.TexlBinding.Visitor.Run()
[...]

PR #1559, opened 2023-06-06
Fix crash when CurrentUICulture is Invariant

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1553
CultureInfo.CurrentUICulture.Name is empty (make locale empty) when CurrentUICulture is set to invariant. This makes TryGetErrorResource throw exception for empty locale. 
This PR fix to set "en-US" default when CultureInfo.CurrentUICulture.Name is empty (invariant case).


LucGenetier:
✅ No public API change.

LucGenetier (approved):
:shipit:

PR #1566, opened 2023-06-09
Move expansion of Canvas-Specific control type into INameResolver from IExternalControl

lesaltzm:
This change is a breaking-internals change, but it allows for safer testing of ongoing analysis improvements. It should have no impact on non-canvas hosts, other than up taking a new member, if they directly implement INameResolver, and IIRC, they should be using ReadOnlySymbolTable instead of directly using INameResolver.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/INameResolver.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/INameResolver.cs
@@ -67,9 +67,9 @@ namespace Microsoft.PowerFx.Core.Binding
         bool LookupParent(out NameLookupInfo lookupInfo);
 
         /// <summary>
-        /// In Canvas specifically, this is used to retrieve the full, derived, type
-        /// of a control object where the output schema may be dependent on a control inputs and heirarchy. 
-        /// All Non-Canvas hosts should not implement this method.
+        /// In Power Apps specifically, this is used to retrieve the full, derived, type
+        /// of a control object where the output schema may be dependent on a control inputs and hierarchy. 
+        /// All Non-Power Apps hosts should not implement this method.
         /// </summary>
         /// <param name="control">The control to get the full type of.</param>
         /// <param name="controlType">output: the expanded control type.</param>
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/INameResolver.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/INameResolver.cs
@@ -66,6 +66,15 @@ namespace Microsoft.PowerFx.Core.Binding
         // Looks up the parent control for the current context.
         bool LookupParent(out NameLookupInfo lookupInfo);
 
+        /// <summary>
+        /// In Canvas specifically, this is used to retrieve the full, derived, type
+        /// of a control object where the output schema may be dependent on a control inputs and heirarchy. 
+        /// All Non-Canvas hosts should not implement this method.
+        /// </summary>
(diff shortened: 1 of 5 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Binding/INameResolver.cs
```diff
@@ -66,6 +66,15 @@ internal interface INameResolver
...
         bool LookupParent(out NameLookupInfo lookupInfo);
 
+        /// <summary>
+        /// In Canvas specifically, this is used to retrieve the full, derived, type
+        /// of a control object where the output schema may be dependent on a control inputs and heirarchy. 
+        /// All Non-Canvas hosts should not implement this method.
+        /// </summary>
+        /// <param name="control">The control to get the full type of.</param>
+        /// <param name="controlType">output: the expanded control type.</param>
+        bool LookupExpandedControlType(IExternalControl control, out DType controlType);
```
Shpakh-MSFT, line 76: Why not add a separate interface for this so that it's not forcing other hosts to even think about it?

src/libraries/Microsoft.PowerFx.Core/Binding/INameResolver.cs
```diff
@@ -66,6 +66,15 @@ internal interface INameResolver
         // Looks up the parent control for the current context.
         bool LookupParent(out NameLookupInfo lookupInfo);
 
+        /// <summary>
+        /// In Canvas specifically, this is used to retrieve the full, derived, type
```
Shpakh-MSFT, line 70: >Canvas [](http://example.com/codeflow?start=15&length=6)

Can you say Power Apps here instead? We may get other products in future which will also have some kind of Canvas.

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Binding/INameResolver.cs
```diff
@@ -66,6 +66,15 @@ internal interface INameResolver
...
+        /// of a control object where the output schema may be dependent on a control inputs and heirarchy.
```
CarlosFigueiraMSFT, line 71: >heirarchy [](http://example.com/codeflow?start=97&length=9)

[typo] hierarchy

src/libraries/Microsoft.PowerFx.Core/Binding/INameResolver.cs
```diff
@@ -66,6 +66,15 @@ internal interface INameResolver
         // Looks up the parent control for the current context.
         bool LookupParent(out NameLookupInfo lookupInfo);
 
+        /// <summary>
+        /// In Canvas specifically, this is used to retrieve the full, derived, type
+        /// of a control object where the output schema may be dependent on a control inputs and heirarchy. 
+        /// All Non-Canvas hosts should not implement this method.
```
CarlosFigueiraMSFT, line 72: >All Non-Canvas hosts should not implement this method. [](http://example.com/codeflow?start=12&length=54)

Since this is internal, will there by any non-Canvas/PowerApps host that can implement it (only PA has internals visibility)?

lesaltzm: The dataverse target also has internals visibility.

LucGenetier:
✅ No public API change.

Shpakh-MSFT (approved):
:shipit:

Issue #1571, opened 2023-06-12
Does TryCoerceToRecord loop the wrong fields?


xieofxie:
https://github.com/microsoft/Power-Fx/blob/ace1d1fa33c34eb40a96404de59b67562779250f/src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs#L238

In my understanding, when we coerce source to target, we should check that every target's field exists in source but not the way around. In current implementation, for the passing scenario like below, the newResult (`{![key:n]}`) doesn't have type recordType (`{![key:n, value:s]}`).

```
var record = FormulaValue.NewRecordFromFields(new NamedValue("key", FormulaValue.New(1)));
var recordType = RecordType.Empty().Add(new NamedFormulaType("key", FormulaType.Number)).Add(new NamedFormulaType("value", FormulaType.String));
record.TryCoerceTo(recordType, out var newResult);
```

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/1572

Issue #1570, opened 2023-06-12
When source and target are of same type, should return true for CanPotentiallyCoerceToTargetType


xieofxie:
https://github.com/microsoft/Power-Fx/blob/ace1d1fa33c34eb40a96404de59b67562779250f/src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs#L345

Since they are not aggregated type, we should return true to support more types like Decimal, Guid etc.

https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/1572

PR #1572, opened 2023-06-12
Check target mismatch for TryCoerceToRecord

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1571
and: https://github.com/microsoft/Power-Fx/issues/1570


src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs
```diff
@@ -238,6 +238,12 @@ internal static bool TryCoerceTo(this FormulaValue value, FormattingInfo formatI
         public static bool TryCoerceToRecord(this RecordValue value, RecordType targetType, out RecordValue result)
         {
             result = null;
+
+            if (!CheckRecordTargetMatching(value.Type, targetType))
+            {
+                return false;
+            }
+
             int n = value.Fields.Count();
```
xieofxie, line 247: Do we want to support this scenario? I think this is a common scenario when value has more fields then targetType.
Now if we don't change the loop after this line, only an exact match will work and I don't think it is expected
```
var record = FormulaValue.NewRecordFromFields(new NamedValue("key", FormulaValue.New(1)), new NamedValue("value", FormulaValue.New(2)));
var recordType = RecordType.Empty().Add(new NamedFormulaType("key", FormulaType.Number));
record.TryCoerceTo(recordType, out var newResult);
```

nguhoa: Updated code. we will fill blank for extra/missing field

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs
```diff
@@ -311,12 +319,7 @@ public static bool CanPotentiallyCoerceToRecordType(this RecordType source, Reco
         {
```
xieofxie, line 319: Will we allow source to target if source has more fields than target?
This should align with `TryCoerceToRecord` method

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs
```diff
@@ -238,16 +238,16 @@ internal static bool TryCoerceTo(this FormulaValue value, FormattingInfo formatI
...
+            
+            foreach (var sourceField in value.Fields)
             {
-                var fieldName = value.Fields.ElementAt(i).Name;
+                var fieldName = sourceField.Name;
                 var fieldValue = value.GetField(fieldName);
                 if (!targetType.TryGetFieldType(fieldName, out FormulaType fieldType))
                 {
-                    return false;
+                    recordResult.Add(new NamedValue(fieldName, new BlankValue(IRContext.NotInSource(FormulaType.Blank))));
```
xieofxie, line 249: Now we will return result with a union of value type and targetType, is it expected?

I feel it could lead to unexpected scenario. Imagine this: we have A `{ key : value, keya: value }` and B `{ key: value, keyb: value}` and we want to coerce them both to C `{ key: value}`. But now we have new A and new B of different types.

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs
nguhoa: We allow and TryCoerceToRecord support as well.
We allow source has more fields than target or source has less fields than target. Basically, a record can coerce to any record. The missing field will be blank value.

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs
jas-valgotar, line 249: >FormulaType.Blank [](http://example.com/codeflow?start=100&length=17)

Should this be the type associated with fieldValue?

src/tests/Microsoft.PowerFx.Interpreter.Tests/TypeCoercionTest.cs
```diff
@@ -156,12 +169,31 @@ public void TryCoerceToRecordTest(string field1, string field2, bool expectedSuc
...
             {
                 Assert.True(isSucceeded);
                 Assert.Equal(FormulaValue.New(double.Parse(field1)).Value, result.GetField(fieldName1).ToObject());
                 Assert.Equal(FormulaValue.New(bool.Parse(field2)).Value, result.GetField(fieldName2).ToObject());
+
+                bool isSucceededExtraField = inputRecord.TryCoerceToRecord(extraFieldTargetType, out RecordValue extraFieldResult);
+                Assert.True(isSucceededExtraField);
+                Assert.Equal(FormulaValue.New(double.Parse(field1)).Value, extraFieldResult.GetField(fieldName1).ToObject());
+                Assert.Equal(FormulaValue.New(bool.Parse(field2)).Value, result.GetField(fieldName2).ToObject());
+                Assert.Equal(FormulaType.Blank, extraFieldResult.GetField(extraFieldName).Type);
```
jas-valgotar, line 191: >FormulaType.Blank [](http://example.com/codeflow?start=29&length=17)

FormulaType.String?

src/tests/Microsoft.PowerFx.Interpreter.Tests/TypeCoercionTest.cs
```diff
@@ -156,12 +169,31 @@ public void TryCoerceToRecordTest(string field1, string field2, bool expectedSuc
...
+
+                bool isSucceededMissingField = inputRecord.TryCoerceToRecord(missingFieldTargetType, out RecordValue missingFieldResult);
```
jas-valgotar, line 193: >isSucceededMissingField [](http://example.com/codeflow?start=21&length=23)

Can we also check the Fields count? I expect it should be 1?

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs
nguhoa: updated

src/tests/Microsoft.PowerFx.Interpreter.Tests/TypeCoercionTest.cs
nguhoa: updated

src/tests/Microsoft.PowerFx.Interpreter.Tests/TypeCoercionTest.cs
nguhoa: Add count check.

src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs
```diff
@@ -238,16 +238,15 @@ internal static bool TryCoerceTo(this FormulaValue value, FormattingInfo formatI
         public static bool TryCoerceToRecord(this RecordValue value, RecordType targetType, out RecordValue result)
         {
             result = null;
-            int n = value.Fields.Count();
-            var recordResult = new NamedValue[n];
-
-            for (int i = 0; i < n; i++)
+            var recordResult = new List<NamedValue>();
+            
+            foreach (var sourceField in value.Fields)
```
jas-valgotar, line 243: >foreach (var sourceField in value.Fields) [](http://example.com/codeflow?start=12&length=41)

We should directly loop over targetType,
If fields is present in `value` add it, else fill the `blank()`.
That way we avoid separate loop at line 272, And logic is easy to reason about!

nguhoa: If we loop target inside source, it is o(n^2) computation, seperate loop only cost o(2*n). I don't think we want it.

jas-valgotar: Not inside the source, I meant to say just looping over target, O(n) (instead of current O(2n))

nguhoa: i see, you mean use target instead of source. make sense then.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

jas-valgotar (approved):
:shipit:

PR #1582, opened 2023-06-14
wip Jas/1212 dv record

jas-valgotar:


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Types/AggregateType.cs
```diff
@@ -58,6 +58,11 @@ public virtual bool TryGetFieldType(string name, out FormulaType type)
             return true;
         }
 
+        internal bool TryGetBackingDType(string name, out DType type)
```
MikeStall, line 61: >TryGetBackingDType [](http://example.com/codeflow?start=22&length=18)

Do we have issues if a derived type overrides TryGetFieldType?

src/libraries/Microsoft.PowerFx.Core/Public/Values/NamedValue.cs
```diff
@@ -15,7 +17,24 @@ public class NamedValue
     {
         public string Name { get; }
 
-        public FormulaValue Value { get; }
+        public FormulaValue Value => _value ?? _getFormulaValue().ConfigureAwait(false).GetAwaiter().GetResult();
```
MikeStall, line 20: >.GetResult() [](http://example.com/codeflow?start=100&length=12)

may need to revisit in future PR. 

Existing path calls .Result already.

MikeStall (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Types/AggregateType.cs
```diff
@@ -58,6 +58,26 @@ public virtual bool TryGetFieldType(string name, out FormulaType type)
...
```
CarlosFigueiraMSFT, line 61: >name [](http://example.com/codeflow?start=48&length=4)

suggestion: fieldName (as is defined in GetFieldType). The function above (TryGetFieldType) is explicit named about fields, this one is not.

src/libraries/Microsoft.PowerFx.Core/Public/Values/NamedValue.cs
```diff
@@ -25,7 +44,19 @@ public NamedValue(KeyValuePair<string, FormulaValue> pair)
         public NamedValue(string name, FormulaValue value)
```
CarlosFigueiraMSFT, line 44: >value [](http://example.com/codeflow?start=52&length=5)

Now that this is being used, consider throwing an ArgumentNullException for this argument.

src/libraries/Microsoft.PowerFx.Core/Public/Values/NamedValue.cs
```diff
@@ -25,7 +44,19 @@ public NamedValue(KeyValuePair<string, FormulaValue> pair)
...
         {
             Name = name ?? throw new ArgumentNullException(nameof(name));
-            Value = value;
+            _backingDType = value.Type._type;
```
CarlosFigueiraMSFT, line 47: >value.Type._type [](http://example.com/codeflow?start=28&length=16)

The Value getter indicates that _value can be null. Will this crash here?

jas-valgotar: you are correct, thanks for spotting it.

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
```diff
@@ -43,18 +43,21 @@ public virtual bool TryGetPrimaryKey(out string key)
         private IEnumerable<NamedValue> GetFields()
         {
             foreach (var fieldName in Type.FieldNames)
-            {
-                var formulaValue = GetField(fieldName);
-                yield return new NamedValue(fieldName, formulaValue);
+            {
+                // below will always succeed.
```
CarlosFigueiraMSFT, line 47: >// below will always succeed [](http://example.com/codeflow?start=16&length=28)

How can we guarantee that this will always succeed? AggregateType.TryGetBackingDType can return false.

jas-valgotar: `TryGetBackingDType` can return false but since fieldName is being enumerated from `Type` TryGetBackingDType should always succeed here.

src/libraries/Microsoft.PowerFx.Core/Public/Types/AggregateType.cs
jas-valgotar: updated

src/libraries/Microsoft.PowerFx.Core/Public/Values/NamedValue.cs
jas-valgotar: updated.

src/libraries/Microsoft.PowerFx.Core/Public/Values/NamedValue.cs
CarlosFigueiraMSFT, line 20: >_getFormulaValue() [](http://example.com/codeflow?start=47&length=18)

Should we cache this value instead of calling _getFormulaValue every time Value is accessed in the lazy scenario?

CarlosFigueiraMSFT: Makes sense.

PR #1587, opened 2023-06-15
Update UserAgent with a more precise version

LucGenetier:
Use the product version embedded in the DLL instead of the assembly version
Defaults to assembly version in case we couldn't access the DLL for any security reason

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/PowerPlatformConnectorClient.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/PowerPlatformConnectorClient.cs
@@ -20,23 +20,7 @@ namespace Microsoft.PowerFx.Connectors
         /// <summary>
         /// For telemetry - assembly version stamp. 
         /// </summary>
-        public static string Version
-        {
-            get
-            {
-                if (!string.IsNullOrEmpty(_version))
-                {
-                    return _version;
-                }
-
-                // InformationVersion example: "0.2.6-preview.20230615-1002+048250dc4404bdb749f992cd0869f8821c79e3fe"
-                _version = typeof(PowerPlatformConnectorClient).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion.Split('+')[0];
-
-                return _version;
-            }
-        }
-
-        private static string _version = null;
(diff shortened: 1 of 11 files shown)


src/libraries/Microsoft.PowerFx.Connectors/PowerPlatformConnectorClient.cs
```diff
@@ -18,7 +19,32 @@ public class PowerPlatformConnectorClient : HttpClient
...
+                    return _version;
+                }                
+
+                try
+                {
+                    // Will return "0.2.6-preview.20230615-1002"
+                    // ProductVersion example "0.2.6-preview.20230615-1002+048250dc4404bdb749f992cd0869f8821c79e3fe"
+                    _version = FileVersionInfo.GetVersionInfo(typeof(PowerPlatformConnectorClient).Assembly.Location).ProductVersion.Split('+')[0];
+                }
+                catch (Exception)
```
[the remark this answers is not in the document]
LucGenetier: Here we are accessing the file system and trying to read the DLL, there is no guarantee we'll be allowed to do that from a security standpoint.
I prefer this extra safety net, just in case...

src/libraries/Microsoft.PowerFx.Connectors/PowerPlatformConnectorClient.cs
```diff
@@ -18,7 +20,23 @@ public class PowerPlatformConnectorClient : HttpClient
...
+            get
+            {
+                if (!string.IsNullOrEmpty(_version))
+                {
+                    return _version;
+                }
+
+                // InformationVersion example: "0.2.6-preview.20230615-1002+048250dc4404bdb749f992cd0869f8821c79e3fe"
+                _version = typeof(PowerPlatformConnectorClient).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion.Split('+')[0];
+
```
nguhoa, line 34: can you set this in constructor and set the _version variable to readonly?

LucGenetier: Version is static and our constructors aren't
Not that we couldn't have one static constructor but I don't see the value here

LucGenetier: Fixed

PR #1589, opened 2023-06-15
Update the error message validation for .txt tests

nguhoa:
Fix issue https://github.com/microsoft/Power-Fx-Dataverse/issues/219.
Update logic of error message validation for .txt tests to make the order of message sentences does not matter.

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Mod_Float.txt
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Mod_Float.txt
@@ -32,4 +32,4 @@
 -5.8180083079168396E-201
 
 >> Mod(1E+400, 3)
-Errors: Error 4-10: Numeric value is too large.|Error 0-14: The function 'Mod' has some invalid arguments.|Error 4-10: Invalid argument type (Error). Expecting a Number value instead.
+Errors: Error 4-10: Numeric value is too large.|Error 4-10: Invalid argument type (Error). Expecting a Number value instead.|Error 0-14: The function 'Mod' has some invalid arguments.
--- a/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/BaseRunner.cs
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/BaseRunner.cs
@@ -209,18 +209,30 @@ namespace Microsoft.PowerFx.Core.Tests
                     var expectedCompilerError = expected.StartsWith("Errors: Error") || expected.StartsWith("Errors: Warning"); // $$$ Match error message. 
                     if (expectedCompilerError)
                     {
-                        var msg = $"Errors: " + string.Join("\r\n", runResult.Errors.Select(err => err.ToString()).ToArray());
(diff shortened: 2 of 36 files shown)


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Mod_Float.txt
```diff
@@ -32,4 +32,4 @@
 -5.8180083079168396E-201
 
 >> Mod(1E+400, 3)
-Errors: Error 4-10: Numeric value is too large.|Error 0-14: The function 'Mod' has some invalid arguments.|Error 4-10: Invalid argument type (Error). Expecting a Number value instead.
```
MikeStall, line 35: With the more fleixble comparsion in baserunner.cs, are these changes needed now?

nguhoa: not need, just for testing that even message return are different, we still can pass the check.

nguhoa: Maybe i just modify Mod test and no change for text_format to test this change here.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/BaseRunner.cs
```diff
@@ -208,19 +208,31 @@ private async Task<(TestResult, string)> RunAsync2(TestCase testCase)
...
                         // for tests that are run with and without NumberIsFloat set.
-                        if (actualStr.Contains(expected))
+                        if (isValid)
                         {
                             // Compiler errors result in exceptions
                             return (TestResult.Pass, null);
                         }
-                        else if (NumberIsFloat && expected.StartsWith("Errors:") &&
-                                 actualStr.Contains(Regex.Replace(expected, "(?<!Number,)(\\s|'|\\()Decimal(\\s|'|,|\\.|\\))", "$1Number$2")))
+                        else if (NumberIsFloat && actualStr.Contains(Regex.Replace(expected, "(?<!Number,)(\\s|'|\\()Decimal(\\s|'|,|\\.|\\))", "$1Number$2")))
```
CarlosFigueiraMSFT, line 235: >NumberIsFloat && actualStr.Contains(Regex.Replace(expected, "(?<!Number,)(\\s|'|\\()Decimal(\\s|'|,|\\.|\\))", "$1Number$2") [](http://example.com/codeflow?start=33&length=124)

We should move this check to inside the loop (line 218) - this way it will work with out-of-order messages as well

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestHelpers/BaseRunner.cs
nguhoa: updated.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
🚢

LucGenetier:
✅ No public API change.

Issue #1592, opened 2023-06-16
Backwards compat shims 


MikeStall:
Add some shims to undo some binary breaking changes. 

This is to unblock Dataverse Low code plugins.

PR #1593, opened 2023-06-16
Allow better mutation controls. 

MikeStall:
Before, Set() and Collect()  were both tied ot a single IsMutable concept. 
Need to decouple to allow scenarios like:

Set(Accounts, ....)  // Fail
Collect(Accounts, ... ) // Succeed
Set(ThisRecord.field1, ...) // succeed

Collect([1,2,3,4], ..) // fail
Set([1,2,3,4], ..) // fail

Set(namedFormula1, ...); // fail 

Set(x, [1,2,3,4], ..) // Succeed
Collect(x, {5}) // Succeed

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/NameSymbol.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/NameSymbol.cs
@@ -24,14 +24,14 @@ namespace Microsoft.PowerFx
     [DebuggerDisplay("{Name} ({Owner.DebugName}:{SlotIndex})")]
     internal class NameSymbol : ISymbolSlot
     {
-        public NameSymbol(string name, bool mutable)
+        public NameSymbol(string name, SymbolProperties props)
         {
             Name = name;
-            IsMutable = mutable;
+            Props = props ?? throw new ArgumentNullException(nameof(props));
         }
 
-        public NameSymbol(DName name, bool mutable)
-            : this(name.Value, mutable)
+        public NameSymbol(DName name, SymbolProperties props)
+            : this(name.Value, props)
         {
         }
 
@@ -40,7 +40,7 @@ namespace Microsoft.PowerFx
         /// <summary>
         /// Can this variable be passed to Set(). If not, then this is a binding failure. 
         /// </summary>
-        public bool IsMutable { get; private set; }
+        public SymbolProperties Props { get; private set; }
 
         public ReadOnlySymbolTable Owner { get; set; }
 
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolProperties.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolProperties.cs
@@ -1,0 +1,29 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using Microsoft.PowerFx.Core;
+using Microsoft.PowerFx.Core.Binding;
+using Microsoft.PowerFx.Core.Binding.BindInfo;
+using Microsoft.PowerFx.Core.Entities;
+using Microsoft.PowerFx.Core.Functions;
+using Microsoft.PowerFx.Core.Types.Enums;
+using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx
+{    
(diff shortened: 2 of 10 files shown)


MikeStall:
Partial fix for #1592. 
Provides finer granularity on recent changes around mutation enforcement (#1499)

src/libraries/Microsoft.PowerFx.Core/Public/Config/NameSymbol.cs
```diff
@@ -24,14 +24,14 @@ namespace Microsoft.PowerFx
     [DebuggerDisplay("{Name} ({Owner.DebugName}:{SlotIndex})")]
     internal class NameSymbol : ISymbolSlot
     {
-        public NameSymbol(string name, bool mutable)
+        public NameSymbol(string name, SymbolProperties props)
         {
             Name = name;
```
LucGenetier, line 29: >name [](http://example.com/codeflow?start=19&length=4)

Could add  a check on name as it can't be null

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolProperties.cs
```diff
@@ -0,0 +1,29 @@
...
+    public class SymbolProperties
+    {
+        // Can this symbol be reassigned with a Set() function. 
+        public bool CanSet { get; init; }
+
+        // Can this symbol be passed to Collect, Patch() for mutation 
+        // This only applies to symbols of a tabular type. 
+        public bool CanMutate { get; init; }
+
+        // Is Copyable?
```
LucGenetier, line 27: >// Is Copyable? [](http://example.com/codeflow?start=8&length=15)

?

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolProperties.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using Microsoft.PowerFx.Core;
+using Microsoft.PowerFx.Core.Binding;
+using Microsoft.PowerFx.Core.Binding.BindInfo;
+using Microsoft.PowerFx.Core.Entities;
+using Microsoft.PowerFx.Core.Functions;
+using Microsoft.PowerFx.Core.Types.Enums;
+using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Types;
```
LucGenetier, line 14: <nit> probably don't need all that list

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -77,17 +78,34 @@ internal override bool TryGetVariable(DName name, out NameLookupInfo symbol, out
             return _variables.TryGetValue(lookupName, out symbol);
         }
 
+        // Exists for binary backcompat.
+        public ISymbolSlot AddVariable(string name, FormulaType type, bool mutable = false, string displayName = null)
```
LucGenetier, line 82: >tring name, FormulaType type, [](http://example.com/codeflow?start=40&length=29)

should probably check these params as it's a public API

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -77,17 +78,34 @@ internal override bool TryGetVariable(DName name, out NameLookupInfo symbol, out
...
         /// Value must be provided at runtime.
         /// This can throw an exception in case of there is a conflict in name with existing names.
         /// </summary>
         /// <param name="name"></param>
         /// <param name="type"></param>
-        /// <param name="mutable"></param>
+        /// <param name="props"></param>
         /// <param name="displayName"></param>
-        public ISymbolSlot AddVariable(string name, FormulaType type, bool mutable = false, string displayName = null)
+        public ISymbolSlot AddVariable(string name, FormulaType type, SymbolProperties props, string displayName = null)
```
LucGenetier, line 101: >string name, FormulaType type, SymbolProperties props, [](http://example.com/codeflow?start=39&length=54)

same

src/libraries/Microsoft.PowerFx.Core/Public/Config/NameSymbol.cs
MikeStall: this is internal, caller has it.

LucGenetier (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1596, opened 2023-06-16
Backwards compat shims for Dataverse

### LINKED ISSUE #1592: Backwards compat shims

MikeStall:
Fix #1592 

We should be able to revert this change after DV moves off the old surface.


LucGenetier (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1597, opened 2023-06-16
Allow constant nodes (not only constant string expressions) to be used in IsMatch

CarlosFigueiraMSFT:
This was the previous behavior - we should revert to what we had before. Since IsMatch doesn't need to know the value of the regular expressions at compile-time (unlike Match[All], which need it to determine the output schema), the constant check can be less strict.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IsMatch.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IsMatch.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-using System;
 using System.Collections.Generic;
 using Microsoft.PowerFx.Core.App.ErrorContainers;
 using Microsoft.PowerFx.Core.Binding;
@@ -10,7 +9,6 @@ using Microsoft.PowerFx.Core.Localization;
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
 using Microsoft.PowerFx.Syntax;
-using static Microsoft.PowerFx.Syntax.PrettyPrintVisitor;
 
 #pragma warning disable SA1649 // File name should match first type name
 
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IsMatch.cs
```diff
@@ -10,6 +10,7 @@
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
 using Microsoft.PowerFx.Syntax;
+using static Microsoft.PowerFx.Syntax.PrettyPrintVisitor;
```
nguhoa, line 13: i think you don't need this.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IsMatch.cs
```diff
@@ -41,19 +42,12 @@ public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
...
 
-        public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DType[] argTypes, IErrorContainer errors, out DType returnType, out Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
+        public override void CheckSemantics(TexlBinding binding, TexlNode[] args, DType[] argTypes, IErrorContainer errors)
         {
-            bool fValid = base.CheckTypes(context, args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);            
-
-            TexlNode regExNode = args[1];
-            
-            if ((argTypes[1].Kind != DKind.String && argTypes[1].Kind != DKind.OptionSetValue) || !BinderUtils.TryGetConstantValue(context, regExNode, out var nodeValue))                
+            if ((argTypes[1].Kind != DKind.String && argTypes[1].Kind != DKind.OptionSetValue) || !binding.IsConstant(args[1]))
```
nguhoa, line 45: >argTypes[1] [](http://example.com/codeflow?start=17&length=11)

are we sure that argTypes is not null and length is more than 1? should we check before refer index 1?

CarlosFigueiraMSFT: The function arity (defined in the constructor) defines that it should have at least 2 (and at most 3) arguments, so argTypes.Length is guaranteed to be between 2 and 3.

LucGenetier:
✅ No public API change.

PR #1602, opened 2023-06-21
Allow named formulas to be added to constant cache

rick-nguyen:
Named formulas would always be marked false because of the existing logic for FirstNameNodes always being false unless enum.


LucGenetier:
✅ No public API change.

lesaltzm (approved):

LucGenetier:
✅ No public API change.

PR #1603, opened 2023-06-22
Fix CollectionTableValue to use PrimaryKey when available

LucGenetier:
Related to DV issue 207

With this fix, the following expression are near instantaneous
image

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
@@ -257,9 +257,29 @@ namespace Microsoft.PowerFx.Types
         {
             var ret = true;
 
-            if (baseRecord.TryGetPrimaryKey(out string basePrimaryKey) && currentRecord.TryGetPrimaryKey(out string recordPrimaryKeyValue) && basePrimaryKey == recordPrimaryKeyValue)
+            if (currentRecord.TryGetPrimaryKey(out string currentRecordPrimaryKeyValue))
             {
-                return true;
+                if (currentRecord is not IHasPrimaryKeyName hpk)
+                {
+                    // TryGetPrimaryKey tries to retrieve the primary key value, while IHasPrimaryKey allows the retrieval of the primary key name
+                    // We shouldn't reach this point as IHasPrimaryKey should be implemented on RecordValue, it's just in case, for completeness.
+                    if (baseRecord.TryGetPrimaryKey(out string baseRecordPrimaryKey) && baseRecordPrimaryKey == currentRecordPrimaryKeyValue)
+                    {
+                        return true;
+                    }
+                }
+                else
+                {
+                    string pKey = hpk.GetPrimaryKeyName();
+                    FormulaValue pKeyValue = await baseRecord.GetFieldAsync(pKey, cancellationToken).ConfigureAwait(false);
+
+                    if (pKeyValue?.ToObject()?.ToString() == currentRecordPrimaryKeyValue)
+                    {
+                        return true;
+                    }
+                }
+
+                return false;
             }
 
             if (baseRecord.Fields.Count() != currentRecord.Fields.Count())
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
@@ -254,9 +254,7 @@ namespace Microsoft.PowerFx.Types
         }
 
         protected static async Task<bool> MatchesAsync(RecordValue currentRecord, RecordValue baseRecord, CancellationToken cancellationToken)
-        {
-            var ret = true;
-
+        {            
             if (currentRecord.TryGetPrimaryKey(out string currentRecordPrimaryKeyValue))
             {
                 if (currentRecord is not IHasPrimaryKeyName hpk)
@@ -311,7 +309,10 @@ namespace Microsoft.PowerFx.Types
                 }
                 else if (baseRecordField.Value is RecordValue baseRecordValue && currentFieldValue is RecordValue currentRecordValue)
                 {
-                    ret = await MatchesAsync(currentRecordValue, baseRecordValue, cancellationToken).ConfigureAwait(false);
+                    if (!await MatchesAsync(currentRecordValue, baseRecordValue, cancellationToken).ConfigureAwait(false))
+                    {
+                        return false;
+                    }
                 }
                 else
                 {
@@ -319,7 +320,7 @@ namespace Microsoft.PowerFx.Types
                 }
             }
 
-            return ret;
+            return true;
         }
     }
 }
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
@@ -224,7 +224,7 @@ namespace Microsoft.PowerFx.Types
         {
             cancellationToken.ThrowIfCancellationRequested();
 
-            if (this is IMutationCopy && mutationCopy)
+            if (mutationCopy && this.TryShallowCopy(out _))
             {
                 for (int index = 0; index < _sourceList.Count; index++)
                 {
@@ -257,10 +257,11 @@ namespace Microsoft.PowerFx.Types
         {            
             if (currentRecord.TryGetPrimaryKey(out string currentRecordPrimaryKeyValue))
             {
-                if (currentRecord is not IHasPrimaryKeyName hpk)
+                string pKey = currentRecord.GetPrimaryKeyName();
+
+                if (string.IsNullOrEmpty(pKey))
                 {
-                    // TryGetPrimaryKey tries to retrieve the primary key value, while IHasPrimaryKey allows the retrieval of the primary key name
-                    // We shouldn't reach this point as IHasPrimaryKey should be implemented on RecordValue, it's just in case, for completeness.
+                    // TryGetPrimaryKey tries to retrieve the primary key value, while GetPrimaryKeyName allows the retrieval of the primary key name
                     if (baseRecord.TryGetPrimaryKey(out string baseRecordPrimaryKey) && baseRecordPrimaryKey == currentRecordPrimaryKeyValue)
                     {
                         return true;
@@ -268,7 +269,6 @@ namespace Microsoft.PowerFx.Types
                 }
                 else
                 {
-                    string pKey = hpk.GetPrimaryKeyName();
                     FormulaValue pKeyValue = await baseRecord.GetFieldAsync(pKey, cancellationToken).ConfigureAwait(false);
 
                     if (pKeyValue?.ToObject()?.ToString() == currentRecordPrimaryKeyValue)
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
@@ -153,27 +153,49 @@ namespace Microsoft.PowerFx.Types
             if (_sourceList == null)
             {
                 return await base.RemoveAsync(recordsToRemove, all, cancellationToken).ConfigureAwait(false);
-            }
+            }           
 
             foreach (RecordValue recordToRemove in recordsToRemove)
             {
                 var found = false;
 
-                foreach (var item in _enumerator)
+                if (recordToRemove.TryGetPrimaryKey(out string basePrimaryKey))
                 {
-                    cancellationToken.ThrowIfCancellationRequested();
-
-                    var dRecord = Marshal(item);
-
-                    if (await MatchesAsync(dRecord.Value, recordToRemove, cancellationToken).ConfigureAwait(false))
+                    for (var index = 0; index < _sourceList.Count; index++)
                     {
-                        found = true;
+                        cancellationToken.ThrowIfCancellationRequested();
+                        RecordValue record = Marshal(_sourceIndex[index]).Value;
 
-                        deleteList.Add(item);
-
-                        if (!all)
+                        if (record.TryGetPrimaryKey(out string recordPrimaryKeyValue) && basePrimaryKey == recordPrimaryKeyValue)
                         {
-                            break;
+                            found = true;
+                            deleteList.Add(_sourceIndex[index]);
+
+                            if (!all)
+                            {
+                                break;
+                            }
+                        }
+                    }
+                }
+                else
+                {
+                    foreach (T item in _enumerator)
+                    {
+                        cancellationToken.ThrowIfCancellationRequested();
+
+                        DValue<RecordValue> dRecord = Marshal(item);
+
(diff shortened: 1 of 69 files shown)


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/DatabaseTests.cs
```diff
@@ -0,0 +1,212 @@
...
+        private (PowerFxConfig config, RecalcEngine engine, CheckResult checkResult) CheckExpression(string expr)
+        {
+            PowerFxConfig config = new PowerFxConfig(Features.PowerFxV1);
+            RecalcEngine engine = new RecalcEngine(config);
+
+            config.EnableSetFunction();
+            config.SymbolTable.EnableMutationFunctions();
+
+            RecordType recordType = TestDatabaseRecordValue.CustomRecordType;
+            config.SymbolTable.AddVariable("t", recordType.ToTable(), new SymbolProperties() { CanSet = true, CanMutate = true });
```
MikeStall, line 79: >config.SymbolTable.AddVariable( [](http://example.com/codeflow?start=12&length=31)

engine.UpdateVariable() ? may be simpler

src/tests/Microsoft.PowerFx.Interpreter.Tests/DatabaseTests.cs
```diff
@@ -0,0 +1,212 @@
...
+        {
+            TestDatabaseRecordValue[] records = new TestDatabaseRecordValue[]
+            {
+                new TestDatabaseRecordValue(1, "Luc", "Val1"),
+                new TestDatabaseRecordValue(2, "Mike", "Val2")
+            };
+            TestDatabaseTableValue database = new TestDatabaseTableValue(TestDatabaseRecordValue.CustomRecordType, new List<DValue<RecordValue>>(records.Select(r => DValue<RecordValue>.Of(r))));
+
+            SymbolValues values = new SymbolValues(config.SymbolTable);
+            values.SymbolTable.TryLookupSlot("t", out ISymbolSlot slot);
```
MikeStall, line 98: >values.SymbolTable.TryLookupSlot("t", out ISymbolSlot slot); [](http://example.com/codeflow?start=12&length=60)

coudl get slow from ST.AddVariable

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
```diff
@@ -223,14 +245,34 @@ protected override async Task<DValue<RecordValue>> PatchCoreAsync(RecordValue ba
         /// <remarks>A derived class may override if there's a more efficient way to find the match than by linear scan.</remarks>
         protected virtual async Task<RecordValue> FindAsync(RecordValue baseRecord, CancellationToken cancellationToken, bool mutationCopy = false)
         {
-            if (this is IMutationCopy && mutationCopy)
+            cancellationToken.ThrowIfCancellationRequested();
+
+            if (baseRecord.TryGetPrimaryKey(out string basePrimaryKey))
             {
                 for (var index = 0; index < _sourceList.Count; index++)
```
MikeStall, line 252: >for (var index = 0; index < _sourceList.Count; index++) [](http://example.com/codeflow?start=16&length=55)

_sourceList may be null

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
```diff
@@ -279,14 +320,23 @@ await foreach (var baseRecordField in baseRecord.GetFieldsAsync(cancellationToke
...
                     }
                 }
                 else if (baseRecordField.Value is RecordValue baseRecordValue && currentFieldValue is RecordValue currentRecordValue)
                 {
                     ret = await MatchesAsync(currentRecordValue, baseRecordValue, cancellationToken).ConfigureAwait(false);
                 }
+                else if (baseRecordField.Value is TableValue baseTableValue && currentFieldValue is TableValue currentTableValue)
+                {
+                    // Let's not scan throw entire table, but just check if the row count is the same.
+                    ret = baseTableValue.Count() == currentTableValue.Count();
```
MikeStall, line 333: this impacts language semantics - double check?
Alternative - another way to compare? Or we just ignore tables completely?

src/tests/Microsoft.PowerFx.Interpreter.Tests/DatabaseTests.cs
```diff
@@ -0,0 +1,212 @@
...
+using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Types;
+using Xunit;
+
+namespace Microsoft.PowerFx.Interpreter.Tests
+{
+    public class DatabaseTests
+    {
+        [Fact]
+        public void PatchFirst()
```
MikeStall, line 20: >PatchFirst [](http://example.com/codeflow?start=20&length=10)

test matrix: 2*2 = 4 cases:

- arg0 Table - "dataverse table" vs. in-memory table 
dv table means TableValue overrides key virtuals 

- arg1 baseRecord - dataverse record vs. in-memory record.
dv record means override TryGetPrimaryKey

MikeStall: Also:

// We may need clarification on execptation here, when baseRecord includes another field that doesn't match
Patch( x, { contactid: x.contactid, address1_city: "othercity"} , {address1_city: "Chicago"} )

src/tests/Microsoft.PowerFx.Interpreter.Tests/DatabaseTests.cs
LucGenetier: Works great

src/tests/Microsoft.PowerFx.Interpreter.Tests/DatabaseTests.cs
LucGenetier: UpdateVariable doesn't allow providing SymbolProperties, leading to IMutationCopy requirement which makes things way more complicated

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
```diff
@@ -223,14 +245,34 @@ protected override async Task<DValue<RecordValue>> PatchCoreAsync(RecordValue ba
         /// <remarks>A derived class may override if there's a more efficient way to find the match than by linear scan.</remarks>
         protected virtual async Task<RecordValue> FindAsync(RecordValue baseRecord, CancellationToken cancellationToken, bool mutationCopy = false)
         {
-            if (this is IMutationCopy && mutationCopy)
+            cancellationToken.ThrowIfCancellationRequested();
+
+            if (baseRecord.TryGetPrimaryKey(out string basePrimaryKey))
```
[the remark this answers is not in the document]
LucGenetier: I added the following lines to MatchesAsync but, while this technically works (all UTs are Ok), we lose the performance benefit in this case.
We end up in `RecordValue copyRecord = (RecordValue)record.MaybeShallowCopy();` part of the code which is VERY slow

            if (baseRecord.TryGetPrimaryKey(out string basePrimaryKey) && currentRecord.TryGetPrimaryKey(out string recordPrimaryKeyValue) && basePrimaryKey == recordPrimaryKeyValue)
            {
                return true;
            }

The shallow copy scans all fields of the Entity, following relationships... hence many network calls

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
LucGenetier: After verifying with DV, we don't need this change

src/tests/Microsoft.PowerFx.Interpreter.Tests/DatabaseTests.cs
```diff
@@ -0,0 +1,212 @@
...
+                {
+                    Val = field.Value.ToObject().ToString();
+                }
+            }
+
+            return Task.FromResult(DValue<RecordValue>.Of(this));
+        }
+    }
+
+    internal class TestDatabaseTableValue : CollectionTableValue<DValue<RecordValue>>
```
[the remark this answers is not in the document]
LucGenetier: InMemoryTableValue is implementing IMutationCopy which complicates this scenario where we want to imitate DVTableValue

Ideally we should directly derive from TableValue and implement all CRUD APIs here to have an InMemoryDatabaseValue, based on a Dictionary, with PrimaryKey, etc

src/tests/Microsoft.PowerFx.Interpreter.Tests/DatabaseTests.cs
LucGenetier: Test matrix -> Patch.txt is covering most combinations
Asked Greg about the clarification

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/DatabaseTests.cs
MikeStall: ok - although we could add that overload.

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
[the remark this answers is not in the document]
MikeStall: 1. By skipping maybeShallowCopy, I'm wondering if there is actually a correctness issue here - we may be missing some test cases (see comments in test file)?

2. "The shallow copy scans all fields of the Entity, following relationships... hence many network calls."     A shallow copy most certainly should not be making network calls; this may be hitting some of the other DV bugs that greg filed.

src/tests/Microsoft.PowerFx.Interpreter.Tests/DatabaseTests.cs
MikeStall: But Patch.txt is not covering them with this combination of overloads (TestDatabaseRecordValue)

MikeStall:
>                     ret = await MatchesAsync(currentRecordValue, baseRecordValue, cancellationToken).ConfigureAwait(false);

not your PR  - but this seems fishy ... 

with your other changes, I thought "can we just remove ret" altogether, And then I saw this line. And this is very suspicious...

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs:294 in cbf2421. [](commit_id = cbf2421306a1eefce63bcc4856f69c77f781bf39, deletion_comment = False)

MikeStall (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValue.cs
```diff
@@ -80,7 +80,7 @@ public string ToExpression()
     /// <summary>
     /// Indicates that a FormulaValue should be copied before being mutated, for Copy on Write semantics.
     /// </summary>
-    internal interface IMutationCopy
+    public interface IMutationCopy
```
MikeStall, line 83: >public [](http://example.com/codeflow?start=4&length=6)

this is one thing to consider...

Should we keep this internal. and instead expose as a virtual on RecordValue?

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/DatabaseTests.cs
```diff
@@ -0,0 +1,457 @@
...
+            TestDatabaseTableValue database = new TestDatabaseTableValue(records);
+
+            SymbolValues values = new SymbolValues(config.SymbolTable);
+            values.Set(dbSlot, database);
+
+            return (values, database);
+        }
+    }
+
+    internal class TestDatabaseRecordValue : RecordValue, IHasPrimaryKeyName
```
LucGenetier, line 168: >IHasPrimaryKeyName [](http://example.com/codeflow?start=58&length=18)

Key change

src/tests/Microsoft.PowerFx.Interpreter.Tests/DatabaseTests.cs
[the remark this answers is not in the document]
LucGenetier: class completely rewritten

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValue.cs
LucGenetier: I think this should be public to allow our external customers/users to implement this if needed

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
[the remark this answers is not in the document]
LucGenetier: This is resolved now

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
```diff
@@ -26,7 +26,7 @@ public abstract class RecordValue : ValidFormulaValue, IMutationCopy
         /// Fields and their values directly available on this record. 
         /// The field names should match the names on <see cref="Type"/>. 
         /// </summary>
-        public IEnumerable<NamedValue> Fields => GetFields();
+        public virtual IEnumerable<NamedValue> Fields => GetFields();
```
MikeStall, line 29: >virtual [](http://example.com/codeflow?start=15&length=7)

This is a concern.

We get these from the type, which has the potential virtual. 
We shouldn't have virtual fields per-value.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
LucGenetier: Removed, was for initial testingn only

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
```diff
@@ -251,8 +254,31 @@ protected virtual async Task<RecordValue> FindAsync(RecordValue baseRecord, Canc
...
         protected static async Task<bool> MatchesAsync(RecordValue currentRecord, RecordValue baseRecord, CancellationToken cancellationToken)
-        {
-            var ret = true;
+        {            
+            if (currentRecord.TryGetPrimaryKey(out string currentRecordPrimaryKeyValue))
+            {
+                string pKey = currentRecord.GetPrimaryKeyName();
+
+                if (string.IsNullOrEmpty(pKey))
+                {
```
MikeStall, line 263: just curious: what's an example that hits the true/false branches?

LucGenetier: if you have your own RecordValue and don't override GetPrimaryKeyName

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

PR #1607, opened 2023-06-23
Component Function property - Block access to all DottedNameNode

anhtruong-MSFT:
In order to support function to function reference, we need to allow component control to be returned as a lookupInfo. This, however, also allows reference to all properties of the component. Since we only want function to function reference, we need to restrict all DottedNameNode in the binder.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -3361,12 +3361,15 @@ namespace Microsoft.PowerFx.Core.Binding
                         return;
                     }
 
-                    bool isComponentDataProperty = template.IsComponent &&
+                    bool isBindingPropertyFunctionPropertyOrParameter = (_txb.Property?.IsScopeVariable ?? false) || (_txb.Property?.IsScopedProperty ?? false);
+                    bool isReferencedPropertyDataProperty = !property.IsScopedProperty && !property.IsScopeVariable && property.PropertyCategory == PropertyRuleCategory.Data;
+                    bool isFunctionPropertyReferencingDataProperty = template.IsComponent &&
                         (_txb.Document?.Properties?.EnabledFeatures?.IsEnhancedComponentFunctionPropertyEnabled ?? false) &&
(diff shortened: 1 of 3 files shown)


LucGenetier:
✅ No public API change.

VamsiModem (dismissed):

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -3361,6 +3361,15 @@ public override void PostVisit(DottedNameNode node)
                         return;
                     }
 
+                    bool isBindingPropertyFunctionPropertyOrParameter = template.IsComponent &&
+                        (_txb.Document?.Properties?.EnabledFeatures?.IsEnhancedComponentFunctionPropertyEnabled ?? false) &&
+                        ((_txb.Property?.IsScopeVariable ?? false) || (_txb.Property?.IsScopedProperty ?? false));
```
Shpakh-MSFT, line 3366: Did you validate this with Action and Event properties? Both also have scope parameters.

Shpakh-MSFT: This will likely broke those property scenarios. You need to ensure that current property is function property first.

LucGenetier:
✅ No public API change.

Shpakh-MSFT (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -3361,6 +3361,16 @@ public override void PostVisit(DottedNameNode node)
                         return;
                     }
 
+                    bool isBindingPropertyFunctionPropertyOrParameter = template.IsComponent &&
+                        (_txb.Document?.Properties?.EnabledFeatures?.IsEnhancedComponentFunctionPropertyEnabled ?? false) &&
```
Shpakh-MSFT, line 3365: Please add a comment here about the logic.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

Shpakh-MSFT (approved):
:shipit:

PR #1611, opened 2023-06-27
Connector swagger parser improvements

LucGenetier:
Out of 2515 swagger files: from 678 unsupported files to only 124 where
14 parse errors (duplicate keys or malformed format)
73 unsupported content types (octect-stream, pdf, text, csv, xml, zip, mp3, jpeg...)
14 unsupported parameter format (file)
2 deviation from swagger specification: enum integer (enums are in description field)
1 unsupport OpenApi version
6 duplicated parameters (2 params with same name in different locations)
2 malformatted swagger files
12 OpenApi nuget bug (version string taken for a date)

+
Better handling of unsupported swagger files (detection, error messages...)
Better support for numberIsFloat flag
Refactoring of dynamic intellisense (consolidation of code)
Better support of enums
Better support for option sets
Better support of string and integer formats
Code cleanup

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/ArgumentMapper.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/ArgumentMapper.cs
@@ -81,6 +81,11 @@ namespace Microsoft.PowerFx.Connectors
             {
                 bool hiddenRequired = false;
 
+                if (param == null)
+                {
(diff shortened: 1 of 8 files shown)


LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1613, opened 2023-06-27
Default Parser options should have the expression limit.

jas-valgotar:


LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT:
Do we need to update Power Apps to pass this value?

jas-valgotar:
I should've created Draft, not merging this. Needed this to discuss a DVPlugin issue with Mike.

---
In reply to: [1610309509](https://github.com/microsoft/Power-Fx/pull/1613#issuecomment-1610309509) [](http://example.com/codeflow?ancestors=1610309509,1610309509)

PR #1614, opened 2023-06-27
Ensure that the binding knows the name of the property or named formula it's on, if the name resolver had that information

lesaltzm:
In Canvas Apps, the binding has information about which property it corresponds to, if it's a control property. However, this isn't present for named formulas. Ensure that the property name or named formula name is added, along with the entity information that's already present.


VamsiModem (approved):

LucGenetier:
✅ No public API change.

PR #1615, opened 2023-06-27
Add error messages for formula columns Text function.

gregli-msft:


LucGenetier:
✅ No public API change.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4222,6 +4222,14 @@
     <value>Cannot reference virtual table {0} in formula columns.</value>
     <comment>Error Message. Used in formula columns for a formula that references a virtual table.</comment>
   </data>
+  <data name="FormulaColumns_TextNumberMissingFormat" xml:space="preserve">
```
CarlosFigueiraMSFT, line 4225: This is not the place where those strings should be... PFx assemblies should only have resources that are used by PFx. If this is used only by formula columns, then it should be on that project.

MikeStall (approved):

src/strings/PowerFxResources.en-US.resx
MikeStall: This is being tracked by https://github.com/microsoft/Power-Fx-Dataverse/issues/192

LucGenetier:
✅ No public API change.

PR #1616, opened 2023-06-27
Add new ECS flag to negate recent changes to function property

anhtruong-MSFT:
To prepare for future work that would fix dataflow issue with component funciton property. We need to use an ECS to control rollout and rollback of this fix.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/App/IExternalEnabledFeatures.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/App/IExternalEnabledFeatures.cs
@@ -22,6 +22,8 @@ namespace Microsoft.PowerFx.Core.App
         bool IsDynamicSchemaEnabled { get; }
 
         bool IsEnhancedComponentFunctionPropertyEnabled { get; }
+
+        bool IsComponentFunctionPropertyDataflowEnabled { get; }
     }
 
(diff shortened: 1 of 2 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/App/IExternalEnabledFeatures.cs
```diff
@@ -37,5 +39,7 @@ internal sealed class DefaultEnabledFeatures : IExternalEnabledFeatures
         public bool IsDynamicSchemaEnabled => true;
 
         public bool IsEnhancedComponentFunctionPropertyEnabled => true;
+
+        public bool IsComponentFunctionPropertyDataflowEnabled => true;
```
Shpakh-MSFT, line 43: >true [](http://example.com/codeflow?start=66&length=4)

Should this be false?

Shpakh-MSFT (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/App/IExternalEnabledFeatures.cs
anhtruong-MSFT: DefaultEnabledFeatures is not used by canvas host so this value here can be treated as dummy value.

Issue #1569, opened 2023-06-09
Ban behavior functions in a LookUp / Filter predicate


MikeStall:
https://learn.microsoft.com/en-us/power-platform/power-fx/reference/function-filter-lookup

The predicate in Lookup/Filter may run an unspecified number of numbers.  Semantically, it could run on every row. For delegation, it may run once and results captured and passed in. 

For side-effecting predicates, that could lead to major randomness. 

eg: the Collect() inside the predicate is side-effecting and should be illegal

`LookUp(t1, LocalId=Collect(t1, {  Price : 200}).LocalId).Price`

The make may expect it run  once, but if it's non-delegated, it could run many times. 

This case would require the predicate run every time because of the per-record dependency on Field2. 
`LookUp(t1, LocalId=Collect(t1, {  Price : ThisRecord.Field2}).LocalId).Price`

A maker can be explicit and hoist the behavior property themselves. :

`With({val : Collect(,...)}, LookUp(t1, LocalId=val.LocalId).Price);

MikeStall:
CC: @gregli-msft / @CarlosFigueiraMSFT

PR #1617, opened 2023-06-28
Ban behavior within filter funtions

anderson-joyle:
Issue #1569.


LucGenetier:
✅ No public API change.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5286,4 +5286,8 @@
   <data name="AboutPlainText_text" xml:space="preserve">
     <value>The text to have its HTML and XML tags removed.</value>
   </data>
+  <data name="ErrFilterFunctionBahaviorAsPredicate" xml:space="preserve">
+    <value>Filter functions don't allow behavior function as predicate.</value>
```
CarlosFigueiraMSFT, line 5290: >don't allow behavior function [](http://example.com/codeflow?start=28&length=29)

Suggestion: "...don't allow behavior functions as predicates" (plural in 'function' and 'predicate')

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/FilterDelegationBase.cs
```diff
@@ -184,6 +184,19 @@ protected bool IsValidDelegatableFilterPredicateNode(TexlNode dsNode, TexlBindin
             return true;
         }
 
+        public override void CheckSemantics(TexlBinding binding, TexlNode[] args, DType[] argTypes, IErrorContainer errors)
```
CarlosFigueiraMSFT, line 187: >CheckSemantics [](http://example.com/codeflow?start=29&length=14)

While I think this is the right thing to do, this is technically a breaking change for Power Apps, so we should only enforce this check with PowerFxV1CompatibilityRules.

MikeStall, line 187: >override [](http://example.com/codeflow?start=15&length=8)

Putting on this base class make sense. 

It looks like this will catch: 
CountIf, Filter, LookUp.  (and possibly anything in PA derived that's not in Fx).

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/FilterFunctions_V1CompatDisabled.txt
```diff
@@ -0,0 +1,7 @@
+﻿#SETUP: EnableExpressionChaining,MutationFunctionsTestSetup,disable:PowerFxV1CompatibilityRules
+
+>> With({t1:Table({a:1}), t2:Table({a:1})}, LookUp(t1, a=Collect(t2, {a:2}).a))
+Blank()
+
+>> With({t1:Table({a:1}), t2:Table({a:1})}, Filter(t1, a=Collect(t2, {a:2}).a))
+Table()
```
CarlosFigueiraMSFT: We probably don't need this test - we don't want to specify that this behavior is "expected" - we just allow it for legacy reasons

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/FilterDelegationBase.cs
```diff
@@ -184,6 +184,23 @@ protected bool IsValidDelegatableFilterPredicateNode(TexlNode dsNode, TexlBindin
...
+        {
+            if (binding.Features.PowerFxV1CompatibilityRules)
+            {
+                for (int i = 1; i < args.Length; i++)
+                {
+                    var node = args[i];
+
+                    // If a filter function contains a side effect call as predicate, this is a compilation compilation error.
```
CarlosFigueiraMSFT, line 195: >compilation compilation [](http://example.com/codeflow?start=96&length=23)

duplicate word

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
🚢

LucGenetier:
✅ No public API change.

PR #1621, opened 2023-06-28
Fix recent constant change affecting named enums

rick-nguyen:
Previous PR for named formulas would mark named enums as false even if they were true: https://github.com/microsoft/Power-Fx/pull/1602


src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -2807,13 +2806,14 @@ public override void Visit(FirstNameNode node)
                     return;
                 }
 
+                var isConstantNamedFormula = lookupInfo.Kind == BindKind.Enum;
```
lesaltzm, line 2809: naming: this is just `isConstant`. An enum is not a named formula.

lesaltzm: Or, just set this to false, since you have the enum check lower down.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

Issue #1610, opened 2023-06-26
Fix the precheck for out of memory in Substitute


nguhoa:
With the following expression, we expected precheck return error before run it and throw exception of out of memory:
Len(With({one: "aaaaaaaaaaaaaaaaaa"}, Substitute(Substitute(Substitute(Substitute(Substitute(Substitute(Substitute(Substitute(Substitute(Substitute(one, "a", one), "a", one), "a", one), "a", one), "a", one), "a", one), "a", one), "a", one), "a", one), "a", one)))

Error:
System.OutOfMemoryException: Exception of type 'System.OutOfMemoryException' was thrown.
   at System.Text.StringBuilder.ToString()
   at Microsoft.PowerFx.Functions.Library.SubstituteWorker(EvalVisitor eval, IRContext irContext, StringValue source, StringValue match, StringValue replacement, Int32 instanceNum)
   at Microsoft.PowerFx.Functions.Library.Substitute(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
   at Microsoft.PowerFx.Functions.Library.<>c__DisplayClass466_0`1.<StandardErrorHandling>b__0(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, T[] args)
   at Microsoft.PowerFx.Functions.Library.<>c__DisplayClass463_0`1.<<StandardErrorHandlingAsync>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.PowerFx.EvalVisitor.Visit(CallNode node, EvalVisitorContext context)
   at Microsoft.PowerFx.EvalVisitor.Visit(CallNode node, EvalVisitorContext context)
   at Microsoft.PowerFx.EvalVisitor.Visit(CallNode node, EvalVisitorContext context)
   at Microsoft.PowerFx.EvalVisitor.Visit(CallNode node, EvalVisitorContext context)
   at Microsoft.PowerFx.EvalVisitor.Visit(CallNode node, EvalVisitorContext context)
   at Microsoft.PowerFx.EvalVisitor.Visit(LazyEvalNode node, EvalVisitorContext context)
   at Microsoft.PowerFx.Types.LambdaFormulaValue.EvalInRowScopeAsync(EvalVisitorContext context)
   at Microsoft.PowerFx.Functions.Library.With(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
[...]

PR #1623, opened 2023-06-28
Fix precheck issue in substitute out of memory.

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1610
- Fix logic to throw overflow exception if result length of substitute is more than int.MaxValue.
- Check if result length is more than int.MaxValue (StringBuilder capacity as in https://learn.microsoft.com/en-us/dotnet/api/system.text.stringbuilder.maxcapacity?view=net-6.0), then return runtime error in substitute precheck.
- Added tests for 2 cases above.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
@@ -940,28 +940,21 @@ namespace Microsoft.PowerFx.Functions
                 return sourceLen;
             }
 
-            try
+            checked
             {
-                checked
+                if (replaceAll)
                 {
-                    if (replaceAll)
-                    {
-                        // Replace all instances. 
-                        // Maximum possible length of Substitute, convert all the Match to Replacement. 
-                        // Unicode, so 2B per character.                        
-                        // Round up as conservative estimate. 
-                        maxLenChars = (int)Math.Ceiling((double)sourceLen / matchLen) * replacementLen;
-                    }
-                    else
-                    {
-                        // Only replace 1 instance 
-                        maxLenChars = sourceLen - matchLen + replacementLen;
-                    }
+                    // Replace all instances. 
+                    // Maximum possible length of Substitute, convert all the Match to Replacement. 
+                    // Unicode, so 2B per character.                        
+                    // Round up as conservative estimate. 
+                    maxLenChars = (int)Math.Ceiling((double)sourceLen / matchLen) * replacementLen;
                 }
-            }
-            catch (OverflowException e)
-            {
-                throw e;
+                else
+                {
+                    // Only replace 1 instance 
+                    maxLenChars = sourceLen - matchLen + replacementLen;
+                }
             }
             
             // If not match found, will still be source length 
@@ -996,7 +989,7 @@ namespace Microsoft.PowerFx.Functions
             {
                 maxLenChars = SubstituteGetResultLength(sourceLen, matchLen, replacementLen, instanceNum < 0);
             }
-            catch
+            catch (OverflowException)
             {
                 return CommonErrors.OverflowError(irContext);
             }
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
@@ -15,6 +15,7 @@ using Microsoft.PowerFx.Core.IR;
 using Microsoft.PowerFx.Core.Localization;
 using Microsoft.PowerFx.Core.Utils;
 using Microsoft.PowerFx.Types;
+using static Microsoft.PowerFx.Governor;
 
 namespace Microsoft.PowerFx.Functions
 {
@@ -929,9 +930,9 @@ namespace Microsoft.PowerFx.Functions
         }
 
         // This is static analysis before actually executing, so just use string lengths and avoid contents. 
-        internal static int SubstituteGetResultLength(int sourceLen, int matchLen, int replacementLen, bool replaceAll)
+        internal static double SubstituteGetResultLength(int sourceLen, int matchLen, int replacementLen, bool replaceAll)
         {
-            int maxLenChars;
+            double maxLenChars;
 
             if (matchLen > sourceLen)
             {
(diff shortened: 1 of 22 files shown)


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/SandboxTests.cs
```diff
@@ -180,10 +196,16 @@ public async Task InfiniteLoop(string expr)
...
+        [InlineData(612220032, 1, 18, true, 11019960576)]
+        [InlineData(int.MaxValue, 1, 100, true, 214748364700)]
+        [InlineData(100, 10, int.MaxValue, true, 21474836470)]
+        [InlineData(612220032, 0, 18, true, 612220032)]
+        [InlineData(0, 1, int.MaxValue, true, 0)]
+        [InlineData(5, 3, 0, true, 5)]
         [InlineData(4, 1, 5, false, 8)] // just first,  (4-1+5)
         [InlineData(4, 5, 3, false, 4)]
-        public void TestSubstitutePrePoll(int sourceLen, int matchLen, int replacementLen, bool replaceAll, int expectedMaxLenChars)
+        public void TestSubstitutePrePoll(int sourceLen, int matchLen, int replacementLen, bool replaceAll, double expectedMaxLenChars)
```
MikeStall, line 208: >double [](http://example.com/codeflow?start=108&length=6)

int64?  (still a whole number)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -929,9 +930,9 @@ public static FormulaValue Split(IRContext irContext, StringValue[] args)
         }
 
         // This is static analysis before actually executing, so just use string lengths and avoid contents. 
-        internal static int SubstituteGetResultLength(int sourceLen, int matchLen, int replacementLen, bool replaceAll)
+        internal static double SubstituteGetResultLength(int sourceLen, int matchLen, int replacementLen, bool replaceAll)
```
MikeStall, line 933: >double [](http://example.com/codeflow?start=24&length=6)

int64

MikeStall, line 933: >SubstituteGetResultLength [](http://example.com/codeflow?start=31&length=25)

For overflow,  should this be a checked context? 


We should still catch the overflow exception

[https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/checked-and-unchecked](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/checked-and-unchecked)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -988,7 +989,13 @@ private static FormulaValue Substitute(EvalVisitor runner, EvalVisitorContext co
             var matchLen = match.Value.Length;
             var replacementLen = replacement.Value.Length;
 
-            var maxLenChars = SubstituteGetResultLength(sourceLen, matchLen, replacementLen, instanceNum < 0);
+            var maxLen = SubstituteGetResultLength(sourceLen, matchLen, replacementLen, instanceNum < 0);
+
+            if (!TryGetIntFromDouble(maxLen, out int maxLenChars))
+            {
+                throw new GovernorException($"Memory overuse: Used {maxLen} characters vs. {int.MaxValue} characters allowed.");
+            }
```
MikeStall, line 997: we could catch overflow here and return a runtime error. 

CommonErrors.OverflowError

src/tests/Microsoft.PowerFx.Interpreter.Tests/SandboxTests.cs
nguhoa: keep this as int.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: keep int.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: add check overflow.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -933,34 +933,37 @@ internal static int SubstituteGetResultLength(int sourceLen, int matchLen, int r
...
+                    }
                 }
             }
-            else
+            catch (OverflowException e)
             {
-                // Only replace 1 instance 
-                maxLenChars = sourceLen - matchLen + replacementLen;
+                throw e;
             }
```
MikeStall, line 965: Do we need this?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -987,8 +990,17 @@ private static FormulaValue Substitute(EvalVisitor runner, EvalVisitorContext co
             var sourceLen = source.Value.Length;
             var matchLen = match.Value.Length;
             var replacementLen = replacement.Value.Length;
+            var maxLenChars = sourceLen;
+
+            try
+            {
+                maxLenChars = SubstituteGetResultLength(sourceLen, matchLen, replacementLen, instanceNum < 0);
+            }
+            catch
```
MikeStall, line 999: >catch [](http://example.com/codeflow?start=12&length=5)

This should be: 
catch (OverflowException)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: removed

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
nguhoa: updated

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

Issue #1514, opened 2023-05-18
Filter should be limited to two arguments


gregli-msft:
Today, Filter appears to have an implementation bug that it only supports one argument:

```
> Set( y, [{idx:1, tbl:[1,2,3]}, {idx:2, tbl:[4,5,6]}] )
y:
  idx       tbl
 ===== =========
    1   <table>
    2   <table>

> Filter( y, idx=1 )

  idx       tbl
 ===== =========
    1   <table>

> Filter( y, idx=1, idx=2 )
One or more errors occurred. (Index was outside the bounds of the array.)

> Filter( y, true )

  idx       tbl
 ===== =========
    1   <table>
    2   <table>

> Filter( y, true, true )
One or more errors occurred. (Index was outside the bounds of the array.)
```

We shuoldn't fix this.  :)  The maker should use the And operator (or function or &&) to put multiple arguments together.  

We should make this a compile time error under the V1FxCompat flag.

Limiting to two arguments allows us to add a reduction argument like LookUp has.  Which enables us to push shaping of the result up the stack more easily when making a network call.   The other option is to not have the third arguments for either LookUp or Filter and rely on With and ForAll to do the shaping.

gregli-msft:
Moving to GA 1.0 since the SQL compiler may support this now and we should close that gap.

PR #1625, opened 2023-06-28
Filter function limited to two args

gregli-msft:
Addressing https://github.com/microsoft/Power-Fx/issues/1514.  

Advantages:
- This brings Filter and LookUp into alignment.  
- Using And to combine predicates is easier to understand and read than knowing that extra arguments are And'd together.
- Limiting to two arguments allows us to add a reduction argument like LookUp has.  Which enables us to push shaping of the result up the stack more easily when making a network call.

For C# clients, the Filter function was already effectively limited to two args because the interpreter couldn't handle it (a bug), so no change there.  

For Canvas, this is a breaking Power Fx 1.0 change (under Power FxV1CompatibiltyRules flag) that we'll need to manage later, but only ~13% of filter calls use the three or more argument form.  A back compat converter should be possible.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Filter.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Filter.cs
@@ -58,6 +58,16 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
 
             var fArgsValid = base.CheckTypes(context, args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
 
+            // Only a small percentage of makers use the multi-argument version of Filter in Canvas.
+            // The C# interpreter never supported it.  Blocking this and pushing makers to the more readable
+            // use of "And" to tie multiple predicates together, consistent with LookUp, and freeing up the 
+            // third argument to perform an efficient projection as is also done for LookUp.
+            if (args.Length > 2 && context.Features.PowerFxV1CompatibilityRules)
+            {
+                errors.EnsureError(DocumentErrorSeverity.Severe, args[2], TexlStrings.ErrFilterFunction_OnlyTwoArgs);
+                fArgsValid = false;
+            }
+
             // The first Texl function arg determines the cursor type, the scope type for the lambda params, and the return type.
             fArgsValid &= ScopeInfo.CheckInput(context.Features, args[0], argTypes[0], errors, out var typeScope);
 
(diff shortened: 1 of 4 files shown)


src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Filter.cs
```diff
@@ -58,6 +58,16 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
 
             var fArgsValid = base.CheckTypes(context, args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
 
+            // Only a small percentage of makers use the multi-argument version of Filter in Canvas.
+            // The C# interpreter never supported it.  Blocking this and pushing makers to the more readable
+            // use of "And" to tie multiple predicates together, consistent with LookUp, and freeing up the 
+            // third argument to perform an efficient projection as is also done for LookUp.
+            if (args.Length > 2 && context.Features.PowerFxV1CompatibilityRules)
```
CarlosFigueiraMSFT, line 65: If we plan to backcompat this change into Power Apps, we should add a new feature instead of using this one

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5302,4 +5302,8 @@
     <value>Filter functions don't allow behavior functions as predicates.</value>
     <comment>When a filter function (like LookUp) contains a function as predicate that will trigger a side effect.</comment>
   </data>
+  <data name="ErrFilterFunction_OnlyTwoArgs" xml:space="preserve">
+    <value>Use the And operator to combine multiple predicates into the second argument of Filter.</value>
+    <comment>Limiting the Filter funciton to just two arguments, like the LookUp function, allowing us to add a third argument later also like LookUp.</comment>
```
CarlosFigueiraMSFT, line 5307: Add `{Locked=And}` to the comment; this tells the translation team that this term should not be translated.

LucGenetier:
✅ No public API change.

src/strings/PowerFxResources.en-US.resx
gregli-msft: Done.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Filter.cs
gregli-msft: We talked and because there is a relatively large number of usres of the old form, we should avoid the back compat converter and leave it under V1 compatibility flag.

CarlosFigueiraMSFT (dismissed):
🚢

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

PR #1627, opened 2023-06-28
Create support for delegation warning for offline 

ManojLalavat:
This PR is need to show auther warning when he write powerfx functions which are not supported offline yet.

image


src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/DelegationValidationStrategy.cs
```diff
@@ -402,6 +402,12 @@ protected bool IsValidCallNodeInternal(CallNode node, TexlBinding binding, Opera
             var callInfo = binding.GetInfo(node);
             if (callInfo?.Function != null && ((TexlFunction)callInfo.Function).IsRowScopedServerDelegatable(node, binding, metadata))
             {
+                // Following condition is only applicable for data verse offline support.
```
ian-legler, line 405: >// Following condition is only applicable for data verse offline support. [](http://example.com/codeflow?start=16&length=73)

It doesn't appear that we are actually checking anywhere that the DataSource is dataverse and the app is offline enabled.

ManojLalavat: fixed in the pac PR

jelopezf (approved):

jack-work:
Partly due to lack of staff at the moment, it's the decision of PFx that we'd prefer to stop the delegation-warning-in-Power-Fx buck here.  I have proposed an alternative that allows Manoj to write this logic in PA Client rather than here.  It's not a perfect solution, but it will suffice for the moment.

Ian let me know if you want to weigh in on this.

CarlosFigueiraMSFT:
We talked offline, this has been resolved in another way (using #1661)

Issue #1608, opened 2023-06-25
regression of DateTimeValue, DateValue function


xieofxie:
We should get 1,1,1 for the following format instead 1899, 12, 30

![image](https://github.com/microsoft/Power-Fx/assets/2876650/69eeaf18-9635-47de-b034-794851011d5b)

xieofxie:
0002 did work.. it looks like 0001 causes some bug

![image](https://github.com/microsoft/Power-Fx/assets/2876650/a2711cf5-93c2-4749-942f-a43d37b1d95b)

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/1629

PR #1629, opened 2023-06-29
Fix to use epoch for time value only in datetimevalue/datevalue.

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1608.
Fix the check to make sure we use epoch for time only value.
Add tests to covers all possible cases.


LucGenetier:
✅ No public API change.

LucGenetier (approved):

LucGenetier:
✅ No public API change.

PR #1632, opened 2023-06-29
Added empty table/record boolean column access test

SidAnandMSFT:
### Issue

This is the issue tracking this: https://github.com/microsoft/Power-Fx-Dataverse/issues/189

### Change Description

This PR adds a test that creates an empty record and an empty table and shows that on accessing a column key the evaluated result is Blank rather than false which is different from the behavior observed on power apps canvas.


src/tests/Microsoft.PowerFx.Interpreter.Tests/TableTests.cs
```diff
@@ -74,6 +74,25 @@ private FormulaValue Run_Collect_Worflow(bool serialize)
             return RunExpr("Last(collection).Col2", recalEngine, false);
         }
 
+        [Fact]
+        public async Task PowerFx_Empty_Table_Record_Boolean_Column_Evaluates_False()
+        {
+            RecordType schemaType = RecordType.Empty().Add("BooleanColumn", FormulaType.Boolean);
+            FormulaValue formulaValue = FormulaValue.NewRecordFromFields(schemaType, new List<NamedValue>());
```
nguhoa, line 81: you need to define the field. 

        [Fact]

        public async Task PowerFx_Empty_Table_Record_Boolean_Column_Evaluates_False()

        {

            RecordType schemaType = RecordType.Empty().Add("BooleanColumn", FormulaType.Boolean);
            var fields = new List<NamedValue>
            {
                new NamedValue("BooleanColumn", FormulaValue.New(false))
            }; 


            FormulaValue formulaValue = FormulaValue.NewRecordFromFields(schemaType, fields);

            var recalEngine = new RecalcEngine();

            var result = RunExpr("BooleanRecord.BooleanColumn", recalEngine, true, formulaValue, "BooleanRecord");

 

            Assert.Equal(false, result.ToObject());

 

            var list = new List<RecordValue>
            {   
                FormulaValue.NewRecordFromFields(schemaType, fields)
            };

 

            FormulaValue tableFormulaValue = FormulaValue.NewTable(schemaType, list);

            recalEngine = new RecalcEngine();

            result = RunExpr("First(BooleanTable).BooleanColumn", recalEngine, true, tableFormulaValue, "BooleanTable");


            Assert.Equal(false, result.ToObject());

        }

with this test, you can see that you can get the result of false to the result as you expected

PR #1633, opened 2023-06-29
DV 2011 - Dont allow setting a variable with expandable type

anderson-joyle:
To support https://github.com/microsoft/Power-Fx-Dataverse/issues/211.


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/SetFunction.cs
```diff
@@ -80,6 +78,17 @@ public override void CheckSemantics(TexlBinding binding, TexlNode[] args, DType[
                         return;
                     }
 
+                    if (arg1.IsAggregate)
+                    {
+                        var record = arg1.ToRecord();
+
+                        if (record.GetAllNames(DPath.Root).Any(name => name.Type.IsExpandEntity))
```
MikeStall, line 85: >if (record.GetAllNames(DPath.Root).Any(name => name.Type.IsExpandEntity)) [](http://example.com/codeflow?start=24&length=73)

Can we encapsulate this as a helper on RecordValue? 
We may need to reuse this in other places.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5306,4 +5306,8 @@
     <value>Use the And operator to combine multiple predicates into the second argument.</value>
     <comment>{Locked=And} Limiting the Filter funciton to just two arguments, like the LookUp function, allowing us to add a third argument later also like LookUp.</comment>
   </data>
+  <data name="ErrSetVariableWithRelationshipNotAllow" xml:space="preserve">
+    <value>Can't assign variable from expandable types. For instance, a record from a datasource containing relationships.</value>
+    <comment>This will prevent traversing relationships, which can cause performance problems.</comment>
```
MikeStall, line 5311: >comment [](http://example.com/codeflow?start=5&length=7)

I think <comment> is meant as a hint to localization team; not shown to end users.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5306,4 +5306,8 @@
     <value>Use the And operator to combine multiple predicates into the second argument.</value>
     <comment>{Locked=And} Limiting the Filter funciton to just two arguments, like the LookUp function, allowing us to add a third argument later also like LookUp.</comment>
   </data>
+  <data name="ErrSetVariableWithRelationshipNotAllow" xml:space="preserve">
+    <value>Can't assign variable from expandable types. For instance, a record from a datasource containing relationships.</value>
```
MikeStall, line 5310: >Can't assign variable from expandable types. For instance, a record from a datasource containing relationships. [](http://example.com/codeflow?start=11&length=111)

may need to tweak

anderson-joyle: Suggestions?

src/strings/PowerFxResources.en-US.resx
anderson-joyle: Yes. I was just not sure what to include here.

MikeStall (approved):

LucGenetier:
✅ No public API change.

Issue #1317, opened 2023-04-05
Need to decide if an untyped Blank() is a scalar or a table


gregli-msft:
Which impacts the result of `Left(["abc","def"],Blank())` for which JS and C# currently disagree:
* JS: `["",""]`
* C#: `Blank()`

That despite using the same compiler picking Left_TS for the overload.

It would be fine to pick a scalar or a table.  We just need to decide and consistently apply it.

PR #1635, opened 2023-06-30
Treat untyped blank as a scalar rather than a table

gregli-msft:
Addresses https://github.com/microsoft/Power-Fx/issues/1317 based on excellent suggestions from Carlos.  We will now match Canvas semantics and lean toward an untyped blank being a scalar rather than a table.


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/UntypedBlankAsTable.txt
```diff
@@ -1,42 +1,47 @@
...
 Table({Value:"bc"},{Value:"ef"})
 
 >> Right(["abc","def"],Blank())
-#SKIP: Waiting on https://github.com/microsoft/Power-Fx/issues/1317, JS and C# disagree
+Table({Value:""},{Value:""})
 
 >> Right("abc",[2,3])
 Table({Value:"bc"},{Value:"abc"})
 
+// suspect, may be different on Canvas requiring an override
```
gregli-msft, line 42: There are three tests that give different results in Canvas.  We need to validate that these are the right answers and that Canvas should snap to these semantics with Power Fx 1.0.   @CarlosFigueiraMSFT

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/UntypedBlankAsTable.txt
```diff
@@ -1,42 +1,47 @@
...
 Table({Value:"ab"},{Value:"de"})
 
 >> Left(["abc","def"],Blank())
-#SKIP: Waiting on https://github.com/microsoft/Power-Fx/issues/1317, JS and C# disagree
+Table({Value:""},{Value:""})
 
 >> Left("abcd",[2,3])
 Table({Value:"ab"},{Value:"abc"})
 
+// suspect, may be different on Canvas requiring an override
```
gregli-msft, line 29: There are three tests that give different results in Canvas.  We need to validate that these are the right answers and that Canvas should snap to these semantics with Power Fx 1.0.  @CarlosFigueiraMSFT

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/TabularOverloadsBlanksAndErrors.txt
```diff
@@ -252,10 +251,11 @@ Table({Value:0},{Value:0},{Value:0})
 Table()
 
 >> Concatenate(Blank(), Filter([1], Value > 10))
-Blank()
+Table()
 
 >> Concatenate(Blank(), "test")
 "test"
 
+// suspect, may be different on Canvas requiring an override
```
gregli-msft, line 259: There are three tests that give different results in Canvas.  We need to validate that these are the right answers and that Canvas should snap to these semantics with Power Fx 1.0.  @CarlosFigueiraMSFT

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/TabularOverloadsBlanksAndErrors.txt
CarlosFigueiraMSFT: Checking why this doesn't work properly in canvas.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/UntypedBlankAsTable.txt
CarlosFigueiraMSFT: This works as expected in Power Apps

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/UntypedBlankAsTable.txt
CarlosFigueiraMSFT: This works as expected in Power Apps

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/FindT.txt
```diff
@@ -7,37 +6,37 @@ Errors: Error 0-24: The function 'Find' has some invalid arguments.|Error 14-18:
 Errors: Error 0-25: The function 'Find' has some invalid arguments.|Error 5-9: Invalid argument type (Table). Expecting a Text value instead.|Error 11-15: Invalid argument type (Table). Expecting a Text value instead.
 
 >> Find(If(false, ["blank table"], Blank()), If(false, ["blank table"], Blank()), [1])
-Blank()
+Table({Value:1})
```
CarlosFigueiraMSFT, line 9: This is not right... The first (and second) argument is a blank table. For tabular functions, if a blank table is input, then the result should be blank

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/UntypedBlankAsTable.txt
gregli-msft: My fault, I neglected to turn on Excel Compat.  The test is OK.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/UntypedBlankAsTable.txt
gregli-msft: My fault, I neglected to turn on Excel Compat.  The test is OK.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/TabularOverloadsBlanksAndErrors.txt
gregli-msft: C# has the correct behavior.  Carlos is looking into why Canvas is different.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/FindT.txt
gregli-msft: Resolved by recent Carlos changes.

LucGenetier:
✅ No public API change.

Shpakh-MSFT:
How would this work for below expression? Would return type for below expression be Void?
If(false, [1,2], Blank())

CarlosFigueiraMSFT:
This change affects mostly tabular functions - the logic to determine the return type of If is not changed. In this case, since *[Value:n] _accepts_ ObjNull (the type of Blank()), the result will be *[Value:n].

---
In reply to: [1624171294](https://github.com/microsoft/Power-Fx/pull/1635#issuecomment-1624171294) [](http://example.com/codeflow?ancestors=1624171294,1624171294)

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -578,6 +578,8 @@ internal void AssertValid()
 
         public bool IsTable => Kind == DKind.Table || Kind == DKind.ObjNull || Kind == DKind.LazyTable;
 
+        public bool IsTableStrict => Kind == DKind.Table || Kind == DKind.LazyTable;
```
bryancroteau-MSFT, line 581: In the future, we _might_ have to add a Boolean flag with Kind == DKind.Table || Kind == DKind.ObjNull without the LazyTable one. Then, isn't that one also "Strict?" We are currently not using the word "Strict" anywhere in this document. Could we make it more descriptive? Maybe IsNonBlankTable?

CarlosFigueiraMSFT: Updated it to `IsTableNonObjNull`

bryancroteau-MSFT (approved):

Shpakh-MSFT (changes requested):
:clock1:

LucGenetier:
✅ No public API change.

Shpakh-MSFT (approved):
:shipit:

PR #1642, opened 2023-07-06
[Connectors] Allow Complex types inside arrays

madevoge:


LucGenetier (approved):
:shipit:

LucGenetier:
>         [Fact]

Maybe add an extra test (not necessarily in that file) with the real swagger file that you're using so that we know it's going to work for real after this change

---
Refers to: src/tests/Microsoft.PowerFx.Connectors.Tests/OpenApiJsonSerializerTests.cs:296 in 1168309. [](commit_id = 1168309dfb12b898ce1f3aec8010d62457c95434, deletion_comment = False)

PR #1644, opened 2023-07-06
Improve Intellisense logic for better filtering

LucGenetier:


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall:
1. why is User is special (and not other "globals") 
2. This is in a PA path - curious. why hasn't Power Apps hit this bug?

LucGenetier:
✅ No public API change.

MikeStall:
superceded by #1770

Issue #1640, opened 2023-07-05
More tests for And/Or with behavior functions


MikeStall:
And/Or shortcircuit on errors. 
See https://github.com/microsoft/Power-Fx/blob/f70acfb6b00082486adf5953c978d953ca74a0df/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/AndOrCases.txt#L287

But we should also verify with behavior functions :

false && Collect()  // doesn't collect
true || Collect()  // doesn't collect

Collect1() && Collect2()  // 1, then 2
Collect2() || Collect2()  // 1, then 2

PR #1646, opened 2023-07-06
Testing And/Or along with behavior functions

anderson-joyle:
Issue #1640.


MikeStall (approved):

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationScripts/AndOr.txt
```diff
@@ -0,0 +1,72 @@
+﻿// Case to test how shortcut verification work along with behavior functions
```
CarlosFigueiraMSFT, line 1: All of the `And` tests are short-circuiting (false && <something>), while all of the `Or` tests are not (false Or <something>). We should have tests that short-circuit the Or cases as well.

PR #1647, opened 2023-07-06
Fix unixtime parsing and add real openapi test sample

madevoge:


src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -54,7 +54,12 @@ public static FormulaValue FromJson(JsonElement element, FormulaType formulaType
                     return FormulaValue.NewBlank(formulaType);
 
                 case JsonValueKind.Number:
-                    if ((skipTypeValidation && numberIsFloat) || formulaType is NumberType)
+                    if (formulaType is DateTimeType)
+                    {
+                        // unixtime
+                        return DecimalValue.New(element.GetInt64());
```
MikeStall, line 60: >DecimalValue [](http://example.com/codeflow?start=31&length=12)

This has a type mismatch (FormulaType says we expect a DateTime, but we're returning a Decimal)  - will that be a problem?

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -54,7 +54,12 @@ public static FormulaValue FromJson(JsonElement element, FormulaType formulaType
                     return FormulaValue.NewBlank(formulaType);
 
                 case JsonValueKind.Number:
-                    if ((skipTypeValidation && numberIsFloat) || formulaType is NumberType)
+                    if (formulaType is DateTimeType)
+                    {
+                        // unixtime
```
MikeStall, line 59: >unixtime [](http://example.com/codeflow?start=27&length=8)

There are lots of ways to convert between date/number - is this specified in the swagger spec?
Or do we need some additional metadata in the swagger to know which conversion format to use?

MikeStall:
Superceded by #1649

PR #1649, opened 2023-07-07
[Connectors] Fix unixtime management

LucGenetier:
"UnixTime" integer type will be interpreted as a Decimal instead of a DateTime type, leaving the conversion to the caller.


LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #1655, opened 2023-07-12
Taking expression out of document uris and moving it to a separate property in the LSP payloads

akshar2401:
It was recently discovered that any expression beyond ~30K characters essentially breaks down the language server and all the editing features no longer work. This is due to the fact that expression is included in the document uri which itself is wrapped into json rpc payloads which are then sent to language server from the clients such as new formula bar. The parser used today to parse the Uris enforces that maximum length constraint and language server cannot do anything if that parser generates an exception when uri is length is beyond the maximum length. This pr attempts to move expression out of the document uris and into its own separate property in the json rpc payloads. I did not add any new property but instead used "Text" property that was already present in the extended version of TextDocument. I refactored the object models to pull common properties into base classes to avoid repeating them. I changed the type of TextDocument to its extended version which already had the "Text" property to represent the expression. This extended version is used for custom initial fixup protocol. Additionally, i noticed that an option was added to allow consumers to pass in custom logger Action/callback. This callback is an optional and defaults to dummy function. The log strings generated include payloads sent to language server and all of the suggestions in the case of completion requests. This is dangerous thing to do as raw payloads contain PII information and adding all the suggestions in the log string is very expensive as there are usually big number of suggestions. We would be emitting huge strings in those cases. Temporarily, i did not default logger callback to a dummy function as it would still result in creation of these huge log strings with no effect. I kept logger null and only invoke it when it is not null, thus avoid creating those hug string when we don't need to. This way consumers of language server who do not use this logger callback are not hit by this performance degradation by default. But we should consider reverting these changes or coming up with better approach. 

Note that this pr only makes language server work with large expressions like the ones with 30K characters or more. This doesn't mean language server is faster when dealing with these large expressions. It is infact  slow and we need to come up with a better way to exchange the larger expressions between language server and its clients.

The solution in this pr is backwards compatible and i checked with the cards team and confirmed that changes made in this pr would not break their scenarios

Update: I originally changed the type of "TextDocument" field on the params object to its extended version which had more fields including "Text" which represents an expression. I was seeing Public API Changes warning in the pr so i changed the type of "TextDocument" field back to its original version and added a separate property called "Text" at the same level as "TextDocument". The object models used by Language Server class should never have been public and should completely be internal like the ones are for Semantic tokenization. But i didn't convert them to internal for now as cards backend code still refer them to directly and didn't wanna break their compilation.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
+++ b/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
@@ -691,7 +691,7 @@ namespace Microsoft.PowerFx.LanguageServerProtocol
         /// <returns>Expression.</returns>
         private static string GetExpression(LanguageServerRequestBaseParams requestParams, NameValueCollection queryParams)
         {
-            return requestParams?.TextDocument?.Text ?? queryParams.Get("expression");
+            return requestParams?.Text ?? queryParams.Get("expression");
         }
 
         /// <summary>
--- a/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
+++ b/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
@@ -689,7 +689,7 @@ namespace Microsoft.PowerFx.LanguageServerProtocol
         /// <param name="requestParams">Request params.</param>
         /// <param name="queryParams">Query Params.</param>
         /// <returns>Expression.</returns>
-        private static string GetExpression(BaseParams requestParams, NameValueCollection queryParams)
+        private static string GetExpression(LanguageServerRequestBaseParams requestParams, NameValueCollection queryParams)
         {
             return requestParams?.TextDocument?.Text ?? queryParams.Get("expression");
         }
--- a/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
+++ b/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Specialized;
 using System.Linq;
 using System.Text.Json;
 using System.Web;
@@ -61,7 +62,7 @@ namespace Microsoft.PowerFx.LanguageServerProtocol
             _sendToClient = sendToClient;
             _scopeFactory = scopeFactory;
 
-            _logger = logger ?? ((string s) => { });
+            _logger = logger;
         }
 
         /// <summary>
@@ -71,7 +72,10 @@ namespace Microsoft.PowerFx.LanguageServerProtocol
         {
             Contracts.AssertValue(jsonRpcPayload);
 
-            _logger($"[PFX] OnDataReceived Received: {jsonRpcPayload ?? "<null>"}");
+            // Only log when logger is provided and not null
+            // Creating log string is expensive here, especially for the completion requests
+            // Only create log strings when logger is not null
+            _logger?.Invoke($"[PFX] OnDataReceived Received: {jsonRpcPayload ?? "<null>"}");
 
             string id = null;
             try
@@ -86,14 +90,14 @@ namespace Microsoft.PowerFx.LanguageServerProtocol
 
                     if (!element.TryGetProperty("method", out var methodElement))
                     {
-                        _logger($"[PFX] OnDataReceived CreateErrorResult InvalidRequest (method not found)");
+                        _logger?.Invoke($"[PFX] OnDataReceived CreateErrorResult InvalidRequest (method not found)");
                         _sendToClient(JsonRpcHelper.CreateErrorResult(id, JsonRpcHelper.ErrorCode.InvalidRequest));
                         return;
                     }
 
                     if (!element.TryGetProperty("params", out var paramsElement))
                     {
-                        _logger($"[PFX] OnDataReceived CreateErrorResult InvalidRequest (params not found)");
+                        _logger?.Invoke($"[PFX] OnDataReceived CreateErrorResult InvalidRequest (params not found)");
                         _sendToClient(JsonRpcHelper.CreateErrorResult(id, JsonRpcHelper.ErrorCode.InvalidRequest));
                         return;
                     }
@@ -131,7 +135,7 @@ namespace Microsoft.PowerFx.LanguageServerProtocol
                             HandleRangeDocumentSemanticTokens(id, paramsJson);
                             break;
                         default:
-                            _logger($"[PFX] OnDataReceived CreateErrorResult InvalidRequest (unknown method)");
+                            _logger?.Invoke($"[PFX] OnDataReceived CreateErrorResult InvalidRequest (unknown method)");
                             _sendToClient(JsonRpcHelper.CreateErrorResult(id, JsonRpcHelper.ErrorCode.MethodNotFound));
                             break;
                     }
@@ -139,7 +143,7 @@ namespace Microsoft.PowerFx.LanguageServerProtocol
             }
             catch (Exception ex)
             {
-                _logger($"[PFX] OnDataReceived Exception: {ex.GetType().FullName} - {ex.Message} - {ex.StackTrace}");
+                _logger?.Invoke($"[PFX] OnDataReceived Exception: {ex.GetType().FullName} - {ex.Message} - {ex.StackTrace}");
 
                 LogUnhandledExceptionHandler?.Invoke(ex);
 
@@ -150,7 +154,7 @@ namespace Microsoft.PowerFx.LanguageServerProtocol
 
         private void HandleCommandExecutedRequest(string id, string paramsJson)
         {
-            _logger($"[PFX] HandleCommandExecutedRequest: id={id ?? "<null>"}, paramsJson={paramsJson ?? "<null>"}");
+            _logger?.Invoke($"[PFX] HandleCommandExecutedRequest: id={id ?? "<null>"}, paramsJson={paramsJson ?? "<null>"}");
 
             if (id == null)
             {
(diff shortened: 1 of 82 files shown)


LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.LanguageServerProtocol.Protocol.CodeActionParams.get_TextDocument<br />❌CP0002 M:Microsoft.PowerFx.LanguageServerProtocol.Protocol.TextDocumentPositionParams.get_TextDocument

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -61,7 +62,7 @@ public LanguageServer(SendToClient sendToClient, IPowerFxScopeFactory scopeFacto
             _sendToClient = sendToClient;
             _scopeFactory = scopeFactory;
 
-            _logger = logger ?? ((string s) => { });
+            _logger = logger;
```
LucGenetier, line 65: >_logger = logger; [](http://example.com/codeflow?start=12&length=17)

Key change
This suppresses the need to compute 's' string

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/BaseParams.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+namespace Microsoft.PowerFx.LanguageServerProtocol.Protocol
+{
+    /// <summary>
+    /// Base class to represent the request params.
+    /// </summary>
+    public class BaseParams
```
LucGenetier, line 9: >BaseParams [](http://example.com/codeflow?start=17&length=10)

I'd recommend a more specialized name

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/LanguageServerTests.cs
```diff
@@ -446,12 +446,15 @@ private void CheckBehaviorError(string sentToClientData, bool expectBehaviorErro
...
+        [InlineData("Color.AliceBl", 13, false, true)]
+        [InlineData("Behavior(); Color.AliceBl", 25, true, true)]
 
         // $$$ This test generates an internal error as we use an behavior function but we have no way to check its presence
-        [InlineData("Behavior(); Color.AliceBl", 25, false)]
-        public void TestCompletion(string text, int offset, bool withAllowSideEffects)
+        [InlineData("Behavior(); Color.AliceBl", 25, false, true)]
+        [InlineData("Color.AliceBl", 13, false, false)]
+        [InlineData("Behavior(); Color.AliceBl", 25, true, false)]
+        [InlineData("Behavior(); Color.AliceBl", 25, false, false)]
```
LucGenetier, line 456: Looks like you're duplicating all tests
What about: 
foreach (bool addExprToUri in new[] { true, false })
{ 
...
}

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/LanguageServerTests.cs
```diff
@@ -913,12 +1078,15 @@ public void TestGetPosition()
         }
 
         [Theory]
-        [InlineData("Power(", 6, "Power(2,", 8, false)]
-        [InlineData("Behavior(); Power(", 18, "Behavior(); Power(2,", 20, true)]
+        [InlineData("Power(", 6, "Power(2,", 8, false, true)]
```
LucGenetier, line 1081: >true [](http://example.com/codeflow?start=55&length=4)

Ditto

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
akshar2401: Not sure if you want me to do something here but yes that is the intention. Even if we default logger to no-op function and it is not provided, we would be creating strings that wouldn't be logged. These strings contain PII and are expensive to create especially with dumping the whole payload and all suggestions in the completion function. If we default to null and not a dummy/no-op function, then strings would not be created and logged. Both solutions have same effect of not logging anything but the latter saves us from introducing performance degradation for all consumers of language server by design/default. If _logger is provided then strings would be created and logged no matter what we default it to. That said, we should take another look at this. Do we want to include PII information in these log strings? Do we need to emit huge strings just for logging purposes?

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/LanguageServerTests.cs
akshar2401: What's wrong in writing a theory test? After it was meant for cases like this. Both "true" and "false" are mapped to different tests so we can identify which one failed since both are independent of each other? Not a fan of creating "sub tests" within a test by looping over test cases. If it was the recommended way then all the InlineDatas would be converted to a list of test cases and a loop would be to execute them one by one. Wherever possible, i prefer to use InlineData and write data driven tests. Anyways, i am fine with using loops here.

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/LanguageServerTests.cs
akshar2401: Sure the code is duplicate but they test different scenarios. Duplicated test checks the expression being different in uri and in the new "Text" property and "Text" property is preferred. So if expression are present both in uri and Text property but they are different and a completion operation is performed for example, then results for the expressio in "Text" property is returned instead of the one in uri

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/LanguageServerTests.cs
akshar2401: Refactored

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/LanguageServerTests.cs
```diff
@@ -668,8 +775,10 @@ public void TestCodeAction()
         }
 
         // Test a codefix using a customization, ICodeFixHandler
-        [Fact]
-        public void TestCodeActionWithHandler()
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
```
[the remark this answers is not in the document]
akshar2401: Refactored

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/LanguageServerTests.cs
akshar2401: Refactored to Avoid duplication

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/PublicSurfaceTests.cs
```diff
@@ -78,6 +78,7 @@ public void Test()
                 "Microsoft.PowerFx.LanguageServerProtocol.Protocol.TextEdit",
                 "Microsoft.PowerFx.LanguageServerProtocol.Protocol.VersionedTextDocumentIdentifier",
                 "Microsoft.PowerFx.LanguageServerProtocol.Protocol.WorkspaceEdit",
+                "Microsoft.PowerFx.LanguageServerProtocol.Protocol.BaseParams",
```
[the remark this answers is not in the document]
akshar2401: Done

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/BaseParams.cs
akshar2401: Changed the name to be more specific

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.LanguageServerProtocol.Protocol.CodeActionParams.get_TextDocument<br />❌CP0002 M:Microsoft.PowerFx.LanguageServerProtocol.Protocol.TextDocumentPositionParams.get_TextDocument

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/LanguageServerRequestBaseParams.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+namespace Microsoft.PowerFx.LanguageServerProtocol.Protocol
+{
+    /// <summary>
+    /// Base class to represent the request params.
+    /// </summary>
+    public class LanguageServerRequestBaseParams
```
jas-valgotar, line 9: >public [](http://example.com/codeflow?start=4&length=6)

Can this be kept internal?

akshar2401: I would love to do that but i have no other option but to make it public. The children of this class are public and in use in the backend code of power apps cards. I cannot have an internal parent class and public children. I get compiler errors. As much as i love to, I cannot make the children internal as well since they are used outside of PowerFx. We should work with appropriate owners of cards code to remove this dependency

jas-valgotar (approved):

MikeStall:
> Public API changes: ❌CP0002 M:Microsoft.PowerFx.LanguageServerProtocol.Protocol.CodeActionParams.get_TextDocument❌CP0002 M:Microsoft.PowerFx.LanguageServerProtocol.Protocol.TextDocumentPositionParams.get_TextDocument

Can we avoid the API level breaking changes?

LucGenetier (approved):
:shipit:

LucGenetier:
✅ No public API change.

Issue #1653, opened 2023-07-12
questions about CanPotentiallyCoerceTo


xieofxie:
1. what does DateTimeNoTimeZone type mean? I didn't find related data value type
2. what does Any type mean? It seems non of the types could be coerce to it
3. will we support Number type to Time type? It looks like Number could be converted to DateTime.
4. Perhaps a bug here https://github.com/microsoft/Power-Fx/blob/76c3cade7d5a95cd083edabeb9d582e6a84192d9/src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs#L371C9-L371C9
It should be:
```
if (target == xx)
{
    return xx.AllowedListConvertToXx(source);
}
```
Check https://github.com/microsoft/Power-Fx/blob/76c3cade7d5a95cd083edabeb9d582e6a84192d9/src/libraries/Microsoft.PowerFx.Interpreter/Marshal/TypeCoercionProvider.cs#L99

One of them must be wrong

nguhoa:
Thanks for pointing out the issues.
1: DateTimeNoTimeZone is type that does not use local timezone for datetime.
2: we only can coerce from specific type to the list of types that it allows, we don't allow coerce from one to Any.
3: Number to Time is not supported.
4: We fix it in PR: https://github.com/microsoft/Power-Fx/pull/1656

nguhoa:
Fixed in PR https://github.com/microsoft/Power-Fx/pull/1656

PR #1656, opened 2023-07-12
Check coerce from source to target instead of the opposite.

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1653.
Check coerce from source to target instead of the opposite.
Add some more tests.


MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1661, opened 2023-07-14
Implement Abstract DelegationHintProvider

jack-work:
In the upcoming months there are plans to add multiple warnings which pertain exclusively to Canvas into Power Fx.  While we decide on a strategy for allowing Canvas to extend the warning capabilities provided by the Binder, I am proposing the creation of a new data structure that extends the allows consumers to add custom delegation hints by providing an abstract class which implements virtual methods and invokes them during delegation warning visitation on an instance provided baller the caller on TexlBinding.Run.

Currently, the data structure implements two virtuals, one to visit callndoes and one to visit binary op nodes.  This decision was a consequence of the needs of [this PR](https://github.com/microsoft/Power-Fx/pull/1627) is reaching an imminent deadline.  The call node visitor needs extra information from the binder, which is curious, and while writing this I believe we may want to simply do a second visitation over some rules in Canvas rather than extending the binder.  However, this data structure gives us a means of keeping the new offline-profile delegation warning implementations out of the open source while we develop a long term plan that is most extensible for our users.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -16,6 +16,7 @@ using Microsoft.PowerFx.Core.Entities.QueryOptions;
 using Microsoft.PowerFx.Core.Errors;
 using Microsoft.PowerFx.Core.Functions;
 using Microsoft.PowerFx.Core.Functions.Delegation;
+using Microsoft.PowerFx.Core.Functions.Delegation.DelegationStrategies;
 using Microsoft.PowerFx.Core.Glue;
 using Microsoft.PowerFx.Core.Localization;
 using Microsoft.PowerFx.Core.Texl;
@@ -266,7 +267,8 @@ namespace Microsoft.PowerFx.Core.Binding
             bool updateDisplayNames = false,
             bool forceUpdateDisplayNames = false,
             IExternalRule rule = null,
-            Features features = null)
+            Features features = null,
+            DelegationHintProvider delegationHintProvider = null)
         {
             Contracts.AssertValue(node);
             Contracts.AssertValue(bindingConfig);
@@ -327,6 +329,7 @@ namespace Microsoft.PowerFx.Core.Binding
             HasLocalScopeReferences = false;
             TransitionsFromAsyncToSync = false;
             Rule = rule;
+            DelegationHintProvider = delegationHintProvider;
             if (resolver != null)
             {
                 EntityPath = resolver.CurrentEntityPath;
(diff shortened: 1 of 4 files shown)


src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/DelegationValidationStrategy.cs
```diff
@@ -400,6 +400,11 @@ protected bool IsValidCallNodeInternal(CallNode node, TexlBinding binding, Opera
             }
 
             var callInfo = binding.GetInfo(node);
+            if (binding.DelegationHintProvider.TryGetWarning(node, callInfo.Function, out var warning))
```
CarlosFigueiraMSFT, line 403: >DelegationHintProvider [](http://example.com/codeflow?start=24&length=22)

This can be null, so need to use something like `if(binding.DelegationHintProvider?.TryGetWarning(...) == true)`

src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/OpDelegationStrategy.cs
```diff
@@ -310,6 +310,11 @@ public virtual bool IsSupportedOpNode(TexlNode node, OperationCapabilityMetadata
                 return false;
             }
 
+            if (binding.DelegationHintProvider.TryGetWarning(binaryOpNode, out var warning))
```
CarlosFigueiraMSFT, line 313: >DelegationHintProvider [](http://example.com/codeflow?start=24&length=22)

Can be null - see comment on DelegationValidationStrategy

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -266,7 +267,8 @@ public IExternalControl Control
             bool updateDisplayNames = false,
             bool forceUpdateDisplayNames = false,
             IExternalRule rule = null,
-            Features features = null)
+            Features features = null,
+            DelegationHintProvider DelegationHintProvider = null)
```
CarlosFigueiraMSFT, line 271: >DelegationHintProvider [](http://example.com/codeflow?start=35&length=22)

[nit] `delegationHintProvider` (camelCase)

src/libraries/Microsoft.PowerFx.Core/Binding/DelegationWarningProvider.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// </summary>
+    internal abstract class DelegationHintProvider
+    {
+        public virtual bool TryGetWarning(BinaryOpNode node, out ErrorResourceKey? warningText)
+        {
+            warningText = null;
+            return false;
+        }
+        
+        public virtual bool TryGetWarning(CallNode node, TexlFunction function, out ErrorResourceKey? warningText)
```
CarlosFigueiraMSFT, line 24: >TexlFunction function [](http://example.com/codeflow?start=57&length=21)

We don't really need to pass the function (it can be retrieved from the node), so we can make the interface simpler

CarlosFigueiraMSFT: Nevermind, we don't necessarily have a binding in the provider.

CarlosFigueiraMSFT (dismissed):
:shipit:

lesaltzm (dismissed):

lesaltzm (approved):

CarlosFigueiraMSFT (approved):
🚢

PR #1662, opened 2023-07-14
Update StringInterpolate.txt

sanjeevgoyalmsft:


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/StringInterpolate.txt
```diff
@@ -19,7 +19,7 @@
 "This string {{ has }} curly braces"
 
 >> $""
-""
+Error
```
CarlosFigueiraMSFT, line 22: This is not an error - this is a valid interpolated string.

ian-legler:
These do not appear to be valid changes to the test files.  If these are scenarios where Dataverse SQL diverges from other PowerFx implementations, the correct fix is to update the SQL compiler to match the expected results.

MikeStall:
discussed in scrum

PR #1666, opened 2023-07-17
Update untyped blanks to be treated as scalar when it's possible to be a table or scalar

CarlosFigueiraMSFT:
We updated all other functions to do that, but the statistical functions still behavior still favor the tabular overload: in `Min(Blank(), 1)` the blank value is interpreted as a table (and the operation returns `Blank()`), while there is an overload that takes numbers instead.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/FunctionScopeInfo.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/FunctionScopeInfo.cs
@@ -108,10 +108,21 @@ namespace Microsoft.PowerFx.Core.Functions
             }
             else if (_function.ParamTypes[0].IsTable)
             {
-                if (!typeScope.IsTable)
+                if (features.PowerFxV1CompatibilityRules)
                 {
-                    errors.Error(callNode, TexlStrings.ErrNeedTable_Func, _function.Name);
-                    fArgsValid = false;
+                    if (!typeScope.IsTableNonObjNull)
+                    {
+                        errors.Error(callNode, TexlStrings.ErrNeedTable_Func, _function.Name);
+                        fArgsValid = false;
+                    }
+                }
+                else
+                {
+                    if (!typeScope.IsTable)
+                    {
+                        errors.Error(callNode, TexlStrings.ErrNeedTable_Func, _function.Name);
+                        fArgsValid = false;
+                    }
                 }
 
                 // This assumes that the lambdas operate on the individual records
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/FunctionScopeInfo.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/FunctionScopeInfo.cs
@@ -110,6 +110,7 @@ namespace Microsoft.PowerFx.Core.Functions
             {
                 if (features.PowerFxV1CompatibilityRules)
                 {
+                    // Untyped blank values should not be used to define the scope
                     if (!typeScope.IsTableNonObjNull)
                     {
                         errors.Error(callNode, TexlStrings.ErrNeedTable_Func, _function.Name);
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/FunctionScopeInfo.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/FunctionScopeInfo.cs
@@ -108,22 +108,13 @@ namespace Microsoft.PowerFx.Core.Functions
             }
             else if (_function.ParamTypes[0].IsTable)
             {
-                if (features.PowerFxV1CompatibilityRules)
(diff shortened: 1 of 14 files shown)


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Distinct_V1Compat.txt
```diff
@@ -0,0 +1,9 @@
+﻿#SETUP: PowerFxV1CompatibilityRules
+
+// Untyped blanks are not allowed in the argument that defines the row scope
+>> Distinct(Blank(), true)
```
CarlosFigueiraMSFT, line 4: >Distinct(Blank(), true) [](http://example.com/codeflow?start=3&length=23)

This is a breaking change: we won't allow untyped blanks to define the row scope. There isn't a real scenario where this would be useful, besides language semantic tests.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/FilterFunctions_V1Compat.txt
```diff
@@ -0,0 +1,5 @@
+#SETUP: TableSyntaxDoesntWrapRecords,PowerFxV1CompatibilityRules
+
+// Untyped blanks are not allowed in the argument that defines the row scope
+>> Filter(Blank(), Blank())
```
CarlosFigueiraMSFT, line 4: >Filter(Blank(), Blank()) [](http://example.com/codeflow?start=3&length=24)

This is a breaking change: we won't allow untyped blanks to define the row scope. There isn't a real scenario where this would be useful, besides language semantic tests.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Lookup_V1Compat.txt
```diff
@@ -0,0 +1,10 @@
+﻿#SETUP: DisableMemChecks,TableSyntaxDoesntWrapRecords,PowerFxV1CompatibilityRules
+
+// Untyped blanks are not allowed in the argument that defines the row scope
+>> LookUp(Blank(), Blank())
```
CarlosFigueiraMSFT, line 4: >ookUp(Blank(), Blank() [](http://example.com/codeflow?start=4&length=22)

This is a breaking change: we won't allow untyped blanks to define the row scope. There isn't a real scenario where this would be useful, besides language semantic tests.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/FunctionScopeInfo.cs
```diff
@@ -107,11 +107,23 @@ public virtual bool CheckInput(Features features, TexlNode inputNode, DType inpu
...
-                {
-                    errors.Error(callNode, TexlStrings.ErrNeedTable_Func, _function.Name);
-                    fArgsValid = false;
+            {
+                if (features.PowerFxV1CompatibilityRules)
+                {
+                    // Untyped blank values should not be used to define the scope
+                    if (!typeScope.IsTableNonObjNull)
+                    {
+                        errors.Error(callNode, TexlStrings.ErrNeedTable_Func, _function.Name);
```
lesaltzm, line 116: Could we combine the condition so we don't duplicate this block?

`if ((features.PowerFxV1CompatibilityRules && !typeScope.IsTableNonObjNull) || 
      !typeScope.IsTable)`?

lesaltzm (approved):

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/FilterFunctions.txt
```diff
@@ -49,7 +49,7 @@ Table()
 Table()
 
 // Filter blank table
->> Filter(Blank(), Blank())
+>> Filter(If(1<0,[1,2,3]), Blank())
```
MikeStall, line 52: What is  `Filter(Blank(), Blank())` ?  If that's now a compiler error, we can still include the test and enforce that.

MikeStall (dismissed):

src/libraries/Microsoft.PowerFx.Core/Functions/FunctionScopeInfo.cs
CarlosFigueiraMSFT: Updated it

lesaltzm (approved):

LucGenetier:
✅ No public API change.

PR #1672, opened 2023-07-18
Fix the pack failure due to - The icon file 'icon.png' does not exist in the package - error

ckkashyap:


lesaltzm (changes requested):
You can use src/buildLocalPackages.cmd to generate packages locally. 
That script, as well as our build pipeline adds these properties.

Is there another reason we'd need this change?

PR #1676, opened 2023-07-19
Fix First/Last tests

CarlosFigueiraMSFT:
The tests for First/Last were not running properly - most tests were only run with the `disable:FirstLastNRequiresSecondArguments` setup, which is not run on the V1 groups. This change updates the tests to ensure that most tests should run on all configurations.


MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1678, opened 2023-07-20
Fix #1663 - warn that EnableRegExFunctions is still preview.

### LINKED ISSUE #1663: config.EnableRegExFunctions() doesn't indicate that these functions are Preview
IsMatch, Match, and MatchAll functions are not finished, we need to reconcile the regular expression language here with what is in Canvas.  

We need some way to indicate to a host that these functions are not GA.  Currently, all a host needs to do is call `config.EnableRegExFunctions()` which looks like `config.EnableSetFunction();` and `config.EnableParseJSONFunction();` which are GA.

Either a Engine.Features flag needs to be used, or the name of the enable function needs to be descriptive, something like `config.EnableRegExFunctions_Experimental();`

MikeStall:
Fix #1663


LucGenetier:
✅ No public API change.

LucGenetier (approved):

Issue #1669, opened 2023-07-17
Custom SetProperty function with number value


NavneetThekkumpat:
Running Testengine with this SetProperty
``` 
SetProperty(Rating1.Value, 5);
```
PowerFx.Engine.**Check** and **Eval** functions fail with the below error
<img width="900" height="280" alt="image" src="https://github.com/microsoft/Power-Fx/assets/98551644/42540f95-e2aa-45c8-8b33-f0ea4f9bfb58">

**Workaround right now is setting NumberIsFloat ParserOptions to true**  and it runs successful.

jas-valgotar:
This is related to https://github.com/microsoft/Power-Fx/issues/1659
This should be a fix here: https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/CustomSetPropertyFunction.cs#L72

PR #1682, opened 2023-07-21
Allow Coercion in Set() for numbers.

### LINKED ISSUE #1659: Set() doesn't coerce Decimal to Number and vice versa!
Should `Set(x, 1); Set(x, Float(1))` work?
Today it doesn't. Since Set(x, 1) will type `x` as `decimal`.

jas-valgotar:
-Fixes #1659 #1669

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/CustomSetPropertyFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/CustomSetPropertyFunction.cs
@@ -69,6 +69,12 @@ namespace Microsoft.PowerFx
 
             var arg1 = argTypes[1];
 
+            // coerce number to decimal and vice versa.
+            if (arg0.IsNumeric && arg1.IsNumeric)
+            {
+                return CheckType(context, args[1], argTypes[1], argTypes[0], errors, ref nodeToCoercedTypeMap);
+            }
+
             if (!arg0.Accepts(arg1, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: context.Features.PowerFxV1CompatibilityRules))
             {
                 errors.EnsureError(DocumentErrorSeverity.Critical, args[1], ErrBadType);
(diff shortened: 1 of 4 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/CustomSetPropertyFunction.cs
```diff
@@ -69,6 +69,12 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
 
             var arg1 = argTypes[1];
 
+            // coerce number to decimal and vice versa.
```
CarlosFigueiraMSFT, line 72: >// coerce number to decimal and vice versa. [](http://example.com/codeflow?start=12&length=43)

Why limit coercion to numbers only? Can we allow coercions to other types (e.g., set a GUID to a string variable)?

jas-valgotar: Higher chance of storing runtime errors unknowingly for other types, so I thought initially we could start with numeric and that behavior change feels _almost_ transparent to the end user. 
I am open to expanding it to other types. What's your advice?

MikeStall: I'm leaning towards just numbers, but on the fence. 

 

I know we originally hoped dec/float was "just" ordinary coercion. But we really do have 2 concepts here:

Numeric coercion (float-->dec) is "closer" than general coercion (str-->float). 

 

general coercion is really more of "pasring" than just coercion. And practically, it has a higher chance of maker error, requires a locale, and can introduce more runtime errors that could have been caught at design time.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #1687, opened 2023-07-21
Ensure that untyped blank is interpreted as a scalar for the 'in' operator

CarlosFigueiraMSFT:
Untyped blanks, when used in contexts where they can be interpreted as either a scalar or a complex type (record / table), should be interpreted as scalars. This wasn't working for the `in` operator, and this fixes it. Since this is technically a breaking change (although I can't see why someone would use that), the update is made under the PowerFxV1CompatibilityRules feature.


LucGenetier:
✅ No public API change.

MikeStall (approved):

LucGenetier:
✅ No public API change.

Issue #978, opened 2023-01-09
Review Interpreter override .txt cases 


MikeStall:
Review interpreter's overrides here: 

https://github.com/microsoft/Power-Fx/tree/main/src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterExpressionTestCases

Are there things to fix here?

PR #1689, opened 2023-07-21
Reviewed and updated some overrides test files

gregli-msft:
Resolved https://github.com/microsoft/Power-Fx/issues/978.  After this PR, all overrides are understood with a couple being fixed by an upcoming Carlos Blank handling PR.

This PR will require adding an override file to Canvas, which will be done separately.  Canvas is doing the wrong thing, the override should be there.

Finally, REPL updated to better handle Unicode characters.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1690, opened 2023-07-21
Users/ianleg/date minmax

ian-legler:


src/libraries/Microsoft.PowerFx.Core/Public/Config/Features.cs
```diff
@@ -60,7 +60,12 @@ public sealed class Features
         /// Updates the FirstN/LastN functions to require a second argument, instead of
         /// defaulting to 1.
         /// </summary>
-        internal bool FirstLastNRequiresSecondArguments { get; set; }
+        internal bool FirstLastNRequiresSecondArguments { get; set; }
+
+        /// <summary>
+        /// Allow delegation for Min and Max tabular functions with Date and DateTime fields
+        /// </summary>
+        internal bool AllowDateTimeMinMaxDelegation { get; set; }
```
lesaltzm, line 68: This, and AllowImpure/AllowAsync delegation probably shouldn't be in Features?
They're PA specific, and I'd imagine belong more with the existing IExternal*Features

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/StatisticalTableFunction.cs
```diff
@@ -69,38 +71,60 @@ public override bool IsServerDelegatable(CallNode callNode, TexlBinding binding)
                 }
 
                 return false;
-            }
+            }
+
+            var args = callNode.Args.Children.VerifyValue();
+            var fieldType = binding.GetType(args[1]);
+
+            // if a non-numeric feature is enabled and the type is not a number, also check for the specific sub type capability
```
lesaltzm, line 79: nit: comments should start with a capital letter (I thought this was enforced by the linter, I guess not?)

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/StatisticalTableFunction.cs
```diff
@@ -69,38 +71,60 @@ public override bool IsServerDelegatable(CallNode callNode, TexlBinding binding)
...
+            if (binding.IsFullRecordRowScopeAccess(args[1]))
+            {
+                return GetDottedNameNodeDelegationStrategy().IsValidDottedNameNode(args[1].AsDottedName(), binding, null, null);
+            }
+
+            var firstNameStrategy = GetFirstNameNodeDelegationStrategy().VerifyValue();
+            return firstNameStrategy.IsValidFirstNameNode(args[1].AsFirstName(), binding, null);
+        }
+
+        protected virtual IList<DType> SupportedTypes(TexlBinding binding)
```
lesaltzm, line 117: Likely this should take an IExternalEnabledFeatures or IExternalDocumentProperties object, not the Binding. A binding is a bit more than what should be needed here.

lesaltzm, line 117: Consider: `bool IsSupportedTypeForDelegation(TexlBinding binding, DType)`

Might be better than having the `contains` check in the calling code?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/StatisticalTableFunction.cs
```diff
@@ -69,38 +71,60 @@ public override bool IsServerDelegatable(CallNode callNode, TexlBinding binding)
...
+        {
+            return new List<DType> { DType.Number };
```
CarlosFigueiraMSFT, line 119: >DType.Number  [](http://example.com/codeflow?start=37&length=13)

Do we support it for Decimal as well?

CarlosFigueiraMSFT, line 119: >return new List<DType> { DType.Number }; [](http://example.com/codeflow?start=12&length=40)

How often is IsServerDelegatable called? If a lot, we should consider caching the value of the supported types.

CarlosFigueiraMSFT: Or use McCall's suggestion of using IsSupportedTypeForDelegation(..., DType), and doing simple type comparison in this method.

PR #1693, opened 2023-07-25
Serialization changes to support DV 282

anderson-joyle:
To support https://github.com/microsoft/Power-Fx-Dataverse/issues/282.


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Utils/TypeUtils.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Linq;
+using Microsoft.PowerFx.Core.Types;
+
+namespace Microsoft.PowerFx.Core.Utils
+{
+    internal static class TypeUtils
```
CarlosFigueiraMSFT, line 9: >TypeUtils [](http://example.com/codeflow?start=26&length=9)

This is only needed in StructuralPrint, so there's no need to make it its own class - it can be a "normal" private method in StructuralPrint.

CarlosFigueiraMSFT: Nevermind, saw that it's used in the interpreter (SetFunction). Any reason why it can't be made part of DType itself?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Utils/TypeUtils.cs
anderson-joyle: No particular reason. This was previously implemented in the interpreter and then I had to move it to core.

src/libraries/Microsoft.PowerFx.Core/Logging/StructuralPrint.cs
```diff
@@ -81,14 +81,21 @@ public override LazyList<string> Visit(FirstNameNode node, Precedence parentPrec
...
+            }
+
             if (node.Ident.AtToken == null)
-            {
+            {
+                var type = _binding?.GetType(node);
+
+                if (type != null && type.AggregateHasExpandedType())
+                {
+                    return LazyList<string>.Of($"#$firstnameexp$#");
```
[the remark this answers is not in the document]
anderson-joyle: Please check [https://github.com/microsoft/Power-Fx-Dataverse/pull/283](https://github.com/microsoft/Power-Fx-Dataverse/pull/283). This is the PR containing the test cases.

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.ParseResult.GetAnonymizedFormula

src/libraries/Microsoft.PowerFx.Core/Utils/TypeUtils.cs
CarlosFigueiraMSFT: In the interpreter it made sense to be an extension method - we didn't want to "pollute"  the core project with something that would not be used there. But now that it is, this can be moved to the DType class itself instead of adding a new utility class (we already have quite a few of them)

src/libraries/Microsoft.PowerFx.Core/Logging/StructuralPrint.cs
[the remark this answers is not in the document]
CarlosFigueiraMSFT: Code in this repository should be tested in this repository. We shouldn't rely on other repositories to test that, otherwise we may make changes which would break the other repositories without knowing about it.

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.ParseResult.GetAnonymizedFormula

jas-valgotar (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Parser/ParseResult.cs
```diff
@@ -110,13 +110,5 @@ internal ParseResult(TexlNode root, List<TexlError> errors, bool hasError, List<
...
             return $"Err#{++i} {sb}";
         }));
-
-        /// <summary>
-        /// Converts the current formula into an anonymized format suitable for logging.
-        /// </summary>
-        public string GetAnonymizedFormula()
-        {
-            return StructuralPrint.Print(Root);
-        }
```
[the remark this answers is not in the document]
CarlosFigueiraMSFT: Great point, we shouldn't remove that (sorry, I was the one who suggested it, I didn't realize it was a public method in a public class)

src/libraries/Microsoft.PowerFx.Core/Logging/StructuralPrint.cs
```diff
@@ -81,12 +81,18 @@ public override LazyList<string> Visit(FirstNameNode node, Precedence parentPrec
                 return LazyList<string>.Of(sanitizedName);
             }
 
-            var info = _binding?.GetInfo(node);
+            var info = _binding?.GetInfo(node);
+
+            if (IsExpandedType(_binding, node))
+            {
+                return LazyList<string>.Of($"#$fne$#");
```
LucGenetier, line 88: >$"#$fne$#" [](http://example.com/codeflow?start=43&length=10)

Maybe add a comment on the meaning of this string

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -3848,6 +3848,20 @@ public bool TryGetSimilarName(DName name, FieldNameKind fieldNameKind, out strin
...
+        public bool AggregateHasExpandedType()
+        {
+            var ret = false;
+
+            if (IsAggregate)
+            {
+                var record = ToRecord();
+
+                ret = record.GetAllNames(DPath.Root).Any(name => name.Type.IsExpandEntity);
+            }
```
LucGenetier, line 3860: Will it work if the type is a record with a field being a record itself and having an expandEntity itself (ie. Is this recursive?)

LucGenetier (approved):
:shipit:

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.ParseResult.GetAnonymizedFormula

src/libraries/Microsoft.PowerFx.Core/Parser/ParseResult.cs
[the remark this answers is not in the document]
anderson-joyle: Hosts will still be able to serialize expressions, before or after binding. From what I checked, PA and Cards are not calling it. Other hosts can call the following:

Anonymize before binding:
`check.SetText(expression).ApplyGetLogging()`

Anonymize after binding:
```
check.SetText(script).SetBindingInfo().ApplyBinding();
check.ApplyGetLogging();
```

I can add the code back of course. Just wanted all of us on the same page is terms of capabilities.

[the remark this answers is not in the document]
jas-valgotar: I would slightly lean towards keeping the code, This feels like a Risk without much reward,
While you can do `check.SetText(expression).ApplyGetLogging()`, `ParseResult.GetAnonymizedFormula()` feels cleaner.
Thoughts @CarlosFigueiraMSFT @LucGenetier ?

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
anderson-joyle: > Will it work if the type is a record with a field being a record itself and having an expandEntity itself
Something like `{name: "John", account: First(Accounts)}` will work.

> Is this recursive?
I don't believe it is. We can address this in a new issue since this would be beyond the scope of this one.

src/libraries/Microsoft.PowerFx.Core/Logging/StructuralPrint.cs
anderson-joyle: Comment added.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

PR #1694, opened 2023-07-25
Reorganizing tests to be explicit about dependencies - and remove skipped tests

CarlosFigueiraMSFT:
We have a lot of tests that are currently marked as `#skip` that don't need to be. This change removes *all* of the skipped tests from the core test project, instead updates them to the expected result, and adds explicit overrides in the interpreter project for issues. This reduced the number of skipped interpreter tests from 197 to 21.

This change also splits the 'Power Fx V1' setup with the decimal support. We will introduce PFxV1 support in Power Apps before we introduce decimal support, so we want to be able to run those tests that don't require the decimal support there as well.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
```diff
@@ -28,7 +28,7 @@ public class FileExpressionEvaluationTests : PowerFxTest
         // See the README.md in the ExpressionTestCases directory for more details.
 
         [InterpreterTheory]
-        [TxtFileData("ExpressionTestCases", "InterpreterExpressionTestCases", nameof(InterpreterRunner), "TableSyntaxDoesntWrapRecords,ConsistentOneColumnTableResult,NumberIsFloat")]
+        [TxtFileData("ExpressionTestCases", "InterpreterExpressionTestCases", nameof(InterpreterRunner), "TableSyntaxDoesntWrapRecords,ConsistentOneColumnTableResult,NumberIsFloat,DecimalSupport")]
```
gregli-msft, line 31: Shouldn't this be run without DecimalSupport to test the compiler as Canvas would use it?  Or is DecimalSupport an interpreter only setting?  Either way, it would be good to comment as this looks a little funny for now.

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
```diff
@@ -42,14 +42,14 @@ public void Canvas_Float(ExpressionTestCase testCase)
...
-        [TxtFileData("ExpressionTestCases", "InterpreterExpressionTestCases", nameof(InterpreterRunner), "PowerFxV1,disable:NumberIsFloat")]
+        [TxtFileData("ExpressionTestCases", "InterpreterExpressionTestCases", nameof(InterpreterRunner), "PowerFxV1,disable:NumberIsFloat,DecimalSupport")]
         public void V1_Decimal(ExpressionTestCase testCase)
         {
             RunExpressionTestCase(testCase, Features.PowerFxV1, numberIsFloat: false);
         }
 
         [InterpreterTheory]
-        [TxtFileData("ExpressionTestCases", "InterpreterExpressionTestCases", nameof(InterpreterRunner), "PowerFxV1,NumberIsFloat")]
+        [TxtFileData("ExpressionTestCases", "InterpreterExpressionTestCases", nameof(InterpreterRunner), "PowerFxV1,NumberIsFloat,DecimalSupport")]
```
gregli-msft, line 52: Might be good to have a comment here too, that DecimalSupport means it is avaiable, but not used unless explicitly called upon because of NumberIsFloat.

src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterExpressionTestCases/OpMatrix_Eq_Float_V1CompatDisabled_Overrides.txt
```diff
@@ -0,0 +1,6 @@
+#override: OpMatrix_Eq_Float_V1CompatDisabled.txt
+#SETUP: disable:PowerFxV1CompatibilityRules
+
+// Bug in pre-V1, fixed in V1
```
gregli-msft, line 4: I'm not sure why we need to differentiate pre-V1?  Canvas would have PowerFxV1Compat turned off, don't the #SETUP directives take care of this?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/OpMatrix_Unary_Decimal.txt
```diff
@@ -57,13 +57,13 @@ Error({Kind:ErrorKind.InvalidArgument})
 2.000000000000000000000002
 
 >> -Date(2000,1,1)
-#SKIP: waiting on https://github.com/microsoft/Power-Fx/issues/1294, expected: 36526
```
gregli-msft, line 60: I understand these results.  Can we also add some tests for the numeric values, just to make sure that is coming out correctly.  In this case, `Value(-Date(2000,1,1) )` returning -36526, and the others.

CarlosFigueiraMSFT: Added tests with the value and more comments

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
CarlosFigueiraMSFT: Added a comment

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
CarlosFigueiraMSFT: It's an interpreter support flag. Added a comment to clarify it.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterExpressionTestCases/OpMatrix_Eq_Float_V1CompatDisabled_Overrides.txt
gregli-msft: Ah, got it.  Thanks!

gregli-msft (approved):

Issue #1685, opened 2023-07-21
Remove in ForAll return type issue?


jorisdg:
Set(list, Table({ Name: "One", ID: 1}, { Name: "Two", ID: 2}))
Set(list2, Table(First(list)))

ForAll(list2, Remove(list, ThisRecord))

the return value of this expression:
![image](https://github.com/microsoft/Power-Fx/assets/16008040/f4324f4d-2236-4ef6-85bf-d5408a054d43)

if you get that value and try to iterate the fields on the row like this:
![image](https://github.com/microsoft/Power-Fx/assets/16008040/e7e55b95-b0b6-4cbd-9b4d-fa01f1da0ab1)

you get an exception:
{"InMemoryRecordValue.TryGetField(Value): Wrong field type. Returned b, expected *[ID:w, Name:s]."}

at Microsoft.PowerFx.Types.RecordValue.<GetFields>d__3.MoveNext()\r\n   at System.Linq.Enumerable.Count[TSource](IEnumerable`1 source)\r\n   at (my code)

anderson-joyle:
Other functions are producing a similar error result:
`ForAll(t, Coalesce(Blank()))`

PR #1695, opened 2023-07-26
Fixing Remove in ForAll return type issue

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/1685.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
@@ -1970,9 +1970,9 @@ namespace Microsoft.PowerFx.Functions
                 }
 
                 // Handle the Unknown return type functions: Remove, Coelesce
-                else if (!forceSingleColumn && arg is BlankValue && !tableType.FieldNames.Any())
+                else if (!forceSingleColumn && arg is BlankValue bv2 && !tableType.FieldNames.Any())
                 {
-                    return DValue<RecordValue>.Of(new InMemoryRecordValue(IRContext.NotInSource(recordType)));
+                    return DValue<RecordValue>.Of(bv2);
                 }
 
                 // Handle the single-column-table case. 
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
@@ -1962,6 +1962,7 @@ namespace Microsoft.PowerFx.Functions
 
             return args.Select(arg =>
             {
+                
                 if (!forceSingleColumn && arg is RecordValue rv)
                 {
                     return DValue<RecordValue>.Of(rv);
@@ -1971,9 +1972,20 @@ namespace Microsoft.PowerFx.Functions
                     return DValue<RecordValue>.Of(bv);
                 }
 
-                // Handle the single-column-table case. 
+                // Handle the single-column-table case.
                 var columnName = tableType.SingleColumnFieldName;
-                var defaultField = new NamedValue(columnName, arg);
+                NamedValue defaultField = null;
+
+                if (arg is BlankValue)
+                {
+                    var fieldType = tableType.GetFieldType(columnName);
+                    defaultField = new NamedValue(columnName, FormulaValue.NewBlank(fieldType));
+                }
+                else
+                {
+                    defaultField = new NamedValue(columnName, arg);
+                }                                 
+                
                 return DValue<RecordValue>.Of(new InMemoryRecordValue(IRContext.NotInSource(recordType), new List<NamedValue>() { defaultField }));
             });
         }
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
@@ -1961,8 +1961,7 @@ namespace Microsoft.PowerFx.Functions
             var fieldCounter = tableType.FieldNames.Count();
 
             return args.Select(arg =>
-            {
-                
+            {                
                 if (!forceSingleColumn && arg is RecordValue rv)
                 {
                     return DValue<RecordValue>.Of(rv);
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
@@ -1958,21 +1958,17 @@ namespace Microsoft.PowerFx.Functions
         {
             var tableType = (TableType)irContext.ResultType;
             var recordType = tableType.ToRecord();
+            var fieldCounter = tableType.FieldNames.Count();
+
             return args.Select(arg =>
             {
                 if (!forceSingleColumn && arg is RecordValue rv)
                 {
                     return DValue<RecordValue>.Of(rv);
                 }
-                else if (!forceSingleColumn && arg is BlankValue bv && tableType.FieldNames.Count() > 1)
+                else if (!forceSingleColumn && arg is BlankValue bv && (fieldCounter == 0 || fieldCounter > 1))
                 {
                     return DValue<RecordValue>.Of(bv);
-                }
-
-                // Handle the Unknown return type functions: Remove, Coelesce
-                else if (!forceSingleColumn && arg is BlankValue bv2 && !tableType.FieldNames.Any())
-                {
-                    return DValue<RecordValue>.Of(bv2);
                 }
 
                 // Handle the single-column-table case. 
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
@@ -1964,7 +1964,7 @@ namespace Microsoft.PowerFx.Functions
                 {
                     return DValue<RecordValue>.Of(rv);
                 }
-                else if (!forceSingleColumn && arg is BlankValue bv && bv.Type._type.IsAggregate)
+                else if (!forceSingleColumn && arg is BlankValue bv && bv.Type._type.IsRecord)
                 {
                     return DValue<RecordValue>.Of(bv);
                 }
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
@@ -1958,33 +1958,20 @@ namespace Microsoft.PowerFx.Functions
         {
             var tableType = (TableType)irContext.ResultType;
             var recordType = tableType.ToRecord();
-            var fieldCounter = tableType.FieldNames.Count();
-
             return args.Select(arg =>
-            {                
+            {
                 if (!forceSingleColumn && arg is RecordValue rv)
                 {
                     return DValue<RecordValue>.Of(rv);
                 }
-                else if (!forceSingleColumn && arg is BlankValue bv && (fieldCounter == 0 || fieldCounter > 1))
+                else if (!forceSingleColumn && arg is BlankValue bv && bv.Type._type.IsAggregate)
                 {
                     return DValue<RecordValue>.Of(bv);
                 }
 
-                // Handle the single-column-table case.
+                // Handle the single-column-table case. 
                 var columnName = tableType.SingleColumnFieldName;
-                NamedValue defaultField = null;
-
-                if (arg is BlankValue)
-                {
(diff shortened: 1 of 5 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1967,6 +1967,12 @@ public static IEnumerable<DValue<RecordValue>> StandardTableNodeRecords(IRContex
                 else if (!forceSingleColumn && arg is BlankValue bv && tableType.FieldNames.Count() > 1)
                 {
                     return DValue<RecordValue>.Of(bv);
+                }
+
+                // Handle the Unknown return type functions: Remove, Coelesce
+                else if (!forceSingleColumn && arg is BlankValue bv2 && !tableType.FieldNames.Any())
```
CarlosFigueiraMSFT, line 1973: >&& !tableType.FieldNames.Any() [](http://example.com/codeflow?start=69&length=30)

This can be merged into the previous if statement (`...arg is BlankValue bv && (fieldCount == 0 || fieldCount > 1))`)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1967,6 +1967,12 @@ public static IEnumerable<DValue<RecordValue>> StandardTableNodeRecords(IRContex
                 else if (!forceSingleColumn && arg is BlankValue bv && tableType.FieldNames.Count() > 1)
```
CarlosFigueiraMSFT, line 1967: >tableType.FieldNames.Count() [](http://example.com/codeflow?start=71&length=28)

Can be cached outside the loop (e.g., `var fieldCount = tableType.FieldNames.Count();` after the declaration of tableType) since the value doesn't change

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1967,6 +1967,12 @@ public static IEnumerable<DValue<RecordValue>> StandardTableNodeRecords(IRContex
...
                 {
                     return DValue<RecordValue>.Of(bv);
+                }
+
+                // Handle the Unknown return type functions: Remove, Coelesce
```
CarlosFigueiraMSFT, line 1972: >Coelesce [](http://example.com/codeflow?start=69&length=8)

[typo] `Coalesce`

CarlosFigueiraMSFT, line 1972: >Handle the Unknown return type functions [](http://example.com/codeflow?start=19&length=40)

It's not really an unknown return type - we do know the type. It's mostly untyped blank values.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
CarlosFigueiraMSFT, line 1967: >tableType.FieldNames.Count() > 1 [](http://example.com/codeflow?start=71&length=32)

Do we need this restriction at all? For example, today this throws an exception: `ForAll([1,2,3], If(Value <> 2, {a:Value}))`

src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationScripts/Remove.txt
```diff
@@ -0,0 +1,14 @@
+﻿>> Set(list, Table({ Name: "One", ID: 1}, { Name: "Two", ID: 2}))
+Table({ID:1,Name:"One"},{ID:2,Name:"Two"})
+
+>> Set(list2, Table(First(list)))
+Table({ID:1,Name:"One"})
+
+>> ForAll(list2, Remove(list, ThisRecord))
+Table(Blank())
```
CarlosFigueiraMSFT, line 8: >able(Blank()) [](http://example.com/codeflow?start=1&length=13)

We should add this test, to show that `list` was properly updated:

    >> list
    Table({ID:2,Name:"Two"})

src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationScripts/Remove.txt
```diff
@@ -0,0 +1,14 @@
...
+
+>>Set(x, [1,2,3])
+Table({Value:1},{Value:2},{Value:3})
+
+>> ForAll(x,Remove(x, ThisRecord))
```
CarlosFigueiraMSFT, line 13: >ForAll(x,Remove(x, ThisRecord)) [](http://example.com/codeflow?start=3&length=31)

This is dangerous - we should not let the collection being 'looped' to be updated. Similar to the exception thrown in C# when we do something like

    var coll = new List<int>() { 1, 2, 3 };
    foreach (var i in coll)
    {
        coll.Add(i);
    }

or

    var coll = new List<int>() { 1, 2, 3 };
    foreach (var i in coll)
    {
        coll.Remove(i);
    }

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
anderson-joyle: I would need to investigate this situation as part of another issue.

src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationScripts/Remove.txt
anderson-joyle: Issue [https://github.com/microsoft/Power-Fx/issues/1696](https://github.com/microsoft/Power-Fx/issues/1696) will address this.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
CarlosFigueiraMSFT: This is actually the same issue - blank values that are not handled properly.

src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationScripts/Remove.txt
CarlosFigueiraMSFT: In any case, let's not add this test. Tests should always show what is expected in the product - and that is not.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ForAll.txt
```diff
@@ -109,3 +109,12 @@ Error({Kind:ErrorKind.Numeric})
...
 Error({Kind:ErrorKind.InvalidArgument})
+
+>> ForAll(Table(Blank()),ThisRecord)
+Table(Blank())
+
+>> ForAll(Table(Blank()),Coalesce(ThisRecord))
+Table(Blank())
+
+>> ForAll([1,2,3], If(Value <> 2, {a:Value}))
+Table({a:1},{a:Blank()},{a:3})
```
CarlosFigueiraMSFT, line 120: This is wrong - the expected result should be `Table({a:1},Blank(),{a:3})`. The type of the second argument in ForAll is a record, so its result should not be wrapped.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1,2262 +1,2275 @@
...
+                BuiltinFunctionsCore.TruncT,
+                NoErrorHandling(MultiSingleColumnTable(SimpleFunctionImplementations[BuiltinFunctionsCore.RoundDown], ReplaceBlankWithCallZeroButFloatZeroForSpecificIndices(1)))
+            },
+        };
+
+        public static IEnumerable<DValue<RecordValue>> StandardTableNodeRecords(IRContext irContext, FormulaValue[] args, bool forceSingleColumn)
+        {
+            var tableType = (TableType)irContext.ResultType;
+            var recordType = tableType.ToRecord();
+            var fieldCounter = tableType.FieldNames.Count();
```
MikeStall, line 1961: This file looks like it has a ton of whitespace changes - was that intentional? (it makes it harder to see the changes you actually made; and it will interfere with git blame)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1,2262 +1,2275 @@
...
+
+            return args.Select(arg =>
+            {                
+                if (!forceSingleColumn && arg is RecordValue rv)
+                {
+                    return DValue<RecordValue>.Of(rv);
+                }
+                else if (!forceSingleColumn && arg is BlankValue bv && (fieldCounter == 0 || fieldCounter > 1))
```
MikeStall, line 1969: `fieldCounter >= 0` ?

Or just:
`tableType.FieldNames.Any()`  ?

anderson-joyle: The original logic was checking if the field counter was different from 1. Originally I decided not to change much but turns out this the root of the problem.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ForAll.txt
```diff
@@ -109,3 +109,26 @@ Error({Kind:ErrorKind.Numeric})
...
+Table(Blank())
+
+>> ForAll([1,2,3], If(Value <> 2, {a:Value}))
+Table({a:1},Blank(),{a:3})
+
+>> Concat(ForAll([1,2,3,4,5],If(Value = 1 Or Value = 3 Or Value = 5, {a:Value},Value = 4, {a:Blank()},Blank())),If(IsBlank(ThisRecord), "<<blank record>>",IsBlank(ThisRecord.a), "<<blank value>>",Text(ThisRecord.a)),",")
+"1,<<blank record>>,3,<<blank value>>,5"
+
+>> Concat(ForAll([1,2,3],If(Value=1 Or Value=3,Sequence(Value),Blank())),If(IsBlank(ThisRecord),"<<blank record>>",IsBlank(Value),"<<blank value>>",$"Table with{CountRows(Value)}elements"),", ")
+"Table with1elements, <<blank value>>, Table with3elements"
```
CarlosFigueiraMSFT, line 126: This is covering the same codepath as the test on line 133; your call whether to keep it or remove this one

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ForAll.txt
```diff
@@ -109,3 +109,26 @@ Error({Kind:ErrorKind.Numeric})
...
+Table(Blank())
+
+>> ForAll(Table(Blank()),Coalesce(ThisRecord))
+Table(Blank())
+
+>> ForAll([1,2,3], If(Value <> 2, {a:Value}))
+Table({a:1},Blank(),{a:3})
+
+>> Concat(ForAll([1,2,3,4,5],If(Value = 1 Or Value = 3 Or Value = 5, {a:Value},Value = 4, {a:Blank()},Blank())),If(IsBlank(ThisRecord), "<<blank record>>",IsBlank(ThisRecord.a), "<<blank value>>",Text(ThisRecord.a)),",")
+"1,<<blank record>>,3,<<blank value>>,5"
```
CarlosFigueiraMSFT, line 123: This is covering the same codepath as the test on line 129; your call whether to keep it or remove this one

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ForAll.txt
```diff
@@ -109,3 +109,26 @@ Error({Kind:ErrorKind.Numeric})
...
+
+>> Concat(ForAll([1,2,3],If(Value=1 Or Value=3,Sequence(Value),Blank())),If(IsBlank(ThisRecord),"<<blank record>>",IsBlank(Value),"<<blank value>>",$"Table with{CountRows(Value)}elements"),", ")
```
CarlosFigueiraMSFT, line 125: >{CountRows(Value)} [](http://example.com/codeflow?start=160&length=18)

[minor] space between the number

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1698, opened 2023-07-26
Add null check

anhtruong-MSFT:
A customer reported their app crashing when opening for editing.
The cause is a null reference exception was raised in the CheckSemantic of Filter function.We need to add null check for ViewInfo in Filter CheckSemantic

[Incident-409006522 Details - IcM (microsofticm.com)](https://portal.microsofticm.com/imp/v3/incidents/details/409006522/home)

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Filter.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Filter.cs
@@ -111,8 +111,8 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
                     if (dataSourceInfo != null)
                     {
                         // Verify the view belongs to the same datasource
-                        var viewInfo = argTypes[i].ViewInfo.VerifyValue();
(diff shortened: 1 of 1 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Filter.cs
```diff
@@ -111,8 +111,8 @@ public override void CheckSemantics(TexlBinding binding, TexlNode[] args, DType[
                     if (dataSourceInfo != null)
                     {
                         // Verify the view belongs to the same datasource
-                        var viewInfo = argTypes[i].ViewInfo.VerifyValue();
-                        if (viewInfo.RelatedEntityName != dataSourceInfo.Name)
+                        var viewInfo = argTypes[i].ViewInfo?.VerifyValue();
```
Shpakh-MSFT, line 114: >?.VerifyValue() [](http://example.com/codeflow?start=59&length=15)

Remove the check VerifyValue if you are handling it being null.

PR #1701, opened 2023-07-27
Add delegation warnings for countif and math functions/ Match functions for offline

ManojLalavat:


VamsiModem (approved):

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -723,6 +723,11 @@ public virtual bool IsServerDelegatable(CallNode callNode, TexlBinding binding)
         {
             Contracts.AssertValue(callNode);
             Contracts.AssertValue(binding);
+
+            if (binding.DelegationHintProvider?.TryGetWarning(callNode, this, out var warning) ?? false)
```
ian-legler, line 727: This is a tricky pattern. It might be better to have a helper function like TryGetDelegationWarning that takes all the arguments and wraps this logic and handles nulls, making it harder to mess up

ManojLalavat: This is what used and suggested by powerfx folks. but I can try to see If I can wrap all usage in future PR

VamsiModem (approved):

PR #1702, opened 2023-07-27
Remove the GetUniqueTexlRuntimeName virtual from TexlFunction

CarlosFigueiraMSFT:
This is only used by Power Apps, no need to be in Power Fx


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
```diff
@@ -180,6 +180,12 @@ public IService ParentService
             }
         }
 
+        // Used in tests
+        internal (string Namespace, string Name) GetName()
+        {
+            return (base.Namespace.Name.Value, base.Name);
```
lesaltzm, line 186: If it's just in those tests, can't you change the tests to, for example: 

 

func =>
                {
                    return
                        func.Namespace.Name.Value == "LQA" &&
                        func.Name == "GetAnswersFromText";
                }

lesaltzm (approved):

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
CarlosFigueiraMSFT: Updated it to add internals visibility to the connectors test project

LucGenetier:
✅ No public API change.

PR #1709, opened 2023-07-28
Default rules for scientific notation and decimal, with tests

gregli-msft:
We've decided that:
- Decimal will always show all digits.  As it is an exact representation and limited to 29 digits (either big or small) this seems most appropriate.
- Floating point scientific notation will use `1E+01` notation with a capital `E` to be consistent with Excel.
- The point at which a floating point value switches to scientific notation will be platform dependent.  We will guarantee that 1e15 to 1e-4 will be decimal.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

Issue #1703, opened 2023-07-28
Text function: Block use of format string when input is non-numeric at compile time


gregli-msft:
Today we allow `Text` to be called with a string, Boolean, GUID, etc. along with a format string.  The format string appears to be ignored.   Canvas does the same.
```
> Text( "123.456", "$ ##.####" )
"123.456"

> Text( true, "$ ##.####" )
"true"

> Text( GUID(), "$ ##.####" )
"b9866596-378c-4d19-b79b-b967f39996a7"
```
This is dangerous because Excel supports coercion from string to number in this scenario.  Which is fraught with peril as this is a runtime check:
![image](https://github.com/microsoft/Power-Fx/assets/15149773/b72e8403-2930-49f9-beae-1d390b0ab671)

Where we are different from Excel is that we are strongly typed.  We can use that to provide a better, less error prone experience.

The desired experience:
1. For all types, `Text` without a format string works as it does today.
2. For number, date, date/time, and time - no change.  Format string works as it always has.
3. For all other types, where a format is provided, we will generate a compile time error.  This avoids us needing to support the string coercion to number that Excel supports today, something that we can add in the future if the need arises as we would be removing a portion of this block.

There are tests for this change in https://github.com/microsoft/Power-Fx/blob/gregli/text-excel/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_ExcelCompat.txt.

nguhoa:
Fixed in 1711

PR #1711, opened 2023-07-29
Text: Block use of format string for non-numeric at compile time

nguhoa:
Fixed issue: https://github.com/microsoft/Power-Fx/issues/1703
- Block use of format string for non-numeric/date/datetime/time at compile time.
- Updated existing test that allow format string to non numeric to not supported
- Add new tests


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text.txt
```diff
@@ -4,9 +4,6 @@
 >> Text(123.4, "#.000")
 "123.400"
 
->> Text(Blank(), "#.000")
```
nguhoa, line 7: Not remove these tests, just move and update tests of this file of format string for non numeric input to Text_Format file.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
```diff
@@ -155,30 +159,38 @@ internal static void ValidateFormatArgs(string name, CheckTypesContext checkType
             }
             else if (BinderUtils.TryGetConstantValue(checkTypesContext, args[1], out var formatArg))
             {
-                if (checkTypesContext.Features.PowerFxV1CompatibilityRules)
+                if (AllowedListToUseFormatString.Contains(argTypes[0]) || string.IsNullOrEmpty(formatArg))
```
CarlosFigueiraMSFT, line 162: >if (AllowedListToUseFormatString.Contains(argTypes[0]) || string.IsNullOrEmpty(formatArg)) [](http://example.com/codeflow?start=16&length=90)

We should perform the check even if the format is not constant. For example, `Text(true, varFormat)` should be disallowed, or something like `Text("Hello", Match("d", "d").FullMatch)`.

CarlosFigueiraMSFT: And we should only perform the check if PowerFxV1CompatibilityRules is true; legacy should not change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ParseJson.txt
```diff
@@ -490,22 +490,21 @@ Blank()
 true
 
 >> Text(ParseJSON("1.234"), "#.##")
-"1.23"
+Errors: Error 0-32: The function 'Text' has some invalid arguments.|Warning 25-31: Not supported format for non numeric and datetime in 'Text'.
```
CarlosFigueiraMSFT, line 493: That's an interesting decision we need to make. We have done work to treat untyped objects (that hold numbers) the same as numbers themselves, so we should consider enabling it (at compile time) for numbers, and returning a runtime error if a text/boolean is passed.

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
```diff
@@ -1483,7 +1483,6 @@ public void TexlFunctionTypeSemanticsVarPTypedGlobalWithCoercion()
         [InlineData("Text(123, \"$000.000\")")]
         [InlineData("Text(123, \"dddd, mm/dd/yy, at hh:mm:ss am/pm\")")]
         [InlineData("Text(\"hello world\")")]
-        [InlineData("Text(\"hello world\", \"mm/dd/yyyy\")")]
```
CarlosFigueiraMSFT, line 1486: >    [InlineData("Text(\"hello world\", \"mm/dd/yyyy\")")] [](http://example.com/codeflow?start=4&length=57)

Instead of deleting it, we should move it to the negative test.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
nguhoa: Updated to check this only when PowerFxV1CompatibilityRules is true. I add check at runtime that will cover when format is not constant.  and add tests for this case.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ParseJson.txt
nguhoa: add check for untyped object as well and revert this file change.

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
nguhoa: moved this to negative test.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -339,6 +339,12 @@ public static FormulaValue Text(FormattingInfo formatInfo, IRContext irContext,
             if (args.Length > 1 && args[1] is StringValue fs)
             {
                 formatString = fs.Value;
+
+                if (!string.IsNullOrEmpty(formatString) && !TextFormatUtils.AllowedListToUseFormatString.Contains(args[0].Type._type))
```
CarlosFigueiraMSFT, line 343: >IsNullOrEmpty [](http://example.com/codeflow?start=28&length=13)

Can the format string ever be null?

nguhoa: i think no, i use it to check empty string.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_Format.txt
```diff
@@ -767,4 +767,16 @@ Error({Kind:ErrorKind.BadLanguageCode})
...
+Error({Kind:ErrorKind.InvalidArgument})
+
+>> Text("Hello", If(1 < 0, "d", Blank()))
+""
+
+>> Text("Hello", If(1 < 0, "d", ""))
+"Hello"
+
+>> Text("Hello", If(1 > 0, "d", ""))
+Error({Kind:ErrorKind.InvalidArgument})
```
CarlosFigueiraMSFT, line 782: Wouldn't those be a compile-time error if PowerFxV1 is enabled?

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5216,4 +5216,8 @@
   <data name="ErrSetVariableWithRelationshipNotAllowRecord" xml:space="preserve">
     <value>This record cannot be copied. Instead, create an inline record using {...} notation to copy the individual fields desired.</value>
   </data>
+  <data name="ErrNotSupportedFormat_Func" xml:space="preserve">
+    <value>Not supported format for non numeric and datetime in '{0}'.</value>
```
CarlosFigueiraMSFT, line 5220: >>Not supported format for non numeric and datetime in '{0}' [](http://example.com/codeflow?start=10&length=59)

Suggestion "The second argument of the Text function is only supported for numeric and date/time values"

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5216,4 +5216,8 @@
...
+    <comment>Error Message.</comment>
```
CarlosFigueiraMSFT, line 5221: >Error Message. [](http://example.com/codeflow?start=13&length=14)

Please give more context to the string, to prevent translation bugs.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
```diff
@@ -157,16 +159,24 @@ internal static void ValidateFormatArgs(string name, CheckTypesContext checkType
...
-                        HasDateTimeFmt = false,
-                        HasNumericFmt = false
-                    };
+                        var textFormatArgs = new TextFormatArgs
+                        {
+                            FormatCultureName = null,
+                            FormatArg = null,
+                            HasDateTimeFmt = false,
+                            HasNumericFmt = false
+                        };
```
CarlosFigueiraMSFT, line 170: Not your changes (it was already there), but this initialization is not needed - the value of textFormatArgs will be set by the call to TextFormatUtils.IsValidFormatArg. You can remove this, and change the call below to `...IsValidFormatArg(formatArg, formatCulture: null, defaultLanguage: null, out var textFormatArgs)`

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_Format.txt
nguhoa: I see, i thought i will let it check at runtime so if it pass empty string format then the case still valid.
I will update logic to make it fail at compile time.

src/strings/PowerFxResources.en-US.resx
nguhoa: updated

src/strings/PowerFxResources.en-US.resx
nguhoa: updated as well

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -37,6 +39,8 @@ internal class TextFormatArgs
 
     internal sealed class TextFormatUtils
     {
+        internal static readonly IReadOnlyList<DType> AllowedListToUseFormatString = new DType[] { DType.Number, DType.Decimal, DType.DateTime, DType.Date, DType.Time };
```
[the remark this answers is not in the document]
CarlosFigueiraMSFT: Probably a good question for @gregli-msft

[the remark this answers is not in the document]
nguhoa: As Mike mentioned we can do compile error for now and we can decide what we do with blank later since Greg is OOF and I will be OOF next week.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
```diff
@@ -143,41 +145,45 @@ public override bool CheckTypes(CheckTypesContext checkTypesContext, TexlNode[]
...
+            if (checkTypesContext.Features.PowerFxV1CompatibilityRules && argTypes[0] != DType.UntypedObject && !TextFormatUtils.AllowedListToUseFormatString.Contains(argTypes[0]))
             {
-                // Coerce enum values to string
-                CollectionUtils.Add(ref nodeToCoercedTypeMap, args[1], DType.String);
-            }
-            else if (!DType.String.Accepts(argTypes[1], exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: checkTypesContext.Features.PowerFxV1CompatibilityRules))
-            {
-                errors.EnsureError(DocumentErrorSeverity.Severe, args[1], TexlStrings.ErrStringExpected);
-                isValid = false;
+                if (!hasConstFormat || !string.IsNullOrEmpty(formatArg))
```
CarlosFigueiraMSFT, line 151: >if (!hasConstFormat || !string.IsNullOrEmpty(formatArg)) [](http://example.com/codeflow?start=16&length=56)

In this case, we don't need to check the format. Text can only be called with a first argument of type text / Boolean if there is a single argument (i.e., as a pure conversion function). Something such as Text(aTextValue, "") should be compile-time error as well.

nguhoa: Updated here

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

PR #1712, opened 2023-08-01
Connectors: Updates for Deprecated and Pageable functions

LucGenetier:
1. Deprecated functions
- add info to ServiceFunction
- make sure Intellisense does not suggest deprecated functions
2. Pageable functions
- add info to ServiceFunctions
- implement paging logic


src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/SuggestionHandlers/DottedNameNodeSuggestionHandler.cs
```diff
@@ -177,7 +177,7 @@ private static bool TryGetNamespaceFunctions(TexlNode node, TexlBinding binding,
                 Contracts.AssertValid(firstNameInfo.Name);
 
                 var namespacePath = new DPath().Append(firstNameInfo.Name);
-                functions = binding.NameResolver.LookupFunctionsInNamespace(namespacePath);
+                functions = binding.NameResolver.LookupFunctionsInNamespace(namespacePath).Where(f => f is not ISupportsDeprecatedFunctions sdf || !sdf.IsDeprecated);
```
MikeStall, line 180: >.Where(f => f is not ISupportsDeprecatedFunctions sdf || !sdf.IsDeprecated) [](http://example.com/codeflow?start=90&length=75)

This is in common code with Power Apps - how is PA solving this problem?

MikeStall: discussed offline: 
Power Apps does not yet do this. If it's deprecated, they completely ignore the method - so it's not callable. 

Here, it's still callable, but but just doesn't show in intellisense.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/OpenApiParser.cs
```diff
@@ -358,34 +360,18 @@ internal static List<ServiceFunction> Parse(string functionNamespace, OpenApiDoc
...
-                        cacheTimeoutMs,
-                        isHidden,
-                        parameterOptions,
-                        argMapper.OptionalParamInfo,
-                        argMapper.RequiredParamInfo,
-                        parameterDefaultValues,
-                        "action", //  funcTemplate.ActionName,??
-                        numberIsFloat,
-                        argMapper._parameterTypes)
+#pragma warning disable SA1117 // parameters should be on same line or all on different lines
```
MikeStall, line 365: why this change? particularly if we have a style cop rule against it?

LucGenetier: makes the code unreadable (for me at least,  you might have a different opinion)

we should get rid of this rule + add another one: no more than 10 params to any constructor

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/ConfigExtensions.cs
```diff
@@ -27,7 +26,8 @@ public static class ConfigExtensions
         /// <param name="httpClient">Required iff we want to invoke the API. A client to invoke the endpoints described in the api document. This must handle auth or any other tranforms the API expects.</param>
         /// <param name="cache">A cache to avoid redundant HTTP gets.</param>
         /// <param name="numberIsFloat">Number is float.</param>
-        public static IReadOnlyList<FunctionInfo> AddService(this PowerFxConfig config, string functionNamespace, OpenApiDocument openApiDocument, HttpMessageInvoker httpClient = null, ICachingHttpClient cache = null, bool numberIsFloat = false)
+        /// <param name="maxRows">Max rows.</param>
+        public static IReadOnlyList<FunctionInfo> AddService(this PowerFxConfig config, string functionNamespace, OpenApiDocument openApiDocument, HttpMessageInvoker httpClient = null, ICachingHttpClient cache = null, bool numberIsFloat = false, int maxRows = 1000)
```
anderson-joyle, line 30: >1000 [](http://example.com/codeflow?start=260&length=4)

Can we set this as a constant somewhere? There are other places that this value will be referenced from.

LucGenetier: Fixed

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.ConfigExtensions.AddService(Microsoft.PowerFx.PowerFxConfig,System.String,Microsoft.OpenApi.Models.OpenApiDocument,System.Net.Http.HttpMessageInvoker,Microsoft.PowerFx.Connectors.ICachingHttpClient,System.Boolean)<br />❌CP0002 M:Microsoft.PowerFx.ConfigExtensions.AddService(Microsoft.PowerFx.PowerFxConfig,System.String,Microsoft.PowerFx.Connectors.ConnectorFunction,System.Net.Http.HttpMessageInvoker,Microsoft.PowerFx.Connectors.ICachingHttpClient)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ConnectorFunction.#ctor(Microsoft.OpenApi.Models.OpenApiOperation,System.Boolean,System.String,System.String,System.String,System.Net.Http.HttpMethod,System.String,System.Net.Http.HttpClient,System.Boolean,System.Boolean)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.OpenApiParser.GetFunctions(Microsoft.OpenApi.Models.OpenApiDocument,System.Net.Http.HttpClient,System.Boolean,System.Boolean)

src/libraries/Microsoft.PowerFx.Core/Functions/IHasUnsupportedFunctions.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+namespace Microsoft.PowerFx.Core.Functions
+{
+    internal interface IHasUnsupportedFunctions
+    {
+        bool IsDeprecated { get; }
+
+        bool IsSupported { get; }
```
MikeStall, line 10: >bool [](http://example.com/codeflow?start=8&length=4)

can this be `IsNotSupported`
Because anybody that doesn't implement this interface really has a default value of IsNotSupported=false

src/libraries/Microsoft.PowerFx.Core/Functions/IHasUnsupportedFunctions.cs
```diff
@@ -0,0 +1,14 @@
...
+
+        string NotSupportedReason { get; }
```
MikeStall, line 12: >string [](http://example.com/codeflow?start=8&length=6)

Is this is a message we show users in intellisense?
Does that mean we need to worry about the locale?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/IAsyncTexlFunction2.cs
```diff
@@ -12,5 +12,7 @@ namespace Microsoft.PowerFx.Core.Functions
     internal interface IAsyncTexlFunction2
     {
         Task<FormulaValue> InvokeAsync(FormattingInfo context, FormulaValue[] args, CancellationToken cancellationToken);
+
+        Task<FormulaValue> InvokeAsync(string url, CancellationToken cancellationToken);
```
MikeStall, line 16: >InvokeAsync [](http://example.com/codeflow?start=27&length=11)

this should probably be on its own interface.

src/libraries/Microsoft.PowerFx.Core/Functions/IHasUnsupportedFunctions.cs
LucGenetier: I can add a comment to mention this (internal use only)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/IAsyncTexlFunction2.cs
LucGenetier: Will do IAsyncTexlFunction3 : IAsyncTexlFunction2 with it

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
```diff
@@ -198,8 +198,8 @@ internal async Task<FormulaValue> GetFieldAsync(FormulaType fieldType, string fi
         /// <param name="fieldName">Name of the field.</param>        
         /// <param name="result"></param>
         /// <returns>true if field is present, else false.</returns>
-        protected abstract bool TryGetField(FormulaType fieldType, string fieldName, out FormulaValue result);
-        
+        protected internal abstract bool TryGetField(FormulaType fieldType, string fieldName, out FormulaValue result);
```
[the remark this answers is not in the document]
LucGenetier: In PagedRecordValue.cs, line 31 when calling TryGetField we get "Compiler Error CS1540" if we are only "protected".

src/libraries/Microsoft.PowerFx.Interpreter/Functions/IAsyncTexlFunction2.cs
MikeStall: We could give it a more specific name - this is just for URLs and pageable entries?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -210,14 +210,19 @@ public static FormulaValue CountRows(IRContext irContext, FormulaValue[] args)
...
-            {
-                var error = table.Rows.Where(r => r.IsError).Select(r => r.Error).FirstOrDefault();
-                if (error != null)
-                {
-                    return error;
-                }
-
-                var count = table.Count();
+            {
+                int count = 0;
```
MikeStall, line 214: >int [](http://example.com/codeflow?start=16&length=3)

this is good - takes 2 pass and does in 1 pass.

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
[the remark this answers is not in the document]
MikeStall: right, but why do we need the change that's causing CS1540?

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
```diff
@@ -46,18 +47,27 @@ internal sealed class ServiceFunction : BuiltinFunction, IAsyncTexlFunction2
         private readonly WeakReference<IService> _parentService;
         private readonly string _actionName;
         private readonly bool _numberIsFloat;
+        private readonly string _pageLink;
```
MikeStall, line 50: >string [](http://example.com/codeflow?start=25&length=6)

Can you comment what exactly does this field mean?

Is it a url? a string used to create a url?

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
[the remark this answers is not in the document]
LucGenetier: We define PagedRecord/TableValue based on any RecordValue hence the need to call a virtual.

What alternative would we have if not doing that?

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
LucGenetier: None of that, I'll add a comment

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
```diff
@@ -425,26 +440,59 @@ public bool TryGetParamDefaultValue(string paramName, out Tuple<string, DType> d
...
+            if (result is RecordValue rv && IsPageable && rv.TryGetField(FormulaType.String, _pageLink, out FormulaValue pageLink))
+            {
+                string nextLink = (pageLink as StringValue)?.Value;
+                result = new PagedRecordValue(rv, () => GetNextPage(nextLink, cancellationToken), _maxRows);
+            }
+
+            return result;
+        }
+
+        private FormulaValue GetNextPage(string nextLink, CancellationToken cancellationToken)
```
MikeStall, line 462: >FormulaValue [](http://example.com/codeflow?start=16&length=12)

can this be more strongly typed? Record or Table?

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
```diff
@@ -46,18 +47,27 @@ internal sealed class ServiceFunction : BuiltinFunction, IAsyncTexlFunction2
...
         public Dictionary<string, TypedName> OptionalParamInfo => _optionalParamInfo;
         public override Capabilities Capabilities => Capabilities.OutboundInternetAccess | Capabilities.EnterpriseAuthentication | Capabilities.PrivateNetworkAccess;
         public override bool IsHidden => _isHidden;
         public override bool IsSelfContained => !_isBehaviorOnly;
-
-        public ServiceFunction(IService parentService, DPath theNamespace, string name, string localeSpecificName, string description,
-            DType returnType, BigInteger maskLambdas, int arityMin, int arityMax, bool isBehaviorOnly, bool isAutoRefreshable, bool isDynamic, bool isCacheEnabled, int cacheTimetoutMs, bool isHidden,
-            Dictionary<TypedName, List<string>> parameterOptions, ServiceFunctionParameterTemplate[] optionalParamInfo, ServiceFunctionParameterTemplate[] requiredParamInfo,
-            Dictionary<string, Tuple<string, DType>> parameterDefaultValues, string actionName = "", bool numberIsFloat = false, params DType[] paramTypes)
+        public bool IsPageable => !string.IsNullOrEmpty(_pageLink);
```
MikeStall, line 62: >IsPageable [](http://example.com/codeflow?start=20&length=10)

Can you comment what the invariants are here?

For example, does Pageable mean that at the Fx level, it must return a TableValue?

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
```diff
@@ -46,18 +47,27 @@ internal sealed class ServiceFunction : BuiltinFunction, IAsyncTexlFunction2
         private readonly WeakReference<IService> _parentService;
         private readonly string _actionName;
         private readonly bool _numberIsFloat;
+        private readonly string _pageLink;
+        private readonly bool _isDeprecated;
```
MikeStall, line 51: >_isDeprecated [](http://example.com/codeflow?start=30&length=13)

nt: can we comment this one: 
- it's still callable, 
- but we issue a warning if used 
- hidden from  intellisense

This also helps explain the different with "supported":
- not callable
- shown in intellisense with error

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
LucGenetier: Could be a RecordValue

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
LucGenetier: No. It means it implements x-ms-pageable and it must return a RecordValue

MikeStall:
> {

please shrink these files. < 100 lines. 

We're just testing paging logic, so 64k lines or 100 lines gives us the same test benefit. 



---
Refers to: src/tests/Microsoft.PowerFx.Connectors.Tests/Responses/AzureBlobStorage_Paging_Response1.json:1 in bccd211. [](commit_id = bccd211826bba5d9843ff817cc466e13b6fb56d7, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Public/Values/PagedRecordValue.cs
```diff
@@ -0,0 +1,137 @@
...
+                {
+                    if (rowCount > MaxRows)
+                    {
+                        break;
+                    }
+
+                    // Get the next page and if an error occurs, let's stop
+                    if (ptv.GetNextRecord() is not RecordValue nextRecord)
+                    {
+                        break;
```
MikeStall, line 95: >break; [](http://example.com/codeflow?start=24&length=6)

yield error here?
Concerned we may be swallowing errors

src/libraries/Microsoft.PowerFx.Core/Public/Values/PagedRecordValue.cs
```diff
@@ -0,0 +1,137 @@
...
+
+using System;
+using System.Collections.Generic;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Core.Public.Values
+{
+    // PagedRecordValue is a RecordValue that can be paged through.
+    // It is composed of the "nextLink" string and a table with the results
+    internal class PagedRecordValue : RecordValue
```
MikeStall, line 12: >PagedRecordValue [](http://example.com/codeflow?start=19&length=16)

These shouldn't be in Core - can you move to Connectors?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/IAsyncTexlFunction2.cs
MikeStall: actually - why does this even need to be on an interface at all? If the only caller is already in ServiceFunction, it doesn't need to be on an interface.

src/libraries/Microsoft.PowerFx.Core/Public/Values/PagedRecordValue.cs
```diff
@@ -0,0 +1,137 @@
...
+                foreach (DValue<RecordValue> currentRow in CurrentTable.Rows)
+                {
+                    rowCount++;
+                    if (rowCount > MaxRows)
+                    {
+                        break;
+                    }
+
+                    yield return currentRow;
+                }
```
MikeStall, line 81: can we move this into the loop and share with same code on line 116?

src/libraries/Microsoft.PowerFx.Core/Public/Values/PagedRecordValue.cs
```diff
@@ -0,0 +1,137 @@
...
+                            {
+                                break;
+                            }
+
+                            yield return currentRow;
+                        }
+                    }
+                    else
+                    {
+                        // This is dead code but we need to set ptv to null to avoid infinite loop
```
MikeStall, line 130: >need to set ptv to null to avoid infinite loo [](http://example.com/codeflow?start=52&length=45)

or just return :)

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
```diff
@@ -425,26 +440,59 @@ public bool TryGetParamDefaultValue(string paramName, out Tuple<string, DType> d
...
+                return FormulaValue.NewBlank();
             }
 
+            FormulaValue result = _invoker.InvokeAsync(nextLink, cancellationToken).ConfigureAwait(false).GetAwaiter().GetResult();
+
+            if (result is RecordValue rv && IsPageable && rv.TryGetField(FormulaType.String, _pageLink, out FormulaValue pageLink))
+            {
+                string nextLink2 = (pageLink as StringValue)?.Value;
+                result = new PagedRecordValue(rv, () => GetNextPage(nextLink2, cancellationToken), _maxRows);
+            }
```
MikeStall, line 475: this is repeated 3 times, can we simplify?
And the one on 494 is missing a null check

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
```diff
@@ -425,26 +440,59 @@ public bool TryGetParamDefaultValue(string paramName, out Tuple<string, DType> d
...
+
+            if (result is ErrorValue ev && (er = ev.Errors.FirstOrDefault(e => e.Kind == ErrorKind.Network)) != null)
+            {
+                result = FormulaValue.NewError(new ExpressionError() { Kind = er.Kind, Severity = er.Severity, Message = $"{Namespace.ToDottedSyntax()}.{Name} failed: {er.Message}" }, ev.Type);
+            }
+
+            if (result is RecordValue rv && IsPageable && rv.TryGetField(FormulaType.String, _pageLink, out FormulaValue pageLink))
+            {
+                string nextLink = (pageLink as StringValue)?.Value;
+                result = new PagedRecordValue(rv, () => GetNextPage(nextLink, cancellationToken), _maxRows);
```
MikeStall, line 456: >cancellationToken [](http://example.com/codeflow?start=78&length=17)

Don't close over it; instead pass in? 
 (cancellationToken) => GetNextPage(nextLink, cancellationToken)

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
```diff
@@ -425,26 +440,59 @@ public bool TryGetParamDefaultValue(string paramName, out Tuple<string, DType> d
...
+
+        private FormulaValue GetNextPage(string nextLink, CancellationToken cancellationToken)
+        {
+            if (string.IsNullOrEmpty(nextLink))
             {
-                throw new InvalidOperationException($"Function {Name} can't be invoked.");
+                return FormulaValue.NewBlank();
             }
 
+            FormulaValue result = _invoker.InvokeAsync(nextLink, cancellationToken).ConfigureAwait(false).GetAwaiter().GetResult();
```
MikeStall, line 469: >.GetResult(); [](http://example.com/codeflow?start=118&length=13)

Let's push this higher and make the method async.

src/libraries/Microsoft.PowerFx.Core/Public/Values/PagedRecordValue.cs
```diff
@@ -0,0 +1,137 @@
...
+using System.Collections.Generic;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Core.Public.Values
+{
+    // PagedRecordValue is a RecordValue that can be paged through.
+    // It is composed of the "nextLink" string and a table with the results
+    internal class PagedRecordValue : RecordValue
+    {
+        public RecordValue CurrentPage { get; private set; }
```
MikeStall, line 14: >private set; } [](http://example.com/codeflow?start=46&length=14)

can you make private readonly fields?

'private set' is still mutable

src/libraries/Microsoft.PowerFx.Core/Public/Values/PagedRecordValue.cs
```diff
@@ -0,0 +1,137 @@
...
+            GetNextRecord = getNextRecord;
+            MaxRows = maxRows;
+        }
+
+        protected internal override bool TryGetField(FormulaType fieldType, string fieldName, out FormulaValue result)
+        {
+            bool b = CurrentPage.TryGetField(fieldType, fieldName, out result);
+
+            // If we retrieve the table (via "value" field), we need to wrap it in a PagedTableValue
+            if (fieldType is TableType && result is TableValue tv)
```
MikeStall, line 34: >if (fieldType is TableType && result is TableValue tv) [](http://example.com/codeflow?start=12&length=54)

can we use a more specific check here? like check on exact field name? Is it always called "value"?

src/libraries/Microsoft.PowerFx.Core/Public/Values/PagedRecordValue.cs
```diff
@@ -0,0 +1,137 @@
...
+            {
+                int rowCount = 0;
+
+                // Enumerate rows of the current page
+                foreach (DValue<RecordValue> currentRow in CurrentTable.Rows)
+                {
+                    rowCount++;
+                    if (rowCount > MaxRows)
+                    {
+                        break;
```
MikeStall, line 77: >break [](http://example.com/codeflow?start=24&length=5)

return?

src/libraries/Microsoft.PowerFx.Core/Public/Values/PagedRecordValue.cs
```diff
@@ -0,0 +1,137 @@
...
+                    }
+
+                    yield return currentRow;
+                }
+
+                PagedTableValue ptv = this;
```
MikeStall, line 83: >PagedTableValue ptv = this; [](http://example.com/codeflow?start=16&length=27)

consider making this method static and passing in 'this'. 

It's too easy for the following code to refer to a field on this rather than field on ptv.

src/libraries/Microsoft.PowerFx.Core/Public/Values/PagedRecordValue.cs
```diff
@@ -0,0 +1,137 @@
...
+
+    internal class PagedTableValue : TableValue
+    {
+        public TableValue CurrentTable { get; private set; }
+
+        public Func<FormulaValue> GetNextRecord { get; private set; }
+
+        public FormulaType FieldType { get; private set; }
+
+        public string FieldName { get; private set; }
```
MikeStall, line 51: Can you update name to be clear which field these refer to?   The next link's, or the value? 

If it's the value, can FieldType be TableType?

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
MikeStall: After looking at the rest of the PR, 
could we do something even more strongly typed, like 
class Segment  { 
  TableValue value;
  string nextLink;
}

Essentially, consolidate all the "decoding" from RecordValue in one place. 
That may reduce some of the duplication and simplify the logic in Page*Value classes.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5009,6 +5009,14 @@
     <value>This feature is deprecated and is no longer supported.</value>
     <comment>An error message for deprecated features.</comment>
   </data>
+  <data name="WarnDeprecatedFunction" xml:space="preserve">
+    <value>In namespace {1}, function {0} is deprecated.</value>
```
MikeStall, line 5013: >namespace {1}, function {0}  [](http://example.com/codeflow?start=14&length=28)

nit:
I think we can just say  "{0}  is deprecated."   where 0 includes namespace - if present. 
rather than call out namespace separately .

MikeStall:
>         public IAsyncTexlFunction2 _invoker;

can this be init property?  (so that it's immutable) 

And maybe just ScopedHttpFunctionInvoker instead of the interface?



---
Refers to: src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs:439 in bccd211. [](commit_id = bccd211826bba5d9843ff817cc466e13b6fb56d7, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Public/Values/PagedRecordValue.cs
LucGenetier: FieldType is a TableType but not necessarily = TableType

src/libraries/Microsoft.PowerFx.Interpreter/Functions/IAsyncTexlFunction2.cs
LucGenetier: Resolved offline
Removed the API from interface

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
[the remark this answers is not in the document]
LucGenetier: I'll remove internal on this API and make TryGetFieldAsync internal instead

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
LucGenetier: I modifed & simplified the code to have Segment=PagedRecordValue

src/libraries/Microsoft.PowerFx.Core/Public/Values/PagedRecordValue.cs
LucGenetier: I simplified the code and don't think we need this

src/libraries/Microsoft.PowerFx.Core/Public/Values/PagedRecordValue.cs
LucGenetier: Refactored

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.ConfigExtensions.AddService(Microsoft.PowerFx.PowerFxConfig,System.String,Microsoft.OpenApi.Models.OpenApiDocument,System.Net.Http.HttpMessageInvoker,Microsoft.PowerFx.Connectors.ICachingHttpClient,System.Boolean)<br />❌CP0002 M:Microsoft.PowerFx.ConfigExtensions.AddService(Microsoft.PowerFx.PowerFxConfig,System.String,Microsoft.PowerFx.Connectors.ConnectorFunction,System.Net.Http.HttpMessageInvoker,Microsoft.PowerFx.Connectors.ICachingHttpClient)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ConnectorFunction.#ctor(Microsoft.OpenApi.Models.OpenApiOperation,System.Boolean,System.String,System.String,System.String,System.Net.Http.HttpMethod,System.String,System.Net.Http.HttpClient,System.Boolean,System.Boolean)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.OpenApiParser.GetFunctions(Microsoft.OpenApi.Models.OpenApiDocument,System.Net.Http.HttpClient,System.Boolean,System.Boolean)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.ConfigExtensions.AddService(Microsoft.PowerFx.PowerFxConfig,System.String,Microsoft.OpenApi.Models.OpenApiDocument,System.Net.Http.HttpMessageInvoker,Microsoft.PowerFx.Connectors.ICachingHttpClient,System.Boolean)<br />❌CP0002 M:Microsoft.PowerFx.ConfigExtensions.AddService(Microsoft.PowerFx.PowerFxConfig,System.String,Microsoft.PowerFx.Connectors.ConnectorFunction,System.Net.Http.HttpMessageInvoker,Microsoft.PowerFx.Connectors.ICachingHttpClient)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ConnectorFunction.#ctor(Microsoft.OpenApi.Models.OpenApiOperation,System.Boolean,System.String,System.String,System.String,System.Net.Http.HttpMethod,System.String,System.Net.Http.HttpClient,System.Boolean,System.Boolean)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.OpenApiParser.GetFunctions(Microsoft.OpenApi.Models.OpenApiDocument,System.Net.Http.HttpClient,System.Boolean,System.Boolean)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.ConfigExtensions.AddService(Microsoft.PowerFx.PowerFxConfig,System.String,Microsoft.OpenApi.Models.OpenApiDocument,System.Net.Http.HttpMessageInvoker,Microsoft.PowerFx.Connectors.ICachingHttpClient,System.Boolean)<br />❌CP0002 M:Microsoft.PowerFx.ConfigExtensions.AddService(Microsoft.PowerFx.PowerFxConfig,System.String,Microsoft.PowerFx.Connectors.ConnectorFunction,System.Net.Http.HttpMessageInvoker,Microsoft.PowerFx.Connectors.ICachingHttpClient)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ConnectorFunction.#ctor(Microsoft.OpenApi.Models.OpenApiOperation,System.Boolean,System.String,System.String,System.String,System.Net.Http.HttpMethod,System.String,System.Net.Http.HttpClient,System.Boolean,System.Boolean)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.OpenApiParser.GetFunctions(Microsoft.OpenApi.Models.OpenApiDocument,System.Net.Http.HttpClient,System.Boolean,System.Boolean)

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
```diff
@@ -46,18 +46,40 @@ internal sealed class ServiceFunction : BuiltinFunction, IAsyncTexlFunction2
...
+        // Those functions will not appear in Intellisense but can be called
+        // Check will return a warning when used
+        private readonly bool _isDeprecated;
+
+        // Functions that we do not support for any reason (see _notSupportedReason)
+        // Deprecated functions are also marked as not supported but will work fine.
+        // We return an error when used
+        private readonly bool _isSupported;
+        private readonly string _notSupportedReason;
+        private readonly int _maxRows;
```
MikeStall, line 63: >_maxRows [](http://example.com/codeflow?start=29&length=8)

can you comment what this field means?

For pageable functions, this limits the amount of paging.

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
```diff
@@ -198,9 +198,9 @@ internal async Task<FormulaValue> GetFieldAsync(FormulaType fieldType, string fi
         /// <param name="fieldName">Name of the field.</param>        
         /// <param name="result"></param>
         /// <returns>true if field is present, else false.</returns>
-        protected abstract bool TryGetField(FormulaType fieldType, string fieldName, out FormulaValue result);
-        
-        protected virtual Task<(bool Result, FormulaValue Value)> TryGetFieldAsync(FormulaType fieldType, string fieldName, CancellationToken cancellationToken)
+        protected abstract bool TryGetField(FormulaType fieldType, string fieldName, out FormulaValue result);
+
+        protected internal virtual Task<(bool Result, FormulaValue Value)> TryGetFieldAsync(FormulaType fieldType, string fieldName, CancellationToken cancellationToken)
```
MikeStall, line 203: >internal [](http://example.com/codeflow?start=18&length=8)

do we need this? Can't they just call the public method (GetFieldAsync)?

src/tests/Microsoft.PowerFx.Connectors.Tests/Responses/AzureBlobStorage_Paging_Response1.json
```diff
@@ -0,0 +1,71 @@
...
+      "LastModified": "2023-07-31T13:28:59Z",
+      "Size": 0,
+      "MediaType": "text/plain",
+      "IsFolder": false,
+      "ETag": "\"0x8DB91CA1915457C\"",
+      "FileLocator": "JTJmY29udGFpbmVyJTJmMTAwMC50eHQ=",
+      "LastModifiedBy": null
+    }    
+  ],
+  "nextLink": "https://tip1-shared-002.azure-apim.net/apim/azureblob/d95489a91a5846f4b2c095307d86edd6/v2/datasets/pfxdevstgaccount1/foldersV2/container?nextPageMarker=2!72!MDAwMDA4ITU0OTgudHh0ITAwMDAyOCE5OTk5LTEyLTMxVDIzOjU5OjU5Ljk5OTk5OTlaIQ--&useFlatListing=False",
```
MikeStall, line 69: just double checking - this high-entropy url doesn't contain any secrets?

src/libraries/Microsoft.PowerFx.Connectors/PagesRecordValueAndTable.cs
```diff
@@ -0,0 +1,131 @@
...
+            if (fieldName == TableFieldName)
+            {
+                result = new PagedTableValue(this);
+            }
+
+            return b;
+        }
+    }
+
+    internal class PagedTableValue : TableValue
```
MikeStall, line 53: >PagedTableValue [](http://example.com/codeflow?start=19&length=15)

nit: 
PagesREcordValueAndTable.cs is  a unusual naming convention. 

We could just put PagedTableValue as a private nested class in PagedRecordValue (since that's the only place that should create it) and then call this file PagedRecordValue.cs

src/libraries/Microsoft.PowerFx.Connectors/PagesRecordValueAndTable.cs
```diff
@@ -0,0 +1,131 @@
...
+                    {
+                        yield return DValue<RecordValue>.Of(ErrorValue.NewError(new ExpressionError() { Kind = ErrorKind.Custom, Message = $"Invalid next page, type = {nextPage.GetType().Name}" }));
+                        yield break;
+                    }
+                    
+                    (bool b, FormulaValue nextTableValue) = nextRecord.TryGetFieldAsync(currentRecord.CurrentTable.Type, currentRecord.TableFieldName, CancellationToken).ConfigureAwait(false).GetAwaiter().GetResult();
+
+                    if (!b || nextTableValue is not TableValue newTable)
+                    {
+                        yield break;
```
MikeStall, line 112: >yield break; [](http://example.com/codeflow?start=24&length=12)

error?

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
LucGenetier: you're correct, that's the name of the field
I'll fix the comment

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
LucGenetier: No, this generates an error (like for TryGetField)

LucGenetier: Will use GetField instead

src/tests/Microsoft.PowerFx.Connectors.Tests/Responses/AzureBlobStorage_Paging_Response1.json
LucGenetier: nope, it's not a SAS token

MikeStall (approved):
:shipit:

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.ConfigExtensions.AddService(Microsoft.PowerFx.PowerFxConfig,System.String,Microsoft.OpenApi.Models.OpenApiDocument,System.Net.Http.HttpMessageInvoker,Microsoft.PowerFx.Connectors.ICachingHttpClient,System.Boolean)<br />❌CP0002 M:Microsoft.PowerFx.ConfigExtensions.AddService(Microsoft.PowerFx.PowerFxConfig,System.String,Microsoft.PowerFx.Connectors.ConnectorFunction,System.Net.Http.HttpMessageInvoker,Microsoft.PowerFx.Connectors.ICachingHttpClient)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ConnectorFunction.#ctor(Microsoft.OpenApi.Models.OpenApiOperation,System.Boolean,System.String,System.String,System.String,System.Net.Http.HttpMethod,System.String,System.Net.Http.HttpClient,System.Boolean,System.Boolean)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.OpenApiParser.GetFunctions(Microsoft.OpenApi.Models.OpenApiDocument,System.Net.Http.HttpClient,System.Boolean,System.Boolean)

PR #1714, opened 2023-08-01
Ensure that untyped blank is interpreted as scalar

### LINKED ISSUE #1713: <<Blank() in "hello">> should work
The Blank value should be interpreted as a scalar string, not as an aggregate type.

CarlosFigueiraMSFT:
We already fixed it in most places, but found one more. This fixes #1713


LucGenetier:
✅ No public API change.

MikeStall (approved):

Issue #1706, opened 2023-07-28
Text function: Block [$-en-US]


gregli-msft:
There are two problems with this notation.

First, we have the usage of this syntax backwards from Excel.  Excel uses this to specify the *output* locale, which we use the third argument to Text for.  We use it to mean the *input* locale.  I believe long term we probably want a different syntax for this that does not conflict with Excel, such as `[@-en-US]`.

Second, this notation changes more than just thousands and decimal separators, it also impacts the placeholder letters used for day, month, and year.  I might say that we could get away with just using en-US placeholders, but the specification for something like `[@-fr-FR` is right there, it seems like an obvious bug to ignore it.

Consider this spreadsheet in English:
![image](https://github.com/microsoft/Power-Fx/assets/15149773/913d9a69-5130-4ead-a28d-e985dc56e457)

Here is how it looks in French:
![image](https://github.com/microsoft/Power-Fx/assets/15149773/b04d8ce4-dc8c-440f-92ac-89f816551bf5)

Literally the same workbook, changing behavior depending on the viewer's locale.  This is exactly what the `[@-en-US]` specifier is designed to address.  

But for now, we should just block `[$-en-US]` and not introduce `[@-en-US]`.  We'll have to resolve this when we get to Power Fx 1.0 in Canvas.  We can live with just en-US for the time being for these format strings.  The maker can still specify the output locale with the third argument to Text, they just need to specify the format string itself in English (mm, dd, yy, command, and dot) until we get this done.

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/1716

Issue #1705, opened 2023-07-28
Text function: Block unsupported placeholders


gregli-msft:
Excel supports many more format placeholders than we do.  Some of these would clearly add value to Power Fx, such as `?` and `/` but we aren't in a position to do all this work now.  Others we may want to add down the road, and we'd like to keep our options open.

Until we are ready to implement all of these, we should block the following placeholders:
1. `?` - Whitespace padded digit (similar to `0` which zoer pads).
2. `#/`, `0/`, and `?/` - Fraction notation.  We can't outright block all `/` because it is common in `mm/dd/yy`, but we can block it if it is used in a numeric context.
3. `[`...`]` - Used in Excel for many scenarios, including conditional sections and elapsed time.  We are going to block them all for now and bring them back when needed.  This means we will be blocking [$-en-US], the subject of [#1706 ](https://github.com/microsoft/Power-Fx/issues/1706).
6. A third `;` - More than three sections.  Excel supports up to four, with the last being what to display if the input is text.  As we are strongly typed this may not make any sense long term.
4. `_` - Whitespace padding.  Sucks up the following character, we need to decide what to do long term, if anything.
5. `*` - Alignment.  Again, sucks up characters around it, we need to decide what to do long term, if anything.
6. `:` when used next to number placeholders - Time notation.  As with `/`, we can't outright block it because it is popular for `hh:mm:ss`.  We'll allow only if adjacent to a time value placeholder (allowing for whitespace).  Excel's rules on this aren't very clear.
7. `@` and `General` (case insensitive) - General text placeholder.  These things are just weird, I haven't deduced how exactly they work.  They seem to do similar things, but act on numbers and strings in opposite ways.  For now, block.

We also need to check the placeholders that C# supports that we do not, and pass these through as regular characters.  At present, we are interpreting permille and single ticks when we should not be.  For example in src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterExpressionTestCases/TextOverride.txt there is `Text(100, "#%‰'eE:")` which is waiting on this issue to be resolved.

Tests are available here: https://github.com/microsoft/Power-Fx/blob/gregli/text-excel/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_ExcelCompat.txt

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/1716

Issue #1704, opened 2023-07-28
Text function: Add consistent escaping to format string


gregli-msft:
Escaping in the Text format string is essential.  As Excel and Power Fx pluck more characters for placeholders, it is critical that makers have a way to reliably insert the character that they want.

Excel handles escaping of placeholder characters in two ways:
- Backslash before each character
- Double quoted string within the format string
It does not support single ticks for escaping, the ticks are passed through.

![image](https://github.com/microsoft/Power-Fx/assets/15149773/6410b275-239a-41b7-942d-57091036c4ad)

Now the C# interpreter:
```
1> Text( 123.456, "mmm ddd yyy" )
"May Wed 1900"

2> Text( 123.456, """mmm"" ddd yyy" )
"mmm Wed 1900"

3> Text( 123.456, "\m\m\m ddd yyy" )
"\5\5\5 Wed 1900"

4> Text( 123.456, "'mmm' ddd yyy" )
"'May' Wed 1900"

5> Text( 123.456, "#.##" )
"123.46"

6> Text( 123.456, "#.""##""" )
"123##"

7> Text( 123.456, "#.\#\#" )
"123##"

8> Text( 123.456, "#.'##'" )
"123##"

9> Text( 123.456, "#.##'" )
"123.46"
```
Lines 3, 8, and 9 are incorrect, and lines 6-8 are missing the decimal point.  Canvas gets many of these wrong in other interesting ways.

Tests for these scenarios can be found in [Power-Fx/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_ExcelCompat.txt at gregli/text-excel · microsoft/Power-Fx (github.com)](https://github.com/microsoft/Power-Fx/blob/gregli/text-excel/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_ExcelCompat.txt)

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/1716

PR #1716, opened 2023-08-02
Text: Add consistent escaping and block unsupported characters

nguhoa:
Fix 3 issues: https://github.com/microsoft/Power-Fx/issues/1704, https://github.com/microsoft/Power-Fx/issues/1705, https://github.com/microsoft/Power-Fx/issues/1706

1. Update text format utils to have consistent escaping of format string with Excel. 
2. This update logic includes backsplash, double quote, single quote, e, decimal point, group separator point. 
3. Blocked locale in format.
4. Block unsupported characters.
5. Make it work the same with Excel for permile, single tick and others.
6. Add handling for scaling factor.
7. Add around 420 tests (mostly come from https://github.com/microsoft/Power-Fx/blob/gregli/text-excel/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_ExcelCompat.txt - some other tests were added to cover the missing cases)


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_Format.txt
```diff
@@ -43,12 +43,6 @@
 >> Text(0, "short day") 
 "00ort 30a99"
 
->> Text(DateTime(1983, 6, 3, 2, 53, 9, 0), "dddd, mmm "" dd, yyy") 
-"Friday, Jun "" 03, 1983"
-
->> Text(30470, "DDDD, MMM "" DD, YYY") 
-"Friday, Jun "" 03, 1983"
-
 >> Text(DateTime(1983, 6, 3, 2, 53, 9, 0), "dddd, mmm """" dd, yyy")
```
CarlosFigueiraMSFT, line 46: I see that those tests have been moved to 'Text_Format_PowerFxV1Compat.txt' - it would be good to have them in 'Text_Format_PowerFxV1CompatDisabled.txt' with the existing result.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_Format.txt
```diff
@@ -767,4 +740,548 @@ Error({Kind:ErrorKind.BadLanguageCode})
 Error({Kind:ErrorKind.BadLanguageCode})
 
 >> Text(1234.5678, "[$-fr-FR # ##]0,00", "vi-VI")
-Error({Kind:ErrorKind.BadLanguageCode})
\ No newline at end of file
+Error({Kind:ErrorKind.BadLanguageCode})
+﻿// Text function consistency with Excel
```
CarlosFigueiraMSFT, line 744: [minor] Can you add an empty line before this new section?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_Format.txt
nguhoa: Add more test to disabled file

CarlosFigueiraMSFT:
>             }

This is now dead code - we should remove, and add it back when we support it

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs:94 in 5ab9272. [](commit_id = 5ab92726c728b36d44ee4a2587fb2c3d44cd4f7a, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -85,21 +97,154 @@ public static bool IsValidFormatArg(string formatString, CultureInfo formatCultu
...
-                textFormatArgs.HasNumericFmt = formatWithoutZeroSubseconds.IndexOfAny(new char[] { '0', '#' }) >= 0;
+                return false;
             }
 
-            if (textFormatArgs.HasDateTimeFmt && textFormatArgs.HasNumericFmt)
+            for (int i = 0; i < formatStr.Length; i++)
             {
-                return false;
+                // Do not allow format string start with '/'
+                if (i == 0 && formatStr[i] == '/')
```
CarlosFigueiraMSFT, line 115: >i == 0 && formatStr[i] == '/' [](http://example.com/codeflow?start=20&length=29)

Can be moved outside the `for` so that the check isn't performed for every character in the format string

nguhoa:
>             }

should i remove it to private function so we have record  and call it later or should I just delete it?

---
In reply to: [1664702289](https://github.com/microsoft/Power-Fx/pull/1716#issuecomment-1664702289) [](http://example.com/codeflow?ancestors=1664702289,1664702289)

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs:94 in ba59822. [](commit_id = ba59822cfcdb4232e09d626cfff5df8a51aa6bb0, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -85,21 +97,154 @@ public static bool IsValidFormatArg(string formatString, CultureInfo formatCultu
...
-            textFormatArgs.HasDateTimeFmt = textFormatArgs.FormatArg.IndexOfAny(new char[] { 'm', 'M', 'd', 'D', 'y', 'Y', 'h', 'H', 's', 'S', 'a', 'A', 'p', 'P' }) >= 0;
-            textFormatArgs.HasNumericFmt = textFormatArgs.FormatArg.IndexOfAny(new char[] { '0', '#' }) >= 0;
-            if (textFormatArgs.HasDateTimeFmt && textFormatArgs.HasNumericFmt)
+            bool hasNumericCharacters = false;
+            int decimalPointIndex = -1;
+            int sectionCount = 0;
+            var formatStr = textFormatArgs.FormatArg;
+            bool hasColonWithNum = false;
+
+            //Block "gen"/"general"
```
CarlosFigueiraMSFT, line 106: >"gen"/"general" [](http://example.com/codeflow?start=20&length=15)

They keyword in Excel is "general" - not "gen". We can search for that explicitly

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
nguhoa: updated here.

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
nguhoa: yes, update

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -85,21 +97,154 @@ public static bool IsValidFormatArg(string formatString, CultureInfo formatCultu
...
+                    // Block all unsupported characters
+                    return false;
+                }
+                else if (formatStr[i] == ':')
+                {
+                    hasColonWithNum = true;
+                }
+                else if (formatStr[i] == ';')
+                {
+                    // Does not allow number of section are more than 2.
```
CarlosFigueiraMSFT, line 160: >// Does not allow number of section are more than 2. [](http://example.com/codeflow?start=20&length=52)

Do we support more than 1 section? Would the results of `Text(12345.6789,"0.00000;#.######")` and `Text(12345.6789,"#.######;0.00000")` be different?

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -85,21 +97,154 @@ public static bool IsValidFormatArg(string formatString, CultureInfo formatCultu
...
+                }
+                else if (i == formatStr.Length - 1)
+                {
+                    // If format string ends with backsplash but no following character (both numeric and datetime format), format is invalid.
+                    if (formatStr[i] == '\\')
+                    {
+                        return false;
+                    }
+
+                    // If format string ends with e or e+ (not escaping character) then format is invalid (only applied for numeric format)
```
CarlosFigueiraMSFT, line 175: >only applied for numeric format [](http://example.com/codeflow?start=107&length=31)

If it is a date format I don't think we are doing the right thing. Excel treats "e" as a year specifier (`=TEXT(NOW(), "yyyy-mm-dd e")` returns "2023-08-03 2023", and if we don't handle it properly, we should block it as well (and add tests)

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -85,21 +97,154 @@ public static bool IsValidFormatArg(string formatString, CultureInfo formatCultu
...
+                    else
+                    {
+                        // Update \c to "c" to match with                       
+                        formatStr = formatStr.Insert(i + 2, "\"");
+                        formatStr = formatStr.Insert(i + 1, "\"");
+                        formatStr = formatStr.Remove(i, 1);
+                        i += 2;
+                    }
+                }
+                else if (formatStr[i] == '\"' && i < formatStr.Length - 1)
```
CarlosFigueiraMSFT, line 197: >formatStr[i] == '\"' && i < formatStr.Length -  [](http://example.com/codeflow?start=25&length=47)

If we see a `"` and this is the last character then we should fail. Does it? A test such as `Text(123.456, "0.0""")` should catch that.

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -85,21 +97,154 @@ public static bool IsValidFormatArg(string formatString, CultureInfo formatCultu
...
+                        return false;
+                    }
+                }
+
+                if (textFormatArgs.HasDateTimeFmt && textFormatArgs.HasNumericFmt)
+                {
+                    // Check if the date time format contains '0's after the seconds specifier, which
+                    // is used for fractional seconds - in which case it is valid
+                    var formatWithoutZeroSubseconds = _formatWithoutZeroSubsecondsRegex.Replace(formatStr, m => m.Groups[1].Success ? string.Empty : m.Groups[1].Value);
+                    textFormatArgs.HasNumericFmt = formatWithoutZeroSubseconds.IndexOfAny(_numericCharacters.ToArray()) >= 0;
```
CarlosFigueiraMSFT, line 214: >_numericCharacters.ToArray() [](http://example.com/codeflow?start=90&length=28)

No need to call `ToArray()` on every iteration

nguhoa: i will move this out of the loop

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
nguhoa: Yes, we support upto 3 sections. If only 2 section, the first one will be used for zero (same with excel) as:
> Text(12345.6789,"0.00000;#.######")
"12345.67890"

> Text(12345.6789,"#.######;0.00000")
"12345.6789"

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
nguhoa: Add this check and add tests.

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
nguhoa: Yes, block e here and add tests.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_Format_PowerFxV1CompatDisabled.txt
```diff
@@ -8,18 +8,586 @@ Error({Kind:ErrorKind.InvalidArgument})
...
+>> Text(1234567.1234567, "0.00e")
+Error({Kind:ErrorKind.InvalidArgument})
+
+>> Text(0, "longdate")
+Error({Kind:ErrorKind.InvalidArgument})
+
+>> Text(0, "shortdate")
+Error({Kind:ErrorKind.InvalidArgument})
+
+>> Text(DateTime(1970, 1, 1, 14, 32, 25, 678), "longdatetime")
```
CarlosFigueiraMSFT, line 40: Should we change it the behavior with Power Fx V1 disabled?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_Format.txt
```diff
@@ -639,132 +594,963 @@ Error({Kind:ErrorKind.InvalidArgument})
...
 
->> Text(1234.5678, "[$-en-US]", "en-US")
-"1234.5678"
+// Escape with backslash.
+>> Text(1234567.1234567, "\#,###.###")
+"#,1234567.123"
 
->> Text(1234.5678, "[$-fr-FR]", "fr-FR")
-"1234,5678"
+>> Text(1234567.1234567, "\\#,###.###")
```
CarlosFigueiraMSFT, line 710: >Text(1234567.1234567, "\\#,###.###") [](http://example.com/codeflow?start=3&length=36)

Lots of those new tests are not supported in Power Apps today (which is what uses PowerFxV1CompatDisabled). Can we add them to a different file, so it will be easier to exclude them in that platform until the support is added?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_Format_PowerFxV1CompatDisabled.txt
nguhoa: the blocking unsupported character like this is happened at runtime and at runtime i don't think we check PowerFxV1 flag. Do we want to check flag for all blocking and only do it when flag enable?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_Format.txt
nguhoa: move to new test file.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1722, opened 2023-08-03
Provide a delegation warning SortByColumns in an unsupported data source

ian-legler:
Previously, a complicated enough data source would not report a delegation warning.  Provide a warning in that scenario


LucGenetier:
✅ No public API change.

Issue #1493, opened 2023-05-09
expose property keys for IUntypedObject


xieofxie:
So it could be used together with `TryGetProperty`

MikeStall:
This is a tactical case from #1140.

MikeStall:
```
    interface IUntypedObject2
    {
        bool TryGetPropertyNames(out string[] names);
    }
```

PR #1727, opened 2023-08-04
Add GetProperties method to IUntypedObject

Robulane:
This PR solves this work item: https://github.com/microsoft/Power-Fx/issues/1493


src/libraries/Microsoft.PowerFx.Core/Public/Values/UntypedObjectValue.cs
```diff
@@ -43,7 +43,12 @@ public interface IUntypedObject
         // It is validated to be in a culture invariant standard format number, possibly with dot decimal separator and "E" exponent.
         // This method need not be implemetned if ExteranlType.UntypedNumber is not used.
         // GetDouble and GetDecimal can be called on an ExternalType.UntypedNumber.
-        string GetUntypedNumber();
+        string GetUntypedNumber();
+
+        /// <summary>
+        /// Returns the list of available properties if the current instance is an object.
+        /// </summary>
+        IEnumerable<string> GetProperties();
```
MikeStall, line 51: Some UO implementations may not be able to enumerate their properties - such as if they're lazy or exist over an open system.  Can you make this:

`bool TryGetProperties(out IEnumerable<string> names)`

This also avoids the NotImpl exceptions and lets us just return false.

CarlosFigueiraMSFT, line 51: Suggestion: `GetPropertyNames` - TryGetProperty gets the property value; this is returning the property names, not their values.

MikeStall (approved):
:shipit:

CarlosFigueiraMSFT (dismissed):
🚢

src/libraries/Microsoft.PowerFx.Core/Public/Values/UntypedObjectValue.cs
```diff
@@ -43,7 +43,13 @@ public interface IUntypedObject
         // It is validated to be in a culture invariant standard format number, possibly with dot decimal separator and "E" exponent.
         // This method need not be implemetned if ExteranlType.UntypedNumber is not used.
         // GetDouble and GetDecimal can be called on an ExternalType.UntypedNumber.
-        string GetUntypedNumber();
+        string GetUntypedNumber();
```
LucGenetier, line 46: >GetUntypedNumber [](http://example.com/codeflow?start=15&length=16)

I might name it GetRawValue() to avoid any confusion with *Untyped*Object

src/libraries/Microsoft.PowerFx.Json/Functions/JsonUntypedObject.cs
```diff
@@ -103,6 +103,12 @@ public bool TryGetProperty(string value, out IUntypedObject result)
             var res = _element.TryGetProperty(value, out var je);
             result = new JsonUntypedObject(je);
             return res;
+        }
+
+        public bool TryGetPropertyNames(out IEnumerable<string> result)
+        {
+            result = _element.EnumerateObject().Select(x => x.Name);
```
LucGenetier, line 110: >EnumerateObject [](http://example.com/codeflow?start=30&length=15)

While this is consistent with the rest of the code, we don't check if ValueKind is Object

LucGenetier (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Json/Functions/JsonUntypedObject.cs
Robulane: Added the check

src/libraries/Microsoft.PowerFx.Core/Public/Values/UntypedObjectValue.cs
Robulane: let's do it in a separate cleanup pr

LucGenetier (approved):
:shipit:

PR #1731, opened 2023-08-07
x-ms-pageable and deprecated function updates

LucGenetier:
- paging
  - implement paging
  - make functions using paging supported

- deprecated / unsupported function
  - deprecated: intellisense will return a warning in Check()
  - unsupported: intellisense will return an error in Check()


LucGenetier:
✅ No public API change.

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Connectors.ConnectorFunction.#ctor(Microsoft.OpenApi.Models.OpenApiOperation,System.Boolean,System.String,System.String,System.String,System.Net.Http.HttpMethod,System.String,System.Net.Http.HttpClient,System.Boolean,System.Boolean)

MikeStall (approved):
:shipit:

PR #1739, opened 2023-08-09
Support more connectors

LucGenetier:
- ignore x-ms-notification-content
- ignore x-ms-capabilities
- ignore x-ms-test-value
- ignore x-ms-url-encoding
- improve TestAllConnectors test
- do not throw an exception when the return type isn't supported but use IsSupported flag instead


LucGenetier:
✅ No public API change.

Robulane (approved):

.gitignore
```diff
@@ -357,3 +357,11 @@ tool/
...
 # ApiCompat suppression files
 suppress.*.xml
+
+# Connetor analysis reports
+Analysis.txt
+Report.json
+Report_*.json
+
+# global.json
+global.json
```
MikeStall, line 367: what file is this?

.gitignore
```diff
@@ -357,3 +357,11 @@ tool/
 
 # ApiCompat suppression files
 suppress.*.xml
+
+# Connetor analysis reports
+Analysis.txt
+Report.json
+Report_*.json
```
MikeStall, line 364: could we ignore by directory?

.gitignore
LucGenetier: [https://learn.microsoft.com/en-us/dotnet/core/tools/global-json](https://learn.microsoft.com/en-us/dotnet/core/tools/global-json)

It's set during build pipeline but I don't want it to be checked-in if I define it locally

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1742, opened 2023-08-10
Removed reference to binding in UserDefinedFunction

VamsiModem:


src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunctionLibrary.cs
```diff
@@ -0,0 +1,85 @@
...
+
+namespace Microsoft.PowerFx.Core.Functions
+{
+    internal class UserDefinedFunctionLibrary
+    {
+        private TexlFunctionSet _library;
+
+        private readonly Dictionary<string, TexlBinding> _bindings;
+
+        public Dictionary<string, List<TexlFunction>>.KeyCollection FunctionNames => _library.FunctionNames;
```
MikeStall, line 22: >string [](http://example.com/codeflow?start=26&length=6)

can you comment what 'string' is? 
function name? (with / without namespace?)

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -30,21 +30,33 @@ internal class UserDefinedFunction : TexlFunction
     {
         private readonly bool _isImperative;
         private readonly IEnumerable<UDFArg> _args;
-        private TexlBinding _binding;
+        private readonly UserDefinedFunctionLibrary _library;
 
-        public override bool IsAsync => _binding?.IsAsync(UdfBody) ?? false;
+        public override bool IsAsync
```
MikeStall, line 35: >IsAsync [](http://example.com/codeflow?start=29&length=7)

is there any danger in this value being able to mutate on us?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunctionLibrary.cs
MikeStall, line 22: >Dictionary<string, List<TexlFunction>>.KeyCollection [](http://example.com/codeflow?start=15&length=52)

Can this be IEnumerable<string> ?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunctionLibrary.cs
```diff
@@ -0,0 +1,85 @@
...
+            _bindings = new Dictionary<string, TexlBinding>();
+        }
+
+        public bool TryAdd(UserDefinedFunction function, TexlBinding binding)
+        {
+            Contracts.AssertValue(function);
+
+            if (_library.AnyWithName(function.Name))
+            {
+                return false;
```
MikeStall, line 38: > return false; [](http://example.com/codeflow?start=15&length=14)

do we need to replace when the binding is updated?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -30,21 +30,33 @@ internal class UserDefinedFunction : TexlFunction
     {
         private readonly bool _isImperative;
         private readonly IEnumerable<UDFArg> _args;
-        private TexlBinding _binding;
+        private readonly UserDefinedFunctionLibrary _library;
```
MikeStall, line 33: >UserDefinedFunctionLibrary [](http://example.com/codeflow?start=25&length=26)

this adds a level of indirection, but can we remove the dependency entirely?
Ie, if we're just doing intellisense at the callsite, we don't need this. 
And if we're doing codegen , we should already have it through a separate channel. 

(this begs the issue for IsAsync)

VamsiModem: Makes sense, the only reason we have reference binding/library is to derive the value for IsAsync as it is used in several places in the PAClient.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
VamsiModem: we are not deriving IsAsync from Binding or library anymore.

LucGenetier:
✅ No public API change.

PR #1746, opened 2023-08-11
Fix issue with Option Set Value inside Filter()

jas-valgotar:
Fixes https://github.com/microsoft/Power-Fx-Dataverse/issues/300


MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1753, opened 2023-08-14
Connector/Intellisense: do not suggest x-ms-visibility:internal functions

LucGenetier:


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/Shared/ServiceFunction.cs
```diff
@@ -55,6 +55,9 @@ internal sealed class ServiceFunction : BuiltinFunction, IAsyncTexlFunction2, IH
         // Check will return a warning when used
         private readonly bool _isDeprecated;
 
+        // For functions that have "x-ms-visibility" set to "internal" in swagger file.
+        private readonly bool _isInternal;
```
MikeStall, line 59: How is PowerApps handling this?

They must hit the same issue when importing Sharepoint swagger.

LucGenetier: They do

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

Issue #1760, opened 2023-08-16
Text: Only allow DateTimeFormat enum instead of string value of it


nguhoa:
In Text, we currently allow both DateTimeFormat enum (LongDateTime) and string value of it ("'longdatetime'").

> Text( Now(), "ZZZ 'longdatetime' ZZZ" )
"ZZZ 'long14ateti8e' ZZZ"

> Text( Now(), "'longdatetime'" )
"Monday, August 14, 2023 10:32:19 AM"

> Text( Now(), "'longdatetime' ZZZ" )
"'long14ateti8e' ZZZ"

> Text( Now(), DateTimeFormat.LongDateTime )
"Monday, August 14, 2023 10:56:29 AM"

We will only support DateTimeFormat enum and not allow string alternative of it.

gregli-msft:
Slightly more info...

Before we had strongly typed enums, enums and strings/numbers were interchangeable.  Which meant that the Format parameter to the Text function needed a way to support `DateTimeFormat.LongDate` as a string.  The solution was to have a special string, `'longdate'` in this case, that the Text function would key off of.  This is hacky and contradicts the normal handling of single ticks and other letters in the format string.  

For example, and see how the addition of another character changes the meaning completely:
```
1> Text( Now(), DateTimeFormat.LongDate )
"Wednesday, August 16, 2023"

2> Text( Now(), "'longdate'" )
"Wednesday, August 16, 2023"

3> Text( Now(), "'longdate'z" )
"'long16ate'z"
```

Times have changed and we now have strongly typed enums.  We should use it to differentiate from custom format strings.  As the [docs describe](https://learn.microsoft.com/en-us/power-platform/power-fx/reference/function-text), one should be able to use either the enum or a custom format string, but not both.  

In other words, example 2 above should look like example 3, without the extra `z`.

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/1761

PR #1761, opened 2023-08-16
Only allow DateTimeFormat enum instead of string value of it

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1760, https://github.com/microsoft/Power-Fx/issues/1768
Update logic to only allow DateTimeFormat enum and not treat the string value of it the same way with enum.
Update Test.cs to not convert from enum to string for DateTimeFormat.
Move existing test (with new behavior) and add new tests to Text_ExcelCompat.txt.


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
```diff
@@ -39,7 +39,8 @@ public void Canvas_Float(ExpressionTestCase testCase)
             var features = new Features()
             {
                 TableSyntaxDoesntWrapRecords = true,
-                ConsistentOneColumnTableResult = true
+                ConsistentOneColumnTableResult = true,
+                StronglyTypedBuiltinEnums = true,
```
lesaltzm, line 43: Canvas does not have StronglyTypedBuiltInEnums by default (or at all) at the moment, is this right?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
```diff
@@ -150,14 +150,9 @@ internal static void ValidateFormatArgs(string name, CheckTypesContext checkType
                 errors.EnsureError(DocumentErrorSeverity.Moderate, args[1], TexlStrings.ErrNotSupportedFormat_Func, name);
                 isValid = false;
             }
-            else
+            else if (!BuiltInEnums.DateTimeFormatEnum.FormulaType._type.Accepts(argTypes[1], exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: checkTypesContext.Features.PowerFxV1CompatibilityRules))
             {
-                if (BuiltInEnums.DateTimeFormatEnum.FormulaType._type.Accepts(argTypes[1], exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: checkTypesContext.Features.PowerFxV1CompatibilityRules))
-                {
-                    // Coerce enum values to string
-                    CollectionUtils.Add(ref nodeToCoercedTypeMap, args[1], DType.String);
```
lesaltzm, line 158: Is this coercion not needed? Double check running this against canvas, if possible.

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
nguhoa: yes, will remove this.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_ExcelCompat.txt
```diff
@@ -197,4 +197,236 @@ Table({Value:"08:03:05"},{Value:"00:00:00"})
...
+"'long15ateti8e' ZZZ"
+
+>> Text(DateTime(2023, 8, 15, 2, 53, 9, 0), "LongDateTime")
+"Long15ateTi8e"
+
+>> Text(DateTime(2023, 8, 15, 2, 53, 9, 0), "longdatetime")
+"long15ateti8e"
+
+>> Text(DateTime(2023, 8, 15, 2, 53, 9, 0), "'utc'")
+"'utc'"
```
MikeStall, line 260: new v1 behavior, Should be in:
#SETUP: PowerFxV1CompatibilityRules

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/NotYetReady/Text_Format.txt
```diff
@@ -1,98 +0,0 @@
-﻿>> Text(DateTime(1983, 6, 3, 2, 53, 9, 0), "utc") // utc1
-"1983-06-03T02:53:09.000Z"
-
->> Text(30470.519531251, "utc") // excelUtc1
-"1983-06-03T12:28:07.500Z"
```
MikeStall, line 5: old v0 behavior, 

new v1 behavior, Should be in:
#SETUP: disable:PowerFxV1CompatibilityRules

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Text.cs
nguhoa: I add flag StronglyTypedBuiltinEnums check and put back code so canvas can go with old flow.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/NotYetReady/Text_Format.txt
nguhoa: update to use StronglyTypedBuiltinEnums flag enable and disable.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_ExcelCompat.txt
nguhoa: move to test files.

LucGenetier:
✅ No public API change.

gregli-msft:
Addresses https://github.com/microsoft/Power-Fx/issues/1768

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_ExcelCompat.txt
```diff
@@ -197,4 +191,204 @@ Table({Value:"08:03:05"},{Value:"00:00:00"})
...
```
gregli-msft, line 234: Interestingly, 
``` 
> Text(DateTime(2023, 8, 15, 2, 53, 9, 0), "'utc'")
"'utc'"

> Text(123, "'utc'")
"utc"
```
What happened to the single ticks in the output of the second example?  This isn't a super important case or anything, I'm just worried there is something going on in there that we don't understand.

nguhoa: This PR does not ignore ' in numeric format  as we will change in PR: https://github.com/microsoft/Power-Fx/pull/1716. So when this PR and that PR merge, this test result will change to "'utc'".

MikeStall:
>         public bool HasDateTimeFmt { get; set; }

Is it possible to normalize and merge these flags? 

Ie, why does the consumer need to be aware?


---
Refers to: src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs:37 in 2755d07. [](commit_id = 2755d07fde2175e6dde5996e69138bd5c10bd3c2, deletion_comment = False)

LucGenetier:
✅ No public API change.

nguhoa:
>         public bool HasDateTimeFmt { get; set; }

remove enum flag. not all cases need this flag, it is only for specific case when datetimeenum is used. So i only update in librarytext.

---
In reply to: [1682678729](https://github.com/microsoft/Power-Fx/pull/1761#issuecomment-1682678729) [](http://example.com/codeflow?ancestors=1682678729)

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs:37 in 2755d07. [](commit_id = 2755d07fde2175e6dde5996e69138bd5c10bd3c2, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -487,50 +501,54 @@ public static bool TryText(FormattingInfo formatInfo, IRContext irContext, Formu
...
-                    }
-                    catch (FormatException)
-                    {
-                        return false;
-                    }
-
-                    break;
+                switch (format.ToLowerInvariant())
+                {
+                    case "'shortdatetime24'":
```
MikeStall, line 521: >case [](http://example.com/codeflow?start=20&length=4)

nit-  (perhaps in future PR) - these are "magical" constants. It would be nice if we can share them across the source. 

Ie, instead of encoding as ("hasEnum" bool + string version of the enum), can we actually have a resolved enum and avoid the sttring literal compare?

nguhoa: As discussed offline. I will leave this extra hasEnum here and will fix the enum type of datetimeformat enum for all source and refer then break this function to handle enum separately.

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #1763, opened 2023-08-16
Fix issue with "Text(5/2)" in French locale

LucGenetier:
Only occurs on Windows 10


MikeStall (approved):

LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #1772, opened 2023-08-17
Remove enhanced delegation external feature

jelopezf:
Enhanced delegation has been out and enabled by default for a long time now. Considering it's enabled in around 87% of Power Apps and there is confidence in the feature, we're getting rid of it as a setting and just making it the default behavior.


LucGenetier:
✅ No public API change.

VamsiModem (approved):

ian-legler (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1778, opened 2023-08-21
Sort suggestion correctly

jas-valgotar:


LucGenetier:
✅ No public API change.

jas-valgotar:
The sorting is eventually done on UFB's Monaco editor and will need to be taken care of there. Hence closing this PR.

PR #1781, opened 2023-08-22
Do not suggest deferred types in Collect(| Intellisense

jas-valgotar:
-While Suggesting Collect(| do not suggest deferred type, since we only want to suggest Table Types and even though Deferred Type is coercable to Table Types, it has no real meaning in this case.
-This also helps with not suggesting row scope symbols since they are also deferred types.


LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #1786, opened 2023-08-24
Add Contextual Suggestion for Binary Op

jas-valgotar:
-Adds contextual suggestions for Binary Op node.
Today 
`Num = |` doesn't suggest global number types.

This PR will suggest global types depending on left hand side of type for binary op nodes.


MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1791, opened 2023-08-25
Enable field suggestion for custom functions.

### LINKED ISSUE #1787: Intellisense: suggest fields when entering types
Collect, Patch know the schema for the record types and will show intellisense to suggest fields. 
Request to extend this to arbitrary record parameters, especially when working with custom functions. 

image

jas-valgotar:
Fixes #1787 
-If the custom function provided ParamType, now intellisense will suggest the corresponding columns. _including nested columns_, and columns inside a table type.
-Adds virtual `TryGetTypeForArgSuggestionAt(int argIndex, out DType type) ` to get Type intellisense will be used to match suggestions.
-Fixes Patch and Remove for suggestions.

NOTE: PA would need to override the `TryGetTypeForArgSuggestionAt()` for mutation functions definition in same way it's done here.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
@@ -156,6 +156,8 @@ namespace Microsoft.PowerFx.Core.Functions
         // Return true if the function uses an input's column names to inform Intellisense's suggestions.
         public virtual bool CanSuggestInputColumns => false;
 
+        public virtual bool IsInputColumnSuggestionArg0 => false;
+
         // Return true if the function expects a screen's context variables to be suggested within a record argument.
         public virtual bool CanSuggestContextVariables => false;
 
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
@@ -156,7 +156,26 @@ namespace Microsoft.PowerFx.Core.Functions
         // Return true if the function uses an input's column names to inform Intellisense's suggestions.
         public virtual bool CanSuggestInputColumns => false;
 
-        public virtual bool IsInputColumnSuggestionArg0 => false;
+        /// <summary>
+        /// If this returns false, the Intellisense will use Arg[0] type to suggest the type of the argument.
+        /// e.g. Collect(), Remove(), etc.
+        /// </summary>
+        /// <param name="argIndex"></param>
+        /// <param name="type"></param>
+        /// <returns></returns>
+        public virtual bool TryGetTypeForArgSuggestionAt(int argIndex, out DType type)
+        {
+            var maxArgIndex = (ParamTypes?.Count() ?? 0) - 1;
+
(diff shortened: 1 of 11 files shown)


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -166,7 +166,28 @@ internal abstract class TexlFunction : IFunction
...
+        public virtual bool CanSuggestInputColumns => false;
+
+        /// <summary>
+        /// If this returns false, the Intellisense will use Arg[0] type to suggest the type of the argument.
+        /// e.g. Collect(), Remove(), etc.
+        /// </summary>
+        /// <param name="argIndex"></param>
+        /// <param name="type"></param>
+        /// <returns></returns>
+        public virtual bool TryGetTypeForArgSuggestionAt(int argIndex, out DType type)
```
MikeStall, line 178: >TryGetTypeForArgSuggestionAt [](http://example.com/codeflow?start=28&length=28)

Does this apply to all functions? Or can this only be called in limited scenarios?

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
jas-valgotar: Yes, for a consistent experience, we should suggest global on all functions. not just Collect(| etc...

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -160,7 +160,28 @@ internal abstract class TexlFunction : IFunction
...
```
jas-valgotar, line 172: >public virtual bool TryGetTypeForArgSuggestionAt(int argIndex, out DType type) [](http://example.com/codeflow?start=8&length=78)

PA will need to override this in mutation functions.

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/RemoveFunction.cs
```diff
@@ -26,7 +26,9 @@ internal abstract class RemoveFunctionBase : BuiltinFunction
 
         public override bool RequiresDataSourceScope => true;
 
-        public override bool CanSuggestInputColumns => true;
+        public override bool CanSuggestInputColumns => true;
+
+        public override bool ManipulatesCollections => true;
```
lesaltzm, line 31: Manipulates Collections is a canvas concept? why is it added to this?

PR #1793, opened 2023-08-27
add clock service

MikeStall:
Add an IClockService, similar to existing IRandomService. 
Call clock service instead of DateTime.UtcNow directly. This abstracts our direct dependency on the machine. It also allows the host to override, which can be very useful for deterministic testing Today(), IsToday(), Now() - especially around UTC , datetime boundaries.


LucGenetier:
✅ No public API change.

nguhoa (approved):

PR #1797, opened 2023-08-30
Add meaningful errors to UDF when return type & arg type are missing

VamsiModem:


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4928,6 +4928,14 @@
     <value>The stated function return type does not match the return type of the function body.</value>
     <comment>This error message shows up when expected return type does not match with actual return type</comment>
   </data>
+  <data name="ErrUDF_MissingReturnType" xml:space="preserve">
+    <value>Missing function return type, for example the ":Number" in "f( x:Text ):Number = Value( x );"</value>
```
CarlosFigueiraMSFT, line 4932: >Text [](http://example.com/codeflow?start=76&length=4)

`Text`, `Number` and `Value` should not be translated in this example, so we should tag them as `{Locked=...} in the comments.

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -1720,14 +1729,18 @@ private bool EatTid(TokKind tid)
 
         // Returns the current token if it's of the given kind and moves to the next token.
         // If the token is not the right kind, reports an error, leaves the token, and returns null.
```
CarlosFigueiraMSFT, line 1731: >reports an error [](http://example.com/codeflow?start=47&length=16)

Please update the comment to account for the new argument. May be a good time to convert this comment to the `///` format.

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -1718,16 +1731,25 @@ private bool EatTid(TokKind tid)
...
 
-        // Returns the current token if it's of the given kind and moves to the next token.
-        // If the token is not the right kind, reports an error, leaves the token, and returns null.
-        private Token TokEat(TokKind tid)
+        /// <summary>
+        /// Returns the current token if it's of the given kind and moves to the next token.
+        /// If the token is not the right kind, reports an error, leaves the token, and returns null.
+        /// </summary>
+        /// <param name="tid">TokenKind.</param>
+        /// <param name="addError">Will not add token mis-match error when true.</param>
```
CarlosFigueiraMSFT, line 1739: >Will not add token mis-match error when true. [](http://example.com/codeflow?start=35&length=45)

The other way around - it will not add the error if the value is `false`.

CarlosFigueiraMSFT (approved):
🚢

LucGenetier:
✅ No public API change.

PR #1798, opened 2023-08-31
Add coercion support for UDFs

VamsiModem:
1. Adds coercion support for UDFs
2. UDFs are not bound on creation
3. Added IR tests for UDF body/param coercion


MikeStall:
Can you add a test to demonstrate what this change is doing?

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
```diff
@@ -65,15 +65,15 @@ public bool ProcessUserDefnitions(out UserDefinitionResult userDefinitionResult)
...
+            var functions = CreateUserDefinedFunctions(parseResult.UDFs, out var errors, shouldBindBody);
 
             errors.AddRange(parseResult.Errors ?? Enumerable.Empty<TexlError>());
             userDefinitionResult = new UserDefinitionResult(functions, errors, parseResult.NamedFormulas);
 
             return true;
         }
 
-        private IEnumerable<UserDefinedFunction> CreateUserDefinedFunctions(IEnumerable<UDF> uDFs, out List<TexlError> errors)
+        private IEnumerable<UserDefinedFunction> CreateUserDefinedFunctions(IEnumerable<UDF> uDFs, out List<TexlError> errors, bool shouldBindBody = false)
```
CarlosFigueiraMSFT, line 76: > = false [](http://example.com/codeflow?start=146&length=8)

This is a private method, and the only call is passing a value to this argument, so there's no need to define a default value for it.

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
```diff
@@ -48,14 +48,14 @@ public static ParseUserDefinitionResult Parse(string script, ParserOptions parse
...
-        public static bool ProcessUserDefinitions(string script, INameResolver globalNameResolver, IBinderGlue documentBinderGlue, BindingConfig bindingConfig, ParserOptions parserOptions, out UserDefinitionResult userDefinitionResult, Features features = null)
+        public static bool ProcessUserDefinitions(string script, INameResolver globalNameResolver, IBinderGlue documentBinderGlue, BindingConfig bindingConfig, ParserOptions parserOptions, out UserDefinitionResult userDefinitionResult, Features features = null, bool shouldBindBody = false)
         {
             var userDefinitions = new UserDefinitions(script, globalNameResolver, documentBinderGlue, bindingConfig, parserOptions, features);
 
             return userDefinitions.ProcessUserDefnitions(out userDefinitionResult);
         }
 
-        public bool ProcessUserDefnitions(out UserDefinitionResult userDefinitionResult)
+        public bool ProcessUserDefnitions(out UserDefinitionResult userDefinitionResult, bool shouldBindBody = false)
```
CarlosFigueiraMSFT, line 58: >ProcessUserDefnitions [](http://example.com/codeflow?start=20&length=21)

It was already wrong, but can you fix the typo in this method name?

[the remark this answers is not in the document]
VamsiModem: Optional params always came at the end.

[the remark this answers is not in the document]
CarlosFigueiraMSFT: But this doesn't have to be optional - see other comment. This is only called once, from inside this method, with the argument that *should have been passed*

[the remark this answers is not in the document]
VamsiModem: Yup, changed that!

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -98,9 +98,19 @@ public void CheckTypesOnDeclaration(CheckTypesContext context, DType actualBodyR
...
-            if (!ReturnType.Kind.Equals(actualBodyReturnType.Kind) && !actualBodyReturnType.CoercesTo(ReturnType, true, false, context.Features.PowerFxV1CompatibilityRules))
+            if (!ReturnType.Kind.Equals(actualBodyReturnType.Kind))
             {
-                errorContainer.EnsureError(DocumentErrorSeverity.Severe, UdfBody, TexlStrings.ErrUDF_ReturnTypeDoesNotMatch);
+                if (actualBodyReturnType.CoercesTo(ReturnType, true, false, context.Features.PowerFxV1CompatibilityRules))
+                {
+                    if (!_binding.CoercedToplevelType.IsValid)
+                    {
+                        _binding.SetCoercedToplevelType(ReturnType);
+                    }
```
Shpakh-MSFT, line 108: Can you please add some comments about this logic?

Shpakh-MSFT:
2. Add a param that would allow to skip binding on UDF declaration.

>> Is this needed for some other change?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -98,9 +98,19 @@ public void CheckTypesOnDeclaration(CheckTypesContext context, DType actualBodyR
             Contracts.AssertValue(actualBodyReturnType);
             Contracts.AssertValue(errorContainer);
 
-            if (!ReturnType.Kind.Equals(actualBodyReturnType.Kind) && !actualBodyReturnType.CoercesTo(ReturnType, true, false, context.Features.PowerFxV1CompatibilityRules))
+            if (!ReturnType.Kind.Equals(actualBodyReturnType.Kind))
             {
-                errorContainer.EnsureError(DocumentErrorSeverity.Severe, UdfBody, TexlStrings.ErrUDF_ReturnTypeDoesNotMatch);
+                if (actualBodyReturnType.CoercesTo(ReturnType, true, false, context.Features.PowerFxV1CompatibilityRules))
+                {
+                    if (!_binding.CoercedToplevelType.IsValid)
```
Shpakh-MSFT, line 105: >if (!_binding.CoercedToplevelType.IsValid) [](http://example.com/codeflow?start=20&length=42)

Why do you need this check?

Shpakh-MSFT:
>             var parserOptions = new ParserOptions()

Any more tests for coercion?

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/UserDefinedFunctionTests.cs:47 in fb95b3f. [](commit_id = fb95b3faf975f4e501c8848f9d006ef653114c77, deletion_comment = False)

CarlosFigueiraMSFT:
>         public void TestUDFNamedFormulaCounts(string script, int udfCount, int namedFormulaCount, bool expectErrors)

We should have some setup handler for the TXT tests that deal with UDFs, and have e2e tests for those. For example, if we define something like `Mul(x:Number, y:Number):Number = x*y;`, we can have the following tests:

    // "Simple" invocation (Sqrt forces number)
    >> Mul(Sqrt(9), Sqrt(16))
    12

    // Coercion from decimal (when enabled)
    >> Mul(9, 16)
    144

    // Coercion to argument types
    >> Mul("4", Date(1900, 1, 3))
    20

    // Coercion from result
    >> Concatenate("Result is ", Mul(5, 6))
    "Result is 30"

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/UserDefinedFunctionTests.cs:45 in 313981b. [](commit_id = 313981b94ca0350b339ad6b0e2bbddc82bfdadd3, deletion_comment = False)

VamsiModem:
> 2. Add a param that would allow to skip binding on UDF declaration.
> 
> > > Is this needed for some other change?

This is just an optimization, we don't have to bind on declaration, especially for powerFx, it may be useful for other hosts.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
VamsiModem: Just being defensive here, as have an assertion in SetCoercedToplevelType()

lesaltzm: Why would we only need to set it if it's not valid? when would it have been set otherwise?

lesaltzm: Wouldn't the assert failing be the correct behavior?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -98,9 +98,19 @@ public void CheckTypesOnDeclaration(CheckTypesContext context, DType actualBodyR
             Contracts.AssertValue(actualBodyReturnType);
             Contracts.AssertValue(errorContainer);
 
-            if (!ReturnType.Kind.Equals(actualBodyReturnType.Kind) && !actualBodyReturnType.CoercesTo(ReturnType, true, false, context.Features.PowerFxV1CompatibilityRules))
+            if (!ReturnType.Kind.Equals(actualBodyReturnType.Kind))
```
lesaltzm, line 101: Why Equals instead of Accepts?

Wouldn't it be fine to return, e.g. a Record with fewer fields?

lesaltzm:
> > Can you add a test to demonstrate what this change is doing?
> 
> These changes are very specific to PAClient, especially coercion, because PFX still uses old engine for UDFs.

This is not correct. You could absolutely add tests where the result type of a UDF needs to be coerced, that are testable in pfx by either validating the coercion map in the binding, or even in the interpreter.

lesaltzm (changes requested):
see above

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
```diff
@@ -48,14 +48,14 @@ public static ParseUserDefinitionResult Parse(string script, ParserOptions parse
             return TexlParser.ParseUserDefinitionScript(script, parserOptions);
         }
 
-        public static bool ProcessUserDefinitions(string script, INameResolver globalNameResolver, IBinderGlue documentBinderGlue, BindingConfig bindingConfig, ParserOptions parserOptions, out UserDefinitionResult userDefinitionResult, Features features = null)
+        public static bool ProcessUserDefinitions(string script, INameResolver globalNameResolver, IBinderGlue documentBinderGlue, BindingConfig bindingConfig, ParserOptions parserOptions, out UserDefinitionResult userDefinitionResult, Features features = null, bool shouldBindBody = false)
```
lesaltzm, line 51: What's the point of this flag?
 Also, this UDF codepath is mostly new code, there should be documentation comments for most of these methods/parameters.

lesaltzm: When would it make sense to not bind, when calling this?
I think this pattern is somewhat off, would be good to consider reworking to similar to Engine's handling of expression:

Parse(string) -> ParseResult
Check(ParseResult) -> CheckResult (does binding)
Eval(CheckResult) -> ....

lesaltzm: The single method doing everything with a bunch of params feels odd

VamsiModem: For UDFs, we bind the body on declaration, to verify if provided return type matches with actual return type. This may be useful for other hosts, but for power apps this results in duplicate binding.

VamsiModem:
> > > Can you add a test to demonstrate what this change is doing?
> > 
> > 
> > These changes are very specific to PAClient, especially coercion, because PFX still uses old engine for UDFs.
> 
> This is not correct. You could absolutely add tests where the result type of a UDF needs to be coerced, that are testable in pfx by either validating the coercion map in the binding, or even in the interpreter.

In this case we are not updating coercion map, we are just setting the CoercedToplevelType, that is used later in PAClient to do coercion codegen.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
```diff
@@ -103,21 +96,9 @@ private IEnumerable<UserDefinedFunction> CreateUserDefinedFunctions(IEnumerable<
                 userDefinedFunctions.Add(func);
             }
 
-            BindUserDefinedFunctions(userDefinedFunctions, ReadOnlySymbolTable.NewDefault(texlFunctionSet) as INameResolver, errors);
```
VamsiModem, line 106: User defined functions shouldn't be bound on creation, this will result in double binding, instead hosts can bind body by calling UserDefinedfunction.BindBody()

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
VamsiModem: Added comments & cleaned up the API

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
VamsiModem: Changed to Accepts()

lesaltzm (approved):

Issue #1800, opened 2023-08-31
Exception is popping for blocked entities for single org hosts


anderson-joyle:
`First('Appointments').appointment_PostRoles`

```
System.InvalidCastException: Unable to cast object of type 'Microsoft.PowerFx.Types.BindingErrorType' to type 'Microsoft.PowerFx.Types.TableType'.
   at Microsoft.PowerFx.Dataverse.DataverseRecordValue.ResolveOneToManyRelationship(OneToManyRelationshipMetadata relationshipMetadata, FormulaType fieldType, CancellationToken cancellationToken)
   at Microsoft.PowerFx.Dataverse.DataverseRecordValue.TryGetFieldAsync(FormulaType fieldType, String fieldName, CancellationToken cancellationToken)
   at Microsoft.PowerFx.Types.RecordValue.GetFieldAsync(FormulaType fieldType, String fieldName, CancellationToken cancellationToken)
   at Microsoft.PowerFx.EvalVisitor.Visit(RecordFieldAccessNode node, EvalVisitorContext context)
   at Microsoft.PowerFx.ParsedExpression.EvalAsync(CancellationToken cancellationToken, IRuntimeConfig runtimeConfig)
   at FxPlayground.Core.Session.TryEvalAsync(String expression, UserInfo user, CancellationToken cancel) in D:\dev\pa2\PowerFxPlayground\src\FxPlaygroundCore\Session.cs:line 576
   at PowerFxService.Controllers.EvaluationController.HandleRequestAsync(String session, RequestBody body) in D:\dev\pa2\PowerFxPlayground\src\WebDemo\Controllers\EvaluationController.cs:line 159
```

This error started to pop up after https://github.com/microsoft/Power-Fx-Dataverse/issues/264.
That fix revealed a `DType.Error` not being handled.

PR #1801, opened 2023-08-31
Handles error when trying to get entity schema

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/1800.

### DIFF
--- a/src/strings/PowerFxResources.en-US.resx
+++ b/src/strings/PowerFxResources.en-US.resx
@@ -5240,8 +5240,4 @@
     <value>Set a variable with an online datasource value (table or record) can cause multiple requests to the server. Soon this will be blocked.</value>
     <comment>Warning for makers that a certain capability should be avoided.</comment>
   </data>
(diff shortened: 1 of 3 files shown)


src/strings/PowerFxResources.en-US.resx
```diff
@@ -5240,4 +5240,8 @@
     <value>Set a variable with an online datasource value (table or record) can cause multiple requests to the server. Soon this will be blocked.</value>
     <comment>Warning for makers that a certain capability should be avoided.</comment>
   </data>
+  <data name="ErrInvalidRelationshipEntity" xml:space="preserve">
+    <value>It was not possible to get the schema for a relationship entity. It might be a private one.</value>
```
MikeStall, line 5244: What error does Power Apps give in this scenario?

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #1802, opened 2023-08-31
Restrict usage of type 'String' for return type & param types in UDFs

VamsiModem:


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
```diff
@@ -96,8 +99,14 @@ private IEnumerable<UserDefinedFunction> CreateUserDefinedFunctions(IEnumerable<
                 {
                     continue;
                 }
-                
-                var func = new UserDefinedFunction(udfName.Value, udf.ReturnType.GetFormulaType()._type, udf.Body, udf.IsImperative, udf.Args);
```
Shpakh-MSFT, line 100: >GetFormulaType [](http://example.com/codeflow?start=81&length=14)

Why not fix GetFormulaType to return valid type here?
Why not update String FormulaType to Text?

lesaltzm (approved):

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
VamsiModem: Modified GetFormulaType() to return valid type

lesaltzm (approved):

LucGenetier:
✅ No public API change.

PR #1807, opened 2023-09-01
mark more obsolete behavior

MikeStall:
Move to internals.


gregli-msft (approved):

LucGenetier:
Public API changes: <br />❌CP0001 T:Microsoft.PowerFx.Interpreter.UDF.DefineFunctionsResult<br />❌CP0002 M:Microsoft.PowerFx.RecalcEngine.DefineFunctions(System.String,System.Boolean)

EvanPriceSencom:
What was the motivator for removing the ability to register user-defined functions? Was this functionality moved elsewhere? Defining reflection functions is generally the more powerful option, however there are benefits to users being able to define helper functions without the need to recompile and redeploy the C# solution.

MikeStall:
@EvanPriceSencom - this was a proof-of-concept implementation. We're in the middle of a fully productized implementation that will be public soon, but the APIs for that will look different.

EvanPriceSencom:
> @EvanPriceSencom - this was a proof-of-concept implementation. We're in the middle of a fully productized implementation that will be public soon, but the APIs for that will look different.

Understood, thank you for the reply. Looking forward to what comes of it; things are fantastic thus far.

PR #1811, opened 2023-09-06
[Performance Improv]: Moving Skipping Semantic Tokens logic closer to core algorithm and Allow Skipping Function tokens "Partially"

akshar2401:
An ability to skip some of the semantic tokens allows new formula bar to complement language server really well when it comes to overall tokenization and syntax highlighting work. It is heavily used by new formula bar to lighten the load on language server to compute tokens for syntax highlighting. **This ability results in better performance and huge reduction in payload sizes exchanged between new formula bar and language server for semantic tokenization. This ability basically translates to "Hey Language Server, I am formula bar and i promise you i can recognize certain token types correctly so i want you to skip those tokens and do not worry about computing them or sending them back to me".** Since the implementation of semantic tokenization, this has been a crucial part of semantic tokenization that has allowed language server to not worry about token types like strings, string interpolations, comments and other such tokens which include multi line tokens that tend to increase the time needed for language server to compute tokens and send them back. 

**This pr further improves the performance by moving this skipping ability into the core logic that compute tokens. This allows us to skip tokens early and avoid unnecessary computations and keep the tokens that need to be send back very small in size. Unlike old formula bar, new formula bar is very good and efficient at recognizing many powerfx tokens on its own and thus it uses this skipping ability to avoid computing them on the server side as it has already computed those tokens when the server responds with its set of tokens. One such token is the function names. New formula bar can now recognize function names that are immediately followed by ( (for example Max in Max(1,2,3). This is the most common case in different formulas. This allows us new formula bar to tell language server to skip those functions that are followed by (. Therefore, this pr includes changes that partially skips function tokens that are immediately followed by (. It doesn't skip the ones that are not followed by ( and have new lines or spaces in between (for example Max in Max    (1,2,3)). This alone resulted in a significant reduction about (40% in terms of number of tokens being exchanged between server and new formula bar) for some of the example formulas.**

On top of that, this also helps UDFs. New formula bar would be able to identify functions in UDf syntax and color the m without the support of the server.** 

Semantic tokenization is the hottest activity in formula bar and language server. Our telemetry shows 4 million requests in the last 7 days in production. Any optimizations is good one for semantic tokenization and skipping ability would play a significant role

New formula bar can recognize functions in UDF syntax without the help of server
UFBWithoutServer

New formula bar can recognize functions that are immediately followed by ( which is the most natural case
UFBwithoutServer2

Non intuitive but new formula bar needs server to recognize functions that are not followed by (. 
UFBWithoutServer3

From the images, you can see how both formula bar and language server complement each other. When one fails, other fills in for it. This complementary is at the heart of rich syntax highlighting feature


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Tokenization.cs
```diff
@@ -342,5 +381,58 @@ private static TokenType MapNodeKindToTokenType(NodeKind kind)
...
         }
+
+        /// <summary>
+        /// Adds token to the given list of tokens if it is feasible (token should not be skipped).
+        /// </summary>
+        /// <param name="tokens">List of tokens to which given token might be added.</param>
+        /// <param name="token">Token to be added if feasible.</param>
+        /// <param name="tokenTypesToSkip">Optional: Token types to be skipped.</param>
+        /// <param name="expression">Optional: Expression.</param>
+        private static void AddTokenIfPossible(this ICollection<ITokenTextSpan> tokens, ITokenTextSpan token, IReadOnlyCollection<TokenType> tokenTypesToSkip = null, string expression = null)
```
jas-valgotar, line 392: >void [](http://example.com/codeflow?start=23&length=4)

can we return bool?

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Tokenization.cs
```diff
@@ -342,5 +381,58 @@ private static TokenType MapNodeKindToTokenType(NodeKind kind)
...
+        /// <summary>
+        /// Determines where the given token should be skipped or not.
+        /// </summary>
+        /// <param name="token">Token to be skipped.</param>
+        /// <param name="tokenTypesToSkip">Optional: Token types to be skipped.</param>
+        /// <param name="expression">Optional: Expression.</param>
+        /// <returns>True if this token should be skipped or false otherwise.</returns>
+        private static bool ShouldBeSkipped(ITokenTextSpan token, IReadOnlyCollection<TokenType> tokenTypesToSkip = null, string expression = null)
+        {
+            var shouldBeSkippedBasedOnTokenType = token.TokenType.ShouldBeSkipped(tokenTypesToSkip);
```
jas-valgotar, line 420: >token.TokenType.ShouldBeSkipped(tokenTypesToSkip); [](http://example.com/codeflow?start=50&length=50)

Do we need to worry about nulls?token?.TokenType?.ShouldBeSkipped(tokenTypesToSkip);

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Tokenization.cs
```diff
@@ -342,5 +381,58 @@ private static TokenType MapNodeKindToTokenType(NodeKind kind)
...
+            
+            // Client requesting to skip function type means that it can successfully recognize function tokens that are immediately 
+            // followed by open paren (
+            // In that case, "only" skip function tokens that are immediately followed by open paren
+            // For example, Max in "Max(1,2,3)" would be skipped while Max in "Max (1,2,3)" would not be skipped as there's a space between 
+            // Max and open paren (
+            // Function name immediately followed by open parent is the most common case
+            // Thus, Doing this drastically reduces the number of tokens we have to send back to client
+            // thus increasing the performance of semantic tokenization
+            if (shouldBeSkippedBasedOnTokenType && token.TokenType == TokenType.Function)
```
jas-valgotar, line 430: >token.TokenType [](http://example.com/codeflow?start=51&length=15)

Same, can `token` be null?

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Tokenization.cs
```diff
@@ -188,10 +200,26 @@ private static bool CanHideLeftHandSideOfDottedName(NameInfo nameInfo, TexlBindi
         /// <param name="tokens">Collection in which tokens created from th given nodes would be stored.</param>
         /// <param name="binding">Binding which would be used to tokenize the operation and determine the type of each token.</param>
         /// <param name="compilerGeneratedNodes">Collection that contains ids of compiler generate nodes.</param>
-        private static void ExtractTokensFromStringInterpolationNodesList(ICollection<ITokenTextSpan> tokens, TexlBinding binding, ICollection<int> compilerGeneratedNodes)
+        /// <param name="tokenTypesToSkip">Options: Collection of token types to potentially skip in the final result. Tokens with these token types may or may not be skipped depending on their types.</param>
+        private static void ExtractTokensFromStringInterpolationNodesList(ICollection<ITokenTextSpan> tokens, TexlBinding binding, ICollection<int> compilerGeneratedNodes, IReadOnlyCollection<TokenType> tokenTypesToSkip = null)
```
jas-valgotar, line 204: > IReadOnlyCollection<TokenType> tokenTypesToSkip = null) [](http://example.com/codeflow?start=171&length=56)

I am assuming, all the signature change doesn't break PA.

jas-valgotar (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Tokenization.cs
akshar2401: It is not necessary TBH. It won't be used. This is another Add function on the collection of tokens and Adds don't return bool. don't wanna do something that is not needed. Let me know if you think otherwise

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Tokenization.cs
akshar2401: That's not possible at all here. We are the ones who create these token objects inside a "private" function so need to worry about that.

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Tokenization.cs
akshar2401: Nope it doesn't. Plus i modified an internal method which is only used language server in a private function. this is not anywhere near to "public" and PA doesn't use only language server does

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/Tokenization.cs
akshar2401: not it cannot be null

PR #1813, opened 2023-09-07
Fix Engine.UpdateVariable for Int and add more overloads

gregli-msft:
Consistent with how we pull data in from connectors, UpdateVariable with an int should return a Decimal instead of a Number.  In addition, this PR adds additional overloads for common data types that a C# host might want to use.


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -93,14 +93,39 @@ public void UpdateVariable(string name, decimal value)
...
         }
 
         public void UpdateVariable(string name, long value)
         {
             UpdateVariable(name, new DecimalValue(IRContext.NotInSource(FormulaType.Decimal), value));
         }
 
+        public void UpdateVariable(string name, string value)
+        {
+            UpdateVariable(name, new StringValue(IRContext.NotInSource(FormulaType.String), value));
```
MikeStall, line 106: We should call the FormulaValue.New() overloads. That more closely mimics the public surface. 

(separately, we should get rid of IRContext)

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -93,14 +93,39 @@ public void UpdateVariable(string name, decimal value)
 
         public void UpdateVariable(string name, int value)
         {
-            UpdateVariable(name, new NumberValue(IRContext.NotInSource(FormulaType.Number), value));
+            UpdateVariable(name, new DecimalValue(IRContext.NotInSource(FormulaType.Decimal), value));
```
Shpakh-MSFT, line 96: >new DecimalValue(IRContext.NotInSource(FormulaType.Decimal), [](http://example.com/codeflow?start=33&length=60)

Is this not a breaking change?

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Connectors.Tests/ConnectorWizardTests.cs
```diff
@@ -139,7 +139,7 @@ public void ConnectorWizardTest()
             // Finally contruct the expression
             string expression = function.GetExpression("SQL", parameters);
 
-            Assert.Equal(@"SQL.ExecuteProcedureV2(""default"", ""default"", ""sp_2"", { p1: Float(50), p2: ""abc"" })", expression);
+            Assert.Equal(@"SQL.ExecuteProcedureV2(""default"", ""default"", ""sp_2"", { p1: Decimal(50), p2: ""abc"" })", expression);
```
MikeStall, line 142: @LucGenetier - fyi, this may be a merge conflict for you

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Match.cs
```diff
@@ -162,7 +162,7 @@ private bool TryCreateReturnType(TexlNode regExNode, string regexPattern, IError
 
                 if (!startMatchHidden)
                 {
-                    propertyNames.Add(new TypedName(DType.Number, ColumnName_StartMatch));
+                    propertyNames.Add(new TypedName(DType.Decimal, ColumnName_StartMatch));
```
CarlosFigueiraMSFT, line 165: >Decimal [](http://example.com/codeflow?start=58&length=7)

We should only change if Not(context.NumberIsFloat)

gregli-msft: Great call.  I backed out the regular expression changes for now.  The change in the compiler is trivial, but the change in the interpeter is more involved.  Regular expressions are not a part of Power Fx 1.0 so it can wait.  Tracked with https://github.com/microsoft/Power-Fx/issues/1816

src/tests/Microsoft.PowerFx.Connectors.Tests/ConnectorWizardTests.cs
gregli-msft: Nope, seems ok.

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNew.cs
```diff
@@ -42,9 +42,9 @@ public static DecimalValue New(long number)
             return new DecimalValue(IRContext.NotInSource(FormulaType.Decimal), (decimal)number);
         }
 
-        public static NumberValue New(int number)
+        public static DecimalValue New(int number)
```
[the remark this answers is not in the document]
gregli-msft: Added comment.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
gregli-msft: Great call.  Done.  Which led to many more changes, but also some more bugs being discovered.

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNew.cs
```diff
@@ -38,13 +38,16 @@ public static DecimalValue New(decimal number)
...
         {
-            // it is safe to convert 64-bit Long to Decimal
+            // it is safe to convert 64-bit Long to Decimal, but not to Float
             return new DecimalValue(IRContext.NotInSource(FormulaType.Decimal), (decimal)number);
         }
 
-        public static NumberValue New(int number)
-        {
-            return new NumberValue(IRContext.NotInSource(FormulaType.Number), number);
+        public static DecimalValue New(int number)
```
MikeStall, line 45: We're hitting issues trying to integrate upstack - _changing return type is a breaking change_. So this leads to binary breaking changes and is blocking a DV update. 

@LucGenetier - need to determine why our API compat checks in PR didn't flag this. 

(Nice explanation here https://stackoverflow.com/questions/15362424/method-with-same-name-and-signature-but-different-return-type-in-c-sharp)

PR #1814, opened 2023-09-07
Removed unused RequiresBindingContext virtual from TexlFunction

CarlosFigueiraMSFT:
This virtual property is only used in Power Apps - no need to have it in Power Fx.


LucGenetier:
✅ No public API change.

Shpakh-MSFT (approved):
:shipit:

PR #1815, opened 2023-09-07
Wiring up NL2Fx in LSP

MikeStall:
Expose NL features through LSP. 

1. introduce `CheckContextSummary` - this captures the set of symbols, with some semantic filtering, to be shared with NL / authoring tools.  It has other context like whether to allow side-effects, v1 semantics, etc.  It's based on the Engine so that it can also be fully overwridden by other derived engines (like SQL compiler) which don't use symbol table and use rule scope instead 
2. expose `SymbolEntry` as public facing counterpart to NameSymbol
3. Add NamedFormulaType.IsExpandEntity so that we can more effectively filter to avoid expanded types. RecordValues hav ethis, but not types. 
4. Custom $/nl2fx message on LSP for handling NL requests.


src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolEntry.cs
```diff
@@ -0,0 +1,37 @@
...
+    public class SymbolEntry
+    {
+        public string Name { get; init; }
+
+        public string DisplayName { get; init; }
+
+        /// <summary>
+        /// Prefer <see cref="DisplayName"/>if present, else show <see cref="Name"/>. 
+        /// </summary>
+        public string BestName => string.IsNullOrEmpty(this.DisplayName) ? this.Name : this.DisplayName;
```
jas-valgotar, line 29: >BestName [](http://example.com/codeflow?start=22&length=8)

super nit: should we call it PreferredName? (non blocking, just a suggestion)

src/libraries/Microsoft.PowerFx.Core/Binding/BindInfo/NameLookupInfo.cs
```diff
@@ -43,5 +44,25 @@ public NameLookupInfo(BindKind kind, DType type, DPath path, int upCount, object
...
+            x = new SymbolEntry
+            {
+                Name = ns.Name,
+                DisplayName = this.DisplayName,                
+                Properties = ns.Props,
+                Type = FormulaType.Build(this.Type),
+                Slot = ns
+            };
+            return true;
+        }
```
jas-valgotar, line 66: Test for public surface? 
matybe rename to [TryGet/TryConvertTo]SymbolEntry?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
```diff
@@ -402,6 +402,24 @@ IEnumerable<TexlFunction> INameResolver.LookupFunctionsInNamespace(DPath nameSpa
...
+                {
+                    if (variable.Value.TryToSymbolEntry(out var x))
+                    {
+                        names.Add(x);
+                    }
+                }
+            }
+        }
+
+        internal class EnumerateNamesOptions
```
jas-valgotar, line 419: >EnumerateNamesOptions [](http://example.com/codeflow?start=23&length=21)

How is this used?

src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
```diff
@@ -402,6 +402,24 @@ IEnumerable<TexlFunction> INameResolver.LookupFunctionsInNamespace(DPath nameSpa
             return _functions.WithNamespace(nameSpace);
         }
 
+        internal virtual void EnumerateNames(List<SymbolEntry> names, EnumerateNamesOptions opts)
```
jas-valgotar, line 405: >List [](http://example.com/codeflow?start=45&length=4)

IList<>?

src/libraries/Microsoft.PowerFx.Core/Binding/BindInfo/NameLookupInfo.cs
jas-valgotar: Oops, missed it.

src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
MikeStall: yes, I wanted future options to be easy, since this is a policy-driven enumeration. 
This is internal, so less penalty to have it.

jas-valgotar (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
MikeStall: IList doesn't have Add() !

MikeStall: I'm wrong - it is present, via ICollection<T>

src/libraries/Microsoft.PowerFx.Core/Public/Types/AggregateType.cs
```diff
@@ -122,11 +122,18 @@ public IEnumerable<NamedFormulaType> GetFieldTypes()
...
             return FieldNames.Select(field =>
             {
-                var fieldType = GetFieldType(field);
-                
                 var displayName = DType.TryGetDisplayNameForColumn(_type, field, out var dName)
                     ? dName
                     : null;
+
+                if (this.TryGetBackingDType(field, out var backingDType))
+                {
```
MikeStall, line 130: @jas-valgotar  - this one is worth double checking - it mirrors change you made for RecordValues.

src/libraries/Microsoft.PowerFx.Core/Public/Types/AggregateType.cs
```diff
@@ -122,11 +122,18 @@ public IEnumerable<NamedFormulaType> GetFieldTypes()
...
+                    var t = new TypedName(backingDType, new DName(field));
+                    return new NamedFormulaType(t, displayName);
+                }
```
jas-valgotar, line 133: This isn't used, but ideally we should
t
if (this.TryGetBackingDType(field, out var backingDType))
{
t = new TypedName(backingDType, new DName(field));
}
else
{
// ideally code never hits this, but just to be safe
t = new TypedName(ObjNull/Deferred, new DName(field));
}
.
return

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Types/AggregateType.cs
```diff
@@ -122,11 +122,18 @@ public IEnumerable<NamedFormulaType> GetFieldTypes()
...
+
+                var fieldType = GetFieldType(field);
```
jas-valgotar, line 135: >fieldType [](http://example.com/codeflow?start=20&length=9)

This is not used, but this should not matter, if above was false, this would throw... 
Most likely we can remove it with if/else I mentioned

PR #1818, opened 2023-09-07
Do not suggest Deprecated Connector functions

jas-valgotar:
-If a path is deprecated or internal for Intellisense, do not suggest it.
-Dotted Name already was blocking it, FirstNameNode also blocks it now.


MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1821, opened 2023-09-11
Reworking decimal tests for DV Decimal compatibility

gregli-msft:
Most of the decimal specific tests aren't running when testing the SQL compiler because the range of decimal used was often 1.000000000000000000000001 which will round to 1 with a DV decimal.  Existing tests are duplicated and modified to be in a compatible range.  These tests are remaining here because they also work in C# (which the bigger range) and this ensures consistency.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1824, opened 2023-09-11
Create Cx build that avoids breaking change

MikeStall:
#1813 made a breaking change to FormulaValue.New(int) . 
Create a Cx build that avoids the call to this method so that we don't hit binary breaks.


LucGenetier:
✅ No public API change.

PR #1836, opened 2023-09-13
Add SortText to Completion Results

jack-work:
The intellisense results that are provided to the client are in sorted order.  Names whose types match the expected type of an expression context are prioritied over those that do not match types.  However, LSP sorts the completion items lexicographically unless a value for SortText is provided.  To maintain this ordering, the server simply embeds the index of the intellisense suggestion in its response to the client, thereby preserving order for all clients.


src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -300,22 +301,24 @@ private void HandleCompletionRequest(string id, string paramsJson)
...
-                });
-            }
+                    Kind = GetCompletionItemKind(item.Kind),
 
-            _sendToClient(JsonRpcHelper.CreateSuccessResult(id, new
-            {
-                items,
+                    // The order of the results should be preserved.  To do that, we embed the index
+                    // into the sort text, which clients may sort lexigraphically.
+                    SortText = index.ToString(CultureInfo.InvariantCulture),
```
jas-valgotar, line 316: >SortText [](http://example.com/codeflow?start=20&length=8)

This may intersect with the work, [https://msazure.visualstudio.com/OneAgile/_workitems/edit/24935881](https://msazure.visualstudio.com/OneAgile/_workitems/edit/24935881)
Is this change for the same?

akshar2401: Yes it is. We are coordinating on this. Jack was gracious enough to take care of this :)

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -300,22 +301,24 @@ private void HandleCompletionRequest(string id, string paramsJson)
...
+
+                    // If the current position is in front of a single quote and the completion result starts with a single quote,
+                    // we don't want to make it harder on the end user by inserting an extra single quote.
+                    InsertText = item.DisplayText.Text is { } label && label[0] == '\'' && precedingCharacter == '\'' ? label.Substring(1) : item.DisplayText.Text
```
akshar2401, line 320: Could we use the TexlLexer.IsIdentDelimiter(label[0]) instead of hard coding the delimiter? also helps if fx changes ident delimiter in future (less likely as it is GAed now)

jack-work: Yes this is a good idea

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
jack-work: Navaneeth is taking it to finish though :)

NavneetThekkumpat: Yes , my initial work fixed it in the  UFB side (thought was the approach decided). But definitely this is a better and the right way to fix(on the server side). We would need to make changes in UFB to respect this (server side) completion result and filtering/sorting that comes with it.

PR #1837, opened 2023-09-13
Fix Intellisense when Parameters are defined and Config is also intia…

jas-valgotar:
Fixes https://dynamicscrm.visualstudio.com/OneCRM/_workitems/edit/3590732


MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1838, opened 2023-09-14
Expose comments through ParseUserDefinitionResult

VamsiModem:
Return comments in user definitions, so they can be colored properly


src/tests/Microsoft.PowerFx.Core.Tests/UserDefinedFunctionTests.cs
```diff
@@ -153,5 +153,20 @@ public void TestCoercionWithUDFBody(string udfScript, string expectedIR)
...
+        public void TestCommentsFromUserDefinitionsScript(string script, int udfCount)
+        {
+            var parserOptions = new ParserOptions()
+            {
+                AllowsSideEffects = false
+            };
+
+            var parseResult = UserDefinitions.Parse(script, parserOptions);
+
+            Assert.Equal(udfCount, parseResult.Comments.Count());
```
CarlosFigueiraMSFT, line 169: We should have at least one test where we validate the span of the comments to make sure that they are correct.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
🚢

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #1839, opened 2023-09-14
Mark the Coalesce function as asynchronous to account for lazy evaluation

CarlosFigueiraMSFT:
Since we made the Coalesce function lazy, we need to mark it as asynchronous so that it may need to block while evaluating some of its arguments, since they can be asynchronous themselves.


LucGenetier:
✅ No public API change.

lesaltzm (approved):

PR #1840, opened 2023-09-14
Add Capabilities flag to LSP 

MikeStall:
Add Capabilities flag to LSP. 
Add Fx2NL hook on NLHandler. 
Update NL2Fx messages .

### DIFF
--- a/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
+++ b/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
@@ -135,14 +135,15 @@ namespace Microsoft.PowerFx.LanguageServerProtocol
                         case CustomProtocolNames.CommandExecuted:
                             HandleCommandExecutedRequest(id, paramsJson);
                             break;
+
                         case CustomProtocolNames.GetCapabilities:
-                            HandleGetCapabilities(id, paramsJson);
+                            new GetCapabilitiesHandler(this).Handle(id, paramsJson);
                             break;
                         case CustomProtocolNames.NL2FX:
-                            HandleNL2FxRequest(id, paramsJson);
+                            new NL2FxHandler(this).Handle(id, paramsJson);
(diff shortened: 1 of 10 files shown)


src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -135,9 +135,17 @@ public void OnDataReceived(string jsonRpcPayload)
                         case CustomProtocolNames.CommandExecuted:
                             HandleCommandExecutedRequest(id, paramsJson);
                             break;
+
+                        case CustomProtocolNames.GetCapabilities:
```
akshar2401, line 139: Happened to come across this. Why do we need Custom capabilities protocol when language server has a well defined initialize protocol that includes capabilities? I understand the need we need to get this done quickly but we should follow the protocol whenever we can

akshar2401: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.LanguageServerProtocol.NLHandler.NL2Fx(Microsoft.PowerFx.LanguageServerProtocol.NL2FxParameters,System.Threading.CancellationToken)

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
akshar2401: Chatted offline. Given the need of the hour, it should be okay to get this in as long as we follow up to do it the way protocol specifies.

MikeStall: Tracking with #1843

Issue #1792, opened 2023-08-26
OptionSetValue ToExpression needs to escape 


MikeStall:
Defined option set as:
```
    Language:
       - 'en-us'
       - 'fr-fr'
```
      
Actual:
`Language.en-us`

Expected
`Language.'en-us'`

PR #1841, opened 2023-09-14
Escaping option set when serialized

anderson-joyle:
Issue #1792


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Values/OptionSetValue.cs
```diff
@@ -77,7 +77,7 @@ public string DisplayName
 
         public override void ToExpression(StringBuilder sb, FormulaValueSerializerSettings settings)
         {
-            sb.Append($"{Type._type.OptionSetInfo.EntityName}.{Option}");
+            sb.Append($"{Type._type.OptionSetInfo.EntityName}.'{Option}'");
```
MikeStall, line 80: I think there's a helper that will do this intelligently? Ie, only add if needed?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Values/OptionSetValue.cs
MikeStall: This would also be incorrect if option contains ' since it wouldn't get properly escaped.

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.RecordValue.ToExpressionField(System.String)

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -619,7 +619,7 @@ public static string EscapeName(string name)
                     return sb.ToString();
                 }
 
-                if (!IsKeyword(name, out var kind))
+                if (!IsKeyword(name, out _) && !IsReservedKeyword(name))
```
MikeStall, line 622: ?
Could we hit this before?

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -59,27 +59,33 @@ private static object RegExSetup(PowerFxConfig config)
...
-                    { "option_2", "Option2" }
+            OptionSet optionSet = new OptionSet("OptionSet", DisplayNameUtility.MakeUnique(new Dictionary<string, string>()
+            {
+                    { "option_1", "Option1" },
+                    { "option_2", "Option2" },
+                    { "option-3", "Option-3" },
+                    { "Is", "Option IS" }, // Reserved word
+                    { "Self", "Option SELF" }, // Keyword
+                    { "is", "Option is low case" }, // Not a reserved word
+                    { "Self", "Option self low case" }, // Not a keyword
```
CarlosFigueiraMSFT, line 77: >Self [](http://example.com/codeflow?start=23&length=4)

I don't think we can have two keys with the same name. Did you mean to call it `self`?

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
anderson-joyle: Yes. Thanks for catching this.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -1,27 +1,27 @@
-﻿// Copyright (c) Microsoft Corporation.
-// Licensed under the MIT license.
-
-using System;
-using System.Collections.Generic;
+﻿// Copyright (c) Microsoft Corporation.
```
lesaltzm, line 1: Why are there so many diffs in this file? Line ending change?

lesaltzm (approved):

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
anderson-joyle: Dont know to be honest. From CodeFlow, everything looks good.

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
```diff
@@ -322,6 +322,7 @@ public override void ToExpression(StringBuilder sb, FormulaValueSerializerSettin
             sb.Append("}");
         }
 
+        [Obsolete("This will be removed in a future version.")]
```
CarlosFigueiraMSFT, line 325: >This will be removed in a future version. [](http://example.com/codeflow?start=19&length=41)

If someone is using it, what should they use instead?

CarlosFigueiraMSFT (dismissed):
:shipit:

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1153, opened 2023-03-07
Parser can restart on named formula errors

rick-nguyen:
Changes identifier lexer with semicolon as new delimiter end
Changes parser to add an error name formula to the list for a given identifier once, and continue parsing as normal after next semicolon (or terminates if we hit Eof)

PR #1846, opened 2023-09-15
Allow unified user definitions parser to restart

rick-nguyen:
Similar to previous PR: https://github.com/microsoft/Power-Fx/pull/1153

Restart no longer works with named formulas under the unified parser. This PR fixes that and allows UDFs to continue processing as well if the parser encounters an error.


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -135,7 +136,14 @@ private bool ParseUDFArgs(out HashSet<UDFArg> args)
                 if (TokEat(TokKind.Colon, addError: false) == null)
                 {
                     CreateError(_curs.TokCur, TexlStrings.ErrUDF_MissingParamType);
-                    break;
+
+                    // If the result was an error, keep moving cursor until end of named formula expression
```
lesaltzm, line 140: nit: end of expression (not just named formula)?

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -135,7 +136,14 @@ private bool ParseUDFArgs(out HashSet<UDFArg> args)
                 if (TokEat(TokKind.Colon, addError: false) == null)
                 {
                     CreateError(_curs.TokCur, TexlStrings.ErrUDF_MissingParamType);
-                    break;
+
+                    // If the result was an error, keep moving cursor until end of named formula expression
+                    while (_curs.TidCur != TokKind.Semicolon && _curs.TidCur != TokKind.Eof)
```
lesaltzm, line 141: If it's missing a param type, could we try to move to the `)` or `;`? maybe the body is still valid? not for this pr, but something to consider, udfs have some other syntax elements that can be used.

lesaltzm (approved):

LucGenetier:
✅ No public API change.

Shpakh-MSFT (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -144,11 +152,24 @@ private bool ParseUDFArgs(out HashSet<UDFArg> args)
                 if (varType == null)
                 {
                     CreateError(_curs.TokCur, TexlStrings.ErrUDF_MissingParamType);
+
+                    // If the result was an error, keep moving cursor until end of expression
+                    while (_curs.TidCur != TokKind.Semicolon && _curs.TidCur != TokKind.Eof)
+                    {
+                        _curs.TokMove();
+                    }
```
Shpakh-MSFT, line 160: This could be a helper as it is repeated multiple times.

PR #1850, opened 2023-09-18
Fix in ConnectorType, missing Fields

LucGenetier:


Robulane (approved):

LucGenetier:
✅ No public API change.

PR #1854, opened 2023-09-18
Add  insertText and SortText to the Intellisense completionResult

NavneetThekkumpat:
Original code by @jack-work from  https://github.com/microsoft/Power-Fx/pull/1836
Testing and validating in this PR.
 
Adding sortText and insertText to the completion result. https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -318,22 +319,24 @@ private void HandleCompletionRequest(string id, string paramsJson)
             // Avoid this if possible
             _logger?.Invoke($"[PFX] HandleCompletionRequest: Suggest results: Count:{result.Suggestions.Count()}, Suggestions:{string.Join(", ", result.Suggestions.Select(s => $@"[{s.Kind}]: '{s.DisplayText.Text}'"))}");
 
-            var items = new List<CompletionItem>();
-
-            foreach (var item in result.Suggestions)
+            var precedingCharacter = expression[cursorPosition - 1];
```
jas-valgotar, line 322: >cursorPosition [](http://example.com/codeflow?start=48&length=14)

Can  cursorPosition be 0?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -318,22 +319,24 @@ private void HandleCompletionRequest(string id, string paramsJson)
...
-            _sendToClient(JsonRpcHelper.CreateSuccessResult(id, new
-            {
-                items,
+                    // The order of the results should be preserved.  To do that, we embed the index
+                    // into the sort text, which clients may sort lexigraphically.
+                    SortText = index.ToString(CultureInfo.InvariantCulture),
+
+                    // If the current position is in front of a single quote and the completion result starts with a single quote,
+                    // we don't want to make it harder on the end user by inserting an extra single quote.
+                    InsertText = item.DisplayText.Text is { } label && TexlLexer.IsIdentDelimiter(label[0]) && precedingCharacter == TexlLexer.IdentifierDelimiter ? label.Substring(1) : item.DisplayText.Text
```
jas-valgotar, line 338: >TexlLexer.IsIdentDelimiter(label[0]) && precedingCharacter == TexlLexer.IdentifierDelimiter ? label.Substring(1) : item.DisplayText.Text [](http://example.com/codeflow?start=71&length=136)

Any test case that demos this?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -318,22 +319,24 @@ private void HandleCompletionRequest(string id, string paramsJson)
...
 
-            _sendToClient(JsonRpcHelper.CreateSuccessResult(id, new
-            {
-                items,
+                    // The order of the results should be preserved.  To do that, we embed the index
+                    // into the sort text, which clients may sort lexigraphically.
+                    SortText = index.ToString(CultureInfo.InvariantCulture),
+
+                    // If the current position is in front of a single quote and the completion result starts with a single quote,
+                    // we don't want to make it harder on the end user by inserting an extra single quote.
```
tomjschneider, line 337: I feel like this logic should be best left up to the client that is integrating Power Fx and not handled here. If I type `'H` and accept the suggestion `'Hello'`, `'Hello'` is still the correct `insertText`, not `Hello'`. This feels more like a client-side issues that should be handled there.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
NavneetThekkumpat: Removed this logic

jas-valgotar (dismissed):
:shipit:

tomjschneider (approved):

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
akshar2401: Why was this removed?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
akshar2401: Again why the removal?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
akshar2401: This logic is PowerFx specific and should leave in its language server not Monaco. We had a similar problem which looks like got fixed. If you did Connector. and selected the function from intellisense it would insert Connector.Connector.Foobar and not just FooBar. Who is responsible for not inserting Connector here again? Monaco doesn't know what Connector is.

akshar2401: To not confuse and block you @NavneetThekkumpat , Merge this pr as you have got the approvals and continue to handle this on the fx bar side. what matters is the end result which is that we need to solve the problem whether that is here or in Fx bar.

tomjschneider: Discussed in a call - I have come around to Akshar's suggestion and think this does make sense to handle on the Language Server, which has more available context for what the user is doing, particularly in cases when completion is manually triggered.

LucGenetier:
✅ No public API change.

jas-valgotar (approved):
:shipit:

akshar2401 (approved):

LucGenetier:
✅ No public API change.

PR #1855, opened 2023-09-18
NamedFormula edit experience improvement

Shpakh-MSFT:
This PR adds a base index in token span. This is needed so that when Formulas property rule changes, we don't need to rebind each and every named formula every time. We can only rebind the rules which are changed. Currently we need to do that as span for each named formula rule is not relative to named formula identifier token. This change adds the baseindex in each token which then can be used to report the span in whole formula rule, for error reporting.


src/libraries/Microsoft.PowerFx.Core/Localization/Span.cs
```diff
@@ -19,22 +19,28 @@ namespace Microsoft.PowerFx.Syntax
...
         public int Min { get; }
 
         /// <summary>
-        /// End index of this span.
+        /// End index of this span from base.
         /// </summary>
         public int Lim { get; }
 
-        internal Span(int min, int lim)
+        internal Span(int min, int lim, int? basemin = null)
```
MikeStall, line 36: >basemin [](http://example.com/codeflow?start=45&length=7)

nit - can name match property?

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -6,6 +6,7 @@
 using System.Globalization;
 using System.Linq;
 using System.Xml.Linq;
+using Microsoft.CodeAnalysis;
```
MikeStall, line 9: was this a VS error?

Shpakh-MSFT: Probably.

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -997,9 +1004,28 @@ private TexlNode ParseAs(TexlNode left, ITexlSource leftTrivia)
                 rhsIdentifier);
         }
 
+        private Token GetNewOrExistingToken(Token existingToken)
+        {
+            var isNamedFormulaMode = _flagsMode.Peek().HasFlag(Flags.NamedFormulas);
+            if (existingToken == null || !isNamedFormulaMode || existingToken.Span.BaseIndex != null)
```
Shpakh-MSFT, line 1010: >existingToken.Span.BaseIndex != null) [](http://example.com/codeflow?start=64&length=37)

Todo: Add comment about this check.

This supposed to be set only once. This is to avoid cloning token twice if baseIndex is already set for this token.

src/libraries/Microsoft.PowerFx.Core/Localization/Span.cs
```diff
@@ -44,6 +50,7 @@ internal Span(Span span)
 
             Min = span.Min;
             Lim = span.Lim;
+            BaseIndex = span.BaseIndex;
```
lesaltzm, line 53: Consider naming this `offset`? it's a little closer to what it actually is.

PR #1856, opened 2023-09-18
Allow Polymorphic Types to union.

jas-valgotar:
When combining two records:

1. If "field1" in Record1 is of a polymorphic type, and
2. "field1" in Record2 is of a specific Record type,

Then, the resulting type of "field1" should be the same specific Record type as that of "field1" in Record2.

However, please note that there is a potential runtime failure if the actual type resolved for the polymorphic type does not match the expected union type.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1863, opened 2023-09-20
Fix connection deprecated function logic 

LucGenetier:
When 2 functions share the same Path, we need to separate their deprecation status
One could be deprecated and not the other


LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #1865, opened 2023-09-20
NamedFormula edit experience improvement

Shpakh-MSFT:
This PR adds a base index in NF formula parse results for NF identifier. This is needed so that when Formulas property rule changes, we don't need to rebind each and every named formula every time. We can only rebind the rules which are changed. Currently we need to do that as span for each named formula rule is not relative to named formula identifier token. With this base index information, we can then just use new base Index and previous base index to calculate the offset for each NF edit, which we can then use for the error reporting.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1872, opened 2023-09-21
Fix string escaping bug

CarlosFigueiraMSFT:
The current string escaping logic is buggy: an expression such as `"{""123""}"` is not escaped properly. This fixes it, and adds some tests to validate it.


anderson-joyle (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1876, opened 2023-09-22
Sort suggestion correctly

jas-valgotar:


LucGenetier:
✅ No public API change.

Issue #1823, opened 2023-09-11
RoundUp doesn't error for large negative number of digits


gregli-msft:
Two problems, with decimal
```
> RoundUp( 123.12345678919, -28 )
10000000000000000000000000000

> RoundUp( 123.12345678919, -29 )
0
```
We should generate a runtime overflow error for the second case, consistent with #NUM! in Excel.

For floating point
```
> RoundUp( Float(123.12345678919), -14 )
100000000000000

> RoundUp( Float(123.12345678919), -15 )
999999999999999.9

> RoundUp( Float(123.12345678919), -16 )
123.12345678919

> RoundUp( Float(123.12345678919), -300 )
123.12345678919

> RoundUp( Float(123.12345678919), -400 )
123.12345678919
```
Not at all consistent with Excel, all numbers should be 1 with a variable number of zeros following.  -14, -15, and -16 is an interesting sequence.  For -400, we should generate an error.

PR #1877, opened 2023-09-22
Draft. Round functions changes

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/1823.
Open questions:
- The current code imposes a limit to positive digits up to 28 digits for decimals. Should an expression like `RoundUP(123, 9999)` also result in a runtime error?
- RoundDown has a similar scenario where the function will return arg[0] in case the digit limit is exceeded. Should we also return a runtime error in this case? If the answer is `Yes`, then `Trunc` function will also need some adjustments (contained in this draft PR). 

These questions are based on same tests in PA.
I left some comments to give a better understanding of the scenario.


src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -916,6 +915,17 @@ public static FormulaValue AbsDecimal(IRContext irContext, DecimalValue arg)
             return new DecimalValue(irContext, val);
         }
 
+        public static FormulaValue Trunc(IRContext irContext, FormulaValue[] args)
```
anderson-joyle, line 918: >Trunc [](http://example.com/codeflow?start=35&length=5)

In case we need to change `RoundDown`, `Trunc` function will have its own library function to skip some conditions.

anderson-joyle:
>                 return dec;

Should this also result in a runtime error?

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs:1016 in 858e8ca. [](commit_id = 858e8caa247a3231c7529afb0d44bc389f606d97, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1628,14 +1628,14 @@ static Library()
                 BuiltinFunctionsCore.Trunc,
                 StandardErrorHandling<FormulaValue>(
                     BuiltinFunctionsCore.Trunc.Name,
-                    expandArguments: InsertDefaultValues(outputArgsCount: 2, fillWith: new NumberValue(IRContext.NotInSource(FormulaType.Number), 0)),
+                    expandArguments: NoArgExpansion,
```
anderson-joyle, line 1631: >NoArgExpansion [](http://example.com/codeflow?start=37&length=14)

`Trunc` function will have its own target function and wont need default values being injected.

PR #1879, opened 2023-09-23
Create Microsoft.PowerFx.Repl

MikeStall:
Create a reusable nuget, Microsoft.PowerFx.Repl, for core REPL functionality.  This facilitates sharing the repl in other scenarios, such as with dataverse or web uis. 

The host's basic repl loop is then:
```
var repl = new PowerFxRepl { Engine = _engine };
while(true) {
   await repl.WritePromptAsync();
   string line = Console.ReadLine();
   await repl.HandleLineAsync(line); // executes
}
```

REPL includes ability to define new variables `Set(x,y)`.  This starts to overlap with .txt Mutation scripts - so those now include Repl and use it to replace the hacky TryMatchSet() behavior. 

This has extensibility hooks so that tools/repl can add internal functionality. 

The core design principle behind the REPL:
[1]  stateless - all state is in the associated Engine. 
[2] Services - all IO, Policy should be handled in a service object,like: 
- IReplOutput for writing output. 
- MultilineProcessor - for determining if an input is complete
- ValueFormatter  - for formatting results. 

We expect to make breaking changes to the REPL interface as we get feedback and apply to other situations. 

Some remaining items in #1883.


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Repl/MultilineProcessor.cs
```diff
@@ -0,0 +1,38 @@
...
+        // false if we're on a subsequent line. 
+        public bool IsFirstLine => _commandBuffer.Length == 0;
+
+        // Return null if we need more input. 
+        // else return string containing multiple lines together. 
+        public virtual string HandleLine(string line)
+        {
+            _commandBuffer.AppendLine(line);
+
+            // $$$ fix this check and apply ReadFormula logic.
```
MikeStall, line 25: @gregli-msft  - fyi

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

gregli-msft (approved):

LucGenetier:
✅ No public API change.

PR #1880, opened 2023-09-24
Gregli/repl3

gregli-msft:


MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Repl/PsuedoFunctions.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using System;
+using System.Threading;
+using Microsoft.PowerFx.Syntax;
+
+namespace Microsoft.PowerFx.Repl
+{
+    public interface IPseudoFunction
+    {
+        public abstract void Execute(CallNode callNode, PowerFxReplBase repl, ReadOnlySymbolTable extraSymbolTable, CancellationToken cancel);
```
MikeStall, line 12: >CallNode callNode, PowerFxReplBase repl, ReadOnlySymbolTable extraSymbolTable, CancellationToken cancel); [](http://example.com/codeflow?start=37&length=105)

can probably tweak this. Maybe just pass in checkResult?

gregli-msft: That could work.  Engine isn't public on CheckResult and I'm not sure we have an internal relationship?  I went ahead and merged and we can make this better later.

PR #1884, opened 2023-09-25
Fixes sorting issue.

jas-valgotar:
-sortText does lexicographical sort and hence we need to prepend string with 0s

otherwise,
e.g. if you had more than 10 suggestions... it would have sorted to
0
1
10
11
2
3
4
5
6
7
8
9


MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1886, opened 2023-09-25
Replacing UDF parser function

anderson-joyle:


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -218,13 +218,13 @@ public async Task<FormulaValue> EvalAsync(string expressionText, CancellationTok
         internal DefineFunctionsResult DefineFunctions(string script, bool numberIsFloat = false)
         {
             var parsedUDFS = new Core.Syntax.ParsedUDFs(script, numberIsFloat: numberIsFloat);
```
VamsiModem, line 220: Unused?

src/libraries/Microsoft.PowerFx.Core/Syntax/ParsedUDFs.cs
VamsiModem, line 1: This will become obsolete after the code is removed from RecalcEngine (See comment below)

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
VamsiModem, line 220: Unused?

VamsiModem (approved):

LucGenetier:
✅ No public API change.

PR #1887, opened 2023-09-25
Added Help and Notify provider model

gregli-msft:


src/tools/Repl/Program.cs
```diff
@@ -344,5 +347,54 @@ public FormulaValue Execute(StringValue option, BooleanValue value)
...
+
+<identifier> = <formula> defines a named formula with automatic recalc.
+    Example: F = m * a
+
+Available functions (case sensitive):
+";
+
+                var post =
+@"
+Available operators: = <> <= >= + - * / % && And || Or ! Not in exactin
```
CarlosFigueiraMSFT, line 372: Can you add the `^` operator, which is also supported?

PR #1888, opened 2023-09-25
Moved pseudo functions until after Check to reuse parse, added exception handling to IR function

gregli-msft:


MikeStall (approved):
:shipit:

PR #1891, opened 2023-09-26
Fix logic for handling decimal/thousands separator in Text function

CarlosFigueiraMSFT:
The current logic for handling separators is incorrect (see newly added test, which currently fails, but works in Excel) - this fixes it.


nguhoa (approved):

CarlosFigueiraMSFT:
While the validation now succeeds, the result is still different from Excel, so closing this one for now.

PR #1892, opened 2023-09-26
Connector fixes for Outlook 365

LucGenetier:
- when an HTTP body contains parameters with default values (and the parameter isn't specified in the expression), we were not adding it to the body
- some function return "binary" (or octet-stream) data - this was not supported and is now implemented (requires OpenApi nuget upgrade)
- in some cases functions with an internal body parameter were wrongly reported as optional (instead of required)
- functions leading to a PATCH request were all failing as we need "PATCH" to be entirely upper case
- DateTimeValue constructor was not happy with "no-time-zone" data type


MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1893, opened 2023-09-26
Extend Language Server with custom notification that publishes control tokens

NavneetThekkumpat:
- Extending  language server to add a custom notification protocol that publishes control tokens with the semantic tokens.
- Each published control token should have a range of its presence in the formula and the control name, along with the version
- The usage of version is to make sure the client handles only ​current semantic request version controls.
- The idea is to make use of this control token data that can be forwarded to the studio and can be used to highlight the tokens that are only controls.
- An example payload will look this 
``` 
  "payload":
                "[
                  {jsonrpc:2.0,id:<guid-id>,result:{resultId:null,data:[<semantictokendata>]}},
                  {jsonrpc:2.0,
                  method:$/publishControlTokens,
                  params:
                  {
                    version:<versionnumber>,
                    controls:[
                      {name:Button1,Ranges:[[0,0,0,7]]},
                      {name:Button2,Ranges:[[0,8,0,15]]},
                      {name:Button3,Ranges:[[1,0,1,7]]},
                      {name:Button4,Ranges:[[2,0,2,7]]},
                      {name:Button5,Ranges:[[3,0,3,7]]},
                      {name:Button6,Ranges:[[3,8,3,15]]}
                      ]
                  }
                 }]"
```


src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Schemas/ControlToken.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Microsoft.PowerFx.LanguageServerProtocol.Schemas
+{
+    internal class ControlToken
+    {
+        // Control name
+        public string Name { get; set; }
+
+        // Representing the range of the control token
+        public LinkedList<List<uint>> EncodedTokenIndices { get; set; }
```
akshar2401, line 17: Why not use Range class here which is language server compliant? Also linked list here looks a bit weird choice. You certainly don't need to lots of elements towards the end of the list.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -638,10 +639,10 @@ private void HandleRangeDocumentSemanticTokens(string id, string paramsJson)
                 return;
             }
 
-            HandleSemanticTokens(id, semanticTokensParams, TextDocumentNames.RangeDocumentSemanticTokens);
+            HandleSemanticTokens(id, semanticTokensParams, TextDocumentNames.RangeDocumentSemanticTokens, false);
         }
 
-        private void HandleSemanticTokens<T>(string id, T semanticTokensParams, string method)
+        private void HandleSemanticTokens<T>(string id, T semanticTokensParams, string method, bool isFullDocument = false)
```
akshar2401, line 645: Why the need to pass this in? IsRangeSemanticTokens == false means it is full semantic tokens. We already check that in handleSemantictokens function. If you are trying to avoid string comparison method == TextDocumentNames.... then remove the isRangeSemanticTokens and use the negation of isFullSemanticTokens

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -690,8 +691,36 @@ private void HandleSemanticTokens<T>(string id, T semanticTokensParams, string m
                 tokens = tokens.Where(token => !(token.EndIndex <= startIndex || token.StartIndex >= endIndex));
             }
 
-            var encodedTokens = SemanticTokensEncoder.EncodeTokens(tokens, expression, eol);
+            var controlTokens = isFullDocument ? new LinkedList<ControlToken>() : null;
```
akshar2401, line 694: Nice to see to this covered. We should not spend time adding and computing control tokens for the range.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -690,8 +691,36 @@ private void HandleSemanticTokens<T>(string id, T semanticTokensParams, string m
...
+
+            var version = queryParams.Get("version") ?? string.Empty;
+
+            // Send PublishControlTokens notification
+            _sendToClient(JsonRpcHelper.CreateNotification(
+                CustomProtocolNames.PublishControlTokens,
+                new PublishControlTokensParams()
+                {
+                    Version = version,
+                    Data = controlTokens
```
akshar2401, line 722: Name this as Controls and not Data. Semantic tokens is general while this is specific.

src/tests/Microsoft.PowerFx.Interpreter.Tests/Helpers/SemanticTokenTestHelper.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) Microsoft Corporation.
```
akshar2401, line 1: There's already a helper for semantic tokens test. Just move this there

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SemanticTokensEncoder.cs
```diff
@@ -160,7 +162,8 @@ public static IEnumerable<uint> EncodeTokens(IEnumerable<ITokenTextSpan> tokens,
         /// <param name="currentLineNumber">Curent Line Number in the expression.</param>
         /// <param name="previousTokenLineNumber">Line number of the previous encoded token.</param>
         /// <param name="previousTokenStartIdx">Starting position of the previous encoded token.</param>
-        private static void EncodeAndAddToken(ICollection<uint> encodedTokens, ITokenTextSpan currentToken, uint? currentTokenStartLine, uint? currentTokenStartPosition, uint? endPosition, uint currentLineNumber,  ref uint previousTokenLineNumber, ref uint previousTokenStartIdx)
+        /// <param name="controlTokens">Collection to add control token data.</param>
+        private static void EncodeAndAddToken(ICollection<uint> encodedTokens, ITokenTextSpan currentToken, uint? currentTokenStartLine, uint? currentTokenStartPosition, uint? endPosition, uint currentLineNumber, ref uint previousTokenLineNumber, ref uint previousTokenStartIdx, ICollection<ControlToken> controlTokens = null)
```
akshar2401, line 166: Make sure you test this in studio that semantic tokens are not affected at all by this (one off errors). While harmless changes, you never know.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Schemas/ControlToken.cs
akshar2401, line 17: Just call it range. EncodedTokenIndices is a concept of semantic tokens and shouldn't be a concept for this notification. We wouldn't decode these indices. we would just use them

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/SemanticTokensEncoderTests.cs
```diff
@@ -5,6 +5,8 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Numerics;
+using System.Reflection.Emit;
```
akshar2401, line 8: Seems like unused imports. make sure to go thru unused imports. Happens a lot with visual studio

akshar2401:
Looks like you copied the whole signal r payload in your description. This looks different when http is used. Just show the notification payload. {jsonrpc, method, ...}

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/PublishControlTokensParams.cs
```diff
@@ -0,0 +1,23 @@
...
+using Microsoft.PowerFx.LanguageServerProtocol.Schemas;
+
+namespace Microsoft.PowerFx.LanguageServerProtocol.Protocol
+{
+    internal class PublishControlTokensParams
+    {
+        /// <summary>
+        /// The Version for which token information is reported.
+        /// </summary>
+        public string Version { get; set; } = string.Empty;
```
akshar2401, line 16: Not for now but in future we should send back the whole uri once we move to workspace architecture in the new formula bar.

NavneetThekkumpat:
> Looks like you copied the whole signal r payload in your description. This looks different when http is used. Just show the notification payload. {jsonrpc, method, ...}

I agree, but I included the payload within it anyways.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
NavneetThekkumpat: Ah good point, lets stick with this bool. Ill change the string comparison that is already existing to the bool here.

src/tests/Microsoft.PowerFx.Interpreter.Tests/Helpers/SemanticTokenTestHelper.cs
NavneetThekkumpat: Ah missed that one. Lets use that.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Schemas/ControlToken.cs
NavneetThekkumpat: I agree, I can change the Linkedlist since it may not be necessary. But Range class expects position in int. But I believe for our use case restricting to uint is a good idea.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SemanticTokensEncoder.cs
```diff
@@ -190,8 +194,46 @@ private static void EncodeAndAddToken(ICollection<uint> encodedTokens, ITokenTex
...
+        /// <param name="controlTokens">Collection to add control tokens to.</param>
+        /// <param name="currentToken">Current token.</param>
+        /// <param name="currentLineNumber">current line number of the token.</param>
+        /// <param name="tokenStartPosition">Start Position of the token.</param>
+        /// <param name="tokenEndPosition">End Position of the token.</param>
+        internal static void HandleControlTokensData(ICollection<ControlToken> controlTokens, ITokenTextSpan currentToken, uint currentLineNumber, uint tokenStartPosition, uint tokenEndPosition)
+        {
+            if (currentToken.TokenType == TokenType.Control)
+            {
+                var controlTokenIfPresent = controlTokens.Where(controlToken => controlToken.Name == currentToken.TokenName);
```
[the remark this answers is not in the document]
akshar2401: We have done a lot of things to make semantic tokenization optimal and that includes avoid expensive (even little expensive) operations whenever we can. We should not undo all of that by ignoring it thinking this won't have that major impact

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/MockSymbolTable.cs
```diff
@@ -0,0 +1,44 @@
...
+using Microsoft.PowerFx.Core.Binding;
+using Microsoft.PowerFx.Core.Binding.BindInfo;
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.Utils;
+
+/** 
+ * Just a small handy mock of symbol table to be able to customize binder to compute different token types.
+ * Might not be 100% correct but it works and allows testing against different token types.
+*/
+internal class MockSymbolTable : ReadOnlySymbolTable
```
akshar2401, line 14: I added this in semantic tokens encoder tests file because i didn't intend it to be used outside. I see that you need it outside of that file now. In that case, please add a comment that this should only be used for semantic tokens related tests. This is very specific to that area. People can create new mocks by referencing this but we should just keep it "small handy mock".

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Schemas/ControlToken.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Microsoft.PowerFx.LanguageServerProtocol.Schemas
+{
+    internal class ControlToken
+    {
+        // Control name
+        public string Name { get; set; }
+
+        // Representing the range of the control token
+        public List<uint[]> Range { get; set; }
```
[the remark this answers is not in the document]
akshar2401: also nit: Just add a comment here explaining [0,1,2,3]. It is bit unnatural but we are doing it to keep payload sizes smaller. May be given an example. Also mention which index is start line number, which index is end line number so we know how to interpret this on the client side

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Schemas/ControlTokenDictionary.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.Text;
+using System.Xml.Linq;
+using Microsoft.PowerFx.Core.Texl.Intellisense;
+using Microsoft.PowerFx.LanguageServerProtocol.Protocol;
+
+namespace Microsoft.PowerFx.LanguageServerProtocol.Schemas
+{
+    internal class ControlTokenDictionary : IEnumerable<KeyValuePair<string, ControlToken>>
```
akshar2401, line 14: You do not need to implement IEnumerable of KeyValuePair. Just implement IEnumerable<ControlToken>. This way in LanguageServer.cs you won't have to do GetControlTokens(), you can just serialize the ControlTokenDictionary as IEnumerable<ControlToken>. Make things simple. Otherwise looks good to me

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -684,14 +683,42 @@ private void HandleSemanticTokens<T>(string id, T semanticTokensParams, string m
...
+        }
+
+        /// <summary>
+        /// Handles publishing a control token notification if any control tokens found.
+        /// </summary>
+        /// <param name="controlTokenDict">Collection to add control tokens to.</param>
+        /// <param name="queryParams">Collection of query params.</param>
+        private void PublishControlTokenNotification(ControlTokenDictionary controlTokenDict, NameValueCollection queryParams)
+        {
+            if (controlTokenDict == null || controlTokenDict.Count() == 0 || queryParams == null)
```
akshar2401, line 707: Low hanging fruit. But Count() would actually iterate to get the count. Alternatively you can just implement Count inside ControlTokenDictionary which just returns the underlying dictionary's size which doesn't require iterating

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Schemas/ControlTokenDictionary.cs
NavneetThekkumpat: I think I am going to leave this, since it helps in easy lookup.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SemanticTokensEncoder.cs
```diff
@@ -190,8 +195,45 @@ private static void EncodeAndAddToken(ICollection<uint> encodedTokens, ITokenTex
...
+
+        /// <summary>
+        /// Encodes the token and adds it to the given <paramref name="controlTokenDict"/> dictionary.
+        /// </summary>
+        /// <param name="currentToken">Current token.</param>
+        /// <param name="currentLineNumber">current line number of the token.</param>
+        /// <param name="tokenStartPosition">Start Position of the token.</param>
+        /// <param name="tokenEndPosition">End Position of the token.</param>
+        /// <param name="controlTokenDict">Controls Token Dictionary that keeps a collection of control tokens data.</param>
+        internal static void HandleControlTokensData(ITokenTextSpan currentToken, uint currentLineNumber, uint tokenStartPosition, uint tokenEndPosition, ControlTokenDictionary controlTokenDict)
```
akshar2401, line 216: nit: private

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/SemanticTokensEncoderTests.cs
```diff
@@ -197,12 +200,46 @@ public void TestMultilineTokensAreEncodedCorrectly(string expression)
...
+        [InlineData("9 + 9", 0)]
+        [InlineData("Max(10, 20, 30)", 0)]
+        [InlineData("Color.AliceBlue", 0)]
+        [InlineData("9 + 9\n Max(1, 3, 19); \n Color.AliceBlue", 0)]
+        [InlineData("/*jj*/\n\r\n\r\n//yes", 0)]
+        [InlineData("Label2.Text", 1, new string[] { "Label2,0,0,0,6" })]
+        [InlineData("NestedLabel1", 1, new string[] { "NestedLabel1,0,0,0,12" })]
+        [InlineData("Label2.Text;\nNestedLabel1", 2, new string[] { "Label2,0,0,0,6", "NestedLabel1,1,0,1,12" })]
+        [InlineData("Label2.Text;Label2.Text", 1, new string[] { "Label2,0,0,0,6,0,12,0,18" })]
+        public void TestControlTokensAreComputedCorrectlyWhenPresent(string expression, int expectedNumberOfControlTokens, string[] expectedControlTokenIndicesArray = null)
```
akshar2401, line 219: Can we add one more test case here? gallery.Selected.Label2 where both gallery and Label2 are controls. Look into the MockSymbolTable in TOkenizatonTests.cs to mock this.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Schemas/ControlTokenDictionary.cs
akshar2401: I will leave it up to you but it is against what's commonly done. if you get any further feedback on this then you can consider it again but feel free to leave it like this for now

akshar2401 (approved):

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Schemas/ControlToken.cs
```diff
@@ -0,0 +1,22 @@
...
+    internal class ControlToken
+    {
+        // Control name
+        public string Name { get; set; }
+
+        // Representing the ranges of the control token.
+        // Eg: [0,0,0,6] >> start line 0, start character 0, end line 0, end character 6.
+        // [0,0,0,6,1,1,1,7] >> First control range start line 0, start character 0, end line 0, end character 6
+        // Second control range start line 1, start character 1, end line 1, end character 7.
+        public List<uint[]> Ranges { get; set; }
```
MikeStall, line 20: >List [](http://example.com/codeflow?start=15&length=4)

curious - how often do we have more than 1 for a single control?

NavneetThekkumpat: Interesting one I guess maybe when there are multi line tokens this scenario might be prevalent. Other than that I think if the formula has multiple occurrence of the same control but accessing different properties of the same control. In that case we would still need to highlight all these control tokens(of the same control).

NavneetThekkumpat: We could potentially add some telemetry in the formula bar package to see the behavior on this one. If this scenario is low maybe we can change the approach here and keep it light.
I am fine with making the changes right away and go the other way.

akshar2401: Controls cannot be multi line tokens (they are lexed as identifiers which are single line only). I have seen formulas with the same control name in many cases. This grouping helps us reduce the overall payload size which is more important. old formula bar has there problems where it starts to act weird when dealing with huge payloads. We do not have native compression in signalR which we use today for communication, so this is critical for the performance. This was also one of the factors behind using indices for sortText instead of long strings. Additionally, we see about 3000 JSON payloads passing failures across UFB and Language Server every 7 days. This is due to big payloads which often hit the parser limits (we do have telemetry on this). But keeping them as small as possible is ideal for us

akshar2401: With this custom protocol, we do have flexibility to keep it tight

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Schemas/ControlTokenDictionary.cs
MikeStall: This pattern is a a bit weird...

Particularly that it's called "ControlTokenDictionary", but doesn't implement IDict* interfaces.

One approach is just:
class ControlTokens {

  public void Add(); 
  public IEnumerable<...> Tokens  {get;set; }
}

That's also gives you a path to add future other enumerations , like 
  public IEnumerable<string> TokensNames ;

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SemanticTokensEncoder.cs
```diff
@@ -190,8 +195,45 @@ private static void EncodeAndAddToken(ICollection<uint> encodedTokens, ITokenTex
...
+        {
+            if (currentToken.TokenType == TokenType.Control)
```
MikeStall, line 218: >if (currentToken.TokenType == TokenType.Control) [](http://example.com/codeflow?start=12&length=48)

This is the policy for adding the tokens?

So all token types of Control,  and those occur where DType.IsControl .  So the host creates these tokens by having firstName tokens of DType.IsControl =true?

akshar2401: Yes. It depends on how that first name was bound during binding. Binding is the heart of these tokens. For each token, a binding is queried to decide which token type it should be decorated with. For controls, it is BindKind.Control which i think has a DType.IsControl == true.

MikeStall (approved):
:shipit:

MikeStall (approved):
:shipit:

akshar2401 (approved):

akshar2401 (approved):

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SemanticTokensEncoder.cs
```diff
@@ -160,7 +164,8 @@ public static IEnumerable<uint> EncodeTokens(IEnumerable<ITokenTextSpan> tokens,
         /// <param name="currentLineNumber">Curent Line Number in the expression.</param>
         /// <param name="previousTokenLineNumber">Line number of the previous encoded token.</param>
         /// <param name="previousTokenStartIdx">Starting position of the previous encoded token.</param>
-        private static void EncodeAndAddToken(ICollection<uint> encodedTokens, ITokenTextSpan currentToken, uint? currentTokenStartLine, uint? currentTokenStartPosition, uint? endPosition, uint currentLineNumber,  ref uint previousTokenLineNumber, ref uint previousTokenStartIdx)
+        /// <param name="controlTokens">Control Tokens that keeps a collection of control tokens data.</param>
+        private static void EncodeAndAddToken(ICollection<uint> encodedTokens, ITokenTextSpan currentToken, uint? currentTokenStartLine, uint? currentTokenStartPosition, uint? endPosition, uint currentLineNumber, ref uint previousTokenLineNumber, ref uint previousTokenStartIdx, ControlTokens controlTokens)
```
jas-valgotar, line 168: >controlTokens [](http://example.com/codeflow?start=293&length=13)

If this is optional, let's assign null here, I saw a test that has to pass null...

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Schemas/ControlToken.cs
jas-valgotar, line 20: >List [](http://example.com/codeflow?start=15&length=4)

IList or ICollection maybe?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SemanticTokensEncoder.cs
NavneetThekkumpat: I did pass null to the function that calls this one.  Since this is a private one, we do not have tests on this particularly .

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/LanguageServerTests.cs
```diff
@@ -2195,6 +2200,56 @@ public void TestEmptyRangeSemanticTokensResponseReturnedWhenRangeIsNullOrInvalid
...
+                // Assert
+                Assert.Equal("2.0", notification.Jsonrpc);
+                Assert.Equal("$/publishControlTokens", notification.Method);
+                Assert.Equal("someVersionId", notification.Params.Version);
+
+                var controlTokenList = notification.Params.Controls;
+                Assert.Equal(expectedNumberOfControlTokens, controlTokenList.Count());
+                foreach (var controlToken in controlTokenList)
+                {
+                    Assert.Equal(typeof(ControlToken), controlToken.GetType());
```
jas-valgotar, line 2242: >Assert.Equal(typeof(ControlToken), controlToken.GetType()); [](http://example.com/codeflow?start=20&length=59)

super nit: Assert.IsAssignableFrom<>()

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Schemas/ControlToken.cs
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.PowerFx.LanguageServerProtocol.Protocol;
+
+namespace Microsoft.PowerFx.LanguageServerProtocol.Schemas
+{
+    internal class ControlToken
```
jas-valgotar, line 11: >ControlToken [](http://example.com/codeflow?start=19&length=12)

Can we add the ctor and mark fields readonly?

NavneetThekkumpat: I can add one, but we do not use this right now anywhere.

jas-valgotar: The reason is, we want field read-only. right now you don't need it becuase you are directly assigning fields, which in the long run are prone to bugs. 
Meaning sometimes people may change the value of fields at a wired point without realizing it.
Basically, we always prefer read-only fields unless its really need to be not one.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SemanticTokensEncoder.cs
jas-valgotar: If any function needs to pass null, let's assign it a default value of null.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Schemas/ControlToken.cs
NavneetThekkumpat: Fair enough, making the changes.

jas-valgotar (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1898, opened 2023-09-27
Prevent accidental type literal parse

rick-nguyen:
Visit isn't implemented yet for TypeLiteral nodes, causing an exception in Pretty print. This is a mitigation using ParserOptions() to avoid that.


src/libraries/Microsoft.PowerFx.Core/Parser/ParserOptions.cs
```diff
@@ -44,16 +44,22 @@ public class ParserOptions
...
+        public ParserOptions()
+        {
+        }
+
+        public ParserOptions(CultureInfo culture, bool allowsSideEffects = false, int maxExpressionLength = 0, bool allowParseAsTypeLiteral = false)
+        {
+            Culture = culture;
+            AllowsSideEffects = allowsSideEffects;
+            MaxExpressionLength = maxExpressionLength;
+            AllowParseAsTypeLiteral = allowParseAsTypeLiteral;
```
Shpakh-MSFT, line 61: >AllowParseAsTypeLiteral [](http://example.com/codeflow?start=12&length=23)

Does this need to be part of ctor? I don't see other flags are set this way. Default value of bool is false anyway, I think.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -71,14 +72,27 @@ private TexlParser(IReadOnlyList<Token> tokens, Flags flags, Features features =
...
+        public static ParseUDFsResult ParseUDFsScript(string script, CultureInfo loc = null, bool numberIsFloat = false)
+        {
+            Contracts.AssertValue(script);
+            Contracts.AssertValueOrNull(loc);
+
+            // UDFs always support ReservedKeywords, no back compat concern
+            var formulaTokens = TokenizeScript(script, loc, Flags.NamedFormulas | (numberIsFloat ? Flags.NumberIsFloat : 0));
+            var parser = new TexlParser(formulaTokens, Flags.NamedFormulas | (numberIsFloat ? Flags.NumberIsFloat : 0), new ParserOptions());
+
+            return parser.ParseUDFs(script);
```
Shpakh-MSFT, line 87: Why this change?

rick-nguyen: Existing API that was removed by Anderson recently, I kept it by mistake

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -55,13 +55,14 @@ public enum Flags
...
         private readonly Features _features;
+        private readonly ParserOptions _parserOptions;
 
         // Represents temporary extra trivia, for when a parsing method
         // had to parse tailing trivia to do 1-lookahead. Will be
         // collected by the next call to ParseTrivia.
         private ITexlSource _extraTrivia;
 
-        private TexlParser(IReadOnlyList<Token> tokens, Flags flags, Features features = null)
+        private TexlParser(IReadOnlyList<Token> tokens, Flags flags, ParserOptions parserOptions, Features features = null)
```
Shpakh-MSFT, line 65: > Flags flags, ParserOptions parserOptions [](http://example.com/codeflow?start=55&length=41)

We have two ways now which could be conflicting. Can we have a single way of defining the behavior? Can we add a flag as well for this?

LucGenetier:
✅ No public API change.

Shpakh-MSFT (approved):
:shipit:

PR #1902, opened 2023-09-28
Add PowerFxVersion function

LucGenetier:
image

With real nugets, we'll get this function return "Microsoft.PowerFx.Connectors.dll 1.1.0-build.20230928-1002"


src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -2542,5 +2547,12 @@ public static FormulaValue IsBlankOrError(IRContext irContext, FormulaValue[] ar
 
             return new BooleanValue(irContext, false);
         }
+
+        public static async ValueTask<FormulaValue> PowerFxVersion(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
+        {
+            IEnumerable<Assembly> powerFxAssemblies = AppDomain.CurrentDomain.GetAssemblies().Where(a => a.FullName.StartsWith("Microsoft.PowerFx", StringComparison.OrdinalIgnoreCase));
```
MikeStall, line 2553: Do not call AppDomain.CurrentDomain  or anything on AppDomain.  Consider it a banned API.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
LucGenetier: What alternate API do you suggest?

PR #1903, opened 2023-09-29
Fixes parser restart bug with UDFs

VamsiModem:
Parser wasn't restarting when a return type is missing in UDF, this PR fixes that


LucGenetier:
✅ No public API change.

Shpakh-MSFT:
>         [InlineData("a = Abs(1.2);\nAdd(a: Number, b: Number):Number = a + b;\nb = Add(1, 2);", 2, 1, 0)]

Can you adda test for following?

a = Abs(1.2);\nAdd(a: Number, b: Number):/*  Number */Number

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/UserDefinedFunctionTests.cs:195 in 514b4b2. [](commit_id = 514b4b2472f98c47c5ffb453b7882dadc2b4984e, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -301,7 +301,10 @@ private ParseUserDefinitionResult ParseUDFsAndNamedFormulas(string script, Parse
                     {
                         CreateError(_curs.TokCur, TexlStrings.ErrUDF_MissingReturnType);
                         udfs.Add(new UDF(thisIdentifier.As<IdentToken>(), colonToken, returnType: null, new HashSet<UDFArg>(args), body: null, _hasSemicolon, parserOptions.NumberIsFloat, isValid: false));
-                        break;
+
+                        // Return type not found, move to next user definition
+                        MoveToNextUserDefinition();
+                        continue;
```
Shpakh-MSFT, line 307: Why do we need this for return type?

VamsiModem: When a return type is not found it's still considered a parser error and we must move to the next user definition.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

Shpakh-MSFT (approved):
:shipit:

PR #1906, opened 2023-09-29
add AsDouble() 

MikeStall:
Add helpers to say:

`double x = value.AsDouble();`  

or AsDecimal. 

this avoids having to cast and guess whether it's decimal or double.


LucGenetier:
✅ No public API change.

jas-valgotar (approved):
:shipit:

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValue.cs
```diff
@@ -88,6 +89,68 @@ public bool TryGetPrimitiveValue(out object val)
...
+                    return b;
+                }
+            }
+
+            throw new InvalidOperationException($"Can't coerce to double from {this.Type._type.GetKindString()})");
+        }
+
+        public double AsDouble()
+        {
+            if (this is BlankValue)
```
CarlosFigueiraMSFT, line 114: This is dangerous - blanks are different than 0.0, and it is ideal to mix those two concepts.

MikeStall: Caller can always get blank if they care. But a common pattern is missing is default, some should make that easy.

Issue #1634, opened 2023-06-29
Check for Runtime mismatch for all type of Function Invocation.


jas-valgotar:
Today we check for runtime mismatch while Invoking _only_ TexlFunction.
https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs#L288

This causes us to miss issues in other invocations and we tend to find those issues later when those functions are wrapped inside TexlFunction. (e.g. https://github.com/microsoft/Power-Fx/issues/1598)

We should check the return type on all of the invocations (e.g. IAsyncTexlFunction)

PR #1910, opened 2023-10-02
UDF functions

anderson-joyle:


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -276,6 +278,11 @@ internal void AddFunction(TexlFunction function)
 
             // Add any associated enums 
             EnumStoreBuilder?.WithRequiredEnums(new TexlFunctionSet(function));
+        }
+
+        internal void AddUserFunction(TexlFunction function)
+        {
+            AddFunction(function);
         }
```
jas-valgotar, line 286: Why do we need this? can't we  directly call AddFunction()?

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -410,307 +409,128 @@ public void FormulaCantRedefine()
...
+            "foo(Table( { Value: \"Strawberry\" }, { Value: \"Vanilla\" } ))",
+            true,
+            null)]
+        [InlineData(
+            "foo(x:Number):Number = x + 1;",
+            "foo(Float(1))",
+            false,
+            2.0)]
+
+        public void UserDefiniedFunctionTest(string udfExpression, string expression, bool expectedError, object expected, bool isDecimal = false)
```
jas-valgotar, line 502: >expectedError [](http://example.com/codeflow?start=91&length=13)

Instead of bool, can we have param string[] errorMsg at the end, that way we know we are giving actionable errors

src/tests/Microsoft.PowerFx.Interpreter.Tests/SerializationTests.cs
```diff
@@ -18,6 +19,34 @@ namespace Microsoft.PowerFx.Interpreter.Tests
 {
     public class SerializationTests : PowerFxTest
     {
+        [Fact]
+        public void DateT3333333333333imeSerializationTests()
```
jas-valgotar, line 23: >DateT3333333333333imeSerializationTests [](http://example.com/codeflow?start=20&length=39)

typo?

src/tests/Microsoft.PowerFx.Interpreter.Tests/SerializationTests.cs
```diff
@@ -18,6 +19,34 @@ namespace Microsoft.PowerFx.Interpreter.Tests
...
+            config.SymbolTable.AddUserFunction(script, CultureInfo.CurrentCulture);
+
+            //var check = engine.Check("Add(1,2)");
+            //Assert.True(check.IsSuccess);
+
+            //check = engine.Check("Foo(1)");
+            //Assert.True(check.IsSuccess);
+
+            //check = engine.Check("y");
+            //Assert.True(check.IsSuccess);
```
jas-valgotar, line 39: remove commented code?

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -410,307 +409,128 @@ public void FormulaCantRedefine()
...
+        [InlineData(
+            "foo(x:Number):Number = x + 1;",
+            "foo(Table( { Value: \"Strawberry\" }, { Value: \"Vanilla\" } ))",
+            true,
+            null)]
+        [InlineData(
+            "foo(x:Number):Number = x + 1;",
+            "foo(Float(1))",
+            false,
+            2.0)]
```
jas-valgotar, line 500: nit: move test with other successful tests.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -410,307 +409,128 @@ public void FormulaCantRedefine()
...
+            null)]
+
+        // Syntax error
+        [InlineData(
+            "foo():Blank = x[",
+            null,
+            true,
+            null)]
+
+        // Incorrect parameters
```
jas-valgotar, line 485: >// Incorrect parameters [](http://example.com/codeflow?start=8&length=23)

Can we please add some tests with coercion?

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -410,307 +409,128 @@ public void FormulaCantRedefine()
...
-            var config = new PowerFxConfig();
-            var recalcEngine = new RecalcEngine(config);
-            Assert.False(recalcEngine.DefineFunctions("foo():Blank = foo();").Errors.Any());
-            var result = recalcEngine.Eval("foo()");
-            Assert.IsType<ErrorValue>(result);
-        }
-
-        [Fact]
-        public void DefHailstoneSequence()
+        [Theory]
```
jas-valgotar, line 414: >Theory [](http://example.com/codeflow?start=9&length=6)

Can we add test just to demo what happens when there already exist internal function with same name as udf? do we support overloading? should we?

jas-valgotar, line 414: >Theory [](http://example.com/codeflow?start=9&length=6)

Can we add test, demoing chaining of different udfs?
e.g. Odd(foo())

jas-valgotar, line 414: >Theory [](http://example.com/codeflow?start=9&length=6)

Any tests that demos runtime local? Ie number in fr?

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -267,16 +269,33 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
...
+                userDefinedFunc.BindBody(symbolsTable, new Glue2DocumentBinderGlue(), bindingConfig: BindingConfig.Default);
+
+                (var irnode, var ruleScopeSymbol) = userDefinedFunc.GetIRTranslator();
+
+                var expr = new ParsedExpression(irnode, ruleScopeSymbol, context.StackDepthCounter)
+                {
+                    _allSymbols = symbolsTable
+                };
+                
+                result = expr.Eval(symbolValues);
```
jas-valgotar, line 297: >expr.Eval(symbolValues); [](http://example.com/codeflow?start=25&length=24)

I think we should pass here runtime config with local?
Also, I would use Async api here.

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -353,6 +360,22 @@ public void AddHostObject(string name, FormulaType type, Func<IServiceProvider,
...
             _variables.Add(hostDName, info);
+        }
+
+        internal void AddUserFunction(string script, CultureInfo locale)
+        {
+            if (UserDefinitions.ProcessUserDefinitions(script, new ParserOptions() { AllowsSideEffects = true, Culture = locale }, out var userDefinitionResult) && !userDefinitionResult.HasErrors)
+            {
+                foreach (var udf in userDefinitionResult.UDFs)
+                {
+                    //udf.BindBody(this, new Glue2DocumentBinderGlue(), BindingConfig.Default);
```
jas-valgotar, line 371: >//udf.BindBody(this, new Glue2DocumentBinderGlue(), BindingConfig.Default); [](http://example.com/codeflow?start=20&length=75)

remove comment?

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -267,16 +269,33 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
...
             }
+            else if (func is UserDefinedFunction userDefinedFunc)
+            {
+                var recordValue = userDefinedFunc.GetRuntimeRecordValue(args, _cancellationToken);
+
+                var symbolsTableRecord = ReadOnlySymbolTable.NewFromRecord(recordValue.Type);
+                var symbolValuesRecord = SymbolValues.NewFromRecord(symbolsTableRecord, recordValue);
+
+                var symbolsTable = ReadOnlySymbolTable.Compose(_symbolValues.SymbolTable, symbolsTableRecord);
+                var symbolValues = ReadOnlySymbolValues.Compose(_symbolValues, symbolValuesRecord);
```
jas-valgotar, line 285: Like the way you leveraged the RecordValue :)

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -42,6 +41,15 @@ internal class UserDefinedFunction : TexlFunction
 
         public override bool IsSelfContained => !_isImperative;
 
+        private ReadOnlySymbolTable _symbols;
+
+        public ReadOnlySymbolTable AllSymbols => _symbols;
+
+        public void SetSymbols(ReadOnlySymbolTable symbols)
```
jas-valgotar, line 48: >SetSymbols [](http://example.com/codeflow?start=20&length=10)

Where is this used?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
anderson-joyle: This was the original entry point. I've moved to another place and forgot to remove this.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
anderson-joyle: I'll leave this at it is for now since we are moving this to txt files.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
anderson-joyle: I'll leave this at it is for now since we are moving this to txt files.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
anderson-joyle: I'll leave this at it is for now since we are moving this to txt files.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
anderson-joyle: The test case `DateT3333333333333imeSerializationTests` (at SerializationTests) had a similar test case. This test was left by mistaken tough but I will include another on to the new txt files.

anderson-joyle: Confirmed with @VamsiModem. Overloads are not supported yet but will eventually.

anderson-joyle: I'll add some in txt files.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
anderson-joyle: Removed. Thanks for catching this.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -148,25 +158,59 @@ public override IEnumerable<StringGetter[]> GetSignatures()
...
         }
 
+        public (IntermediateNode topNode, ScopeSymbol ruleScopeSymbol) GetIRTranslator()
+        {
+            return IRTranslator.Translate(_binding);
+        }
+
+        public RecordValue GetRuntimeRecordValue(FormulaValue[] args, CancellationToken cancellationToken)
+        {
+            var argsArray = _args.ToArray();
```
VamsiModem, line 168: >var argsArray = _args.ToArray(); [](http://example.com/codeflow?start=12&length=32)

Any specific reason to change to array? 
you could use ElementAt() right?

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -276,6 +278,11 @@ internal void AddFunction(TexlFunction function)
 
             // Add any associated enums 
             EnumStoreBuilder?.WithRequiredEnums(new TexlFunctionSet(function));
+        }
+
+        internal void AddUserFunction(TexlFunction function)
```
VamsiModem, line 283: >AddUserFunction [](http://example.com/codeflow?start=22&length=15)

naming nit: AddUserDefinedFunction?

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -214,7 +215,31 @@ public async Task<FormulaValue> EvalAsync(string expressionText, CancellationTok
             return result;
         }
 
-        // Preview functionality 
+        internal void AddUserFunction(string script, CultureInfo locale)
+        {
+            if (UserDefinitions.ProcessUserDefinitions(script, new ParserOptions() { AllowsSideEffects = true, Culture = locale }, out var userDefinitionResult) && !userDefinitionResult.HasErrors)
```
VamsiModem, line 220: >, [](http://example.com/codeflow?start=109&length=1)

naming nit: AddUserDefinedFunction

VamsiModem, line 220: We don't want to allow side effects for phase 1 of UDFs

VamsiModem, line 220: >ProcessUserDefinitions [](http://example.com/codeflow?start=32&length=22)

With the changes made recently, ProcessUserDefinitions will always returns true which is wrong and needs to be refactored.

ProcessUserDefinitions() can just return UserDefinitionResult instead of bool, with that change in place the above condition can be re-written

var result = UserDefinitions.ProcessUserDefinitions(script, new ParserOptions() { AllowsSideEffects = true, Culture = locale });

if (result.HasErrors){ ... }

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -214,7 +215,31 @@ public async Task<FormulaValue> EvalAsync(string expressionText, CancellationTok
...
+                foreach (var udf in userDefinitionResult.UDFs)
+                {
+                    _symbolTable.AddUserFunction(udf);
+                }
+            }
+            else
+            {
+                var sb = new StringBuilder();
+
+                sb.AppendLine("Something went worng when parsing user definied functions.");
```
VamsiModem, line 231: >definied [](http://example.com/codeflow?start=70&length=8)

nit: spelling

src/tests/Microsoft.PowerFx.Interpreter.Tests/SerializationTests.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
```
VamsiModem, line 6: do we still need this change?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
anderson-joyle: To avoid extra iterations by calling `.Count()`. Also, as far as I know `ElementAt` will iterate through the enumerable every time it is called.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

VamsiModem (dismissed):

jas-valgotar:
>                     }

Can we share this check with UDF eval on L299? 

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs:318 in 2695975. [](commit_id = 26959751d64d9c62983efdb3f038b2ac171ae7f9, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -4,16 +4,18 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Globalization;
+using System.Linq;
```
jas-valgotar, line 8: unused?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -187,7 +231,16 @@ public bool Lookup(DName name, out NameLookupInfo nameInfo, NameLookupPreference
...
+
+                    if (_fvArgs != null)
+                    {
+                        var formulaValue = _fvArgs[value.ArgIndex];
+                        nameInfo = new NameLookupInfo(BindKind.PowerFxResolvedObject, type, DPath.Root, 0, formulaValue);
+                    }
+                    else
+                    {                        
+                        nameInfo = new NameLookupInfo(BindKind.PowerFxResolvedObject, type, DPath.Root, 0, new UDFParameterInfo(type, value.ArgIndex, value.NameIdent.Name));
+                    }
```
jas-valgotar, line 243: Can you add comment for this part, what it is trying to do?

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -267,16 +269,35 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
...
+
+                userDefinedFunc.SetFormulaValues(args);
+                userDefinedFunc.BindBody(symbolsTable, new Glue2DocumentBinderGlue(), bindingConfig: BindingConfig.Default);
```
jas-valgotar, line 288: Why 2 separate call, why can't we pass it in BindBody()?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
jas-valgotar: Why do we need to call the count? can't we foreach?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -148,25 +158,59 @@ public override IEnumerable<StringGetter[]> GetSignatures()
...
+        }
+
         /// <summary>
         /// NameResolver that combines global named resolver and params for user defined function.
         /// </summary>
         private partial class UserDefinitionsNameResolver : INameResolver
         {
             private readonly INameResolver _globalNameResolver;
             private readonly IReadOnlyDictionary<string, UDFArg> _args;
+            private readonly FormulaValue[] _fvArgs;
```
jas-valgotar, line 189: >FormulaValue [](http://example.com/codeflow?start=29&length=12)

Why not IEnumberable<>?

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -267,16 +269,35 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
...
+
+                (var irnode, var ruleScopeSymbol) = userDefinedFunc.GetIRTranslator();
+
+                var expr = new ParsedExpression(irnode, ruleScopeSymbol, context.StackDepthCounter)
+                {
+                    _allSymbols = symbolsTable
+                };
+
+                RuntimeConfig runtimeConfig = new RuntimeConfig(symbolValues, CultureInfo);
```
jas-valgotar, line 297: >RuntimeConfig [](http://example.com/codeflow?start=16&length=13)

nit: `var`

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
MikeStall: Right, ElementAt is O[N].
The extra ToArray() usually isnt a problem. 

But if we're already creating a List<>, then we could just foreach and add to the list.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -108,14 +111,21 @@ public TexlBinding BindBody(INameResolver nameResolver, IBinderGlue documentBind
...
             bindingConfig = bindingConfig ?? new BindingConfig(this._isImperative);
-            _binding = TexlBinding.Run(documentBinderGlue, UdfBody, UserDefinitionsNameResolver.Create(nameResolver, _args), bindingConfig, features: features, rule: rule);
+            _binding = TexlBinding.Run(documentBinderGlue, UdfBody, UserDefinitionsNameResolver.Create(nameResolver, _args, _fvArgs), bindingConfig, features: features, rule: rule);
 
             CheckTypesOnDeclaration(_binding.CheckTypesContext, _binding.ResultType, _binding);
 
             return _binding;
         }
 
+        public void SetFormulaValues(FormulaValue[] formulaValues)
```
MikeStall, line 124: This could be an issue since we will need to reuse a binding across multiple invokes.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -267,16 +269,35 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
...
+            {
+                var recordValue = userDefinedFunc.GetRuntimeRecordValue(args, _cancellationToken);
+
+                var symbolsTableRecord = ReadOnlySymbolTable.NewFromRecord(recordValue.Type);
+                var symbolValuesRecord = SymbolValues.NewFromRecord(symbolsTableRecord, recordValue);
+
+                var symbolsTable = ReadOnlySymbolTable.Compose(_symbolValues.SymbolTable, symbolsTableRecord);
+                var symbolValues = ReadOnlySymbolValues.Compose(_symbolValues, symbolValuesRecord);
+
+                userDefinedFunc.SetFormulaValues(args);
```
MikeStall, line 287: >userDefinedFunc.SetFormulaValues(args); [](http://example.com/codeflow?start=16&length=39)

This could be a problem - we should treat userDefinedFunc as immutable; consider if the same function is called  multiple times:

Foo(1,2, Foo(3,4,5));

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -214,7 +215,34 @@ public async Task<FormulaValue> EvalAsync(string expressionText, CancellationTok
             return result;
         }
 
-        // Preview functionality 
+        internal void AddUserDefinedFunction(string script, CultureInfo locale)
```
MikeStall, line 218: >locale [](http://example.com/codeflow?start=72&length=6)

To clarify, can we rename to 'parseCulture'?

MikeStall, line 218: Can we move to be on SymbolTable?
A practical scenario here is calling on Engine (not RecalcEngine) and demosntrating bind-only scenarios. 

Ok to have a helper on RecalcEngine that passes through to this.SymbolTable.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -148,25 +158,59 @@ public override IEnumerable<StringGetter[]> GetSignatures()
             return new[] { _args.Select<UDFArg, TexlStrings.StringGetter>(key => _ => key.NameIdent.Name.Value).ToArray() };
         }
 
+        public (IntermediateNode topNode, ScopeSymbol ruleScopeSymbol) GetIRTranslator()
+        {
+            return IRTranslator.Translate(_binding);
+        }
+
+        public RecordValue GetRuntimeRecordValue(FormulaValue[] args, CancellationToken cancellationToken)
```
MikeStall, line 166: >FormulaValue[] args, [](http://example.com/codeflow?start=49&length=20)

Should we verify  these are the same length (args, _args)? Probably will never fire, but just  in case?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
MikeStall: might be nice to have count, indexing.  Maybe 
IReadOnlyCollection?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -187,7 +231,16 @@ public bool Lookup(DName name, out NameLookupInfo nameInfo, NameLookupPreference
...
                 {
                     var type = value.TypeIdent.GetFormulaType()._type;
-                    nameInfo = new NameLookupInfo(BindKind.PowerFxResolvedObject, type, DPath.Root, 0, new UDFParameterInfo(type, value.ArgIndex, value.NameIdent.Name));
+
+                    if (_fvArgs != null)
+                    {
+                        var formulaValue = _fvArgs[value.ArgIndex];
+                        nameInfo = new NameLookupInfo(BindKind.PowerFxResolvedObject, type, DPath.Root, 0, formulaValue);
+                    }
+                    else
```
MikeStall, line 240: Instead of changing here, make a change in EvalVisitor, Visit(ResolvedObjectNode node)

MikeStall:
>         public override async ValueTask<FormulaValue> Visit(ResolvedObjectNode node, EvalVisitorContext context)

Per my other comments - (in UserDefinitionsNameResolver) 
I think this functions will need to be update to know about parameters. 



---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs:726 in 2695975. [](commit_id = 26959751d64d9c62983efdb3f038b2ac171ae7f9, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -267,16 +269,35 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
...
+                (var irnode, var ruleScopeSymbol) = userDefinedFunc.GetIRTranslator();
+
+                var expr = new ParsedExpression(irnode, ruleScopeSymbol, context.StackDepthCounter)
+                {
+                    _allSymbols = symbolsTable
+                };
+
+                RuntimeConfig runtimeConfig = new RuntimeConfig(symbolValues, CultureInfo);
+
+                result = await expr.EvalAsync(_cancellationToken, runtimeConfig).ConfigureAwait(false);
```
MikeStall, line 299: >expr [](http://example.com/codeflow?start=31&length=4)

we probably also want to sue the same EvalVisitor on the udf.  (similar to what we do when running a lambda)
It has too much state (service provider, current culture, etc). 


For example, a test case is:

`Temp() : string => Language();`

Then this should return true:
`Language() = Temp()`

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
anderson-joyle: We could. I choose this way just to make it more explicit.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
anderson-joyle: I can add a verification but this can decrease our test coverage.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
anderson-joyle: Due to the new design this has been removed.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
anderson-joyle: Rollback to original.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
anderson-joyle: Changes applied to EvalVisitor.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
anderson-joyle: Removed.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
anderson-joyle: Removed. Added stack to EvalVisitor.

anderson-joyle:
>                     }

I'll check. This brings the question: why not all `result` is checked this way?

---
In reply to: [1753920608](https://github.com/microsoft/Power-Fx/pull/1910#issuecomment-1753920608) [](http://example.com/codeflow?ancestors=1753920608)

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs:318 in 2695975. [](commit_id = 26959751d64d9c62983efdb3f038b2ac171ae7f9, deletion_comment = False)

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -42,6 +40,10 @@ internal class UserDefinedFunction : TexlFunction
 
         public override bool IsSelfContained => !_isImperative;
 
+        private ReadOnlySymbolTable _symbols;
+
+        public ReadOnlySymbolTable AllSymbols => _symbols;
```
MikeStall, line 45: do we still need?

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -267,16 +272,25 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
...
-                result = await udtf.InvokeAsync(args, _cancellationToken, context.StackDepthCounter.Increment()).ConfigureAwait(false);
-            }
             else if (func is CustomTexlFunction customTexlFunc)
             {
                 // If custom function throws an exception, don't catch it - let it propagate up to the host.
                 result = await customTexlFunc.InvokeAsync(FunctionServices, args, _cancellationToken).ConfigureAwait(false);
             }
+            else if (func is UserDefinedFunction userDefinedFunc)
+            {
+                userDefinedFunc.BindBody(_symbolValues.SymbolTable, new Glue2DocumentBinderGlue(), bindingConfig: BindingConfig.Default);
```
MikeStall, line 282: > userDefinedFunc.BindBody( [](http://example.com/codeflow?start=15&length=26)

we need to move this call up. All binding should be done before we begin eval.

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -189,6 +191,38 @@ public void AddConstant(string name, FormulaValue data)
...
 
             _variables.Add(name, info); // can't exist
+        }
+
+        /// <summary>
+        /// Adds an user defnied function.
+        /// </summary>
+        /// <param name="script">String representation of the user defined function.</param>
+        /// <param name="parseCulture">CultureInfo to parse the script againts.</param>
+        internal void AddUserDefinedFunction(string script, CultureInfo parseCulture)
```
MikeStall, line 201: >AddUserDefinedFunction [](http://example.com/codeflow?start=22&length=22)

bind here. 
Probably should take in a ReadOnlySymbolTable here, which we use to bind the body.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -267,16 +272,25 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
...
+
+                _formulaValuesStack.Push(args);
+
+                (var irnode, _) = userDefinedFunc.GetIRTranslator();
+
+                this.CheckCancel();
+
+                result = await irnode.Accept(this, context).ConfigureAwait(false);
+
+                _formulaValuesStack.Pop();
```
MikeStall, line 292: > _formulaValuesStack.Pop(); [](http://example.com/codeflow?start=15&length=27)

nit - I'd put this in a finally - just in case an exception is thrown.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -43,6 +46,8 @@ internal class EvalVisitor : IRNodeVisitor<ValueTask<FormulaValue>, EvalVisitorC
 
         public Governor Governor { get; private set; }
 
+        private readonly Stack<FormulaValue[]> _formulaValuesStack = new Stack<FormulaValue[]>();
```
MikeStall, line 49: >_formulaValuesStack [](http://example.com/codeflow?start=47&length=19)

this is good. 
Let's take it a step further and wrap in something like UdfStackFrame (see chat)

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -1511,6 +1279,236 @@ public double NextDouble()
             }
         }
 
+        #region Skypped tests.
+        [Fact(Skip = "DefineFunctions method is deprecated")]
```
MikeStall, line 1283: Do these run?

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
MikeStall: We don't do overloads - but we should still test and verify we handle "appropriately"   (either fail, redefine, etc)

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
MikeStall: We should have both - it's good to have at least a few C# tests for UDFs - these are often easier to debug, and give serve as documentation for how to use it. 

But .txts are good for being comprehensive.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -410,307 +409,143 @@ public void FormulaCantRedefine()
...
+                {
+                    var decimalValue = decimal.Parse(expected.ToString());
+                    Assert.Equal(decimalValue, result.ToObject());
+                }
+                else
+                {
+                    Assert.Equal(expected, result.ToObject());
+                }
+            }
+            catch (Exception ex)
```
MikeStall, line 545: >catch (Exception ex) [](http://example.com/codeflow?start=12&length=20)

future PR: we should be able to detect user errors without having to catch an exception (like CheckResult.IsSuccess)

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -410,307 +409,143 @@ public void FormulaCantRedefine()
...
-            var config = new PowerFxConfig()
+                if (isDecimal)
+                {
+                    var decimalValue = decimal.Parse(expected.ToString());
+                    Assert.Equal(decimalValue, result.ToObject());
+                }
+                else
+                {
+                    Assert.Equal(expected, result.ToObject());
+                }
```
MikeStall, line 543: nit - the new FormulaValue.AsDouble() helper may simplify the testing here.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
anderson-joyle: I changed the approach here. Looking forward for your opinion.

jas-valgotar:
>                     }

Correct, it should be, there is already an issue field for it
[https://github.com/microsoft/Power-Fx/issues/1634](https://github.com/microsoft/Power-Fx/issues/1634)

---
In reply to: [1756302329](https://github.com/microsoft/Power-Fx/pull/1910#issuecomment-1756302329) [](http://example.com/codeflow?ancestors=1756302329,1753920608)

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs:318 in d7573a1. [](commit_id = d7573a139aaf1b025a7fd5a76a91052c829ec599, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -148,6 +152,27 @@ public override IEnumerable<StringGetter[]> GetSignatures()
             return new[] { _args.Select<UDFArg, TexlStrings.StringGetter>(key => _ => key.NameIdent.Name.Value).ToArray() };
         }
 
+        public (IntermediateNode topNode, ScopeSymbol ruleScopeSymbol) GetIRTranslator()
+        {
+            return IRTranslator.Translate(_binding);
+        }
+
+        public RecordValue GetRuntimeRecordValue(FormulaValue[] args, CancellationToken cancellationToken)
```
jas-valgotar, line 160: >GetRuntimeRecordValue [](http://example.com/codeflow?start=27&length=21)

can be removed now?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -148,6 +152,27 @@ public override IEnumerable<StringGetter[]> GetSignatures()
             return new[] { _args.Select<UDFArg, TexlStrings.StringGetter>(key => _ => key.NameIdent.Name.Value).ToArray() };
         }
 
+        public (IntermediateNode topNode, ScopeSymbol ruleScopeSymbol) GetIRTranslator()
+        {
+            return IRTranslator.Translate(_binding);
```
jas-valgotar, line 157: >_binding [](http://example.com/codeflow?start=42&length=8)

Can we add null check on binding and add msg, First Call BindBody()

anderson-joyle:
>                     }

I see. Better if we address all these type checks in #1634, IMO. 

---
In reply to: [1757880902](https://github.com/microsoft/Power-Fx/pull/1910#issuecomment-1757880902) [](http://example.com/codeflow?ancestors=1757880902,1753920608)

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs:318 in 2695975. [](commit_id = 26959751d64d9c62983efdb3f038b2ac171ae7f9, deletion_comment = False)

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
anderson-joyle: These were changed to skip while the old code was not yet removed. These test cases are now removed.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -161,12 +186,23 @@ public static INameResolver Create(INameResolver globalNameResolver, IEnumerable
                 return new UserDefinitionsNameResolver(globalNameResolver, args);
             }
 
+            public static INameResolver Create(INameResolver globalNameResolver, IEnumerable<UDFArg> args, FormulaValue[] fvArgs)
+            {
+                return new UserDefinitionsNameResolver(globalNameResolver, args, fvArgs);
+            }
```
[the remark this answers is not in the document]
anderson-joyle: This was a previous attempt of injecting `FormulaValue[]` to `UDFParameterInfo`. It is not longer used. Thanks for catching this.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
anderson-joyle: Removed.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
anderson-joyle: `IRTranslator.Translate` already calls `Contracts.AssertValue(binding);`.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
anderson-joyle: I'll add overloads test in my next PR. I've filed [https://github.com/microsoft/Power-Fx/issues/1940](https://github.com/microsoft/Power-Fx/issues/1940) to address all the pending comments.

src/libraries/Microsoft.PowerFx.Interpreter/UDF/UDFStackFrame.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public UDFStackFrame(UserDefinedFunction func, FormulaValue[] args)
+        {
+            _func = func;
+            _args = args;
+        }
+        
+        private UDFStackFrame()
+        {
+        }
```
MikeStall, line 29: nit - not needed.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -338,6 +299,7 @@ internal DefineFunctionsResult DefineFunctions(IEnumerable<UDFDefinition> udfDef
             return new DefineFunctionsResult(expressionErrors, binders.Select(binder => new FunctionInfo(binder.Function)));
         }
 
+        [Obsolete("This has been deprecated and will soon be deleted.")]
         internal DefineFunctionsResult DefineFunctions(params UDFDefinition[] udfDefinitions)
```
MikeStall, line 303: >DefineFunctions [](http://example.com/codeflow?start=39&length=15)

just delete?

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1912, opened 2023-10-03
Use exact floating point numbers in Text/ExcelCompat tests

CarlosFigueiraMSFT:
We run the .txt tests in different platforms, including those without decimal support. In one of those we were seeing different (off-by-one) result in some tests because the floating point representation of the numbers used in the test was not exact. Since the purpose of the tests was not to test floating point arithmetic, this change updates the tests to use values that have an exact representation in floating point to avoid this problem.


LucGenetier:
✅ No public API change.

jas-valgotar (approved):
:shipit:

PR #1917, opened 2023-10-06
Refactor type determination for output properties that infer type from primary input property

Ananya-Gupta24:
Currently during binding, output properties that infer their type from the primary input property have their type retrieved from the control's template. This PR changes the logic for this determination and handles it similar to pass through input property type calculation


LucGenetier:
✅ No public API change.

lesaltzm (dismissed):

LucGenetier:
✅ No public API change.

lesaltzm (approved):

LucGenetier:
✅ No public API change.

PR #1921, opened 2023-10-09
Move EDate and EOMonth functions to the core library

CarlosFigueiraMSFT:
They were added as 'feature gated' for timing constraints, but it should be part of the core library.


LucGenetier:
✅ No public API change.

gregli-msft (approved):

PR #1925, opened 2023-10-10
Send back empty control tokens

akshar2401:
We previously anticipated wrongly that we do not need to send back empty control tokens. We actually need to do send back empty control tokens and that's the only way we know we went from having atleast one control token in a formula to zero control tokens in the formula. This is a small change that we need to do

### DIFF
--- a/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
+++ b/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
@@ -704,7 +704,7 @@ namespace Microsoft.PowerFx.LanguageServerProtocol
         /// <param name="queryParams">Collection of query params.</param>
         private void PublishControlTokenNotification(ControlTokens controlTokensObj, NameValueCollection queryParams)
         {
-            if (controlTokensObj == null || controlTokensObj.Size() == 0 || queryParams == null)
+            if (controlTokensObj == null || queryParams == null)
             {
                 return;
             }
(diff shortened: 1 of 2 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -704,7 +704,7 @@ private void HandleSemanticTokens<T>(string id, T semanticTokensParams, string m
         /// <param name="queryParams">Collection of query params.</param>
         private void PublishControlTokenNotification(ControlTokens controlTokensObj, NameValueCollection queryParams)
         {
-            if (controlTokensObj == null || controlTokensObj.Size() == 0 || queryParams == null)
```
NavneetThekkumpat, line 707: We could potentially remove the Size method from ControlTokens, since it is not used anymore

NavneetThekkumpat: Also we are initializing controlTokensObj to null and this check here would also block the sending for empty control tokens. Might have to remove the check

akshar2401: It is used in your test so didn't remove it. Wanted to but your test needs it so let's keep. No it won't it is only null when we do range semantic tokens. It is never null when we do full semantic tokens. We only do control highlighting in full semantic tokens

NavneetThekkumpat (approved):

jas-valgotar (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #1926, opened 2023-10-10
Add newline for REPL output

gregli-msft:
The current REPL library doesn't provide whitespace between the expression and the output.  With the REPL here, we have the ability to color the prompt which helps.  to visually group the expression with the output.  But with the PAC CLI, we don't have this ability and it all bleeds together.  This is particularly true for Power Fx where an expression could cause recalc of other values which are rightly shown in the output.

In short, we want to change this:
image

Into the slightly longer, but much easier to parse and read:
image

We could have fixed this just in the PAC CLI, however I prefer this whitespace here too and consistency is good.  This also ensures that table formatting works correctly with added newlines.  It also mirrors the tests cases.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Repl/Services/StandardFormatter.cs
+++ b/src/libraries/Microsoft.PowerFx.Repl/Services/StandardFormatter.cs
@@ -28,6 +28,10 @@ namespace Microsoft.PowerFx.Repl.Services
         // Set to Int32.MaxInt to remove restriction.
         public int MaxTableRows { get; set; } = 10;
 
+        // Append a newline to the end of a formatted table.
+        // If the REPL does not include a newline for the prompt, set this to true for proper spacing.
+        public bool FormatTableNewLine { get; set; } = false;
+
         private string FormatRecordCore(RecordValue record)
         {
             return FormatStandardRecord(record, TryGetSpecialFieldNames(record));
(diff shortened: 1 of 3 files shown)


src/libraries/Microsoft.PowerFx.Repl/Services/StandardFormatter.cs
```diff
@@ -28,6 +28,10 @@ public class StandardFormatter : ValueFormatter
         // Set to Int32.MaxInt to remove restriction.
         public int MaxTableRows { get; set; } = 10;
 
+        // Append a newline to the end of a formatted table.
+        // If the REPL does not include a newline for the prompt, set this to true for proper spacing.
+        public bool FormatTableNewLine { get; set; } = false;
```
MikeStall, line 33: Does anybody ever set this to true?
It would be nice if we didn't need a separate flag here - we're basically double capturing this (in Repl.Prompt, and here).

MikeStall (approved):

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Repl/Services/StandardFormatter.cs
gregli-msft: No, nobody sets the FormulatTableNewLine flag today.  If Prompt didn't include "\n", then it should be set to true.  In theory the formatter could go look at the REPL Prompt property but that seemed like more dependency than this was worth and currently the formatters are a leaf node and unaware of the hosting REPL which is a good thing.

Issue #1929, opened 2023-10-10
REPL will throw an exception for an illegal named formula


gregli-msft:
If an illegal named formula is entered in the REPL, such as `If( Mid("asdf",Value,1) = "a", "X", "Y" )`, then the REPL will throw an exception which is not caught.  In the PAC CLI, this will cause an odd termination as shown below.

Two things should likely happen:
1. Catching the named formula and giving a better error message without exiting.
2. Catching any unhandled exception that happens in the REPL, should these be caught?  

![image](https://github.com/microsoft/Power-Fx/assets/15149773/4d3a1c01-2049-422a-9f57-422ded379db4)

PR #1930, opened 2023-10-10
Catch exceptions for SetFormula in the REPL

gregli-msft:
[REPL will throw an exception for an illegal named formula · Issue #1929 · microsoft/Power-Fx (github.com)](https://github.com/microsoft/Power-Fx/issues/1929).


LucGenetier:
✅ No public API change.

MikeStall (approved):

Issue #1950, opened 2023-10-13
Docs: tracking supported functions per sku 


MikeStall:
Enable manifests where each power-fx host can describe which functions it enables. This can then feed into documentation. 

See also: #1730

MikeStall:
resolved

PR #1951, opened 2023-10-13
Documenting which functions an engine implements. 

MikeStall:
define a .json schema for how an engine can list what it implements. This file gets checked in. 
The idea is that each host would have a .json in their repo. So adding new functions would require updating this file, but that's a healthy friction to create awareness of what functions are included. 

Adds a test helper that hosts can use to verify this json file actually matches their engine. 

See #1950


src/tests/Microsoft.PowerFx.Interpreter.Tests/Docs/InterpreterBase.json
```diff
@@ -0,0 +1,140 @@
+{
+  "MinVersion": "1.3.0-local",
```
MikeStall, line 2: This is a sample file for the interpreter. 
But each host would have their own file, base on their engine's configuration.

LucGenetier:
✅ No public API change.

LucGenetier (approved):

LucGenetier:
✅ No public API change.

gregli-msft (approved):

PR #1955, opened 2023-10-13
add UDF tests

MikeStall:
Add some tests for UDFs. 

BindBody() should only be called once - enforce that.


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -106,6 +111,11 @@ public TexlBinding BindBody(INameResolver nameResolver, IBinderGlue documentBind
                 throw new ArgumentNullException(nameof(documentBinderGlue));
             }
 
+            if (_binding != null)
```
MikeStall, line 114: @VamsiModem

VamsiModem (dismissed):

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -197,13 +197,13 @@ public void AddConstant(string name, FormulaValue data)
...
         /// </summary>
         /// <param name="script">String representation of the user defined function.</param>
-        /// <param name="parseCulture">CultureInfo to parse the script againts.</param>
-        /// <param name="symbolTable">Extra symbols to bind UDF.</param>
-        /// <param name="extraSymbolTable"></param>
-        internal void AddUserDefinedFunction(string script, CultureInfo parseCulture, ReadOnlySymbolTable symbolTable = null, ReadOnlySymbolTable extraSymbolTable = null)
+        /// <param name="parseCulture">CultureInfo to parse the script againts. Default is invariant.</param>
+        /// <param name="symbolTable">Extra symbols to bind UDF. Commonly coming from Engine.</param>
+        /// <param name="extraSymbolTable">Additional symbols to bind UDF.</param>
+        internal void AddUserDefinedFunction(string script, CultureInfo parseCulture = null, ReadOnlySymbolTable symbolTable = null, ReadOnlySymbolTable extraSymbolTable = null)
```
anderson-joyle, line 206: >parseCulture = null [](http://example.com/codeflow?start=72&length=19)

Do we need to handle `parseCulture ==  null`? Something like `parseCulture ?? CultureInfo.InvariantCulture`?
This will hit `TokenizeScript` at some point, which will assign `CultureInfo.CurrentCulture` in case `parseCulture` is null.
[https://github.com/microsoft/Power-Fx/blob/6deed89bf822892620a3d923e6e62f60d1f26d02/src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs#L479](https://github.com/microsoft/Power-Fx/blob/6deed89bf822892620a3d923e6e62f60d1f26d02/src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs#L479)

anderson-joyle (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
MikeStall: set invariant

VamsiModem (approved):

LucGenetier:
✅ No public API change.

PR #1957, opened 2023-10-14
Update required range for exponential notation

CarlosFigueiraMSFT:
We had defined a certain range where the default number-to-text conversion should go from decimal numbers to exponential notation, while leaving some to be implementation-dependent (until 1e10 should use decimal notation; starting from 1e20 should use exponential). In JavaScript, the exponential notation is used starting from 1e**21**, so we can relax the requirement and start requiring exponential notation from that as well.


LucGenetier:
✅ No public API change.

gregli-msft (approved):

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #1958, opened 2023-10-15
[Suggestion] Add public TryGetValue method to recalcengine

kklocker:
Was just playing around with the recalc engine, and noticed that i got an internal null-reference error when trying to fetch a non-existing value using the existing `GetValue` method (see image)
image

This pr just introduces a new method to the RecalcEngine that returns a boolean if it fails to find the value.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
@@ -320,6 +320,24 @@ namespace Microsoft.PowerFx
             return value;
         }
 
+        /// <summary>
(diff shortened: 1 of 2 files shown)


kklocker:

MikeStall (approved):

MikeStall:
Thank you for the contribution! I'm curious what are you using Power Fx for?

kklocker:
@MikeStall Personal project where I am sort of building a hierarchy of models with associated PowerFx formulas :) But also slightly keen on replacing internal expression parsers and evaluators in our applications at work, although I think I have to have a pretty sofisticated POC for that :P

PR #1962, opened 2023-10-16
Fixes IndexOutOfRangeException in Lexer

VamsiModem:
We have been getting a lot of IndexOutOfRangeException exceptions and this PR fixes that.  Not really sure on how to test this, verified this change with Rick and confirmed it's a valid fix.


rick-nguyen (approved):

LucGenetier:
✅ No public API change.

PR #1968, opened 2023-10-18
Deprecate Coercion to Primary Output properties

lesaltzm:
This adds a Feature to allow us to disable coercion to Primary Output Properties. 
This is a feature that only affected Canvas apps, as it depends on the Control type, which does not exist in non-canvas hosts. Unfortunately, to accomplish this, we need to pass `Features` throughout the DType tree so that CoercesTo can have access to this flag.


src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -353,18 +358,21 @@ private static BinderCheckTypeResult CheckTypeCore(IErrorContainer errorContaine
                 return new BinderCheckTypeResult() { Coercions = coercions };
             }
 
-            // If the node is a control, we may be able to coerce its primary output property
-            // to the desired type, and in the process support simplified syntax such as: label1 + slider4
-            IExternalControlProperty primaryOutProp;
-            if (nodeType is IExternalControlType controlType && node.AsFirstName() != null && (primaryOutProp = controlType.ControlTemplate.PrimaryOutputProperty) != null)
+            if (!features.PrimaryOutputPropertyCoercionDeprecated)
```
CarlosFigueiraMSFT, line 361: >PrimaryOutputPropertyCoercionDeprecated [](http://example.com/codeflow?start=26&length=39)

Is this (and the one one line 296) the only significative change in the code (besides passing features instead of just PFxV1)?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
lesaltzm: yup.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (dismissed):
🚢

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

PR #1969, opened 2023-10-18
Fixes AsType() for interpreter

jas-valgotar:
-Legacy analysis requires a hack that changes the return type to include ExpandInfo, this causes the runtime type mismatch in the interpreter. 
-Hence putting it behind a new flag


LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #1972, opened 2023-10-19
DRAFT: add lex helpers

MikeStall:


MikeStall:
want to avoid clients writing code like this: https://github.com/microsoft/power-fx-surveys-sample/blob/main/SurveyEngine/ParseUtility.cs

MikeStall:
Superceded by #2074

PR #1973, opened 2023-10-19
Fix tons of bugs with Signature Help: Duplicates, Wrong Descriptions

akshar2401:
While playing with new Formula Bar, I found a tons of bugs in signature help that went unnoticed for a long time. We need to send correct signatures for PowerFx coders so they are aware of all overloads and their descriptions. This is a larger issue and not just in Canvas but almost all consumptions of Language Server and new formula bar suffer from this. 
Below is a break down of these bugs

**1) Incorrect Descriptions: Due to a faulty implementation of an algorithm to compute signatures in IntellisenseResult.cs, All the overload signatures got the same signature when in fact some of them have different descriptions.** 
Look at the example below (Two overloads of Max function):

**Before**
image

**After**
image

**Before**
image

**After**
image

Above is just an example of Max function. This is the case with many of these fx functions

****2) Duplicate Signatures: Some of the overloads have the same descriptions and same number of parameters with the same names. These overloads are different in the function libraries and due to documentational issues as @jas-valgotar pointed out , we end up with duplicate singatures. To solve this, I am differentiating between signatures by their labels (for example, Max(number, number, number, ... )). and documentation**

**In the below example, in before, Boolean() function has 9  signatures while in after, after deduping all the signatures, we are only down to 6**. Carefully notice that some signatures look the same after deduping but they are not because the descriptions/documentations are different.**

**Before** 
DuplicateSignatures

**After**
DedupSignatures2

**3) Signature/Label was created using the different overload instead of the one for which signature/label is being generated and this resulted in elipsis not showing up when they should:** 

**Before**:
**_Clearly, Max has variable number of arguments and elipsis should be there but it is not. This is because in the picture below we are matching Max(source, expression) overload which only has 2 arguments. We used that overload to create label for Max(number, number, ....) overload. This also points to an important problem. Shouldn't Max(source, expression) be the first and active signature?_**

image

**After**
image


src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/SignatureHelp/SignatureInformation.cs
```diff
@@ -1,14 +1,42 @@
...
+
+        public bool Equals(SignatureInformation other)
+        {
+            if (other == null)
+            {
+                return false;
+            }
+
+            // For now, label is enough to compare two signature information
+            if (Label != other.Label)
```
akshar2401, line 24: I've been flip flopping on the logic here. In an ideal world, we should compare more attributes of the signature to decide whether they are equal or not and use those attributes to compute hash code as well. From my testing, I haven't found the case where two overloads map to the same signature/label but have different documentation/description. If that was the case, then we would need to consider documentation as well. Then there is active parameter and its label and its description. I don't think we need to implement such a complex comparison/hash code computation logic and Label should be fine. I leaned towards keeping it simple for now but let me know if you think i should make this more future proof. This simplicity was influenced by the testing i did and things i saw. But of course, someone with more knowledge about powerfx and its functions and their signatures can chime in here. @jas-valgotar wanted your attention here when you get a chance

akshar2401: We need to also consider Documentation when comparing two signatures. 
I found examples for Boolean() function where two overloads have same signatures but different documentation
Look at the example below
image

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseResult.cs
```diff
@@ -144,14 +147,15 @@ internal IntellisenseResult(IIntellisenseData data, List<IntellisenseSuggestion>
                         funcDisplayString.Append(')');
                         var signatureInformation = new SignatureInformation()
                         {
-                            Documentation = func.Description,
+                            // Use overload description and not IntellisenseData.CurrFunc overload
+                            Documentation = possibleOverload.Description ?? func.Description,
```
jas-valgotar, line 151: >ossibleOverload.Description ?? func.Description, [](http://example.com/codeflow?start=45&length=48)

Can we also add func.Description when the Description is empty string?

src/tests/Microsoft.PowerFx.Core.Tests/IntellisenseTests/TestSignatures/11.json
```diff
@@ -0,0 +1,66 @@
...
+        }
+      ]
+    },
+    {
+      "Label": "Boolean(input)",
+      "Documentation": "Converts a column of text values to a column of boolean values.",
+      "Parameters": [
+        {
+          "Label": "input",
+          "Documentation": ""
```
jas-valgotar, line 19: >"" [](http://example.com/codeflow?start=27&length=2)

@MikeStall this highlights we are missing documentation for functions,
filed this to track: [https://github.com/microsoft/Power-Fx/issues/1975](https://github.com/microsoft/Power-Fx/issues/1975)

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseResult.cs
akshar2401: Not sure if i understand. Am i not doing that already? If possibleOverload.Description is empty then this would fallback to func.Description.

akshar2401: Oh oops never mind. this is null collasceing operator so it won't cover empty string

akshar2401: done

src/tests/Microsoft.PowerFx.Core.Tests/IntellisenseTests/TestSignatures/11.json
akshar2401: Be careful when doing this. If this is assigned a different description then other versions of input for the same function and its overloads then we would need to factor in the description of the active parameter when comparing two signature information.

akshar2401: I've left that out to avoid the complexity here.

LucGenetier:
✅ No public API change.

jas-valgotar (approved):
:shipit:

lesaltzm (approved):

lesaltzm (approved):
Please run the pa client tests locally first and make sure they're passing before checking this in!

akshar2401:
Going to hold off on to this as some tests are failing

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

akshar2401:
> Please run the pa client tests locally first and make sure they're passing before checking this in!

Merging this. I have the fixes ready and my pipeline tests are passing with a private fx build that has these changes

PR #1959, opened 2023-10-16
Support records in [Add/Show/Drop/Rename]Columns functions

CarlosFigueiraMSFT:
Allow those functions to work in records in addition to tables.

PR #1976, opened 2023-10-20
ShowColumns - Support to display names

anderson-joyle:


MikeStall (approved):

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/ShowColumns.txt
```diff
@@ -1,4 +1,4 @@
-﻿#SETUP: SupportColumnNamesAsIdentifiers
+﻿#SETUP: SupportColumnNamesAsIdentifiers, MutationFunctionsTestSetup
 
 >> ShowColumns(Table({a:1,b:2},{a:3,b:4}), b)
```
CarlosFigueiraMSFT, line 3: Instead of adding this new setup handler here, can you please create another test file? Power Apps doesn't support the mutation functions test set up, so that will make it not able to run any of the tests

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/ShowColumns.cs
```diff
@@ -118,7 +118,9 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
                     continue;
                 }
 
-                var columnName = new DName(expectedColumnName);
```
CarlosFigueiraMSFT, line 121: This check should only be done if the support column names as identifiers is true

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/ShowColumns.cs
anderson-joyle: Should we also change this in `DropColumn` function?

CarlosFigueiraMSFT: Yep. I actually have a PR out that will merge the logic for both functions (#1959 ), we should wait until that is in before making this change.

LucGenetier:
✅ No public API change.

anderson-joyle:
This has been fixed by https://github.com/microsoft/Power-Fx/pull/1959.

PR #1977, opened 2023-10-21
add FunctionsList to repl

MikeStall:


CarlosFigueiraMSFT (dismissed):
🚢

src/libraries/Microsoft.PowerFx.Repl/Functions/HelpFunction.cs
```diff
@@ -61,21 +61,13 @@ public class HelpProvider
     {
         public const string FormulaRefURL = "https://aka.ms/Power-Fx-Formula-Reference";
 
-        public static IEnumerable<string> FunctionsList(PowerFxREPL repl)
```
CarlosFigueiraMSFT, line 64: This is also referenced in the REPL's Program.cs file, it will need to be updated to reference the new function.

MikeStall: fixed

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

PR #1978, opened 2023-10-23
Do not remove trailing whitespaces around the string literals in interpolated strings when "Removing Whitespaces/Unformatting"

### LINKED ISSUE #1645: Removing Whitespace removes white spaces between string literals and island edges in interpolated string
Recently found a behavior where unformatting/remove whitespace a formatted expression removes whitespaces between string literal and start of the island { and end of the island and string literal that follows it in an interpolated string. These whitespaces should be consumed in the string literal and therefore should be preserved.
Before: 
image

After:
image

Most likely, 
the bug is here. For normal string literals, both ends have " so Trim() has no effect. String bodies within interpolated strings are tokenized as string literals in LexStringInterpolationBody function in TexlLexer. These string bodies do not necessarily start and end with " like normal string literals do. if these string bodies end with spaces then this Trim would remove them and this is clearly a side effect of trimming here. String bodies extracted from interpolated strings should not be trimmed on both ends. If this is not the case, then it is possible that we do not capture spaces around string bodies in LexStringInterpolationBody. The spans of these string bodies might start after all the spaces on the left and end before the spaces on the right. If this is the case indeed then we need to fix and expand the span to cover the spaces around string bodies in interpolated strings. But still trimming would again remove them. So depending on what is the actual cause, the fix would be either combination of fixing the spans and preventing trimming in case of string bodies or just preventing the trimming.
image

akshar2401:
This pr fixes #1645. The Remove Whitespace functionality used to blindly remove training whitespaces around the string literals inside interpolated strings. This is erroneous behavior as those whitespaces are part of string literals and should not be deleted. For example [dollar symbol]"1 + 2 = {1 + 2}" should stay as [dollar symbol]"1 + 2 = {1 + 2}" but instead becomes [dollar symbol]"1 + 2 ={1 + 2}"

Before:
UnformattingBeforeFix

After:
UnformattingFix

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
@@ -415,6 +415,12 @@ namespace Microsoft.PowerFx.Syntax
             return result;
         }
 
+        /// <summary>
+        /// Given the expression and list of tokens that make up the expression, generates the minified version of the given expression.
+        /// </summary>
+        /// <param name="text">Expression to minify.</param>
+        /// <param name="tokens">List of tokens that make up the given expression.</param>
+        /// <returns>Minified Expression.</returns>
         public string GetMinifiedScript(string text, List<Token> tokens)
         {
             Contracts.AssertValue(text);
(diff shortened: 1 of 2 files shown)


LucGenetier:
✅ No public API change.

akshar2401:
Ran tests in canvas server code and this doesn't break any tests or any internals. Low risk change

jas-valgotar:
Thanks for running tests to verify the impact.

---
In reply to: [1775378623](https://github.com/microsoft/Power-Fx/pull/1978#issuecomment-1775378623) [](http://example.com/codeflow?ancestors=1775378623)

LucGenetier:
✅ No public API change.

Issue #1922, opened 2023-10-09
Text function doesn't handle sections properly for date/time values


CarlosFigueiraMSFT:
The result of `Text(Date(2021,9,8), "yyyy-mm-dd;mm/dd/yyyy")` is currently `"2021-09-08;09/08/2021"` in the C# interpreter, but it should be `"2021-09-08"` (since the date is greater than 0).

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/1992

PR #1992, opened 2023-10-26
Fix handling section for DateTime format in Text

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1922


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_ExcelCompat.txt
```diff
@@ -2084,3 +2084,283 @@ Error({Kind:ErrorKind.InvalidArgument})
...
+"1899-12-30;12/30/1899;December 30 1899"
+
+>> Text(1246.234, "hh:mm:ss.00;hh/mm/ss;hh mm ss")
+"05:36:57.60"
+
+>> Text(-34.56, "hh:mm:ss.00;hh/mm/ss;hh mm ss")
+"10/33/36"
+
+>> Text(0, "hh:mm:ss.00;hh/mm/ss;hh mm ss")
+"00 00 0"
```
CarlosFigueiraMSFT, line 2348: This should be "00 00 00" since we have 2 's', right?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_ExcelCompat.txt
```diff
@@ -2084,3 +2084,283 @@ Error({Kind:ErrorKind.InvalidArgument})
...
+"T:00:00:00.00"
+
+>> Text(Time(14, 32, 25, 678), "\T\:hh:mm:ss.00;\Z\:hh/mm/ss;\K\:hh mm ss")
+"T:14:32:25.67"
+
+>> Text(Time(-5, 23, 15, 356), "\T\:hh:mm:ss.00;\Z\:hh/mm/ss;\K\:hh mm ss")
+"\Z\:19/23/15"
+
+>> Text(Time(0, 0, 0, 0), "\T\:hh:mm:ss.00;\Z\:hh/mm/ss;\K\:hh mm ss")
+"\K\:00 00 0"
```
CarlosFigueiraMSFT, line 2276: This should be "K:00 00 00" : we have 2 's', and the characters K and : are escaped

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_ExcelCompat.txt
```diff
@@ -2084,3 +2084,283 @@ Error({Kind:ErrorKind.InvalidArgument})
...
+>> Text(25, "yyyy-mm-dd\;mm/dd/yyyy\;mmmm dd yyyy")
+"1900-01-24;01/24/1900;January 24 1900"
+
+>> Text(-25, "yyyy-mm-dd\;mm/dd/yyyy\;mmmm dd yyyy")
+"1899-12-05;12/05/1899;December 05 1899"
+
+>> Text(0, "yyyy-mm-dd\;mm/dd/yyyy\;mmmm dd yyyy")
+"1899-12-30;12/30/1899;December 30 1899"
+
+>> Text(1246.234, "hh:mm:ss.00;hh/mm/ss;hh mm ss")
```
CarlosFigueiraMSFT, line 2341: >234 [](http://example.com/codeflow?start=13&length=3)

Can you use some fraction that has an exact floating point representation so that it will have the same result in platforms without decimal (such as Power Apps)? For example, 1246.22265625, which would return 05:20:37.50

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_ExcelCompat.txt
```diff
@@ -2084,3 +2084,283 @@ Error({Kind:ErrorKind.InvalidArgument})
...
+"05:36:57.60"
+
+>> Text(-34.56, "hh:mm:ss.00;hh/mm/ss;hh mm ss")
```
CarlosFigueiraMSFT, line 2344: >-34.56 [](http://example.com/codeflow?start=8&length=6)

Similar to the previous one, can you use a number with exact floating point representation? -34.55859375 would return 13/24/22

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_ExcelCompat.txt
```diff
@@ -2084,3 +2084,283 @@ Error({Kind:ErrorKind.InvalidArgument})
...
+"10/33/36"
```
CarlosFigueiraMSFT, line 2345: >10/33/36 [](http://example.com/codeflow?start=1&length=8)

Negative values are made positive when using the second section, so this time should be after noon (13:26:...)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_ExcelCompat.txt
nguhoa: updated. thanks

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_ExcelCompat.txt
nguhoa: updated. thanks

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_ExcelCompat.txt
nguhoa: updated here as well.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_ExcelCompat.txt
nguhoa: fix logic and update test

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_ExcelCompat.txt
nguhoa: updated.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
🚢

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
🚢

PR #1993, opened 2023-10-26
Move logic to determine record support to FunctionScopeInfo

CarlosFigueiraMSFT:
It was previously defined in TexlFunction, but FunctionScopeInfo is a better place for it.


anderson-joyle (approved):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #1996, opened 2023-10-27
Turn off SA1407 rule

LucGenetier:
This rule should be excluding real obvious operator precedence like `a*b+c`


LucGenetier:
✅ No public API change.

Issue #1924, opened 2023-10-09
Text function: thousands grouping incorrectly applied when preceded by an escaped numeric format char


CarlosFigueiraMSFT:
The result of `Text(12345.34567, "0\#,0")` should be `"1234#,5"` - since the `,` character is not between two numeric format chars, it should not be interpreted as a thousands separator. The function currently returns `"12,34#5"`.

nguhoa:
Fixed in 1997

PR #1997, opened 2023-10-27
Fix handling thousand group with escaping characters

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1924


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Utils/TextFormatUtils.cs
```diff
@@ -120,7 +120,7 @@ public static bool IsValidFormatArg(string formatString, CultureInfo formatCultu
                     mCount = 0;
                 }
 
-                if (formatStr[i] == 'a' || formatStr[i] == 'A')
+                if ((formatStr[i] == 'a' || formatStr[i] == 'A') && (i == 0 || formatStr[i - 1] != '\''))
```
CarlosFigueiraMSFT, line 123: >(i == 0 || formatStr[i - 1] != '\'' [](http://example.com/codeflow?start=68&length=35)

Is this change related to thousands grouping character and escaping?

nguhoa: This is not issue of thousand grouping but when i test it and i found that I only want to do the am pm restrict when it is not in escaping ''. so i just fix here.

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2000, opened 2023-10-30
Add Intellisense in Repl, with TAB, ESC, UP, DOWN keys management

LucGenetier:


src/libraries/Microsoft.PowerFx.Repl/Repl.cs
```diff
@@ -55,7 +56,7 @@ public class PowerFxREPL
         public ParserOptions ParserOptions { get; set; } = new ParserOptions() { AllowsSideEffects = true };
 
         // example override, switching to [1], [2] etc.
-        public virtual string Prompt => "\n>> ";
```
jas-valgotar, line 58: >"\n>> [](http://example.com/codeflow?start=40&length=5)

Why this change? does this mean new prompt will now not start on a new line?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Repl/Repl.cs
LucGenetier: Was creating a mess when rewriting the current line :)

LucGenetier:
✅ No public API change.

jas-valgotar (approved):

LucGenetier:
✅ No public API change.

PR #2003, opened 2023-10-30
IsBlank can operate on controls without coercing

lesaltzm:
This scenario can occur, e.g. if a component has an input property expecting a `Screen` type, and is passed `Blank()`. In that case, IsBlank() is a valid operation on the control type, even if we're not coercing to the primary output property.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IsBlank.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IsBlank.cs
@@ -49,6 +49,9 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
             {
                 if (context.Features.PrimaryOutputPropertyCoercionDeprecated)
                 {
+                    // Even when primary output property coercion is not supported,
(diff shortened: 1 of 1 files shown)


Ananya-Gupta24 (approved):

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IsBlank.cs
```diff
@@ -49,7 +49,7 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
             {
                 if (context.Features.PrimaryOutputPropertyCoercionDeprecated)
                 {
-                    return false;
+                    return true;
```
Shpakh-MSFT, line 52: >true [](http://example.com/codeflow?start=27&length=4)

Can you please add a comment here about it?

LucGenetier:
✅ No public API change.

Ananya-Gupta24 (approved):

Issue #1873, opened 2023-09-22
Intellisense: suggest enums 


MikeStall:
When typing = and the LHS is an option set, we should suggest an option set. :

`ThisRecord.Status =  `

Shoud suggest `'Status (Accounts)'`  . 

A full expression is:

`ThisRecord.Status = 'Status (Accounts)'.Active`

PR #2006, opened 2023-10-31
Enhances optionset intellisense.

jas-valgotar:
Fixes https://github.com/microsoft/Power-Fx/issues/1873

Earlier:
(Filter(Accounts, ThisRecord.Status =  |) suggested all the column names, but now it would only suggest the relevant Option set.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

lesaltzm (approved):

LucGenetier:
✅ No public API change.

Issue #1945, opened 2023-10-12
Text function: block formats where exponential notation has decimals


CarlosFigueiraMSFT:
For an expression such as `Text(13579.34567, "0.00E+0.0")`, Excel does interpret the decimal number after 'E+' and writes out the result as 1.36E+4.0. This is technically correct, but useless, as the value is always a whole number. C# currently has a wrong result: 1.35E+48 (the '8' comes from the '7' after the 135, which is rounded up). To avoid confusion, we should block this (return an error) if such format is passed to the function.

nguhoa:
Another different between c# and excel in exponential notation when it has escaping character. 
If I have the following example that after e+ we have escaping character:
	Text(1234567.1234567, "0.000E+00\k000")
- Excel will get all 0000 after e+ as 0006 and put escaping character in the place after they get 0006 so the result is: 1.235E+00k006
- C# only get 00 before escaping character and after e+. character after escaping character is directly get from input number and the value before e also different so the result is: "1.234E+06k567".
 
But without escaping character both excel and c# behave the same:
	Text(1234567.1234567, "0.000E+00")
return 1.235E+06
 
So we will block this escaping character after exponential notation as well.

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/2007

Issue #1948, opened 2023-10-12
Text function: escaped characters in the exponential notation format 'breaks' the notation


CarlosFigueiraMSFT:
In an expression such as `Text(12345.34567,"0.00E+\@0")`, the C# interpreter returns an incorrect result: 12345.34E+@6 - the escaped character "broke" the exponential notation. We should either fix it to work like Excel (which returns 1.23E@+4), or block this (return an error)

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/2007

Issue #1947, opened 2023-10-12
Text function: block formats where exponential notation is mixed with scaling modifiers


CarlosFigueiraMSFT:
In an expression such as `Text(12345.34567,"0.00E+0%")` we have inconsistent behavior between Excel and C#. Since this is not a realistic scenario, we should just block this and return an error instead.

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/2007

PR #2007, opened 2023-10-31
Block exponential notation with decimals and escaping character

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1945
and https://github.com/microsoft/Power-Fx/issues/1948
and https://github.com/microsoft/Power-Fx/issues/1947


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_ExcelCompat_PowerFxV1CompatDisabled.txt
```diff
@@ -965,4 +965,179 @@ Error({Kind:ErrorKind.InvalidArgument})
 "'long1ateti1e'"
 
 >> Text(Date(1970, 1, 1), DateTimeFormat.LongDateTime24)
-"'long1ateti1e24'"
\ No newline at end of file
+"'long1ateti1e24'"
+
+// Block exponential notation with decimals, escaping character and scaling modifier
```
CarlosFigueiraMSFT, line 970: We don't need to add more tests to this file. Power Fx in C# never runs with V1 disabled, and the difference in the Text function is mostly a runtime difference. Those are interesting tests to be running in Power Apps, but we don't need them here.

nguhoa: Removed the change in this file. thanks

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_ExcelCompat_PowerFxV1Compat.txt
```diff
@@ -1032,4 +1032,179 @@ true
 "Thursday, January 1, 1970 12:00:00 AM"
 
 >> Text(Date(1970, 1, 1), DateTimeFormat.LongDateTime24)
-"Thursday, January 1, 1970 0:00:00"
\ No newline at end of file
+"Thursday, January 1, 1970 0:00:00"
+
+// Block exponential notation with decimals, escaping character and scaling modifier
+>> Text(1234567.1234567, "0.00E+0.0")
+Errors: Error 0-34: The function 'Text' has some invalid arguments.|Warning 22-33: Incorrect format specifier for 'Text'.
```
CarlosFigueiraMSFT, line 1039: Those are all compile-time errors, but it's possible that the compiler won't catch this issue - for example, if we use an expression like the one below. Can you add some of those tests using that as well? No need to do all of them, since the code is shared, but it would be good to get a few (~25%) to ensure coverage on that path as well.

    >> With({fmt:"0.00E+0.0"},Text(1234567.1234567,fmt))

nguhoa: Added these tests. thanks

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

PR #2011, opened 2023-11-01
Implementation of SortByColumns function

CarlosFigueiraMSFT:
The function is defined in the core library, but didn't have an implementation yet. It does now.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -662,6 +664,291 @@ public static async ValueTask<FormulaValue> SortTable(EvalVisitor runner, EvalVi
...
+            var columnNames = new List<string>();
+            var ascendingSort = new List<bool>();
+            for (var i = 1; i < args.Length; i += 2)
+            {
+                var columnName = ((StringValue)args[i]).Value;
+
+                if (!arg0.Type.FieldNames.Contains(columnName))
+                {
+                    return CreateInvalidSortColumnError(irContext, runner.CultureInfo, columnName);
+                }
```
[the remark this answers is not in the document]
CarlosFigueiraMSFT: Unlike all other functions that take columns as argument, where we need to pass the column as a literal string (or identifier, depending on the flag), in SortByColumns we are allowed to pass arbitrary expressions that return a string value. Thus we can't tell the column name at "compile-time" and need to have this additional check at runtime.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -662,6 +664,291 @@ public static async ValueTask<FormulaValue> SortTable(EvalVisitor runner, EvalVi
...
+
+                if (y is BlankValue)
+                {
+                    return -1;
+                }
+
+                if (x is StringValue sv1 && y is StringValue sv2)
+                {
+#pragma warning disable CA1309 // Use ordinal string comparison
+                    return string.Compare(sv1.Value, sv2.Value, StringComparison.CurrentCulture) * modifier;
```
jas-valgotar, line 884: >CurrentCulture [](http://example.com/codeflow?start=81&length=14)

Why CurrentCulture? and not Invariant or Culture from Runtime Config?

jas-valgotar:
Amazing test cases! :fire:

jas-valgotar (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -571,17 +571,38 @@ public virtual bool IsEcsExcemptedLambda(int index)
 
             return false;
         }
-
+
+        public enum ParamIdentifierStatus
```
anderson-joyle, line 575: >public enum ParamIdentifierStatus [](http://example.com/codeflow?start=8&length=33)

Can we have some documentation?

anderson-joyle:
>                     : expectedColumnName);

nit: This seems sharable among other functions.

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/ShowDropColumnsBase.cs:159 in 798446e. [](commit_id = 798446ed8835c76b401191d7a42597adb3824252, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
CarlosFigueiraMSFT: Good point, added

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
CarlosFigueiraMSFT: Great catch, using the culture from the runner.

CarlosFigueiraMSFT:
>                     : expectedColumnName);

Good catch; since it is not related to this PR will create a new one to address that (this one is already quite large).

---
In reply to: [1811528483](https://github.com/microsoft/Power-Fx/pull/2011#issuecomment-1811528483) [](http://example.com/codeflow?ancestors=1811528483)

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/ShowDropColumnsBase.cs:159 in 798446e. [](commit_id = 798446ed8835c76b401191d7a42597adb3824252, deletion_comment = False)

LucGenetier:
✅ No public API change.

jas-valgotar (approved):

PR #2018, opened 2023-11-02
Make Office365Outlook_V4CalendarPostItem test reliable

LucGenetier:
This test could be unpredictable depending on the your local timezone and/or time in the day


LucGenetier:
✅ No public API change.

jas-valgotar (approved):
:shipit:

Issue #2015, opened 2023-11-02
REPL: Errors in script files are hard to track down


gregli-msft:
Consider this script.  There is an error on line 10, where `Descending` should be prefixed with `SortOrder.`

```
// AFC and NFC
afc = [ {name:"chiefs",wins:6}, {name:"chargers",wins:3}, {name:"raiders",wins:3}, {name:"broncos",wins:3} ]
nfc = [ {name:"seahawks",wins:5}, {name:"49ers",wins:5}, {name:"rams",wins:3}, {name:"cardinals",wins:1} ]

// Top from each conference
topAfc = First(Sort(afc,wins,SortOrder.Descending))
topNfc = First(Sort(nfc,wins,SortOrder.Descending))

// Top team overall
top = First(Sort(Table(topAfc,topNfc),wins,Descending))

// Write it out
Notify( $"Top team: {Proper(top.name)} with {top.wins} wins" )
```

Output:

![image](https://github.com/microsoft/Power-Fx/assets/15149773/2f9f33d7-0ab9-4699-8357-35e7c3dcf9e2)

So to what formulas do these errors belong?  We should emit the line number for each error as a prefix to the error.

Issue #2014, opened 2023-11-02
REPL: Script output is still very verbose showing every individual return value, even when echo is off


gregli-msft:
For a script like this one:

```
// AFC and NFC
afc = [ {name:"chiefs",wins:6}, {name:"chargers",wins:3}, {name:"raiders",wins:3}, {name:"broncos",wins:3} ]
nfc = [ {name:"seahawks",wins:5}, {name:"49ers",wins:5}, {name:"rams",wins:3}, {name:"cardinals",wins:1} ]

// Top from each conference
topAfc = First(Sort(afc,wins,SortOrder.Descending))
topNfc = First(Sort(nfc,wins,SortOrder.Descending))

// Top team overall
top = First(Sort(Table(topAfc,topNfc),wins,SortOrder.Descending))

// Write it out
Notify( $"Top team: {Proper(top.name)} with {top.wins} wins" )
```

The result is still very hard to follow:

![image](https://github.com/microsoft/Power-Fx/assets/15149773/20484c85-e91f-4778-89d9-9be741cd12f9)

We should consider not showing the individual result values if echo = false.

PR #2021, opened 2023-11-03
Further REPL improvements

gregli-msft:
Addresses:
- https://github.com/microsoft/Power-Fx/issues/2015
- https://github.com/microsoft/Power-Fx/issues/2014

Without Echo (default if running in the local REPL in this repo), only Notify calls will display - see first run.  All other results will only be shown if Echo is on - see second run.
image

Line numbers are now shown with errors, if the host passes the line number information in.  This is very helpful if processing a file.
image

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Repl/Repl.cs
+++ b/src/libraries/Microsoft.PowerFx.Repl/Repl.cs
@@ -210,7 +210,7 @@ namespace Microsoft.PowerFx
 
             if (expression != null)
             {
-                await this.HandleCommandAsync(expression, lineNumber: lineNumber);
+                await this.HandleCommandAsync(expression, cancel: cancel, lineNumber: lineNumber);
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/libraries/Microsoft.PowerFx.Repl/Repl.cs
```diff
@@ -214,10 +219,13 @@ public async Task HandleLineAsync(string line, CancellationToken cancel = defaul
         /// </summary>
         /// <param name="expression">expression to run.</param>
         /// <param name="cancel">cancellation token.</param>
+        /// <param name="lineNumber">line number to attribute errors to.</param>
         /// <returns>status object with details.</returns>
         /// <exception cref="InvalidOperationException">invalid.</exception>
-        public virtual async Task<ReplResult> HandleCommandAsync(string expression, CancellationToken cancel = default)
+        public virtual async Task<ReplResult> HandleCommandAsync(string expression, CancellationToken cancel = default, uint? lineNumber = null)
```
MikeStall, line 225: is lineNumber the right abstraction?
Is this just for printing source information?

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Repl/Repl.cs
gregli-msft: This is just for printing source information.  Completely optional, default null doesn't change any behavior.

I'm having the repl host pass this in, since it is responsible for pulling the lines out of a file and feeding them to repl one by one.  It also facilitates if the host does some preprocessing, say to pull islands of Power Fx out of a Jupyter like file.

MikeStall (approved):

PR #2023, opened 2023-11-03
Revert "Add Intellisense in Repl, with TAB, ESC, UP, DOWN keys management"

gregli-msft:
Reverts microsoft/Power-Fx#2000

This needs some more work.
- Not designed for REPL library consumers, like the PAC CLI.
- Isn't working for me.
- Use of ESC and TAB keys breaks multiline detection.
- The extra newline in the prompt was there for a reason when used with the PAC CLI and should be retained.


MikeStall (approved):

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

Issue #1946, opened 2023-10-12
Text function: support '#' in the format for the exponential part


CarlosFigueiraMSFT:
For an expression such as `Text(12345.34567,"0.00E+#")`, Excel does the right thing and returns "1.23E+4". C# is incorrect, returning "12345.34E+6" - which is incorrect (the '6' comes from the '5' after .34, which is rounded up).

nguhoa:
In the case of any non-numeric character between exponential notation and numeric character, C# is in correct with excel, it takes input number into the exponential result. So we will block this case as well.

nguhoa:
Fixed in https://github.com/microsoft/Power-Fx/pull/2024

PR #2024, opened 2023-11-03
Handling # after exponential notation

nguhoa:
Fix issue: https://github.com/microsoft/Power-Fx/issues/1946


LucGenetier:
✅ No public API change.

LucGenetier (approved):
:shipit:

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text_ExcelCompat.txt
```diff
@@ -2824,4 +2824,456 @@ Error({Kind:ErrorKind.InvalidArgument})
...
+"12345"
+
+>> Text(12345.34567,"##;(0.00E+00##)")
+"12345"
+
+>> Text(-12345.34567,"##;(0.00E+00##)")
+"(1.23E+004)"
+
+>> Text(0,"(0.00E+0#);##;(0.00E+000###)")
+"(0.00E+0000)"
```
CarlosFigueiraMSFT, line 2991: >0000 [](http://example.com/codeflow?start=8&length=4)

Why is it 4 zeros? The format only specifies 3 required places, and the value (0) is not larger than that.

CarlosFigueiraMSFT: Same issue in many other test cases

nguhoa: This is how excel does. # after 0 count as another 0. Can you recheck ?

CarlosFigueiraMSFT: Oh yeah, Excel is weird :-|... I guess that's right then :-)

PR #2026, opened 2023-11-06
Fix time test

### LINKED ISSUE #2025: TestWithClockService test is failing
The case `TimeZoneOffset()` for the Pacific Standard Time is failing because it assumes that the result is 420 (UTC-07:00), but we are now out of the daylight saving time, so the current value is now 480.

CarlosFigueiraMSFT:
Fixes #2025


LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #2029, opened 2023-11-06
Fix suggestion API for connectors with optional "value-collection"

Robulane:
There is a bug: suggestions are not returned for the connector CardsForPowerApps. This is because of wrong handling of "value-collection".

Per documentation (https://learn.microsoft.com/en-us/connectors/custom-connectors/openapi-extensions) 

"value-collection" is not required. It's a path string that evaluates to an array of objects in the response payload. If value-collection isn't specified, the response is evaluated as an array.

Same applies to value-title and value-path


MikeStall (approved):
:shipit:

LucGenetier (approved):

Issue #2028, opened 2023-11-06
HttpFunctionInvoker should handle blank values


Robulane:
Blank values are "nulls" in PowerFX and it's possible to assign blank to the variable of any type. However, when calling the connector and providing "Blank" values, the call fails with the exceptions below.

Example 1

System.ArgumentException: Expected TableValue and got BlankValue value, for property dataSources
   at Microsoft.PowerFx.Connectors.Execution.FormulaValueSerializer.WriteProperty(String propertyName, OpenApiSchema propertySchema, FormulaValue fv)
   at Microsoft.PowerFx.Connectors.HttpFunctionInvoker.GetBody(String referenceId, Boolean schemaLessBody, Dictionary`2 map, IConvertToUTC utcConverter, CancellationToken cancellationToken)
   at Microsoft.PowerFx.Connectors.HttpFunctionInvoker.BuildRequest(FormulaValue[] args, IConvertToUTC utcConverter, CancellationToken cancellationToken)
   at Microsoft.PowerFx.Connectors.HttpFunctionInvoker.InvokeAsync(IConvertToUTC utcConverter, String cacheScope, FormulaValue[] args, HttpMessageInvoker localInvoker, CancellationToken cancellationToken, Boolean throwOnError)
   at Microsoft.PowerFx.Connectors.ConnectorFunction.InvokeAsync(FormulaValue[] args, BaseRuntimeConnectorContext runtimeContext, CancellationToken cancellationToken)
   
  Example 2:
  [ConnectorFunctionInvoker.cs:InvokeFunctionAsync:183] Error calling or parsing connector response

System.ArgumentException: Expected StringValue and got BlankValue value, for property to
   at Microsoft.PowerFx.Connectors.Execution.FormulaValueSerializer.WriteProperty(String propertyName, OpenApiSchema propertySchema, FormulaValue fv)
   at Microsoft.PowerFx.Connectors.HttpFunctionInvoker.GetBody(String referenceId, Boolean schemaLessBody, Dictionary`2 map, IConvertToUTC utcConverter, CancellationToken cancellationToken)
   at Microsoft.PowerFx.Connectors.HttpFunctionInvoker.BuildRequest(FormulaValue[] args, IConvertToUTC utcConverter, CancellationToken cancellationToken)
[...]

PR #2033, opened 2023-11-06
HttpFunctionInvoker to handle blank values

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2028.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
@@ -115,6 +115,11 @@ namespace Microsoft.PowerFx.Connectors.Execution
 
         private void WriteProperty(string propertyName, OpenApiSchema propertySchema, FormulaValue fv)
         {
+            if (fv is BlankValue)
+            {
+                return;
+            }
+
             if (propertySchema == null)
             {
                 throw new ArgumentException($"Missing schema for property {propertyName}");
@@ -123,13 +128,7 @@ namespace Microsoft.PowerFx.Connectors.Execution
             switch (propertySchema.Type)
             {
                 case "array":
-
-                    if (fv is BlankValue)
-                    {
-                        StartArray(propertyName);
-                        WriteValue(null);
-                    }
-                    else if (fv is TableValue tableValue)
+                    if (fv is TableValue tableValue)
                     {
                         StartArray(propertyName);
 
@@ -183,10 +182,6 @@ namespace Microsoft.PowerFx.Connectors.Execution
                     {
                         WriteDecimalValue(decimalValue.Value);
                     }
-                    else if (fv is BlankValue)
-                    {
-                        WriteDecimalValue(0);
-                    }
                     else
                     {
                         throw new ArgumentException($"Expected NumberValue (number) and got {fv?.GetType()?.Name ?? "<null>"} value, for property {propertyName}");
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
```diff
@@ -123,39 +123,46 @@ private void WriteProperty(string propertyName, OpenApiSchema propertySchema, Fo
...
             {
                 case "array":
-                    // array
-                    if (fv is not TableValue tableValue)
+
+                    if (fv is BlankValue)
                     {
-                        throw new ArgumentException($"Expected TableValue and got {fv?.GetType()?.Name ?? "<null>"} value, for property {propertyName}");
+                        StartArray(propertyName);
+                        WriteValue(null);
```
jas-valgotar, line 130: >WriteValue(null); [](http://example.com/codeflow?start=24&length=17)

Should this be WriteStringValue(string.Empty) ?
Blank Array in json is `[]`, while this gives `[null]` = [Blank()]

src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
```diff
@@ -243,6 +262,10 @@ private void WriteProperty(string propertyName, OpenApiSchema propertySchema, Fo
                     {
                         WriteDateValue(dv.GetConvertedValue(null));
                     }
+                    else if (fv is BlankValue)
+                    {
+                        WriteStringValue(string.Empty);
+                    }
```
jas-valgotar, line 268: Nicee! got all the places! :)

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
```diff
@@ -176,6 +183,10 @@ private void WriteProperty(string propertyName, OpenApiSchema propertySchema, Fo
                     {
                         WriteDecimalValue(decimalValue.Value);
                     }
+                    else if (fv is BlankValue)
+                    {
+                        WriteDecimalValue(0);
```
Robulane, line 188: This can be dangerous. Empty is not the same as 0, Null is not the same as 0. In c# you can have a nullable property int? and having Null and 0 gives two different meaning. The same applies to the boolean false value or empty arrays.

Imho, we should probably just skip writing properties all together

src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
MikeStall, line 130: >     WriteValue(null); [](http://example.com/codeflow?start=19&length=22)

Should blank() mean writing nothing at all? 

{ y:1, x : null } 

vs:
{ y:1 }

src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
MikeStall: agreed - we could do a top-level check for Blank. 

Also - what about ErrorValue?  We probably need to treat that like blank too (I think that's what we do in SQL case, which also can't represent error)

src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
anderson-joyle: Blank will print no property at all.

Robulane (approved):

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

nguhoa (approved):

PR #2034, opened 2023-11-07
Also mark ScopeCollection references as stateful

lesaltzm:
BindKind.Data is used for non-component collections in legacy PA cases, but for components, ScopeCollection is used. This matches the behavior for those so that both are considered stateful and not lifted out of row scopes.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -2995,7 +2995,8 @@ namespace Microsoft.PowerFx.Core.Binding
                 UpdateBindKindUseFlags(lookupInfo.Kind);
 
                 // Update statefulness of global datasources excluding dynamic datasources.
-                if (lookupInfo.Kind == BindKind.Data && !_txb._glue.IsDynamicDataSourceInfo(lookupInfo.Data))
+                if ((lookupInfo.Kind == BindKind.Data && !_txb._glue.IsDynamicDataSourceInfo(lookupInfo.Data)) ||
+                    lookupInfo.Kind == BindKind.ScopeCollection)
                 {
                     _txb.SetStateful(node, true);
                 }
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -2995,8 +2995,7 @@ namespace Microsoft.PowerFx.Core.Binding
                 UpdateBindKindUseFlags(lookupInfo.Kind);
 
                 // Update statefulness of global datasources excluding dynamic datasources.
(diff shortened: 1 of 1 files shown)


LucGenetier:
✅ No public API change.

Ananya-Gupta24 (dismissed):

Shpakh-MSFT (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -2995,7 +2995,8 @@ public override void Visit(FirstNameNode node)
                 UpdateBindKindUseFlags(lookupInfo.Kind);
 
                 // Update statefulness of global datasources excluding dynamic datasources.
-                if (lookupInfo.Kind == BindKind.Data && !_txb._glue.IsDynamicDataSourceInfo(lookupInfo.Data))
+                if ((lookupInfo.Kind == BindKind.Data && !_txb._glue.IsDynamicDataSourceInfo(lookupInfo.Data)) ||
```
Shpakh-MSFT, line 2998: >ookupInfo.Kind == BindKind.Dat [](http://example.com/codeflow?start=22&length=30)

May be use IsInfoKindDataSource

Shpakh-MSFT:
>                         lookupInfo.Kind == BindKind.Data)

Do we need check here as well?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs:4651 in 715f91d. [](commit_id = 715f91d96ec6bbd7265646002d3f37d810ef3b80, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
Shpakh-MSFT: It's for NameInfo but we need some helper like that.

lesaltzm: good call

lesaltzm:
> > ```
> >                     lookupInfo.Kind == BindKind.Data)
> > ```
> 
> Do we need check here as well?
> 
> Refers to: src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs:4651 in [715f91d](https://github.com/microsoft/Power-Fx/commit/715f91d96ec6bbd7265646002d3f37d810ef3b80). [](commit_id = [715f91d](https://github.com/microsoft/Power-Fx/commit/715f91d96ec6bbd7265646002d3f37d810ef3b80), deletion_comment = False)

I don't think so, that looks like it's mutability, which isn't applicable to ScopeCollections (a canvas-only concept)

Shpakh-MSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2035, opened 2023-11-08
[Connectors] Property handle parameter reference

Robulane:
PowerFX doesn't handle the situation when the parameter reference is referencing a specific field in the record. E.g.

            "x-ms-dynamic-list": {
              "operationId": "GetChannelsForGroup",
              "itemsPath": "value",
              "itemValuePath": "id",
              "itemTitlePath": "displayName",
              "parameters": {
                "groupId": {
                  "parameterReference": "body/recipient/groupId",
                  "required": true
                }
              }
            }


MikeStall (approved):
:shipit:

PR #2049, opened 2023-11-16
Refactor logic to retrieve column names depending on features

CarlosFigueiraMSFT:
The logic was used in many functions, moved it to the base class to be reused.


LucGenetier:
✅ No public API change.

anderson-joyle (approved):
:shipit:

Issue #2048, opened 2023-11-16
Implement Search() function


MikeStall:
This  
`Search(Table,  value, column1)`
 
is same as:
`Filter(Table, value in column1)`

1. there is a  Search.txt in NotYetReady:
2. We should use symbol names, not string names, for the columns. 
3. May need to pull the texl function from PAClient

PR #2051, opened 2023-11-17
Implements and move Search()

jas-valgotar:
Fixes https://github.com/microsoft/Power-Fx/issues/2048


src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -961,6 +961,77 @@ public static async ValueTask<FormulaValue> AsType(EvalVisitor runner, EvalVisit
...
                 return new ErrorValue(irContext, e.ExpressionError);
             }
+        }
+
+        public static async ValueTask<FormulaValue> SearchImpl(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
+        {
+            var source = args[0];
+            var textToSearchArg = args[1];
+
+            if (source is BlankValue || source is ErrorValue || textToSearchArg is BlankValue)
```
MikeStall, line 971: >textToSearchArg is BlankValue [](http://example.com/codeflow?start=64&length=29)

?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Search.cs
```diff
@@ -0,0 +1,263 @@
...
+using Microsoft.PowerFx.Core.Functions.Delegation.DelegationMetadata;
+using Microsoft.PowerFx.Core.Localization;
+using Microsoft.PowerFx.Core.Texl.Builtins;
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Syntax;
+
+namespace Microsoft.PowerFx.Core.Texl.Builtins
+{
+    internal sealed class SearchFunction : FunctionWithTableInput
```
MikeStall, line 23: >SearchFunction [](http://example.com/codeflow?start=26&length=14)

this is existing code from PA? Any differences?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -961,6 +961,77 @@ public static async ValueTask<FormulaValue> AsType(EvalVisitor runner, EvalVisit
...
+            {
+                return source;
+            }
+
+            var sourceTable = (TableValue)args[0];
+            var textToSearch = ((StringValue)textToSearchArg).Value.ToLower(runner.CultureInfo);
```
MikeStall, line 977: >((StringValue)textToSearchArg) [](http://example.com/codeflow?start=31&length=30)

what if textToSearch is error?  (did IR pipeline already check that)

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Search.cs
```diff
@@ -0,0 +1,263 @@
...
+            if (!sourceType.GetAllNames(DPath.Root).Any(name => name.Type.Kind == DKind.String))
+            {
+                errors.EnsureError(DocumentErrorSeverity.Severe, args[0], TexlStrings.ErrSearchWrongTableType);
+                fValid = false;
+            }
+
+            for (var i = 2; i < argLen; i++)
+            {
+                var nameArg = args[i];
+                if (!base.TryGetColumnLogicalName(argTypes[0], supportColumnNamesAsIdentifiers, nameArg, errors, out var columnName))
```
CarlosFigueiraMSFT, line 126: >columnName [](http://example.com/codeflow?start=121&length=10)

Also need to validate that the column type is string (the method IsValidSearchableColumnType below is not used).

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Search.cs
```diff
@@ -0,0 +1,263 @@
...
+                string columnName;
+                DType columnType;
+                if (binding.Features.SupportColumnNamesAsIdentifiers)
+                {
+                    columnName = ((FirstNameNode)args[i]).Ident.Name;
+                    columnType = dsType.GetType(new DName(columnName));
+                }
+                else
+                {
+                    columnType = binding.GetType(args[i]);
```
CarlosFigueiraMSFT, line 211: >columnType [](http://example.com/codeflow?start=20&length=10)

We already validated that the column types should be string (in CheckTypes), so there's no need to ask the binder again.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
CarlosFigueiraMSFT, line 971: >source is ErrorValue [](http://example.com/codeflow?start=40&length=20)

This should not be true, as the wrapper would return the error if the argument is an error.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -961,6 +961,77 @@ public static async ValueTask<FormulaValue> AsType(EvalVisitor runner, EvalVisit
...
+
+            if (string.IsNullOrEmpty(textToSearch))
+            {
+                return source;
+            }
+
+            // If no column names are provided, search all columns
+            bool searchAllColumns = args.Length < 3;
```
CarlosFigueiraMSFT, line 985: >bool searchAllColumns = args.Length < 3 [](http://example.com/codeflow?start=12&length=39)

The minArity of the function is set to 3, so this should never be true.

CarlosFigueiraMSFT:
We should add some tests in TexlTest.cs to validate both positive and negative cases for this function.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Search.cs
CarlosFigueiraMSFT: Adding some compilation tests in TexlTests.cs should cover this case.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Search.cs
```diff
@@ -0,0 +1,263 @@
...
+            var args = Contracts.VerifyValue(callNode.Args.Children);
+            int cargs = args.Count;
+
+            FilterOpMetadata metadata = Contracts.VerifyValue(dataSource.DelegationMetadata.FilterDelegationMetadata);
+            for (int i = 2; i < cargs; i++)
+            {
+                if (!IsValidSearchableColumnNode(args[i], binding, dataSource))
+                {
+#pragma warning disable CA1305 // Specify IFormatProvider
+                    var message = string.Format("ColumnSearchable: false");
```
CarlosFigueiraMSFT, line 253: >string.Format("ColumnSearchable: false") [](http://example.com/codeflow?start=34&length=40)

This is the same as `"ColumnSearchable: false"` so there's no need to use string.Format (and then no need to disable/restore the warning)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
CarlosFigueiraMSFT: The StandardErrorHandlingAsync wrapper has already blocked this, we shouldn't get error arguments here.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
jas-valgotar: good point, updated.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Search.cs
```diff
@@ -0,0 +1,273 @@
...
+            for (var i = 2; i < argLen; i++)
+            {
+                var nameArg = args[i];
+                if (!base.TryGetColumnLogicalName(argTypes[0], supportColumnNamesAsIdentifiers, nameArg, errors, out var columnName))
+                {
+                    fValid = false;
+                    continue;
+                }
+
+                var columnType = sourceType.GetType(columnName);
```
CarlosFigueiraMSFT, line 132: >var columnType = sourceType.GetType(columnName) [](http://example.com/codeflow?start=16&length=47)

There is an overload of TryGetLogicalColumnName that also returns the column type, so you can use that and avoid this call.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Search.cs
```diff
@@ -0,0 +1,271 @@
...
+using Microsoft.PowerFx.Core.App.ErrorContainers;
+using Microsoft.PowerFx.Core.Binding;
+using Microsoft.PowerFx.Core.Entities;
+using Microsoft.PowerFx.Core.Entities.QueryOptions;
+using Microsoft.PowerFx.Core.Errors;
+using Microsoft.PowerFx.Core.Functions;
+using Microsoft.PowerFx.Core.Functions.Delegation;
+using Microsoft.PowerFx.Core.Functions.Delegation.DelegationMetadata;
+using Microsoft.PowerFx.Core.Localization;
+using Microsoft.PowerFx.Core.Texl.Builtins;
```
anderson-joyle, line 16: Not used?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
```diff
@@ -1383,6 +1383,17 @@ static Library()
                     returnBehavior: ReturnBehavior.AlwaysEvaluateAndReturnResult,
                     targetFunction: RoundDown)
             },
+            {
+                BuiltinFunctionsCore.Search,
+                StandardErrorHandlingAsync<FormulaValue>(
+                    BuiltinFunctionsCore.Search.Name,
+                    expandArguments: NoArgExpansion,
+                    replaceBlankValues: NoOpAlreadyHandledByIR,
+                    checkRuntimeTypes: ShowColumnsTypeChecker,
```
CarlosFigueiraMSFT, line 1392: >ShowColumnsTypeChecker [](http://example.com/codeflow?start=39&length=22)

The ShowColumnsTypeChecker will allow a record in the first argument, which we don't allow in Search.

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
```diff
@@ -3620,6 +3620,54 @@ public void TestAttachmentDottedNameNodeBinding(string script, string expectedSc
...
+        {
+            var symbol = new SymbolTable();
+            symbol.AddVariable("T", new TableType(TestUtils.DT("*[Name:s]")));
+            symbol.AddVariable("T2", new TableType(TestUtils.DT("*[Name:s,Age:n,Address:s]")));
+
+            TestSimpleBindingSuccess(
+                script,
+                TestUtils.DT(expectedType),
+                symbol,
+                new Features { ConsistentOneColumnTableResult = true });
```
CarlosFigueiraMSFT, line 3639: >ConsistentOneColumnTableResult = true [](http://example.com/codeflow?start=31&length=37)

There's no need to use this feature as we don't have any functions in the tests that depend on it.

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
```diff
@@ -3620,6 +3620,54 @@ public void TestAttachmentDottedNameNodeBinding(string script, string expectedSc
             symbol.AddEntity(new TestDataSource("DS", schema));
 
             TestSimpleBindingSuccess(script, expectedType, symbol);
+        }
+
+        [Theory]
+        [InlineData("Search(T, \"a\", \"Name\")", "*[Name:s]")]
```
CarlosFigueiraMSFT, line 3626: >\"Name\" [](http://example.com/codeflow?start=38&length=8)

We should have some tests with column names as identifiers as well.

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
```diff
@@ -3620,6 +3620,54 @@ public void TestAttachmentDottedNameNodeBinding(string script, string expectedSc
...
+                new Features { ConsistentOneColumnTableResult = true });
+        }
+
+        [Theory]
+
+        // missing 3rd arg
+        [InlineData("Search(T, \"a\")", "*[]")]
+
+        // Age is number, only string supported.
+        [InlineData("Search(T2, \"a\", \"Age\"", "*[Address:s, Age:n, Name:s]")]
```
CarlosFigueiraMSFT, line 3648: >" [](http://example.com/codeflow?start=46&length=1)

Missing closing parenthesis (here and in the next few tests)

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
```diff
@@ -3620,6 +3620,54 @@ public void TestAttachmentDottedNameNodeBinding(string script, string expectedSc
...
+        [InlineData("Search(T3, \"a\", \"Age\"", "*[Age:n]")]
+
+        // 3rd onwards Arg should always be constant string.
+        [InlineData("Search(T2, \"a\", If(1<0, \"Name\")", "*[Address:s, Age:n, Name:s]")]
+
+        // 2nd Arg should only be a string.
```
CarlosFigueiraMSFT, line 3654: >// 2nd Arg should only be a string. [](http://example.com/codeflow?start=8&length=35)

Should we allow coercion in the second argument? In all functions that expect a string value we let other primitive types to be passed.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
```diff
@@ -3620,6 +3620,100 @@ public void TestAttachmentDottedNameNodeBinding(string script, string expectedSc
...
+                script,
+                TestUtils.DT(expectedType),
+                symbol);
+        }
+
+        [Theory]
+        [InlineData("Search(T, \"a\", Name)", "*[Name:s]")]
+        [InlineData("Search(T, If(1<0,\"a\"), Name)", "*[Name:s]")]
+        [InlineData("Search(T2,\"a\", Name, Address)", "*[Name:s,Age:n,Address:s]")]
+        public void TexlFunctionTypeSemanticsSearch_StronglyTypedBuiltinEnums(string script, string expectedType)
```
CarlosFigueiraMSFT, line 3645: >StronglyTypedBuiltinEnums [](http://example.com/codeflow?start=52&length=25)

`SupportColumnNamesAsIdentifiers` - not strongly typed enums

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
```diff
@@ -3620,6 +3620,100 @@ public void TestAttachmentDottedNameNodeBinding(string script, string expectedSc
...
+
+        // 3rd onwards Arg should always be constant string.
+        [InlineData("Search(T2, \"a\", If(1<0, Name))", "*[Address:s, Age:n, Name:s]")]
+
+        // 2nd Arg should only be a string.
+        [InlineData("Search(T2, 1, Name)", "*[]")]
+
+        // missing field.
+        [InlineData("Search(T2, \"a\", Name, Address, 'Does not exist')", "*[Address:s, Age:n, Name:s]")]
+        public void TexlFunctionTypeSemanticsSearch_StronglyTypedBuiltinEnums_Negative(string script, string expectedSchema)
```
CarlosFigueiraMSFT, line 3705: >StronglyTypedBuiltinEnums [](http://example.com/codeflow?start=52&length=25)

`SupportColumnNamesAsIdentifiers` - not strongly typed enums

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Search.cs
jas-valgotar: thanks for pointing that out, updated.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Search.cs
jas-valgotar: some difference just to support column name as id. Largely the same!

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
jas-valgotar: We could!
I am not worried abt semantic change here as its widening the restriction.
Just concerned as to end users may end up getting tricked in it if they don't fully grasp the coercion!
e.g. `Search(table, Date(), column,..)` would start working and they would need to know what exactly Text(Date()) would actually look like...
cc @MikeStall any thoughts?

jas-valgotar:
Currently we don't show Intellisense for this functions. I have filed [https://github.com/microsoft/Power-Fx/issues/2061](https://github.com/microsoft/Power-Fx/issues/2061) to track it.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Search.cs
jas-valgotar: Added texl tests.

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
jas-valgotar: Added.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs
jas-valgotar: Makes sense, Thanks for pointing it out.

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

CarlosFigueiraMSFT (approved):
:shipit:

PR #2057, opened 2023-11-18
Prevent assert on DType.TryGetPath

CarlosFigueiraMSFT:
A call to TryGetPath was asserting on a valid invocation (last case in the newly added tests). It shouldn't assert - it should just return false in that case.


LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #2059, opened 2023-11-20
move docs schema to Core so that it's more accessible

MikeStall:


LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

Issue #2052, opened 2023-11-17
Stack trace for post item failure


anderson-joyle:
Stack trace for post item failure:

`office365Outlook.V4CalendarPostItem("Calendar", "Test Event", DateTime(2023, 11, 13, 11, 00, 00), DateTime(2023, 11, 13, 11, 30, 00), "PST")`

```
System.ArgumentException: Expected StringValue and got ErrorValue value, for property start
   at Microsoft.PowerFx.Connectors.Execution.FormulaValueSerializer.WriteProperty(String propertyName, OpenApiSchema propertySchema, FormulaValue fv)
   at Microsoft.PowerFx.Connectors.HttpFunctionInvoker.GetBody(String referenceId, Boolean schemaLessBody, Dictionary`2 map, IConvertToUTC utcConverter, CancellationToken cancellationToken)
   at Microsoft.PowerFx.Connectors.HttpFunctionInvoker.BuildRequest(FormulaValue[] args, IConvertToUTC utcConverter, CancellationToken cancellationToken)
   at Microsoft.PowerFx.Connectors.HttpFunctionInvoker.<InvokeAsync>d__10.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.PowerFx.Connectors.ConnectorFunction.<InvokeInternalAsync>d__93.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.PowerFx.Connectors.ConnectorTexlFunction.<InvokeAsync>d__31.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.PowerFx.EvalVisitor.<Visit>d__36.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

PR #2062, opened 2023-11-20
Handling date-no-tz type as datetime

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2052.

This fix does not address the exception since https://github.com/microsoft/Power-Fx/pull/2033 was fixed previously.
This PR addresses a new situation when running the expression described on this issue.


MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #2065, opened 2023-11-21
Update logic to determine whether a LookUp call with a reduction formula argument can be delegated

ian-legler:
Currently, the logic to determine whether a LookUp function call with a reduction formula argument is incorrect, purely checking whether the node is marked as NOT delegatable.

This means that most scenarios are marked delegable, but uses of the LookUp function as part of the filter in a delegatable function (e.g. LookUp, Filter, etc.) can lead to incorrect code gen.

Updating the logic to try to determine whether the full function can be delegated, in context.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/FilterDelegationBase.cs
```diff
@@ -64,7 +64,7 @@ public override bool TryGetDelegationMetadata(CallNode node, TexlBinding binding
             return true;
         }
 
-        protected bool IsValidDelegatableFilterPredicateNode(TexlNode dsNode, TexlBinding binding, FilterOpMetadata filterMetadata, bool generateHints = true)
+        protected bool IsValidDelegatableFilterPredicateNode(TexlNode dsNode, TexlBinding binding, FilterOpMetadata filterMetadata, bool generateHints = true, bool enforceBoolean = true)
```
lesaltzm, line 67: Maybe a comment about when enforceBoolean should be false?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Lookup.cs
```diff
@@ -80,7 +82,32 @@ public override bool IsServerDelegatable(CallNode callNode, TexlBinding binding)
...
 
-            FilterOpMetadata metadata = null;
+            if (!TryGetFilterOpDelegationMetadata(callNode, binding, out var metadata))
+            {
+                return false;
+            }
+
+            var args = callNode.Args.Children.VerifyValue();
+
+            // without lookup reduction delegation, follow the  legacy logic to determine if the function call is delegatable
```
lesaltzm, line 92: nit: comments start with an initial capital letter

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Lookup.cs
```diff
@@ -2,13 +2,15 @@
 // Licensed under the MIT license.
 
 using System.Collections.Generic;
+using System.Globalization;
```
lesaltzm, line 5: unused using?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Lookup.cs
```diff
@@ -101,25 +128,54 @@ public override bool IsServerDelegatable(CallNode callNode, TexlBinding binding)
...
+        {
+            if (!TryGetFilterOpDelegationMetadata(callNode, binding, out var metadata))
             {
                 return false;
             }
 
-            return IsValidDelegatableFilterPredicateNode(args[1], binding, metadata);
+            // use a variation of the filter predicate logic to determine if the reduction formula is delegatable, without enforcing the return type must be boolean
+            return IsValidDelegatableFilterPredicateNode(reductionNode, binding, metadata, generateHints: false, enforceBoolean: false);
         }
```
lesaltzm, line 154: Maybe I misunderstood, but I thought we had said that the reduction formula was going to always be evaluated locally?
Why would it share anything with the filter predicate logic?

ian-legler: When we had our discussion, I had missed a critical element.  The primary issue is that the LookUp can be used with a reduction formula inside another Filter or LookUp call.  We properly delegate trivial reduction formulas in that scenario today, so we can't break that.  My proposed fix is to use a flavor of the same logic that determines whether a filter predicate is delegable to determine if the reduction can be code gen'd properly

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Lookup.cs
```diff
@@ -101,25 +129,54 @@ public override bool IsServerDelegatable(CallNode callNode, TexlBinding binding)
...
-            if (args.Count > 2 && binding.IsDelegatable(args[2]))
-            {
-                SuggestDelegationHint(args[2], binding);
-                return false;
-            }
+            return true;
+        }
 
-            if (args.Count < 2)
+        public override bool IsEcsExcemptedLambda(int index)
```
CarlosFigueiraMSFT, line 135: >Excempted [](http://example.com/codeflow?start=34&length=9)

Typo - `Exempted`. Since this is defined in the base class it's ok to leave it like that, but can you at least fix the comment?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/FilterDelegationBase.cs
```diff
@@ -64,7 +64,10 @@ public override bool TryGetDelegationMetadata(CallNode node, TexlBinding binding
             return true;
         }
 
-        protected bool IsValidDelegatableFilterPredicateNode(TexlNode dsNode, TexlBinding binding, FilterOpMetadata filterMetadata, bool generateHints = true)
+        // Determine whether a node can be delegated as part of a filter predicate.
+        // The enforeceBoolean flag determines whether to enforce the return type of the node.  If the node is part of a filter predicate directly, it must return a boolean type.
+        // If the node is used in other places, such as a LookUp reduction formula inside a filter, it can return any type.
```
CarlosFigueiraMSFT, line 69: >LookUp reduction formula inside a filter [](http://example.com/codeflow?start=58&length=40)

Suggestion: '... such as a reduction / projection formula inside a LookUp'. There is no reduction formula in a filter, only in a lookup.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Lookup.cs
```diff
@@ -80,7 +83,32 @@ public override bool IsServerDelegatable(CallNode callNode, TexlBinding binding)
...
 
-            FilterOpMetadata metadata = null;
+            if (!TryGetFilterOpDelegationMetadata(callNode, binding, out var metadata))
+            {
+                return false;
+            }
+
+            var args = callNode.Args.Children.VerifyValue();
+
+            // Without lookup reduction delegation, follow the  legacy logic to determine if the function call is delegatable
```
CarlosFigueiraMSFT, line 93: >   [](http://example.com/codeflow?start=62&length=2)

[minor] two spaces --> 1 space

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Lookup.cs
ian-legler: This looks like a merge artifact

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/FilterDelegationBase.cs
ian-legler: Clarified the scenario in the comment

LucGenetier:
✅ No public API change.

lesaltzm (approved):

LucGenetier:
✅ No public API change.

PR #2066, opened 2023-11-21
Fix logic to determine selects for a some functions with column names as identifiers

CarlosFigueiraMSFT:
The logic to detect which columns need to be selected from data source was not working properly in a few functions if column names as identifiers were used. This fixes it, and adds a test with a data source to the core test library.


MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

MikeStall (dismissed):

LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #2071, opened 2023-11-27
Fix delegation validation for Search function

CarlosFigueiraMSFT:
The logic to determine the delegability of the Search function does not work properly with column names as identifiers. This fixes it.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

lesaltzm (approved):

PR #2074, opened 2023-11-29
TextFirst parser mode

gregli-msft:
An alternative parser mode for hosts that wish to start in a "text first" manner:

image

The semantic rules:
- The input buffer is in general treated as a text string.
- There is no need to escape `"`, `{`, or `}`.
- String interpolation are supported, starting with `${` rather than the normal `{`.  This is to allow `{` to be used without interpretation.  String interpolation can be nested by using standard string interpolation within.
- A single `=` starts a normal formula.


LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.ParserOptions.#ctor(System.Globalization.CultureInfo,System.Boolean,System.Int32)

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -970,13 +980,35 @@ public LexerImpl(TexlLexer lex, string text, StringBuilder sb, Flags flags)
...
+                    }
+
+                    if (ich < _charCount && text[ich] == '=')
+                    {
+                        // skip '=' and treat the rest as a normal formula
+                        CurrentPos = ich + 1;
+                    }
+                    else
+                    {
+                        // CurrentPos doesn't move (stays at start of buffer) and treat all of it as text first
```
CarlosFigueiraMSFT, line 1004: >CurrentPos doesn't move (stays at start of buffer) [](http://example.com/codeflow?start=27&length=50)

Can you add a test where we have whitespaces in the beginning? Something like

    >>      hello
    "     hello"

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -970,13 +980,35 @@ public LexerImpl(TexlLexer lex, string text, StringBuilder sb, Flags flags)
...
+                        _initialLexerMode = LexerMode.TextFirst;
+                    }
+                }
             }
 
             // If the mode stack is empty, this is already an parse, use NormalMode as a default
```
CarlosFigueiraMSFT, line 1010: >NormalMode [](http://example.com/codeflow?start=73&length=10)

update comment

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
gregli-msft: Done.  I couldn't add it in the .txt files as we strip the leading whitespace when running the tests, but I added it in C# tests.

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
gregli-msft: I don't understand this comment, can you say more?  I believe this logic is correct, in the case identified it should be TextFirst mode.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
CarlosFigueiraMSFT: The comment says that we use "Normal" mode as a default; after your changes we will use the initial lexer mode (normal or text first). The logic is correct, just the comment is a little stale

gregli-msft: Done.  Sorry, what a GitHub newbie I am.  I thought you were referring to the green changed code above.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
🚢

PR #2075, opened 2023-11-29
Remove unused member from DateAddFunction class

CarlosFigueiraMSFT:
The SubDayStringList field is not used anywhere, can be removed.


LucGenetier:
✅ No public API change.

anderson-joyle (approved):
:shipit:

PR #2077, opened 2023-11-30
Add tests for IfError to cover legacy bug

CarlosFigueiraMSFT:
Found a bug in IfError in Power Apps, which uses the legacy (pre-Power Fx 1.0) logic. It may be risky to fix it, so just adding some tests to document it.


LucGenetier:
✅ No public API change.

anderson-joyle (changes requested):
:clock1:

anderson-joyle (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2078, opened 2023-11-30
Prune expected returntype intellisense.

jas-valgotar:
Expected returntype can have Dataverse type as expected return type. And that can cause intellisense to explode as that type can potentially point to different DV Tables.

This PR mitigates that by limiting intellisense to top-level fields, so intellisense never drills into nested fields. That helps us with the speed and out primary use case for the expected return type is to suggest top-level field anyways.


MikeStall (approved):

LucGenetier:
✅ No public API change.

Issue #2056, opened 2023-11-18
[pac pfx] More information when getting runtime type mismatches


scottdurow:
If I use the following with `pac pfx`
```
With({
    contact:LookUp(Contacts,Email="someone_j@example.com")
    },
    Collect(Accounts,
        {
            'Account Name':"Test",
            'Primary Contact':contact,
            'Owner':contact.'Created By'
        }
    )
);
```
I get the error `<Error: Runtime type mismatch>`

Could there be a little bit more information returned (e.g. Expected type Polymorphic Lookup)
Also - what is the correct way of setting this polymorphic lookup?

jas-valgotar:
That does look like a bug,
As a workaround, I checked the expression without With() works fine!
Ie.
```
Collect(Accounts,
        {
            'Account Name':"Test",
            'Primary Contact':LookUp(Contacts,Email="someone_j@example.com"),
            'Owner':LookUp(Contacts,Email="someone_j@example.com").'Created By'
        }
)
```

PR #2079, opened 2023-11-30
Change polymorphic union

jas-valgotar:
Fixes https://github.com/microsoft/Power-Fx/issues/2056

The union should be Polymorphic, as this union is needed by the mutation function and when the mutation function receives the record after updating, that runtime field will become polymorphic again.


src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -2828,7 +2828,7 @@ private static DType UnionCore(ref bool fError, DType leftType, DType rightType,
                     ((field1Type == Polymorphic && field2Type.IsRecord) ||
                     (field2Type == Polymorphic && field1Type.IsRecord)))
                 {
-                    fieldType = field1Type == Polymorphic ? field2Type : field1Type;
+                    fieldType = Polymorphic;
```
CarlosFigueiraMSFT, line 2831: >Polymorphic [](http://example.com/codeflow?start=32&length=11)

Since you're working with polymorphic types, we should add tests for the AsType function. Currently we don't have any tests for that.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
jas-valgotar: I do have some on DV repo https://github.com/microsoft/Power-Fx-Dataverse/blob/main/src/PowerFx.Dataverse.Tests/PluginExecutionTests/Functions/AsTypeFunctionTests.cs
I will see if I can add them here...

MikeStall (approved):

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
jas-valgotar: Will add them in seprate PR
https://github.com/microsoft/Power-Fx/issues/2082

Issue #2083, opened 2023-12-01
ShowColumns() doesn't work properly with Display Name Identifier.


jas-valgotar:
`First(ShowColumns(Accounts, 'Account Name')).'Account Name'` should work. 
but it only allows logical names, Ie forces to use `First(ShowColumns(Accounts, 'Account Name')).name`

PR #2084, opened 2023-12-01
Disables single column access syntax.

jas-valgotar:
This disables single column access syntex, I.e. Accounts.'Account Name'.
Alternatively, users can still use ShowColumns() e.g. ShowColumns(Accounts, 'Account Name')


src/tests/Microsoft.PowerFx.Core.Tests/DisplayNameTests.cs
```diff
@@ -202,16 +202,14 @@ public void DisableDisplayNames()
         [InlineData("If(DisplayB, DisplayNum, 1234)", "If(DisplayB, DisplayNum, 1234)", true)]
         [InlineData("If(DisplayB, Num, 1234)", "If(DisplayB, DisplayNum, 1234)", true)]
         [InlineData("Sum(Nested, Inner)", "Sum(NestedDisplay, InnerDisplay)", true)]
-        [InlineData("Sum(Nested /* The source */ , Inner /* Sum over the InnerDisplay column */)", "Sum(NestedDisplay /* The source */ , InnerDisplay /* Sum over the InnerDisplay column */)", true)]        
-        [InlineData("First(Nested.Inner).Inner", "First(NestedDisplay.InnerDisplay).InnerDisplay", true)]
```
jas-valgotar, line 206: >[InlineData("First(Nested.Inner).Inner", "First(NestedDisplay.InnerDisplay).InnerDisplay", true)] [](http://example.com/codeflow?start=8&length=97)

Filed [https://github.com/microsoft/Power-Fx/issues/2083](https://github.com/microsoft/Power-Fx/issues/2083)

LucGenetier:
✅ No public API change.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5453,4 +5453,8 @@
     <value>Table doesn't contain any column of text type.</value>
     <comment>Error Message.</comment>
   </data>
+  <data name="ErrDeprecatedDotUseShowColumn" xml:space="preserve">
+    <value>Deprecated use of '.'. Instead use ShowColumn().</value>
+    <comment>Error Message.</comment>
```
CarlosFigueiraMSFT, line 5458: >. [](http://example.com/codeflow?start=26&length=1)

Please add `{Locked=ShowColumns}` to prevent it from being translated.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5453,4 +5453,8 @@
     <value>Table doesn't contain any column of text type.</value>
     <comment>Error Message.</comment>
   </data>
+  <data name="ErrDeprecatedDotUseShowColumn" xml:space="preserve">
+    <value>Deprecated use of '.'. Instead use ShowColumn().</value>
```
CarlosFigueiraMSFT, line 5457: >Instead use ShowColumn() [](http://example.com/codeflow?start=34&length=24)

Suggestion: 'Please use the 'ShowColumns' function instead'.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -5453,4 +5453,8 @@
     <value>Table doesn't contain any column of text type.</value>
     <comment>Error Message.</comment>
   </data>
+  <data name="ErrDeprecatedDotUseShowColumn" xml:space="preserve">
```
CarlosFigueiraMSFT, line 5456: >ShowColumn [](http://example.com/codeflow?start=33&length=10)

[nit] the function name is ShowColumns (with an 'S' at the end), we should be consistent.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -3338,6 +3338,14 @@ public override void PostVisit(DottedNameNode node)
                     return;
                 }
 
+                // Disable SingleColumnsAccess syntax. I.e. [1,2,3].Values.
```
CarlosFigueiraMSFT, line 3341: >[1,2,3].Values [](http://example.com/codeflow?start=60&length=14)

We should give a better example, such as `[{a:1,b:2},{a:3,b:4]].a`. Same with the example in the line below.

src/tests/Microsoft.PowerFx.Core.Tests/BindingEngineTests.cs
```diff
@@ -220,10 +220,10 @@ public void CheckDottedBindError()
         {
             var config = new PowerFxConfig();
             var engine = new Engine(config);
-            var result = engine.Check("[1,2,3].foo");
+            var result = engine.Check("First([1,2,3]).foo");
```
CarlosFigueiraMSFT, line 223: >First([1,2,3]).foo [](http://example.com/codeflow?start=39&length=18)

It may be good to have the existing test here, just validating the new error message.

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
```diff
@@ -3859,6 +3859,35 @@ public void TestSilentValidDelegatableFilterPredicateNode(string script, bool wa
...
                 symbol.RemoveFunction(silentFilterFunction);
             }
+        }
+
+        [Theory]
+        [InlineData("T.A", "*[A:n]", true)]
+        [InlineData("T.B", "*[B:![C:s]]", true)]
+        [InlineData("T.A", "e", false)]
+        [InlineData("T.B", "e", false)]
+        public void TestSignleColumnAccessDependentOnTableType(string script, string expectedSchema, bool isBindingSuccess)
```
CarlosFigueiraMSFT, line 3869: >Signle [](http://example.com/codeflow?start=24&length=6)

[typo] Single

CarlosFigueiraMSFT, line 3869: >isBindingSuccess [](http://example.com/codeflow?start=106&length=16)

Instead of specifying here what we expect, consider calling this parameter 'usePowerFxV1Rules' or something similar, to make it clearer. And in the validation you can check that if PFxV1, then error, otherwise success.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -749,22 +749,22 @@ public void CheckLambdaBindError()
         public void CheckDottedBindError()
```
CarlosFigueiraMSFT, line 749: >CheckDottedBindError [](http://example.com/codeflow?start=20&length=20)

This is the same test as the one in BindingEngineTests.cs in core. Do we need both?

jas-valgotar: Its because engines are different, engine vs recalcengine.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
jas-valgotar: updated thanks!

src/strings/PowerFxResources.en-US.resx
jas-valgotar: updated, wasn't aware of this {Locked=} syntax. Thanks for pointing it out.

src/tests/Microsoft.PowerFx.Core.Tests/BindingEngineTests.cs
jas-valgotar: Adding a new test, keeping the existing one.

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
jas-valgotar: oops

CarlosFigueiraMSFT (approved):
:shipit:

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #2090, opened 2023-12-05
Remove resources only used in Power Apps

CarlosFigueiraMSFT:
Some resources from Power Fx are only used in Power Apps, so they can be moved to PA's repository instead of this one here. This cleans up some PA-specific enum resources.


LucGenetier:
✅ No public API change.

rick-nguyen (approved):

PR #2094, opened 2023-12-07
Allow variables in SortByColumns arguments

CarlosFigueiraMSFT:
The current SortByColumns implementation does not allow anything other than identifiers from being used as column names. In Power Apps we allow arbitrary expressions (literal strings or other expressions that return strings) from being used, so we will allow this here as well.

This also defines the legacy behavior for SortByColumns: it will only support column names as identifiers if that feature ("SupportColumnNamesAsIdentifiers") is enabled, *and* PowerFxV1Compatibility rules is also enabled - unlike other functions that take column names, which only look at the first feature.


LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT:
Closing this one, opening as #2095

PR #2096, opened 2023-12-09
Update Void accept behavior and take advantage in Repl

gregli-msft:
Several Power Fx functions have no real return value.  To date, Canvas apps have returned Boolean values that aren't seen and likely this behavior is not well known.  With the Repl, these values are being printed as part of the stream and it can be confusing. to see `Set( x, 3 )` and then `true`.  Similarly, it is seems odd in a strongly typed language to return a value that really has no meaning (the `true` from Set) which could be a coding error, the kind of thing that strong typing should catch.

Changes to be made:
1. Void will accept all types, including Void.  In Power Apps, behavior properties would be of type Void, and they work with any formula, dropping the computed value.
2. Void will not be accepted by any other types (except Void).  Trying to use the output from Set will result in an error.
3. The following functions will return Void: Set, Clear, Remove, Notify (Repl), Help (Repl), Exit (Repl).  Future functions that have no return value will return Void.

There is a tiny loss of functionality with this change.  The result from Remove can no longer be put in a variable.  Which normally doesn't matter as Remove returns no value.  But, if Remove encounters an error, that error can no longer be put in a variable.  The error will still be reported and can be caught and processed with IfError and App.OnError in Power Apps.  With an IfError, the error record can be extracted and that can be put in a variable.  The team discussed this and we do not believe this is a common pattern, if used at all.

With this change, it is also not possible to use Boolean logic to short circuit the use of these functions.  For example, it used to be possible to `Back() && Clear( table )` and the Clear would only be executed if the Back() returned true.  We don't think the intent here is very clear, especially for a non-pro developer, and we'd prefer it be written as `If( Back(), Clear( table )`.  Usage on Canvas apps is very low for this pattern.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -121,7 +121,7 @@ public ISymbolSlot AddVariable(string name, FormulaType type, SymbolProperties p
 
             if (type is Types.Void)
             {
-                throw new NotSupportedException();
+                throw new InvalidOperationException($"Can't set {name} to a Void value.");
```
CarlosFigueiraMSFT, line 124: >$"Can't set {name} to a Void value." [](http://example.com/codeflow?start=52&length=36)

This needs to be localized. Currently we don't have proper localization in MS.PFx.Interpreter, but we do have for MS.PFx.Core, and we should continue with it.

src/libraries/Microsoft.PowerFx.Core/Functions/IAsyncTexlFunction3.cs
```diff
@@ -0,0 +1,16 @@
...
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Core.Functions
+{
+    // A Texl function capable of async invokes, using IRContext. 
+    internal interface IAsyncTexlFunction3
```
CarlosFigueiraMSFT, line 12: >IAsyncTexlFunction3 [](http://example.com/codeflow?start=23&length=19)

We really need to get rid of the IAsyncTexlFunction*... but that's outside the scope of this change

CarlosFigueiraMSFT:
>                     // where Foo() is a peudo function, get CheckResult for just 'expr' and pass in. 

Not in your changes, but since you're touching this file, can you fix this typo?

---
Refers to: src/libraries/Microsoft.PowerFx.Repl/Repl.cs:291 in ab70632. [](commit_id = ab7063278aa773041afe9507a9123566b088bf48, deletion_comment = False)

CarlosFigueiraMSFT:
>         public void DTypeUnion_PowerFxV1CompatRules_VoidNotCompatibleWithAnything(bool usePFxV1CompatRules)

Please update the test name

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/TypeSystemTests/DTypeTests.cs:2892 in ab70632. [](commit_id = ab7063278aa773041afe9507a9123566b088bf48, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
gregli-msft: My understanding could be way off here.  But I believe that the compiler type checking should catch this before we get to the symbol table, that this is only an internal error, and that makers should not see this.  All the other uses of InvalidOperationException appear to be internal and are not localized, including those from Core.  I only upgraded it from NotSupportedException as that seemed far lamer, with no real information.

gregli-msft:
Good comments, thank you!  I addressed the other two comments as well.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Clear.txt
```diff
@@ -11,9 +11,6 @@
   If(CountRows(t1) = 0, "Cleared", "Failed")
 "Cleared"
 
->> Clear(t1);Clear(t2)
```
anderson-joyle, line 14: PAD possible regression.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/SetFunction.cs
```diff
@@ -40,7 +40,7 @@ public RecalcEngineSetFunction()
               "Set",
               TexlStrings.AboutSet,
               FunctionCategories.Behavior,
-              DType.Boolean,
+              DType.Unknown,
```
anderson-joyle, line 43: Possible PAD breaking.

Issue #1725, opened 2023-08-03
File types 


MikeStall:
Do we need a file type?   
- relationship to DType.File?   / Dataverse type?
- getting files from connectors
- common type for PAD, PVA 

This may be a purely SDK issue - from a language perspective, these would look like Records. Existing type checks should also ensure that we can't assign between real files and records.

PR #2097, opened 2023-12-11
Add a blob value / type and tests

CarlosFigueiraMSFT:


LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.BlobType)

MikeStall:
Issue #1725. 
fyi, PAD was also interested in files.

src/libraries/Microsoft.PowerFx.Core/Public/Values/BlobValue.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        public override object ToObject()
+        {
+            throw new InvalidOperationException();
+        }
+
+        public override string ToString()
+        {
+            return "o";
```
MikeStall, line 35: >return "o"; [](http://example.com/codeflow?start=12&length=11)

?

src/tests/Microsoft.PowerFx.Interpreter.Tests/BlobTests.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        [Fact]
+        public async void CreateBlobTest()
+        {
+            _symbolTable.AddVariable("b", FormulaType.Blob, mutable: true);
+            var opts = new ParserOptions() { AllowsSideEffects = true };
+            var evaluator = _engine.Check("Set(b, BlobCreation(\"myid\"))", opts, _symbolTable).GetEvaluator();
+            var result = await evaluator.EvalAsync(default, _runtimeConfig).ConfigureAwait(false);
+
+            // Set() returns constant 'true;
```
MikeStall, line 50: >// Set() returns constant 'true; [](http://example.com/codeflow?start=12&length=32)

unrelated - did we have some change to make this void?

src/tests/Microsoft.PowerFx.Interpreter.Tests/BlobTests.cs
```diff
@@ -0,0 +1,60 @@
...
+            Assert.Equal(true, result.ToObject());
+
+            evaluator = _engine.Check("b", opts, _symbolTable).GetEvaluator();
+            result = await evaluator.EvalAsync(default, _runtimeConfig).ConfigureAwait(false);
+            var blobValue = result as BlobValue;
+            Assert.NotNull(blobValue);
```
MikeStall, line 56: super nit: can we just cast? Then we'll get a more meaningful exception?

src/libraries/Microsoft.PowerFx.Core/Public/Values/BlobValue.cs
```diff
@@ -0,0 +1,43 @@
...
+// Licensed under the MIT license.
+
+using System;
+using System.Diagnostics.Contracts;
+using System.Text;
+using Microsoft.PowerFx.Core.IR;
+
+namespace Microsoft.PowerFx.Types
+{
+    public class BlobValue : PrimitiveValue<string>
```
MikeStall, line 11: >PrimitiveValue<string> [](http://example.com/codeflow?start=29&length=22)

Changing base class later is a big deal, so we should be confident this is the pattern we want for future scenarios.

madevoge:
Should we extend IValueVisitor to include BlobValue?

Issue #1000, opened 2023-01-18
Update function


anderson-joyle:
Update function implementation.
Doc: [https://learn.microsoft.com/en-us/power-platform/power-fx/reference/function-update-updateif](https://learn.microsoft.com/en-us/power-platform/power-fx/reference/function-update-updateif)

As a reference, check how [PR 595](https://github.com/microsoft/Power-Fx/pull/595). Other fixes came in place after this PR, but it is a good starting point.

Originally test cases were added to [Patch.txt](https://github.com/microsoft/Power-Fx/blob/main/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch.txt) file. We are now trying to walk away from handlers (﻿#SETUP: EnableExpressionChaining,MutationFunctionsTestSetup), so these are for reference only.

For your new test cases, please create a new file under [...Tests/MutationScripts](https://github.com/microsoft/Power-Fx/tree/main/src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationScripts).

coletaylor1:
Can be post GA

Harshit-Lalwani:
Hi @anderson-joyle ,  I'm interested in picking this up. 

A few questions:

- There's a  good first issue label is on this, but looking at the analogous Remove function implementation (PR #3008)  , this seems to require working through the full TexlFunction hierarchy, async interpreter-layer code, and mutation semantics coordination with Power Apps. Would you still consider this appropriate for a first-time contributor?

- Should I follow the Remove/Patch pattern (separate UpdateImpl class in MutationUtils.cs) for the implementation structure?

Happy to take this on if it's still open

PR #2098, opened 2023-12-11
Update function

pooya1380m:
Update function #1000

I implemented the Update function
but I don't now how to write Update.txt , I don't know how test.txt work 
can some body explain it for me


pooya1380m:

anderson-joyle:
Thanks for your PR!
Regarding the txt file test, please create a new `Update.txt` file at `...src\tests\Microsoft.PowerFx.Core.Tests\ExpressionTestCases`.
The syntax should be
```
>> POWER-FX_EXPRESSION_1
EXPECTED_RESULT_1

>> POWER-FX_EXPRESSION_2
EXPECTED_RESULT_2

.
.
.

>> POWER-FX_EXPRESSION_N
EXPECTED_RESULT_N
```

The test framework should catch the file automatically and run the expressions in it.

pooya1380m:
Thanks


>> Patch(t1, r1, {Field2:"mars"});
   t1
Table({Field1:1,Field2:"mars",Field3:DateTime(2022,1,1,0,0,0,0),Field4:true})

How does this work?

MikeStall:
@pooya1380m  - thanks for the contribution. I'm curious - are you using Power Fx libraries for something?

anderson-joyle:
At the top of `Patch.txt` file we have `MutationFunctionsTestSetup` setup. You can check its implementation here:
https://github.com/microsoft/Power-Fx/blob/c37f0963973b1ad80e829f0c8bb80567bfcb805d/src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs#L146C13-L146C13

This setup creates a table variable `t1` populated with some records in advance. We can then reference `t1` within the txt test file.

Please also check txt files at `...src\tests\Microsoft.PowerFx.Interpreter.Tests\MutationScripts`. The big difference is that here the result from an expression persists to the next one.

pooya1380m:
> @pooya1380m  - thanks for the contribution. I'm curious - are you using Power Fx libraries for something? 
No
I'm interested in the concept of Power Fx and would like to contribute to it

pooya1380m:
I want to know if I used the correct black value for my tests ?

PR #2099, opened 2023-12-12
[Informational] [DO NOT MERGE] Prototype Markdown support for Language Server

akshar2401:


src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/SignatureHelp/SignatureInformation.cs
```diff
@@ -48,5 +63,42 @@ public override int GetHashCode()
...
+    public class MarkdownDocumentation
+    {
+        public string Value { get; set; }
+    }
+
+    public class MarkdownSignatureInfomation
+    {
+        public string Label { get; set; }
+
+        public MarkdownDocumentation Documentation { get; set; }
```
MikeStall, line 93: >MarkdownDocumentation [](http://example.com/codeflow?start=15&length=21)

Is the key here that if Documentation is a string, then UFB treats as plain text, and if it's an object (MarkdownDocumentation) , then UFB treats it as markdown?

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/SignatureHelp/SignatureInformation.cs
```diff
@@ -48,5 +63,42 @@ public override int GetHashCode()
...
+            return false;
+        }
+    }
+
+    public class MarkdownDocumentation
+    {
+        public string Value { get; set; }
+    }
+
+    public class MarkdownSignatureInfomation
```
MikeStall, line 89: >MarkdownSignatureInfomation [](http://example.com/codeflow?start=17&length=27)

Note this is missing some other things on SignatureInformation (like equals). Maybe some other pattern of class sharing?

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/SignatureHelp/SignatureInformation.cs
akshar2401: Yeah i left it out for the POC sake but this needs to be fully supported as per the protocol. Happy to provide my thoughts on a complete solution. Yes there's an missing property here called kind. It could be plaintext or markdown (GFM to be precise). It should be added in the final solution.

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/SignatureHelp/SignatureInformation.cs
akshar2401: Yeah this was my "quick" solution but my idea was to have a base class called BaseSignatureInformation which would have only Label and Parameters field and Equals method that only compares label and then there would be child classes that would extend BaseSignatureInformation and add Documentation fields with two different types (string and MarkdownDocumentaiton) respectively. They both would overwrite the equals and gethashcode().

akshar2401: Equals() and GetHashCode() are crucial to avoid duplicate signatures and this was the solution towards duplicate signatures. we need to make sure these methods are accurate with the planned changes to support markdown or it would case regression.

PR #2100, opened 2023-12-12
New functions to retrieve column names and values for untyped objects

CarlosFigueiraMSFT:
We currently have support for untyped objects, but to access their fields they need to be known at build time. This makes the support really untyped - we can access the column names of untyped records and also retrieve a column value given its name as a string.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/ColumnNames.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
+        {
+            yield return new[] { TexlStrings.ColumnNamesArg1 };
+        }
+    }
+
+    // Column(record:O, property_name:s): O
+    internal class ColumnFunction_UO : BuiltinFunction
```
Shpakh-MSFT, line 35: It's a separate function. So probably needs to be in it's own file like other functions.

CarlosFigueiraMSFT: We have examples of other functions bundled together with similar functions. DateTime.cs has lots of date/time related functions; UntypedObejcts.cs has ParseJSON and Index_UO; MinMax.cs has Min/Max; and some others. I don't have a problem splitting it out if you think this is important, though.

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
```diff
@@ -3981,6 +3981,65 @@ public void TestSingleColumnAccessDependentOnTableType(string script, string exp
...
+        }
+
+        [Theory]
+        [InlineData("Column(ParseJSON('Hello'), 'a')")] // Compiles, would fail at runtime
+        [InlineData("Column(ParseJSON('[]'), 'Value')")] // Compiles, would fail at runtime
+        [InlineData("Column(ParseJSON('Hello').a, 'a')")] // Compiles, would fail at runtime
+        [InlineData("Column(ParseJSON('{''a'':1}'), 'a')")]
+        [InlineData("Column(ParseJSON('{''a'':1}').a, 'a')")] // Compiles, would fail at runtime
+        [InlineData("Column(Blank(), 'a')")]
+        [InlineData("Column(ParseJSON('{''a'':{''b'':1}}'), 'a')")]
```
Shpakh-MSFT, line 4020: >{''b'':1 [](http://example.com/codeflow?start=46&length=8)

Test with nested column reference
ParseJSON('{''a'':{''b'':1}}').a , 'b'

Shpakh-MSFT (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/ColumnNames.cs
Shpakh-MSFT: In the past at least we covered function and all it's overloads in single file. It might have changed now.

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
CarlosFigueiraMSFT: This is the same as the test above the Blank table (at least at compile time). I can add another runtime test for that.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/ColumnNames.cs
CarlosFigueiraMSFT: Split it into two different files

Shpakh-MSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2103, opened 2023-12-13
Block unsupported column types for sorting

CarlosFigueiraMSFT:
Today the Sort[ByColumns] functions allow (at compile-time) sorting by many different types, including media, image, blob, ... Those are clearly not sortable - and the implementation doesn't support those - so we should block them at compile-time.

This also adds support for Time columns in the Sort function which was missing, and merges duplicate SortByColumns tests.


MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #2104, opened 2023-12-13
Remove resources from translated file that don't exist in en-US

CarlosFigueiraMSFT:
During development we add resources from the en-US file, and as they are translated they are added to the resources for other languages. But when we remove resources from en-US, they remain on the satellite resource files. This cleans them up for now (although we should have it automated to prevent this problem from happening again).


MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #2107, opened 2023-12-14
Fixed string interpolation with user definitions

VamsiModem:
If we have unterminated double quotes in a string interpolation island in user definitions, that would cause an infinite loop and this PR fixes that


lesaltzm (approved):

LucGenetier:
✅ No public API change.

Issue #2004, opened 2023-10-31
REPL: Get function help 


MikeStall:
We can get function names, but we need information such as the function's description, documentation link, category, and any other relevant details.  Ideally also parameter information.

MikeStall:
We should be able to pull the raw information from the same source as LSP / intellisense.

MikeStall:
For category, this is an array. 

We should expose here:
https://github.com/microsoft/Power-Fx/blob/c4ed8db676f213958ad2f58aee3c57e042077ca1/src/libraries/Microsoft.PowerFx.Core/Public/FunctionInfo.cs#L42


See localized strings here: 
https://github.com/microsoft/Power-Fx/blob/c4ed8db676f213958ad2f58aee3c57e042077ca1/src/strings/PowerFxResources.en-US.resx#L2372

MikeStall:
Fixed with #2154

PR #2108, opened 2023-12-14
Add Description, HelpIink to FunctionInfo

MikeStall:
Partial for #2004. 

Not doing categories yet because we need to discuss how to support that longer term. Is there a curated set? Do we expose the category enum?

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests/FunctionInfoTests.cs
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/FunctionInfoTests.cs
@@ -1,0 +1,29 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Xunit;
+
+namespace Microsoft.PowerFx.Core.Tests
+{
+    public class FunctionInfoTests
+    {
+        [Fact]
+        public void FunctionInfo()
+        {
+            // Get any function info. 
+            var engine = new Engine();
+            var infos = engine.FunctionInfos.ToArray();
+            var infoAbs = infos.Where(info => info.Name == "Mid").First();
+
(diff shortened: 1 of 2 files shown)


src/tests/Microsoft.PowerFx.Core.Tests/FunctionInfoTests.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    public class FunctionInfoTests
+    {
+        [Fact]
+        public void FunctionInfo()
+        {
+            // Get any function info. 
+            var engine = new Engine();
+            var infos = engine.FunctionInfos.ToArray();
+            var infoAbs = infos.Where(info => info.Name == "Mid").First();
```
CarlosFigueiraMSFT, line 20: >infoAbs [](http://example.com/codeflow?start=16&length=7)

Why not call it `infoMid` since you are testing the Mid function, not Abs?

CarlosFigueiraMSFT, line 20: >infos.Where(info => info.Name == "Mid").First(); [](http://example.com/codeflow?start=26&length=48)

Shorter version: infos.Single(info => info.Name == "Mid")

MikeStall: fixed - I originally started with Abs, but Mid is more interesting.

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/FunctionInfoTests.cs
MikeStall: I actually tried this and it's somehow returning 2 identical copies! That's weird, will investigate separately.

LucGenetier:
✅ No public API change.

Issue #2101, opened 2023-12-12
Weeknum() 


MikeStall:
Implement Weeknum() in C#. 

https://learn.microsoft.com/en-us/power-platform/power-fx/reference/function-weeknum

There are some WeekNum.txt, but in not-yet-ready

PR #2109, opened 2023-12-14
WeekNum function

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2101.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DateTime.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DateTime.cs
@@ -272,7 +272,7 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
         public override bool HasPreciseErrors => true;
 
         public WeekNumFunction()
-            : base("WeekNum", TexlStrings.AboutWeekNum, FunctionCategories.DateTime, DType.Number, 0, 1, 2, DType.DateTime, DType.Number)
+            : base("WeekNum", TexlStrings.AboutWeekNum, FunctionCategories.DateTime, DType.Number, 0, 1, 2, DType.DateTime, BuiltInEnums.StartOfWeekEnum.FormulaType._type)
         {
         }
 
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
@@ -688,9 +688,7 @@ namespace Microsoft.PowerFx.Functions
 
             if (startOfWeek <= 0 || startOfWeek > 17 || (startOfWeek > 3 && startOfWeek < 11))
             {
-                return CommonErrors.GenericInvalidArgument(
-                    irContext,
-                    "Expected a value from the StartOfWeek enumeration to indicate how to number the weekdays.");
+                return CommonErrors.StartOfWeekInvalid(irContext);
             }
 
             var zeroIndex = false;
@@ -717,6 +715,54 @@ namespace Microsoft.PowerFx.Functions
             }
 
             return NumberOrDecimalValue(irContext, weekday);
+        }
+
+        public static FormulaValue WeekNum(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
+        {
+            var timeZoneInfo = runner.TimeZoneInfo;
+            var startOfWeek = 0d;
+
+            if (args[0] is BlankValue)
+            {
+                return NumberOrDecimalValue(irContext, 0);
+            }
+
+            var arg0 = runner.GetNormalizedDateTime(args[0]);
+            var dow = arg0.DayOfWeek;
+
+            if (args[1] is ErrorValue)
+            {
+                return args[1];
+            }
+
+            startOfWeek = GetDoubleFromFormulaValue(args[1], 1);
+
+            if (startOfWeek <= 0 || startOfWeek > 17 || (startOfWeek > 3 && startOfWeek < 11) || startOfWeek != Math.Floor(startOfWeek))
+            {
+                return CommonErrors.StartOfWeekInvalid(irContext);
+            }
+
+            if (startOfWeek == 3)
+            {
+                return CommonErrors.GenericInvalidArgument(
+                    irContext,
+                    "The MondayZero value, from the StartOfWeek enumeration, is not supported in the WeekNum function.");
+            }
+
+            var beginningOfYear = new DateTime(arg0.Year, 1, 1);
+            var weekdayFormulaValue = Weekday(runner, context, irContext, new FormulaValue[] { FormulaValue.New(beginningOfYear), new NumberValue(IRContext.NotInSource(FormulaType.Number), 1) });
+
+            if (weekdayFormulaValue is ErrorValue)
+            {
+                return weekdayFormulaValue;
+            }
+
+            var weekdayResult = GetDoubleFromFormulaValue(weekdayFormulaValue);
+            var dateDiff = arg0.Date.Subtract(beginningOfYear).TotalDays;
+            var dayOfWeek = (weekdayResult - WeekStartDay(startOfWeek) + 7) % 7;
+            var weeknum = Math.Floor((dateDiff + dayOfWeek) / 7) + 1;
+
+            return NumberOrDecimalValue(irContext, (int)weeknum);
         }
 
         public static FormulaValue EDate(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
@@ -777,6 +823,42 @@ namespace Microsoft.PowerFx.Functions
             else
             {
                 throw CommonExceptions.RuntimeMisMatch;
+            }
+        }
+
+        private static double WeekStartDay(double startOfWeek)
+        {
+            if (startOfWeek == 1 || startOfWeek == 2)
+            {
+                return startOfWeek;
+            }
+
+            if (startOfWeek >= 11 && startOfWeek <= 16)
+            {
+                return startOfWeek - 9;
+            }
+
+            if (startOfWeek == 17)
+            {
+                return 1;
+            }
+
+            return 0;
+        }
+
+        private static double GetDoubleFromFormulaValue(FormulaValue fv, double defaultDouble = 0)
+        {
+            if (fv is NumberValue numberValue)
+            {
+                return numberValue.Value;
+            }
+            else if (fv is DecimalValue decimalValue)
+            {
+                return (double)decimalValue.Value;
+            }
+            else
+            {
+                return defaultDouble;
             }
         }
     }
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
@@ -719,15 +719,14 @@ namespace Microsoft.PowerFx.Functions
 
         public static FormulaValue WeekNum(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
         {
-            var timeZoneInfo = runner.TimeZoneInfo;
             var startOfWeek = 0d;
+            var arg0 = _epoch;
 
-            if (args[0] is BlankValue)
+            if (args[0] is not BlankValue)
             {
-                return NumberOrDecimalValue(irContext, 0);
+                arg0 = runner.GetNormalizedDateTime(args[0]);
             }
 
-            var arg0 = runner.GetNormalizedDateTime(args[0]);
             var dow = arg0.DayOfWeek;
 
             if (args[1] is ErrorValue)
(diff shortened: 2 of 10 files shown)


LucGenetier:
✅ No public API change.

MikeStall (dismissed):
:shipit:

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/WeekNum.txt
```diff
@@ -11525,76 +11525,76 @@
 
 // **************** SAME DAY - ALL HOURS TESTCASES  ****************
 
->> WeekNum(DateTimeValue("1/7/2021 0:00:00:00"),15)
+>> WeekNum(DateTimeValue("1/7/2021 0:00:00"),15)
```
carlosfigueira, line 11528: We should not force the expression to use the enumeration (strongly-typed enum) in this function. Calls with the numeric equivalent should fail.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
```diff
@@ -719,6 +717,54 @@ public static FormulaValue Weekday(EvalVisitor runner, EvalVisitorContext contex
...
         }
 
+        public static FormulaValue WeekNum(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
+        {
+            var timeZoneInfo = runner.TimeZoneInfo;
+            var startOfWeek = 0d;
+
+            if (args[0] is BlankValue)
+            {
+                return NumberOrDecimalValue(irContext, 0);
```
carlosfigueira, line 727: Blank as the first argument should be interpreted as the "date zero" (1899-12-30), so the result may not be 0

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
```diff
@@ -719,6 +717,54 @@ public static FormulaValue Weekday(EvalVisitor runner, EvalVisitorContext contex
...
+            var dow = arg0.DayOfWeek;
+
+            if (args[1] is ErrorValue)
+            {
+                return args[1];
+            }
+
+            startOfWeek = GetDoubleFromFormulaValue(args[1], 1);
+
+            if (startOfWeek <= 0 || startOfWeek > 17 || (startOfWeek > 3 && startOfWeek < 11) || startOfWeek != Math.Floor(startOfWeek))
```
carlosfigueira, line 740: If we force the maker to use the enumeration value statically, we won't need those checks

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
```diff
@@ -719,6 +717,54 @@ public static FormulaValue Weekday(EvalVisitor runner, EvalVisitorContext contex
             return NumberOrDecimalValue(irContext, weekday);
         }
 
+        public static FormulaValue WeekNum(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
+        {
+            var timeZoneInfo = runner.TimeZoneInfo;
```
carlosfigueira, line 722: unused

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/WeekNum.txt
carlosfigueira, line 11528: Instead of using the DateTimeValue, we should use the DateTime (or Date) function in tests. The former is locale-dependent, and we want the tests to be run in most locations. For example, that value could be interpreted both as January 7th and July 1st depending on whether the tests are run. If we use the Date[Time] function, the result will not be ambiguous.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/WeekNum.txt
```diff
@@ -11525,76 +11525,76 @@
```
carlosfigueira, line 11525: I'm all for having many tests, but in this case we're going overboard. Do we really need 4000-ish cases to cover this function? As long as we cover the interesting cases (early dates, beginning / end of year, end of February / beginning of March, different startofweek arguments), we should have complete coverage of this function. We should also have test for earlier dates (i.e., before 1000) and later dates (after 9000) which I couldn't find (maybe there are, but it's hard to find in the middle of so many unnecessary tests.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/WeekNum.txt
```diff
@@ -11680,7 +11677,7 @@ Error({Kind:ErrorKind.InvalidArgument})
 
 // Valid date passed as DateTimeValue parameter
 >> WeekNum(1/2/2022,StartOfWeek.Sunday)
```
carlosfigueira, line 11679: "Valid date passed as DateTimeValue" - this is incorrect. We are not passing a valid date here, we are passing a number (`1/2/2022` which is (1/2)/2022. I'm don't think it adds value to the test for this function itself.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/WeekNum.txt
anderson-joyle: This file was previously in the "NotReadyYet" folder. We can leave it as is for now since we need to unblock other teams and work on cutting some fat from this txt file in an after-match issue. @MikeStall thoughts?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
anderson-joyle: Yes. but then we are breaking the semantics since PA allows it. Is there an easy way to handle this from PA side?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/WeekNum.txt
anderson-joyle: Thanks for catching this.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
carlosfigueira: The Weekday function handles that. See the tests, where we have two versions of tests with numbers as StartOfWeek argument, one with StronglyTypedBuiltinEnums enabled one with it disabled (the former returns errors, the latter works)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/WeekNum.txt
carlosfigueira: I typically don't like to leave things to do later - they have the tendency of getting pushed out by some other priorities. Do the other teams really need this now, or can they wait another 1-2 days?

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/WeekNum_StronglyTypedEnum.txt
```diff
@@ -0,0 +1,173 @@
...
+7
+
+>> WeekNum(Date(2020, 3, 30))
+14
+
+>> WeekNum(Date(2020, 7, 26))
+31
+
+// Same day - all hours
+>> WeekNum(DateTime(2021,1,7,0,0,0), StartOfWeek.Friday)
```
carlosfigueira, line 93: All the tests that work with both StronglyTypedEnums on and off can go to a "common" test file (WekNum.txt). Only the differences should be in the split test files.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/WeekNum_StronglyTypedEnum.txt
```diff
@@ -0,0 +1,173 @@
...
+>> WeekNum(DateTime(2021,1,7,9,0,0), StartOfWeek.Friday)
+1
+
+>> WeekNum(DateTime(2021,1,7,10,0,0), StartOfWeek.Friday)
+1
+
+>> WeekNum(DateTime(2021,1,7,11,0,0), StartOfWeek.Friday)
+1
+
+>> WeekNum(DateTime(2021,1,7,12,0,0), StartOfWeek.Friday)
```
carlosfigueira, line 128: Instead of having the hours from 0-12, it is enough to have a few interesting points at the day:
- DateTime(2021,1,7,0,0,0)
- DateTime(2021,1,7,12,0,0)
- DateTime(2021,1,7,23,59,59,999)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/WeekNum_StronglyTypedEnum.txt
```diff
@@ -0,0 +1,173 @@
...
+Errors: Error 25-26: Invalid argument type (Decimal). Expecting a Enum (StartOfWeek) value instead.
+
+>> WeekNum(Date(2022,1,2),1E308)
+Errors: Errors: Error 23-28: Numeric value is too large.|Error 23-28: Invalid argument type (Error). Expecting a Enum (StartOfWeek) value instead.
+
+>> WeekNum(Date(2016, 1, 3),"")
+Errors: Error 25-27: Invalid argument type (Text). Expecting a Enum (StartOfWeek) value instead.
+
+>> WeekNum(Date(2020, 5, 3),StartOfWeek.MondayZero)
+Error({Kind:ErrorKind.InvalidArgument})
```
carlosfigueira, line 173: We should have tests for the edge of valid values (Date(0001,1,1) and Date(9999,12,31))

carlosfigueira: And for weeknum around leap days

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
```diff
@@ -50,6 +50,7 @@
 
   <ItemGroup>
     <None Remove="expressiontestcases\README.md" />
+    <None Remove="ExpressionTestCases\WeekNum_StronglyTypedEnumDisabled.txt" />
```
MikeStall, line 53: > <None Remove="ExpressionTestCases\WeekNum_StronglyTypedEnumDisabled.txt" /> [](http://example.com/codeflow?start=3&length=76)

nit - vs issue?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DateTime.cs
```diff
@@ -272,7 +272,7 @@ internal sealed class WeekNumFunction : BuiltinFunction
         public override bool HasPreciseErrors => true;
 
         public WeekNumFunction()
-            : base("WeekNum", TexlStrings.AboutWeekNum, FunctionCategories.DateTime, DType.Number, 0, 1, 2, DType.DateTime, DType.Number)
+            : base("WeekNum", TexlStrings.AboutWeekNum, FunctionCategories.DateTime, DType.Number, 0, 1, 2, DType.DateTime, BuiltInEnums.StartOfWeekEnum.FormulaType._type)
```
MikeStall, line 275: >BuiltInEnums.StartOfWeekEnum.FormulaType._type [](http://example.com/codeflow?start=124&length=46)

is this ok? Could this impact Power Apps?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/WeekNum_StronglyTypedEnum.txt
```diff
@@ -0,0 +1,173 @@
+#SETUP: disable:NumberIsFloat,StronglyTypedBuiltinEnums
+
+// Blank args
+>> WeekNum(Blank(),Blank())
```
MikeStall, line 4: we should include na error example. 

WeekNum(error,valid)

WeekNum(valid,error)

Easy way to generate typed-errors is:

`If(1/0, Date(1900,11,22))`

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DateTime.cs
anderson-joyle: I spoke with PA team member, and it shouldn't. The `Weekday` function works similar to this scenario. I'm also validating this by building PA against these changes.

MikeStall (dismissed):

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

PR #2117, opened 2023-12-19
Add AI disclaimer

MikeStall:
Fix #1202

The disclaimer is markdown (including a help link) in the resources.

1. TexlFunction gets a virtual flag noting whether the function needs a disclaimer.  Default impl is based on a whitelist of AI functions, but we can override this flag next.
2. LSP will get flag and append , and send markdown to formula bar. 

This productizes the POC from #2099 .

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
@@ -279,7 +279,19 @@ namespace Microsoft.PowerFx.Core.Functions
         /// <summary>
         /// This function requires an AI disclaimer. 
         /// </summary>
-        public virtual bool ShowAIDisclaimer => this.Name.StartsWith("AI", StringComparison.OrdinalIgnoreCase);
+        public virtual bool ShowAIDisclaimer => _aiWhitelist.Contains(this.Name);
+
+        // Move away from whitelist: https://github.com/microsoft/Power-Fx/issues/2118
+        private static readonly ISet<string> _aiWhitelist = new HashSet<string>()
+        {
+            "AIClassify", 
+            "AIExtract",
+            "AIReply",
+            "AISentiment",
+            "AISummarize",
+            "AISummarizeRecord",
+            "AITranslate"
+        };
 
         // A forward link to the function help.
         public virtual string HelpLink =>
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
@@ -279,9 +279,7 @@ namespace Microsoft.PowerFx.Core.Functions
         /// <summary>
         /// This function requires an AI disclaimer. 
         /// </summary>
-        // $$$ Pass this flag in
-        //public bool UseAIDisclaimer => this.Name.StartsWith("AI");
-        public bool ShowAIDisclaimer => true;
+        public virtual bool ShowAIDisclaimer => this.Name.StartsWith("AI", StringComparison.OrdinalIgnoreCase);
 
         // A forward link to the function help.
         public virtual string HelpLink =>
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
@@ -276,6 +276,13 @@ namespace Microsoft.PowerFx.Core.Functions
         // A description associated with this function.
         public string Description => _description(null);
 
+        /// <summary>
+        /// This function requires an AI disclaimer. 
+        /// </summary>
+        // $$$ Pass this flag in
+        //public bool UseAIDisclaimer => this.Name.StartsWith("AI");
+        public bool ShowAIDisclaimer => true;
+
         // A forward link to the function help.
         public virtual string HelpLink =>
 
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/MarkdownString.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/MarkdownString.cs
@@ -1,0 +1,23 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.PowerFx.Intellisense
+{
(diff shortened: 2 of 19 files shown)


src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SignatureInformation.cs
```diff
@@ -3,13 +3,46 @@
...
+            {
+                this.Parameters = Array.ConvertAll(info.Parameters, x => new ParameterInformation(x));
+            }
+
+            this.Label = info.Label;
+            this.Documentation = info.Documentation;
+            if (info.GetDisclaimerMarkdown != null)
+            {
+                // Append disclaimer to create a final markdown string to send to LSP. 
+                MarkdownString disclaimer = info.GetDisclaimerMarkdown();
```
MikeStall, line 34: This is the key change

MikeStall:
>                         var listSep = TexlLexer.GetLocalizedInstance(CultureInfo.CurrentCulture).LocalizedPunctuatorListSeparator + " ";

LSP elsewhere is still using CurrentCulture. . 

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseResult.cs:97 in 4b742e0. [](commit_id = 4b742e03086fba4af356e113bba93f5848add149, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/MarkdownString.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.PowerFx.Intellisense
+{
+    /// <summary>
+    /// A string that represents Github flavored markdown. 
+    /// For use in displaying in intellisense. 
+    /// </summary>
+    public class MarkdownString
```
MikeStall, line 14: >MarkdownString [](http://example.com/codeflow?start=17&length=14)

introducing a wrapper so that the type system can help us with "plain text" vs. "markdown".

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -274,7 +274,12 @@ public virtual async Task<ConnectorSuggestions> GetConnectorSuggestionsAsync(For
         public string LocaleInvariantName { get; }
 
         // A description associated with this function.
-        public string Description => _description(null);
+        public string Description => _description(null);
+
+        /// <summary>
+        /// This function requires an AI disclaimer. 
+        /// </summary>
+        public virtual bool ShowAIDisclaimer => this.Name.StartsWith("AI", StringComparison.OrdinalIgnoreCase);
```
akshar2401, line 282: if i write a UDF that starts with "AI", would this be true for that? i assume udfs would be created during analysis phase. Do they inherit from this TexlFunction? Can this be overwritten in derived classes?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -375,7 +375,9 @@ private void HandleSignatureHelpRequest(string id, string paramsJson)
             var cursorPosition = GetPosition(expression, signatureHelpParams.Position.Line, signatureHelpParams.Position.Character);
             var result = scope.Suggest(expression, cursorPosition);
 
-            _sendToClient(JsonRpcHelper.CreateSuccessResult(id, result.SignatureHelp));
+            var lspObj = new SignatureHelp(result.SignatureHelp);
```
akshar2401, line 378: nit: may be use a better name here

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SignatureInformation.cs
```diff
@@ -3,13 +3,46 @@
...
+                string original = info.Documentation;
+
+                this.Documentation = new MarkdownStringHolder
+                {
+                    Value = original + "\r\n" + disclaimer.Markdown
```
akshar2401, line 39: I will need to verify if \r\n would always work or not. UFB or more like monaco randomly decides which line feed to use \n or \r\n.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
akshar2401: or what if we add a function in future that starts with AI or one of the third party connectors function is prefixed with AI? If this would be overwritten in derived classes then it is fine but startswith(AI) feels like an inefficient approach

MikeStall: This is just starting with a whitelist. But I can spell them out until we get the flag propagated from swagger.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SignatureInformation.cs
akshar2401: use \n\n here

MikeStall: updated. And wrapped in a handy static field.

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/MarkdownString.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// <returns></returns>
+        public static MarkdownString operator +(MarkdownString left, MarkdownString right) => Add(left, right);
+
+        /// <summary>
+        /// Concatenate two markdown strings. 
+        /// </summary>
+        /// <param name="left"></param>
+        /// <param name="right"></param>
+        /// <returns></returns>
+        public static MarkdownString Add(MarkdownString left, MarkdownString right)
```
MikeStall, line 92: There's a stylecop rule that require Add() if you have op+.

jas-valgotar (approved):
:shipit:

akshar2401 (approved):

LucGenetier:
✅ No public API change.

jas-valgotar (approved):

PR #2123, opened 2023-12-22
Adding a new isReactHooksSingleton argument to transport type attribute constructor

pacrozie:
Adding a new argument to the transport type attribute to support React Hooks for the Power Apps Canvas client.


src/libraries/Microsoft.PowerFx.Transport.Attributes/TransportTypeAttribute.cs
```diff
@@ -32,13 +32,14 @@ public class TransportTypeAttribute : Attribute
 
         public bool EnablePublicMembersByDefault { get; }
 
-        public TransportTypeAttribute(TransportKind kind = TransportKind.ByValue, bool enablePublicMembersByDefault = true, string customTypescriptBaseClass = null, string customDtoName = null, bool isMethodCustomizationEnabled = false)
+        public TransportTypeAttribute(TransportKind kind = TransportKind.ByValue, bool enablePublicMembersByDefault = true, string customTypescriptBaseClass = null, string customDtoName = null, bool isMethodCustomizationEnabled = false, bool isReactHooksSingleton = false)
```
MikeStall, line 35: not sure if this matters, but adding a default parameter is a binary breaking change. 
I'm not sure why our API compat tool is not catching and flagging that....

lesaltzm (dismissed):

src/libraries/Microsoft.PowerFx.Transport.Attributes/TransportTypeAttribute.cs
lesaltzm: The API review tool I think isn't covering anything other than Core and Interpreter? Anyway, this should be fine, nothing uses this other than PFx and canvas code.

MikeStall (approved):

lesaltzm (approved):

PR #2124, opened 2023-12-22
[THROW WAY] [DO NOT MERGE] Refactor Nl2FX to support more dynamic NlHandler creation

akshar2401:


src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -63,13 +65,14 @@ public class LanguageServer
         /// </summary>
         public NLHandler NL2FxImplementation { get; set; }
```
MikeStall, line 66: >NLHandler [](http://example.com/codeflow?start=15&length=9)

I'm ok with adding the factory. 


TBD - do we deprecate /obsolete this ?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Public/INLHandlerFactory.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT license.
+
+using System.Collections.Specialized;
+using Microsoft.PowerFx.Intellisense;
+
+namespace Microsoft.PowerFx.LanguageServerProtocol.Public
+{
+    public interface INLHandlerFactory
+    {
+        public NLHandler GetNlHandler(IPowerFxScope uri);
```
MikeStall, line 11: >IPowerFxScope [](http://example.com/codeflow?start=38&length=13)

this makes sense. 
Scope uri tells us which formula bar we have, and the context could be dependent on that.

PR #2125, opened 2023-12-27
Office 365 connector extensive testing and fixes

LucGenetier:
> [!CAUTION]
> BREAKING CHANGE
> - Internal functions are no more visible externally 
  Functions having _x-ms-visibility_ extension with "internal" value are not visible in PowerApps
  We were making all these functions visible and this is now fixed
  **⚠️ Some customers had access to functions they won't find them anymore** 
 Tabular connectors are heavily impacted here - those connectors shouldn't be exposed as they aren't presented like PowerApps as a table
  It is possible to expose internal functions via _ConnectorSettings.IncludeInternalFunctions_ (default value false) but this should only be reserved for tests and never set in production
  FYI, these internal functions are heavily used in dynamic Intellisense
- Serialization of "date-time" and "date-no-tz" swagger types was incorrect
  - "date-time" fields are converted properly (wasn't the case) & sent as UTC on the wire (`yyyy-MM-ddTHH:mm:ss.fffZ` format), precision has been fixed (trucated to 1 ms)
    This is what PowerApps is doing
  - "date-no-tz" fields are sent on the wire with `yyyy-MM-ddTHH:mm:ss.fff` format (no Z, truncated to 1ms)
    PowerApps has a bug under investigation (PA swallows seconds, ms and us, potentially leading to incorrect results)
- Deserialization of "date-time" fields was having some timezone conversion issues
  This is fixed
- Connector parameters are case-sensitive and we were managing them as case-insensitive
  This is fixed
- Some parameters with identical names (which are renamed with _1, _2... like in PA) where improperly serialized in HTTP payload
  This is fixed
- Connector parameters with array parameters in circular references where improperly identified as strings (instead of array)
  This is fixed
- O365 Outlook _GetSensitivityLabels_ hidden function doesn't work due to a very probable bug in its swagger definition - this is under investigation
  This impacts _SendEmail_ and _SendEmailV2_ functions where dynamic intellisense will not work properly for _Sensitivity_ param (no suggestion returned)
- Connector MediaKind is now identified in swagger files (_x-ms-media-kind_ extension)
  Unused for now, but this is to properly support image params later


src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
```diff
@@ -125,6 +125,8 @@ public string NotSupportedReason
         /// https://learn.microsoft.com/en-us/connectors/custom-connectors/openapi-extensions#x-ms-visibility.
         /// </summary>
         public bool IsInternal => Operation.IsInternal();
+        
+        public bool IsHidden => Visibility.ToVisibility() == Connectors.Visibility.Internal;
```
MikeStall, line 129: What's the difference between IsInternal and IsHidden?
(especially since Hidden checks the internal flag)

LucGenetier: Absolutely none :)
I've removed IsHidden and all "hidden" notions to use IsInternal instead
Good catch!

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/Execution/OpenApiJsonSerializer.cs
```diff
@@ -60,8 +60,14 @@ protected override void WriteBooleanValue(bool booleanValue)
 
         protected override void WriteDateTimeValue(DateTime dateTimeValue)
         {
-            // ISO 8601            
-            _writer.WriteStringValue(dateTimeValue.ToString("o", CultureInfo.InvariantCulture));
+            // ISO 8601                        
+            _writer.WriteStringValue(dateTimeValue.ToString("yyyy-MM-ddTHH:mm:ss.fffZ", CultureInfo.InvariantCulture));
```
MikeStall, line 64: >"yyyy-MM-ddTHH:mm:ss.fffZ" [](http://example.com/codeflow?start=60&length=26)

nit - since these are used in multiple places, can you define as a const field in the base class?

src/libraries/Microsoft.PowerFx.Connectors/Public/BaseRuntimeConnectorContext.cs
```diff
@@ -17,11 +17,13 @@ public abstract class BaseRuntimeConnectorContext
 
         internal virtual bool ReturnRawResults { get; } = false;
 
+        internal virtual MediaKind MediaKind { get; set; } = MediaKind.NotBinary;
```
MikeStall, line 20: >virtual MediaKind MediaKind  [](http://example.com/codeflow?start=17&length=28)

What exactly does this flag mean?

1. Is it global for all connectors?
2. when does the host set it to the non-default value?

LucGenetier: this is x-ms-media-kind extension which defines if a function returns a byte array if this is a video, an image...
it's per function and defined in the swagger file

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2130, opened 2024-01-04
Increase local NuGet version to a large number

CarlosFigueiraMSFT:
As we increase our official NuGet version, the version from a local build should have a higher version number. This increases it to a very large number, which should be enough until our major version update.


MikeStall (approved):

LucGenetier:
✅ No public API change.

Issue #2127, opened 2023-12-29
Don't copy derived RecordValues to inmemory record values


MikeStall:
Hosts can have derived RecordValues which can have key state (like a file handle). When Collecting, we don't want to copy to an in-memory record value because that loses the handle. 

Ie, this should pass:

```
RecordValue x = new FileObjectRecordValue  ( ...);
engine.UpdateVariable("myFile", x)
var result= engine.Eval("Collect(Files, myFile)") 
Assert.True(Object.ReferenceEquals(x, result) 
```

See #2092 for related examples. 

1. We should look at IR and see why the copy is happening (is it a coercion? If "Files" is declared correctly, there shouldn't be any coercion needed
2. Ensure it's possible and expand test in 2092 to show how to do it. What does FileObjectRecordValue  need to derive from, etc.

PR #1145, opened 2023-03-03
Fixes #1049 Wrong Mutation semantics

jas-valgotar:
- Fixes #1049 
- Now Collect function appends the deep copy of the provided record.

PR #2131, opened 2024-01-08
Don't copy derived RecordValues to inmemory record values

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2127.


src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/CollectFunction.cs
```diff
@@ -299,10 +299,9 @@ public virtual async Task<FormulaValue> InvokeAsync(FormulaValue[] args, Cancell
 
             var tableValue = arg0 as TableValue;
             var recordValue = arg1 as RecordValue;
-            var recordValueCopy = FormulaValue.NewRecordFromFields(recordValue.Fields);
```
MikeStall, line 302: fwiw, this line was added in #1145 
This current PR is effectively reverting 1145. 

Strange that removing it doesn't cause any test failures.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/CollectFunction.cs
jas-valgotar: What does this revert mean? 
With({r:{Value:5}}, Collect(t1, r); Patch(t1, Last(t1), {Value:-1}); Notify(r.Value))

What will be r.Value now? since its by ref now, I assume -1?
NOTE: PA doesn't copy by ref and will return 5.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/CollectFunction.cs
anderson-joyle: We are still getting `{Value:5}`. I added these test cases and some extra ones to this PR.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/CollectFunction.cs
anderson-joyle: When patching, Fx is copying the record here https://github.com/microsoft/Power-Fx/blob/c8ebad07563535b4a12be979a0f6a2f20081b467/src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs#L240.
We are hitting the same problem when executing `Patch`.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/CollectFunction.cs
```diff
@@ -298,11 +298,12 @@ public virtual async Task<FormulaValue> InvokeAsync(FormulaValue[] args, Cancell
             }
 
             var tableValue = arg0 as TableValue;
-            var recordValue = arg1 as RecordValue;
-            var recordValueCopy = FormulaValue.NewRecordFromFields(recordValue.Fields);
+            var recordValue = arg1 as RecordValue;
+
+            var recordValueToAppend = (RecordValue)recordValue.MaybeShallowCopy();
```
jas-valgotar, line 303: >RecordValue [](http://example.com/codeflow?start=39&length=11)

nit: already a cast at L301

anderson-joyle: Yes but `MaybeShallowCopy` returns a FormulaValue. That's why this extra cast.

LucGenetier:
✅ No public API change.

jas-valgotar (dismissed):

jas-valgotar (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2133, opened 2024-01-09
Add UniChar function

adithyaselv:
Adds UniChar funcion to Power Fx.

**Examples** 
```
>> UniChar(Hex2Dec("1F600"))
"😀"

>> UniChar(119137)
"𝅘𝅥𝅯"
```

**User scenarios** 
* Text based custom icons within the app.
* Display multilingual texts.

**References**
[UNICHAR function - Microsoft Support](https://support.microsoft.com/en-us/office/unichar-function-ffeb64f5-f131-44c6-b332-5cd72f0659b8)
[Char.ConvertFromUtf32(Int32) Method (System) | Microsoft Learn](https://learn.microsoft.com/en-us/dotnet/api/system.char.convertfromutf32?view=net-8.0)
[Unicode - Surrogate pair (UTF-16) (datacadamia.com)](https://datacadamia.com/data/type/text/surrogate)


src/strings/PowerFxResources.en-US.resx
```diff
@@ -4377,4 +4377,26 @@
...
+  <data name="AboutUnicharT" xml:space="preserve">
+    <value>Converts a table of numbers to table of Unicode strings</value>
+    <comment>Description of 'Unichar' function (table overload).</comment>
+  </data>
+  <data name="UnicharTArg1" xml:space="preserve">
+    <value>column_of_numbers</value>
+    <comment>function_parameter - First parameter for the Unichar function -  a table column of numbers to be converted into unicode strings.</comment>
+  </data>
+  <data name="AboutUnichar_column_of_numbers" xml:space="preserve">
+    <value>a table column of numbers to be converted into unicode strings.</value>
```
CarlosFigueiraMSFT, line 4400: >strings [](http://example.com/codeflow?start=66&length=7)

characters

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Unichar.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            return base.GetGenericArgPreprocessor(index);
+        }
+
+        public override bool IsSelfContained => true;
+
+        public override bool IsStateless => true;
+
+        public UnicharFunction()
+            : base("Unichar", TexlStrings.AboutUnichar, FunctionCategories.Text, DType.String, 0, 1, 1, DType.Number)
```
CarlosFigueiraMSFT, line 30: >Unichar [](http://example.com/codeflow?start=20&length=7)

Should we use `UniChar`? This is a compound noun ("Unicode Char"), and for other functions we use pascal casing (e.g., LookUp, AddColumns, TimeValue, IsMatch, and so on)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -1187,6 +1187,46 @@ private static DateTime ConvertToUTC(DateTime dateTime, TimeZoneInfo fromTimeZon
...
+                    Kind = ErrorKind.InvalidArgument
+                });
+            }
+            else if (isPartialSurrogate)
+            {
+                return new ErrorValue(irContext, new ExpressionError()
+                {
+                    Message = $"Input value {number} is an invalid argument",
+                    Span = irContext.SourceContext,
+                    Kind = ErrorKind.InvalidArgument
```
CarlosFigueiraMSFT, line 1220: For compatibility with Excel we can return ErrorKind.NotApplicable here

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -1187,6 +1187,46 @@ private static DateTime ConvertToUTC(DateTime dateTime, TimeZoneInfo fromTimeZon
             return resultDateTime.UtcDateTime;
         }
 
+        public static FormulaValue Unichar(IRContext irContext, NumberValue[] args)
+        {
+            double number = args[0].Value;
+
+            // Unicode upper and lower bounds
+            const double lowerBound = 1; // Excel returns error for Unichar(0)
+            const double upperBound = 1114111; // 0x10FFFF
```
CarlosFigueiraMSFT, line 1196: Excel returns #NA for 1114111 and 1114110 - UNICHAR(1114109) returns a valid value. C# and JS do support those values, so I think we're ok in supporting them in Power Fx.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -1187,6 +1187,46 @@ private static DateTime ConvertToUTC(DateTime dateTime, TimeZoneInfo fromTimeZon
...
+            if (isOutOfRange)
+            {
+                return new ErrorValue(irContext, new ExpressionError()
+                {
+                    Message = $"Input value {number} falls outside the allowable range",
+                    Span = irContext.SourceContext,
+                    Kind = ErrorKind.InvalidArgument
+                });
+            }
+            else if (isPartialSurrogate)
```
CarlosFigueiraMSFT, line 1214: No need to use `else` here, since the previous block returns already

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -1187,6 +1187,46 @@ private static DateTime ConvertToUTC(DateTime dateTime, TimeZoneInfo fromTimeZon
...
+                return new ErrorValue(irContext, new ExpressionError()
+                {
+                    Message = $"Input value {number} is an invalid argument",
+                    Span = irContext.SourceContext,
+                    Kind = ErrorKind.InvalidArgument
+                });
+            }
+            else
+            {
+                string unicode = char.ConvertFromUtf32((int)Math.Floor(number));
```
CarlosFigueiraMSFT, line 1225: Numeric arguments are rounded down for this function, so `UniChar(1114111)` should be the same as `UniChar(1114111.99)` - but this is not the case with the current implementation (it would return an error)

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4377,4 +4377,26 @@
     <value>**Disclaimer:** AI-generated content can have mistakes. Make sure it's accurate and appropriate before using it. [See terms](https://go.microsoft.com/fwlink/?linkid=2225491)</value>
     <comment>The disclaimer we show on AI functions. This is github flavored markdown. So ** marks text in bold. </comment>
   </data>
+  <data name="AboutUnichar" xml:space="preserve">
+    <value>Converts number to Unicode string</value>
```
CarlosFigueiraMSFT, line 4381: Suggestion: "Returns the Unicode character that is referenced by the given numeric value."

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4377,4 +4377,26 @@
...
+  </data>
+  <data name="AboutUnichar_number" xml:space="preserve">
+    <value>number to be converted into unicode.</value>
+  </data>
+  <data name="UnicharArg1" xml:space="preserve">
+    <value>number</value>
+    <comment>function_parameter - First parameter for the Unichar function - the number to be converted into unicode string.</comment>
+  </data>
+  <data name="AboutUnicharT" xml:space="preserve">
+    <value>Converts a table of numbers to table of Unicode strings</value>
```
CarlosFigueiraMSFT, line 4392: >to table of [](http://example.com/codeflow?start=39&length=11)

...to **a** table of Unicode...

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4377,4 +4377,26 @@
     <value>**Disclaimer:** AI-generated content can have mistakes. Make sure it's accurate and appropriate before using it. [See terms](https://go.microsoft.com/fwlink/?linkid=2225491)</value>
     <comment>The disclaimer we show on AI functions. This is github flavored markdown. So ** marks text in bold. </comment>
   </data>
+  <data name="AboutUnichar" xml:space="preserve">
+    <value>Converts number to Unicode string</value>
+    <comment>Description of 'Unichar' function.</comment>
+  </data>
+  <data name="AboutUnichar_number" xml:space="preserve">
+    <value>number to be converted into unicode.</value>
```
CarlosFigueiraMSFT, line 4385: >unicode [](http://example.com/codeflow?start=39&length=7)

Unicode (capitalization)

CarlosFigueiraMSFT, line 4385: >number [](http://example.com/codeflow?start=11&length=6)

Number (capitalization)

CarlosFigueiraMSFT, line 4385: >>number to be converted into unicode. [](http://example.com/codeflow?start=10&length=37)

Suggestion: "Number to be converted into a Unicode string"

src/strings/PowerFxResources.en-US.resx
CarlosFigueiraMSFT, line 4400: >a [](http://example.com/codeflow?start=11&length=1)

`A` - capitalization

CarlosFigueiraMSFT, line 4400: >unicode [](http://example.com/codeflow?start=58&length=7)

`Unicode` - capitalization

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Unichar.cs
```diff
@@ -0,0 +1,76 @@
...
+            Contracts.Assert(MinArity <= args.Length && args.Length <= MaxArity);
+
+            var fValid = base.CheckTypes(context, args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
+            Contracts.Assert(returnType.IsTable);
+
+            // Typecheck the input table
+            fValid &= CheckNumericColumnType(context, args[0], argTypes[0], errors, ref nodeToCoercedTypeMap);
+
+            // Synthesize a new return type
+            returnType = DType.CreateTable(new TypedName(DType.String, GetOneColumnTableResultName(context.Features)));
```
CarlosFigueiraMSFT, line 71: > GetOneColumnTableResultName(context.Features))) [](http://example.com/codeflow?start=70&length=48)

New functions that return a single-column table all have the schema `*[Value:<type>]` - i.e., the column name is always `Value`, regardless of the features.

CarlosFigueiraMSFT:
> #SETUP: ConsistentOneColumnTableResult

Can be removed (see comment on Unichar.cs about result column name)

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/UnicharT.txt:1 in 75605b0. [](commit_id = 75605b0e381fc4ce57cb3fbd98dc71a8934bf10e, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
adithyaselv: I thought `Math.Floor` would help there and tried this test https://github.com/microsoft/Power-Fx/pull/2133/files#diff-1ac69fe955d21389cbada90d26edbc3abd7a4692f8f6c9a975e8fef93b068c8aR68-R69 , https://github.com/microsoft/Power-Fx/pull/2133/files#diff-1ac69fe955d21389cbada90d26edbc3abd7a4692f8f6c9a975e8fef93b068c8aR13-R14 . But i see what you mean. I will add this specific recommended test case and fix

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
LucGenetier: FWIW, I would use hex notation in the code itself instead of using the decimal form and a comment

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
LucGenetier: Do we need to separate "> 0x10FFFF" and [0xD800-0xDFFF] cases?

Maybe a sinple test/error/message is enough

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Unichar.cs
adithyaselv: sounds good to me

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4377,4 +4377,26 @@
     <value>**Disclaimer:** AI-generated content can have mistakes. Make sure it's accurate and appropriate before using it. [See terms](https://go.microsoft.com/fwlink/?linkid=2225491)</value>
     <comment>The disclaimer we show on AI functions. This is github flavored markdown. So ** marks text in bold. </comment>
   </data>
+  <data name="AboutUniChar" xml:space="preserve">
+    <value>Returns the Unicode character that is referenced by the given numeric value.</value>
+    <comment>Description of 'Unichar' function.</comment>
+  </data>
+  <data name="AboutUniChar_number" xml:space="preserve">
+    <value>Numberic value to be converted into Unicode character.</value>
```
CarlosFigueiraMSFT, line 4385: >Numberic [](http://example.com/codeflow?start=11&length=8)

[typo] Numeric

CarlosFigueiraMSFT, line 4385: >into Unicode character [](http://example.com/codeflow?start=42&length=22)

...into **a** Unicode character (missing article)

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Unichar.cs
```diff
@@ -0,0 +1,76 @@
...
+            Contracts.Assert(MinArity <= args.Length && args.Length <= MaxArity);
+
+            var fValid = base.CheckTypes(context, args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
+            Contracts.Assert(returnType.IsTable);
+
+            // Typecheck the input table
+            fValid &= CheckNumericColumnType(context, args[0], argTypes[0], errors, ref nodeToCoercedTypeMap);
+
+            // Synthesize a new return type
+            returnType = DType.CreateTable(new TypedName(DType.String, ColumnName_Value));
```
anderson-joyle, line 71: >ColumnName_Value [](http://example.com/codeflow?start=71&length=16)

Not related to this PR directly but I noticed that we have 2 places where we can get this constant from:
  - TexlFunction.ColumnName_Value (internal)
  - TableValue.ValueName (public)

Should we unify this constant in a single place?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Unichar.cs
```diff
@@ -0,0 +1,76 @@
...
+        }
+    }
+
+    // UniChar(arg:*[n]) : *[s]
+    internal sealed class UniCharTFunction : BuiltinFunction
+    {
+        public override bool IsSelfContained => true;
+
+        public UniCharTFunction()
+            : base("UniChar", TexlStrings.AboutUniCharT, FunctionCategories.Table, DType.EmptyTable, 0, 1, 1, DType.EmptyTable)
```
anderson-joyle, line 46: >DType.EmptyTable [](http://example.com/codeflow?start=110&length=16)

Aren't we expecting a decimal here?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -1187,6 +1187,46 @@ private static DateTime ConvertToUTC(DateTime dateTime, TimeZoneInfo fromTimeZon
...
         }
 
+        public static FormulaValue UniChar(IRContext irContext, NumberValue[] args)
+        {
+            // Excel floors the input number
+            double number = Math.Floor(args[0].Value);
+
+            // Unicode upper and lower bounds
+            const double lowerBound = 0x1; // Excel returns error for UniChar(0)
+            const double upperBound = 0x10FFFF; // 1114111
```
anderson-joyle, line 1197: Any reason why we are not checking these bounds at compilation time?

CarlosFigueiraMSFT: The argument may not be a number; it may be an expression or a variable, so we don't know it in all cases. We could check it at compile time if we have a literal number argument, but we still need a runtime check.

anderson-joyle: Make sense. There is no need for a compilation check in this case.

CarlosFigueiraMSFT (approved):
🚢

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Unichar.cs
adithyaselv: Good point. I think we should ideally replace all occurrences of ColumnName_Value with the ValueName. I can create a follow up clean up PR

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
adithyaselv: @LucGenetier just being consistent with the Excel behavior , as Excel Unichar returns "#VALUE" and "#NA" errors

PR #2134, opened 2024-01-09
Refactor NL2Fx Handling in Language Server to support more dynamic NL Handler Creation based on IPowerFxScope

akshar2401:
This pr performs necessary wiring to enable creating NL Handler for Nl2Fx and FX2NL based on the current IPowerFxScope language server is operating on. This is necessary to support Nl2FX scenarios in Power Apps and the current approach which uses Nl2FxImplementation property is not a thread safe. With the plan to perform Nl2Fx operations in parallel and given that we only create 1 instance of Language Server SDK, two concurrent Nl2FX tasks can overwrite Nl2FxImplementation unexpectedly. The new approach in this pr scopes the creation of handler to every Nl2FX task. It allows the creation of handler based on the scope itself which has "host" specific entities which are required to create a more meaningful NlHandler. For Power Apps, "host" contains document, control and property which are critical to context building for Nl2Fx model inside NL Handler.


src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -35,6 +35,11 @@ public class LanguageServer
     {
         private const char EOL = '\n';
 
+        /// <summary>
+        /// This const represents the dummy formula that is used to create an infrastructure needed to get the symbols for Nl2Fx operation.
+        /// </summary>
+        public const string Nl2FxDummyFormula = "\"Dummy Nl2Fx Check\"";
```
akshar2401, line 41: For Nl2FX, we run scope.Check() multiple times. Initially, we run it on Dummy expression to get access to CheckResult which has necessary infrastructure to build the context (symbol tables). And then we run Check() on expressions returned from Nl2Fx model inside FinalCheck() function. Power Apps needs to distinguish between these Check calls as it doesn't use symbol tables but rather uses traditional name resolver. This change is not harmful yet gives deterministic way to distinguish between Check call being made on a dummy expression vs Check calls inside FinalCheck(). This change is not harmful because it doesn't matter what the dummy expression is. "1" as a dummy expression was too common.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -63,7 +68,9 @@ public class LanguageServer
         /// </summary>
         public NLHandler NL2FxImplementation { get; set; }
```
akshar2401, line 69: Keeping this for backward compat reasons and using it as a backup to factory. If i remove this, then it might introduce compile time error in other consumers of Language Server.

akshar2401: Given we only create one SDK instance, Nl2FxImplementation property being overwritten from the outside is not a thread safe. The factory approach makes the create of NlHandler dynamic and scopes it the IPowerFxInstance which is different for any language server task so we should be thread safe

akshar2401:
Tested this in doc server. no compiler or test failures with these changes

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.LanguageServerProtocol.LanguageServer.#ctor(Microsoft.PowerFx.LanguageServerProtocol.LanguageServer.SendToClient,Microsoft.PowerFx.Core.IPowerFxScopeFactory,System.Action{System.String})

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -63,7 +68,9 @@ public class LanguageServer
...
 
-        public LanguageServer(SendToClient sendToClient, IPowerFxScopeFactory scopeFactory, Action<string> logger = null)
+        internal readonly INLHandlerFactory _nlHandlerFactory;
+
+        public LanguageServer(SendToClient sendToClient, IPowerFxScopeFactory scopeFactory, Action<string> logger = null, INLHandlerFactory nlHandlerFactory = null)
```
jas-valgotar, line 73: > public LanguageServer(SendToClient sendToClient, IPowerFxScopeFactory scopeFactory, Action<string> logger = null, INLHandlerFactory nlHandlerFactory = null) [](http://example.com/codeflow?start=7&length=157)

Can we add new ctor, to avoid binary breaking change? 
 
or

@MikeStall Can we take this breaking change and remove NL2FxImplementation entirely since its not thread safe?

jas-valgotar (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
akshar2401: This doesn't seem to introduce any compile time error in Doc server atleast. For example, Action<void> logger = null was not part of language server ctor initially and it was added as an optional param by Luc. It never broke anything.  The new nlhandler factory is following the same and i do not think it would break any consumer of this sdk

MikeStall: It's a binary-breaking change.  But not a source-breaking change
DocServer didn't hit it because you recompiled - but hosts that just pull updated nuget w/o recompile will break. 
 
Can we make it a init property?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -464,7 +472,7 @@ protected override CustomGetCapabilitiesResult Handle(IPowerFxScope scope, Custo
             {
                 var result = new CustomGetCapabilitiesResult();
 
-                var nl = _parent.NL2FxImplementation;
+                var nl = _parent._nlHandlerFactory?.GetNLHandler(scope) ?? _parent.NL2FxImplementation;
```
MikeStall, line 475: > _parent._nlHandlerFactory?.GetNLHandler(scope) ?? _parent.NL2FxImplementation; [](http://example.com/codeflow?start=24&length=79)

nit - can we put this behind a helper?

     var nl = _parent.GetNLHandler(scope);

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.LanguageServerProtocol.LanguageServer.#ctor(Microsoft.PowerFx.LanguageServerProtocol.LanguageServer.SendToClient,Microsoft.PowerFx.Core.IPowerFxScopeFactory,System.Action{System.String})

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
MikeStall, line 41: >public [](http://example.com/codeflow?start=8&length=6)

Does this need to be public?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -60,9 +65,21 @@ public class LanguageServer
...
+        private INLHandlerFactory _nlHandlerFactory;
+
+        /// <summary>
+        /// A factory to get the NLHandler from the given scope.
+        /// </summary>
+        public INLHandlerFactory NLHandlerFactory
+        {
+            get => _nlHandlerFactory;
+            init => _nlHandlerFactory = value;
+        }
```
MikeStall, line 81: can we simplify as { get; init;l }

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
akshar2401: Made this interval and set it to GUID as we discussed offline

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -35,6 +35,11 @@ public class LanguageServer
     {
         private const char EOL = '\n';
 
+        /// <summary>
+        /// This const represents the dummy formula that is used to create an infrastructure needed to get the symbols for Nl2Fx operation.
+        /// </summary>
+        internal static string Nl2FxDummyFormula = '"' + Guid.NewGuid().ToString() + '"';
```
MikeStall, line 41: >Guid.NewGuid().ToString() [](http://example.com/codeflow?start=57&length=25)

Probably ok for this to be a fixed guid value. 
That may make debugging easier too.

tomjschneider: Using a static GUID had been my interpretation of this as well. That lets us search code for a very specific string to see if they use this.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/BaseNLParams.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT license.
+
+namespace Microsoft.PowerFx.LanguageServerProtocol.Protocol
+{
+    public class BaseNLParams
+    {
+        /// <summary>
+        /// Additional context for NL operation. Usually, a stringified JSON object.
+        /// </summary>
+        public string NLContext { get; set; } = null;
```
tomjschneider, line 11: nit: Maybe just `Context` since it is already under NL specific requests? Dont' feel strongly about it though.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
akshar2401: Okay are you both suggesting to generate a GUID let's say "blahblah" and that becomes our dummy formula for eternity?

tomjschneider: Correct (although not for eternity, just for however long this dummy formula code lives for.)

akshar2401: I guess it would be there all the time now. When i come up with design with more hooks which involves one of the "pre nl2fx" hook during which this dummy formula is used, we would still need to provide the default hook implementation which is just going to be running Check on this dummy formula

akshar2401: Done

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2144, opened 2024-01-16
Add helpers for getting function parameter names.

MikeStall:
Add public helpers for getting function intellisense and parameter names.
These exist on TexlFunction, but is not exposed.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/FunctionInfo.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/FunctionInfo.cs
@@ -2,8 +2,13 @@
 // Licensed under the MIT license.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
+using System.Globalization;
+using System.Linq;
+using System.Text;
 using Microsoft.PowerFx.Core.Functions;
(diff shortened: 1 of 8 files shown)


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/FunctionInfo.cs
```diff
@@ -39,11 +44,98 @@ internal FunctionInfo(TexlFunction fnc)
...
         public string Description => _fnc.Description;
 
+        /// <summary>
+        /// Get a short description of the function.
+        /// </summary>
+        /// <param name="culture"></param>
+        /// <returns></returns>
+        public string GetDescription(CultureInfo culture = null)
+        {
+            culture ??= CultureInfo.InvariantCulture;
```
LucGenetier, line 57: >CultureInfo.InvariantCulture [](http://example.com/codeflow?start=24&length=28)

Shouldn't we get the ParseCultureInfo via FunctionInfo constructor instead of this default?

LucGenetier (approved):
:shipit:

PR #2147, opened 2024-01-16
Fix Formula Bar now showing correct active parameter in Signature Help Regression

akshar2401:
A regression was caused during AI Disclaimer changes that resulted in formula bar always showing first parameter in signature help as active parameter even when a user is not typing a first parameter and is on let's say a fourth parameter.

This happened due to the addition of "ActiveParameter" property in SignatureInformation model. There are two "active parameter" properties as shown below
image

On the left is active parameter for specific signature information which we never had but was introduced with Ai disclaimer change. On the right is a global one that we used and applied to all signatures. 

Active parameter for specific signature info takes precedence over global one so in this case since we introduced this property for each signature info but never set it to anything, it was always being set to "0" causing formula bar to set first parameter as an active parameter

This pr removes that active parameter property for each specific signature info as we never had it and we always used global one

Current Regression: 
image

After this Pr:
image

### DIFF
--- a/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SignatureInformation.cs
+++ b/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/SignatureInformation.cs
@@ -85,12 +85,5 @@ namespace Microsoft.PowerFx.LanguageServerProtocol.Protocol
         /// The parameters of this signature.
         /// </summary>
         public ParameterInformation[] Parameters { get; set; }
-
-        /// <summary>
-        /// The index of the active parameter.
-        ///
-        /// If provided, this is used in place of `SignatureHelp.activeParameter`.
-        /// </summary>
-        public uint ActiveParameter { get; set; }
     }
 }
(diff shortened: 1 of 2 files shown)


MikeStall (approved):
:shipit:

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.LanguageServerProtocol.Protocol.SignatureInformation.get_ActiveParameter<br />❌CP0002 M:Microsoft.PowerFx.LanguageServerProtocol.Protocol.SignatureInformation.set_ActiveParameter(System.UInt32)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.LanguageServerProtocol.Protocol.SignatureInformation.get_ActiveParameter<br />❌CP0002 M:Microsoft.PowerFx.LanguageServerProtocol.Protocol.SignatureInformation.set_ActiveParameter(System.UInt32)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.LanguageServerProtocol.Protocol.SignatureInformation.get_ActiveParameter<br />❌CP0002 M:Microsoft.PowerFx.LanguageServerProtocol.Protocol.SignatureInformation.set_ActiveParameter(System.UInt32)

Issue #2151, opened 2024-01-17
Function Picker enhancements


gekarapa:
In Pad, we are currently working on implementing a function picker. To support this feature:

- [x] there's a need to extend the function info, with the localized function category names. 
- [ ] Additionally, we require a list of unique function info for each function, including a more generic description that accommodates all function's overloads.

MikeStall:
Related to #2004

MikeStall:
Add this:
https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Core/Public/FunctionInfo.cs


string[] FunctionInfo.GetCategoryNames(CultureInfo locale)

anderson-joyle:
> Additionally, we require a list of unique function info for each function, including a more generic description that accommodates all function's overloads.

This change won't be included. Hosts can group by the function name and pick one description sample to cover all overloads.

PR #2154, opened 2024-01-18
Get function category names

anderson-joyle:
Addressing part 1 of https://github.com/microsoft/Power-Fx/issues/2151.
> ...extend the function info, with the localized function category names.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/Utils.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/Utils.cs
@@ -1,7 +1,10 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+using System.Globalization;
 using System.Numerics;
+using Microsoft.PowerFx.Core.Localization;
+using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
 
 namespace Microsoft.PowerFx.Core.Functions
@@ -14,6 +17,11 @@ namespace Microsoft.PowerFx.Core.Functions
             Contracts.Assert(bitIndex >= 0);
 
             return !(value & (BigInteger.One << bitIndex)).IsZero;
+        }
+
+        public static string GetLocalizedName(this FunctionCategories category, CultureInfo culture)
+        {            
+            return StringResources.Get(category.ToString(), culture.Name);
         }
     }
 }
--- a/src/strings/PowerFxResources.en-US.resx
+++ b/src/strings/PowerFxResources.en-US.resx
@@ -2400,33 +2400,43 @@
   </data>
   <data name="None" xml:space="preserve">
     <value>None</value>
+    <comment>Function category name</comment>
   </data>
   <data name="Text" xml:space="preserve">
     <value>Text</value>
+    <comment>Function category name</comment>
   </data>
   <data name="Logical" xml:space="preserve">
     <value>Logical</value>
+    <comment>Function category name</comment>
   </data>
   <data name="Table" xml:space="preserve">
     <value>Table</value>
+    <comment>Function category name</comment>
   </data>
   <data name="Behavior" xml:space="preserve">
     <value>Behavior</value>
+    <comment>Function category name</comment>
   </data>
   <data name="DateTime" xml:space="preserve">
-    <value>Date &amp; Time</value>
+    <value>Date and Time</value>
+    <comment>Function category name</comment>
   </data>
   <data name="MathAndStat" xml:space="preserve">
-    <value>Math &amp; Statistical</value>
+    <value>Math and Statistical</value>
+    <comment>Function category name</comment>
   </data>
   <data name="UserDefined" xml:space="preserve">
-    <value>UserDefined</value>
+    <value>Defined by the user</value>
+    <comment>Function category name</comment>
   </data>
   <data name="Information" xml:space="preserve">
     <value>Information</value>
+    <comment>Function category name</comment>
   </data>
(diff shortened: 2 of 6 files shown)


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/FunctionTests.cs
```diff
@@ -36,6 +38,42 @@ public void TestAllBuiltinFunctionsHaveParameterDescriptions()
...
+                    {
+                        Assert.Contains(StringResources.Get(category.ToString(), locale_enus.Name), categoriesNames_enus);
+                        Assert.Contains(StringResources.Get(category.ToString(), locale_ptbr.Name), categoriesNames_ptbr);
+                    }
+                }
+            }
+        }
+
+        [Fact]
+        public void FunctionCategoriesResourcesTest()
```
MikeStall, line 67: >FunctionCategoriesResourcesTest [](http://example.com/codeflow?start=20&length=31)

this is good.

src/tests/Microsoft.PowerFx.Core.Tests/FunctionTests.cs
```diff
@@ -36,6 +38,42 @@ public void TestAllBuiltinFunctionsHaveParameterDescriptions()
...
+        [Fact]
+        public void FunctionCategoriesLocaleTest()
+        {
+            var texlFunctionsLibrary = BuiltinFunctionsCore.BuiltinFunctionsLibrary;
+            var locale_enus = CultureInfo.CreateSpecificCulture("en-US");
+            var locale_ptbr = CultureInfo.CreateSpecificCulture("pt-BR");
+
+            foreach (var func in texlFunctionsLibrary)
+            {
+                var info = new FunctionInfo(func);
```
MikeStall, line 50: I don't think we need to call on every function in the library - lots of overlap. 
We can just call on one specific one - and assert the actual contents. 

Can we move this to FunctionInfoTests.cs ?

src/strings/PowerFxResources.en-US.resx
```diff
@@ -2398,6 +2398,9 @@
     <value>Invalid json pointer.</value>
     <comment>Generic json pointer parsing error.</comment>
   </data>
+  <data name="None" xml:space="preserve">
+    <value>None</value>
```
CarlosFigueiraMSFT, line 2402: >None [](http://example.com/codeflow?start=11&length=4)

Please add comments explaining the context in which this value will be displayed to the user (and preferably to the other types as well). Without them the translators don't have any context on what the term means, which will cause bad translations (e.g., wrong gender, wrong syntactical category, ...)

CarlosFigueiraMSFT: For example, this is being used as a function category, which should be a feminine noun in Portuguese, but the current translation for the term Logical (2 down) is a masculine adjective.

CarlosFigueiraMSFT:
>     <value>Date &amp; Time</value>

I know this was already there, but can you change it to `Date and time` or `Date & time`? We use sentence casing throughout the platform. Same with `Math & statistical` below

---
Refers to: src/strings/PowerFxResources.en-US.resx:2417 in 55e6a55. [](commit_id = 55e6a55d159464430c547477a354b99e74a3dd6f, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Functions/Utils.cs
```diff
@@ -15,5 +18,10 @@ public static bool TestBit(this BigInteger value, int bitIndex)
 
             return !(value & (BigInteger.One << bitIndex)).IsZero;
         }
+
+        public static string GetLocalizedName(this FunctionCategories category, CultureInfo culture)
```
MikeStall, line 22: >GetLocalizedName [](http://example.com/codeflow?start=29&length=16)

Nice! Extension method is sweet

LucGenetier:
✅ No public API change.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -2398,37 +2398,54 @@
...
+    <comment>Function category name</comment>
```
CarlosFigueiraMSFT, line 2403: >Function category name [](http://example.com/codeflow?start=13&length=22)

Suggestion for names and descriptions

- None: Function category name - function without pre-defined categories
- Text: Function category name - function that works on text values
- Logical: Function category name - function that works on Boolean values
- Table: Function category name - function that works on tabular data
- Behavior: Function category name - function that needs to be executed in behavior contexts (i.e., not purely functional, with side effects)
- Date and time: Function category name - function that works on date/time values
    - Please use "Date and ***time*** (sentence casing)
- Math and statistical: Function category name - functions for mathematical and/or statistical operations
    - Please use "Math and ***statistical*** (sentence casing)
- User-defined: Function category name - function that is defined by the user, not a built-in function
- Information: Function category name - function that returns information about the environment where it is running
- Color: Function category name - function that works on or returns color values
- Service: Function category name - function that comes from a web service, typically using REST (REpresentational State Transfer) communication
- Component: Function category name - function defined in components within the environment where it is running

MikeStall (dismissed):
:shipit:

src/strings/PowerFxResources.en-US.resx
anderson-joyle: Great suggestions. Thanks.

CarlosFigueiraMSFT (approved):
🚢

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #2122, opened 2023-12-21
Unifying Collect function

anderson-joyle:

PR #2156, opened 2024-01-19
Collect function

anderson-joyle:
This is a rebased version of https://github.com/microsoft/Power-Fx/pull/2122, with all comments addressed.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -144,7 +144,7 @@ public static IEnumerable<DType> GetPrimitiveTypes()
         /// Eventually, all display names should come from this centralized source.
         /// We should not be using individual DataSource/OptionSet/View references.
         /// </summary>
-        internal DisplayNameProvider DisplayNameProvider { get; private set; }
+        internal DisplayNameProvider DisplayNameProvider { get; set; }
```
MikeStall, line 147: >set [](http://example.com/codeflow?start=64&length=3)

making this mutable isn't ideal...

anderson-joyle: Thanks for catching this. It was a previous test. I'll run another code clean-up before merging.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Collect.cs
```diff
@@ -0,0 +1,472 @@
...
+            Contracts.AssertValue(argTypes);
+            Contracts.Assert(args.Length == argTypes.Length);
+            Contracts.AssertValue(errors);
+            Contracts.Assert(MinArity <= args.Length && args.Length <= MaxArity);
+
+            var fValid = base.CheckTypes(context, args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
+
+            // Need a collection for the 1st arg
+            DType collectionType = argTypes[0];
+            if (!collectionType.IsTable)
```
MikeStall, line 250: Triple check this ... this is where we could break PA.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Collect.cs
```diff
@@ -0,0 +1,472 @@
...
+        {
+            return CreateInvariantFieldName(features, dKind);
+        }
+
+        public override DType GetCollectedType(Features features, DType argType)
+        {
+            return GetCollectedTypeForGivenArgType(features, argType);
+        }
+
+        internal override IntermediateNode CreateIRCallNode(PowerFx.Syntax.CallNode node, IRTranslator.IRTranslatorContext context, List<IntermediateNode> args, ScopeSymbol scope)
```
MikeStall, line 445: >CreateIRCallNode [](http://example.com/codeflow?start=43&length=16)

This is interesting. 

@jas-valgotar , fyi

src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationFunctionsTests.cs
```diff
@@ -364,6 +364,52 @@ public void DontCopyDerivedRecordValuesTest(string expression)
...
+            var check = engine.Check(expression, options: new ParserOptions() { AllowsSideEffects = true });
+
+            // Compilation will be successful, but the function will not be executed.
+            // This is because PA depends on the CheckType to determine if the function is valid.
+            Assert.True(check.IsSuccess);
+
+            var evaluator = check.GetEvaluator();
+
+            // Runtime exception
+            Assert.Throws<AggregateException>(() => evaluator.Eval());
```
MikeStall, line 410: >AggregateException [](http://example.com/codeflow?start=26&length=18)

we can avoid AggregateException if we use ThrowsAsync<>  and await

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Collect.cs
jas-valgotar: What is this doing? can you also add comment?
Is it Converting
Collect(t, 1, 2,) -> Collect(t, {Value: 1}, {Value: 2}) ?

jas-valgotar: Also is there a test for it?

src/libraries/Microsoft.PowerFx.Core/Utils/MutationUtils.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        /// <summary>
+        /// Adds specific errors for mutation functions.
+        /// </summary>
+        public static void CheckSemantics(TexlBinding binding, TexlFunction function, TexlNode[] args, DType[] argTypes, IErrorContainer errors)
+        {
+            var targetArg = args[0];
+
+            // control!property are not valid targets for Collect, Remove, etc.
```
CarlosFigueiraMSFT, line 45: >control!property [](http://example.com/codeflow?start=15&length=16)

`control.property` - we deprecated the `!` operator

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Collect.cs
anderson-joyle: > What is this doing? can you also add comment? Is it Converting Collect(t, 1, 2,) -> Collect(t, {Value: 1}, {Value: 2}) ?

Correct. PA handles scalar values at runtime. We are transforming scalars into `{Value:n}` in IR.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Collect.cs
```diff
@@ -0,0 +1,472 @@
...
+
+        public override bool TryGetDataSourceNodes(PowerFx.Syntax.CallNode callNode, TexlBinding binding, out IList<FirstNameNode> dsNodes)
+        {
+            Contracts.AssertValue(callNode);
+            Contracts.AssertValue(binding);
+
+            dsNodes = new List<FirstNameNode>();
+            if (callNode.Args.Count != 2)
+            {
+                return false;
```
CarlosFigueiraMSFT, line 347: >return false; [](http://example.com/codeflow?start=16&length=13)

If this returns false, the out argument should not be initialized (i.e., dsNode = null)

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Collect.cs
```diff
@@ -0,0 +1,472 @@
...
+        public static DType GetCollectedTypeForGivenArgType(Features features, DType argType, TexlNode arg, ref Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
+        {
+            var singleColumnRecordType = GetCollectedTypeForGivenArgType(features, argType);
+
+            if (!argType.IsPrimitive)
+            {
+                return argType;
+            }
+
+            //CollectionUtils.Add(ref nodeToCoercedTypeMap, arg, singleColumnRecordType);
```
CarlosFigueiraMSFT, line 399: > //CollectionUtils.Add(ref nodeToCoercedTypeMap, arg, singleColumnRecordType); [](http://example.com/codeflow?start=11&length=78)

Remove

src/libraries/Microsoft.PowerFx.Core/Utils/MutationUtils.cs
anderson-joyle: This was copied from PA.

CarlosFigueiraMSFT: In Power Apps we still support the `!` operator for legacy purposes; in Power Fx (C#) we don't.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Collect.cs
```diff
@@ -0,0 +1,472 @@
...
+            DType itemType = DType.Invalid;
+            DType datasourceType = argTypes[0];
+
+            var argc = args.Length;
+
+            for (var i = 1; i < argc; i++)
+            {
+                DType argType = GetCollectedType(features, argTypes[i]);
+
+                // !!! How is it possible for an argtype to be a primitive and an aggregate at the same time?
```
CarlosFigueiraMSFT, line 185: >// !!! How is it possible for an argtype to be a primitive and an aggregate at the same time? [](http://example.com/codeflow?start=16&length=93)

Can you elaborate on this question?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Collect.cs
```diff
@@ -0,0 +1,472 @@
...
+
+        public override bool IsAsyncInvocation(PowerFx.Syntax.CallNode callNode, TexlBinding binding)
+        {
+            Contracts.AssertValue(callNode);
+            Contracts.AssertValue(binding);
+
+            return Arg0RequiresAsync(callNode, binding);
+        }
+
+        public static DType GetCollectedTypeForGivenArgType(Features features, DType argType, TexlNode arg, ref Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
```
CarlosFigueiraMSFT, line 390: >GetCollectedTypeForGivenArgType [](http://example.com/codeflow?start=28&length=31)

This is only used in CollectScalarFunction; does it need to be in this base class?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Collect.cs
```diff
@@ -0,0 +1,472 @@
...
+        public static DType GetCollectedTypeForGivenArgType(Features features, DType argType)
+        {
+            Contracts.Assert(argType.IsValid);
+
+            if (!argType.IsPrimitive)
+            {
+                return argType;
+            }
+
+            // Passed a scalar; make a record out of it, using a name that depends on the type.
```
CarlosFigueiraMSFT, line 413: >, using a name that depends on the type. [](http://example.com/codeflow?start=55&length=40)

This part of the comment can be changed, since the name is only 'Value' now (wrapped record is {Value:<primitive>})

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Collect.cs
```diff
@@ -0,0 +1,472 @@
...
+
+        public override bool ManipulatesCollections => true;
+
+        public override bool ModifiesValues => true;
+
+        public override bool IsSelfContained => false;
+
+        public override bool RequiresDataSourceScope => true;
+
+        public virtual bool IsScalar => false;
```
CarlosFigueiraMSFT, line 42: >IsScalar [](http://example.com/codeflow?start=28&length=8)

This is not used anywhere

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Collect.cs
```diff
@@ -0,0 +1,467 @@
...
+        }
+
+        public virtual DType GetCollectedType(Features features, DType argType)
+        {
+            Contracts.Assert(argType.IsValid);
+
+            return argType;
+        }
+
+        public bool TryGetUnifiedCollectedTypeCanvas(TexlNode[] args, DType[] argTypes, IErrorContainer errors, Features features, out DType collectedType)
```
CarlosFigueiraMSFT, line 113: >public [](http://example.com/codeflow?start=8&length=6)

This doesn't seem to be used in any other class; if that's the case, please make it private. Same for V1

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalDataSource.cs
```diff
@@ -19,6 +19,8 @@ internal interface IExternalDataSource : IExternalEntity, IExternalPageableSymbo
 
         bool RequiresAsync { get; }
 
+        bool IsWritable { get; }
```
anderson-joyle, line 22: >bool IsWritable { get; } [](http://example.com/codeflow?start=8&length=24)

Originally pushed by [https://github.com/microsoft/Power-Fx/pull/2246](https://github.com/microsoft/Power-Fx/pull/2246). Including it here since it was meant to fix things in this PR anyway. This does not break PA.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -2873,12 +2873,20 @@ public override void Visit(FirstNameNode node)
...
+                    }
+                    else if (lookupInfo.Data is IExternalDataSource ds)
+                    {
+                        _txb.SetMutable(node, ds.IsWritable);
+                    }
+                }
+                else if (lookupInfo.Kind == BindKind.ScopeCollection)
+                {
+                    _txb.SetMutable(node, true);
                 }
```
anderson-joyle, line 2890: These have been added since PFx introduced a new semantic check for mutable args. Some of mutable collection that PA creates were failing the semantic check.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Collect.cs
```diff
@@ -0,0 +1,464 @@
...
+        /// <summary>
+        /// Since Arg1 and Arg2 depends on type of Arg1 return false for them.
+        /// </summary>
+        public override bool TryGetTypeForArgSuggestionAt(int argIndex, out DType type)
+        {
+            if (argIndex == 0)
+            {
+                type = DType.EmptyTable;
+                return true;
+            }
```
anderson-joyle, line 59: This differs from PA. PA handles suggestions for functions with `ParmTypes == 0` differently.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Collect.cs
anderson-joyle: Collect always expects a table as arg0.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Collect.cs
```diff
@@ -0,0 +1,490 @@
...
+                binding.EntityScope.TryGetDataSource(args[0], out IExternalDataSource dataSourceInfo) &&
+                !dataSourceInfo.IsClearable)
+            {
+                errors.EnsureError(args[0], TexlStrings.ErrInvalidDataSourceForFunction);
+            }
+        }
+    }
+
+    // ClearCollect(collection:*[...], item1, ...)
+    internal class ClearCollectScalarFunction : ClearCollectFunction
```
anderson-joyle, line 476: >ClearCollectScalarFunction [](http://example.com/codeflow?start=19&length=26)

// !!!TODO [TexlRuntimeNameOverride(Suffix = "_S")]

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -165,8 +165,24 @@ private async Task<FormulaValue> TryHandleSet(CallNode node, EvalVisitorContext
             var arg0 = node.Args[0];
             var arg1 = node.Args[1];
 
-            var newValue = await arg1.Accept(this, context).ConfigureAwait(false);
-
+            var newValue = await arg1.Accept(this, context).ConfigureAwait(false);
+
+            if (arg0.IRContext.IsMutation && arg0 is RecordFieldAccessNode rfan)
```
MikeStall, line 170: >IRContext [](http://example.com/codeflow?start=21&length=9)

Some of these look like previous changes? Is this rebased on latest?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
```diff
@@ -229,6 +232,111 @@ public async Task<FormulaValue> InvokeAsync(FormulaType irContext, FormulaValue[
...
+        {
+            return await new CollectProcess().Process(irContext, args, cancellationToken).ConfigureAwait(false);
+        }
+    }
+
+    internal class CollectScalarImpl : CollectScalarFunction, IAsyncTexlFunction3
+    {
+        public async Task<FormulaValue> InvokeAsync(FormulaType irContext, FormulaValue[] args, CancellationToken cancellationToken)
+        {
+            return await new CollectProcess().Process(irContext, args, cancellationToken).ConfigureAwait(false);
```
MikeStall, line 247: >new CollectProcess().Process [](http://example.com/codeflow?start=25&length=28)

nit: should this function just be static?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
```diff
@@ -229,6 +232,111 @@ public async Task<FormulaValue> InvokeAsync(FormulaType irContext, FormulaValue[
...
+            }
+
+            if (arg0 is not TableValue)
+            {
+                return CommonErrors.RuntimeTypeMismatch(IRContext.NotInSource(arg0.Type));
+            }
+
+            var tableValue = arg0 as TableValue;
+
+            List<DValue<RecordValue>> resultRows = new List<DValue<RecordValue>>();
```
MikeStall, line 286: > List<DValue<RecordValue>> resultRows [](http://example.com/codeflow?start=11&length=37)

interpreter only has v1 semantics. 
Collect() always returns single-record in v1.

MikeStall (approved):

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Collect.cs
```diff
@@ -0,0 +1,492 @@
...
+            {
+                errors.EnsureError(DocumentErrorSeverity.Severe, args[0], TexlStrings.ErrInvalidArgs_Func, Name);
+                fValid = false;
+            }
+
+            DType collectedType = null;
+
+            // Get the unified collected type on the RHS. This will generate appropriate
+            // document errors for invalid arguments such as unsupported aggregate types.
+            if (context.Features.PowerFxV1CompatibilityRules && !context.AnalysisMode)
```
CarlosFigueiraMSFT, line 268: >context.Features.PowerFxV1CompatibilityRules &&  [](http://example.com/codeflow?start=16&length=48)

I don't think this is necessary. Regardless of whether we're in V1 mode, we should only allow type expansion if we're in analysis mode.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Collect.cs
```diff
@@ -0,0 +1,492 @@
...
+            {
+                fValid &= TryGetUnifiedCollectedTypeV1(args, argTypes, errors, context.Features, out collectedType);
```
CarlosFigueiraMSFT, line 270: >TryGetUnifiedCollectedTypeV1 [](http://example.com/codeflow?start=26&length=28)

Suggestion: rename to TryGetUnifiedCollectedTypeDontAllowTypeExpansion (and the next TryGetUnifiedCollectedTypeAllowTypeExpansion) - or, if it won't be too complicated, merge the two functions and have the 'allowTypeExpansion' as an argument

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Collect.cs
```diff
@@ -0,0 +1,492 @@
...
+            }
+            else
+            {
+                fValid &= TryGetUnifiedCollectedTypeCanvas(args, argTypes, errors, context.Features, out collectedType);
```
CarlosFigueiraMSFT, line 274: >TryGetUnifiedCollectedTypeCanvas [](http://example.com/codeflow?start=26&length=32)

See other comment for naming suggestion. Other hosts may, in the future, allow for type expansion as well

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Collect.cs
```diff
@@ -0,0 +1,492 @@
...
+            if (fError)
+            {
+                fValid = false;
+                if (!SetErrorForMismatchedColumns(collectionType, collectedType, args[1], errors, context.Features))
+                {
+                    errors.EnsureError(DocumentErrorSeverity.Severe, args[0], TexlStrings.ErrTableDoesNotAcceptThisType);
+                }
+            }
+
+            if (returnType.IsAggregate)
```
CarlosFigueiraMSFT, line 290: >returnType.IsAggregate [](http://example.com/codeflow?start=16&length=22)

Is it necessary? When would the returnType **not** be an aggregate?

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -646,6 +646,8 @@ public bool IsActivityPointer
 
         public bool HasPolymorphicInfo => PolymorphicInfo != null;
 
+        public bool IsSingleColumnTable => IsTable && GetNames(DPath.Root).Count() == 1;
```
CarlosFigueiraMSFT, line 649: >IsSingleColumnTable [](http://example.com/codeflow?start=20&length=19)

Not used anywhere. Can we remove it?

src/libraries/Microsoft.PowerFx.Core/Utils/MutationUtils.cs
```diff
@@ -65,5 +70,44 @@ public static void CheckSemantics(TexlBinding binding, TexlFunction function, Te
                 return;
             }
         }
+
+        public static string GetScalarSingleColumnNameForType(Features features, DKind kind)
```
CarlosFigueiraMSFT, line 74: >GetScalarSingleColumnNameForType [](http://example.com/codeflow?start=29&length=32)

Since this is only used in the CollectFunction, can we move it there (as a private method)? There's no need to share it with others if it is not used by more than one function.

src/libraries/Microsoft.PowerFx.Core/Utils/MutationUtils.cs
```diff
@@ -65,5 +70,44 @@ public static void CheckSemantics(TexlBinding binding, TexlFunction function, Te
...
+                DKind.Hyperlink or
+                DKind.Media or
+                DKind.Blob or
+                DKind.PenImage => features.ConsistentOneColumnTableResult ? TableValue.ValueName : "Url",
+
+                _ => TableValue.ValueName
+            };
+        }
+
+        public static List<IntermediateNode> CreateIRCallNodeCollect(CallNode node, IRTranslator.IRTranslatorContext context, List<IntermediateNode> args, ScopeSymbol scope)
```
CarlosFigueiraMSFT, line 88: >CreateIRCallNodeCollect [](http://example.com/codeflow?start=45&length=23)

Same question as above - since this is only used by CollectFunction, can it live there?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
```diff
@@ -229,6 +232,111 @@ public async Task<FormulaValue> InvokeAsync(FormulaType irContext, FormulaValue[
...
+                }
+                else if (arg is ErrorValue)
+                {
+                    return arg;
+                }
+            }
+
+            if (resultRows.Count == 0)
+            {
+                return FormulaValue.NewBlank(arg0.Type);
```
CarlosFigueiraMSFT, line 326: >arg0.Type [](http://example.com/codeflow?start=45&length=9)

This may not be the correct type for all cases. Assuming that `coll` is a collection with schema `*[a:s]`, then those should be the result for the following expressions:

- `Collect(coll, If(1<0, { a:"hello" }))`
  - `![a:s]` - single record, record result
- `Collect(coll, If(1<0, [{ a:"hello" }]))`
  - `*[a:s]` - table or multiple records, table result

CarlosFigueiraMSFT:
>             var engine = new Engine(new PowerFxConfig(isPFxV1 ? Features.PowerFxV1 : Features.None));

Do we need this?

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests.Shared/TexlTests.cs:147 in cb757a3. [](commit_id = cb757a3b949ac86a58d1d2bae68551cccb52545d, deletion_comment = False)

CarlosFigueiraMSFT:
>             engine.Config.SymbolTable.AddFunction(new PatchAggregateSingleTableFunction());

[minor] can we just call EnableMutationFunctions instead of adding each function individually? If we don't call those, then we don't need the Patch functions either.

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests.Shared/TexlTests.cs:158 in cb757a3. [](commit_id = cb757a3b949ac86a58d1d2bae68551cccb52545d, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
anderson-joyle: It is.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/TexlTests.cs
```diff
@@ -1,4457 +1,4487 @@
...
+        [InlineData("Collect(t2, 2)", "![Value:w]")]
+        [InlineData("Collect(t2, 2, 3, 4)", "*[Value:w]")]
+        [InlineData("Collect(t1, {a:1})", "*[a:w, b:s]", false)]
+        [InlineData("Collect(t1, {a:1},{a:2})", "*[a:w, b:s]", false)]
+        [InlineData("Collect(t1, Table({a:1}))", "*[a:w, b:s]", false)]
+        [InlineData("Collect(t2, 2)", "*[Value:w]", false)]
+        [InlineData("Collect(t2, 2, 3, 4)", "*[Value:w]", false)]
+        public void TexlMutationFunctionsV1Tests(string expression, string expectedType, bool isPFxV1 = true)
+        {
+            var engine = new Engine(new PowerFxConfig(isPFxV1 ? Features.PowerFxV1 : Features.None));
```
CarlosFigueiraMSFT, line 147: Do we need to run tests in different PFx versions? The new "AnalysisMode" should be the one that makes it different.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/TexlTests.cs
```diff
@@ -1,4457 +1,4487 @@
...
+            var options = new ParserOptions() { AllowsSideEffects = true };
+
+            DType.TryParse("*[a:w,b:s]", out var expectedDType);
+            DType.TryParse("*[Value:w]", out var expectedDTypeScalar);
+
+            engine.Config.SymbolTable.AddFunction(new CollectFunction());
```
CarlosFigueiraMSFT, line 153: [minor] can we just call EnableMutationFunctions instead of adding each function individually? If we don't call those, then we don't need the Patch functions either.

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/MutationScripts/ClearCollect.txt
```diff
@@ -86,7 +86,19 @@ Errors: Error 22-24: The specified column 'a' does not exist.|Error 0-12: The fu
...
 >> ClearCollect(If(false,Blank()),r1)
 Errors: Error 31-33: The specified column 'a' does not exist.|Error 0-12: The function 'ClearCollect' has some invalid arguments.
+
+>> Set(t2, [1,2,3])
+Table({Value:1},{Value:2},{Value:3})
+
+>> ClearCollect(t2, 99)
+{Value:99}
+
+>> 0;t2
```
CarlosFigueiraMSFT, line 100: >0;t2 [](http://example.com/codeflow?start=3&length=4)

Curious: why `0;t2` and not just `t2`?

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/MutationScripts/Coalesce_CoalesceShortCircuit.txt
```diff
@@ -1,3 +1,4 @@
+#SETUP: PowerFxV1CompatibilityRules
```
CarlosFigueiraMSFT, line 1: >PowerFxV1CompatibilityRules [](http://example.com/codeflow?start=8&length=27)

Curious: why do we need this? Same question for other test files with this single change

src/libraries/Microsoft.PowerFx.Core/Binding/BindingConfig.cs
```diff
@@ -19,13 +19,16 @@ internal class BindingConfig
 
         public bool UseThisRecordForRuleScope { get; }
 
-        public bool NumberIsFloat { get; }
+        public bool NumberIsFloat { get; }
+
+        public bool AnalysisMode { get; }
```
lesaltzm, line 24: Probably would be better as an enum that "self documents" a bit better. 

Maybe { InferredTypesAllowed, DeclaredTypesOnly } or something like that.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Collect.cs
```diff
@@ -0,0 +1,492 @@
...
+        public override bool ArgMatchesDatasourceType(int argNum)
+        {
+            return argNum >= 1;
+        }
+
+        public override bool IsLazyEvalParam(TexlNode node, int index, Features features)
+        {
+            // First argument to mutation functions is Lazy for datasources that are copy-on-write.
+            // If there are any side effects in the arguments, we want those to have taken place before we make the copy.
+            return index == 0;
```
lesaltzm, line 82: Does this change codegen for canvas?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
```diff
@@ -229,6 +232,111 @@ public async Task<FormulaValue> InvokeAsync(FormulaType irContext, FormulaValue[
...
+    {
+        public async Task<FormulaValue> InvokeAsync(FormulaType irContext, FormulaValue[] args, CancellationToken cancellationToken)
+        {
+            return await new CollectProcess().Process(irContext, args, cancellationToken).ConfigureAwait(false);
+        }
+    }
+
+    internal class CollectProcess
+    {
+        internal async Task<FormulaValue> Process(FormulaType irContext, FormulaValue[] args, CancellationToken cancellationToken)
```
lesaltzm, line 253: Should this be static? Doesn't seem like you need to construct the instance of collectprocess.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -3146,32 +3146,36 @@
     <comment>Display text representing the Contrast value of TeamsTheme enum (TeamsTheme_Contrast_Name). The possible values for this enumeration are: Default, Dark, Contrast.</comment>
   </data>
   <data name="AboutCollect" xml:space="preserve">
-    <value>Adds records to a data source.</value>
-    <comment>Description of 'Clear' function.</comment>
+    <value>Adds one or more items to the specified 'collection'. The items can be from a different table or collection (e.g. Collect(collection, source_collection)), or one or more records Collect(collection, {key1: val1, key2: val2, ...}, ...).</value>
+    <comment>Description of 'Collect' function.</comment>
```
lesaltzm, line 3150: Lock Collect? And do we need ticks around collection?

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4623,6 +4627,16 @@
       A partial operator is the 2nd part of a statement `[Partial Op]` and can be one of "And", "Or", "Table" or "Record".
       It's used to determine how to combine multiple expressions with the same name and operator.</comment>
   </data>
+  <data name="AboutCollect_data_source" xml:space="preserve">
+    <value>The data source that you want to add data to.</value>
```
lesaltzm, line 4631: not sure if we want "you" in this description. Maybe "The datasource to add data to"

src/tests/Microsoft.PowerFx.Core.Tests.Shared/TexlTests.cs
anderson-joyle: These particular test cases test if the return type will be a table or record. This result depends on the PFxV1 flag, not the analysis flag.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/TexlTests.cs
anderson-joyle: `EnableMutationFunctions` is an extension method from the Interpreter. These are Core tests.

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/MutationScripts/ClearCollect.txt
anderson-joyle: This is to avoid getting a "duplicated test case" error. So every time we want to check the content of a variable, we call `0;t2`, `1;t2`, and so on.

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/MutationScripts/Coalesce_CoalesceShortCircuit.txt
anderson-joyle: So the test cases in these files will return the expected type (table or record).

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Collect.cs
anderson-joyle: It didn't affect any test cases in Canvas. Can you think of a specific test in Canvas so we can check this?

MikeStall (approved):

PR #2157, opened 2024-01-20
Added Count to ErrorContainer

petrochuk:
Issue: During tight loops we sometimes need only error count and at this time we can only call GetErrors() which returns IEnumerable and the iterate though each error using Count() extension. It would improve performance if we can have native Count property


anderson-joyle (approved):
:shipit:

PR #764, opened 2022-10-14
Add support to selectively remove functions from the default symbol table

anderson-joyle:
Issue #588

PR #2158, opened 2024-01-22
Don't suggestion removed functions

anderson-joyle:
Complement to https://github.com/microsoft/Power-Fx/pull/764


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/IntellisenseTests/SuggestTest.cs
```diff
@@ -594,5 +612,19 @@ public override int GetHashCode()
                 return 1;
             }
         }
+
+        // Helper to set Engine.SupportedFunctions.
+        private class Engine2 : Engine
```
MikeStall, line 617: We don't really want products to need to derive from an Engine just to update the functions. Let's consider another pattern.

PR #2159, opened 2024-01-22
Create public pretty print to use in NL2FX

nguhoa:


src/libraries/Microsoft.PowerFx.Interpreter/PrettyPrint.cs
```diff
@@ -0,0 +1,20 @@
...
+using System;
+using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Core.Parser;
+using Microsoft.PowerFx.Syntax;
+
+namespace Microsoft.PowerFx.Interpreter
+{
+    public class PrettyPrint
+    {
+        public static string GetPrettyPrint(string input)
```
CarlosFigueiraMSFT, line 13: >GetPrettyPrint [](http://example.com/codeflow?start=29&length=14)

We need to have a better name for this. Calling `PrettyPrint.GetPrettyPrint` does not follow the rest of the project.

src/libraries/Microsoft.PowerFx.Interpreter/PrettyPrint.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Core.Parser;
+using Microsoft.PowerFx.Syntax;
+
+namespace Microsoft.PowerFx.Interpreter
```
CarlosFigueiraMSFT, line 9: >Interpreter [](http://example.com/codeflow?start=28&length=11)

Should this be in the interpreter or in the core? We don't need the interpreter to parse an expression (it can be done entirely in core)

src/tests/Microsoft.PowerFx.Interpreter.Tests/PrettyPrintTests.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Collections.Generic;
+using System.Text;
+using Xunit;
+
+namespace Microsoft.PowerFx.Interpreter.Tests
+{
+    public class PrettyPrintTests
+    {
+        [Theory]
+        [InlineData("((1 + 2) + 3)", "1 + 2 + 3")]
```
CarlosFigueiraMSFT, line 14: >((1 + 2) + 3)", "1 + 2 + 3" [](http://example.com/codeflow?start=21&length=27)

Looking at the tests, is it really doing pretty printing, or only removing unnecessary parentheses? If the latter, we should rename it. If the former, we should have more comprehensive tests.

src/libraries/Microsoft.PowerFx.Interpreter/PrettyPrint.cs
nguhoa: I don't think we need this in core. I need it to be used in NL2FX and I used it for our final post processing expression so I want input is expression

src/tests/Microsoft.PowerFx.Interpreter.Tests/PrettyPrintTests.cs
nguhoa: Trying to test if it remove non necessary characters in expression. We already have existing test for multiple cases in public void TestPrettyPrint(string script, string expected) in FormatterTests file. 
The reason I create this function so I can call it from our internal powerfx repo for NL2FX purpose. The original prettyprint function is internal that cannot be used.
Update the name to be less confused.

src/libraries/Microsoft.PowerFx.Interpreter/PrettyPrint.cs
MikeStall: This should be in Core.  But probably on an existing class. I can look around...

(NL2Fx doesn't use interpreter. )

MikeStall: actually, it looks like TexlNode.ToString() already does this. 
So we don't need this function.

nguhoa: Thanks to point out. abandon this PR since don't need.

PR #2160, opened 2024-01-23
Block void values in the pattern / matches arguments in the Switch function

CarlosFigueiraMSFT:
With the completion of #2096 we started accepting void arguments in the pattern (first argument) and matches (even-numbered arguments) for the Switch function, so something like this would not have any compile-time errors: `Switch(If(1<0, 1, {}), 1, "ok", "not ok")`. This is an invalid expression - the first argument (a void value) cannot be compared with anything else. This fixes that, and adds some additional tests to prevent this from happening again.


LucGenetier:
✅ No public API change.

anderson-joyle (approved):
:shipit:

PR #2162, opened 2024-01-23
Allow access to design properties via Gallery.Selected

lesaltzm:
This removes a pointless restriction on property access via Gallery.Selected.
These properties were already accessible directly on children of the gallery, and this restriction doesn't have any runtime justification for existing. It also blocks deprecation of primary output properties (Icon has a design primary output prop, others might as well)


CarlosFigueiraMSFT (approved):
🚢

LucGenetier:
✅ No public API change.

PR #2164, opened 2024-01-24
Connector Yaml comparer

LucGenetier:
Add Yaml comparer and load results in SQL
Update OpenApiReaderSettings in swagger readers to exclude KeyMustBeRegularExpression rule


LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #2165, opened 2024-01-24
[Draft/Prototype] Add "attribute-like" support to Named Formulas

lesaltzm:


LucGenetier:
✅ No public API change.

PR #2168, opened 2024-01-25
[Connectors] Add warnings

LucGenetier:
When swagger file type "binary" is used, we are now generating a warning that we might return incorrect results as we default to a string
Updated Yaml comparer accordingly


LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Connectors.SupportsConnectorErrors.#ctor(System.String)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Connectors.SupportsConnectorErrors.#ctor(System.String)

src/libraries/Microsoft.PowerFx.Connectors/Internal/ConnectorErrors.cs
```diff
@@ -10,8 +12,8 @@ public ConnectorErrors()
         { 
         }
 
-        public ConnectorErrors(string error)
-            : base(error)
+        public ConnectorErrors(string error, ErrorResourceKey warning = default)
```
MikeStall, line 15: >, ErrorResourceKey warning = default) [](http://example.com/codeflow?start=43&length=37)

Can we do in non-breaking way?

Public API changes:
❌CP0002 M:Microsoft.PowerFx.Connectors.SupportsConnectorErrors.#ctor(System.String)

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4051,6 +4051,10 @@
     <value>In namespace {1}, function {0} is deprecated.</value>
     <comment>A warning message when a function is deprecated.</comment>
   </data>
+  <data name="WarnFunctionUseBinary" xml:space="preserve">
+    <value>This function uses 'binary' format and might not work as expected.</value>
+    <comment>A warning message when a function is using binary type.</comment>
+  </data>
```
MikeStall, line 4057: We'd like to have in connector resource, but mechanics aren't working. 

We did something similar for DV, but for some reason, not working here:
[https://github.com/microsoft/Power-Fx-Dataverse/pull/247](https://github.com/microsoft/Power-Fx-Dataverse/pull/247)

src/libraries/Microsoft.PowerFx.Connectors/Internal/ConnectorErrors.cs
LucGenetier: internal class, no one is using this class

src/strings/PowerFxResources.en-US.resx
LucGenetier: done

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Connectors.SupportsConnectorErrors.#ctor(System.String)

MikeStall (approved):

Issue #2161, opened 2024-01-23
Text first expression tokens


gekarapa:
To enable syntax highlighting in Pad effectively, it is crucial to distinguish between the literal string parts of the root Text first expression and those present in interpolation expressions. This distinction is necessary for addressing scenarios that demand specific (or different) highlighting for the literal string parts within the text-first expressions.
![Screenshot 2024-01-23 175542](https://github.com/microsoft/Power-Fx/assets/133134761/2f1184b7-bc27-4d77-853c-060f116996b7)

```[tasklist]
### Suggestions
- [ ] Introduce a new `TokenType`, such as `TextFirstStrLit`, to represent the string literal parts of the text-first expressions.
- [ ] Introduce TextFirstStrInterpStart and TextFirstStrInterpEnd can help mark the beginning and end of interpolation expressions.
```

MikeStall:
Note that all this information is available through the LSP today. 

So we should see how LSP handles it, and make that lower layer public.

anderson-joyle:
This also will cover https://github.com/microsoft/Power-Fx/issues/2155.

PR #2171, opened 2024-01-26
New TextFirst token kinds to syntax highlight

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2161.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Lexer/TokKind.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Lexer/TokKind.cs
@@ -48,6 +48,12 @@ namespace Microsoft.PowerFx.Syntax
         StrLit,
 
         /// <summary>
+        /// String literal when TextFirst is active.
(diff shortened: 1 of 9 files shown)


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Lexer/TokKind.cs
```diff
@@ -45,7 +45,13 @@ public enum TokKind
...
         /// <code>"Hello world"</code>
         /// </summary>
-        StrLit,
+        StrLit,
+
+        /// <summary>
+        /// String literal when TextFirst is active.
+        /// <code>Hello world</code>
+        /// </summary>
+        TextFirstStrLit,
```
MikeStall, line 54: adding new token kinds is a big deal. 
We should chat.

PR #2172, opened 2024-01-29
New TextFirst flag for syntax highlight

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2161

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Lexer/ITextFirstFlag.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Lexer/ITextFirstFlag.cs
@@ -1,0 +1,17 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.PowerFx.Syntax
+{
+    /// <summary>
+    /// Indicates that the object is in active TextFirst context.
+    /// </summary>
+    public interface ITextFirstFlag
+    {
+        bool IsTextFirst { get; }
+    }
+}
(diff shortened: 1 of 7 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Lexer/ITextFirstFlag.cs
```diff
@@ -0,0 +1,17 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.PowerFx.Syntax
+{
+    /// <summary>
+    /// Indicates that the object is in active TextFirst context.
+    /// </summary>
+    public interface ITextFirstFlag
```
MikeStall, line 13: >public [](http://example.com/codeflow?start=4&length=6)

let's remove interface and just keep on STrLitToken directly

src/tests/Microsoft.PowerFx.Core.Tests/LexerTests.cs
```diff
@@ -605,5 +605,30 @@ public void TextFirstBasicFormulaStrLit()
             Assert.True(lit.Span.Min == 1);
             Assert.True(lit.Span.Lim == str.Length);
         }
+
+        [Theory]
+        [InlineData("Hello ${\"Word\"}", 0, 1, 5)]
```
MikeStall, line 610: >0, 1, 5 [](http://example.com/codeflow?start=41&length=7)

per discussion, there are some tricks to make this test more maintainable

MikeStall (dismissed):
:shipit:

LucGenetier:
✅ No public API change.

PR #2173, opened 2024-01-30
New IRVisitor test to show how we could escape HTML

CarlosFigueiraMSFT:
Trying to implement a scenario using IR rewriting to handle HTML escaping automatically:
- Literal strings are not escaped
- Variables are escaped (can potentially have unsafe data)
- Data from tables are escaped (can potentially have unsafe data)
- IR tree is traversed and whenever there are unsafe string nodes being concatenated, those nodes are escaped.


MikeStall (approved):

src/tests/Microsoft.PowerFx.Interpreter.Tests/IRTests/RewritingIRVisitorTests.cs
```diff
@@ -239,5 +242,438 @@ public override IntermediateNode Visit(DecimalLiteralNode node, Context context)
...
+            var objActual = result.ToObject();
+            if (objActual is not string)
+            {
+                objActual = objActual.ToString();
+            }
+
+            Assert.Equal(objActual, expectedResult);
+        }
+
+        private class EscapeHtmlTransform : IRTransform
```
MikeStall, line 326: Great test. 
Looks like it's using the same infra we use for IRrewriting for delegation .

MikeStall (approved):

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/IRTests/RewritingIRVisitorTests.cs
CarlosFigueiraMSFT: 👍

PR #2174, opened 2024-01-30
Remove time zone conversion from DateValue parsing

rsullivan00:
This PR removes time zone conversion from `DateValue` parsing. 

This fixes issues with `DateValue(str)` being 1-day off for GMT time zone during Daylight Savings Time, which also caused issues with `DateAdd(str, 1)` in GMT.

I don't see why we would need to consider time zones in date (not datetime) parsing, but please let me know if I'm missing something and I can restore any scenarios that would be broken with this change.

### Fixed scenarios

The following tests included in this PR were failing with these results. All are GMT time zone tests.
```
// FAILING, actual result is DateTime(2024,10,27,23,0,0,0)
>> DateAdd("2024-10-27", 1)
DateTime(2024,10,28,0,0,0,0)

// FAILING, actual result is Date(2024,10,25)
>> DateValue("2024-10-26")
Date(2024,10,26)

// FAILING, actual result is Date(2024,10,26)
>> DateValue("2024-10-27")
Date(2024,10,27)
```

Other new tests were already passing, but are included to specify intended behavior in related scenarios.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
@@ -566,10 +566,6 @@ namespace Microsoft.PowerFx.Functions
                     result = _epoch.Add(result.TimeOfDay);
                 }
 
-                var tzi = runner.TimeZoneInfo;
-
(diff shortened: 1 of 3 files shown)


rsullivan00:
Closing in favor of https://github.com/microsoft/Power-Fx/pull/2175, thanks @CarlosFigueiraMSFT !

PR #2175, opened 2024-01-30
Fix handling of GMT time zone

CarlosFigueiraMSFT:
There were some parts of the code that were mistakenly treating the GMT (i.e., London) time zone as UTC. Those are the same for part of the year, but not for all of it, which was causing some issues with parsing and adding date values. This change addresses it.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/DateTimeValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/DateTimeValue.cs
@@ -60,11 +60,11 @@ namespace Microsoft.PowerFx.Types
                 // https://github.com/microsoft/Power-Fx/issues/1931
                 return DateTime.SpecifyKind(value, DateTimeKind.Unspecified);
             }
-            else if (value.Kind == DateTimeKind.Unspecified && timeZoneInfo.BaseUtcOffset == TimeSpan.Zero)
+            else if (value.Kind == DateTimeKind.Unspecified && timeZoneInfo.Equals(TimeZoneInfo.Utc))
             {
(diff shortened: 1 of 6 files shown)


rsullivan00 (approved):

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateTimeValue.cs
```diff
@@ -50,21 +50,21 @@ internal static DateTime GetConvertedDateTimeValue(DateTime value, TimeZoneInfo
             }
             
             // Since we can't convert LocalKind time to UTC, if the time was of kind local just change kind.
-            if (value.Kind == DateTimeKind.Local && timeZoneInfo.BaseUtcOffset == TimeSpan.Zero)
```
MikeStall, line 53: >BaseUtcOffset == TimeSpan.Zero) [](http://example.com/codeflow?start=65&length=31)

Just curious - 
 Why was original code wrong?

MikeStall: nit:

Maybe a IsUtc() extension method? 
(that also gives a place to comment why not to use != TimeSpan.Zero

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Values/DateTimeValue.cs
CarlosFigueiraMSFT: Today, in London, BaseUtcOffset is zero. But not throughout the entire year.

Issue #1819, opened 2023-09-08
"Union All" tables 


babrekel:
In my scenario I have multiple tables that I would like to concatenate in a purely functional manner, without use of `Collect`. While trying, I was expecting the `Concat` or `Concatenate` function to be able to do this. Note: I am biased in that I have a background in .NET development. 

Note that the semantics I'm looking for is "Union All" without deduplication. 

```
Test1 = [{a:5,b:6}]
Test2 = [{a:5,b:6}]

// expected
Concat(Test1, Test2) // Type error
Concatenate(Test1, Test2) // Type error

// workaround:
ForAll(Sequence(CountRows(Test1)+CountRows(Test2)), If(Value<=CountRows(Test1),Index (Test2,Value), Index(Test2, Value - CountRows(Test1)))) // Table({a:5,b:6},{a:5,b:6}) 
```

rnwood:
This is an obvious gap in available functions. The community are using Ungroup() as an ugly workaround.

PR #2179, opened 2024-01-31
Accept tables as input in Table function (Table Concatenation)

adithyaselv:
Gives users ability to merge multiple Tables/Records into a single table without using side effects. returns a new table containing the rows from input Tables/Records.  

Fixes : #1819 

Examples 

``` 

> Table([{a:0, b:false, c:"Hello"}],  
        [{a:1, b:true, c:"World"}]) 

Table({a:0,b:false,c:"Hello"},{a:1,b:true,c:"World"}) 

 

// Mutations  

// Source Table mutations will not affect the result table 

> Set(T1, [{a:1, b:"Hello"}, {a:2, b:"World"}]); 

> Set(T2, [{a:5, b:"Five"}]); 

> Set(T3, Table(T1, T2)) 

> ClearCollect(T1, {a:3, b:"Power Fx"}) 

> T3 

[{a:1, b:"Hello"}, {a:2, b:"World"}, {a:5, b:"Five"}] 

```


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4464,4 +4464,23 @@
       A partial operator is the 2nd part of a statement `[Partial Op]` and can be one of "And", "Or", "Table" or "Record".
       It's used to determine how to combine multiple expressions with the same name and operator.</comment>
   </data>
+  <data name="ErrNeedTable" xml:space="preserve">
+    <value>Argument needs to be a table.</value>
```
CarlosFigueiraMSFT, line 4468: >Argument needs to be a table. [](http://example.com/codeflow?start=11&length=29)

We already have an error that we can use for this scenario. If you run something like `CountRows({a:1})` you should get it.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4464,4 +4464,23 @@
...
+  <data name="AboutTableConcatenate" xml:space="preserve">
+    <value>Creates a new table by combining the all the rows from the specified input tables.</value>
+    <comment>Description of 'TableConcatenate' function.</comment>
+  </data>
+  <data name="TableConcatenateArg1" xml:space="preserve">
+    <value>table</value>
+    <comment>function_parameter - Argument of the TableConcatenate function - A table whose rows will be appended to the resulting table.</comment>
+  </data>
+  <data name="AboutTableConcatenate_table" xml:space="preserve">
+    <value>A table whose rows will be appended to the resulting table.</value>
```
CarlosFigueiraMSFT, line 4479: >value [](http://example.com/codeflow?start=5&length=5)

Every new resource must have a comment explaining what the term means and the context in which the string is shown to the user. That prevents translation issues that we had in the past.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/TableConcatenate.cs
```diff
@@ -0,0 +1,127 @@
...
+namespace Microsoft.PowerFx.Core.Texl.Builtins
+{
+    internal sealed class TableConcatenateFunction : BuiltinFunction
+    {
+        public override bool IsSelfContained => true;
+
+        public override bool SupportsParamCoercion => false;
+
+        public TableConcatenateFunction()
+            : base("TableConcatenate", TexlStrings.AboutTableConcatenate, FunctionCategories.Table, DType.EmptyTable, 0, 0, int.MaxValue)
```
CarlosFigueiraMSFT, line 28: >0 [](http://example.com/codeflow?start=121&length=1)

Does it make sense to call it without any arguments?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -676,6 +676,20 @@ public static async ValueTask<FormulaValue> SortTable(EvalVisitor runner, EvalVi
...
 
+        public static FormulaValue TableConcatenate(IRContext irContext, FormulaValue[] args)
+        {
+            var tables = Array.ConvertAll(
+                args,
+                arg => arg switch
+                {
+                    TableValue r => r.Rows,
+                    BlankValue b => new List<DValue<RecordValue>> { DValue<RecordValue>.Of(b) },
+                    _ => new List<DValue<RecordValue>> { DValue<RecordValue>.Of((ErrorValue)arg) },
```
CarlosFigueiraMSFT, line 687: >(ErrorValue) [](http://example.com/codeflow?start=80&length=12)

If you know this is going to be an ErrorValue, can you match it in the switch statement? And have the `_` "else" clause throw an exception (since it should never be hit)?

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/TableConcatenate.txt
```diff
@@ -0,0 +1,29 @@
...
+// Invalid input - record instead of table 
+>> TableConcatenate({c:"Hello", d: {x: "World"}}) 
+Errors: Error 17-45: Invalid argument type (Record). Expecting a Table value instead.|Error 0-46: The function 'TableConcatenate' has some invalid arguments.
+
+// Invalid input - trying to coerce guid and datetime 
+>> TableConcatenate([{a:Date(2024,1,1)}], [{a:GUID("some-guid-value-1234")}]) 
+Errors: Error 39-73: Incompatible type. The item you are trying to put into a table has a type that is not compatible with the table.|Error 0-74: The function 'TableConcatenate' has some invalid arguments.
+
+>> TableConcatenate(Blank())
+Table(Blank())
```
CarlosFigueiraMSFT, line 29: This doesn't look right. CountRows(blank table) returns 0; CountRows(TC(t1, t2, t3)) should be the same as CountRows(t1) + CountRows(t2) + CountRows(t3), and in this test it would return 1.

In the Collect function, if we get a blank argument we jut ignore it. I think we can do the same in TableConcatenate: any blank/null arguments would not be part of the result.

CarlosFigueiraMSFT, line 29: >Table [](http://example.com/codeflow?start=0&length=5)

Additional interesting tests:

- TC( [1, 2], If(1<0, Table({Value:{a:2}})) )
    - The second argument is a _typed blank table_, but the schema is incompatible. It should fail (at compile-time) even if the second argument is blank.
- TC( [1, 2], Blank(), [4, 5], Blank(), [7, 8] )
    - Multiple blank arguments
- TC( [1,2], If(1/0<2,[3,4]), [5,6] )
    - Tests with error arguments
    - I think we can return an error if any of the tables are erroneous. That is consistent with other functions that take multiple tables
- TC( [1, 2, 3/0, 4], [5, Sqrt(-1), 7, 8] )
    - Tests with errors inside property values
    - Those are valid records, it should work
    - Result will be Table({Value:1},{Value:2},{Value:Error({Kind:ErrorKind.Div0}),{Value:4},...}
- TC( Filter([2,1,0,-1,-2], 1/Value>0), Filter([-2,-1,0,1,2], Log(Value)>0) )
    - Tests with error records
    - Those are valid tables

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
```diff
@@ -4120,6 +4120,17 @@ public void TestBlobFunction(string expression, string expectedType)
             Assert.True(result.IsSuccess);
         }
 
+        [Fact]
+        public void TexlFunctionTypeSemanticTableConcat()
```
CarlosFigueiraMSFT, line 4124: >TexlFunctionTypeSemanticTableConcat [](http://example.com/codeflow?start=20&length=35)

This is a negative test - that's good. We should have positive tests as well (simple cases, type union, ...), that work as a spec for the behavior of this function

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -676,6 +676,21 @@ public static async ValueTask<FormulaValue> SortTable(EvalVisitor runner, EvalVi
...
 
+        public static FormulaValue TableConcatenate(IRContext irContext, FormulaValue[] args)
+        {
+            var tables = Array.ConvertAll(
+                args,
+                arg => arg switch
+                {
+                    TableValue r => r.Rows,
+                    BlankValue b => new List<DValue<RecordValue>> { DValue<RecordValue>.Of(b) },
+                    ErrorValue e => new List<DValue<RecordValue>> { DValue<RecordValue>.Of(e) },
```
CarlosFigueiraMSFT, line 687: This doesn't look right - see comments in the .txt test file

adithyaselv: Makes sense will fix.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -2883,15 +2887,15 @@
...
   </data>
   <data name="AboutTable" xml:space="preserve">
-    <value>Creates a table from the specified records, with as many columns as there are unique record fields. For example: Table({key1: val1, key2: val2, ...}, ...)</value>
+    <value>Creates a table from the specified records and tables, with as many columns as there are unique record fields. For example: Table({key1: val1, key2: val2, ...}, ...)</value>
     <comment>Description of 'Table' function.</comment>
   </data>
   <data name="TableArg1" xml:space="preserve">
-    <value>record</value>
-    <comment>function_parameter - Argument of the Table function - a record that will become a row in the resulting table.</comment>
+    <value>record/table</value>
```
CarlosFigueiraMSFT, line 2894: >record/table [](http://example.com/codeflow?start=11&length=12)

Suggestion: `record_or_table`. This is the pattern we use in other functions when the argument can be multiple things.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -2883,15 +2887,15 @@
...
+    <comment>function_parameter - Argument of the Table function - a record/table that will be part of the the resulting table.</comment>
```
CarlosFigueiraMSFT, line 2895: >. [](http://example.com/codeflow?start=126&length=1)

Can you add this extra sentence to the comment: `Translate this string. Maintain as a single word (do not add spaces).` We've had issues in the past and this helped prevent those.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/TableConcatenate.txt
```diff
@@ -0,0 +1,96 @@
+
+>> Table([{a:0, b:false, c:"Hello"}], [{a:1, b:true, c:"World"}])
```
CarlosFigueiraMSFT, line 2: >>> Table([{a:0, b:fals [](http://example.com/codeflow?start=0&length=22)

Suggestion: rename this file to Table.txt (there's no TableConcatenate function)

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
```diff
@@ -4135,6 +4130,84 @@ public void TestBlobFunction(string expression, string expectedType)
...
+        [InlineData("Table(T1, Table(T2, T3))", "*[a:n, b:n, c:n, d:n]")]
+        [InlineData("Table(Table(T1, T2), T3)", "*[a:n, b:n, c:n, d:n]")]
+        [InlineData("Table(T1, Table(T2, T4))", "*[a:n, b:n, c:n]")]
+        [InlineData("Table(Table(T1, T2), T4)", "*[a:n, b:n, c:n]")]
+        [InlineData("Table([1,2], If(1/0<2,[3,4]), [5,6])", "*[Value: n]")]
+        [InlineData("Table(Sequence(20000))", "*[Value: n]")]
+        [InlineData("Table(Filter(T1, b = 5))", "*[a:n, b:n, c:n]")]
+        [InlineData("Table({X:1}, [1, 2, 3])", "*[X:n, Value:n]")]
+        public void TexlFunctionTypeSemanticsTableConcatenate(string script, string expectedType)
+        {
```
CarlosFigueiraMSFT, line 4155: This can be merged with TexlFunctionTypeSemanticsTable

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -311,6 +313,22 @@ private static void MutationFunctionsTestSetup(RecalcEngine engine, bool numberI
 
             var nameTableType = TableType.Empty().Add("name", FormulaType.String);
             engine.UpdateVariable("t_name", FormulaValue.NewTable(nameTableType.ToRecord()));
+
+            var r3Table = TableType.Empty().Add("b", numberType);
+            var r3Type = RecordType.Empty()
+                .Add(new NamedFormulaType("a", r3Table, "DisplayNamea"));
```
CarlosFigueiraMSFT, line 319: >DisplayNamea [](http://example.com/codeflow?start=57&length=12)

[typo] DisplayName

CarlosFigueiraMSFT, line 319: >DisplayNamea [](http://example.com/codeflow?start=57&length=12)

Can you use the display name in the tests?

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -173,12 +173,14 @@ private static void MutationFunctionsTestSetup(RecalcEngine engine, bool numberI
...
              * Record rwr3 => {![Field1:n, Field2:![Field2_1:n, Field2_2:s, Field2_3:![Field2_3_1:n, Field2_3_2:s]], Field3:b]}
              * Record r_empty => {}
+             * Record r3 => ![a:*[b:n]]
              * Table t1(r1) => Type (Field1, Field2, Field3, Field4)
              * Table t2(rwr1, rwr2, rwr3)
              * Table t_empty: *[Value:n] = []
              * Table t_empty2: *[Value:n] = []
              * Table t_an_bs: *[a:n,b:s] = []
              * Table t_name: *[name:s] = []
+             * Table t_bs: *[b:s]
```
CarlosFigueiraMSFT, line 183: >* Table t_bs: *[b:s]  [](http://example.com/codeflow?start=13&length=21)

The comment says that we're adding `t_bs`, but the code adds `t_bs1` and `t_bs2`

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/TableConcatenate.txt
adithyaselv: saw TableCoercion.txt so thought I could leave it as is

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
adithyaselv: fixed

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
adithyaselv: added a test . is that enough ?

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Table.cs
```diff
@@ -96,13 +100,29 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
...
+
+        public override void CheckSemantics(TexlBinding binding, TexlNode[] args, DType[] argTypes, IErrorContainer errors)
+        {
+            base.CheckSemantics(binding, args, argTypes, errors);
+
+            for (var i = 0; i < argTypes.Length; i++)
+            {
+                var ads = argTypes[i].AssociatedDataSources?.FirstOrDefault();
+
+                if (ads is IExternalDataSource tDsInfo && tDsInfo is IExternalTabularDataSource)
```
CarlosFigueiraMSFT, line 119: >ads is IExternalDataSource tDsInfo && tDsInfo is IExternalTabularDataSource [](http://example.com/codeflow?start=20&length=75)

This can be simplified as `if (ads is IExternalTabularDataSource)` - IExternalTabularDataSource inherits from IExternalDataSource

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Table.txt
```diff
@@ -0,0 +1,112 @@
+
+>> Table([{a:0, b:false, c:"Hello"}], [{a:1, b:true, c:"World"}])
```
CarlosFigueiraMSFT, line 2: Do you need to specify `#SETUP: TableSyntaxDoesntWrapRecords` in this test, since you are relying on this behavior?

CarlosFigueiraMSFT (dismissed):
:shipit:

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Table.txt
adithyaselv: sounds good better to be explicit

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Table.txt
```diff
@@ -0,0 +1,112 @@
...
+// The second argument is a record, no ambiguity
+>> Table([{a:1}], If(1/0<2,{a:2}), {a:3}, [{a:4}])
+Table({a:1},Error({Kind:ErrorKind.Div0}),{a:3},{a:4})
+
+// The second argument is a table. With an error table is passed in, the result is an error
+>> Table([{a:1}], If(1/0<2,[{a:2}]), {a:3}, [{a:4}])
+Error({Kind:ErrorKind.Div0})
+
+>> Table([{a:1}], {a:Error({Kind:ErrorKind.Custom})}, {a:3}, [{a:4}])
+Table({a:1},{a:Error({Kind:ErrorKind.Custom})},{a:3},{a:4})
```
Shpakh-MSFT, line 97: -Nested record
-Nested Table inside a record

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Table.cs
Shpakh-MSFT: >ads is IExternalDataSource tDsInfo && tDsInfo is IExternalTabularDataSource) [](http://example.com/codeflow?start=20&length=76)

This could be non-delegable filter as well. Important thing to check for delegability of this node. If it's non-delegable then provide warning.

Shpakh-MSFT: Never mind. We already discussed that.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Table.txt
```diff
@@ -0,0 +1,112 @@
...
+
+>> Table([])
+Table()
+
+>> Table(Table(Table()))
+Table()
+
+// Single argument with table
+>> Table([{a:0, b:false, c:"Hello"}])
+Table({a:0,b:false,c:"Hello"})
```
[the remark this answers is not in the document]
adithyaselv: added below

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Table.txt
adithyaselv: added below

LucGenetier:
✅ No public API change.

Shpakh-MSFT (dismissed):
:shipit:

adithyaselv:
missing commits for some reason . trying reopen

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

PR #2180, opened 2024-01-31
Intellisense tokens made public

anderson-joyle:
Complement for https://github.com/microsoft/Power-Fx/issues/2161.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/TokenTextSpan.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/TokenTextSpan.cs
@@ -44,7 +44,6 @@ namespace Microsoft.PowerFx.Core.Texl.Intellisense
             EndIndex = endIndex;
             TokenType = type;
             CanBeHidden = canHide;
-            IsTextFirst = false;
         }
 
         public TokenTextSpan(string name, Token token, TokenType type, bool canHide = false)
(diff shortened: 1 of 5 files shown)


LucGenetier:
✅ No public API change.

MikeStall:
>     internal sealed class TokenTextSpan : ITokenTextSpan, ITextFirstFlag

Could we just make this class public?
(Public classes can have internal interfaces)



---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/TokenTextSpan.cs:26 in b6fddcc. [](commit_id = b6fddcc9f4e5524ce47c8937788e8769ecac2593, deletion_comment = False)

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/TokenTextSpan.cs
```diff
@@ -23,7 +23,7 @@ internal interface ITokenTextSpan
         bool CanBeHidden { get; }
     }
 
-    internal sealed class TokenTextSpan : ITokenTextSpan, ITextFirstFlag
+    public sealed class TokenTextSpan : ITokenTextSpan, ITextFirstFlag
     {
         public string TokenName { get; private set; }
```
MikeStall, line 28: >public string TokenName { get; private set; } [](http://example.com/codeflow?start=8&length=45)

what is this?

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #2181, opened 2024-01-31
Add [Partial] combination attribute support for Named Formulas

lesaltzm:
This is an early prototype of support for a [Partial] attribute that enables combining named formulas.
This is done via name-mangling in the parser, renaming each NF to foo_Guid_0, foo_guid_1, ... and then creating a new hidden nf
Foo = Op(foo_guid0, foo_guid1, ...). This allows us to update the formulas independently in authoring, and process them without real changes to NF handling on the PA side. 

Animation 1

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Lexer/Tokens/IdentToken.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Lexer/Tokens/IdentToken.cs
@@ -26,16 +26,22 @@ namespace Microsoft.PowerFx.Syntax
         /// </summary>
         public DName Name { get; }
 
-        internal const string StrInterpIdent = "Concatenate";
-
-        internal IdentToken(string val, Span span)
+        /// <summary>
+        /// Initializes a new instance of the <see cref="IdentToken"/> class.
+        /// </summary>
+        /// <param name="val">The identifier.</param>
+        /// <param name="span">Source location the identifier is associated with.</param>
+        /// <param name="isNonSourceIdentToken">The compiler sometimes generates identifier tokens that don't map
+        /// 1-1 with specific source locations. This is done for cases like string interpolation, where a fake
+        /// Concatenate call node is added, using a Concatenate ident token that doesn't appear in the actual source.</param>
+        internal IdentToken(string val, Span span, bool isNonSourceIdentToken = false)
             : this(val, span, false, false)
         {
             Contracts.AssertValue(val);
 
             // String interpolation sometimes creates tokens that do not exist in the source code
             // so we skip validating the span length for the Ident that the parser generates
-            Contracts.Assert(val.Length == span.Lim - span.Min || val == StrInterpIdent);
+            Contracts.Assert(val.Length == span.Lim - span.Min || isNonSourceIdentToken);
             _value = val;
             Name = DName.MakeValid(val, out IsModified);
         }
--- a/src/libraries/Microsoft.PowerFx.Core/Parser/PartialAttribute.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Parser/PartialAttribute.cs
@@ -1,0 +1,68 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using Microsoft.PowerFx.Syntax;
+
+namespace Microsoft.PowerFx.Core.Parser
+{
+    internal sealed class PartialAttribute
+    {
+        public enum AttributeOperationKind
+        {
+            Error,
+            PartialAnd,
+            PartialOr,
+            PartialTable,
+            PartialRecord
+        }
+
+        public readonly IdentToken AttributeName;
+
+        // This probably should be a TexlNode, but for now we're prototyping a simple case
+        // of just [Partial And]-type attributes, where they're pairs of name and operation.
+        public readonly Token AttributeOperationToken;
+
+        public readonly AttributeOperationKind AttributeOperation;
+
+        public PartialAttribute(IdentToken attributeName, Token attributeOperationToken)
+        {
+            AttributeName = attributeName;
+            AttributeOperationToken = attributeOperationToken;
+            AttributeOperation = ToKind(attributeOperationToken);
+        }
+
+        private AttributeOperationKind ToKind(Token attributeOperation)
+        {
+            if (attributeOperation is KeyToken keyTok)
+            {
+                if (keyTok.Kind == TokKind.KeyAnd)
+                {
+                    return AttributeOperationKind.PartialAnd;
+                }
+                else if (keyTok.Kind == TokKind.KeyOr)
+                {
+                    return AttributeOperationKind.PartialOr;
+                }
+            }
+            else if (attributeOperation is IdentToken identTok)
+            {
+                if (identTok.Name.Value == "Table")
+                {
+                    return AttributeOperationKind.PartialTable;
+                }
+                else if (identTok.Name.Value == "Record")
+                {
+                    return AttributeOperationKind.PartialRecord;
+                }
+            }
+
+            return AttributeOperationKind.Error;
+        }
+
+        public bool SameAttribute(PartialAttribute other)
+        {
+            return AttributeName.Name.Value == other.AttributeName.Name.Value &&
+                AttributeOperation == other.AttributeOperation;
+        }
+    }
+}
--- a/src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
@@ -39,6 +39,9 @@ namespace Microsoft.PowerFx.Core.Parser
 
             // Text first.  Implemented entirely in the Lexer.
             TextFirst = 1 << 5,
+
+            // Parse supports Attributes on Named Formulas/Udfs.
+            AllowAttributes = 1 << 6,
         }
 
         private bool _hasSemicolon = false;
@@ -88,7 +91,10 @@ namespace Microsoft.PowerFx.Core.Parser
         public static ParseUserDefinitionResult ParseUserDefinitionScript(string script, ParserOptions parserOptions)
         {
             Contracts.AssertValue(parserOptions);
-            var flags = (Flags.NamedFormulas | (parserOptions.NumberIsFloat ? Flags.NumberIsFloat : 0)) | (parserOptions.AllowParseAsTypeLiteral ? Flags.AllowTypeLiteral : 0);
+            var flags = Flags.NamedFormulas |
+                (parserOptions.NumberIsFloat ? Flags.NumberIsFloat : 0) |
+                (parserOptions.AllowParseAsTypeLiteral ? Flags.AllowTypeLiteral : 0) | 
+                (parserOptions.AllowAttributes ? Flags.AllowAttributes : 0);
             var formulaTokens = TokenizeScript(script, parserOptions.Culture, flags);
             var parser = new TexlParser(formulaTokens, flags);
 
@@ -208,6 +214,51 @@ namespace Microsoft.PowerFx.Core.Parser
             return new TypeLiteralNode(ref _idNext, parenOpen, expr, new SourceList(sourceList));
         }
 
+        private PartialAttribute MaybeParseAttribute()
+        {            
+            if (_curs.TidCur != TokKind.BracketOpen)
+            {
+                return null;
+            }
+
+            _curs.TokMove();
+            ParseTrivia();
+
+            var attributeName = TokEat(TokKind.Ident);
+            if (attributeName == null)
+            {
+                CreateError(_curs.TokCur, TexlStrings.ErrNamedFormula_MissingValue);
+                _curs.TokMove();
+                return null;
+            }
+
+            ParseTrivia();
+
+            // For this prototype, the attribute op can be an ident, And or Or. 
+            // Definitely not the long term impl here, but works for now.
+            Token attributeOp;
+            if (_curs.TokCur.Kind == TokKind.Ident ||
+                _curs.TokCur.Kind == TokKind.KeyAnd ||
+                _curs.TokCur.Kind == TokKind.KeyOr)
+            {
+                attributeOp = _curs.TokCur;
+                _curs.TokMove();
+            }
+            else
+            {
+                // Use TokEat here to post an error that we expected an ident.
+                TokEat(TokKind.Ident, addError: true);
+                _curs.TokMove();
+                return null;
+            }
+
+            ParseTrivia();
+            TokEat(TokKind.BracketClose);
+            ParseTrivia();
+
+            return new PartialAttribute(attributeName.As<IdentToken>(), attributeOp);
+        }
+
         private ParseUserDefinitionResult ParseUDFsAndNamedFormulas(string script, ParserOptions parserOptions)
         {
             var udfs = new List<UDF>();
@@ -218,6 +269,12 @@ namespace Microsoft.PowerFx.Core.Parser
 
             while (_curs.TokCur.Kind != TokKind.Eof)
             {
+                PartialAttribute attribute = null;
+                if (_flagsMode.Peek().HasFlag(Flags.AllowAttributes))
+                {
+                    attribute = MaybeParseAttribute();
+                }
+
                 var thisIdentifier = TokEat(TokKind.Ident);
                 if (thisIdentifier == null)
                 {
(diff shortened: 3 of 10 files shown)


src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
```diff
@@ -155,5 +166,121 @@ private bool CheckReturnType(IdentToken returnType, List<TexlError> errors)
...
+            return new ParseUserDefinitionResult(newFormulas, parsed.UDFs, parsed.DefinedTypes, errors, parsed.Comments);
+        }
+
+        private Formula GetPartialCombinedFormula(string name, PartialAttribute.AttributeOperationKind operationKind, IList<NamedFormula> formulas)
+        {
+            return operationKind switch
+            {
+                PartialAttribute.AttributeOperationKind.PartialAnd => GeneratePartialFunction("And", name, formulas),
+                PartialAttribute.AttributeOperationKind.PartialOr => GeneratePartialFunction("Or", name, formulas),
+                PartialAttribute.AttributeOperationKind.PartialTable => GeneratePartialFunction("TableConcatenate", name, formulas),
```
lesaltzm, line 253: Impls for table/record coming in a future PR

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Lexer/Tokens/IdentToken.cs
```diff
@@ -26,16 +26,14 @@ public class IdentToken : Token
         /// </summary>
         public DName Name { get; }
 
-        internal const string StrInterpIdent = "Concatenate";
-
-        internal IdentToken(string val, Span span)
+        internal IdentToken(string val, Span span, bool isNonSourceIdentToken = false)
```
CarlosFigueiraMSFT, line 29: >isNonSourceIdentToken [](http://example.com/codeflow?start=56&length=21)

Can you add a comment explaining what 'isNonSourceIdentToken' is? I could only find out by looking for its callers.

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -95,6 +98,23 @@ public static ParseUserDefinitionResult ParseUserDefinitionScript(string script,
...
+        /// Parses a script with user definitions. Same as <see cref="ParseUserDefinitionScript(string, ParserOptions)"/>.
+        /// but internal for prototyping of attribute support. 
+        /// </summary>
+        /// <param name="script">Script to be parsed.</param>
+        /// <param name="parserOptions">Options for parsing an expression.</param>
+        /// <returns><see cref="ParseUserDefinitionResult"/>.</returns>
+        internal static ParseUserDefinitionResult ParseUserDefinitionScriptWithAttributes(string script, ParserOptions parserOptions)
+        {
+            Contracts.AssertValue(parserOptions);
+            var flags = (Flags.NamedFormulas | Flags.AllowAttributes | (parserOptions.NumberIsFloat ? Flags.NumberIsFloat : 0)) | (parserOptions.AllowParseAsTypeLiteral ? Flags.AllowTypeLiteral : 0);
```
CarlosFigueiraMSFT, line 111: >(Flags.NamedFormulas | Flags.AllowAttributes | (parserOptions.NumberIsFloat ? Flags.NumberIsFloat : 0)) | (parserOptions.AllowParseAsTypeLiteral ? Flags.AllowTypeLiteral : 0); [](http://example.com/codeflow?start=24&length=175)

This is almost the same logic as line 94; can it be refactored into a common place?

src/libraries/Microsoft.PowerFx.Core/Parser/PartialAttribute.cs
```diff
@@ -0,0 +1,68 @@
...
+                else if (identTok.Name.Value == "Record")
+                {
+                    return AttributeOperationKind.PartialRecord;
+                }
+            }
+
+            return AttributeOperationKind.Error;
+        }
+
+        public bool SameAttribute(PartialAttribute other)
```
CarlosFigueiraMSFT, line 62: >SameAttribute [](http://example.com/codeflow?start=20&length=13)

Why not override Equals?

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
```diff
@@ -155,5 +166,121 @@ private bool CheckReturnType(IdentToken returnType, List<TexlError> errors)
...
+                PartialAttribute.AttributeOperationKind.PartialRecord => GeneratePartialFunction("MergeRecords", name, formulas),
+                _ => throw new InvalidOperationException("Unknown partial op while generating merged NF")
+            };
+        }
+
+        private Formula GeneratePartialFunction(string functionName, string name, IList<NamedFormula> formulas)
+        {
+            // We're going to construct these texlnodes by hand so the spans match up with real code locations
+            var script = $"{functionName}({string.Join(", ", Enumerable.Range(0, formulas.Count).Select(i => name + _renamedFormulaGuid + i))})";
```
CarlosFigueiraMSFT, line 262: >,  [](http://example.com/codeflow?start=56&length=2)

Does it work with non-en-us locales, such as fr-FR, where the list separator is `;`?

src/tests/Microsoft.PowerFx.Core.Tests/AttributeParserTests.cs
```diff
@@ -0,0 +1,165 @@
...
+
+            Assert.Equal("Foo1", formulas[0].Ident.Name.Value);
+            Assert.Equal("SomeName1", formulas[0].Attribute.AttributeName.Name.Value);
+            Assert.Equal("Ident1", formulas[0].Attribute.AttributeOperationToken.As<IdentToken>().Name.Value);
+
+            Assert.Equal("Foo2", formulas[1].Ident.Name.Value);
+            Assert.Equal("SomeName2", formulas[1].Attribute.AttributeName.Name.Value);
+            Assert.Equal("Ident2", formulas[1].Attribute.AttributeOperationToken.As<IdentToken>().Name.Value);
+        }
+    }
```
CarlosFigueiraMSFT, line 164: >    [](http://example.com/codeflow?start=0&length=3)

It would be good to have a "full" test, where we actually call the combined NF

src/libraries/Microsoft.PowerFx.Core/Parser/PartialAttribute.cs
```diff
@@ -0,0 +1,68 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using Microsoft.PowerFx.Syntax;
+
+namespace Microsoft.PowerFx.Core.Parser
+{
+    internal class PartialAttribute
```
Shpakh-MSFT, line 8: >class [](http://example.com/codeflow?start=13&length=5)

sealed?

src/libraries/Microsoft.PowerFx.Core/Parser/PartialAttribute.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        public readonly IdentToken AttributeName;
+
+        // This probably should be a TexlNode, but for now we're prototyping a simple case
+        // of just [Partial And]-type attributes, where they're pairs of name and operation.
+        public readonly Token AttributeOperationToken;
+
+        public readonly AttributeOperationKind AttributeOperation;
+
+        public PartialAttribute(IdentToken attributeName, Token attributeOperation)
```
Shpakh-MSFT, line 27: >attributeOperation [](http://example.com/codeflow?start=64&length=18)

attributeOperationToken just to avoid confusion.

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
```diff
@@ -155,5 +166,121 @@ private bool CheckReturnType(IdentToken returnType, List<TexlError> errors)
...
+                    // Not all matched, don't use renamed formulas
+                    newFormulas.AddRange(nameGroup);
+                    continue;
+                }
+
+                newFormulas.AddRange(updatedGroupFormulas);
+                newFormulas.Add(
+                    new NamedFormula(
+                        new IdentToken(name, firstAttribute.AttributeName.Span, isNonSourceIdentToken: true),
+                        GetPartialCombinedFormula(name, firstAttribute.AttributeOperation, updatedGroupFormulas),
```
Shpakh-MSFT, line 239: >GetPartialCombinedFormula [](http://example.com/codeflow?start=24&length=25)

Does it work with the NF edit optimization that we have added?

lesaltzm: Mostly, yes, although optimization isn't the focus of this prototype.

src/libraries/Microsoft.PowerFx.Core/Parser/PartialAttribute.cs
lesaltzm: It's not equality. The tokens are different, and would have different spans. We're just ensuring the ids match

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
lesaltzm: ah, good point. This is just a script, and we're manually constructing the parsetree below, but it should still be the real locale-specific separator.

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -208,6 +228,51 @@ private TypeLiteralNode ParseTypeLiteral()
...
+            ParseTrivia();
+
+            // For this prototype, the attribute op can be an ident, And or Or. 
+            // Definitely not the long term impl here, but works for now.
+            Token attributeOp;
+            if (_curs.TokCur.Kind == TokKind.Ident ||
+                _curs.TokCur.Kind == TokKind.KeyAnd ||
+                _curs.TokCur.Kind == TokKind.KeyOr)
+            {
+                attributeOp = _curs.TokCur;
```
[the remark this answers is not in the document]
lesaltzm: We do later. That isn't a parse error as long as it's a valid identifier, it's handled in ProcessUserDefinitions.

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
```diff
@@ -155,5 +166,121 @@ private bool CheckReturnType(IdentToken returnType, List<TexlError> errors)
...
+                    id++;
+                    updatedGroupFormulas.Add(new NamedFormula(newName, formula.Formula, formula.StartingIndex, formula.Attribute));
+                }
+
+                if (updatedGroupFormulas.Count != nameGroup.Count())
+                {
+                    // Not all matched, don't use renamed formulas
+                    newFormulas.AddRange(nameGroup);
+                    continue;
+                }
```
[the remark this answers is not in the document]
lesaltzm: refactored to be clearer

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
Shpakh-MSFT: This is fine for prototype but when we productize it, would this design work well with optimization? As long as you have thought about it and there is a way, it's fine.

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
```diff
@@ -155,5 +161,124 @@ private bool CheckReturnType(IdentToken returnType, List<TexlError> errors)
...
+                    if (!firstAttribute.SameAttribute(formula.Attribute))
+                    {
+                        errors.Add(new TexlError(formula.Attribute.AttributeOperationToken, DocumentErrorSeverity.Severe, TexlStrings.ErrOperationDoesntMatch));
+                        continue;
+                    }
+
+                    var newName = new IdentToken(name + _renamedFormulaGuid + id, formula.Ident.Span, isNonSourceIdentToken: true);
+                    id++;
+                    updatedGroupFormulas.Add(new NamedFormula(newName, formula.Formula, formula.StartingIndex, formula.Attribute));
+                }
```
Shpakh-MSFT, line 214: You can skip this foreach if 
firstAttribute.AttributeOperation == PartialAttribute.AttributeOperationKind.Error

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
```diff
@@ -155,5 +161,124 @@ private bool CheckReturnType(IdentToken returnType, List<TexlError> errors)
...
+                var name = nameGroup.Key;
+                var firstAttribute = nameGroup.Select(nf => nf.Attribute).FirstOrDefault(att => att != null);
+
+                if (firstAttribute == null || nameGroup.Count() == 1)
+                {
+                    newFormulas.AddRange(nameGroup);
+                    continue;
+                }
+
+                var updatedGroupFormulas = new List<NamedFormula>();
```
Shpakh-MSFT, line 194: >var updatedGroupFormulas = new List<NamedFormula>(); [](http://example.com/codeflow?start=16&length=52)

You can initialize this outside foreach and just clear it here.

lesaltzm: Yes, but IMO this is clearer. That seems like an optimization that offers very little benefit, as this isn't a hot path.

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
lesaltzm: Yes, that was how it was set up previously, but as your earlier comment indicated, that was confusing. Also, doing it this way, we can correctly mark multiple errors, rather than just the first one.

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
Shpakh-MSFT: I don't see how it is clearer. It's extra allocation call per iteration which can be easily be avoided. But if you want to keep it this way then it's fine.

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
Shpakh-MSFT: Sure. Sorry I can't see last iteration here.

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -365,7 +422,7 @@ private ParseUserDefinitionResult ParseUDFsAndNamedFormulas(string script, Parse
                 {
                     // = or ( expected here
                     ErrorTid(_curs.TokCur, TokKind.Equ);
-                    break;
+                    MoveToNextUserDefinition();
```
Shpakh-MSFT, line 425: >MoveToNextUserDefinition [](http://example.com/codeflow?start=20&length=24)

Why are we changing this?

lesaltzm: Ah, this was something that was missed in that optimization pass we did. This fails the whole parse on the first missing start of a function definition, but we should still move to the next udf and try that.

Shpakh-MSFT (dismissed):
:shipit:

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4448,4 +4448,16 @@
...
+    <value>Unsupported attribute kind</value>
+    <comment>Error message when the user attempted to use the attribute syntax on a named formula definition, but we didn't recognize the attribute name they specified.</comment>
+  </data>
+  <data name="ErrOperationDoesntMatch" xml:space="preserve">
+    <value>All Partial operations on matching named formulas must be the same.</value>
+    <comment>{Locked=Partial} Error message when the user attempted to use a partial operator on multiple named formulas, but didn't specify the same operator on each one.</comment>
+  </data>
+  <data name="ErrUnknownPartialOp" xml:space="preserve">
+    <value>Unknown Partial operation.</value>
+    <comment>{Locked=Partial}. Error message when the user attempted to use a Partial operator that we don't recognize.</comment>
```
CarlosFigueiraMSFT, line 4461: >Error message when the user attempted to use a Partial operator that we don't recognize [](http://example.com/codeflow?start=31&length=87)

Please give more context about what this is - translators may not know what a "Partial" operat[ion/or] is and may translate it incorrectly.

CarlosFigueiraMSFT (approved):
Before this is made public, we really should have tests for NFs in this repository.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/AttributeParserTests.cs
lesaltzm: #2186

PR #2183, opened 2024-02-02
Value function support of OptionSetValue and deprecation of OptionSetInfo function

gregli-msft:


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

gregli-msft:
Split up in other PRs

PR #2184, opened 2024-02-02
Add BlobValue and BlobType

LucGenetier:


src/tests/Microsoft.PowerFx.Core.Tests/UserDefinedFunctionTests.cs
```diff
@@ -463,7 +463,7 @@ public void TestUserDefinedFunctionCloning()
         [InlineData("x = $\"{\"1$\"}.{\"}\";\r\nudf():Text = $\"{\"}\";\r\ny = 2;", 2, 1)]
         [InlineData("x = $\"{$\"{$\"{$\"{.12e4}\"}}\"}\";\r\nudf():Text = $\"{\"}\";\r\ny = 2;", 2, 1)]
         [InlineData("x = $\"{$\"{$\"{$\"{.12e4}\"}\"}\"}{$\"Another nested}\";\r\nudf():Text = $\"{\"}\";\r\ny = 2;", 2, 1)]
-        public void TestUserDefinitionStringInterpolationWithUnterminatedIsland(string formula, int nfCount, int udfCount)
+        public void TestUserDefStringInterpWithUnterminatedIsland(string formula, int nfCount, int udfCount)
```
LucGenetier, line 466: >TestUserDefStringInterpWithUnterminatedIsland [](http://example.com/codeflow?start=20&length=45)

For some reason XUnit doesn't like when the test name is too long

src/libraries/Microsoft.PowerFx.Core/Types/DTypeSpecParser.cs
```diff
@@ -14,11 +14,37 @@ internal static class DTypeSpecParser
         private const string _typeEncodings = "?ebnshdipmgo$cDT!*%lLNZPQqVOX-w";
         private static readonly DType[] _types = new DType[]
         {
```
LucGenetier, line 16: >{ [](http://example.com/codeflow?start=8&length=1)

No code change here, just easier to read

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValue.cs
```diff
@@ -106,7 +106,7 @@ public bool AsBoolean()
                 }
             }
 
-            throw new InvalidOperationException($"Can't coerce to double from {this.Type._type.GetKindString()})");
+            throw new InvalidOperationException($"Can't coerce to boolean from {this.Type._type.GetKindString()}");
```
LucGenetier, line 109: >boolean [](http://example.com/codeflow?start=66&length=7)

Not related to blobs but this message is wrong

src/libraries/Microsoft.PowerFx.Core/Public/Values/StringValue.cs
```diff
@@ -12,7 +12,18 @@ namespace Microsoft.PowerFx.Types
     public class StringValue : PrimitiveValue<string>
     {
         // List of types that allowed to convert to StringValue
-        internal static readonly IReadOnlyList<FormulaType> AllowedListConvertToString = new FormulaType[] { FormulaType.String, FormulaType.Number, FormulaType.Decimal, FormulaType.DateTime, FormulaType.Date, FormulaType.Time, FormulaType.Boolean, FormulaType.Guid };
+        internal static readonly IReadOnlyList<FormulaType> AllowedListConvertToString = new FormulaType[]
+        {
+            FormulaType.Blob,
```
LucGenetier, line 17: >FormulaType.Blob, [](http://example.com/codeflow?start=12&length=17)

Just added this one and reordered the list

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.BlobType)<br />❌CP0006 M:Microsoft.PowerFx.Types.IValueVisitor.Visit(Microsoft.PowerFx.Types.BlobValue)CP0002 M:Microsoft.PowerFx.Types.FormulaValueJSON.FromJson(System.String,Microsoft.PowerFx.Types.FormulaType,System.Boolean)<br />❌CP0002 M:Microsoft.PowerFx.Types.FormulaValueJSON.FromJson(System.String,Microsoft.PowerFx.Types.FormulaValueJsonSerializerSettings,Microsoft.PowerFx.Types.FormulaType)<br />❌CP0002 M:Microsoft.PowerFx.Types.FormulaValueJSON.FromJson(System.Text.Json.JsonElement,Microsoft.PowerFx.Types.FormulaType,System.Boolean)<br />❌CP0002 M:Microsoft.PowerFx.Types.FormulaValueJSON.FromJson(System.Text.Json.JsonElement,Microsoft.PowerFx.Types.FormulaValueJsonSerializerSettings,Microsoft.PowerFx.Types.FormulaType)

src/libraries/Microsoft.PowerFx.Core/Functions/IResourceManager.cs
```diff
@@ -0,0 +1,66 @@
...
+
+namespace Microsoft.PowerFx.Core.Functions
+{
+    /// <summary>
+    /// Interface for resource manager.
+    /// </summary>
+    public interface IResourceManager
+    {
+        /// <summary>
+        /// Create a new resource element for string.
```
CarlosFigueiraMSFT, line 15: >string [](http://example.com/codeflow?start=46&length=6)

What is this string? Can you elaborate on that? Is that the content, the URI, something else?

src/libraries/Microsoft.PowerFx.Core/Functions/IResourceManager.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// <summary>
+        /// Create a new resource element for a base64 encoded string.
+        /// </summary>
+        /// <param name="base64str">Base64 encoded string.</param>
+        /// <param name="fileType">Associated FileType.</param>
+        /// <returns>Index of the new element.</returns>
+        int NewElementFromBase64String(string base64str, FileType fileType = FileType.Any);
+
+        /// <summary>
+        /// Get resource element for index.
```
CarlosFigueiraMSFT, line 31: >// Get resource element for index. [](http://example.com/codeflow?start=9&length=34)

If we allow retrieval of an element by index, should we have an enumeration, or a Count property?

src/libraries/Microsoft.PowerFx.Core/Functions/IResourceManager.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// </summary>
+        /// <param name="index">Index.</param>
+        /// <returns>Uri.</returns>
+        Uri GetUri(int index);
+    }
+
+    /// <summary>
+    /// Interface for resource element.
+    /// </summary>
+    public interface IResourceElement
```
CarlosFigueiraMSFT, line 48: >IResourceElement [](http://example.com/codeflow?start=21&length=16)

[minor] Should it be in its own file?

src/libraries/Microsoft.PowerFx.Core/Public/Values/BlobValue.cs
```diff
@@ -0,0 +1,28 @@
...
+    public class BlobValue : FileValue
+    {
+        public BlobValue(IResourceManager resourceManager, int id)
+            : base(resourceManager, id)
+        {
+            Contract.Assert(IRContext.ResultType == FormulaType.Blob);
+        }
+
+        internal BlobValue(IRContext irContext)
+            : base(null, -2)
```
CarlosFigueiraMSFT, line 19: >-2 [](http://example.com/codeflow?start=25&length=2)

Curious: what is this value?

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNew.cs
```diff
@@ -136,6 +137,16 @@ public static ColorValue New(Color value)
...
+
+        internal static BlobValue NewBlob(IResourceManager resourceManager, int id)
+        {
+            return new BlobValue(resourceManager, id);
+        }
+
+        internal static BlobValue NewPDF(IResourceManager resourceManager, int id)
+        {
+            return new BlobValue(resourceManager, id);
         }
```
CarlosFigueiraMSFT, line 150: Do we need a separate function if it does the same as NewBlob?

src/libraries/Microsoft.PowerFx.Core/Functions/IResourceManager.cs
LucGenetier: Yes, the content. I'll update the comment

src/libraries/Microsoft.PowerFx.Core/Functions/IResourceManager.cs
LucGenetier: Do we need to have it in the interface?
Could be left to the implementer

src/libraries/Microsoft.PowerFx.Core/Functions/IResourceManager.cs
LucGenetier: Can do that ;)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -327,6 +328,62 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
...
+                {
+                    return new ErrorValue(irContext, new ExpressionError()
+                    {
+                        Message = $"Cannot convert {args[0].Type._type} to Blob, missing resource manager.",
+                        Span = irContext.SourceContext,
+                        Kind = ErrorKind.InvalidArgument
+                    });
+                }
+
+                return new BlobValue(resourceManager, resourceManager.NewElementFromString(sv.Value));
```
CarlosFigueiraMSFT, line 354: >return new BlobValue(resourceManager, resourceManager.NewElementFromString(sv.Value)); [](http://example.com/codeflow?start=16&length=86)

Can we avoid creating a new element in the resource manager?

In an expression such as `If(false, <blobexpr>, If(false, <stringexpr>, <blobvalue>))` the tree would be something like this:

    If(
        false,
        <blobexpr>
        StringToBlob(
            If(
                false,
                <stringexpr>,
                BlobToString(<blobvalue>)
            )
        )
    )

That means that we would be creating a new element there in this case, where it doesn't have to be

src/libraries/Microsoft.PowerFx.Core/Public/Values/BlobValue.cs
LucGenetier: Some invalid value
I could define is as a constant to be explicit on the intent

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -327,6 +328,62 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
...
         }
 
+        public static FormulaValue TextToBlob(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
+        {
+            runner.CancellationToken.ThrowIfCancellationRequested();
+
+            if (args[0].IsBlank())
+            {
+                return new BlobValue(irContext);
+            }
```
CarlosFigueiraMSFT, line 338: Since you use ReturnBehavior.ReturnBlankIfAnyArgIsBlank in the wrapper in LibraryUnary.cs, this should never be hit. And we should not return a BlobValue in this case - conversion functions propagate blank values (i.e., blank in, blank out)

src/libraries/Microsoft.PowerFx.Core/Functions/IResourceManager.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// <returns>Index of the new element.</returns>
+        int NewElementFromString(string str, FileType fileType = FileType.Any);
+
+        /// <summary>
+        /// Create a new resource element for a base64 encoded string.
+        /// </summary>
+        /// <param name="base64str">Base64 encoded string.</param>
+        /// <param name="fileType">Associated FileType.</param>
+        /// <returns>Index of the new element.</returns>
+        int NewElementFromBase64String(string base64str, FileType fileType = FileType.Any);
```
MikeStall, line 28: >NewElementFromBase64String [](http://example.com/codeflow?start=12&length=26)

Interfaces should be minimal. 
NewElementFromString and NewElementFromBase64String are redundant. 
Can we have 1 "new" method, and then the others are extension methods?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -327,6 +328,62 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
...
+            {
+                return FormulaValue.New(fv.ToString());
+            }
+
+            return new ErrorValue(irContext, new ExpressionError()
+            {
+                Message = $"Cannot convert {args[0].Type._type} to Text",
+                Span = irContext.SourceContext,
+                Kind = ErrorKind.InvalidArgument
+            });
```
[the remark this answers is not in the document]
LucGenetier: Ok

src/libraries/Microsoft.PowerFx.Core/Types/FileType.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+namespace Microsoft.PowerFx.Types
+{
+    public enum FileType
```
MikeStall, line 6: >FileType [](http://example.com/codeflow?start=16&length=8)

We need very strong invariants about what these mean. 

For example, what's the penalty if we just always use Unknown?

jas-valgotar:
We should consider adding coercion semantics tests in TypeCoercionTest.cs if we are introducing a new type.

src/libraries/Microsoft.PowerFx.Core/Functions/IResourceManager.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// <param name="i">Index.</param>
+        /// <returns>Resource element.</returns>
+        IResourceElement GetResource(int i);
+
+        /// <summary>
+        /// Get Uri for resource index.
+        /// </summary>
+        /// <param name="index">Index.</param>
+        /// <returns>Uri.</returns>
+        Uri GetUri(int index);
```
MikeStall, line 42: > Uri GetUri(int index); [](http://example.com/codeflow?start=7&length=23)

Can this be on IResourceElement? 
So that we only have 1 lookup

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs
```diff
@@ -157,6 +157,11 @@ private FunctionDescr Scan()
 
         private static FormulaType GetType(Type t)
         {
+            if (t == typeof(BlobValue))
+            {
+                return FormulaType.Blob;
+            }
+
```
jas-valgotar, line 164: suspisious, What if it's Task<BlobValue>? Ie async custom function.

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
```diff
@@ -48,7 +48,7 @@ public void Canvas_Float(ExpressionTestCase testCase)
                 ConsistentOneColumnTableResult = true
             };
 
-            RunExpressionTestCase(testCase, features, numberIsFloat: true);
+            RunExpressionTestCase(testCase, features, numberIsFloat: true, Summary);
```
MikeStall, line 51: >Summary [](http://example.com/codeflow?start=75&length=7)

what is this flag?

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -31,26 +30,35 @@ public class ExpressionEvaluationTests : PowerFxTest
...
+        // - setVariables: set variables in the symbol values
         // Each of these actions are executed in order (init, update, param, configure)
-        internal static Dictionary<string, (Func<PowerFxConfig, PowerFxConfig> initPfxConfig, Func<PowerFxConfig, object> updatePfxConfig, Func<object, RecordValue> parameters, Action<RecalcEngine, bool> configureEngine)> SetupHandlers = new ()
+        internal static Dictionary<string, 
+            (Func<PowerFxConfig, PowerFxConfig> initPfxConfig, 
+             Func<PowerFxConfig, SymbolTable, object> updatePfxConfig, 
+             Func<object, RecordValue> parameters, 
+             Action<RecalcEngine, bool> configureEngine,
+             Action<RuntimeConfig> runtimeConfig,
+             Action<ReadOnlySymbolValues, ISymbolSlot[]> setVariables)> SetupHandlers = new ()
```
MikeStall, line 42: do mutation tests, like collect.txt, handle this?

src/libraries/Microsoft.PowerFx.Core/Functions/IAsyncTexlFunction5.cs
```diff
@@ -0,0 +1,16 @@
...
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Core.Functions
+{    
+    // Texl function interface with IServiceProvider
+    internal interface IAsyncTexlFunction5
```
anderson-joyle, line 12: >IAsyncTexlFunction5 [](http://example.com/codeflow?start=23&length=19)

Is there a good way to squish these `IAsyncTexlFunction` into less numbers of interfaces?

MikeStall: We'd like to - but that's a separate refactoring

src/libraries/Microsoft.PowerFx.Core/Functions/DefaultResourceElement.cs
```diff
@@ -0,0 +1,113 @@
...
+
+using System;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.PowerFx.Core.Functions
+{
+    public abstract class BaseResourceElement
+    {
+        public abstract bool IsUri { get; }
```
MikeStall, line 12: > public abstract bool IsUri { get; } [](http://example.com/codeflow?start=7&length=36)

do we truly need this? 


Every BaseResourceElement is a uri

src/libraries/Microsoft.PowerFx.Core/Functions/DefaultResourceElement.cs
```diff
@@ -0,0 +1,113 @@
...
+            _base64Str = base64Str;
+            Handle = resourceManager.AddElement(this);
+        }
+
+        public override bool IsUri => false;
+
+        public override Task<string> GetAsBase64StringAsync()
+        {
+            return Task.FromResult(_base64Str);
+        }
```
MikeStall, line 71: +1 This is exactly the kind of optimization we talked about.

src/libraries/Microsoft.PowerFx.Core/Functions/DefaultResourceElement.cs
```diff
@@ -0,0 +1,113 @@
...
+            _uri = uri;
+            Handle = resourceManager.AddElement(this);
+        }
+
+        public override bool IsUri => false;
+
+        public override Task<string> GetAsBase64StringAsync()
+        {
+            return Task.FromResult(_uri.ToString());
+        }
```
MikeStall, line 106: ?

src/libraries/Microsoft.PowerFx.Core/Functions/DefaultResourceElement.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        public override Task<string> GetAsStringAsync()
+        {
+            throw new InvalidOperationException("Cannot get Base64 string from Uri");
```
MikeStall, line 110: >   throw new InvalidOperationException("Cannot get Base64 string from Uri"); [](http://example.com/codeflow?start=9&length=76)

Would this just be http download?

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs
LucGenetier: it won't work :)

src/tests/Microsoft.PowerFx.Interpreter.Tests/FileExpressionEvaluationTests.cs
LucGenetier: should probably be renamed
it's ITestOutputHelper == xUnit logger

src/libraries/Microsoft.PowerFx.Core/Functions/DefaultResourceElement.cs
LucGenetier: I think I can remove it

src/libraries/Microsoft.PowerFx.Core/Functions/DefaultResourceElement.cs
LucGenetier: It could be but I'd prefer throwing (with a correct message, not this one)  till we identify a scenario where we'd need to

src/libraries/Microsoft.PowerFx.Core/Functions/DefaultResourceElement.cs
LucGenetier: it's a bug :)

src/libraries/Microsoft.PowerFx.Core/Functions/DefaultResourceElement.cs
```diff
@@ -0,0 +1,113 @@
...
+{
+    public abstract class BaseResourceElement
+    {
+        public abstract bool IsUri { get; }
+
+        public virtual Uri Uri => new Uri($"appres://blobmanager/{Handle.Handle}");
+
+        public ResourceHandle Handle { get; init; }
+
+        public abstract Task<string> GetAsStringAsync();
```
CarlosFigueiraMSFT, line 18: >GetAsStringAsync [](http://example.com/codeflow?start=37&length=16)

A resource should be based on bytes, not strings.

src/libraries/Microsoft.PowerFx.Core/Functions/IAsyncTexlFunction5.cs
CarlosFigueiraMSFT: Reopening this comment.

I understand that this is a separate work item, but we can at least stop the bleeding ("when you find yourself in a hole, the first thing to do is to stop digging"). IAsyncTexlFunction4 takes the same arguments, except the first one is a TimeZoneInfo. Since it's an internal interface, we can change it to take an IServiceProvider instead, and retrieve the TimeZoneInfo from the provider (and update our implementations of that interface)

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -59,12 +67,35 @@ private static object RegExSetup(PowerFxConfig config)
...
+        }
+
+        private static object BlobSetup(PowerFxConfig config, SymbolTable symbolTable)
+        {
+            config.AddBlobTestFunctions();
+            config.EnableSetFunction();
+
+            return new List<ISymbolSlot>()
+            {
+                symbolTable.AddVariable("blob", FormulaType.Blob, true, "blob"),
```
CarlosFigueiraMSFT, line 82: >true [](http://example.com/codeflow?start=66&length=4)

Blobs shouldn't be mutable (from the point of view of Power Fx). Nor should strings (line below)

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -59,12 +67,35 @@ private static object RegExSetup(PowerFxConfig config)
...
+                symbolTable.AddVariable("str", FormulaType.String, true, "str")
+            };
+        }
+
+        private static void AddBlobVar(ReadOnlySymbolValues symbolValues, ISymbolSlot[] slots)
+        {
+            symbolValues.Set(slots[0], FormulaValue.NewBlank(FormulaType.Blob));
+            symbolValues.Set(slots[1], FormulaValue.NewBlank(FormulaType.String));
+        }
```
CarlosFigueiraMSFT, line 91: This is very loosely coupled with the function above - we need to know that there are 2 items in the slots... Can we pass the result of updatePfxConfig to this method, and it will cast to ISymbolSlot[] inside here?

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -385,7 +421,16 @@ protected override async Task<RunResult> RunAsyncInternal(string expr, string se
                                 config = k.initPfxConfig(config);
                             }
 
-                            object o = k.updatePfxConfig?.Invoke(config);
+                            object o = k.updatePfxConfig?.Invoke(config, symbolTable);
+
+                            if (o is ISymbolSlot slot)
+                            {
+                                slots.Add(slot);
+                            }
```
CarlosFigueiraMSFT, line 429: I don't see this codepath being hit. If it isn't, let's not add it. And if you pass the result of updatePfxConfig to setVariables, you won't need to define those variables here.

src/libraries/Microsoft.PowerFx.Core/Public/Values/BlobValue.cs
CarlosFigueiraMSFT: If this constructor will always fail, why do we need it? Shouldn't we always have a resource manager?

src/libraries/Microsoft.PowerFx.Core/Public/Values/FileValue.cs
```diff
@@ -0,0 +1,41 @@
...
+            {
+                throw new ArgumentException($"ResourceManager does not contain element with Id {handle.Handle}.", nameof(handle));
+            }
+        }
+
+        // if this blob is of Uri type, return the element string wher this Uri is stored, otherwise get it from resource manager
+        public override string ToString() => ResourceElement.Uri.ToString(); //.IsUri ? ResourceElement.String : _resourceManager.GetUri(Handle).ToString();
+
+        // not implemented as there is no official Blob() function yet
+        public override void ToExpression(StringBuilder sb, FormulaValueSerializerSettings settings) => throw new NotImplementedException();
```
CarlosFigueiraMSFT, line 35: Since this is an abstract class we don't need to have this here (move it to BlobValue)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -327,6 +328,32 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
...
+                });
+            }
+
+            return new BlobValue(resourceManager, new StringResourceElement(resourceManager, sv.Value).Handle);
+        }
+
+        public static FormulaValue BlobToText(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
+        {
+            runner.CancellationToken.ThrowIfCancellationRequested();
+            return FormulaValue.New((args[0] as FileValue).ToString());
```
CarlosFigueiraMSFT, line 354: >(args[0] as FileValue).ToString() [](http://example.com/codeflow?start=36&length=33)

Can we return an _identifier_ to the blob, so that if we need to convert it back to a blob, we can check if the resource manager has that value, and if so, return a reference to the same resource?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -327,6 +328,32 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
...
+            {
+                return new ErrorValue(irContext, new ExpressionError()
+                {
+                    Message = $"Cannot convert {args[0].Type._type} to Blob, missing resource manager.",
+                    Span = irContext.SourceContext,
+                    Kind = ErrorKind.InvalidArgument
+                });
+            }
+
+            return new BlobValue(resourceManager, new StringResourceElement(resourceManager, sv.Value).Handle);
```
CarlosFigueiraMSFT, line 348: >StringResourceElement [](http://example.com/codeflow?start=54&length=21)

We should have strings here

src/libraries/Microsoft.PowerFx.Core/Functions/DefaultResourceElement.cs
```diff
@@ -0,0 +1,113 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.PowerFx.Core.Functions
+{
+    public abstract class BaseResourceElement
+    {
+        public abstract bool IsUri { get; }
+
+        public virtual Uri Uri => new Uri($"appres://blobmanager/{Handle.Handle}");
```
CarlosFigueiraMSFT, line 14: >appres [](http://example.com/codeflow?start=44&length=6)

Minor, but we may want to use some other scheme here. "appres" refers to an "app resource", which makes sense in Power Apps (we have an "app", after all), but not necessarily in other places (e.g., CLI). Suggestion: blob:

LucGenetier:
The type isn't that new. 
Coercion rules are already existing and they are tested with the tests I'm adding

---
In reply to: [1924072346](https://github.com/microsoft/Power-Fx/pull/2184#issuecomment-1924072346) [](http://example.com/codeflow?ancestors=1924072346)

src/libraries/Microsoft.PowerFx.Core/Public/Values/BlobValue.cs
LucGenetier: Correct, no need :)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
LucGenetier: Is this only an optimization you're proposing? (and I don't know how to implement it)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
LucGenetier: Why?
A blob could have multiple forms: string, base64string, byte[], ...

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
LucGenetier: This is what we're doing here
FileValue.ToString() => ResourceElement.Uri.ToString()

src/libraries/Microsoft.PowerFx.Core/Functions/DefaultResourceElement.cs
LucGenetier: I'll make it configurable & fxres by default

src/libraries/Microsoft.PowerFx.Core/Functions/DefaultResourceElement.cs
LucGenetier: We can have multiple forms: byte[], string, base64string, Stream...

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
LucGenetier: AFAIK, yes

src/libraries/Microsoft.PowerFx.Core/Functions/IAsyncTexlFunction5.cs
LucGenetier: As part of the refactoring, we'll get rid of all other IAsyncTexlFunctionX

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
LucGenetier: They aren't mutable & can have multiple forms

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
LucGenetier: I fixed it

src/libraries/Microsoft.PowerFx.Core/Public/Values/FileValue.cs
LucGenetier: Yes but this avoids having to create it in BlobValue and later in MediaValue, ImageValue

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
CarlosFigueiraMSFT: A blob is semantically a list of bytes. If we pass a string, how do you get its equivalent byte[]? We can assume some encoding such as UTF-8, but that is not going guaranteed to be correct (some system may be expecting some other encoding, UTF-8 does not offer the best performance for many Eastern languages, ...). We could add an encoding argument, but that would make the interface more complicated. Ideally we should let the encoding of data to the resource manager itself.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
CarlosFigueiraMSFT: Not just an optimization. If you are creating a new resource, would it be "equal" to the previous one? The value `<blobvalue>` should be equal to

    If(
        false,
        <blobexpr>
        StringToBlob(
            If(
                false,
                <stringexpr>,
                BlobToString(<blobvalue>)
            )
        )
    )

Which we can accomplish if we use the URI as an "identity" for the blob value.

LucGenetier: This will not work as BlobToText will not work

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.BlobType)<br />❌CP0006 M:Microsoft.PowerFx.Types.IValueVisitor.Visit(Microsoft.PowerFx.Types.BlobValue)

src/libraries/Microsoft.PowerFx.Core/Functions/IAsyncTexlFunction5.cs
MikeStall: fortunately, these are internal and so we have an option to fix.

src/libraries/Microsoft.PowerFx.Core/Functions/Blob/UriBlob.cs
```diff
@@ -0,0 +1,36 @@
...
+            Uri = uri ?? throw new ArgumentNullException(nameof(uri));
+        }
+
+        public override async Task<byte[]> GetAsByteArrayAsync(CancellationToken token)
+        {
+            token.ThrowIfCancellationRequested();
+
+            if (Uri.Scheme == "https")
+            {
+                using HttpClient httpClient = new HttpClient();
```
MikeStall, line 26: >using HttpClient httpClient = new HttpClient(); [](http://example.com/codeflow?start=16&length=47)

I imagine at some future point we may need a service for this. 
Potentially to auth; but also for lifespan management of HttpClients.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -327,6 +329,29 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
...
+            return new BlobValue(new StringBlob(sv.Value));
+        }
+
+        public static FormulaValue BlobToText(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
+        {
+            if (args[0] is BlobValue bv && bv?.ResourceElement is UriBlob uri)
+            {
+                return FormulaValue.New(uri.Uri.ToString());
+            }
+
```
MikeStall, line 346: Can we just delete this and always fail. 

Just anxious about Uris in general , and a partial working thing may get people's hopes up.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -327,6 +329,29 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
             return Text(runner.GetFormattingInfo(), irContext, args, runner.CancellationToken);
         }
 
+        public static FormulaValue TextToBlob(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
```
MikeStall, line 332: >TextToBlob [](http://example.com/codeflow?start=35&length=10)

Can we rename to UriToBlob/BlobToUri?
to avoid confusion - folks may think "Text' means contents.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -327,6 +329,29 @@ public static FormulaValue Text(EvalVisitor runner, EvalVisitorContext context,
             return Text(runner.GetFormattingInfo(), irContext, args, runner.CancellationToken);
         }
 
+        public static FormulaValue TextToBlob(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
+        {
+            runner.CancellationToken.ThrowIfCancellationRequested();
+
+            StringValue sv = (StringValue)args[0];           
+            return new BlobValue(new StringBlob(sv.Value));
```
MikeStall, line 337: this is contents (not uri)?  Should this fail completely?

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -115,6 +108,11 @@ public static FormulaValue FromJson(JsonElement element, FormulaValueJsonSeriali
                         DateTime dt3 = element.GetDateTime();
                         return DateTimeValue.New(TimeZoneInfo.ConvertTimeToUtc(dt3));
                     }
+                    else if (formulaType is BlobType)
+                    {
+                        BlobElementBase resElem = new Base64Blob(element.GetString());
+                        return BlobValue.NewBlob(resElem);
+                    }
```
MikeStall, line 115: This is asserting that if JSON expects a blob, we assume the element is a base64 encoded string?

Can we be sure that's always true?

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
```diff
@@ -76,4 +76,9 @@
       <LastGenOutput>Resources.Designer.cs</LastGenOutput>
     </EmbeddedResource>
   </ItemGroup>
+  <ItemGroup>
+    <None Update="icon.png">
```
MikeStall, line 80: >  <None Update="icon.png"> [](http://example.com/codeflow?start=2&length=26)

?

src/libraries/Microsoft.PowerFx.Core/Functions/Blob/Base64Blob.cs
```diff
@@ -0,0 +1,34 @@
...
+                throw new ArgumentException("Invalid base64 string", nameof(base64Str));
+            }
+        }
+
+        public override Task<byte[]> GetAsByteArrayAsync(CancellationToken token)
+        {
+            token.ThrowIfCancellationRequested();
+            return Task.FromResult(FromBase64ToBytes(_base64Str));
+        }
+    }
```
CarlosFigueiraMSFT, line 33: >} [](http://example.com/codeflow?start=4&length=1)

This should override `GetAsBase64Async` since we already have that value, to prevent 2 conversions.

src/libraries/Microsoft.PowerFx.Core/Functions/Blob/StreamBlob.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.IO;
+
+namespace Microsoft.PowerFx.Core.Functions
+{
+    public class StreamBlob : ByteArrayBlob
```
CarlosFigueiraMSFT, line 8: >StreamBlob [](http://example.com/codeflow?start=17&length=10)

We should add to the comment that this will consume the entire stream (and not dispose it).

src/libraries/Microsoft.PowerFx.Core/Functions/Blob/StreamBlob.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        public StreamBlob(Stream stream)
+            : base(FromStreamToBytes(stream))
+        {
+        }
+
+        private static byte[] FromStreamToBytes(Stream stream)
+        {
+            using var ms = new MemoryStream();
+            stream.CopyTo(ms);
```
CarlosFigueiraMSFT, line 18: > stream.CopyTo(ms); [](http://example.com/codeflow?start=11&length=19)

We should probably have some limits here

src/libraries/Microsoft.PowerFx.Core/Functions/Blob/UriBlob.cs
```diff
@@ -0,0 +1,36 @@
...
+// Licensed under the MIT license.
+
+using System;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.PowerFx.Core.Functions
+{
+    public class UriBlob : BlobElementBase
```
CarlosFigueiraMSFT, line 11: >UriBlob [](http://example.com/codeflow?start=17&length=7)

Since our partners don't need a URI-based blobs for now, can we just not have that in this first iteration? And add later if needed?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
```diff
@@ -413,6 +413,28 @@ internal static partial class Library
...
+                UnaryOpKind.BlobToText,
+                StandardErrorHandling<BlobValue>(
+                    functionName: null, // internal function, no user-facing name
+                    expandArguments: NoArgExpansion,
+                    replaceBlankValues: DoNotReplaceBlank,
+                    checkRuntimeTypes: ExactValueTypeOrBlank<BlobValue>,
+                    checkRuntimeValues: DeferRuntimeValueChecking,
+                    returnBehavior: ReturnBehavior.ReturnBlankIfAnyArgIsBlank,
+                    targetFunction: BlobToText)
             },
```
CarlosFigueiraMSFT, line 438: If we don't need a URI-based blob, can we remove those conversions?

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -275,6 +275,10 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
             {                
                 result = await asyncFunc4.InvokeAsync(TimeZoneInfo, node.IRContext.ResultType, args, _cancellationToken).ConfigureAwait(false);
             }
+            else if (func is IAsyncTexlFunction5 asyncFunc5)
+            {                
+                result = await asyncFunc5.InvokeAsync(_services, node.IRContext.ResultType, args, _cancellationToken).ConfigureAwait(false);
+            }
             else if (func is IAsyncConnectorTexlFunction asyncConnectorTexlFunction)
```
CarlosFigueiraMSFT, line 282: This hurts my eyes :-)

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
CarlosFigueiraMSFT: I'm also not sure we can make this assumption.

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -31,40 +30,65 @@ public class ExpressionEvaluationTests : PowerFxTest
         // - parameters: to define the parameters for the test case
         //               receives an 'object' if defined in updatePfxConfig
         // - configureEngine: to configure the engine
+        // - runtimeConfigTypes: set of objects to add to RuntimeConfig
+        // - setVariables: set variables in the symbol values
```
CarlosFigueiraMSFT, line 34: Comment doesn't match implementation

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -31,40 +30,65 @@ public class ExpressionEvaluationTests : PowerFxTest
...
+        // - runtimeConfigTypes: set of objects to add to RuntimeConfig
+        // - setVariables: set variables in the symbol values
         // Each of these actions are executed in order (init, update, param, configure)
-        internal static Dictionary<string, (Func<PowerFxConfig, PowerFxConfig> initPfxConfig, Func<PowerFxConfig, object> updatePfxConfig, Func<object, RecordValue> parameters, Action<RecalcEngine, bool> configureEngine)> SetupHandlers = new ()
+        internal static Dictionary<string, 
+            (Func<PowerFxConfig, PowerFxConfig> initPfxConfig, 
+             Func<PowerFxConfig, SymbolTable, object> updatePfxConfig, 
+             Func<object, RecordValue> parameters, 
+             Action<RecalcEngine, bool> configureEngine,
+             Action<RuntimeConfig> runtimeConfig)> SetupHandlers = new ()
```
CarlosFigueiraMSFT, line 41: >runtimeConfig [](http://example.com/codeflow?start=35&length=13)

All of the handlers pass `null` to this element. Do we need it?

src/tests/Microsoft.PowerFx.Interpreter.Tests/BlobTests.cs
```diff
@@ -0,0 +1,59 @@
...
+            ISymbolSlot var1Slot = symbolTable.AddVariable("var1", FormulaType.Blob, false, "var1");            
+
+            BlobValue blob = new BlobValue(new StringBlob("abc"));
+            SymbolValues symbolValues = new SymbolValues(symbolTable);
+            symbolValues.Set(var1Slot, blob);            
+
+            RuntimeConfig runtimeConfig = new RuntimeConfig(symbolValues);
+            RecalcEngine engine = new RecalcEngine(config);
+
+            // IR: If:o(False:b, Lazy(ResolvedObject('var1:SymbolTable_5866477')), Lazy(TextToBlob:o(If:s(False:b, Lazy("a":s), Lazy(BlobToText:s(ResolvedObject('var1:SymbolTable_5866477')))))))
```
CarlosFigueiraMSFT, line 51: >BlobToText [](http://example.com/codeflow?start=133&length=10)

This test ensures that we cannot convert from blob to text; we should have another one that we cannot convert from text to blob.

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
```diff
@@ -4083,6 +4083,58 @@ public void TexlFunctionTypeSemanticColumn_Negative(string expression)
...
+        public void TestFilterWithNamingConflicts(string expression)
+        {
+            var tableType = TestUtils.DT("*[a:n,b:s,c:b]");
+            var symbolTable = new SymbolTable();
+            symbolTable.AddVariable("c", FormulaType.Boolean);
+            symbolTable.AddVariable("boolVar", FormulaType.Boolean);
+            symbolTable.AddVariable("T", new TableType(tableType));
+
+            TestSimpleBindingSuccess(expression, tableType, symbolTable);
+        }
```
CarlosFigueiraMSFT, line 4099: This doesn't seem related to blobs...

src/tests/Microsoft.PowerFx.Core.Tests/BlobTests.cs
```diff
@@ -0,0 +1,138 @@
...
+            {
+                Assert.NotNull(blob);
+                Assert.Equal($"Blob {i++}", blob.GetAsStringAsync(null, CancellationToken.None).Result);                
+            }          
+        }
+
+        [Fact]
+        public void BlobTest_LoadStream()
+        {
+            using Stream file = File.Open("icon.png", FileMode.Open, FileAccess.Read);
```
CarlosFigueiraMSFT, line 132: >"icon.png" [](http://example.com/codeflow?start=42&length=10)

Instead of having a binary file we can have something like `var myStream = new MemoryStream(new byte[] { 0, 1, 2, 3, 4, 5, 6, 7 });`, use that stream and validate that you can retrieve the data using one of the GetAsXXX functions

src/tests/Microsoft.PowerFx.Interpreter.Tests/BlobTests.cs
LucGenetier: I think we need to have TextToBlob working and this is tested elsewhere (AzureBlobStorage.CreateFile expression)

src/libraries/Microsoft.PowerFx.Core/Functions/Blob/UriBlob.cs
LucGenetier: I'll remove this class

src/libraries/Microsoft.PowerFx.Core/Functions/Blob/UriBlob.cs
LucGenetier: I'll remove this class

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
LucGenetier: See AzureBlobStorage.CreateFile example

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
LucGenetier: Would be a breaking change

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
LucGenetier: I'll remove UseBlob completely & BlobToText

src/tests/Microsoft.PowerFx.Core.Tests/Microsoft.PowerFx.Core.Tests.csproj
LucGenetier: not needed anymore as we remove UriBlob

src/libraries/Microsoft.PowerFx.Core/Functions/Blob/Base64Blob.cs
LucGenetier: Absolutely!

src/libraries/Microsoft.PowerFx.Core/Functions/Blob/StreamBlob.cs
LucGenetier: PVA seems to have a need up to 4TB
I recommended the use of URI where possible instead but we should probably allow it

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
LucGenetier: BlobToText yes
Not TextToBlob

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
LucGenetier: Sorry about that, we'll fix  later

src/tests/Microsoft.PowerFx.Core.Tests/TexlTests.cs
LucGenetier: You were the author of this one :)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
LucGenetier: Sorry I thought you wanted rename the UnaryOpKind

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.BlobType)<br />❌CP0006 M:Microsoft.PowerFx.Types.IValueVisitor.Visit(Microsoft.PowerFx.Types.BlobValue)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -217,6 +216,9 @@ public static bool TryFloat(FormattingInfo formatInfo, IRContext irContext, Form
                     }
 
                     break;
+
+                default:
+                    return false;
```
CarlosFigueiraMSFT, line 221: [minor] Not really needed - it would return false below as the result won't be set

src/tests/Microsoft.PowerFx.Interpreter.Tests/BlobTestFunctions.cs
```diff
@@ -0,0 +1,164 @@
...
+    internal class BlobFunctionImpl : BlobFunction, IAsyncTexlFunction5
+    {
+        public Task<FormulaValue> InvokeAsync(IServiceProvider runtimeServiceProvider, FormulaType irContext, FormulaValue[] args, CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();                       
+
+            if (args[0] is BlankValue)
+            {
+                return Task.FromResult<FormulaValue>(BlobValue.NewBlob(bytes: null));
+            }
```
CarlosFigueiraMSFT, line 53: I know this is a test function, but we normally propagate blank values in constructor functions, so this could return blank value as well.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -11,6 +11,7 @@
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.PowerFx.Core.Functions;
```
CarlosFigueiraMSFT, line 14: Not needed

CarlosFigueiraMSFT (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
LucGenetier: Well... next time I'll remove it

src/libraries/Microsoft.PowerFx.Core/Public/Values/BlobValue.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.PowerFx.Core.Functions;
+using Microsoft.PowerFx.Core.IR;
+
+namespace Microsoft.PowerFx.Types
+{
+    public class BlobValue : ValidFormulaValue
+    {
+        internal BlobContent ResourceElement { get; }
```
MikeStall, line 15: >ResourceElement [](http://example.com/codeflow?start=29&length=15)

nit; rename Content

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNew.cs
```diff
@@ -136,6 +139,36 @@ public static ColorValue New(Color value)
         public static VoidValue NewVoid()
         {
             return new VoidValue(IRContext.NotInSource(FormulaType.Void));
-        }
+        }
+
+        /// <summary>
+        /// Creates a new BlobValue from a string.
+        /// </summary>
+        /// <param name="str">String.</param>
```
MikeStall, line 147: >String [](http://example.com/codeflow?start=30&length=6)

better comment. Contents of the blob?
(not the uri)

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.BlobType)<br />❌CP0006 M:Microsoft.PowerFx.Types.IValueVisitor.Visit(Microsoft.PowerFx.Types.BlobValue)

src/libraries/Microsoft.PowerFx.Core/Public/Values/FormulaValueNew.cs
```diff
@@ -136,6 +139,36 @@ public static ColorValue New(Color value)
...
+        }
+
+        /// <summary>
+        /// Creates a new BlobValue from a stream.
+        /// All content of the stream will be copied to the Blob.
+        /// The stream will not be disposed.
+        /// </summary>
+        /// <param name="stream"></param>
+        /// <returns></returns>
+        public static BlobValue NewBlob(Stream stream)
```
MikeStall, line 169: >      public static BlobValue NewBlob(Stream stream) [](http://example.com/codeflow?start=2&length=52)

Can we just remove this and add later?
Else, there's nuance to sort out like:
- when do we read the stream.
- if we call GetBytes() twice - does that double read the stream? (which may not be possible) 
- do we need to optimize out copying to bytes?

LucGenetier: It's all explained in the comment
- all content is copied to the blob
- we only read the bytes once

MikeStall (approved):
:shipit:

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
Public API changes: <br />❌CP0006 M:Microsoft.PowerFx.Types.ITypeVisitor.Visit(Microsoft.PowerFx.Types.BlobType)<br />❌CP0006 M:Microsoft.PowerFx.Types.IValueVisitor.Visit(Microsoft.PowerFx.Types.BlobValue)

PR #2187, opened 2024-02-03
Make AsType strictly need Data Sources as it's second argument

jelopezf:
We initially supported relationships for resiliency to name collisions (we would grab the associated data source info off of them) but this just resulted in further issues for analysis/translation later on. This change makes this a strict requirement but also enhances the messaging around the error and provides fix and documentation in app checker.

image
image


LucGenetier:
✅ No public API change.

PR #2188, opened 2024-02-04
Don't create new DType instances for primitive FormulaType classes

CarlosFigueiraMSFT:
Small perf improvement: we don't need to create new DType instances for primitive types.


LucGenetier:
✅ No public API change.

adithyaselv (approved):
LGTM

PR #2190, opened 2024-02-06
Allow LazyRecord/LazyTable in JSON function

LucGenetier:


LucGenetier:
✅ No public API change.

MikeStall (approved):

src/tests/Microsoft.PowerFx.Json.Tests/JSONTests.cs
```diff
@@ -36,5 +37,52 @@ public void Json_IncludeBinaryData_NoSideEffects()
             Assert.False(result.IsSuccess);
             Assert.Equal("The JSON function cannot serialize binary data in non-behavioral expression.", result.Errors.First().Message);
         }
+
+        public class LazyRecordType : RecordType
```
lesaltzm, line 41: Please add a test where there's a recursive type & value, validate that it errors after a certain depth

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Config/Features.cs
```diff
@@ -119,6 +125,23 @@ internal static Features PowerFxV1AllowSetExpandedTypes
             CoalesceShortCircuit = true,
             PrimaryOutputPropertyCoercionDeprecated = true,
             AsTypeLegacyCheck = false,
+            JsonFunctionAcceptsLazyTypes = false
```
CarlosFigueiraMSFT, line 128: >JsonFunctionAcceptsLazyTypes [](http://example.com/codeflow?start=12&length=28)

We can set it to true in PowerFxV1 - this won't break any existing scenarios (will only make current scenarios that don't work today work, so not a breaking change).

src/tests/Microsoft.PowerFx.Json.Tests/JSONTests.cs
```diff
@@ -36,5 +37,122 @@ public void Json_IncludeBinaryData_NoSideEffects()
...
+            var record = RecordType.Empty();
+            record = record.Add("Property", new LazyRecordType());
+
+            var formulaParams = RecordType.Empty();
+            formulaParams = formulaParams.Add("Var", record);
+
+            var result = engine.Check("JSON(Var)", formulaParams);
+
+            Assert.True(result.IsSuccess);
+        }
```
CarlosFigueiraMSFT, line 78: Can you add a test where we actually check the result of the JSON function, to make sure that we not only compile, but we properly serialize lazy records? Also a test with lazy tables would be good.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Config/Features.cs
```diff
@@ -87,6 +88,11 @@ public sealed class Features
         /// </summary>
         internal bool PrimaryOutputPropertyCoercionDeprecated { get; set; }
 
+        /// <summary>
+        /// This is specific for PVA team and it is a temporary feature.
+        /// </summary>
+        internal bool JsonFunctionAcceptsLazyTypes { get; set; }
```
MikeStall, line 94: We should not have this flag.  #2197

Issue #2178, opened 2024-01-31
Handle Empty Power Fx Expressions as Empty Strings in Text-First Mode


gekarapa:
Currently, empty Power Fx expressions are not being treated as empty strings in text-first mode.

Run the following test, which is expected to succeed:
```
var engine = new RecalcEngine();
var parserOptions = new ParserOptions(System.Globalization.CultureInfo.InvariantCulture) { TextFirst = true };
var result = engine.Check("", parserOptions);

Assert.AreEqual(FormulaType.String, result.ReturnType);
Assert.IsTrue(result.IsSuccess);
```

PR #2192, opened 2024-02-07
Empty script when TextFirst is active

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2178.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
@@ -352,18 +352,25 @@ namespace Microsoft.PowerFx.Syntax
         {
             Contracts.AssertValue(text);
 
-            List<Token> tokens;
+            // In case of empty string script and TextFirst is active, returns a list of tokens representing an empty string.
+            if (text.Length == 0 && flags.HasFlag(Flags.TextFirst))
+            {
+                var span = new Span(0, 0);
+                return new List<Token>
+                {
+                    new StrInterpStartToken(span, true),
+                    new StrInterpEndToken(span, true),
+                    new EofToken(span)
+                };
+            }
+
+            var tokens = new List<Token>();
             StringBuilder sb = null;
 
             try
             {
                 // This StringBuilder is used by the Lexer as a temporary storage for tokenized characters.
                 sb = StringBuilderCache.Acquire(Math.Min(text.Length, DesiredStringBuilderSize));
-
-                if (TryParseTextFirstEmptyString(text, flags, out tokens))
-                {
-                    return tokens;
-                }
 
                 Token tok;
                 var impl = new LexerImpl(this, text, sb, flags);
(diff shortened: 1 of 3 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -394,6 +399,32 @@ public List<Token> GetTokens(string text, Flags flags = Flags.None)
...
+                {
+                    new StrInterpStartToken(span, true),
+                    new StrInterpEndToken(span, true),
+                    new EofToken(span)
+                };
+
+                return true;
+            }
+
+            tokens = new List<Token>();
```
MikeStall, line 426: >           tokens = new List<Token>(); [](http://example.com/codeflow?start=1&length=38)

nit - avoid this. 
We're relying on side-effects of a function/ 
If we rreturn false, then assume out params are all null

src/tests/Microsoft.PowerFx.Core.Tests/LexerTests.cs
```diff
@@ -640,5 +641,16 @@ public void ITextFirstTokenTest()
...
+        [Fact]
+        public void TextFirstEmptyExpressionTest()
+        {
+            var engine = new Engine();
+            var parserOptions = new ParserOptions(CultureInfo.InvariantCulture) { TextFirst = true };
+            var result = engine.Check(string.Empty, parserOptions);
+
+            Assert.Equal(FormulaType.String, result.ReturnType);
+            Assert.True(result.IsSuccess);
+        }
```
CarlosFigueiraMSFT, line 654: Can you also evaluate the expression? Would the result be blank, or an empty string? This test will be the _de facto_ specification for this scenario.

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -394,6 +399,32 @@ public List<Token> GetTokens(string text, Flags flags = Flags.None)
             }
 
             return tokens;
+        }
+
+        /// <summary>
+        /// In case of empty string scritp and TextFirst is active, returns a list of tokens representing an empty string.
```
CarlosFigueiraMSFT, line 405: [typo] "...empty string **scritp** and..." --> script

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -394,6 +399,32 @@ public List<Token> GetTokens(string text, Flags flags = Flags.None)
...
 
             return tokens;
+        }
+
+        /// <summary>
+        /// In case of empty string scritp and TextFirst is active, returns a list of tokens representing an empty string.
+        /// </summary>
+        /// <param name="text">Script.</param>
+        /// <param name="flags">Parser flags.</param>
+        /// <param name="tokens">Tokens object initialized by the caller.</param>
```
CarlosFigueiraMSFT, line 409: It is not initialized by the caller, it is initialized by this method

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -352,13 +352,18 @@ public IReadOnlyList<Token> LexSource(string text, Flags flags = Flags.None)
...
             try
             {
                 // This StringBuilder is used by the Lexer as a temporary storage for tokenized characters.
-                sb = StringBuilderCache.Acquire(Math.Min(text.Length, DesiredStringBuilderSize));
+                sb = StringBuilderCache.Acquire(Math.Min(text.Length, DesiredStringBuilderSize));
+
+                if (TryParseTextFirstEmptyString(text, flags, out tokens))
+                {
+                    return tokens;
+                }
```
CarlosFigueiraMSFT, line 366: This can be moved to before the creation (acquiring) of the StringBuilder - since it's not used.

We can also move it outside the `try` block, as this should not throw.

LucGenetier:
✅ No public API change.

MikeStall (approved):

CarlosFigueiraMSFT (approved):
🚢

LucGenetier:
✅ No public API change.

Issue #1578, opened 2023-06-13
PAD - Ability to exclude a list of functions


LucGenetier:
PAD team might want to exclude some default Fx function names which are not relevant for them
ETA Sept 23

gregli-msft:
What functions are they thinking of?  It's a better maker experience if core Power Fx is the same everywhere.

PR #2193, opened 2024-02-07
Exclude a list of functions

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/1578.
This was already available but now moving to `RecalcEngine` and no more need to extend `Engine` class.


MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #2199, opened 2024-02-08
Expose non-source flag on IdentToken to internals

lesaltzm:
For the Record support in PA Client for merge attributes, we're introducing a function that is only callable from generated expressions, not user code. This adds a flag to tell if an identifier is from source directly, or from something we generated.


src/tests/Microsoft.PowerFx.Core.Tests/ParseTests.cs
```diff
@@ -500,6 +500,7 @@ public void TexlCallHeadNodes()
                     Assert.NotNull(node.AsCall().Head);
                     Assert.True(node.AsCall().Head is Identifier);
                     Assert.True((node.AsCall().Head as Identifier).Namespace.IsRoot);
+                    Assert.False(node.AsCall().Head.Token.IsNonSourceIdentToken);
```
anderson-joyle, line 503: >False [](http://example.com/codeflow?start=27&length=5)

Any chance we add a test case where this is going to be true?

anderson-joyle (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2200, opened 2024-02-09
Named formulas new entry

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2186.


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -206,31 +206,13 @@ public void AddConstant(string name, FormulaValue data)
...
-                sb.AppendLine("Something went wrong when parsing user defined functions.");
-
-                foreach (var error in userDefinitionResult.Errors)
-                {
-                    error.FormatCore(sb);
-                }
-
-                throw new InvalidOperationException(sb.ToString());
-            }
+            (var userDefinedFunctions, _) = UserDefinitions.Process(script, parseCulture);
```
lesaltzm, line 209: I'm wondering if parsing (processing) the UDFs shouldn't be the responsibility of the Engine, rather than the symbol table. Logically, symbol table is more a place to store what symbols exist, so it probably should only come into play after the UserDefinition script is parsed/validated, I think

lesaltzm: Maybe instead of AddUserDefinition taking a script, it should take an instance of UDF? Maybe even one that's already bound?

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
```diff
@@ -87,6 +87,40 @@ private bool ProcessUserDefinitions(out UserDefinitionResult userDefinitionResul
...
         }
 
+        /// <summary>
+        /// Process user script and returns user defined functions and named formulas.
+        /// </summary>
+        /// <param name="script">User script containing UDFs and/or named formulas.</param>
+        /// <param name="parseCulture">CultureInfo to parse the script.</param>
+        /// <returns>Tuple.</returns>
+        /// <exception cref="InvalidOperationException">Throw if the user script contains errors.</exception>
+        public static (IEnumerable<UserDefinedFunction>, IEnumerable<NamedFormula>) Process(string script, CultureInfo parseCulture)
```
lesaltzm, line 97: Not really sure why we'd need this API wrapping ProcessUserDefinition? Is the error throwing better than returning a UserDefinitionResult and letting the caller handle that it might be erroneous?

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -376,10 +376,20 @@ public bool TryGetVariableType(string name, out FormulaType type)
...
             return false;
         }
-
+        
         public void UpdateSupportedFunctions(SymbolTable s)
         {
             SupportedFunctions = s;
         }
+        
+        public void AddNamedFormulas(string script, CultureInfo parseCulture = null)
```
lesaltzm, line 385: Consider instead having this be a general AddUserDefinitions, that adds both named formulas like this as well as any UDFs that might be in that script. At least as it stands, it's fine for them to be defined in the same "file"/string, so having separate APIs seems like a place we could trip up.

lesaltzm, line 385: IIRC, doesn't RecalcEngine have a culture/parser options set already? Or should this take a ParserOptions like Eval appears to.

src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationFunctionsTests.cs
```diff
@@ -380,5 +381,21 @@ public override bool TryShallowCopy(out FormulaValue copy)
                 return true;
             }
         }
+
+        [Theory]
+        [InlineData("x=1;y=2;z=x+y;", "Abs(-z)", 3)]
+        public void NameFormulasTest(string namedFormulas, string expression, decimal expected)
```
lesaltzm, line 387: should this be in MutationFunctionsTests? Doesn't appear to do mutation. A new file of tests for UserDefinitions might be good here

anderson-joyle: Temporary placement. It will be moved.

jorisdg: nit: missing the D in the method name "NamedFormulasTest"

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
anderson-joyle: `Engine.Check()` takes a `ParserOptions`. IMO its better to pass only the `CultureInfo` since side effects are not allowed in UDF/NF.

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
anderson-joyle: This is not only wrapping but also unifying a block of code that would be called from multiple places.
The `throw` is following the original UDF implementation. We can discuss and change it.

MikeStall:
>                 if (check.Parse.Root is BinaryOpNode bo && bo.Op == BinaryOp.Equal && bo.Left.Kind == NodeKind.FirstName)

Can we avoid this code and use the real named formula parser?


---
Refers to: src/libraries/Microsoft.PowerFx.Repl/Repl.cs:311 in 6ffdf50. [](commit_id = 6ffdf502357e1870c4d83954cda9674026190b4b, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Repl/Repl.cs
```diff
@@ -118,6 +118,7 @@ public PowerFxREPL()
             // Hook through the HelpProvider, don't override these Help functions
             this.MetaFunctions.AddFunction(new Help0Function(this));
             this.MetaFunctions.AddFunction(new Help1Function(this));
+            this.MetaFunctions.AddFunction(new SetFormulasFunction(this));
```
MikeStall, line 121: >SetFormulasFunction [](http://example.com/codeflow?start=47&length=19)

Can we avoid new function? See comment below

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
MikeStall, line 97: > (IEnumerable<UserDefinedFunction>, IEnumerable<NamedFormula>) [](http://example.com/codeflow?start=21&length=62)

userDefinitionResult?

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
anderson-joyle: Moved to the engine.

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
```diff
@@ -87,6 +87,40 @@ private bool ProcessUserDefinitions(out UserDefinitionResult userDefinitionResul
...
+        {
+            var options = new ParserOptions()
+            {
+                AllowsSideEffects = false,
+                Culture = parseCulture ?? CultureInfo.InvariantCulture
+            };
+
+            var sb = new StringBuilder();
+
+            UserDefinitions.ProcessUserDefinitions(script, options, out var userDefinitionResult);
```
CarlosFigueiraMSFT, line 107: >ProcessUserDefinitions [](http://example.com/codeflow?start=28&length=22)

`ProcessUserDefinitions` also takes a `Features` argument, it may be worth passing one as well.

src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationFunctionsTests.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
 using System.Data;
 using System.Diagnostics.Contracts;
+using System.Globalization;
```
CarlosFigueiraMSFT, line 8: Not needed?

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -376,10 +380,49 @@ public bool TryGetVariableType(string name, out FormulaType type)
...
         }
+
+        /// <summary>
+        /// Add a set of user-defined formulas and functions to the engine.
+        /// </summary>
+        /// <param name="script"></param>
+        /// <param name="parseCulture"></param>
+        /// <param name="symbolTable"></param>
+        /// <param name="onUpdate"></param>
+        public void AddUserDefined(string script, CultureInfo parseCulture = null, ReadOnlySymbolTable symbolTable = null, Action<string, FormulaValue> onUpdate = null)
```
adithyaselv, line 396: Will it be better to call this method `AddUserDefinitions` ?

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -376,10 +380,49 @@ public bool TryGetVariableType(string name, out FormulaType type)
...
         }
+
+        /// <summary>
+        /// Add a set of user-defined formulas and functions to the engine.
+        /// </summary>
+        /// <param name="script">Script containing user defined functions and/or named formulas.</param>
+        /// <param name="parseCulture">Locale to parse user defined script.</param>
+        /// <param name="symbolTable">Extra SymbolTable if needed.</param>
+        /// <param name="onUpdate">Function to be called when update is triggered.</param>
+        public void AddUserDefinitions(string script, CultureInfo parseCulture = null, ReadOnlySymbolTable symbolTable = null, Action<string, FormulaValue> onUpdate = null)
```
MikeStall, line 396: >AddUserDefinitions [](http://example.com/codeflow?start=20&length=18)

can we move to engine base class?
For parse-only scenarios. And so we can work with different backends.

MikeStall:
>             if (check.Parse.IsSuccess)

In the same way we can check if a parse expression succeeds, we should be able to check if parse declarations succeeds.

---
Refers to: src/libraries/Microsoft.PowerFx.Repl/Repl.cs:276 in 842e908. [](commit_id = 842e908a3ddc6ff6e6ba1d52fb52a864b44c0042, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -298,21 +298,14 @@ private ParseUserDefinitionResult ParseUDFsAndNamedFormulas(string script, Parse
                     _curs.TokMove();
                     ParseTrivia();
 
-                    if (_curs.TidCur == TokKind.Semicolon)
+                    if (_curs.TidCur == TokKind.Semicolon || _curs.TidCur == TokKind.Eof)
```
MikeStall, line 301: > || _curs.TidCur == TokKind.Eof [](http://example.com/codeflow?start=57&length=31)

This changes the public grammar for named formulas (no longer requiring semicolon at end)

consider separate PR? Ensure approvals.

src/libraries/Microsoft.PowerFx.Repl/Repl.cs
```diff
@@ -403,6 +370,23 @@ await this.Output.WriteLineAsync($"Error: Can't set '{name}' to a Void value.",
...
+            {
+                // Checking for named formulas and/or udfs.
+                try
+                {
+                    this.Engine.AddUserDefinitions(expression, parseCulture: this.ParserOptions.Culture, symbolTable: currentSymbolTable, onUpdate: OnFormulaUpdate);
+
+                    return new ReplResult();
+                }
+                catch (InvalidOperationException ex)
+                {
```
MikeStall, line 384: This isn't ideal - we should be able to reliably check for parse errors. See comment above.

anderson-joyle:
This PR created bigger discussions and will be split in smaller pieces.

PR #2198, opened 2024-02-08
Add Blob in connectors 

LucGenetier:

PR #2205, opened 2024-02-14
Fix Date management & remove uselsess exception handler

LucGenetier:


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -402,39 +402,20 @@ public async Task<FormulaValue> InvokeAsync(IConvertToUTC utcConverter, string c
...
-                        Kind = ErrorKind.Internal,
-                        Severity = ErrorSeverity.Critical,
-                        Message = $"In {nameof(HttpFunctionInvoker)}.{nameof(InvokeAsync)} request is null"
-                    });
-                }
+            using HttpRequestMessage request = await BuildRequest(args, utcConverter, cancellationToken).ConfigureAwait(false);
 
-                return await ExecuteHttpRequest(cacheScope, throwOnError, request, localInvoker, cancellationToken).ConfigureAwait(false);
-            }
-            catch (Exception ex)
```
MikeStall, line 422: >  catch (Exception ex) [](http://example.com/codeflow?start=10&length=22)

Can you elaborate on this? 
This will convert some non-exception paths (return new ErrorValue, which a host can ignore) to exception paths (which may crash host). Just confirming that this is safe?

LucGenetier: Yes. This exception handler was added in [https://github.com/microsoft/Power-Fx/pull/2198](https://github.com/microsoft/Power-Fx/pull/2198) and 1/ is not needed at all 2/ it's precisely what Roman is complaining about as throwOnError is false in his case

MikeStall: So does he want the unhandled exception?
Or does he want a returned ErrorValue object?

LucGenetier: He doesn't want the ErrorValue but an Exception

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #2209, opened 2024-02-15
Fix IDE rules to not complain about existing code.

MikeStall:
Opening in VS 17.9.0, something has changed about how VS reads IDE rules and it's now complinaing about 100s of things.


anderson-joyle (approved):
:shipit:

LucGenetier:
✅ No public API change.

Issue #2210, opened 2024-02-15
Anonymizing unknown functions


anderson-joyle:
If a user types an unknown function name (not part of BuiltinCore), it should be sanitized when anonymizing the expression.
Example:
`UserNewFunction(1)` should generate something similar to `#$customfunction$#(#$decimal$#)`.

PR #2211, opened 2024-02-15
Anonymizing unknown functions

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2210.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
@@ -96,7 +96,7 @@ namespace Microsoft.PowerFx.Syntax
             foreach (var udf in uDFs)
             {
                 var udfName = udf.Ident.Name;
-                if (_restrictedUDFNames.Contains(udfName) || texlFunctionSet.AnyWithName(udfName) || BuiltinFunctionsCore.IsKnownPublicFunction(udfName))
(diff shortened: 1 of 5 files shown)


src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
```diff
@@ -98,7 +96,7 @@ private IEnumerable<UserDefinedFunction> CreateUserDefinedFunctions(IEnumerable<
             foreach (var udf in uDFs)
             {
                 var udfName = udf.Ident.Name;
-                if (_restrictedUDFNames.Contains(udfName) || texlFunctionSet.AnyWithName(udfName) || BuiltinFunctionsCore._library.AnyWithName(udfName) || BuiltinFunctionsCore.OtherKnownFunctions.Contains(udfName))
+                if (_restrictedUDFNames.Contains(udfName) || texlFunctionSet.AnyWithName(udfName) || BuiltinFunctionsCore.IsKnownPublicFunction(udfName))
```
anderson-joyle, line 99: >BuiltinFunctionsCore.IsKnownPublicFunction(udfName) [](http://example.com/codeflow?start=101&length=51)

@VamsiModem Now this is also checking `_featureGateFunctions`. Do you see a potential problem?

LucGenetier:
✅ No public API change.

MikeStall:
>                 }

Actually - this original logic seems  wrong.
It means adding any new builtin function is a breaking change! We should followup separately.




---
Refers to: src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs:103 in aa1dd48. [](commit_id = aa1dd4848e9454aafbe2fa99c8481edc687011f8, deletion_comment = False)

MikeStall (dismissed):
:shipit:

MikeStall (approved):
:shipit:

PR #2212, opened 2024-02-16
Seal system records and tables, ensuring we can add fields in the future.

gregli-msft:
Records and tables can now be sealed, ensuring that we won't break formulas if we decide to add something to them at a later date.  

This means that the User record, IfError's FirstError/AllErrors record/table, and select other "system" records and tables that we select in the future:
- will be blocked from being used in `AddColumns` and `RenameColumns`.  As this may result in a field being introduced by the maker that would collide with something we may want to add later.
- can't participate in a type union with other records.  `If( true, {a:1}, User )` will result in a Void type.

`ShowColumns` is the easy workaround that removes the block, as it has the maker specifically select the columns that they want to use and newly introduced columns will not impact it.  Note that `DropColumns` doesn't have the same impact as newly added columns would remain.

This change has no impact on Canvas apps, as it requires V1 compat to be enabled.  While Canvas doesn't have the User record, it does have the User function, for which we should mark as sealed once we enable the V1 flag there.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Config/UserInfo.cs
```diff
@@ -123,6 +123,8 @@ internal static RecordType GetUserTypeWorker(params string[] fields)
                 }
             }
 
+            type._type.IsSealed = true;
```
MikeStall, line 126: This is key change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationScripts/User_SealedRecord.txt
```diff
@@ -0,0 +1,110 @@
...
+
+>> AddColumns( User, Foo, 1 )
+Errors: Error 12-16: Can't add columns to a system record or table of system records. Use ShowColumns to extract the specific columns you need first.|Error 0-26: The function 'AddColumns' has some invalid arguments.
+
+>> RenameColumns( User, FullName, Name )
+Errors: Error 15-19: Can't rename columns of a system record or table of system records. Use ShowColumns to extract the specific columns you need first.|Error 0-37: The function 'RenameColumns' has some invalid arguments.
+
+// DropColumns and ShowColumns work as normal, as they are removing without adding
+
+>> DropColumns( User, FullName )
```
CarlosFigueiraMSFT, line 21: If the input of DropColumns is locked, we should also lock the result, so that. This way one cannot add a new column to a locked type.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1351,6 +1362,11 @@ public DType Add(DName name, DType type)
             Contracts.Assert(name.IsValid);
             type.AssertValid();
 
+            if (IsSealed)
+            {
+                throw new InvalidOperationException();
```
CarlosFigueiraMSFT, line 1367: If we throw an exception, can this break something? Would returning an error type be safer?

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -2709,6 +2725,12 @@ public static DType Union(ref bool fError, DType leftType, DType rightType, bool
             leftType.AssertValid();
             rightType.AssertValid();
 
+            if ((leftType.IsSealed || rightType.IsSealed) && leftType != rightType)
```
CarlosFigueiraMSFT, line 2728: If we are union "to left type", would it be ok if the right type is sealed but the left type is not?

CarlosFigueiraMSFT, line 2728: If leftType and rightType are not the same type, but are identical, should this work? For example, `[DropColumns(User, FullName), DropColumns(User, FullName)]`. We should have a test to validate that.

src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationScripts/User_SealedRecord.txt
CarlosFigueiraMSFT: Nevermind, saw that this is already done.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationScripts/User_SealedRecord.txt
```diff
@@ -0,0 +1,110 @@
...
+
+>> User
+{DataverseUserId:GUID("aa1d4f65-044f-4928-a95f-30d4c8ebf118"),Email:"susan@contoso.com",FullName:"Susan Burk",TeamsMemberId:"29:1DUjC5z4ttsBQa0fX2O7B0IDu30R"}
+
+>> [User]
+Table({DataverseUserId:GUID("aa1d4f65-044f-4928-a95f-30d4c8ebf118"),Email:"susan@contoso.com",FullName:"Susan Burk",TeamsMemberId:"29:1DUjC5z4ttsBQa0fX2O7B0IDu30R"})
+
+// AddColumns and RenameColumns are blocked
+
+>> AddColumns( User, Foo, 1 )
```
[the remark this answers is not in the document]
gregli-msft: Done.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
gregli-msft: Removed.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
gregli-msft: I added tests for this, and tried various combinations.  On the surface, it seems like leftType != rightType shouldn't be enough, but somehow it is working.  Do we cache the DTypes and de-dup them?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
gregli-msft: Oh, it looks like == has been overloaded for DType and parts.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
gregli-msft: No, I think we want to limit it both ways.  If we allow right to merge into left, and then we later add something to right, we might conflict with something that somebody adds later to left.  For example, `If( true, { a: 1 }, User )` is disallowed (results in a Void).  We add "b" to User as a number, and the maker's formula changes to `If( true, { a: 1, b: "asdf" }, User )`, then we have a problem.  I'd prefer that the maker extract the specific fields they want from User via ShowColumns which avoids the problem.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -5726,6 +5726,7 @@ public override void PostVisit(TableNode node)
                         ? DType.CreateTable(exprType.GetNames(DPath.Root))
                         : DType.CreateTable(new TypedName(exprType, TableValue.ValueDName)))
                     : DType.EmptyTable;
+                tableType.IsSealed = exprType.IsSealed;
```
MikeStall, line 5729: > tableType.IsSealed = exprType.IsSealed; [](http://example.com/codeflow?start=15&length=40)

curious - why is this sealed?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/ShowDropColumnsBase.cs
```diff
@@ -169,6 +169,12 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
                 }
             }
 
+            // The result of ShowColumns can be added to (since specifc columns have been chosen), but not DropColumns
+            if (!_isShowColumns)
+            {
+                colsToKeep.IsSealed = argTypes[0].IsSealed;
```
MikeStall, line 175: for example, colsToKeep could point to ReturnType. So this could mutate ReturnType.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -130,6 +130,13 @@ public static IEnumerable<DType> GetPrimitiveTypes()
         // Special case for old enums. 
         public ValueTree ValueTree { get; }
 
+        // The columns of a table or record are sealed and cannot be added to, renamed, or unioned by the formula writer. This allows us to add columns to system records/tables without fear of breaking an existing formula.
+        // All records/tables that are sealed in V1 are also marked sealed in Pre-V1. For the IfError function, since ScopeInfo is defined independent of V1 status, it would be a problem to make sealed V1 dependent.
+        // Instead, a V1 specific check is made at the points that a formula could create a problem, such as AddColumns and record union.
+        // Sealed is NOT checked at internal .Add of fields to a record. This would be a good validation, however we don't have V1 information plumbed down to make that possible.
+        // Note that .Add does not pass through IsSealed. Be sure to set IsSealed after fully creating the type. See the User object in the interpreter, which is sealed, as an example.
+        public bool IsSealed { get; set; }
```
MikeStall, line 138: >set [](http://example.com/codeflow?start=36&length=3)

should this be readonly / init?

My concern is mutating the IsSealed on a type

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
gregli-msft: A table of sealed records should propagate the sealedness.  Shouldn't be able to say `First( Table( User ) )` to avoid the block.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/ShowDropColumnsBase.cs
gregli-msft: colsToKeep is only linked to ReturnType if it isn't _isShowColumns.  In the IsShowColumns case, we start from scratch and build up a record, which because it is ShowColumns, doesn't need to propagate sealedness.  But with the change to making IsSealed immutable this all changed.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
gregli-msft: Yes, I agree.  Change made.

LucGenetier:
✅ No public API change.

MikeStall (dismissed):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #2218, opened 2024-02-20
Add sync APIs for BlobValue

LucGenetier:


src/libraries/Microsoft.PowerFx.Core/Public/Values/BlobValue.cs
```diff
@@ -29,22 +29,46 @@ public override void Visit(IValueVisitor visitor)
 
         public override object ToObject() => this;
 
+        public bool IsString => Content.IsString;
+
+        public bool IsBase64 => Content.IsBase64;
+
+        public bool IsByteArray => Content.IsByteArray;
```
MikeStall, line 36: We need clear invariants on what *exactly* these mean. 

For example, base64 is also a string. 

And byte contents can still be accessed as either Byte or Base64.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/Blob/Base64Blob.cs
```diff
@@ -25,10 +25,26 @@ internal Base64Blob(string base64Str)
             }
         }
 
+        internal override bool IsString => false;
```
MikeStall, line 28: >IsString => false; [](http://example.com/codeflow?start=31&length=18)

why not true?
Base64 are strings. 

What's the rule we're following here.

MikeStall: Chatted offline. What customer really wants is #2220 instead.

PR #2220, opened 2024-02-21
Enable custom BlobValue

madevoge:


src/libraries/Microsoft.PowerFx.Core/Functions/Blob/BlobContent.cs
```diff
@@ -8,17 +8,17 @@
 
 namespace Microsoft.PowerFx.Core.Functions
 {
-    internal abstract class BlobContent
+    public abstract class BlobContent
```
MikeStall, line 11: >public [](http://example.com/codeflow?start=4&length=6)

note you'll need to update PublicSurface_Tests . 

(you'll see this failure when you run tests)

src/tests/Microsoft.PowerFx.Core.Tests/PublicSurfaceTests.cs
```diff
@@ -111,6 +111,7 @@ public void PublicSurface_Tests()
                 "Microsoft.PowerFx.Types.BlankValue",
                 "Microsoft.PowerFx.Types.BlobType",
                 "Microsoft.PowerFx.Types.BlobValue",
+                "Microsoft.PowerFx.Core.Functions.BlobContent",
```
MikeStall, line 114: >     "Microsoft.PowerFx.Core.Functions.BlobContent", [](http://example.com/codeflow?start=11&length=52)

super nit - we should double check if this is the namespace we want

MikeStall (approved):

PR #2230, opened 2024-02-22
Value and Boolean function support for number and Boolean backed Option Sets 

gregli-msft:
At present, there is no good way to get the numeric backing value of a number-backed Option Set.  The OptionSetInfo function indirectly gets the value, as a logical name, but it returns it as a string, and a future PR will likely deprecate this function.  This change allows the direct use of the Value function on option set value if it is backed by a number, including Decimal and Float functions.  Likewise, the Boolean function can be used for option sets backed by a Boolean value.  Text could already be used on option sets backed by text.


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Boolean.cs
```diff
@@ -312,6 +321,92 @@ internal override IR.Nodes.IntermediateNode CreateIRCallNode(PowerFx.Syntax.Call
...
+        public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
+        {
+            yield return new[] { TexlStrings.BooleanLArg1 };
+        }
+
+        public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DType[] argTypes, IErrorContainer errors, out DType returnType, out Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
+        {
+            var fValid = base.CheckTypes(context, args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
+
+            var arg = args[0];
```
adithyaselv, line 348: nit, may be not needed here ?

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
```diff
@@ -55,7 +55,9 @@ internal class BuiltinFunctionsCore
         public static readonly TexlFunction BooleanN = _library.Add(new BooleanNFunction());
         public static readonly TexlFunction BooleanN_T = _library.Add(new BooleanNFunction_T());
         public static readonly TexlFunction BooleanW = _library.Add(new BooleanWFunction());
-        public static readonly TexlFunction BooleanW_T = _library.Add(new BooleanWFunction_T());
+        public static readonly TexlFunction BooleanW_T = _library.Add(new BooleanWFunction_T());
```
lesaltzm, line 58: IIRC this will immediately expose them in PA Client. Do you have the corresponding JS implementations ready?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Boolean.cs
gregli-msft: Good point.  I cleaned up the code, tried to make clearer.

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
gregli-msft: Done.  By making this change, the new functions don't appear in signature help, but I assume that is a testing thing not pulling in the feature gated functions.

LucGenetier:
✅ No public API change.

lesaltzm (approved):

LucGenetier:
✅ No public API change.

PR #2232, opened 2024-02-23
Fix pretty print for tables

LucGenetier:


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #2233, opened 2024-02-23
Give access to another NL assembly

LucGenetier:


LucGenetier:
✅ No public API change.

PR #2235, opened 2024-02-27
Deprecate OptionSetInfo, prepare way for new ChoiceInfo function

gregli-msft:
OptionSetInfo only returns one value, the logical name of an option set choice.  We'd like to redesign this function to accommodate:

1. The logical name is a text wrapped number for number and Boolean backed option sets coming from Dataverse.  We recently added the ability to get at these values with the Value and Boolean functions which is a better way, strongly typed.  https://github.com/microsoft/Power-Fx/pull/2230
1. There are other pieces of information that are worth knowing, for example the color of an option set choice, that can't be accommodated with this design.  
1. Our preferred design is to return an extensible record of information.  We could add an additional second argument with an enum to determine what information to return, with a default value of logical name, but this isn't our first choice.
1. We don't use the term OptionSet any longer in Dataverse, it has been replaced with Choice, which is more consistent with our existing Choices function.

As some hosts may still be using this function, it will remain available as a function that can be enabled by calling config.EnableOptionSetInfo.  Value/Boolean can be used immediately as a workaround, wrapped with Text if needed.  We plan to add a new ChoiceInfo function with a separate PR.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/OptionSetInfo.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/OptionSetInfo.cs
@@ -2,13 +2,18 @@
 // Licensed under the MIT license.
 
 using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.PowerFx.Core.Functions;
+using Microsoft.PowerFx.Core.IR;
 using Microsoft.PowerFx.Core.Localization;
 using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Functions;
+using Microsoft.PowerFx.Types;
 
-namespace Microsoft.PowerFx.Core.Texl.Builtins
+namespace Microsoft.PowerFx.Functions
 {
-    internal sealed class OptionSetInfoFunction : BuiltinFunction
+    internal sealed class OptionSetInfoFunction : BuiltinFunction, IAsyncTexlFunction
     {
         public override bool IsSelfContained => true;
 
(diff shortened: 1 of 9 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Environment/PowerFxConfigExtensions.cs
```diff
@@ -60,5 +61,11 @@ public static void EnableRegExFunctions(this PowerFxConfig config, TimeSpan regE
                 config.AdditionalFunctions.Add(func.Key, func.Value);
             }
         }
+
+        [Obsolete("OptionSetInfo function is deprecated. Use the Value function on an option set backed by a number and the Boolean function on an option set bacekd by a Boolean instead.  A new ChoiceInfo function is in the works for access to logical names.")]
```
LucGenetier, line 65: nit: typo on bacekd

LucGenetier (approved):

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/OptionSetInfo.cs
```diff
@@ -2,13 +2,18 @@
...
 using Microsoft.PowerFx.Core.Localization;
 using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Functions;
+using Microsoft.PowerFx.Types;
 
-namespace Microsoft.PowerFx.Core.Texl.Builtins
+namespace Microsoft.PowerFx.Functions
 {
-    internal sealed class OptionSetInfoFunction : BuiltinFunction
+    internal sealed class OptionSetInfoFunction : BuiltinFunction, IAsyncTexlFunction
```
CarlosFigueiraMSFT, line 16: We should not have any function implementation inside Microsoft.PowerFx.Core. This should be moved to the MS.PFx.Interpreter instead

gregli-msft: Done.

src/libraries/Microsoft.PowerFx.Interpreter/Environment/PowerFxConfigExtensions.cs
gregli-msft: Done.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/OptionSetInfo.cs
```diff
@@ -23,5 +28,20 @@ public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
         {
             yield return new[] { TexlStrings.AboutOptionSetInfoArg1 };
         }
+
+        public async Task<FormulaValue> InvokeAsync(FormulaValue[] args, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
gregli-msft: Done.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

PR #2236, opened 2024-02-28
REPL fix to print blank records

anderson-joyle:
Fix for a small issue I found while using REPL.
`Table({Value:1}, Blank())`


CarlosFigueiraMSFT (approved):
Thanks for fixing this. There is also another problem in that logic, where something like `Table(Blank(), {Value:2}, {Value:3})` does not get the "special treatment for single column table named Value" which it should. Can you also update the logic on lines 122-123 to look at the field names instead of the first element of the table?

LucGenetier:
✅ No public API change.

MikeStall (dismissed):

CarlosFigueiraMSFT (approved):
🚢

PR #2240, opened 2024-03-04
don't suggest fields that are already specified

### LINKED ISSUE #2126: intellisense: don't suggest fields that are already specified
We recently added intellisense support to suggest fields in a record. (#1787 )
It shows all fields. If the field is already specified, it shouldn't be suggested. Ie: 

image

In this way, intellisense will narrow down to the remaining fields.

jas-valgotar:
Fixes #2126

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/SuggestionHandlers/FunctionRecordNameSuggestionHandler.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/SuggestionHandlers/FunctionRecordNameSuggestionHandler.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Linq;
 using Microsoft.PowerFx.Core.Binding.BindInfo;
 using Microsoft.PowerFx.Core.Functions;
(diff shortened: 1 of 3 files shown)


MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/SuggestionHandlers/FunctionRecordNameSuggestionHandler.cs
```diff
@@ -152,9 +153,17 @@ private static bool TryGetParentRecordFieldType(DType aggregateType, TexlNode cu
             internal static bool AddAggregateSuggestions(DType aggregateType, IntellisenseData.IntellisenseData intellisenseData, int cursorPos)
             {
                 var suggestionsAdded = false;
+                var parentRecordNode = intellisenseData.CurNode.Parent as RecordNode;
+                var alreadyUsedFields = parentRecordNode?.Ids.Select(id => id.Name).ToImmutableHashSet();
```
Shpakh-MSFT, line 157: >    var alreadyUsedFields = parentRecordNode?.Ids.Select(id => id.Name).ToImmutableHashSet(); [](http://example.com/codeflow?start=12&length=93)

Consider

                var parentRecordNames = parentRecordNode?.Ids.Select(id => id.Name);
                var alreadyUsedFields = parentRecordNames.ToImmutableHashSet() ?? Enumerable.Empty<DName>().ToImmutableHashSet();


That way, you don't need to check for null in loop below.

Shpakh-MSFT (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/SuggestionHandlers/FunctionRecordNameSuggestionHandler.cs
jas-valgotar: I feel we should keep the existing method as it would give us a slight perf advantage, it would save us from empty object creation and call to contains on that empty collection.

Shpakh-MSFT: empty object is a static object shared across so it doesn't get created every time. Also in future, you don't need worry about this object being null if you extend this code, so from code maintenance pov, I prefer this way.

jas-valgotar: Good point, I'll update it.

LucGenetier:
✅ No public API change.

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #2244, opened 2024-03-07
Mark ScopeCollection as mutable

anderson-joyle:
PA is presenting a bug when mutating a collection with PFxV1 active.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -2882,10 +2882,7 @@ namespace Microsoft.PowerFx.Core.Binding
                 }
                 else if (lookupInfo.Kind == BindKind.ScopeCollection)
                 {
-                    if (lookupInfo.Data is IExternalDataSource ds)
-                    {
-                        _txb.SetMutable(node, ds.IsWritable);
-                    }
+                    _txb.SetMutable(node, true);
                 }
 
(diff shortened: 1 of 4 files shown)


src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -2880,6 +2880,10 @@ public override void Visit(FirstNameNode node)
                         _txb.SetMutable(node, true);
                     }
                 }
+                else if (lookupInfo.Kind == BindKind.ScopeCollection)
+                {
+                    _txb.SetMutable(node, true);
```
anderson-joyle, line 2885: >_txb.SetMutable(node, true); [](http://example.com/codeflow?start=20&length=28)

Maybe we need to check if the datasource is mutable. Something like

```
if (lookupInfo.Data is IExternalDataSource ds)
{
    _txb.SetMutable(node, ds.IsWritable);
}
```

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
anderson-joyle: Only PA creates a `ScopeCollection` collection, and it will always be mutable.

LucGenetier:
✅ No public API change.

Issue #2242, opened 2024-03-05
PowerFx implementation in Dataverse Accelerator (Low Code Plugins) does not yet support IsType


cchannon:
The IsType function in the Dataverse Accelerator implementation of the PowerFx runtime does not yet support IsType (returns the error "IsType is a known but not supported function"). 

This is an issue because AsType on its own is not failure-safe, so the low code plugins cannot safely handle polymorphic attributes. 

I'm willing to help implement this capability with my own Open-Source contributions, just looking for guidance on where to contribute as well as any implementation guidelines the FT team has to offer.

PR #2247, opened 2024-03-07
IsType function from PA

anderson-joyle:
Initial move to support https://github.com/microsoft/Power-Fx/issues/2242.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IsType.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IsType.cs
@@ -1,0 +1,79 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+using Microsoft.PowerFx.Core.App.ErrorContainers;
+using Microsoft.PowerFx.Core.Binding;
+using Microsoft.PowerFx.Core.Binding.BindInfo;
+using Microsoft.PowerFx.Core.Entities;
+using Microsoft.PowerFx.Core.Errors;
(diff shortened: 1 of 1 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IsType.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            // Check if first argument is poly type or an activity pointer
+            if (!argTypes[0].IsPolymorphic && !argTypes[0].IsActivityPointer)
+            {
+                errors.EnsureError(DocumentErrorSeverity.Severe, args[0], TexlStrings.ErrBadType_ExpectedType_ProvidedType, DKind.Polymorphic.ToString(), argTypes[0].GetKindString());
+                return;
+            }
+
+            // !!! This code can be shared among functions that require a connected data source.
+            DType dataSourceType = argTypes[0];
```
adithyaselv, line 56: Just a quick clarification, we want to check here if the record (`arg[0]`) is from a connected datasource as opposed to tableArg (`arg[1]`) that is being done in the PA client side.

anderson-joyle: Good catch!

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

anderson-joyle:
> We should have tests for this function as well if this is moved to this repository.

No breaking changes in PA. Please check PR `9683327`.

PR #2248, opened 2024-03-08
Implement EncodeHTML function

CarlosFigueiraMSFT:
We already have a way to encode URL data with the EncodeUrl function, this adds the capability to do the same with HTML data.


LucGenetier:
✅ No public API change.

anderson-joyle:
What exactly are we breaking with this new function?

CarlosFigueiraMSFT:
> What exactly are we breaking with this new function?

Nothing. But we need to have it implemented in Power Apps before we can complete this PR, as the function is being added to the list of core functions.

LucGenetier:
✅ No public API change.

adithyaselv (approved):
:shipit:

PR #2253, opened 2024-03-13
WIP Defaults function

anderson-joyle:
`Default` function implementation.
Implementing the function since it is going to be used by `Patch` implementation.


LucGenetier:
✅ No public API change.

Issue #2196, opened 2024-02-08
Improve error message when void values are assigned in record.


jas-valgotar:
For Expression where the user by mistake assigns the void type to the record field,
e.g. `{out : Remove()}`
we give the below error.
`Incompatible type. The item you are trying to put into a record has a type that is not compatible with the record.`
which is not actionable. We should tell them which types are expected or least, what is the current type that is not compatible (Ie Void value here).

PR #2255, opened 2024-03-14
Improves error message when putting value inside record and table.

jas-valgotar:
Fixes https://github.com/microsoft/Power-Fx/issues/2196

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
@@ -1077,10 +1077,7 @@ namespace Microsoft.PowerFx.Core.Functions
                         continue;
                     }
 
-                    var fError = false;
-                    DType.Union(ref fError, expectedColumnType, actualColumnType, useLegacyDateTimeAccepts: false, features, allowCoerce: true, unionToLeftTypeOnly: true);
-
-                    if (!fError)
+                    if (DType.TryUnionWithCoerce(expectedColumnType, actualColumnType, features, coerceToLeftTypeOnly: true, out var _, out var _))
                     {
                         continue;
                     }
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
@@ -1054,6 +1054,11 @@ namespace Microsoft.PowerFx.Core.Functions
         // DTypes are guaranteed to be finite, so there is no risk of a call stack overflow
         private bool SetErrorForMismatchedColumnsCore(DType expectedType, DType actualType, TexlNode errorArg, IErrorContainer errors, DPath columnPrefix, Features features)
         {
+            return SetErrorForMismatchedColumnsCore(expectedType, actualType, errorArg, errors, columnPrefix, features, requireAllParamColumnsOverride: RequireAllParamColumns);
+        }
+
+        internal static bool SetErrorForMismatchedColumnsCore(DType expectedType, DType actualType, TexlNode errorArg, IErrorContainer errors, DPath columnPrefix, Features features, bool requireAllParamColumnsOverride)
+        {
             Contracts.AssertValid(expectedType);
             Contracts.AssertValid(actualType);
             Contracts.AssertValue(errorArg);
@@ -1085,7 +1090,8 @@ namespace Microsoft.PowerFx.Core.Functions
                             errorArg,
                             errors,
                             columnPrefix.Append(new DName(errName)),
-                            features);
+                            features,
+                            requireAllParamColumnsOverride);
                     }
 
                     if (expectedColumn.Type.IsExpandEntity
@@ -1105,7 +1111,7 @@ namespace Microsoft.PowerFx.Core.Functions
                 }
 
                 // Second, set column missing message if applicable
-                if (RequireAllParamColumns && !expectedType.AreFieldsOptional)
+                if (requireAllParamColumnsOverride && !expectedType.AreFieldsOptional)
                 {
                     errors.EnsureError(
                         DocumentErrorSeverity.Severe,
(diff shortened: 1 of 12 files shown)


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -3588,7 +3588,7 @@
     <comment>1 How to fix the error. </comment>
   </data>
   <data name="ErrorResource_ErrIncompatibleTypesForEquality_Left_Right_HowToFix_2" xml:space="preserve">
-    <value>If you’re comparing records or tables, the field or column types must match exactly.</value>
+    <value>If you're comparing records or tables, the field or column types must match exactly.</value>
```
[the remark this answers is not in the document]
jas-valgotar: Sure, Updated. Had to change it because xunit test case de-serializer couldn't parse it.

LucGenetier:
✅ No public API change.

src/strings/PowerFxResources.en-US.resx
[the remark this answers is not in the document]
MikeStall: can we avoid this change?
It is customer impacting (barely) - but test infra seems like the wrong motivator to change. Ie, these should be able to be unicode strings with any chars. 

If test is tripping up, you can do something like string.replace() in test to normalize to `'`

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/inTable.txt
```diff
@@ -65,7 +65,7 @@ Errors: Error 7-14: Invalid argument type. Cannot use Record values in this cont
 Errors: Error 12-16: Invalid argument type. Cannot use Boolean values in this context.
 
 >> {a:true} in Table({a:{b:true}})
-Errors: Error 0-8: Invalid argument type. Expecting a Record value, but of a different schema.|Error 0-8: Incompatible type. The 'a' column in the data source you’re updating expects a 'Record' type and you’re using a 'Boolean' type.|Error 12-31: Invalid argument type. Expecting a Table value, but of a different schema.|Error 12-31: Incompatible type. The 'a' column in the data source you’re updating expects a 'Boolean' type and you’re using a 'Record' type.
+Errors: Error 0-8: Invalid argument type. Expecting a Record value, but of a different schema.|Error 0-8: Incompatible type. The 'a' column in the data source you're updating expects a 'Record' type and you're using a 'Boolean' type.|Error 12-31: Invalid argument type. Expecting a Table value, but of a different schema.|Error 12-31: Incompatible type. The 'a' column in the data source you're updating expects a 'Boolean' type and you're using a 'Record' type.
```
CarlosFigueiraMSFT, line 68: >u' [](http://example.com/codeflow?start=161&length=2)

Do you need to make this change, since you are normalizing the single quote on BaseRunner.cs?

CarlosFigueiraMSFT:
>             TestBindingError(script, Features.None, errorMessage, symbolTable);

Not really related to your changes, but we should be using `Features.PowerFxV1`, as no host uses `Features.None`. Same in line 90

---
Refers to: src/tests/Microsoft.PowerFx.Interpreter.Tests/DeferredTypeTests.cs:116 in d1fb31e. [](commit_id = d1fb31ec785684db9b6ee122f80065b5550d0178, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -1040,19 +1040,19 @@ private static bool TryGetArg0AsDsInfo(CallNode callNode, TexlBinding binding, o
             return result;
         }
 
-        protected bool SetErrorForMismatchedColumns(DType expectedType, DType actualType, TexlNode errorArg, IErrorContainer errors, Features features)
+        internal static bool SetErrorForMismatchedColumns(DType expectedType, DType actualType, TexlNode errorArg, IErrorContainer errors, Features features, bool requireAllParamColumns)
```
CarlosFigueiraMSFT, line 1043: >internal [](http://example.com/codeflow?start=8&length=8)

We should revert this to `protected` when the definition for the Collect/Remove functions are moved to the MS.PFx.Core assembly. Please create an issue to track that.

CarlosFigueiraMSFT, line 1043: >requireAllParamColumns [](http://example.com/codeflow?start=163&length=22)

Consider naming it something like `requireAllParamColumnsOverride` to avoid confusion with the instance property of the same name (just different casing for the first letter)

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
jas-valgotar: reverted it to `protected` instead chaining it to static method, so it stays PA non breaking

jas-valgotar:
>             TestBindingError(script, Features.None, errorMessage, symbolTable);

Updated to V1

---
In reply to: [2010778690](https://github.com/microsoft/Power-Fx/pull/2255#issuecomment-2010778690) [](http://example.com/codeflow?ancestors=2010778690)

---
Refers to: src/tests/Microsoft.PowerFx.Interpreter.Tests/DeferredTypeTests.cs:116 in aba9876. [](commit_id = aba9876cdb3ff02478274a5e6222f1d0b8cfd170, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/inTable.txt
jas-valgotar: you are correct, reverted this change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #2257, opened 2024-03-15
Update the Delegation logic for LookUp Call nodes

ian-legler:
Incorrect logic for determining whether a LookUp Call node could be delegated was added in a previous PR https://github.com/microsoft/Power-Fx/pull/2065

Updating the logic to ensure proper handling of edge cases


LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):

PR #2261, opened 2024-03-19
Remove feature that is enabled in all platforms

CarlosFigueiraMSFT:
The CoalesceShortCircuit feature was added to Power Fx to give Power Apps time to catch up. It has caught up already, so we can remove it from the Power Fx repository.


LucGenetier:
✅ No public API change.

anhtruong-MSFT (approved):

LucGenetier:
✅ No public API change.

PR #2264, opened 2024-03-19
Makes DisplayNameProvider public.

jas-valgotar:


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/DisplayNames/SingleSourceDisplayNameProvider.cs
```diff
@@ -11,7 +11,7 @@
 
 namespace Microsoft.PowerFx.Core
 {
-    internal class SingleSourceDisplayNameProvider : DisplayNameProvider
+    public class SingleSourceDisplayNameProvider : DisplayNameProvider
```
MikeStall, line 14: nit: can we keep the class private, and instead have a static builder on the base DisplayNameProvider class?

jas-valgotar: good point, updated.

MikeStall (approved):

LucGenetier:
✅ No public API change.

Issue #2263, opened 2024-03-19
Get literal value from parsed expression


anderson-joyle:
Create a helper to extract a literal value to validate.
Possibly `CheckResult.TryGetAsLiteral(this CheckResult check, out var object value) `.

PR #2265, opened 2024-03-19
Get literal value from parsed expression

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2263.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Utils/Extensions.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Utils/Extensions.cs
@@ -187,13 +187,15 @@ namespace Microsoft.PowerFx.Core.Utils
                     value = booLitNode.Value;
                     return true;
                 case CallNode callNode:
-                    if (callNode.ToString().Contains("GUID("))
+                    if (callNode.IsCall("GUID"))
                     {
-                        var guidStr = callNode.Args.ChildNodes[0].AsStrLit().Value;
-                        if (Guid.TryParse(guidStr, out Guid result))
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Core/Utils/Extensions.cs
```diff
@@ -163,5 +163,45 @@ internal static bool CheckAggregateNames(this DType argType, DType dataSourceTyp
...
+                    value = numLitNode.Value.Value;
+                    return true;
+                case DecLitNode decLitNode:
+                    value = decLitNode.Value.Value;
+                    return true;
+                case BoolLitNode booLitNode:
+                    value = booLitNode.Value;
+                    return true;
+                case CallNode callNode:
+                    if (callNode.ToString().Contains("GUID("))
```
MikeStall, line 190: >callNode.ToString() [](http://example.com/codeflow?start=24&length=19)

nit - don't call ToString() here. 
drill in to the identifier.

src/libraries/Microsoft.PowerFx.Core/Utils/Extensions.cs
```diff
@@ -163,5 +163,45 @@ internal static bool CheckAggregateNames(this DType argType, DType dataSourceTyp
...
+                    {
+                        var guidStr = callNode.Args.ChildNodes[0].AsStrLit().Value;
```
MikeStall, line 192: >.AsStrLit(). [](http://example.com/codeflow?start=65&length=12)

can this return null?

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):

Issue #1550, opened 2023-06-02
Unify Collect /Patch texl functions with PA implementation 


MikeStall:
The Mutation texlFunctions (
https://github.com/microsoft/Power-Fx/tree/main/src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation ) should be shared with PA. 

this may require some refactoring in PA to decouple and make more shareable. 

Could leverage techniques added in recent RegEx PR (#1442) for associating a TexlFunction with an implementation,

anderson-joyle:
This will include https://github.com/microsoft/Power-Fx/issues/1550

PR #2269, opened 2024-03-21
Patch function

anderson-joyle:
Part of https://github.com/microsoft/Power-Fx/issues/1550.

## Breaking Language changes:
None.

## New API features
- For the overloads that requires a base record, if it is not found, `Patch` will create a new record.
- `Patch(dataSource:*[], Record, Updates1, Updates2,…)` overload now can take multiple update records.

- New overloads
  - `Patch(DS, record_with_keys_and_updates)`
  - `Patch(DS, table_of_rows, table_of_updates)`
  - `Patch(DS, table_of_rows_with_updates)`
  - `Patch(Record, Updates1, Updates2,…)`

## Other:
None


src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -254,11 +255,22 @@ internal static TexlFunction FindBestErrorOverload(TexlFunction[] overloads, DTy
...
+                    for (int i = 0; i < cArg; i++)
+                    {
+                        bestCandidate = bestCandidate && candidate.ParamTypes[i].Accepts(argTypes[i], exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: usePowerFxV1CompatibilityRules);
+                    }
+
+                    if (bestCandidate)
+                    {
+                        return candidate;
+                    }
                 }
```
anderson-joyle, line 274: Patch has 5 overloads and binder was suggesting a wrong one when asking for intellisense suggestions.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Patch.cs
```diff
@@ -0,0 +1,595 @@
...
+    }    
+
+    internal abstract class PatchAsyncFunctionCore : PatchAndValidateRecordFunctionBase
+    {
+        public override bool IsAsync => true;
+
+        public override bool ManipulatesCollections => true;
+
+        // !!!TODO Shouldn't this be false?
+        public override bool IsSelfContained => true;
```
CarlosFigueiraMSFT, line 275: >l IsSelfContained => true; [](http://example.com/codeflow?start=27&length=26)

This should be false - and it is the case in Power Apps. The only self-contained Patch overload is the one that merges records (and that is not an asynchronous function)

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Patch.cs
```diff
@@ -0,0 +1,594 @@
...
+using Microsoft.PowerFx.Core.Localization;
+using Microsoft.PowerFx.Core.Texl.Builtins;
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Syntax;
+
+namespace Microsoft.PowerFx.Core.Texl.Builtins
+{
+    #region Base classes
+    internal abstract class PatchAndValidateRecordFunctionBase : FunctionWithTableInput
```
MikeStall, line 22: I'm assuming these are all coming from PAClient?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
```diff
@@ -0,0 +1,279 @@
...
+
+            if (args[0] is LambdaFormulaValue arg0lazy)
+            {
+                arg0 = await arg0lazy.EvalAsync().ConfigureAwait(false);
+            }
+
+            if (arg0 is not TableValue tableValue)
+            {
+                return arg0;
+            }
```
MikeStall, line 33: ? How would this happen?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
```diff
@@ -0,0 +1,279 @@
...
+
+            var arg1Rows = ((TableValue)args[1]).Rows;
+            var arg2Rows = ((TableValue)args[2]).Rows;
+
+            // !!!TODO Do we need to check if args[1] and args[2] counters are equal?
+            // PA does not give any error if they are not equal.
```
MikeStall, line 111: >  // PA does not give any error if they are not equal. [](http://example.com/codeflow?start=10&length=54)

What does PA do if they're not equal? Take the shortest?

src/libraries/Microsoft.PowerFx.Interpreter/ParsedExpression.cs
```diff
@@ -139,6 +141,12 @@ public async Task<FormulaValue> EvalAsync(CancellationToken cancellationToken, I
                 ServiceProvider = new BasicServiceProvider(runtimeConfig?.ServiceProvider, hasInnerServices ? innerServices : null)
             };
 
+            if (_features != null)
+            {
+                innerServices.AddService(_features);
+                hasInnerServices = true;
+            }
```
CarlosFigueiraMSFT, line 148: The local `hasInnerServices` is only used 6 lines above; no need to set it again

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
```diff
@@ -0,0 +1,279 @@
...
+
+namespace Microsoft.PowerFx.Core.Texl.Builtins
+{
+    // Patch(dataSource:*[], Record, Updates1, Updates2,…)
+    internal class PatchImpl : PatchFunction, IAsyncTexlFunction5
+    {
+        public async Task<FormulaValue> InvokeAsync(IServiceProvider runtimeServiceProvider, FormulaType irContext, FormulaValue[] args, CancellationToken cancellationToken)
+        {
+            MutationProcessUtils.ThrowIfPFxV1NotActive(runtimeServiceProvider.GetService<Features>(), "Patch");
+
```
CarlosFigueiraMSFT, line 22: We don't have any interpreter clients that don't have PFxV1 enabled - the only client where this is not enabled today is Power Apps. Is this check necessary? Same question for the other implementations.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
```diff
@@ -0,0 +1,279 @@
...
+            }
+
+            var dvalue = MutationProcessUtils.MergeRecords(args.Skip(2));
+
+            if (!dvalue.IsValue)
+            {
+                return dvalue.ToFormulaValue();
+            }
+
+            // If base record is {}, then collect.
```
CarlosFigueiraMSFT, line 47: >// If base record is {}, then collect. [](http://example.com/codeflow?start=12&length=38)

Not exactly; we will collect if we cannot find the base record in the table. For example, after `ClearCollect(coll, {a:1,b:2},{a:3,b:4}); Patch(coll, {a:2,b:3}, {a:5,b:6})`, coll will have 3 elements - `[{"a":1,"b":2},{"a":3,"b":4},{"a":5,"b":6}]`

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
CarlosFigueiraMSFT, line 111: Currently Power Apps doesn't validate it - it effectively takes the smaller of the lengths and goes with it. Returning an error seems appropriate in this case.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
```diff
@@ -0,0 +1,279 @@
...
+                }
+
+                if (updatesRecord.IsBlank)
+                {
+                    continue;
+                }
+
+                DValue<RecordValue> result = null;
+
+                // If base record is {}, then collect.
```
CarlosFigueiraMSFT, line 148: >// If base record is {}, then collect. [](http://example.com/codeflow?start=16&length=38)

Same comment from PatchImpl

src/tests/Microsoft.PowerFx.Interpreter.Tests/PatchFunctionTests.cs
```diff
@@ -27,26 +27,26 @@ public class PatchFunctionTests : PowerFxTest
...
+        //[Theory]
+        //[InlineData(typeof(PatchFunction))]
+        //public async Task CheckArgsTestAsync(Type type)
+        //{
+        //    var expressionError = new ExpressionError()
+        //    {
+        //        Kind = ErrorKind.ReadOnlyValue,
+        //        Severity = ErrorSeverity.Critical,
+        //        Message = "Something went wrong"
+        //    };
```
CarlosFigueiraMSFT, line 39: Why is all this code commented out?

src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterSuggestTests.cs
```diff
@@ -177,11 +177,12 @@ public void TestSuggestVariableName(string suggestion)
...
         [InlineData("Collect(Table({a:1, b:2}), {|", "a:", "b:")]
         [InlineData("Collect(Table({a:1, 'test space': \"test\"), {|", "a:", "'test space':")]
-        [InlineData("Patch(Table({a:1, b:2}), {|", "a:", "b:")]
-        [InlineData("Patch({a:1, b:2}, {|", "a:", "b:")]
         [InlineData("ClearCollect(Table({a:1, b:2}), {|", "a:", "b:")]
         [InlineData("Remove(Table({a:1, b:2}), {|", "a:", "b:")]
         [InlineData("Error(Ab|  Collect()", "Abs", "Color.OliveDrab", "ErrorKind.NotApplicable", "JSONFormat.FlattenValueTables", "Match.Tab", "Table")]
+
+        //[InlineData("Patch({a:1, b:2}, {|", "a:", "b:")] This test case will demand a binder change to work.
+        //[InlineData("Patch(Table({a:1, b:2}), {|", "a:", "b:")] This test case will demand a binder change to work.
```
CarlosFigueiraMSFT, line 185: We should make the binder change with this PR instead of disabling it.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
```diff
@@ -0,0 +1,279 @@
...
+                }
+
+                resultRows.Add(result);
+            }
+
+            return new InMemoryTableValue(IRContext.NotInSource(tableValue.Type), resultRows);
+        }
+    }
+
+    // If arg1 is pure PFx record, it will return a runtime internal error.
```
CarlosFigueiraMSFT, line 170: >If arg1 is pure PFx record, it will return a runtime internal error. [](http://example.com/codeflow?start=7&length=68)

Do we have tests for that?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
```diff
@@ -0,0 +1,279 @@
...
+
+    // Patch(Record, Updates1, Updates2,…)
+    internal class PatchRecordImpl : PatchRecordFunction, IAsyncTexlFunction5
+    {
+        public async Task<FormulaValue> InvokeAsync(IServiceProvider runtimeServiceProvider, FormulaType irContext, FormulaValue[] args, CancellationToken cancellationToken)
+        {
+            MutationProcessUtils.ThrowIfPFxV1NotActive(runtimeServiceProvider.GetService<Features>(), "Patch");
+
+            return MutationProcessUtils.MergeRecords(args).ToFormulaValue();
+        }
```
CarlosFigueiraMSFT, line 233: This is not a mutation function, it's a pure one (no side-effects, self-contained). The implementation can go to the core library.

src/libraries/Microsoft.PowerFx.Interpreter/Environment/PowerFxConfigExtensions.cs
```diff
@@ -39,7 +39,11 @@ public static void EnableMutationFunctions(this SymbolTable symbolTable)
         {
             symbolTable.AddFunction(new RecalcEngineSetFunction());
             symbolTable.AddFunction(new CollectFunction());
-            symbolTable.AddFunction(new PatchFunction());
+            symbolTable.AddFunction(new PatchImpl());
+            symbolTable.AddFunction(new PatchSingleRecordImpl());
+            symbolTable.AddFunction(new PatchAggregateImpl());
+            symbolTable.AddFunction(new PatchAggregateSingleTableImpl());
+            symbolTable.AddFunction(new PatchRecordImpl());
```
CarlosFigueiraMSFT, line 46: >symbolTable.AddFunction(new PatchRecordImpl() [](http://example.com/codeflow?start=12&length=45)

This is not a mutation function - it's a pure function with no side effects. It should go to the core list.

src/libraries/Microsoft.PowerFx.Core/Utils/MutationUtils.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        public static string GetScalarSingleColumnNameForType(Features features, DKind kind)
+        {
+            return kind switch
+            {
+                DKind.Image or
+                DKind.Hyperlink or
+                DKind.Media or
+                DKind.Blob or
+                DKind.PenImage => features.ConsistentOneColumnTableResult ? TableValue.ValueName : "Url",
```
CarlosFigueiraMSFT, line 76: >ConsistentOneColumnTableResult [](http://example.com/codeflow?start=43&length=30)

This is now enabled in all platforms, we don't need this helper anymore (calls to GetScalarSingleColumnNameForTypes can use TableValue.ValueName directly)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch_V1Compat.txt
```diff
@@ -41,3 +44,238 @@ Errors: Error 13-143: The value passed to the 'Patch' function cannot be changed
     {Properties:{Moon:{Name:"Phobos"}}},
     {Planet:"Jupter"})
 Errors: Error 13-251: The value passed to the 'Patch' function cannot be changed.
+
+>> Patch(t1, r1, r_empty);
```
CarlosFigueiraMSFT, line 48: >>> [](http://example.com/codeflow?start=0&length=2)

Can you add a test before this one just with `t1` - that will make it easier to know what was the initial value just by looking at the test.

CarlosFigueiraMSFT: Same with `r1` and `r2`

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch_V1Compat.txt
```diff
@@ -41,3 +44,238 @@ Errors: Error 13-143: The value passed to the 'Patch' function cannot be changed
...
+Table({Field1:2,Field2:"moon",Field3:DateTime(2022,2,1,0,0,0,0),Field4:false})
+
+>> Patch(t1, r1, r2)
+{Field1:2,Field2:"moon",Field3:DateTime(2022,2,1,0,0,0,0),Field4:false}
+
+>> Patch(t1, r1, {Field5:"Field5"});
+   t1
+Errors: Error 14-31: The specified column 'Field5' does not exist.|Error 0-32: The function 'Patch' has some invalid arguments.
+
+// This test would fail in PA.
```
CarlosFigueiraMSFT, line 81: >This test would fail in PA. [](http://example.com/codeflow?start=3&length=27)

Why would this fail?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch_V1Compat.txt
```diff
@@ -41,3 +44,238 @@ Errors: Error 13-143: The value passed to the 'Patch' function cannot be changed
     {Properties:{Moon:{Name:"Phobos"}}},
     {Planet:"Jupter"})
 Errors: Error 13-251: The value passed to the 'Patch' function cannot be changed.
+
+>> Patch(t1, r1, r_empty);
+   t1
+Table({Field1:1,Field2:"earth",Field3:DateTime(2022,1,1,0,0,0,0),Field4:true})
+
+>> Patch(t1, r1, {Field2:"mars"});
```
CarlosFigueiraMSFT, line 52: >Patch(t1, r1, {Field2:"mars"}); [](http://example.com/codeflow?start=3&length=31)

Here we updated the table `t1`, so that the record which used to match `r1` doesn't match it anymore (we shouldn't be updating r1 when we patch the table). The subsequent Patch calls should not be updating the same record.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch_V1Compat.txt
```diff
@@ -41,3 +44,238 @@ Errors: Error 13-143: The value passed to the 'Patch' function cannot be changed
...
+
+>> Patch(t1, First(t1), {DisplayNameField1:1/0});
+   t1
+Table({Field1:Error({Kind:ErrorKind.Div0}),Field2:"earth",Field3:DateTime(2022,1,1,0,0,0,0),Field4:true})
+
+>> Patch(t1, First(t1), {DisplayNameField1:1/0}, {DisplayNameField1:Blank()}, {DisplayNameField1:0});
+   t1
+Table({Field1:0,Field2:"earth",Field3:DateTime(2022,1,1,0,0,0,0),Field4:true})
+
+>> Patch(t1, First(t1), {DisplayNameField2:"jupter"});
```
CarlosFigueiraMSFT, line 126: >jupter [](http://example.com/codeflow?start=44&length=6)

[minor] can you change it to `jupiter` (proper spelling)?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch_V1Compat.txt
```diff
@@ -41,3 +44,238 @@ Errors: Error 13-143: The value passed to the 'Patch' function cannot be changed
...
+Table({Field1:100000,Field2:"athenas",Field3:DateTime(2022,2,1,0,0,0,0),Field4:true},{Field1:200000,Field2:"zeus",Field3:DateTime(2022,2,1,0,0,0,0),Field4:true})
+
+
+// **********************************************************
+// Patch(DS, table_of_rows_with_updates)
+// **********************************************************
+
+>> Patch(t1, Table(r1))
+Error({Kind:ErrorKind.Internal})
+
```
CarlosFigueiraMSFT, line 269: Why is it an internal error? Internal errors are bugs (something that should never happen), they should not be "spec'd" in tests.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DecimalBoot.txt
```diff
@@ -13,9 +13,6 @@ Blank()
 >> Boolean(Filter([1,2,3], Value > 10))
 Table()
 
->> ForAll(Sequence(5), Collect(t_empty, {Value:ThisRecord.Value})); ForAll(Sequence(CountRows(t_empty)), Collect(t_empty2, Index(t_empty, ThisRecord.Value))); Patch(t_empty, First(t_empty), {Value:99}); Patch(t_empty2, Last(t_empty2), {Value:88}); Concat(t_empty, Value) & "," & Concat(t_empty2, Value)
-"992345,123488"
-
 >> ForAll(Boolean([3, 1/0, 0, Sqrt(-1)]), IfError(Text(Value), $"ErrorKind={FirstError.Kind}"))
```
CarlosFigueiraMSFT, line 16: Why is the following test removed?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch_V1Compat.txt
anderson-joyle: Method `TableValue.PatchSingleRecordCoreAsync`  should be overwritten by a child class. This follows the same pattern of other methods.

Please check [https://github.com/microsoft/Power-Fx/blob/78ad88658ec02308c1f1108b0729f42518f50fb7/src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs#L136](https://github.com/microsoft/Power-Fx/blob/78ad88658ec02308c1f1108b0729f42518f50fb7/src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs#L136)

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/DecimalBoot.txt
anderson-joyle: This test case has PFxV1 deactivated. I moved it to `Patch_V1Compat.txt` file.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Patch.cs
anderson-joyle: Yes, with minor changes.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
anderson-joyle: Yes, this is the topic that I brought to discussion internally. Once I have an answer I'll address this.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
anderson-joyle: `Patch(Blank(),Blank(),Blank())`

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
anderson-joyle: Apparently it works anyway. I'll bring this to a internal discussion.

anderson-joyle: I'll address this internally.

src/tests/Microsoft.PowerFx.Interpreter.Tests/PatchFunctionTests.cs
anderson-joyle: I intended to remove this whole file since `PatchFunction` class no longer exists. On second thought, I modified this test case to the new classes. Thanks for pointing this out.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
anderson-joyle: Yes. `Patch_V1Compat.txt`:
```
>> Patch(t1, Table(r1))
Error({Kind:ErrorKind.Internal})
```

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
anderson-joyle: Although I agree this could be moved, in terms of code maintainability its better to leave it this way IMO. Patch is a mutation function (with this  overload as exception).

src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterSuggestTests.cs
anderson-joyle: I suggest to leave binder change for after this PR. It will demand some extra thinking since there are other functions with multiple overloads.

src/libraries/Microsoft.PowerFx.Interpreter/Environment/PowerFxConfigExtensions.cs
anderson-joyle: Moved to Core.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch_V1Compat.txt
anderson-joyle: Each of these test cases dont persist. Each test case is a standalone one.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch_V1Compat.txt
anderson-joyle: Added at the beginning.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch_V1Compat.txt
anderson-joyle: Old comment. Removed.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
anderson-joyle: PFxV1 will try to match both arg sizes. If they are different, return an error. We are discussing a new `Patch` overload.
https://github.com/microsoft/Power-Fx/issues/2277

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
anderson-joyle: If the record is found in the table, update it. Otherwise, create it. For now, we are matching all fields to find the record. This is still open to discussion.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
anderson-joyle: PFxV1 will try to match the args sizes.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch_V1Compat.txt
CarlosFigueiraMSFT: Ah, you're right, I was mixing it up with the tests that run under MutationScripts

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch_V1Compat.txt
```diff
@@ -39,5 +52,242 @@ Errors: Error 13-143: The value passed to the 'Patch' function cannot be changed
...
+  Patch(t1, Table(r1,r2,Blank()),Table({Field1:100000,Field2:"athenas",Field3:DateTime(2022,2,1,0,0,0,0),Field4:true},{Field1:200000,Field2:"zeus",Field3:DateTime(2022,2,1,0,0,0,0),Field4:true},{Field1:999999,Field2:"zeus",Field3:DateTime(2022,2,1,0,0,0,0),Field4:true}));t1
+Table({Field1:100000,Field2:"athenas",Field3:DateTime(2022,2,1,0,0,0,0),Field4:true},{Field1:200000,Field2:"zeus",Field3:DateTime(2022,2,1,0,0,0,0),Field4:true})
+
+
+// **********************************************************
+// Patch(DS, table_of_rows_with_updates)
+// **********************************************************
+
+>> Patch(t1, Table(r1))
+Error({Kind:ErrorKind.Internal})
```
CarlosFigueiraMSFT, line 280: >Internal [](http://example.com/codeflow?start=22&length=8)

If we don't support this overload, we should either return ErrorKind.NotSupported or (preferably) a compile-time error. ErrorKind.Internal are product bugs, and tests can be seen as product specification - we shouldn't have bugs in the specification itself.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch_V1Compat.txt
```diff
@@ -39,5 +52,242 @@ Errors: Error 13-143: The value passed to the 'Patch' function cannot be changed
...
+{ErrorField:Error({Kind:ErrorKind.Div0}),Field1:2,Field2:"moon",Field3:DateTime(2022,2,1,0,0,0,0),Field4:false}
+
+>> Patch(r1, r2, 1/0)
+Errors: Error 0-18: The function 'Patch' has some invalid arguments.|Error 15-16: Cannot use a non-record value in this context: '1 / 0'.
+
+
+
+
+// **********************************************************
+// Patch(DS, table_of_rows, table_of_updates)
```
CarlosFigueiraMSFT, line 252: >table_of_updates [](http://example.com/codeflow?start=28&length=16)

The list of updates don't need to have all fields from the table. Consider adding some tests where only 1 or 2 fields are updated.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch_V1Compat.txt
```diff
@@ -39,5 +52,242 @@ Errors: Error 13-143: The value passed to the 'Patch' function cannot be changed
...
+Errors: Error 6-23: The value passed to the 'Patch' function cannot be changed.
+
+>> ForAll(Sequence(5), Collect(t_empty, {Value:ThisRecord.Value})); ForAll(Sequence(CountRows(t_empty)), Collect(t_empty2, Index(t_empty, ThisRecord.Value))); Patch(t_empty, First(t_empty), {Value:99}); Patch(t_empty2, Last(t_empty2), {Value:88}); Concat(t_empty, Value) & "," & Concat(t_empty2, Value)
+"992345,123488"
+
+
+
+// **********************************************************
+// Patch(Record, Updates1, Updates2,…)
+// **********************************************************
```
CarlosFigueiraMSFT, line 231: Consider adding more tests with literal records, and with some scenarios where this is interesting.

Some scenario examples:

- Adding an index to a table:

    With(
      { tbl:Table({a:1,b:2}, {a:11,b:22}, {a:111,b:222}) }
      ForAll(
        Sequence(CountRows(tbl)),
        Patch(
          { Index: ThisRecord.Value },
          Index(tbl, ThisRecord.Value)
        )
      )
    )

- Flattening record fields

    With(
      { rec: { Name: "John Doe", Address: { Country: "USA", State: "WA", City: "Seattle", Street: "85 Pike Street" } } },
      Patch(
        DropColumns(rec, Address),
        rec.Address
      )
    )

src/libraries/Microsoft.PowerFx.Core/Utils/MutationUtils.cs
```diff
@@ -0,0 +1,102 @@
...
+                return;
+            }
+        }
+
+        /// <summary>
+        /// Merges all records starting from startIndex in args into a single record. Collisions are resolved by last-one-wins.
+        /// </summary>
+        /// <param name="records"></param>
+        /// <returns></returns>
+        public static DValue<RecordValue> MergeRecords(IEnumerable<FormulaValue> records)
```
CarlosFigueiraMSFT, line 74: >MergeRecords [](http://example.com/codeflow?start=42&length=12)

Does it need to be in the core? This looks like it belongs in the interpreter (MS.PFx.Interpreter).

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
```diff
@@ -0,0 +1,252 @@
...
+            if (arg0 is not TableValue tableValue)
+            {
+                return arg0;
+            }
+
+            var arg1Rows = ((TableValue)args[1]).Rows;
+            var arg2Rows = ((TableValue)args[2]).Rows;
+
+            // !!!TODO Do we need to check if args[1] and args[2] counters are equal?
+            // PA does not give any error if they are not equal.
```
CarlosFigueiraMSFT, line 112: Can be removed.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
```diff
@@ -0,0 +1,252 @@
...
+            if (arg1Rows.Count() != arg2Rows.Count())
+            {
+                return CommonErrors.CustomError(IRContext.NotInSource(tableValue.Type), "Both aggregate args must have the same number of records.");
```
CarlosFigueiraMSFT, line 115: >CustomError [](http://example.com/codeflow?start=36&length=11)

Invalid argument, not a custom error

src/libraries/Microsoft.PowerFx.Core/Utils/MutationUtils.cs
anderson-joyle: I've moved `PatchRecord` overload to core, and this is both core and interpreter use util, although this particular overload is not a mutation per se.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryText.cs
```diff
@@ -833,6 +833,11 @@ public static FormulaValue EncodeHTML(IRContext irContext, StringValue[] args)
             return new StringValue(irContext, encoded);
         }
 
+        public static FormulaValue PatchRecord(IRContext irContext, FormulaValue[] args)
+        {
+            return MutationUtils.MergeRecords(args).ToFormulaValue();
+        }
+
```
CarlosFigueiraMSFT, line 840: This shouldn't be in LibraryText. LibraryTable is not ideal (this is only for records, after all), but is better than text.

anderson-joyle: Thanks for catching this. I probably clicked over the wrong file since I was aiming for LibraryTable originally. The fact they are next to each other validates my excuse! 😁

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch_V1Compat.txt
anderson-joyle: Great suggestion! This unveiled a very tricky bug that took me some time to fix.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalDataSource.cs
```diff
@@ -19,6 +19,8 @@ internal interface IExternalDataSource : IExternalEntity, IExternalPageableSymbo
 
         bool RequiresAsync { get; }
 
+        bool IsWritable { get; }
```
MikeStall, line 22: >       bool IsWritable { get; } [](http://example.com/codeflow?start=1&length=31)

fyi, this will be BreakingInternal. adding tag

src/libraries/Microsoft.PowerFx.Interpreter/ParsedExpression.cs
```diff
@@ -133,6 +135,12 @@ public async Task<FormulaValue> EvalAsync(CancellationToken cancellationToken, I
                 hasInnerServices = true;
             }
 
+            if (_features != null)
+            {
+                innerServices.AddService(_features);
```
MikeStall, line 140: >             innerServices.AddService(_features); [](http://example.com/codeflow?start=3&length=49)

since interpreter only supports V1, do we need feature flags?

anderson-joyle: I decided to leave this check because other hosts can forcibly pass `Features` without the PFxV1 flag active.

MikeStall: oh, I see  - this is used to enforce these aren't executed in pre-v1 mode. cool

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -663,7 +663,8 @@ public override IntermediateNode Visit(FirstNameNode node, IRTranslatorContext c
                         }
 
                     case BindKind.OptionSet:
-                    case BindKind.PowerFxResolvedObject:
+                    case BindKind.PowerFxResolvedObject:
+                    case BindKind.Data:
                         {
```
CarlosFigueiraMSFT, line 668: What is the scenario for this?

src/libraries/Microsoft.PowerFx.Core/Public/Values/InMemoryRecordValue.cs
```diff
@@ -83,7 +92,7 @@ protected async Task<DValue<RecordValue>> UpdateAllowedFieldsAsync(RecordValue c
                 return await base.UpdateFieldsAsync(changeRecord, cancellationToken).ConfigureAwait(false);
             }
 
-            await foreach (var field in changeRecord.GetFieldsAsync(cancellationToken).ConfigureAwait(false))
+            await foreach (var field in changeRecord.GetOriginalFieldsAsync(cancellationToken).ConfigureAwait(false))
```
CarlosFigueiraMSFT, line 95: >GetOriginalFieldsAsync [](http://example.com/codeflow?start=53&length=22)

Can you explain this change? Why didn't the previous code work?

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
```diff
@@ -109,6 +109,16 @@ public async IAsyncEnumerable<NamedValue> GetFieldsAsync([EnumeratorCancellation
                 Func<Task<FormulaValue>> getFormulaValue = async () => await GetFieldAsync(fieldName, cancellationToken).ConfigureAwait(false);
                 yield return new NamedValue(fieldName, getFormulaValue, backingDType);
             }
+        }
+
+        /// <summary>
+        /// Returns only the actual fields present in the record. This prevents fields from wrappers.
```
CarlosFigueiraMSFT, line 115: >This prevents fields from wrappers [](http://example.com/codeflow?start=66&length=34)

It's a little confusing. Does it mean "This prevents fields that are not defined in the record, but have *implicit* blank values because of the type, from being returned."?

src/libraries/Microsoft.PowerFx.Interpreter/ParsedExpression.cs
CarlosFigueiraMSFT: But no hosts can pass Features without the PFxV1 flag active. There is only one public member of the Features class, which is Features.PowerFxV1. Everything else is internal (i.e., only Power Apps can see it, and it doesn't use the interpreter).. I don't think this is necessary.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Patch.cs
```diff
@@ -0,0 +1,601 @@
...
+        {
+            if (arity > 3)
+            {
+                return GetGenericSignatures(arity, TexlStrings.PatchArg_Source, TexlStrings.PatchArg_Record, TexlStrings.PatchArg_Update);
+            }
+
+            return base.GetSignatures(arity);
+        }
+
+        internal override IntermediateNode CreateIRCallNode(PowerFx.Syntax.CallNode node, IRTranslatorContext context, List<IntermediateNode> args, ScopeSymbol scope)
```
CarlosFigueiraMSFT, line 440: >CreateIRCallNode [](http://example.com/codeflow?start=43&length=16)

If this override only calls the base with the same arguments, do we need it?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/MutationUtils.cs
```diff
@@ -31,5 +33,38 @@ public static void CheckForReadOnlyFields(DType dataSourceType, TexlNode[] args,
                 }
             }
         }
+
+        /// <summary>
+        /// Merges all records starting from startIndex in args into a single record. Collisions are resolved by last-one-wins.
```
CarlosFigueiraMSFT, line 38: >starting from startIndex in args [](http://example.com/codeflow?start=31&length=32)

Stale comment? There's no `startIndex` or `args` here.

src/libraries/Microsoft.PowerFx.Core/Public/Values/InMemoryRecordValue.cs
anderson-joyle: `GetFieldsAsync` will return all fields based on the record type. For instance, the expression `Table({a:1},{b:"1"})` will return `*[a:w, b:s]`.
Now, if we traverse the records and iterate using `GetFieldsAsync`, we will get both `a` and `b` fields, which will return `Blank()` depending on the record. This could potentially update unwanted fields in the datasource to `Blank()`.

Example (bug):
`Set(t1, Table({a:1,b:1,c:1},{a:2,b:2,c:2}))`
`Patch(t1, Table({a:1, b:99}))`
`t1 => Table({a:1,b:99,c:Blank()},{a:2,b:2,c:2})`

src/libraries/Microsoft.PowerFx.Core/Public/Values/RecordValue.cs
anderson-joyle: Thats correct. I'll change the description since your suggestion is better.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/MutationUtils.cs
```diff
@@ -31,5 +33,38 @@ public static void CheckForReadOnlyFields(DType dataSourceType, TexlNode[] args,
...
+                }
+
+                if (fv is RecordValue recordValue)
+                {
+                    foreach (var field in recordValue.Fields)
+                    {
+                        mergedFields[field.Name] = field.Value;
+                    }
+                }
+            }
```
CarlosFigueiraMSFT, line 65: Suggestion (more compact):

    switch (fv)
    {
        case ErrorValue errorValue:
            return DValue<RecordValue>.Of(errorValue);
        case RecordValue recordValue:
            foreach (var field in recordValue.Fields)
            {
                mergedFields[field.Name] = field.Value;
            }

            break;
    }

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
anderson-joyle: This was necessary to support a new `TableValue` derived class (`EntityTableValue`). This new class simulates a database table and has `AssociatedDataSource` within its type.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Patch.cs
anderson-joyle: No. Thanks for pointing this out.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
```diff
@@ -0,0 +1,255 @@
...
+            if (result.IsError && result.Error is ErrorValue errorvalue && errorvalue.Errors.Any(err => err.Kind == ErrorKind.NotFound))
+            {
+                return (await tableValue.AppendAsync(dvalue.Value, cancellationToken).ConfigureAwait(false)).ToFormulaValue();
+            }
+
+            return result.ToFormulaValue();
+        }
+    }
+
+    // If arg1 is pure PFx record, it will return a runtime internal error.
```
CarlosFigueiraMSFT, line 60: >runtime internal error [](http://example.com/codeflow?start=52&length=22)

It should not be an internal error (those are bugs), it should be an invalid argument.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch_PatchRecord.txt
```diff
@@ -0,0 +1,27 @@
...
+>> Patch(Blank(), {Field1:1,Field2:"earth",Field3:DateTime(2022,1,1,0,0,0,0),Field4:true}, {Field1:2,Field2:"moon",Field3:DateTime(2022,2,1,0,0,0,0),Field4:false})
+{Field1:2,Field2:"moon",Field3:DateTime(2022,2,1,0,0,0,0),Field4:false}
+
+>> Patch({Field1:1,Field2:"earth",Field3:DateTime(2022,1,1,0,0,0,0),Field4:true}, {Field1:2,Field2:"moon",Field3:DateTime(2022,2,1,0,0,0,0),Field4:false}, {ErrorField:1/0})
+{ErrorField:Error({Kind:ErrorKind.Div0}),Field1:2,Field2:"moon",Field3:DateTime(2022,2,1,0,0,0,0),Field4:false}
+
+>> Patch({Field1:1,Field2:"earth",Field3:DateTime(2022,1,1,0,0,0,0),Field4:true}, {Field1:2,Field2:"moon",Field3:DateTime(2022,2,1,0,0,0,0),Field4:false}, 1/0)
+Errors: Error 153-154: Cannot use a non-record value in this context: '1 / 0'.|Error 0-156: The function 'Patch' has some invalid arguments.
+
+// Flattening record fields
```
CarlosFigueiraMSFT, line 22: More interesting tests:

- Patching error records (will return error): Patch({a:1},If(1/0<2,{b:3}))
- Patch with incompatible types (it returns a compile-time error): Patch({a:1},{a:[2]})
- Patch with compatible types (I don't know if it works, but it will be good to have the tests to spec what it does):Patch({a:1},{a:"2"}), Patch({a:"1"},{a:2})
- Patch with partial updates: Patch({a:1,b:2},{b:3})
- Patch with multiple partial updates: Patch({a:1,b:"hello",c:false}, {b:"hello 2"},{c:true},{d:[1,2,3]})
- Patch with record columns: Patch({name:"John Doe",address:{street:"One Main St",city:"Springfield",country:"USA"}},{address:{city:"Smallville"}}) --> does it replace the entire 'address' record, or only the nested 'city'? I think the former, but we should specify it as well
- Same with tables instead of records
- Records with blank properties: Patch({a:1,b:"hello"},{b:Blank()})

src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationFunctionsTests.cs
```diff
@@ -364,6 +368,112 @@ public void DontCopyDerivedRecordValuesTest(string expression)
...
+            var evaluator = check.GetEvaluator();
+
+            // Runtime exception
+            Assert.ThrowsAsync<InvalidOperationException>(async () => await evaluator.EvalAsync(CancellationToken.None).ConfigureAwait(false)).ConfigureAwait(false);
+        }
+
+        [Theory]        
+        [InlineData("Patch(t1, {accountid:GUID(\"00000000-0000-0000-0000-000000000001\"),name:\"Mary Doe\"});First(t1).name", "Mary Doe")]
+        [InlineData("Patch(t1, {accountid:GUID(\"00000000-0000-0000-0000-000000000001\"),address1_city:\"Seattle\"});First(t1).name", "John Doe")]
+        [InlineData("Patch(t1, {accountid:GUID(\"00000000-0000-0000-0000-000000000003\"),name:\"Microsoft Corporation\",address1_city:\"Seattle\"});Last(t1).name", "Microsoft Corporation")]
```
CarlosFigueiraMSFT, line 396: >Last(t1).name [](http://example.com/codeflow?start=148&length=13)

Can you also add the number of elements in the expression, so that we will show that it added one? Something like `$"{CountRows(t1) - {Last(t1).name}"`?

src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationFunctionsTests.cs
```diff
@@ -364,6 +368,112 @@ public void DontCopyDerivedRecordValuesTest(string expression)
...
+        [InlineData("Patch(t1, Table({accountid:GUID(\"00000000-0000-0000-0000-000000000001\"),name:\"Emily Doe\"},{accountid:GUID(\"00000000-0000-0000-0000-000000000002\"),name:\"Benjamin Doe\"}));Last(t1).name", "Benjamin Doe")]
```
CarlosFigueiraMSFT, line 397: >Last(t1).name [](http://example.com/codeflow?start=198&length=13)

This should also change the first name? Can you update the expression to something like `Concat(t1, $"{name} from {address1_city}", ",")`? That will show all updates.

src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationFunctionsTests.cs
```diff
@@ -364,6 +368,112 @@ public void DontCopyDerivedRecordValuesTest(string expression)
...
+        [InlineData("Patch(t1, Table({accountid:GUID(\"00000000-0000-0000-0000-000000000001\"),address1_city:\"Miami\"},{accountid:GUID(\"00000000-0000-0000-0000-000000000002\"),address1_city:\"Orlando\"}));Last(t1).name", "Sam Doe")]
```
CarlosFigueiraMSFT, line 398: >Last(t1).name [](http://example.com/codeflow?start=207&length=13)

Same comment as the previous test

src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationFunctionsTests.cs
```diff
@@ -364,6 +368,112 @@ public void DontCopyDerivedRecordValuesTest(string expression)
...
+        [InlineData("Patch(t1, Table({accountid:GUID(\"00000000-0000-0000-0000-000000000003\"),name:\"Microsoft Corporation\",address1_city:\"Seattle\"},{accountid:GUID(\"00000000-0000-0000-0000-000000000004\"),name:\"Bill Gates\",address1_city:\"Seattle\"}));Last(t1).name", "Bill Gates")]
+        public void MutationEntityTests(string expression, string expected)
```
CarlosFigueiraMSFT, line 400: >MutationEntityTests [](http://example.com/codeflow?start=20&length=19)

Can you add a test with the Patch(dataSource, <table of records>, <table of updates>) overload?

src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationFunctionsTests.cs
anderson-joyle: Good suggestion. Changed.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch_PatchRecord.txt
```diff
@@ -0,0 +1,66 @@
...
+>> Patch({a:1,b:"hello",c:false}, {b:"hello 2"},{c:true},{d:[1,2,3]})
+{a:1,b:"hello 2",c:true,d:Table({Value:1},{Value:2},{Value:3})}
+
+>> Patch({a:[1,2,3]}, {a:[4,5,6]})
+{a:Table({Value:4},{Value:5},{Value:6})}
+
+>> Patch({name:"John Doe",address:{street:"One Main St",city:"Springfield",country:"USA"}},{address:{street:"Avenida Paulista",city:"Sao Paulo",country:"BRA"}})
+{address:{city:"Sao Paulo",country:"BRA",street:"Avenida Paulista"},name:"John Doe"}
+
+>> Patch({name:"John Doe",address:{street:"One Main St",city:"Springfield",country:"USA"}},{address:{country:"BRA"}})
```
CarlosFigueiraMSFT, line 56: >Patch({name:"John Doe",address:{street:"One Main St",city:"Springfield",country:"USA"}},{address:{country:"BRA"}}) [](http://example.com/codeflow?start=3&length=114)

This is interesting - the record type is compatible, but it doesn't let us change it. We have the same error in Power Apps, so this is not worse than today, but can you start a thread to see what we should do about this? It shouldn't block this PR, though.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch_PatchRecord.txt
```diff
@@ -0,0 +1,66 @@
...
+Errors: Error 88-113: The type of this argument 'address' does not match the expected type 'Record'. Found type 'Record'.|Error 0-114: The function 'Patch' has some invalid arguments.
+
+>> Patch({a:Table({a:1,b:1}),b:Blank()},{b:1})
```
CarlosFigueiraMSFT, line 59: >Patch({a:Table({a:1,b:1}),b:Blank()},{b:1}) [](http://example.com/codeflow?start=3&length=43)

Same issue as the previous test. This should work (but doesn't today)

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
```diff
@@ -136,6 +137,18 @@ private static ErrorValue NotImplemented(IRContext irContext, [CallerMemberName]
...
+        }
+
+        /// <summary>
+        /// Used to append a record to a table when baseRecord contains the primary key and updateRecord contains the fields to update.
+        /// </summary>
+        /// <param name="baseRecord"></param>
+        /// <param name="updateRecord"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public virtual async Task<DValue<RecordValue>> AppendAsync(RecordValue baseRecord, RecordValue updateRecord, CancellationToken cancellationToken)
```
CarlosFigueiraMSFT, line 149: >AppendAsync [](http://example.com/codeflow?start=55&length=11)

We discussed offline - this is probably not needed

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Patch_PatchRecord.txt
anderson-joyle: Issue [https://github.com/microsoft/Power-Fx/issues/2288](https://github.com/microsoft/Power-Fx/issues/2288)

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
anderson-joyle: Removed.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (dismissed):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #2271, opened 2024-03-24
Making PropertyRuleCategory and Extensions public

petrochuk:
Making PropertyRuleCategory and Extensions public so they can be used outside of PowerFx.Core


src/libraries/Microsoft.PowerFx.Core/App/Controls/PropertyRuleCategory.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.PowerFx.Core.App.Controls
     /// Please keep these values in sync with src/AppMagic/js/AppMagic.Controls/Constants.ts.
     /// </summary>
     [TransportType(TransportKind.Enum)]
-    internal enum PropertyRuleCategory
+    public enum PropertyRuleCategory
```
MikeStall, line 14: let me followup with you about why.  Making [TransportType] types public is dangerous.

PR #2283, opened 2024-03-28
Strongly typed enums for function parameters

gregli-msft:
Adds additional flags for semantic control of individual option sets.  See the comments at the top of IExternalOptionSet.cs and the comments and test cases in StronglyTypedEnum_*.txt files.

Note that this does not impact Enums proper (as opposed to Option Sets), at this point only used by Canvas apps.


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Match_StronglyTypedEnums_StronglyTypedEnumsDisabled.txt
```diff
@@ -0,0 +1,98 @@
+﻿#SETUP: RegEx
+
```
CarlosFigueiraMSFT, line 2: This file is named "Match_StronglyTypedEnums_StronglyTypedEnumsDisabled.txt", but it doesn't need any of the StronglyTypedEnum logic. Can this be merged into the "base" Match.txt?

src/tests/Microsoft.PowerFx.Core.Tests/TypeSystemTests/DTypeTests.cs
```diff
@@ -62,6 +63,14 @@ internal class BoolOptionSetInfo : IExternalOptionSet
 
             public DKind BackingKind => DKind.Boolean;
 
+            public bool CanCoerceFromBackingKind => false;
+
+            public bool CanCoerceToBackingKind => false;
```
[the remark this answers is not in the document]
CarlosFigueiraMSFT: Ah, this is test code, nevermind. It may still be good to make it closer to reality, though.

src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalOptionSet.cs
```diff
@@ -20,8 +20,77 @@ internal interface IExternalOptionSet : IExternalEntity
...
+
+        /// <summary>
+        /// Option set can be used in place of the backing kind. 
+        /// This is useful to provide a convenient enum value for the backing kind, for example ErrorKind and Color.
+        /// There is no type safety, effectively the enum is just like using the backing type.
+        /// </summary>
+        bool CanCoerceToBackingKind { get; }
+
+        /// <summary>
+        /// Only applies to string backed option sets.
```
CarlosFigueiraMSFT, line 79: >Only applies to string backed option sets. [](http://example.com/codeflow?start=12&length=42)

We should have checks that enforce that (possibly when adding a new option set to the symbol table)

src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalOptionSet.cs
```diff
@@ -20,8 +20,77 @@ internal interface IExternalOptionSet : IExternalEntity
...
+        //          CanCoerceFromBackindKind and CanContenateStronglyTyped = true.
+        //          Functions that have this option set as a parameter type can be passed the backing kind and
+        //          concatenations between the option set and the backind kind result in the option set type.
+        //          Used by Match.
+
+        /// <summary>
+        /// Backing kind typed values can be used in placed of this option set value.  
+        /// This is also possible for concatenate and numeric compare if those flags are set.
+        /// Examples: a completely custom regular expression string can passed for a Match enum
+        /// and a number can be used in palce of an ErrorKind.
```
CarlosFigueiraMSFT, line 67: >palce [](http://example.com/codeflow?start=40&length=5)

[typo] place

src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalOptionSet.cs
```diff
@@ -20,8 +20,77 @@ internal interface IExternalOptionSet : IExternalEntity
...
+        /// <summary>
+        /// Only applies to string backed option sets.
+        /// All enums can be coerced to strings and concatenated as strings.  In some situations, we intend the maker
+        /// to concatenate enums and not lose the strong typing of that enum, for example with JSONFormat or MatchOptions.
+        /// A concatenate with like enum values results in the enum type being preserved if this flag is true.
+        /// </summary>
+        bool CanConcatenateStronglyTyped { get; }
+
+        /// <summary>
+        /// Only applies to number backed option sets.
```
CarlosFigueiraMSFT, line 87: >Only applies to number backed option sets. [](http://example.com/codeflow?start=12&length=42)

We should have checks that enforce that (possibly when adding a new option set to the symbol table)

src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalOptionSet.cs
```diff
@@ -20,8 +20,77 @@ internal interface IExternalOptionSet : IExternalEntity
         /// </summary>
         IEnumerable<DName> OptionNames { get; }
 
+        /// <summary>
+        /// Backing kind for this option set. Must be Number, String, Boolean, or Color.
```
CarlosFigueiraMSFT, line 24: >Must be Number, String, Boolean, or Color. [](http://example.com/codeflow?start=46&length=42)

We should have checks that enforce that (possibly when adding a new option set to the symbol table)

src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalOptionSet.cs
gregli-msft: Fixed.

src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalOptionSet.cs
gregli-msft: I don't know of a way to enforce this at the interface level.  Having a seperate OptionSetDKind is possible, but gets messy considering DKind is a part of DType.  This is enforced in EnumSymbol through a Contracts.Assert.  OptionSet.cs only supports creating String backed option sets.

src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalOptionSet.cs
gregli-msft: I don't know of a way to enforce this at the interface level.  Having a seperate OptionSetDKind is possible, but gets messy considering DKind is a part of DType.  This is enforced in EnumSymbol through a Contracts.Assert.  OptionSet.cs only supports creating String backed option sets.

src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalOptionSet.cs
gregli-msft: I don't know of a way to enforce this at the interface level.  Having a seperate OptionSetDKind is possible, but gets messy considering DKind is a part of DType.  This is enforced in EnumSymbol through a Contracts.Assert.  OptionSet.cs only supports creating String backed option sets.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Match_StronglyTypedEnums_StronglyTypedEnumsDisabled.txt
gregli-msft: Cleaned up.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/TypeSystemTests/DTypeTests.cs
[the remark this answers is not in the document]
gregli-msft: Fixed.  I made Boolean option sets work like the rest - they need CanCoerceToBackingKind to be true for normal operations.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -82,9 +85,218 @@ private static (ISymbolSlot slot, FormulaValue value) AddBlankVar(SymbolTable sy
...
-            return PowerFxConfig.BuildWithEnumStore(new EnumStoreBuilder().WithDefaultEnums(), new TexlFunctionSet(), config.Features);
+            var store = new EnumStoreBuilder().WithDefaultEnums();
+            var newConfig = PowerFxConfig.BuildWithEnumStore(store, new TexlFunctionSet(), config.Features);
+
+            // There are likewise no built in functions that take Boolean backed option sets as parameters
+            newConfig.AddFunction(new TestXORBooleanFunction());
+            newConfig.AddFunction(config.Features.StronglyTypedBuiltinEnums ? new STE_TestXORYesNoFunction() : new Boolean_TestXORYesNoFunction());
+            newConfig.AddFunction(new TestColorInvertFunction());
+            newConfig.AddFunction(config.Features.StronglyTypedBuiltinEnums ? new STE_TestColorBlueRampInvertFunction() : new Color_TestColorBlueRampInvertFunction());
+
```
CarlosFigueiraMSFT, line 96: Ideally those should be in a different setup. Power Apps can run tests with the default built-in enums. It wont be able to run the test with those new functions.

CarlosFigueiraMSFT:
> #SETUP: StronglyTypedBuiltinEnums,PowerFxV1CompatibilityRules,AllEnumsPlusTestEnumsSetup,RegEx,EnableJsonFunctions

Can we split this test in two, one that only uses built-in enums, and one that uses the extra ones? The former can be run in Power Apps, the latter can't.

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/StronglyTypedEnumParams.txt:1 in 97bb477. [](commit_id = 97bb477b2ba9e089e3a2316bf01f932c6b95db31, deletion_comment = False)

CarlosFigueiraMSFT:
> #SETUP: disable:StronglyTypedBuiltinEnums,disable:PowerFxV1CompatibilityRules,AllEnumsPlusTestEnumsSetup,RegEx,EnableJsonFunctions

Can we split this test in two, one that only uses built-in enums, and one that uses the extra ones? The former can be run in Power Apps, the latter can't.

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/StronglyTypedEnumParams_STEDisabled.txt:1 in 97bb477. [](commit_id = 97bb477b2ba9e089e3a2316bf01f932c6b95db31, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests/TypeSystemTests/DTypeTests.cs
```diff
@@ -6,6 +6,7 @@
 using System.Collections.Immutable;
 using System.Linq;
 using System.Text;
+using System.Threading;
 using Microsoft.PowerFx.Core;
```
CarlosFigueiraMSFT, line 10: Probably not needed

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
gregli-msft: I split the tests into two files, one that uses the old AllEnumsSetup and one that uses the new one with the test enums that covers Boolean and Color enums.  Should be good to go for Power Apps.

src/tests/Microsoft.PowerFx.Core.Tests/TypeSystemTests/DTypeTests.cs
gregli-msft: Done.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -340,23 +340,39 @@ private static BinderCheckTypeResult CheckTypeCore(IErrorContainer errorContaine
...
+            // Option sets need not (and shouldn't be) listed in calls to CheckTypeCore.  For consistency,
+            // we always check option sets here against the backing types.
+            //
+            // Option sets can always coerce to string, V1 and pre-V1.
+            var nodeBackingKind = nodeType.OptionSetInfo?.BackingKind;
+            if (nodeBackingKind != null &&
+                (typeWant.Kind == DKind.String ||
+                ((!usePowerFxV1CompatibilityRules || nodeType.OptionSetInfo.CanCoerceToBackingKind) &&
+                   ((typeWant.Kind == DKind.Boolean && nodeBackingKind == DKind.Boolean) ||
+                    ((typeWant.Kind == DKind.Number || typeWant.Kind == DKind.Decimal) && nodeBackingKind == DKind.Number) ||
```
CarlosFigueiraMSFT, line 360: >DKind.Number [](http://example.com/codeflow?start=109&length=12)

I'm assuming we only have backing kinds of String/Boolean/Number/Color. Do we have checks somewhere to enforce that?

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -783,77 +799,71 @@ private static BinderCheckTypeResult CheckComparisonArgTypesCore(IErrorContainer
...
-                    errorContainer.EnsureError(
-                        DocumentErrorSeverity.Severe,
-                        left.Parent,
-                        TexlStrings.ErrUnOrderedTypeForComparison_Type,
-                        typeLeft.GetKindString());
-
-                    return new BinderCheckTypeResult();
-                }
-                else if ((typeRight.OptionSetInfo?.BackingKind ?? DKind.Number) != DKind.Number)
+                // Comparing values from two different option set values is not supported, with or without V1 compat rules
```
CarlosFigueiraMSFT, line 805: >// Comparing values from two different option set values is not supported, with or without V1 compat rules [](http://example.com/codeflow?start=16&length=106)

Today you can compare two different enums (e.g., `ImageRotation.None = GridStyle.None`, `JSONFormat.IgnoreUnsupportedTypes = TraceOptions.IgnoreUnsupportedTypes`). Will this break?

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -1005,45 +1015,61 @@ private static BinderCheckTypeResult CheckEqualArgTypesCore(IErrorContainer erro
                 return new BinderCheckTypeResult();
             }
 
-            // Special case for comparing option set values, it should produce an error when the base option sets are different
-            if (typeLeft.Kind == DKind.OptionSetValue && typeRight.Kind == DKind.OptionSetValue && !typeLeft.Accepts(typeRight, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: usePowerFxV1CompatibilityRules))
+            // Special case for comparing option set values
+            if (typeLeft.Kind == DKind.OptionSetValue || typeRight.Kind == DKind.OptionSetValue)
             {
+                // Comparing values from two different option set values is not supported, with or without V1 compat rules
```
CarlosFigueiraMSFT, line 1021: >// Comparing values from two different option set values is not supported, with or without V1 compat rules [](http://example.com/codeflow?start=16&length=106)

Today you can compare two different enums (e.g., `ImageRotation.None = GridStyle.None`, `JSONFormat.IgnoreUnsupportedTypes = TraceOptions.IgnoreUnsupportedTypes`). Will this break?

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -1194,7 +1220,7 @@ internal static BinderCheckTypeResult CheckUnaryOpCore(IErrorContainer errorCont
             switch (node.Op)
             {
                 case UnaryOp.Not:
-                    var resNot = CheckTypeCore(errorContainer, node.Child, features, childType, DType.Boolean, /* coerced: */ DType.Number, DType.Decimal, DType.String, DType.OptionSetValue, DType.UntypedObject);
```
CarlosFigueiraMSFT, line 1197: >DType.OptionSetValue [](http://example.com/codeflow?start=169&length=20)

Can't we do something like `Not MyBoolOptionSet.NoValue`?

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -1265,22 +1291,47 @@ public static BinderCheckTypeResult CheckBinaryOpCore(IErrorContainer errorConta
...
+                        if (leftType == DType.OptionSetValue && rightType == DType.OptionSetValue)
+                        {
+                            if (rightType.OptionSetInfo.EntityName == leftType.OptionSetInfo.EntityName &&
+                                rightType.OptionSetInfo.CanConcatenateStronglyTyped)
+                            {
+                                return new BinderCheckTypeResult() { Node = node, NodeType = leftType };
+                            }
+                        }
+                        else if (leftType == DType.OptionSetValue &&
+                                 leftType.OptionSetInfo.CanCoerceFromBackingKind && leftType.OptionSetInfo.CanConcatenateStronglyTyped)
```
CarlosFigueiraMSFT, line 1313: Would that allow something like `JSONFormat.IndentFour & ErrorKind.Div0`? I don't know if we should enable that

CarlosFigueiraMSFT:
>         public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DType[] argTypes, IErrorContainer errors, out DType returnType, out Dictionary<TexlNode, DType> nodeToCoercedTypeMap)

I see we have some tests for Concatenate with string-backed enums (option sets). Can we add tests with enums backed by a different type (e.g., ErrorKind.Div0)

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Concatenate.cs:49 in 0932cfa. [](commit_id = 0932cfac9d2735b87705db1b4e3d6749a0b06861, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Types/Enums/BuiltInEnums.cs
```diff
@@ -48,7 +49,8 @@ internal static class BuiltInEnums
                 { "LongDateTime24", "'longdatetime24'" },
                 { "ShortDateTime24", "'shortdatetime24'" },
                 { "UTC", "utc" }
-            });
+            },
+            canCoerceFromBackingKind: true);
```
CarlosFigueiraMSFT, line 53: >canCoerceFromBackingKind: true [](http://example.com/codeflow?start=12&length=30)

If I understand correctly, that will allow us to create this enum from the string such as "utc" or "'shortdate'". Is that the case?

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
gregli-msft: There is a Contracts.Assert at the top of EnumSymbol.cs, line 60. I assume Contracts.Assert is strong enough?  EnumStoreBuilder only works with EnumSymbols. What one would really like is a restriction at the IExternalOptionSet level, but I don't know how to do that with an interface definition. The way this logic is written in BinderUtils, it wouldn't match for another type, so more restrictive if an unexpected type were to enter.

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
gregli-msft: It does.  I started off with not allowing this, but it gets odd fast as all option sets can coerce to string.  Take for example string interpolation, that is translated into concatenate:

```
>> IR( $"a {JSONFormat.IndentFour} {ErrorKind.Div0}" )
Concatenate:s("a ":s, FieldAccess(ResolvedObject(Microsoft.PowerFx.Core.Types.Enums.EnumSymbol), IndentFour), " ":s, FieldAccess(ResolvedObject(Microsoft.PowerFx.Core.Types.Enums.EnumSymbol), Div0))
```

What I settled on is that we would only concatenate strongly typed within the same enum.  So, yes you can do the above, but you can only use it in a text context:

```
>> JSON( 1, JSONFormat.IndentFour & MatchOptions.IgnoreCase )
Error 31-32: Invalid argument type (Text). Expecting a Enum (JSONFormat) value instead.
Error 31-32: The value 'i' is not supported as an option in the JSON function.
Error 0-4: The function 'JSON' has some invalid arguments.
```

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
gregli-msft: I believe you are asking about https://github.com/microsoft/Power-Fx/blob/0932cfac9d2735b87705db1b4e3d6749a0b06861/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/StronglyTypedEnum_TestEnums.txt#L548 and tests near this one.  Instead of listing OptionSetValue as a coercable type, CheckTypeCore will coerce any option set with a matching backing type.  This simplifies the logic in that function, prevents indirect coercions (string backed option set to Boolean), and makes for a better error message for the maker.

gregli-msft:
> > ```
> >     public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DType[] argTypes, IErrorContainer errors, out DType returnType, out Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
> > ```
> 
> I see we have some tests for Concatenate with string-backed enums (option sets). Can we add tests with enums backed by a different type (e.g., ErrorKind.Div0)
> 
> Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Concatenate.cs:49 in [0932cfa](https://github.com/microsoft/Power-Fx/commit/0932cfac9d2735b87705db1b4e3d6749a0b06861). [](commit_id = [0932cfa](https://github.com/microsoft/Power-Fx/commit/0932cfac9d2735b87705db1b4e3d6749a0b06861), deletion_comment = False)

Yes, for example: https://github.com/microsoft/Power-Fx/blob/0932cfac9d2735b87705db1b4e3d6749a0b06861/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/StronglyTypedEnum_BuiltInEnums.txt#L1098.  As all option sets coerce to string, it is always allowed, the question is what the result will be.

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
gregli-msft: That comment was a tad overzealous.  Without StronglyTypedBuiltinEnums, option sets are converted to their backing kinds before getting this far.  I update the comments.

Moving forward, yes we are already blocking this in main with the original Strongly Typed implementation, which I think is correct.  There are lots of tests added to check this, 
https://github.com/microsoft/Power-Fx/blob/0932cfac9d2735b87705db1b4e3d6749a0b06861/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/StronglyTypedEnum_BuiltInEnums.txt#L219
as well as checking that we don't break this for Canvas
https://github.com/microsoft/Power-Fx/blob/0932cfac9d2735b87705db1b4e3d6749a0b06861/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/StronglyTypedEnum_BuiltInEnums_PreV1.txt#L221

src/libraries/Microsoft.PowerFx.Core/Types/Enums/BuiltInEnums.cs
gregli-msft: Kind of.  From a type system perspective, this accurately describing the semantics for the second argument to Text (that it should be this enum and not another) and that we do want strings to be convertible to DateTimeFormat.CalculatedOptionSetValue for custom formats.  An implementation can check to see if they get back that enum value (or some other bit that says it was calculated) and treat it like a string instead of the enum value.

But in the current implementation, it already checks for enum values separaete from strings, before that coercion would happen.  It probably could be rewritten to use this mechanism and be less special case.

Thinking about IntelliSense, this might be important, but taking that even further we should probably have a bit for "MustBeConstant" too, both for this and for the Match enum, so that we don't recommend variables that are string valued.  I'm not sure how IntelliSense works and if it uses CheckTypes, perhaps this is already covered.

In any case, I have added some tests.  For Pre-V1, the interpreter doesn't do the right thing, hence the override, but the interpreter shouldn't be used in Pre-V1 mode anyway.
https://github.com/microsoft/Power-Fx/blob/5aff062b5fc04ceebab980578fb315f080bd3d31/src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/StronglyTypedEnum_BuiltInEnums.txt#L1192

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
🚢

LucGenetier:
✅ No public API change.

PR #2284, opened 2024-03-28
Change Error Span for "Function X has some invalid arguments"

lesaltzm:
Moves the error span for "some invalid arguments" to just the head of the call node, so it doesn't hide/distract from other more specific errors in the formula bar


LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

lesaltzm:
Do not merge, pending test in PA Client

LucGenetier:
✅ No public API change.

PR #2286, opened 2024-03-29
Transform Text function exception into runtime error

gregli-msft:
If the text formatter used in the Text function ran into a format it did not understand, for example a simple "i", then the code would throw and the exception would not be caught.

The diff is broken up and looks more involved then it is, all that was done is to add a try/catch with the appropriate addition of tabs for the extra block.


src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Text.txt
```diff
@@ -266,4 +266,8 @@ Error({Kind:ErrorKind.BadLanguageCode})
 "domingo enero 31"
 
 >> Text(DateTimeValue("12:34:56"),"yyyy-mm-dd hh:mm:ss")
-"1899-12-30 12:34:56"
\ No newline at end of file
+"1899-12-30 12:34:56"
+
+// This would previously throw an exception in the interpreter by the C# text formatter on the "i"
+>> Text( 12, "i" )
```
CarlosFigueiraMSFT, line 272: Should this be an error at all?

image

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2292, opened 2024-03-29
Only run CheckSemantics if CheckTypes passes

gregli-msft:
We are producing some additional errors that we don't need, that either don't add any value or can be confusing.

For example
```
>> ClearCollect(Float(1)/0,Foo)
Errors: Error 24-27: Name isn't valid. 'Foo' isn't recognized.
|Error 21-22: Invalid argument type (Number). Expecting a Table value instead.
|Error 21-22: The value passed to the 'ClearCollect' function cannot be changed.
|Error 0-12: The function 'ClearCollect' has some invalid arguments.
```
The third error about the value cannot be changed, although technically correct, is not adding any value as there are much bigger issues here.  It comes from CheckSemantics, which doesn't need to be run if we already have errors from CheckTypes.


LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):

LucGenetier:
✅ No public API change.

Shpakh-MSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):

LucGenetier:
✅ No public API change.

PR #2294, opened 2024-03-29
Supporting curly braces when declaring UDFs

rick-nguyen:
Initial PR to support imperative logic and chained expressions in UDFs.


src/tests/Microsoft.PowerFx.Interpreter.Tests/UserDefinedTests.cs
```diff
@@ -24,5 +25,16 @@ public void NamedFormulaEntryTest(string script, string expression, double expec
...
+        [InlineData("test2(b: Boolean): Boolean = { Set(a, b); };")]
+        [InlineData("test2(b: Boolean): Boolean = { Set(a, b); Collect(abc, { bcd: 1 }) };")]
+        [InlineData("test2(b: Boolean): Boolean = { Set(a, b); Collect(abc, { bcd: 1 }) }; num = 3;")]
+        public void UserDefinedFunctions(string script)
+        {
+            var result = UserDefinitions.Process(script, null);
+
+            Assert.False(result.HasErrors);
+        }
     }
```
CarlosFigueiraMSFT, line 39: >} [](http://example.com/codeflow?start=4&length=1)

We should have negative tests (i.e., unclosed curly braces), tests with errors inside braces (we should still parse other definitions)

src/tests/Microsoft.PowerFx.Interpreter.Tests/UserDefinedTests.cs
```diff
@@ -24,5 +25,16 @@ public void NamedFormulaEntryTest(string script, string expression, double expec
...
             Assert.Equal(expected, result.Value);
         }
+
+        [Theory]
+        [InlineData("test2(b: Boolean): Boolean = { Set(a, b); };")]
+        [InlineData("test2(b: Boolean): Boolean = { Set(a, b); Collect(abc, { bcd: 1 }) };")]
+        [InlineData("test2(b: Boolean): Boolean = { Set(a, b); Collect(abc, { bcd: 1 }) }; num = 3;")]
+        public void UserDefinedFunctions(string script)
+        {
+            var result = UserDefinitions.Process(script, null);
```
CarlosFigueiraMSFT, line 35: >UserDefinitions.Process(script, null); [](http://example.com/codeflow?start=25&length=38)

How do we differentiate UDFs that have side effects versus those that don't? For example, in Power Apps we cannot use the former in a data property, but we can use the latter

Shpakh-MSFT: I assume that registered UDF will have AllowSideAffect property set to true. But that doesn't exist today and needs to be added.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT:
We should have tests that actually execute a UDF with side effects

src/tests/Microsoft.PowerFx.Interpreter.Tests/UserDefinedTests.cs
```diff
@@ -24,5 +27,52 @@ public void NamedFormulaEntryTest(string script, string expression, double expec
...
+        {
+            var options = new ParserOptions()
+            {
+                AllowsSideEffects = true,
+                Culture = CultureInfo.InvariantCulture
+            };
+            UserDefinitions.ProcessUserDefinitions(script, options, out var result);
+
+            Assert.False(result.HasErrors);
+        }
```
CarlosFigueiraMSFT, line 45: More interesting tests:

- F1(x:Number): Number = { Set(a, x); x+1; };
  - behavior UDF with a return value
- F1(x:Number): Boolean = { Set(a, x); Today(); };
  - That should be an error - we cannot coerce a date to a Boolean value

Shpakh-MSFT (dismissed):
:shipit:

src/tests/Microsoft.PowerFx.Interpreter.Tests/UserDefinedTests.cs
```diff
@@ -24,5 +27,52 @@ public void NamedFormulaEntryTest(string script, string expression, double expec
...
+        [InlineData("test2(b: Boolean): Boolean = { Set(a, b); };")]
+        [InlineData("test2(b: Boolean): Boolean = { Set(a, b); Collect(abc, { bcd: 1 }) };")]
+        [InlineData("test2(b: Boolean): Boolean = { Set(a, b); Collect(abc, { bcd: 1 }) }; num = 3;")]
+        public void ValidUDFBodyTest(string script)
+        {
+            var options = new ParserOptions()
+            {
+                AllowsSideEffects = true,
+                Culture = CultureInfo.InvariantCulture
+            };
```
Shpakh-MSFT, line 41: Can you add a negative testcase where AllowSideEffects is false?

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -593,6 +593,49 @@ public void FunctionPrecedenceTest(string script, double expected)
...
+
+        // Mismatch return value with side effectful UDF
+        [InlineData(
+            "F1(x:Number) : Boolean = { Set(a, x); Today(); };",
+            null,
+            true,
+            0,
+            true)]
+
+        public void ImperativeUserDefinedFunctionTest(string udfExpression, string expression, bool expectedError, double expected = 0, bool allowSideEffects = false)
```
CarlosFigueiraMSFT, line 614: > = 0 [](http://example.com/codeflow?start=130&length=4)

No need to define the default values for these arguments as they are passed in all cases

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -593,6 +593,49 @@ public void FunctionPrecedenceTest(string script, double expected)
...
+        [InlineData(
+            "F1(x:Number) : Number = { Set(a, x); a+1; };",
+            "F1(123)",
+            false,
+            124,
+            true)]
+
+        // Mismatch return value with side effectful UDF
+        [InlineData(
+            "F1(x:Number) : Boolean = { Set(a, x); Today(); };",
```
CarlosFigueiraMSFT, line 608: >"F1(x:Number) : Boolean = { Set(a, x); Today(); }; [](http://example.com/codeflow?start=12&length=50)

Where do you expect this to throw? During AddUserDefinedFunctio, or during Eval? It would be good if the test could tell us that.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -593,6 +593,49 @@ public void FunctionPrecedenceTest(string script, double expected)
...
+            true,
+            0,
+            true)]
+
+        public void ImperativeUserDefinedFunctionTest(string udfExpression, string expression, bool expectedError, double expected = 0, bool allowSideEffects = false)
+        {
+            var config = new PowerFxConfig()
+            {
+                MaxCallDepth = 100
+            };
```
CarlosFigueiraMSFT, line 619: We don't need this

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -593,6 +593,49 @@ public void FunctionPrecedenceTest(string script, double expected)
...
+            recalcEngine.UpdateVariable("a", 1m);
+
+            try
+            {
+                recalcEngine.AddUserDefinedFunction(udfExpression, CultureInfo.InvariantCulture, symbolTable: recalcEngine.EngineSymbols, allowSideEffects: allowSideEffects);
+
+                var result = recalcEngine.Eval(expression, options: _opts);
+                var fvExpected = FormulaValue.New(expected);
+
+                Assert.Equal(fvExpected.AsDecimal(), result.AsDecimal());
```
CarlosFigueiraMSFT, line 631: >Equal [](http://example.com/codeflow?start=23&length=5)

Above or below this line, Assert.False(expectedError)

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -43,6 +43,8 @@ internal class UserDefinedFunction : TexlFunction
 
         public override bool IsSelfContained => !_isImperative;
 
+        public IEnumerable<UDFArg> Args => _args;
```
CarlosFigueiraMSFT, line 46: >Args [](http://example.com/codeflow?start=35&length=4)

Do we need this? I couldn't find a reference to it

rick-nguyen: Oh yeah you're right. Left in by mistake from a stale local change

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
rick-nguyen: It throws during AddUserDefinedFunction, `CheckTypesOnDeclaration`

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
🚢

PR #2297, opened 2024-04-01
Ensure AssociatedDataSources is nonnull even for DType.Invalid

lesaltzm:
Fixes a longstanding intermmitent Arg Null Exception we've seen, when DType.Invalid leaks into the system.


jas-valgotar (approved):

LucGenetier:
✅ No public API change.

PR #2299, opened 2024-04-01
New exception to support DV not connected

anderson-joyle:
To support https://github.com/microsoft/Power-Fx-Dataverse/issues/455.


src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs
```diff
@@ -322,6 +323,10 @@ public async Task<FormulaValue> InvokeAsync(IServiceProvider serviceProvider, Fo
                 {
                     return CommonErrors.CustomError(IRContext.NotInSource(info.RetType), customFunctionErrorException.Message);
                 }
+                catch (ServiceUnavailableException serviceUnavailableException)
+                {
+                    return FormulaValue.NewError(serviceUnavailableException.ExpressionError);
+                }
```
MikeStall, line 329: can we avoid new exception type and reuse CustomFunctionErrorException with ErrorValue?

LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #2302, opened 2024-04-02
Json fix when server returns a malformed Json payload

LucGenetier:


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Connectors.Tests/PowerPlatformConnectorTests.cs
```diff
@@ -1422,6 +1423,65 @@ public async Task SQL_ExecuteStoredProc_WithUserAgent()
...
+   at System.Text.Json.Utf8JsonReader.ConsumeValue(Byte marker)
+   at System.Text.Json.Utf8JsonReader.ReadFirstToken(Byte first)
+   at System.Text.Json.Utf8JsonReader.ReadSingleSegment()
+   at System.Text.Json.Utf8JsonReader.Read()
+   at System.Text.Json.JsonDocument.Parse(ReadOnlySpan`1 utf8JsonSpan, JsonReaderOptions readerOptions, MetadataDb& database, StackRowStack& stack)
+   at System.Text.Json.JsonDocument.Parse(ReadOnlyMemory`1 utf8Json, JsonReaderOptions readerOptions, Byte[] extraRentedArrayPoolBytes, PooledByteBufferWriter extraPooledByteBufferWriter)
+   at System.Text.Json.JsonDocument.Parse(ReadOnlyMemory`1 json, JsonDocumentOptions options)
+   at System.Text.Json.JsonDocument.Parse(String json, JsonDocumentOptions options)
+   at Microsoft.PowerFx.Types.FormulaValueJSON.FromJson(String jsonString, FormulaValueJsonSerializerSettings settings, FormulaType formulaType) in ...\Microsoft.PowerFx.Json\FormulaValueJSON.cs:line 36", new Regex(@" in .*\\libraries\\").Replace(message, @" in ...\"));
```
MikeStall, line 1480: This is too brittle. 
Callstack and line numbers can change with any update of STJ.

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -32,10 +31,22 @@ public static FormulaValue FromJson(string jsonString, FormulaType formulaType =
...
+                using JsonDocument document = JsonDocument.Parse(jsonString);
+                JsonElement propBag = document.RootElement;
+
+                return FromJson(propBag, settings, formulaType);
+            }
+            catch (JsonException je)
+            {                
+                return new ErrorValue(IRContext.NotInSource(formulaType), new ExpressionError()
+                {
+                    Message = $"{je.GetType().Name} {je.Message} {je.StackTrace}",
```
MikeStall, line 45: >{je.StackTrace}", [](http://example.com/codeflow?start=65&length=17)

This probably shouldn't be part of the error message. 
ErrorValues get returned to the user. Whereas a C# callstack is not meaningful to an end-user.

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -32,10 +31,22 @@ public static FormulaValue FromJson(string jsonString, FormulaType formulaType =
...
-            using JsonDocument document = JsonDocument.Parse(jsonString);
-            JsonElement propBag = document.RootElement;
-
-            return FromJson(propBag, settings, formulaType);
+            try
+            {
+                using JsonDocument document = JsonDocument.Parse(jsonString);
+                JsonElement propBag = document.RootElement;
+
+                return FromJson(propBag, settings, formulaType);
```
MikeStall, line 39: >FromJson [](http://example.com/codeflow?start=23&length=8)

From caller's perspective, there are 4 overloads of FromJson(...) ... which ones catch? 

They should probably all behave consistently (all catch or none catch).

LucGenetier: Only the one that uses a string
Others use JsonElement

MikeStall: per-offline - once we have a JsonElement, we don't get parser errors.

MikeStall (approved):

PR #2303, opened 2024-04-02
Adding semantic check to Patch V1

anderson-joyle:
An extra semantic check missing in https://github.com/microsoft/Power-Fx/pull/2269.


LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #2306, opened 2024-04-04
Tabular connectors - initial work

LucGenetier:


LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.ConfigExtensions.AddActionConnector(Microsoft.PowerFx.PowerFxConfig,Microsoft.PowerFx.Connectors.ConnectorSettings,Microsoft.OpenApi.Models.OpenApiDocument,Microsoft.PowerFx.Connectors.ConnectorLogger)<br />❌CP0002 M:Microsoft.PowerFx.ConfigExtensions.AddActionConnector(Microsoft.PowerFx.PowerFxConfig,System.String,Microsoft.OpenApi.Models.OpenApiDocument,Microsoft.PowerFx.Connectors.ConnectorLogger)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.OpenApiParser.GetFunctions(Microsoft.PowerFx.Connectors.ConnectorSettings,Microsoft.OpenApi.Models.OpenApiDocument,Microsoft.PowerFx.Connectors.ConnectorLogger)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.OpenApiParser.GetFunctions(System.String,Microsoft.OpenApi.Models.OpenApiDocument,Microsoft.PowerFx.Connectors.ConnectorLogger)

src/libraries/Microsoft.PowerFx.Connectors/Environment/PowerFxConfigExtensions.cs
```diff
@@ -90,5 +96,37 @@ public static IReadOnlyList<ConnectorFunction> AddActionConnector(this PowerFxCo
                 throw;
             }            
         }
+
+        public static async Task<ConnectorTableValue> AddTabularConnector(this PowerFxConfig config, string tableName, OpenApiDocument openApiDocument, IReadOnlyDictionary<string, FormulaValue> globalValues, HttpClient client, CancellationToken cancellationToken, ConnectorLogger configurationLogger = null)
```
MikeStall, line 100: >AddTabularConnector [](http://example.com/codeflow?start=54&length=19)

These are key entry points.

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorTableValueWithServiceProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Connectors
+{
+    public class ConnectorTableValueWithServiceProvider : ConnectorTableValue
```
MikeStall, line 14: we should followup on this

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.ConfigExtensions.AddActionConnector(Microsoft.PowerFx.PowerFxConfig,Microsoft.PowerFx.Connectors.ConnectorSettings,Microsoft.OpenApi.Models.OpenApiDocument,Microsoft.PowerFx.Connectors.ConnectorLogger)<br />❌CP0002 M:Microsoft.PowerFx.ConfigExtensions.AddActionConnector(Microsoft.PowerFx.PowerFxConfig,System.String,Microsoft.OpenApi.Models.OpenApiDocument,Microsoft.PowerFx.Connectors.ConnectorLogger)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.OpenApiParser.GetFunctions(Microsoft.PowerFx.Connectors.ConnectorSettings,Microsoft.OpenApi.Models.OpenApiDocument,Microsoft.PowerFx.Connectors.ConnectorLogger)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.OpenApiParser.GetFunctions(System.String,Microsoft.OpenApi.Models.OpenApiDocument,Microsoft.PowerFx.Connectors.ConnectorLogger)

src/libraries/Microsoft.PowerFx.Connectors/Internal/InjectServiceProviderFunction.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        public Task<FormulaValue> InvokeAsync(FormulaValue[] args, IServiceProvider context, CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            // Creation of the table with its context
+            ConnectorTableValueWithServiceProvider connectorTableWithServiceProvider = new ConnectorTableValueWithServiceProvider(args[0] as ConnectorTableValue, context);
+            return Task.FromResult<FormulaValue>(connectorTableWithServiceProvider);
+        }
```
MikeStall, line 41: Could we remove ConnectorTableValueWithServiceProvider altogether and just do:
(This is clearer about lifespans, and would also ensure the cancellation token is passed, which it is not today)

        public async Task<FormulaValue> InvokeAsync(FormulaValue[] args, IServiceProvider context, CancellationToken cancellationToken)
        {
            cancellationToken.ThrowIfCancellationRequested();

            var runtimeContext = context.GetService<BaseRuntimeConnectorContext>();

            // Creation of the table with its context
            ConnectorTableValueWithServiceProvider connectorTableWithServiceProvider = new ConnectorTableValueWithServiceProvider(args[0] as ConnectorTableValue, context);

            ConnectorTableValue table = (ConnectorTableValue)args[0]; // IRRewriter should gaurantee this is always true
            var tabularService = table._tabularService;
            var rows = await tabularService.GetItemsAsync(runtimeContext, cancellationToken).ConfigureAwait(false);

            // Create in-memory 
            var inMemTable = new InMemoryTableValue(table.IRContext, rows);

            return inMemTable;
        }

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorTableValueWithServiceProvider.cs
MikeStall: see suggestion in InjectServiceProviderFunction.
We could simplify and remove this class.

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorTableValue.cs
```diff
@@ -0,0 +1,59 @@
...
+            : base(irContext)
+        {
+        }
+
+        public override IEnumerable<DValue<RecordValue>> Rows => GetRowsInternal().ConfigureAwait(false).GetAwaiter().GetResult();
+
+        protected virtual Task<IEnumerable<DValue<RecordValue>>> GetRowsInternal()
+        {
+            throw new Exception("No HttpClient context");
+        }
```
MikeStall, line 53: Can we remove this extra virtual, and just have Rows throw?

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorTableValueWithServiceProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        internal IServiceProvider ServiceProvider { get; }
+
+        public ConnectorTableValueWithServiceProvider(ConnectorTableValue tableValue, IServiceProvider serviceProvider)
+            : base(tableValue._tabularService, (tableValue.IRContext.ResultType as ConnectorTableType).RecordType)
+        {
+            ServiceProvider = serviceProvider;
+        }
+
+        private IEnumerable<DValue<RecordValue>> _cachedRows;
```
MikeStall, line 22: >IEnumerable [](http://example.com/codeflow?start=16&length=11)

can we change to something that includes length?

That ensures there isnt deferred network calls. 
All network calls should be in the duration of GetRows()

src/libraries/Microsoft.PowerFx.Connectors/Environment/TabularService.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            if (tableName is not StringValue sv)
+            {
+                throw new PowerFxConnectorException("'table' global value is not of type StringValue");
+            }
+
+            TableName = sv.Value;
+        }
+
+        public async Task<RecordType> InitAsync(PowerFxConfig config, string tableName, BaseRuntimeConnectorContext runtimeConnectorContext, CancellationToken cancellationToken)
```
MikeStall, line 41: >InitAsync [](http://example.com/codeflow?start=38&length=9)

move to derived class, since init params are specific to derived.

src/libraries/Microsoft.PowerFx.Connectors/Environment/TabularService.cs
```diff
@@ -0,0 +1,71 @@
...
+        internal abstract Task<RecordType> GetSchemaAsync(BaseRuntimeConnectorContext runtimeConnectorContext, CancellationToken cancellationToken);
+
+        // TABLE DATA SERVICE - CREATE
+        // POST: /datasets/{datasetName}/tables/{tableName}/items?api-version=2015-09-01
+
+        // TABLE DATA SERVICE - READ
+        // GET AN ITEM - GET: /datasets/{datasetName}/tables/{tableName}/items/{id}?api-version=2015-09-01
+
+        // LIST ITEMS - GET: /datasets/{datasetName}/tables/{tableName}/items?$filter=’CreatedBy’ eq ‘john.doe’&$top=50&$orderby=’Priority’ asc, ’CreationDate’ desc
+        public abstract Task<IEnumerable<DValue<RecordValue>>> GetItemsAsync(BaseRuntimeConnectorContext runtimeConnectorContext, CancellationToken cancellationToken);
```
MikeStall, line 63: >BaseRuntimeConnectorContext [](http://example.com/codeflow?start=77&length=27)

IServiceProvider?

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorTableValueWithServiceProvider.cs
LucGenetier: I made the change but notice that CollectionTableValue.Rows is only IEnumerable<DValue<RecordValue>> and this leads to a ToArray() call in GetItemsAsync

src/libraries/Microsoft.PowerFx.Connectors/Internal/InjectServiceProviderFunction.cs
LucGenetier: Current approach defers GetRows()…  will be useful for FirstN() …

src/libraries/Microsoft.PowerFx.Connectors/Environment/TabularService.cs
```diff
@@ -0,0 +1,54 @@
...
+        public abstract Task<RecordType> GetSchemaAsync(CancellationToken cancellationToken);
+
+        // TABLE DATA SERVICE - CREATE
+        // POST: /datasets/{datasetName}/tables/{tableName}/items?api-version=2015-09-01
+
+        // TABLE DATA SERVICE - READ
+        // GET AN ITEM - GET: /datasets/{datasetName}/tables/{tableName}/items/{id}?api-version=2015-09-01
+
+        // LIST ITEMS - GET: /datasets/{datasetName}/tables/{tableName}/items?$filter=’CreatedBy’ eq ‘john.doe’&$top=50&$orderby=’Priority’ asc, ’CreationDate’ desc
+        public abstract Task<ICollection<DValue<RecordValue>>> GetItemsAsync(IServiceProvider serviceProvider, CancellationToken cancellationToken);
```
MikeStall, line 46: >GetItemsAsync [](http://example.com/codeflow?start=63&length=13)

Can you add some kind of "FilterParameters" class that has strings for $filter and the others?

src/libraries/Microsoft.PowerFx.Connectors/Internal/InjectServiceProviderFunction.cs
MikeStall: Ok -but then we need to double check the caching in ConnectorTableValueWithServiceProvider._cachedRows

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorTableValueWithServiceProvider.cs
MikeStall: I don't see how paging could work here:
- the returned rows can't be a deferred object (ie, not IEnumerable) because network calls can't be deferred (they must use the current httpClient). 
- how would the page update the _cachedRows?

This could be an existing problem with our paging in dataverse.

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
```diff
@@ -73,7 +74,8 @@ public TableValue(TableType type)
         internal TableValue(IRContext irContext)
             : base(irContext)
         {
-            Contract.Assert(IRContext.ResultType is TableType);
```
MikeStall, line 76: >TableType [](http://example.com/codeflow?start=52&length=9)

Why this change?

src/libraries/Microsoft.PowerFx.Connectors/Internal/TabularDType.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Connectors
+{
+    // Used in TabularIRVisitor
+    internal class TabularDType : DType
```
MikeStall, line 10: >TabularDType [](http://example.com/codeflow?start=19&length=12)

This is probably fine now; but eventually it should be using DType.AssociatedDAtaSources.

for example, we'll need to capture delegation capabilities.

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorTableType.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Connectors
+{
+    public class ConnectorTableType : FormulaType
```
MikeStall, line 9: >ConnectorTableType [](http://example.com/codeflow?start=17&length=18)

can we make internal? 

- We will probably want to remove this later. 
- callers shouldn't depend on this distinction (vs just TableType)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.ConfigExtensions.AddActionConnector(Microsoft.PowerFx.PowerFxConfig,Microsoft.PowerFx.Connectors.ConnectorSettings,Microsoft.OpenApi.Models.OpenApiDocument,Microsoft.PowerFx.Connectors.ConnectorLogger)<br />❌CP0002 M:Microsoft.PowerFx.ConfigExtensions.AddActionConnector(Microsoft.PowerFx.PowerFxConfig,System.String,Microsoft.OpenApi.Models.OpenApiDocument,Microsoft.PowerFx.Connectors.ConnectorLogger)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.OpenApiParser.GetFunctions(Microsoft.PowerFx.Connectors.ConnectorSettings,Microsoft.OpenApi.Models.OpenApiDocument,Microsoft.PowerFx.Connectors.ConnectorLogger)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.OpenApiParser.GetFunctions(System.String,Microsoft.OpenApi.Models.OpenApiDocument,Microsoft.PowerFx.Connectors.ConnectorLogger)

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorTableValueWithServiceProvider.cs
LucGenetier: We'll deal with that later, not needed now

src/libraries/Microsoft.PowerFx.Connectors/Internal/InjectServiceProviderFunction.cs
LucGenetier: We only cache per expression

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
LucGenetier: To allow ConnectorTableType which is not a TableType (sealed)

src/libraries/Microsoft.PowerFx.Connectors/Internal/TabularDType.cs
LucGenetier: For future, yes
Mostly needed in TabularIRVisitor

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.ConfigExtensions.AddActionConnector(Microsoft.PowerFx.PowerFxConfig,Microsoft.PowerFx.Connectors.ConnectorSettings,Microsoft.OpenApi.Models.OpenApiDocument,Microsoft.PowerFx.Connectors.ConnectorLogger)<br />❌CP0002 M:Microsoft.PowerFx.ConfigExtensions.AddActionConnector(Microsoft.PowerFx.PowerFxConfig,System.String,Microsoft.OpenApi.Models.OpenApiDocument,Microsoft.PowerFx.Connectors.ConnectorLogger)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.OpenApiParser.GetFunctions(Microsoft.PowerFx.Connectors.ConnectorSettings,Microsoft.OpenApi.Models.OpenApiDocument,Microsoft.PowerFx.Connectors.ConnectorLogger)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.OpenApiParser.GetFunctions(System.String,Microsoft.OpenApi.Models.OpenApiDocument,Microsoft.PowerFx.Connectors.ConnectorLogger)<br />❌CP0007 T:Microsoft.PowerFx.Connectors.ODataParameters

src/libraries/Microsoft.PowerFx.Connectors/EngineExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Collections.Generic;
+using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Core.IR.Nodes;
+
+namespace Microsoft.PowerFx.Connectors
+{
+    public static class EngineExtensions
+    {
+        // To support tabular connectors, we need to use an IR transform to inject the ServiceProvider at runtime (which contains the HttpClient)
+        public static void EnableTabularConnectors(this Engine engine)
```
MikeStall, line 13: >static [](http://example.com/codeflow?start=15&length=6)

Can we add an [Obsolete] on this? 
Ie, it's something we're going to change in the future. We're adding to unblock limited use case.

src/libraries/Microsoft.PowerFx.Connectors/Environment/CdpTabularService.cs
```diff
@@ -0,0 +1,94 @@
...
+namespace Microsoft.PowerFx.Connectors
+{
+    // Implements CDP protocol for Tabular connectors
+    public class CdpTabularService : TabularService
+    {
+        public string DataSetName { get; }
+
+        public string TableName { get; }
+
+        protected HttpClient _httpClient;
```
MikeStall, line 22: >protected HttpClient _httpClient; [](http://example.com/codeflow?start=8&length=33)

Is it safe to hold this as a field? 
We may need to get from IServiceProvider  (which is why we pass SP ) 

Maybe:
virtual HttpClient GetClient(IServiceProvider)

src/libraries/Microsoft.PowerFx.Connectors/Environment/TabularService.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public async Task InitAsync(CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();            
+            RecordType = await GetSchemaAsync(cancellationToken).ConfigureAwait(false);
+        }
+
+        // TABLE METADATA SERVICE
+        // GET: /$metadata.json/datasets/{datasetName}/tables/{tableName}?api-version=2015-09-01
+        public abstract Task<RecordType> GetSchemaAsync(CancellationToken cancellationToken);
```
MikeStall, line 37: >public [](http://example.com/codeflow?start=8&length=6)

protected?

Force callers to get through t.InitAsync and then t.TableType

src/libraries/Microsoft.PowerFx.Connectors/Environment/TabularService.cs
```diff
@@ -0,0 +1,54 @@
...
+{
+    public abstract class TabularService
+    {
+        public RecordType RecordType { get; private set; } = null;
+
+        public TableType TableType => _tableType ??= RecordType.ToTable();
+
+        public bool IsInitialized => RecordType != null;
+
+        private TableType _tableType;
```
MikeStall, line 20: Can we make this just 1 field, not 2?

  private TableType _tableType; // <-- single source of truth

  public RecordType RecordType => _tableType.ToRecord();

  public TableType TableType => _tableType

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -123,6 +125,18 @@ public async Task<HttpRequestMessage> BuildRequest(FormulaValue[] args, IConvert
             }
 
             var url = (OpenApiParser.GetServer(_function.Servers, _httpClient) ?? string.Empty) + path + query.ToString();
+
+            // Replace connectionId or other parameters in the URL
+            foreach (Match m in new Regex(@"{(?<p>[^{}]+)}").Matches(url))
```
MikeStall, line 130: > new Regex(@"{(?<p>[^{}]+)}").Matches(url)) [](http://example.com/codeflow?start=31&length=43)

running a regex at this layer on every invoke seems scary. 

Du we really need this? Especially for CDP case if we're already building URIs directly

MikeStall:
>         internal ODataParameters(bool count, string filter, string orderby, int top)

can we make these public?

Callers will need to allocate so they can pass to GetItemsAsync



---
Refers to: src/libraries/Microsoft.PowerFx.Connectors/Obsolete/ODataQueryableTableValue.cs:62 in bf3a280. [](commit_id = bf3a2802a7ac20386da1926437e2b23f64ac9e2f, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorTableValue.cs
```diff
@@ -0,0 +1,49 @@
...
+using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Connectors
+{
+    // Returned by AddTabularConnector
+    // Doesn't contain any ServiceProvider which is runtime only
+    public class ConnectorTableValue : TableValue, IRefreshable
+    {
+        public new TableType Type => _tabularService.TableType;
```
MikeStall, line 16: >public new TableType Type => _tabularService.TableType; [](http://example.com/codeflow?start=8&length=55)

Do we need this? 
Doesn't TableValue.Type already have correct type?

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorTableValueWithServiceProvider.cs
```diff
@@ -0,0 +1,43 @@
...
+        public override IEnumerable<DValue<RecordValue>> Rows => GetRowsInternal().ConfigureAwait(false).GetAwaiter().GetResult();
+
+        private async Task<IEnumerable<DValue<RecordValue>>> GetRowsInternal()
+        {
+            if (_cachedRows != null)
+            {
+                return _cachedRows;
+            }
+
+            _cachedRows = await _tabularService.GetItemsAsync(ServiceProvider, null /* no OData filter */, CancellationToken.None).ConfigureAwait(false);
```
MikeStall, line 33: >null [](http://example.com/codeflow?start=79&length=4)

This is critical - the only reason we can cache _cachedRows is because this is null.

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
MikeStall: Can it be? 

Else, this may end up breaking other code that does this kind of check.

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorTableType.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Connectors
+{
+    internal class ConnectorTableType : FormulaType
```
MikeStall, line 9: >FormulaType [](http://example.com/codeflow?start=40&length=11)

can this derive from TableType?

Because of all the `Is TableType` checks.

src/tests/Microsoft.PowerFx.Connectors.Tests/FileTabularConnector.cs
```diff
@@ -0,0 +1,89 @@
...
+            Assert.Equal("FieldAccess(Last:![line:s](FirstN:*[line:s](InjectServiceProviderFunction:![line:s](ResolvedObject('File:RuntimeValues_XXX')), Float:n(2:w))), line)", ir);
+
+            // Use tabular connector. Internally we'll call ConnectorTableValueWithServiceProvider.GetRowsInternal to get the data
+            FormulaValue result = await check.GetEvaluator().EvalAsync(CancellationToken.None, symbolValues).ConfigureAwait(false);
+            StringValue str = Assert.IsType<StringValue>(result);
+            Assert.Equal("b", str.Value);
+        }
+    }
+
+    internal class FileTabularService : TabularService
```
MikeStall, line 67: >FileTabularService [](http://example.com/codeflow?start=19&length=18)

nice :)

src/tests/Microsoft.PowerFx.Connectors.Tests/PublicSurfaceTests.cs
```diff
@@ -36,9 +37,12 @@ public void PublicSurfaceTest_Connectors()
               "Microsoft.PowerFx.Connectors.ConnectorParameters",              
               "Microsoft.PowerFx.Connectors.ConnectorParameterWithSuggestions",
               "Microsoft.PowerFx.Connectors.ConnectorSchema",
-              "Microsoft.PowerFx.Connectors.ConnectorSettings",
+              "Microsoft.PowerFx.Connectors.ConnectorSettings",              
+              "Microsoft.PowerFx.Connectors.ConnectorTableValue",
+              "Microsoft.PowerFx.Connectors.ConnectorTableValueWithServiceProvider",
```
MikeStall, line 42: can we keep these internal?

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
LucGenetier: This Regex is very simple but I agree it's on the critical path.
We could cache the results (not in Match format of course) so that we would accelerate this a little bit.
We could also have a loop on url.Split('{')

Could be in a later PR or this one, I let you decide

src/libraries/Microsoft.PowerFx.Connectors/EngineExtensions.cs
LucGenetier: ??
What would you want to replace it with?

src/libraries/Microsoft.PowerFx.Connectors/Environment/TabularService.cs
LucGenetier: the init gives us a RecordType, not a TableType, that's why I made the RecordType the source of truth

Why would the opposite be better?

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorTableValue.cs
LucGenetier: Absolutely required, yes!
I dont' remember why but I'll confirm tomorrow

LucGenetier:
>         internal ODataParameters(bool count, string filter, string orderby, int top)

I think this is more than a draft and needs to be improved
This store is really dirty to my eyes

---
In reply to: [2045765729](https://github.com/microsoft/Power-Fx/pull/2306#issuecomment-2045765729) [](http://example.com/codeflow?ancestors=2045765729)

---
Refers to: src/libraries/Microsoft.PowerFx.Connectors/Obsolete/ODataQueryableTableValue.cs:62 in bf3a280. [](commit_id = bf3a2802a7ac20386da1926437e2b23f64ac9e2f, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Connectors/Internal/InjectServiceProviderFunction.cs
MikeStall: beware, an expression may call multiple times. 

`First(Table, 5); First(Table, 10)`
is a single expression. 

Again - I think the code works today (because the cache will just always pull all rows), -  but we should be aware of this nuance because this would be easy to break.

src/libraries/Microsoft.PowerFx.Connectors/EngineExtensions.cs
MikeStall: Not sure, will discuss in phase 3. 
Ideally, this call would not be needed at all. For example, imagine the IR Translator was smart enough to pull rewriters from the type directly (rather than needing a separate registration). 

The point is:
-  it's not intrinsically necessary 
- but fixing it is tricky 
- it's ok for this PR

src/libraries/Microsoft.PowerFx.Connectors/Environment/TabularService.cs
MikeStall: The key is to have 1 source of truth. (only 1 mutable field/property; everthing else is read-only based on that). 
It can be either record or table. 

Although does it need to be table to return the special table type you create?

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
LucGenetier: I imagined there was a good reason
I'll "unseal" TableType and update tormoow

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
MikeStall: It's not just performance; it's risk and and suggests a layering issue and we need to convince ourselves it's safe.  

What about this PR is forcing us to add this? It's for building a request  and CDPTabularService builds the request directly (and I think it skips this method altogether?)

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorTableType.cs
MikeStall: That suggests adding another TableType may really break some invariants. 

Why do we need this? Is existing TableType not sufficient?

src/libraries/Microsoft.PowerFx.Connectors/Internal/InjectServiceProviderFunction.cs
LucGenetier: See SQL_Tabular test, lines 2000-2004

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
LucGenetier: It was a bug + I need it for this PR
We need these replacements for url parameters (like server, database) and I included ConnectionId

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorTableType.cs
LucGenetier: See TabularIRVisitor where we test this type
I could derive from (unsealed) TableType

src/libraries/Microsoft.PowerFx.Connectors/Environment/CdpTabularService.cs
LucGenetier: instead of passing an HttpClient in the constructor, I'll use Func<HttpClient> like for the jwt token in PowerAppsClient

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorTableValue.cs
LucGenetier: Not needed anymore

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorTableType.cs
LucGenetier: we need a way to identify this specific type in IR rewriter, either in having a derived TableType or a specific DKind

LucGenetier:
Public API changes: <br />❌CP0001 T:Microsoft.PowerFx.Connectors.ODataQueryableTableValue<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ODataParameters.AddTo(System.Collections.Specialized.NameValueCollection)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ODataParameters.WithCount<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ODataParameters.WithFilter(System.String)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ODataParameters.WithOrderby(System.String)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ODataParameters.WithTop(System.Int32)<br />❌CP0007 T:Microsoft.PowerFx.Connectors.ODataParameters

LucGenetier:
Public API changes: <br />❌CP0001 T:Microsoft.PowerFx.Connectors.ODataQueryableTableValue<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ODataParameters.AddTo(System.Collections.Specialized.NameValueCollection)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ODataParameters.WithCount<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ODataParameters.WithFilter(System.String)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ODataParameters.WithOrderby(System.String)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ODataParameters.WithTop(System.Int32)<br />❌CP0007 T:Microsoft.PowerFx.Connectors.ODataParameters

LucGenetier:
Public API changes: <br />❌CP0001 T:Microsoft.PowerFx.Connectors.ODataQueryableTableValue<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ODataParameters.AddTo(System.Collections.Specialized.NameValueCollection)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ODataParameters.WithCount<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ODataParameters.WithFilter(System.String)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ODataParameters.WithOrderby(System.String)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ODataParameters.WithTop(System.Int32)<br />❌CP0007 T:Microsoft.PowerFx.Connectors.ODataParameters

src/libraries/Microsoft.PowerFx.Connectors/Internal/TabularDType.cs
MikeStall: For PA/DV, tabular is specified via an existing table DType , but with AssociatedDAtaSources set to extra info

src/libraries/Microsoft.PowerFx.Connectors/Environment/CdpTabularService.cs
```diff
@@ -0,0 +1,126 @@
...
+namespace Microsoft.PowerFx.Connectors
+{
+    // Implements CDP protocol for Tabular connectors
+    public class CdpTabularService : TabularService
+    {
+        public string DataSetName { get; }
+
+        public string TableName { get; }
+
+        protected Func<HttpClient> _getHttpClient;
```
MikeStall, line 22: >Func<HttpClient>  [](http://example.com/codeflow?start=18&length=17)

this may still need the Runtime's IServiceProvider.

src/libraries/Microsoft.PowerFx.Connectors/Environment/CdpTabularService.cs
```diff
@@ -0,0 +1,126 @@
...
+        public override async Task<ICollection<DValue<RecordValue>>> GetItemsAsync(IServiceProvider serviceProvider, CancellationToken cancellationToken)
+        {
+            try
+            {
+                ConnectorLogger executionLogger = serviceProvider.GetService<ConnectorLogger>();
+                executionLogger?.LogInformation($"Entering in {nameof(CdpTabularService)} {nameof(GetItemsAsync)} for {DataSetName}, {TableName}");
+
+                Uri uri = new Uri(_uriPrefix + (_v2 ? "/v2" : string.Empty) + $"/datasets/{DataSetName}/tables/{TableName}/items?api-version=2015-09-01", UriKind.Relative);
+
+                ODataParameters odataParameters = serviceProvider.GetService<ODataParameters>();
```
MikeStall, line 86: >serviceProvider.GetService<ODataParameters>(); [](http://example.com/codeflow?start=50&length=46)

this should be a normal param, not from SP.

src/libraries/Microsoft.PowerFx.Connectors/Environment/CdpTabularService.cs
```diff
@@ -0,0 +1,126 @@
...
+                if (odataParameters != null)
+                {
+                    uri = odataParameters.GetUri(uri);
+                }
+
+                using HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Get, uri);
+                using HttpResponseMessage response = await _getHttpClient().SendAsync(request, cancellationToken).ConfigureAwait(false);
```
MikeStall, line 93: >_getHttpClient() [](http://example.com/codeflow?start=59&length=16)

serviceProvider has the httpclient.

src/libraries/Microsoft.PowerFx.Connectors/Environment/TabularService.cs
```diff
@@ -0,0 +1,52 @@
...
+        public bool IsInitialized => TableType != null;
+
+        public virtual ConnectorTableValue GetTableValue()
+        {
+            return IsInitialized
+                ? new ConnectorTableValue(this, TableType)
+                : throw new InvalidOperationException("Tabular service is not initialized.");
+        }
+
+        public async Task InitAsync(CancellationToken cancellationToken)
```
MikeStall, line 25: > public async Task InitAsync(CancellationToken cancellationToken) [](http://example.com/codeflow?start=7&length=65)

tbd?

What we really want is an async constructor

src/libraries/Microsoft.PowerFx.Connectors/Environment/TabularService.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Connectors
+{
+    public abstract class TabularService
+    {
+        public TableType TableType { get; private set; } = null;
```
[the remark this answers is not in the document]
LucGenetier: We could but we have no need for now
As this is a Tabular service, only TableType should matter

src/libraries/Microsoft.PowerFx.Connectors/Environment/TabularService.cs
LucGenetier: It can be sync and async 
See File tabular example which is sync

LucGenetier:
Public API changes: <br />❌CP0001 T:Microsoft.PowerFx.Connectors.ODataQueryableTableValue<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ODataParameters.AddTo(System.Collections.Specialized.NameValueCollection)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ODataParameters.WithCount<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ODataParameters.WithFilter(System.String)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ODataParameters.WithOrderby(System.String)<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ODataParameters.WithTop(System.Int32)<br />❌CP0007 T:Microsoft.PowerFx.Connectors.ODataParameters

src/libraries/Microsoft.PowerFx.Connectors/Environment/TabularService.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        // TABLE DATA SERVICE - READ
+        // GET AN ITEM - GET: /datasets/{datasetName}/tables/{tableName}/items/{id}?api-version=2015-09-01
+
+        // LIST ITEMS - GET: /datasets/{datasetName}/tables/{tableName}/items?$filter=’CreatedBy’ eq ‘john.doe’&$top=50&$orderby=’Priority’ asc, ’CreationDate’ desc
+        public Task<ICollection<DValue<RecordValue>>> GetItemsAsync(IServiceProvider serviceProvider, ODataParameters oDataParameters, CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            return IsInitialized
```
MikeStall, line 47: super nit: can this just be:

ThrowIfNotInitialized(); 

Since every method will want this

MikeStall (approved):
:shipit:

Issue #2304, opened 2024-04-03
Binder should suggest based on "AllowSideEffects"


anderson-joyle:
Patch has both self-contained and not self-contained overloads.
Depending on the context, the binder should suggest the correct overload.

https://github.com/microsoft/Power-Fx/blob/e33328bf231ded766918f84814d0d320e8ea23ab/src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs#L247

PR #2307, opened 2024-04-04
Binder should suggest based on "AllowSideEffects"

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2304.


LucGenetier:
✅ No public API change.

anderson-joyle:
> > ```
> >     [InlineData("Patch(table2,|", "record2")]
> > ```
> 
> Since we are not in a behavior context here (i.e., allowSideEffect = false), should we suggest this record?
> 
> Refers to: src/tests/Microsoft.PowerFx.Interpreter.Tests/InterpreterSuggestTests.cs:224 in [2d2aeca](https://github.com/microsoft/Power-Fx/commit/2d2aecac3876686665be56a26c5f0b24b445cd33). [](commit_id = [2d2aeca](https://github.com/microsoft/Power-Fx/commit/2d2aecac3876686665be56a26c5f0b24b445cd33), deletion_comment = False)

This is a good question. In this context, the only possible overload will be `Patch(record, record, ...)`. The first argument is wrong, but the engine would still suggest a good `argN`.

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2308, opened 2024-04-05
Heavily Refactor LSP (Still Backwards Compatible) and Make it Async and Customizable

akshar2401:
This pr significantly refactors LSP SDK to truly make it async and bit customizable. This pr also reorganizes existing tests into a much modular structure while still keeping around the old tests as a proof that the changes in this pr are backwards compatible. 

The main goal of this PR is to make LSP async in such as a way that hosts can run expensive operations like nl2fx network calls asynchronously while making sure the operations that need to run synchronously acquire correct host specific locks.

This pr accomplishes it goals by 
1) Refactoring a monolith language server class into small handlers targeting only one specific method. Each handler is async by default and this allows us to not let LSP block threads inside its hosts
2) LSP in this pr identifies critical sections which are delegated to hosts for their executions. Host provides an environment using which LSP can acquire appropriate locks and run those critical sections inside Host environment
3) LSP, now being async, allows hosts to run Nl2Fx or Fx2Nl network calls in a non blocking way


src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Fx2Nl/BackwardsCompatibleFx2NlLanguageServerOperationHandler.cs
```diff
@@ -0,0 +1,45 @@
...
+        public BackwardsCompatibleFx2NlLanguageServerOperationHandler(INLHandlerFactory nLHandlerFactory)
+        {
+            _nLHandlerFactory = nLHandlerFactory;
+        }
+
+        protected override async Task<bool> PreHandleFx2NlAsync(LanguageServerOperationContext operationContext, CancellationToken cancellationToken)
+        {
+            // Existing SDK relies on GetNLHandler to get the handler for Nl2Fx operation.
+            // This is a legacy handler and not to be confused with the new handler pattern.
+            _nlHandler = _nLHandlerFactory.GetNLHandler(operationContext.GetScope(_fx2NlRequestParams.TextDocument.Uri), _fx2NlRequestParams) ?? throw new NullReferenceException("No suitable handler found to handle Fx2Nl");
```
MikeStall, line 29: >_nlHandler [](http://example.com/codeflow?start=12&length=10)

Can we gaurantee that this field is not overwritten by 2 handlers running concurrently?
Could we set it back  to null at line 42, and assert it's null here?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/JsonRpc/LanguageServerInput.cs
```diff
@@ -0,0 +1,80 @@
...
+
+                    if (root.TryGetProperty("jsonrpc", out var jsonrpcElement))
+                    {
+                        input.Jsonrpc = jsonrpcElement.GetString();
+                    }
+
+                    if (root.TryGetProperty("params", out var paramsElement))
+                    {
+                        input.Params = paramsElement.GetRawText();
+                    }
```
MikeStall, line 69: Could we use strong parsing here?
Deserialize<LanguageServerInput>(jsonRpcPAyload)

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -73,19 +72,125 @@ public class LanguageServer
...
+            return outputBuilder.Response;
+        }
+
+        public async Task<string> OnDataReceivedAsync(string jsonRpcPayload, CancellationToken cancellationToken)
+        {
+            var outputBuilder = await OnDataRecievedAsyncInternal(jsonRpcPayload, CancellationToken.None).ConfigureAwait(false);
+            return outputBuilder.Response;
+        }
+
+        private async Task<LanguageServerOutputBuilder> OnDataRecievedAsyncInternal(string jsonRpcPayload, CancellationToken cancellationToken)
```
MikeStall, line 129: >OnDataRecievedAsyncInternal [](http://example.com/codeflow?start=56&length=27)

super nit - typo "Received"

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -73,19 +72,125 @@ public class LanguageServer
...
+            try
+            {
+                if (input.Method == null)
+                {
+                    _loggerInstance?.LogError($"[PFX] OnDataReceived CreateErrorResult InvalidRequest (method not found)");
+                    outputBuilder.AddInvalidRequestError(input.Id, $"Invalid method");
+                    return outputBuilder;
+                }
+
+                var handler = GetHandlerFactory().GetHandler(input.Method, new HandlerCreationContext(LogUnhandledExceptionHandler));
```
MikeStall, line 147: > GetHandlerFactory().GetHandler [](http://example.com/codeflow?start=29&length=31)

This is great. Moving from a giant switch to something more modular.

MikeStall:
General direction here looks great. 

Impressive that:
- minimal test changes , less risk that we broke something. 
- no change to csproj  (ie, we didn't pickup a new dependency)
- moving from a giant switch case to more structured handlers.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/LanguageServerOperationContext.cs
```diff
@@ -0,0 +1,149 @@
...
+using Microsoft.PowerFx.Core;
+using Microsoft.PowerFx.Intellisense;
+
+namespace Microsoft.PowerFx.LanguageServerProtocol.Handlers
+{
+    /// <summary>
+    /// A context for a language server operation with information that might be needed when handling the operation.
+    /// It also has helper methods to simplify the operation handling.
+    /// </summary>
+    public class LanguageServerOperationContext
```
MikeStall, line 15: >LanguageServerOperationContext [](http://example.com/codeflow?start=17&length=30)

Can we document the lifespan of this? I assume this is per-request (given the requestId) , and not something that can be cached.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/CodeActions/BaseCodeActionsLanguageServerOperationHandler.cs
```diff
@@ -0,0 +1,109 @@
...
+using System.Web;
+using Microsoft.PowerFx.LanguageServerProtocol.Protocol;
+using static Microsoft.PowerFx.LanguageServerProtocol.LanguageServer;
+
+namespace Microsoft.PowerFx.LanguageServerProtocol.Handlers
+{
+    /// <summary>
+    /// Handler for code actions.
+    /// </summary>
+    public class BaseCodeActionsLanguageServerOperationHandler : ILanguageServerOperationHandler
```
MikeStall, line 17: > Base [](http://example.com/codeflow?start=16&length=5)

Why do these all start with "Base"? Do folks derive from them?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/ScopeVariations/IAsyncPowerFxScope.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// </summary>
+    public interface IAsyncPowerFxScope : IPowerFxScope
+    {
+        /// <summary>
+        /// Check for errors in the given expression. 
+        /// </summary>
+        /// <param name="expression">The expression to validate.</param>
+        /// <param name="cancellationToken">Cancellation token.</param>
+        /// <returns>CheckResult.</returns>
+        Task<CheckResult> CheckAsync(string expression, CancellationToken cancellationToken);
```
MikeStall, line 22: >CheckAsync [](http://example.com/codeflow?start=26&length=10)

Why would Check be async?  This normally should be an entirely sync method.


I could imagine SuggestAsync being async if we support dynamic intellisense.

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.LanguageServerProtocol.LanguageServer.GetCharPosition(System.String,System.Int32)<br />❌CP0002 M:Microsoft.PowerFx.LanguageServerProtocol.LanguageServer.GetPosition(System.String,System.Int32,System.Int32)<br />❌CP0002 M:Microsoft.PowerFx.LanguageServerProtocol.LanguageServer.GetRange(System.String,Microsoft.PowerFx.Syntax.Span)

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/JsonRpc/LanguageServerInput.cs
akshar2401: Params differ in structure depending on what lsp method we are trying to handle. The parsing of params is left up to the handlers as they know how to parse their respective params structure. If it intellisense, params is deserialized to completion params, if it is signature help, it is deserialized to signature params and so on. This was the case in old sdk as welll. If we want to strongly parse it here then we would need to switch case on lsp method and select the right type. 

It makes sense for each handlers to take the responsibility of the params. Doing it here seems infeasible and violation of separation of concerncs

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/ScopeVariations/IAsyncPowerFxScope.cs
akshar2401: Yes computing check is synchronous ideally but in Document Server, we do this behind a lock and putting it behind a lock in document server makes the check result computation effectively asynchronous. A host can have requirements like this where they wrap traditional check result computation in a async code path.
An example:
image

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/JsonRpc/LanguageServerInput.cs
MikeStall: Let's chat more and I can explain.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Fx2Nl/Fx2NLDTOs.cs
```diff
@@ -0,0 +1,34 @@
...
+using Microsoft.PowerFx.LanguageServerProtocol.Protocol;
+
+namespace Microsoft.PowerFx.LanguageServerProtocol.Handlers
+{
+    /// <summary>
+    /// Represents the result of pre-handling step of Fx2Nl.
+    /// </summary>
+    /// <param name="checkResult">Check result of pre handle step.</param>
+    /// <param name="parameters"> Fx2Nl Parameters computed during pre handle step. </param>
+    public record Fx2NlPreHandleResult(CheckResult checkResult, Fx2NLParameters parameters);
```
MikeStall, line 13: >record [](http://example.com/codeflow?start=11&length=6)

[https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record)

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Fx2Nl/BaseFx2NlLanguageServerOperationHandler.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// <summary>
+        /// Performs core Fx2Nl operation.
+        /// </summary>
+        /// <param name="operationContext">Language Server Operation Context.</param>
+        /// <param name="cancellationToken">Cancellation Token.</param>
+        /// <param name="handleContext">Fx2Nl Handle Context.</param>
+        /// <returns> Extended Fx2Nl Handle Contex.</returns>
+        protected virtual async Task<Fx2NlHandleContext> Fx2NlAsync(LanguageServerOperationContext operationContext, Fx2NlHandleContext handleContext, CancellationToken cancellationToken)
+        {
+            return handleContext;
```
MikeStall, line 56: >   return handleContext; [](http://example.com/codeflow?start=9&length=24)

Throw? / abstract?

akshar2401: made this abstract

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/ILanguageServerOperationHandler.cs
```diff
@@ -0,0 +1,33 @@
...
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.PowerFx.LanguageServerProtocol.Handlers
+{
+    /// <summary>
+    /// A interface representing a handler for a language server operation.
+    /// </summary>
+    public interface ILanguageServerOperationHandler
```
MikeStall, line 12: >public [](http://example.com/codeflow?start=4&length=6)

internal

src/tests/Microsoft.PowerFx.Interpreter.Tests/LanguageServiceProtocol/PublicSurfaceTests.cs
```diff
@@ -40,7 +40,29 @@ public void Test()
...
                 "Microsoft.PowerFx.LanguageServerProtocol.Fx2NLParameters",
-                "Microsoft.PowerFx.LanguageServerProtocol.UsageHints",
+                "Microsoft.PowerFx.LanguageServerProtocol.UsageHints",
+                "Microsoft.PowerFx.LanguageServerProtocol.DiagnosticsHelper",
+                "Microsoft.PowerFx.LanguageServerProtocol.LanguageServerHelper",
+                "Microsoft.PowerFx.LanguageServerProtocol.LanguageServerInput",
+                "Microsoft.PowerFx.LanguageServerProtocol.LanguageServerOutput",
+                "Microsoft.PowerFx.LanguageServerProtocol.LanguageServerOutputBuilder",
+                "Microsoft.PowerFx.LanguageServerProtocol.ILanguageServerLogger",
+                "Microsoft.PowerFx.LanguageServerProtocol.IHostTaskExecutor",
```
MikeStall, line 50: >            "Microsoft.PowerFx.LanguageServerProtocol.IHostTaskExecutor", [](http://example.com/codeflow?start=4&length=73)

IHostTaskExecutor needs to be public, but can we make most of these internal?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Nl2Fx/Nl2FxLanguageServerOperationHandler.cs
```diff
@@ -0,0 +1,132 @@
...
+            _nl2FxResult = await _nlHandler.NL2FxAsync(_nl2FxParameters, cancellationToken).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Performs post-handle operations for NL2FX.
+        /// </summary>
+        /// <param name="operationContext">Language Server Operation Context.</param>
+        private void PostHandleNl2FxResults(LanguageServerOperationContext operationContext)
+        {
+            var nl2FxResult = _nl2FxResult;
```
MikeStall, line 75: >  var nl2FxResult = _nl2FxResult; [](http://example.com/codeflow?start=10&length=33)

If we want to disable FinalCheck,  we could add a policy flag on _nlHandler (which host provides) and check that.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/LanguageServerOperationContext.cs
```diff
@@ -0,0 +1,160 @@
...
+            return _scope ??= _scopeFactory.GetOrCreateInstance(uri);
+        }
+
+        /// <summary>
+        /// A helper method to convert the expression to display format.
+        /// </summary>
+        /// <param name="uri">Uri to use for check result creation.</param>
+        /// <param name="expression">Expression to create check result for.</param>
+        /// <returns>Converted expression.</returns>
+        public string ConvertToDisplay(string uri, string expression)
```
MikeStall, line 80: >ConvertToDisplay [](http://example.com/codeflow?start=22&length=16)

This should probably be under lock

akshar2401: No this is the method i forgot to remove

LucGenetier:
✅ No public API change.

MikeStall (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Host/IHostTaskExecutor.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        /// <summary>
+        ///  Executes a task in host environment if host task executor is available.
+        ///  This is critical to trust between LSP and its hosts.
+        ///  LSP should correctly use this and wrap particulat steps that need to run in host using these methods. 
+        /// </summary>
+        /// <param name="operationContext">Language Server Operation Context.</param>
+        /// <param name="task">Task to run inside host.</param>
+        /// <param name="cancellationToken">Cancellation Token.</param>
+        Task ExecuteTaskAsync(Action task, LanguageServerOperationContext operationContext, CancellationToken cancellationToken);
```
MikeStall, line 37: >Task ExecuteTaskAsync(Action task, [](http://example.com/codeflow?start=8&length=34)

Interfaces should  be minimal -  can we just make this an extension method. It's a convenience over ExecuteTask<void>

akshar2401: I removed this method. Agreed interfaces should be minimal but then we also try to avoid making changes to interfaces that we define out of the fear that it would break hosts (mostly compilation errors) and derive many other interfaces. Having too many interfaces because we don't want to modify existing ones as it breaks hosts is also not ideal as it can confuse hosts on which interfaces to implement. We need to reeval this strategy at some point but for now i have gotten rid of this method

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Fx2Nl/Fx2NlLanguageServerOperationHandler.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        /// <summary>
+        /// Performs core Fx2Nl operation.
+        /// </summary>
+        /// <param name="operationContext">Language Server Operation Context.</param>
+        /// <param name="cancellationToken">Cancellation Token.</param>
+        private async Task Fx2NlAsync(LanguageServerOperationContext operationContext, CancellationToken cancellationToken)
+        {
+            var result = await _nLHandler.Fx2NLAsync(_preHandleCheckResult, _fx2NlParameters, cancellationToken).ConfigureAwait(false);
```
MikeStall, line 77: > _nLHandler.Fx2NLAsync(_preHandleCheckResult, [](http://example.com/codeflow?start=30&length=45)

Subtle bug - usage of checkREsult should be done under the lock. 
[https://msazure.visualstudio.com/DefaultCollection/OneAgile/_git/Power%20Fx%20Internal?path=/src/Microsoft.PowerFx.NaturalLanguage/Helper/LanguageServerExtensions.cs&version=GBmain&line=98&lineEnd=99&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents](https://msazure.visualstudio.com/DefaultCollection/OneAgile/_git/Power%20Fx%20Internal?path=/src/Microsoft.PowerFx.NaturalLanguage/Helper/LanguageServerExtensions.cs&version=GBmain&line=98&lineEnd=99&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)

So CheckResult shouldn't be the field, just the Fx2NLPayload that we computed from it.

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Fx2Nl/Fx2NlLanguageServerOperationHandler.cs
akshar2401: We talked offline and i think this is not a blocker and we should do an immediate follow up on this?

MikeStall (approved):

LucGenetier:
✅ No public API change.

Issue #2309, opened 2024-04-09
Lexer is removing chars


anderson-joyle:
Lexer is removing chars from string inputs.
This should work:
```
var expression = "\"AA" + char.ConvertFromUtf32(8203) + "BB\""; // expression.Length is 7
var tokens = TexlLexer.InvariantLexer.LexSource(expression, TexlLexer.Flags.None);
var stringToken = tokens[0] as StrLitToken; // stringToken.Legth 
Assert.Equal(stringToken.Value.Length, expression.Length - 2);
```

PR #2310, opened 2024-04-09
Removing lexer validation for special chars

anderson-joyle:
https://github.com/microsoft/Power-Fx/issues/2309

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
@@ -1567,7 +1567,7 @@ namespace Microsoft.PowerFx.Syntax
                         _sb.Append(ch);
                         NextChar();
                     }
-                    else if (!CharacterUtils.IsFormatCh(ch))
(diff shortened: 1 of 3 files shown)


LucGenetier:
✅ No public API change.

akshar2401:
UFB (new formula bar) has a logical dependency on the Lexer for its default syntax hightlighting capabilities and like Lexer (before this pr), UFB disallowed format category. After this pr goes in, UFB would no longer have parity with the Lexer. How frequent are the cases where identifiers contain these format category chars?

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -1567,7 +1567,7 @@ private Token LexStringLit()
                         _sb.Append(ch);
                         NextChar();
                     }
-                    else if (!CharacterUtils.IsFormatCh(ch))
+                    else
```
MikeStall, line 1570: To migitate breaks, can we limit this to under v1 flag.

PR #2311, opened 2024-04-09
Detect when a body has multiple parameters and when one of them is blob

LucGenetier:
This is a particular condition we don't support and should detect

Otherwise we end up with this exception
image


odivay (approved):

LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #2312, opened 2024-04-09
WIP: Adds warning on comparing same FirstNameNode

jas-valgotar:
This pull request introduces a warning when comparing two identical FirstNameNode instances, which could be either global or scoped variables. 
This enhancement is particularly beneficial in scenarios involving scoped variables, where it's easy to misunderstand the concept of implicit `ThisRecord`. 

It's important to note that this change is non-breaking, as it only adds a warning without generating an error.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -3862,6 +3862,12 @@ namespace Microsoft.PowerFx.Core.Binding
                 var leftType = _txb.GetType(node.Left);
                 var rightType = _txb.GetType(node.Right);
 
+                // Helps warn no op comparison, e.g. Filter(table, ThisRecord.Value = Value)) or Filter(table, Value = Value)
+                if (IsNoOPFirstNameComparison(node))
+                {
+                    _txb.ErrorContainer.EnsureError(DocumentErrorSeverity.Severe, node, TexlStrings.WrnNoOpFieldComparison);
+                }
+
                 var res = CheckBinaryOpCore(_txb.ErrorContainer, node, _txb.Features, leftType, rightType, _txb.BindingConfig.NumberIsFloat);
 
                 foreach (var coercion in res.Coercions)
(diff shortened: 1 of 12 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -3882,6 +3888,48 @@ public override void PostVisit(BinaryOpNode node)
...
+                    node.Op == BinaryOp.GreaterEqual) && 
+                    IsNoOPFirstNameComparison(node.Left, node.Right))
+                {
+                    return true;
+                }
+
+                return false;
+            }
+
+            private static bool IsNoOPFirstNameComparison(TexlNode left, TexlNode right)
```
MikeStall, line 3907: >IsNoOPFirstNameComparison [](http://example.com/codeflow?start=32&length=25)

this is key change...

MikeStall (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
jas-valgotar: NOTE: I will check the impact of the change with the PowerApps using a private build.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -3862,6 +3862,12 @@ public override void PostVisit(BinaryOpNode node)
                 var leftType = _txb.GetType(node.Left);
                 var rightType = _txb.GetType(node.Right);
 
+                // Helps warn no op comparison, e.g. Filter(table, ThisRecord.Value = Value)) or Filter(table, Value = Value)
+                if (IsNoOPFirstNameComparison(node))
+                {
+                    _txb.ErrorContainer.EnsureError(DocumentErrorSeverity.Severe, node, TexlStrings.WrnNoOpFieldComparison);
```
CarlosFigueiraMSFT, line 3868: Shouldn't it be a warning instead of DocumentErrorSeverity.Severe?

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4575,4 +4575,8 @@
     <value>Delegation warning. The result of this argument '{0}' may be truncated for large data sets before being passed to the '{1}' function.</value>
     <comment>Error message when an argument to non-delegable function has possible delegation and resulting rows may be truncated</comment>
   </data>
+  <data name="WrnNoOpFieldComparison" xml:space="preserve">
+    <value>This comparison will always be constant. (Always yields true or false). Please consider using 'As' keyword for aliasing scoped variable if this not what you want.</value>
```
lesaltzm, line 4579: aliasing isn't a super low-code friendly term. run this by a PM or content person?

PR #2316, opened 2024-04-11
Update function parameter resource comments to instruct that resources should be localized

CarlosFigueiraMSFT:
Based on feedback from the localization team, resources that contain `_` are typically not localized. This updates all function parameters which contain those with explicit instructions to localize them.


LucGenetier:
✅ No public API change.

anderson-joyle (approved):
:shipit:

Issue #2301, opened 2024-04-02
Summarize function


anderson-joyle:
Summarize function implementation.

PR #2317, opened 2024-04-12
Summarize function 

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2301.


src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/GroupByFunction.cs
```diff
@@ -0,0 +1,33 @@
...
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Syntax;
+
+#pragma warning disable SA1402 // File may only contain a single type
+#pragma warning disable SA1649 // File name should match first type name
+
+namespace Microsoft.PowerFx.Core.Texl.Builtins
+{
+    internal sealed class GroupByFunction : TexlFunction
```
MikeStall, line 19: >GroupByFunction [](http://example.com/codeflow?start=26&length=15)

do we need this?

anderson-joyle: No. This will be removed. Thanks for pointing this out.

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -332,6 +334,28 @@ private static void MutationFunctionsTestSetup(RecalcEngine engine, bool numberI
             engine.UpdateVariable("t_bs2", FormulaValue.NewTable(t_bsType.ToRecord()));
         }
 
+        private static void LoadDataTestSetup(RecalcEngine engine, bool numberIsFloat)
```
anderson-joyle, line 337: >LoadDataTestSetup [](http://example.com/codeflow?start=28&length=17)

@MikeStall @carlosfigueira Any better suggestions on how to test load a table with data to run `Summarize` tests? I'm trying to keep `Summarize` tests as sharable as possible with PA. Without this loading stage, each test case in txt will have 20+ lines just to initialize a TableValue.

anderson-joyle:
> {Supplier:"Contoso",  Fruit:"Grapes",  Price:220, Purchase:Date(2015,10,1)}

@MikeStall @carlosfigueira These lines are read by `LoadDataTestSetup` to create and load a TableValue.

---
Refers to: src/tests/Microsoft.PowerFx.Interpreter.Tests/Data/Summarize.txt:1 in f402479. [](commit_id = f402479363b094771ff274963c00e3841092f906, deletion_comment = False)

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
MikeStall: what about mutation scripts (like what we use for patch/collect)?  Those have shared state.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
```diff
@@ -0,0 +1,183 @@
...
+            Contracts.AssertValue(errors);
+            Contracts.Assert(MinArity <= args.Length && args.Length <= MaxArity);
+
+            bool isValid = base.CheckTypes(context, args, argTypes, errors, out returnType, out nodeToCoercedTypeMap);
+            Contracts.Assert(returnType.IsTable);
+
+            // Source table type
+            DType sourceType = argTypes[0];
+
+            if (!sourceType.IsTable)
```
MikeStall, line 95: if it's not a table, should we just issue an error and stop?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
```diff
@@ -0,0 +1,183 @@
...
+                if (!existingType.IsPrimitive)
+                {
+                    errors.EnsureError(DocumentErrorSeverity.Severe, arg, TexlStrings.ErrNeedPrimitive);
+                    isValid = false;
+                    continue;
+                }
+
+                var fError = false;
+                returnType = returnType.Add(ref fError, DPath.Root, columnName, existingType);
+                if (fError)
```
MikeStall, line 143: >if (fError) [](http://example.com/codeflow?start=16&length=11)

is this where we'd catch it if we had a collision in names?
Sumarize(..., expr1 As Column1, expr2 As Column1)

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
```diff
@@ -0,0 +1,183 @@
...
+    // Summarize( Table, GroupByColumn1 [, GroupByColumn2 …], AggregateExpr1 As Name [, AggregateExpr1 As Name …] )
+    internal sealed class SummarizeFunction : FunctionWithTableInput
+    {
+        public override bool IsSelfContained => true;
+
+        public override bool SupportsParamCoercion => false;
+
+        public override bool HasColumnIdentifiers => true;
+
+        private static readonly IReadOnlyCollection<string> _supportedAggregateFunctions = new HashSet<string>()
```
MikeStall, line 24: >_supportedAggregateFunctions [](http://example.com/codeflow?start=60&length=28)

where do we use this?

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
CarlosFigueiraMSFT: Mutation scripts are not run in Power Apps (issue with the JS test runner). At this point I'd rather have the tests with many lines in Summarize.txt instead of having this custom setup - that has the advantage of making the tests self-contained.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -3018,6 +3018,34 @@
...
+  <data name="AboutSummarize" xml:space="preserve">
+    <value>Returns a grouping of the specified table, combining with aggregation functions.</value>
+    <comment>Description of 'Patch' function.</comment>
+  </data>
+  <data name="SummarizeArg1" xml:space="preserve">
+    <value>source</value>
+    <comment>function_parameter - First argument to the Summarize function - the data source (table / collection) to operate on.</comment>
+  </data>
+  <data name="SummarizeArg2" xml:space="preserve">
+    <value>groupby_column</value>
```
CarlosFigueiraMSFT, line 3030: >groupby [](http://example.com/codeflow?start=11&length=7)

Suggestion: `grouping_column`

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Summarize.txt
```diff
@@ -0,0 +1,97 @@
...
+// No grouping columns
+>> Summarize( Sales, CountRows( Distinct( ThisGroup, Supplier ) ) As CountOfSuppliers )
+Errors: Error 0-84: Summarize requires at least one grouping column.
+
+// Unnamed aggregate
+>> Summarize( Sales, Fruit, CountRows( Distinct( ThisGroup, Supplier ) ) )
+Errors: Error 0-71: Summarize arguments must be either existing column names to group by or formulas named with the As keyword.
+
+// Can't reference aggregate column from other aggregates
+// >> Summarize( Sales, Fruit, Average( ThisGroup, Price ) As Avg_Price, Sum( ThisGruop, Price ) As Total_Cost, Total_Cost/Avg_Price As Ratio )
```
CarlosFigueiraMSFT, line 94: >ThisGruop [](http://example.com/codeflow?start=78&length=9)

[typo] ThisGroup

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
anderson-joyle: This can be removed. Other checks are catching this.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
anderson-joyle: On second thought, I removed this for now since the class already validates any aggregation class that returns anything different than an primitive result.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
anderson-joyle: Yes. Test case has been added.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests/MutationFunctionsTests.cs
```diff
@@ -24,6 +24,27 @@ public class MutationFunctionsTests : PowerFxTest
...
+        //                      {Supplier:""Fabrikam"", Fruit:""Lemons"",  Price:30,  Purchase:Date(2015,10,3), Tags: [""Mexixo"",""Seedless""]},
+        //                      {Supplier:""Contoso"",  Fruit:""Bananas"", Price:12,  Purchase:Date(2015,10,3), Tags: [""Mexico""]}) As SL,
+        //                      SL.Fruit, Sum( Sort( SL.ThisGroup, Supplier ), Price ) As OnePrice )";
+
+        //    var check = engine.Check(expression, options: _opts);
+        //    var result = check.GetEvaluator().Eval();
+
+        //    Assert.True(check.IsSuccess);
+        //    Assert.IsAssignableFrom<TableValue>(result);
+        //}
```
anderson-joyle, line 46: Remove

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall:
>         public virtual bool IsLambdaParam(int index)

Is it still safe to call this?
Do existing callers to this return wrong result when use with Summarize()? 



---
Refers to: src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs:575 in c7b39cb. [](commit_id = c7b39cb11a501c78c0f6764083a78759792d4dca, deletion_comment = False)

anderson-joyle:
>         public virtual bool IsLambdaParam(int index)

It is. Existing callers will call virtual, with chains back to the original code.
Looking at this again, it looks a better idea to remove the two asserts in the new overload.

---
In reply to: [2072877255](https://github.com/microsoft/Power-Fx/pull/2317#issuecomment-2072877255) [](http://example.com/codeflow?ancestors=2072877255)

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs:575 in c7b39cb. [](commit_id = c7b39cb11a501c78c0f6764083a78759792d4dca, deletion_comment = False)

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4441,7 +4441,7 @@ public override bool PreVisit(CallNode node)
                 Contracts.Assert(carg > 0);
 
                 // The zeroth arg should not be a lambda. Instead it defines the context type for the lambdas.
-                Contracts.Assert(!maybeFunc.IsLambdaParam(0));
+                Contracts.Assert(!maybeFunc.IsLambdaParam(null, 0));
```
CarlosFigueiraMSFT, line 4444: >null [](http://example.com/codeflow?start=58&length=4)

Why not pass the first argument? If we can avoid null values being passed to the virtual, then we should do that.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -657,6 +663,11 @@ public bool ParameterCanBeIdentifier(Features features, int index)
                 paramIdentifierStatus == ParamIdentifierStatus.PossiblyIdentifier;
         }
 
+        public virtual bool ParameterCanBeIdentifier(TexlNode node, int index, Features features)
```
CarlosFigueiraMSFT, line 666: >ParameterCanBeIdentifier [](http://example.com/codeflow?start=28&length=24)

Please add a comment explaining why this is needed instead of just using the original one

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4579,4 +4607,32 @@
     <value>Delegation warning. The result of this argument '{0}' may be truncated for large data sets before being passed to the '{1}' function.</value>
     <comment>Error message when an argument to non-delegable function has possible delegation and resulting rows may be truncated</comment>
   </data>
+  <data name="ErrNeedPrimitive" xml:space="preserve">
+    <value>Expected a text, numeric, boolean, or date/time value.</value>
+    <comment>Error Message.</comment>
```
CarlosFigueiraMSFT, line 4612: >Error Message. [](http://example.com/codeflow?start=13&length=14)

Please expand the comment to something like "Error message shown to the user when a function expects an expression that returns a primitive value but it doesn't."

Similar to the other error messages

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4579,4 +4607,32 @@
...
+  </data>
+  <data name="ErrSummarizeNoGroupBy" xml:space="preserve">
+    <value>Summarize requires at least one grouping column.</value>
+    <comment>Error Message.</comment>
```
CarlosFigueiraMSFT, line 4616: >Message [](http://example.com/codeflow?start=19&length=7)

Please add `{Locked=Summarize}` to the comment to prevent the translators from localizing the function name. Also mention that this is a function name. Similar comment for other error messages as well.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4579,4 +4607,32 @@
...
+  </data>
+  <data name="ErrSummarizeInvalidArg" xml:space="preserve">
+    <value>Summarize arguments must be either existing column names to group by or formulas named with the As keyword.</value>
```
CarlosFigueiraMSFT, line 4619: >As [](http://example.com/codeflow?start=107&length=2)

Add `{Locked=As}` to the comment as we don't want this to be translated.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4579,4 +4607,32 @@
...
+    <comment>Error Message.</comment>
+  </data>
+  <data name="ErrSummarizeDatasourceAsNode" xml:space="preserve">
+    <value>Summarize data source doesn't support As keyword.</value>
+    <comment>Error Message.</comment>
+  </data>
+  <data name="ErrSummarizeThisGroupColumnName" xml:space="preserve">
+    <value>Column 'ThisGroup' conflicts with a reserved name. Use 'RenameColumns' function to pick a different name.</value>
```
CarlosFigueiraMSFT, line 4627: >ThisGroup [](http://example.com/codeflow?start=19&length=9)

Please lock it in the comments. Same for RenameColumns

CarlosFigueiraMSFT, line 4627: >Use 'RenameColumns' function [](http://example.com/codeflow?start=62&length=28)

Suggestion: Use **the** 'RenameColumns' function...

CarlosFigueiraMSFT, line 4627: >RenameColumns [](http://example.com/codeflow?start=67&length=13)

In this error message the function name 'RenameColumns' and keyword 'ThisGroup' are within single quotes; the previous ones are not. We should be consistent.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4579,4 +4607,32 @@
...
+    <comment>Error Message.</comment>
+  </data>
+  <data name="ErrSummarizeDataSourceContainsThisGroupColumn" xml:space="preserve">
+    <value>The datasource contains a column named 'ThisGroup', which conflicts with a reserved name. Use 'RenameColumns' function to pick a different name.</value>
```
CarlosFigueiraMSFT, line 4631: >datasource [](http://example.com/codeflow?start=15&length=10)

`data source` (with space)

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
```diff
@@ -0,0 +1,232 @@
...
+            Contracts.Assert(returnType.IsTable);
+
+            // Source table type
+            DType sourceType = argTypes[0];
+
+            if (args[0] is AsNode)
+            {
+                isValid = false;
+                errors.EnsureError(DocumentErrorSeverity.Severe, args[0], TexlStrings.ErrSummarizeDataSourceScopeNotSupported);
+            }
```
CarlosFigueiraMSFT, line 104: Why can't we use As in the first argument?

anderson-joyle: We decided for this first phase, scoping `arg0` will be blocked.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
```diff
@@ -0,0 +1,232 @@
...
+            : base("Summarize", TexlStrings.AboutSummarize, FunctionCategories.Table, DType.EmptyTable, 0, 2, int.MaxValue, DType.EmptyTable)
+        {
+            SignatureConstraint = new SignatureConstraint(omitStartIndex: 3, repeatSpan: 1, endNonRepeatCount: 1, repeatTopLength: 6);
+            ScopeInfo = new FunctionThisGroupScopeInfo(this);
+        }
+
+        public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
+        {
+            yield return new[] { TexlStrings.SummarizeArg1, TexlStrings.SummarizeArg2, TexlStrings.SummarizeArg3 };
+            yield return new[] { TexlStrings.SummarizeArg1, TexlStrings.SummarizeArg2, TexlStrings.SummarizeArg3, TexlStrings.SummarizeArg2 };
```
CarlosFigueiraMSFT, line 42: > TexlStrings.SummarizeArg3, TexlStrings.SummarizeArg2 [](http://example.com/codeflow?start=86&length=53)

Arg2 should not come after Arg3 - once we're in the aggregation arguments, we cannot go back to the columns.

anderson-joyle: We can. The `Summarize` function is designed to take grouping by and aggregations in any order after `arg0`.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
```diff
@@ -0,0 +1,232 @@
...
+                // All args starting at index 1 need to be identifiers or aggregate functions (wrapped in AsNode node).
+                switch (arg)
+                {
+                    case AsNode nameNode:
+                        existingType = argType;
+                        columnName = nameNode.Right.Name;
+                        break;
+
+                    case FirstNameNode:
+                        if (!TryGetColumnLogicalName(sourceType, context.Features.SupportColumnNamesAsIdentifiers, arg, errors, out columnName, out existingType))
```
CarlosFigueiraMSFT, line 131: >context.Features.SupportColumnNamesAsIdentifiers [](http://example.com/codeflow?start=65&length=48)

For new functions we should always use column names as identifiers - this should be true.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
```diff
@@ -0,0 +1,232 @@
...
+                        atLeastOneGroupByColumn = true;
+                        break;
+
+                    default:
+                        isValid = false;
+                        errors.EnsureError(DocumentErrorSeverity.Severe, arg, TexlStrings.ErrSummarizeInvalidArg);
+                        continue;
+                }
+
+                if (columnName == thisGroup)
```
CarlosFigueiraMSFT, line 146: >columnName [](http://example.com/codeflow?start=20&length=10)

We should also check that the display name is not ThisGroup. For example, if logicalname = cr3er_ThisGroup and displayname=ThisGroup, I don't think this check would catch that.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
```diff
@@ -0,0 +1,232 @@
...
+
+            if (sourceType.Contains(new DName(thisGroup)))
+            {
+                isValid = false;
+                errors.EnsureError(DocumentErrorSeverity.Severe, args[0], TexlStrings.ErrSummarizeDataSourceContainsThisGroupColumn);
+            }
+
+            var atLeastOneGroupByColumn = false;
+
+            for (int i = 1; i < args.Length; i++)
```
CarlosFigueiraMSFT, line 114: >for [](http://example.com/codeflow?start=12&length=3)

If I understand this loop correctly, it will allow something like `Summarize(MyCollection, CountRows(ThisGroup) As count, Brand, Sum(ThisGroup, Price) As TotalPrice, Year)`, which is more than we said we would do on the discussion. We should block that.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
```diff
@@ -0,0 +1,232 @@
...
+                returnType = returnType.Add(ref fError, DPath.Root, columnName, existingType);
+                if (fError)
+                {
+                    isValid = false;
+                    errors.EnsureError(DocumentErrorSeverity.Severe, arg, TexlStrings.ErrColConflict_Name, columnName);
+                    continue;
+                }
+            }
+
+            if (!atLeastOneGroupByColumn)
```
CarlosFigueiraMSFT, line 170: >if (!atLeastOneGroupByColumn) [](http://example.com/codeflow?start=12&length=29)

We check that we have at least one grouping column, but do we check if we have at least one column to group? For example, would `Summarize(MyTable, CountRows(ThisGroup) As Count, Sum(ThisGroup, 1) As SumOfOnes)` work? If it does, it shouldn't.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
```diff
@@ -0,0 +1,232 @@
...
+            Contracts.Assert(returnType.IsTable);
+            return isValid;
+        }
+
+        public override ParamIdentifierStatus GetIdentifierParamStatus(Features features, int index)
+        {
+            if (!features.SupportColumnNamesAsIdentifiers)
+            {
+                return ParamIdentifierStatus.NeverIdentifier;
+            }
```
CarlosFigueiraMSFT, line 185: Newer functions should always use column names as identifiers, regardless of the feature

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
```diff
@@ -209,7 +209,8 @@ internal class BuiltinFunctionsCore
         public static readonly TexlFunction StdevPT = _library.Add(new StdevPTableFunction());
         public static readonly TexlFunction Substitute = _library.Add(new SubstituteFunction());
         public static readonly TexlFunction SubstituteT = _library.Add(new SubstituteTFunction());
-        public static readonly TexlFunction Sum = _library.Add(new SumFunction());
+        public static readonly TexlFunction Sum = _library.Add(new SumFunction());
+        public static readonly TexlFunction Summarize = _library.Add(new SummarizeFunction());
         public static readonly TexlFunction SumT = _library.Add(new SumTableFunction());
```
CarlosFigueiraMSFT, line 214: This should not be added to the core library directly until Power Apps is ready to consume it. Please add it to _featureGateFunctions instead.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.PowerFx.Core.Entities;
 using Microsoft.PowerFx.Core.IR;
 using Microsoft.PowerFx.Core.Utils;
+
 using Microsoft.PowerFx.Interpreter;
```
CarlosFigueiraMSFT, line 13: [minor] Please remove empty line

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -1162,6 +1163,89 @@ private static FormulaValue Refresh(EvalVisitor runner, EvalVisitorContext conte
...
+        {
+            if (args[0] is BlankValue)
+            {
+                return new BlankValue(irContext);
+            }
+
+            if (args[0] is not TableValue tableValue)
+            {
+                return CommonErrors.RuntimeTypeMismatch(irContext);
+            }
```
CarlosFigueiraMSFT, line 1178: This would be overriding any errors that are passed to the function, and we shouldn't. This expression should return a Div0 error:

    Summarize(
        If(1/0<2, [{a:1,b:2},{a:1,b:3}]),
        a,
        CountRows(ThisGroup) As Count
    )

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
```diff
@@ -0,0 +1,232 @@
...
+                }
+
+                if (columnName == thisGroup)
+                {
+                    isValid = false;
+                    errors.EnsureError(DocumentErrorSeverity.Severe, arg, TexlStrings.ErrSummarizeThisGroupColumnName);
+                    continue;
+                }
+                
+                if (!existingType.IsPrimitive)
```
CarlosFigueiraMSFT, line 153: >IsPrimitive [](http://example.com/codeflow?start=34&length=11)

We should have a stricter test here. Types such as Image, Media, Blob are all considered primitives, but we should not be able to group by those.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Summarize.txt
```diff
@@ -0,0 +1,408 @@
...
+
+
+// === Basic grouping with aggregates =================================================
+
+>> Summarize( 
+  Table({Supplier:"Contoso",  Fruit:"Grapes",  Price:220, Purchase:Date(2015,10,1), Tags: ["Red","Seedless"]},
+  {Supplier:"Fabrikam", Fruit:"Lemons",  Price:31,  Purchase:Date(2015,10,1), Tags: ["Colombia"]},
+  {Supplier:"Contoso",  Fruit:"Lemons",  Price:29,  Purchase:Date(2015,10,2), Tags: ["Peru"]},
+  {Supplier:"Contoso",  Fruit:"Grapes",  Price:210, Purchase:Date(2015,10,2), Tags: ["Green","Seedless"]},
+  {Supplier:"Fabrikam", Fruit:"Lemons",  Price:30,  Purchase:Date(2015,10,3), Tags: ["Mexixo","Seedless"]},
```
CarlosFigueiraMSFT, line 22: >Mexixo [](http://example.com/codeflow?start=86&length=6)

[minor] typo (Mexixo ==> Mexico). Same issue on many other tests.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Summarize.txt
```diff
@@ -0,0 +1,408 @@
...
+  {Supplier:"Contoso",  Fruit:"Bananas", Price:12,  Purchase:Date(2015,10,3), Tags: ["Mexico"]}),
+  Fruit, Concat( Sort( Distinct( ThisGroup, Supplier ), Value , SortOrder.Descending ), Value, ", " ) As Suppliers )
+Table({Fruit:"Grapes",Suppliers:"Contoso"},{Fruit:"Lemons",Suppliers:"Fabrikam, Contoso"},{Fruit:"Bananas",Suppliers:"Contoso"})
+
+
+// This formula is in error
+// >> Summarize( Sales, Fruit, Concat( Sort( Distinct( ThisGroup, Supplier ), Supplier ), Supplier, ", " ) As Suppliers )
+// Errors: BUG should be compile time error, not runtime error as currently generated
+
```
CarlosFigueiraMSFT, line 326: Can we fix this bug?

src/tests/Microsoft.PowerFx.Interpreter.Tests/PowerFxEvaluationTests.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
 using System.Drawing;
 using System.Globalization;
+using System.IO;
 using System.Linq;
```
CarlosFigueiraMSFT, line 9: Likely not necessary

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
anderson-joyle: I went ahead and removed this overload. I make changes to PA if necessary.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
anderson-joyle: Are you concerned about the args order? If that's the case lets catch up offline.

jas-valgotar:
We should add some Intellisense tests and decide on the spec what do we want to suggest where.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Summarize.txt
```diff
@@ -0,0 +1,408 @@
...
+// Aggregates last
+>> Summarize(
+  Table({Supplier:"Contoso",  Fruit:"Grapes",  Price:220, Purchase:Date(2015,10,1), Tags: ["Red","Seedless"]},
+  {Supplier:"Fabrikam", Fruit:"Lemons",  Price:31,  Purchase:Date(2015,10,1), Tags: ["Colombia"]},
+  {Supplier:"Contoso",  Fruit:"Lemons",  Price:29,  Purchase:Date(2015,10,2), Tags: ["Peru"]},
+  {Supplier:"Contoso",  Fruit:"Grapes",  Price:210, Purchase:Date(2015,10,2), Tags: ["Green","Seedless"]},
+  {Supplier:"Fabrikam", Fruit:"Lemons",  Price:30,  Purchase:Date(2015,10,3), Tags: ["Mexixo","Seedless"]},
+  {Supplier:"Contoso",  Fruit:"Bananas", Price:12,  Purchase:Date(2015,10,3), Tags: ["Mexico"]}),
+  Supplier, Fruit, Average( ThisGroup, Price ) As Avg_Price )
+Table({Avg_Price:215,Fruit:"Grapes",Supplier:"Contoso"},{Avg_Price:30.5,Fruit:"Lemons",Supplier:"Fabrikam"},{Avg_Price:29,Fruit:"Lemons",Supplier:"Contoso"},{Avg_Price:12,Fruit:"Bananas",Supplier:"Contoso"})
```
jas-valgotar, line 36: >{Avg_Price:215,Fruit:"Grapes",Supplier:"Contoso"} [](http://example.com/codeflow?start=6&length=49)

Should we have same ordering as defined by end user in function? or we consciously decided to reverse it?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Summarize.txt
```diff
@@ -0,0 +1,408 @@
...
+// Groupping or aggregate columns is not primitive type
+>> Summarize(
+  Table({Supplier:"Contoso",  Fruit:"Grapes",  Price:220, Purchase:Date(2015,10,1), Tags: ["Red","Seedless"]},
+  {Supplier:"Fabrikam", Fruit:"Lemons",  Price:31,  Purchase:Date(2015,10,1), Tags: ["Colombia"]},
+  {Supplier:"Contoso",  Fruit:"Lemons",  Price:29,  Purchase:Date(2015,10,2), Tags: ["Peru"]},
+  {Supplier:"Contoso",  Fruit:"Grapes",  Price:210, Purchase:Date(2015,10,2), Tags: ["Green","Seedless"]},
+  {Supplier:"Fabrikam", Fruit:"Lemons",  Price:30,  Purchase:Date(2015,10,3), Tags: ["Mexixo","Seedless"]},
+  {Supplier:"Contoso",  Fruit:"Bananas", Price:12,  Purchase:Date(2015,10,3), Tags: ["Mexico"]}),
+  Tags, CountRows( Distinct( ThisGroup, Supplier ) ) As CountOfSuppliers )
+Errors: Error 638-642: Expected a text, numeric, boolean, or date/time value.|Error 0-9: The function 'Summarize' has some invalid arguments.
```
jas-valgotar, line 212: We should improve the error to say what was the type found was, e.g. here "Expected ... Found Table."

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Summarize.txt
```diff
@@ -0,0 +1,408 @@
...
+>> Summarize(
+  Table({Supplier:"Contoso",  Fruit:"Grapes",  Price:220, Purchase:Date(2015,10,1), Tags: ["Red","Seedless"]},
+  {Supplier:"Fabrikam", Fruit:"Lemons",  Price:31,  Purchase:Date(2015,10,1), Tags: ["Colombia"]},
+  {Supplier:"Contoso",  Fruit:"Lemons",  Price:29,  Purchase:Date(2015,10,2), Tags: ["Peru"]},
+  {Supplier:"Contoso",  Fruit:"Grapes",  Price:210, Purchase:Date(2015,10,2), Tags: ["Green","Seedless"]},
+  {Supplier:"Fabrikam", Fruit:"Lemons",  Price:30,  Purchase:Date(2015,10,3), Tags: ["Mexixo","Seedless"]},
+  {Supplier:"Contoso",  Fruit:"Bananas", Price:12,  Purchase:Date(2015,10,3), Tags: ["Mexico"]}),
+  Fruit, Sum( ThisGroup, Supplier ) As x,  Sum( ThisGroup, Supplier ) As x)
+Errors: Error 706-708: Column name conflict for 'x'.|Error 0-9: The function 'Summarize' has some invalid arguments.
+
```
jas-valgotar, line 234: What happens if Columns are specified twice? Ie 
`Summerize(Table, Fruit, Fruit, Sum())`

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
anderson-joyle: Good point. Originally I was restricting the aggregation by a list supported functions. Probably this is the way to go.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
```diff
@@ -0,0 +1,232 @@
...
+        public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DType[] argTypes, IErrorContainer errors, out DType returnType, out Dictionary<TexlNode, DType> nodeToCoercedTypeMap)
+        {
+            Contracts.AssertValue(args);
+            Contracts.AssertAllValues(args);
+            Contracts.AssertValue(argTypes);
+            Contracts.Assert(args.Length == argTypes.Length);
+            Contracts.AssertValue(errors);
+            Contracts.Assert(MinArity <= args.Length && args.Length <= MaxArity);
+
+            const string thisGroup = "ThisGroup";
```
jas-valgotar, line 92: >const string thisGroup = "ThisGroup"; [](http://example.com/codeflow?start=12&length=37)

maybe define it on DType or TexlFunction?

anderson-joyle: Moved to `FunctionThisGroupScopeInfo`

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Summarize.txt
anderson-joyle: Dont think order is important. Nothing has been changed to intentionally reverse.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Summarize.txt
anderson-joyle: Column conflict error.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
anderson-joyle: Added test cases for it.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
```diff
@@ -0,0 +1,242 @@
...
+                    continue;
+                }
+                
+                // Restricted supported types
+                if (!(existingType == DType.String || 
+                      existingType == DType.Number || 
+                      existingType == DType.Decimal || 
+                      existingType == DType.Boolean || 
+                      existingType == DType.DateTime ||
+                      existingType == DType.Date))
```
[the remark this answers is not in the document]
anderson-joyle: We should. Added.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -646,10 +652,11 @@ public virtual ParamIdentifierStatus GetIdentifierParamStatus(Features features,
         /// <summary>
         /// Returns whether the parameter can be represented by an identifier.
         /// </summary>
+        /// <param name="node">TexlNode. Used by functions that dont have args based on order e.g. Summarize.</param>
```
CarlosFigueiraMSFT, line 655: >dont [](http://example.com/codeflow?start=64&length=4)

[typo] `don't`

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -646,10 +652,11 @@ public virtual ParamIdentifierStatus GetIdentifierParamStatus(Features features,
         /// <summary>
         /// Returns whether the parameter can be represented by an identifier.
         /// </summary>
+        /// <param name="node">TexlNode. Used by functions that dont have args based on order e.g. Summarize.</param>
         /// <param name="features">The features enabled for the expression.</param>
         /// <param name="index">Parameter's index.</param>
         /// <returns>true if the parameter can be an identifier, false otherwise.</returns>
-        public bool ParameterCanBeIdentifier(Features features, int index)
+        public virtual bool ParameterCanBeIdentifier(TexlNode node, int index, Features features)
```
CarlosFigueiraMSFT, line 659: >virtual [](http://example.com/codeflow?start=15&length=7)

We already have one virtual method in TexlFunction about identifier status for parameters (the one above that). Ideally we should not have multiple virtuals for the same scenario. Can we update the GetIdentifierParamStatus to receive the TexlNode, and not make this one virtual?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
```diff
@@ -0,0 +1,245 @@
...
+        public SummarizeFunction()
+            : base("Summarize", TexlStrings.AboutSummarize, FunctionCategories.Table, DType.EmptyTable, 0, 2, int.MaxValue, DType.EmptyTable)
+        {
+            SignatureConstraint = new SignatureConstraint(omitStartIndex: 3, repeatSpan: 1, endNonRepeatCount: 1, repeatTopLength: 6);
+            ScopeInfo = new FunctionThisGroupScopeInfo(this);
+        }
+
+        public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
+        {
+            yield return new[] { TexlStrings.SummarizeArg1, TexlStrings.SummarizeArg2, TexlStrings.SummarizeArg3 };
```
CarlosFigueiraMSFT, line 42: >TexlStrings.SummarizeArg1, TexlStrings.SummarizeArg2 [](http://example.com/codeflow?start=33&length=52)

In the current implementation we can have something like `Summarize(MyTable, Field)`. If we want to support that, then we should give an intellisense suggestion for that as well.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
```diff
@@ -0,0 +1,245 @@
...
+        {
+            Contracts.Assert(arity >= 0);
+
+            if (arity <= 3)
+            {
+                return base.GetSignatures(arity);
+            }
+
+            // Limit the argCount avoiding potential OOM
+            int argCount = arity > SignatureConstraint.RepeatTopLength + SignatureConstraint.EndNonRepeatCount ? SignatureConstraint.RepeatTopLength + SignatureConstraint.EndNonRepeatCount : arity;
```
CarlosFigueiraMSFT, line 59: >arity > SignatureConstraint.RepeatTopLength + SignatureConstraint.EndNonRepeatCount ? SignatureConstraint.RepeatTopLength + SignatureConstraint.EndNonRepeatCount : arity [](http://example.com/codeflow?start=27&length=169)

Suggestion: Math.Min(arity, SignatureConstraint.RepeatTopLength + SignatureConstraint.EndNonRepeatCount)

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
```diff
@@ -0,0 +1,245 @@
...
+            // Source table type
+            DType sourceType = argTypes[0];
+
+            if (args[0] is AsNode)
+            {
+                isValid = false;
+                errors.EnsureError(DocumentErrorSeverity.Severe, args[0], TexlStrings.ErrSummarizeDataSourceScopeNotSupported);
+            }
+
+            if (sourceType.Contains(FunctionThisGroupScopeInfo.ThisGroup))
```
CarlosFigueiraMSFT, line 105: >if (sourceType.Contains(FunctionThisGroupScopeInfo.ThisGroup)) [](http://example.com/codeflow?start=12&length=62)

Do we also need to check if any of the display names is named 'ThisGroup'? In the test SummarizeDisplayNamesTest we check that we cannot use a column whose display name is ThisGroup as a grouping column. But should it work if we didn't use it?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
```diff
@@ -0,0 +1,245 @@
...
+
+                    default:
+                        isValid = false;
+                        errors.EnsureError(DocumentErrorSeverity.Severe, arg, TexlStrings.ErrSummarizeInvalidArg);
+                        continue;
+                }
+
+                if (columnName.Equals(FunctionThisGroupScopeInfo.ThisGroup))
+                {
+                    isValid = false;
```
CarlosFigueiraMSFT, line 157: >isValid = false; [](http://example.com/codeflow?start=20&length=16)

Will this block ever be executed, or is it caught already on line 105?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
```diff
@@ -0,0 +1,245 @@
...
+                      existingType == DType.Number || 
+                      existingType == DType.Decimal || 
+                      existingType == DType.Boolean || 
+                      existingType == DType.Time || 
+                      existingType == DType.OptionSet || 
+                      existingType == DType.OptionSetValue || 
+                      existingType == DType.DateTime ||
+                      existingType == DType.Date))
+                {
+                    errors.EnsureError(DocumentErrorSeverity.Severe, arg, TexlStrings.ErrNeedPrimitive);
```
CarlosFigueiraMSFT, line 173: >EnsureError [](http://example.com/codeflow?start=27&length=11)

Here we are enforcing that the grouping columns are primitive; do we also need to enforce that the grouping expressions return primitive values? For example, I don't think this should work: `Summarize(MyTable, Name, Filter(ThisGroup, Age >= 18))`

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Summarize.txt
```diff
@@ -0,0 +1,460 @@
...
+// Average
+// Max
+// Min
+// StdevP
+// VarP
+// Count
+// CountA
+// CountRows
+// CountIf
+// Invalid functions
```
CarlosFigueiraMSFT, line 12: Can you give context to this comment (or remove if not necessary)?

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Summarize.txt
```diff
@@ -0,0 +1,460 @@
...
+  {Supplier:"Contoso",  Fruit:"Lemons",  Price:29,  Purchase:Date(2015,10,2), Tags: ["Peru"]},
+  {Supplier:"Contoso",  Fruit:"Grapes",  Price:210, Purchase:Date(2015,10,2), Tags: ["Green","Seedless"]},
+  {Supplier:"Fabrikam", Fruit:"Lemons",  Price:30,  Purchase:Date(2015,10,3), Tags: ["Mexico","Seedless"]},
+  {Supplier:"Contoso",  Fruit:"Bananas", Price:12,  Purchase:Date(2015,10,3), Tags: ["Mexico"]}),
+  Fruit, Average( ShowColumns( ThisGroup, Price ), Price ) As Price )
+Table({Fruit:"Grapes",Price:215},{Fruit:"Lemons",Price:30},{Fruit:"Bananas",Price:12})
+
+// NOT READY YET
+// === As on table =================================================
+
```
CarlosFigueiraMSFT, line 423: Remove if not needed

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
CarlosFigueiraMSFT: Nevermind, found a test for that

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
anderson-joyle: It will be executed.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
```diff
@@ -0,0 +1,244 @@
...
+            if (args[0] is AsNode)
+            {
+                isValid = false;
+                errors.EnsureError(DocumentErrorSeverity.Severe, args[0], TexlStrings.ErrSummarizeDataSourceScopeNotSupported);
+            }
+
+            // Check if the source table contains ThisGroup display or logical column
+            if (sourceType.GetAllNames(DPath.Root).Any(tn => 
+                tn.Name == FunctionThisGroupScopeInfo.ThisGroup || 
+                (DType.TryGetDisplayNameForColumn(sourceType, tn.Name, out var innerDisplayName) && innerDisplayName == FunctionThisGroupScopeInfo.ThisGroup)))
```
CarlosFigueiraMSFT, line 110: Suggestion: if(sourceType.Contains(FunctionThisGroupScopeInfo.ThisGroup) || DType.TryGetLogicalNameForColumn(sourceType, FunctionThisGroupScopeInfo.ThisGroup, out var _))

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
MikeStall: is it about scoping arg0? Or scoping identifiers?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
MikeStall: what was resolution?

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ExpressionTestCases/Summarize.txt
jas-valgotar: I feel we should keep the ordering as how user defined it,
Similar to SQL,
`select f1, f2, Count(*) as f3 from table`
Will always have deterministic ordering as per how user ordered.

anderson-joyle: Let me rephrase myself: we print the expression in a deterministic order.
image

What I meant was this is not something I changed in the Summarize function.

jas-valgotar: Ahh! that makes sense!

PR #2319, opened 2024-04-15
Tabular connectors: Add support for SalesForce and clean code

LucGenetier:


src/libraries/Microsoft.PowerFx.Connectors/Environment/SwaggerTabularService.cs
```diff
@@ -24,41 +24,100 @@ public sealed class SwaggerTabularService : CdpTabularService
...
+
+        // Only used for testing
+        internal static (ConnectorFunction s, ConnectorFunction c, ConnectorFunction r, ConnectorFunction u, ConnectorFunction d) GetFunctions(IReadOnlyDictionary<string, FormulaValue> globalValues, OpenApiDocument openApiDocument)
+        {
+            return new SwaggerTabularService(globalValues).GetFunctions(new PowerFxConfig(), openApiDocument);
+        }
+
+        /* Schema [GET] */
+
+        private const string MetadataServiceRegex = @"/\$metadata\.json/datasets/({[^{}]+}(,{[^{}]+})?|default)/tables/{[^{}]+}$";
```
MikeStall, line 64: >default [](http://example.com/codeflow?start=103&length=7)

We can't assume this name.

src/libraries/Microsoft.PowerFx.Connectors/Environment/SwaggerTabularService.cs
```diff
@@ -24,41 +24,100 @@ public sealed class SwaggerTabularService : CdpTabularService
...
+                return new SwaggerTabularService(globalValues).LoadSwaggerAndIdentifyKeyMethods(new PowerFxConfig(), openApiDocument);
+            }
+            catch (Exception ex)
+            {
+                error = ex.Message;
+                return false;
+            }
+        }
+
+        // Only used for testing
```
MikeStall, line 56: >// Only used for testing [](http://example.com/codeflow?start=8&length=24)

Consider an [Obsolete] to warn people off.

src/libraries/Microsoft.PowerFx.Connectors/Environment/SwaggerTabularService.cs
```diff
@@ -24,41 +24,100 @@ public sealed class SwaggerTabularService : CdpTabularService
         private readonly string _connectionId;
         private IReadOnlyList<ConnectorFunction> _tabularFunctions;
         private ConnectorFunction _metadataService;
+        private ConnectorFunction _createItems;
         private ConnectorFunction _getItems;
+        private ConnectorFunction _updateItem;
+        private ConnectorFunction _deleteItem;
```
MikeStall, line 30: who sets these fields?

src/libraries/Microsoft.PowerFx.Connectors/Environment/SwaggerTabularService.cs
LucGenetier: Will add

src/libraries/Microsoft.PowerFx.Connectors/Environment/SwaggerTabularService.cs
LucGenetier: observations tend to prove the opposite
an alternative would be to have a 2 step approach
- first try with dataset param {...} (or params)
- if no function is identified, try with default

WDYT?

src/libraries/Microsoft.PowerFx.Connectors/Environment/SwaggerTabularService.cs
LucGenetier: internal ConnectorFunction DeleteItem => _deleteItem ??= GetFunction...

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/Environment/SwaggerTabularService.cs
```diff
@@ -72,14 +131,25 @@ public async Task InitAsync(PowerFxConfig config, OpenApiDocument openApiDocumen
             }
         }
 
-        // TABLE DATA SERVICE - CREATE
-        // POST: /datasets/{datasetName}/tables/{tableName}/items?api-version=2015-09-01
+        internal bool LoadSwaggerAndIdentifyKeyMethods(PowerFxConfig config, OpenApiDocument openApiDocument, ConnectorLogger logger = null)
+        {
+            var (s, c, r, u, d) = GetFunctions(config, openApiDocument, logger);
+            return s != null && c != null && r != null && u != null && d != null;
```
MikeStall, line 137: > && u != null && d != null [](http://example.com/codeflow?start=54&length=26)

could a readonly table be missing these?

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2323, opened 2024-04-16
Tabular Connectors - Rework SwaggerTabularConnector and various fixes

LucGenetier:
Rework SwaggerTabularConnector function identification logic
Remove all hardcoded names (table, default...)
Rename DataSetName to DatasetName
Add GetGlobalValueNames which returns the list of expected global value names
Move OpenApiDocument to constructor
Better detection of missing global values
CdpTabularConnetor: add Name and DisplayName (only valid after Init)
ConnectorType: Fix DisplayName and add Summary
Add SalesForce tests


LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Connectors.CdpTabularService.get_DataSetName<br />❌CP0002 M:Microsoft.PowerFx.Connectors.SwaggerTabularService.#ctor(System.Collections.Generic.IReadOnlyDictionary{System.String,Microsoft.PowerFx.Types.FormulaValue})

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Connectors.CdpTabularService.get_DataSetName<br />❌CP0002 M:Microsoft.PowerFx.Connectors.SwaggerTabularService.#ctor(System.Collections.Generic.IReadOnlyDictionary{System.String,Microsoft.PowerFx.Types.FormulaValue})

src/libraries/Microsoft.PowerFx.Connectors/Environment/CdpTabularService.cs
```diff
@@ -16,19 +16,27 @@ namespace Microsoft.PowerFx.Connectors
     // Implements CDP protocol for Tabular connectors
     public class CdpTabularService : TabularService
     {
-        public string DataSetName { get; }
+        public string Name { get; private set; }
```
MikeStall, line 19: >Name [](http://example.com/codeflow?start=22&length=4)

Name is too ambigious... 

Can we comment difference between these names (and maybe with a common example of each)

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Connectors/Environment/CdpTabularService.cs
LucGenetier: Will do in future PR

PR #2327, opened 2024-04-18
Remove Obsolete Warnings from LSP to not break dataverse tests proj build

akshar2401:
@gregli-msft Noticed that adding obsolete on Language Server old ctor and old onDataReceived caused compiler errors/warnings in Dataverse tests projects. Removing these warnings if dataverse doesn't want to suppress it,


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/LanguageServer/LanguageServer.cs
```diff
@@ -73,7 +73,13 @@ public class LanguageServer
 
         private readonly IHostTaskExecutor _hostTaskExecutor;
 
-        [Obsolete("Use the constructor with ILanguageServerOperationHandlerFactory")]
```
gregli-msft, line 76: I think we should leave this in place.  I have updated the Dataverse repo to ignore the warning.

PR #2328, opened 2024-04-18
Intellisense: fix duplicate key exception in expression converter

### LINKED ISSUE #2325: Intellisense failure: An item with the same key has already been added.
This was a stack trace we recieved from Plugins folks.
(Related Plugins vso item: https://dynamicscrm.visualstudio.com/OneCRM/_workitems/edit/3947865/)

An item with the same key has already been added.

System.ArgumentException: An item with the same key has already been added.
   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at Microsoft.PowerFx.Core.ExpressionLocalizationHelper.ConvertExpression(String expressionText, RecordType parameters, BindingConfig bindingConfig, INameResolver resolver, IBinderGlue binderGlue, ParserOptions options, Features flags, Boolean toDisplay)
   at Microsoft.PowerFx.Engine.GetDisplayExpression(String expressionText, ReadOnlySymbolTable symbolTable, CultureInfo culture)
   at Microsoft.PowerFx.LanguageServerProtocol.LanguageServer.HandleInitialFixupRequest(String id, String paramsJson)
   at Microsoft.PowerFx.LanguageServerProtocol.LanguageServer.OnDataReceived(String jsonRpcPayload)

LucGenetier:
Fix #2325


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Localization/ExpressionLocalizationHelper.cs
```diff
@@ -38,71 +44,314 @@ internal static string ConvertExpression(string expressionText, RecordType param
                 features: flags);
 
             foreach (var token in binding.NodesToReplace)
-            {
-                worklist.Add(token.Key.Span, TexlLexer.EscapeName(token.Value));
-            }
-
-            return Span.ReplaceSpans(expressionText, worklist);
+            {
+                // [START] TEMPORARY CODE TO HELP RESOLVE ISSUE #2325
```
MikeStall, line 48: > // [START] TEMPORARY CODE TO HELP RESOLVE ISSUE #2325 [](http://example.com/codeflow?start=15&length=54)

can we make this code permanent?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Binding/BinderExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Collections.Generic;
+using Microsoft.PowerFx.Syntax;
+
+namespace Microsoft.PowerFx.Core.Binding
+{
+    internal static class BinderExtensions
+    {
+        public static void AddIfNotSameSpan(this IList<KeyValuePair<Token, string>> nodes, Token key, string value)            
+        {
+            foreach (KeyValuePair<Token, string> kvp in nodes)
```
MikeStall, line 13: 1. This introduces a O[N^2]
2. We should understand _why_ are we adding duplicates? That maybe the real bug.   
This switches to a "first add wins" policy. If the first and second are adding different values, this could pick the wrong one - when the real bug may be that we shouldn't be adding twice at all.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Types/DTypeSpecParser.cs
```diff
@@ -11,8 +11,8 @@ namespace Microsoft.PowerFx.Core.Types
 {
     internal static class DTypeSpecParser
     {
-        private const string _typeEncodings = "?ebnshdipmgo$cDT!*%lLNZPQqVOX-w";
-        private static readonly DType[] _types = new DType[]
+        internal const string _typeEncodings = "?ebnshdipmgo$cDT!*%lLNZPQqVOX-w";
```
MikeStall, line 14: >internal [](http://example.com/codeflow?start=8&length=8)

Why are we increasing visiblity? Is that related to 2328, or another fix?

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -247,7 +247,7 @@ public IExternalControl Control
 
         // Stores tokens that need replacement (Display Name -> Logical Name) for serialization
         // Replace Nodes (Display Name -> Logical Name) for serialization
-        public IList<KeyValuePair<Token, string>> NodesToReplace { get; }
+        public IDictionary<Token, string> NodesToReplace { get; }
```
MikeStall, line 250: >IDictionary [](http://example.com/codeflow?start=15&length=11)

This ensures that conflicts don't crash  - but the fundamental question is *why* are we getting conflicts?

src/libraries/Microsoft.PowerFx.Core/Types/DTypeSpecParser.cs
LucGenetier: This is used in TestUtils.cs at lines 342-343 in order to implement DT2 helper method which can now support display names and allows the test update in TestDelegationValidation.cs to work as expected.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
LucGenetier: With an expression like "SortByColumns(Accounts, 'Account Name', SortOrder.Ascending)" (coming from TestDelegableExpressions_PowerFxV1 test), 'Account Name' second argument is processed in Binder.cs
- first at line 4574, which ends up in calling GetLogicalNodeNameAndUpdateDisplayNames at line 2813
- then, a second time at line 4616 where we call GetLogicalNodeNameAndUpdateDisplayNames again
GetLogicalNodeNameAndUpdateDisplayNames ends up at line 2735 adding the same token (with the same span and string) twice.

The code in ExpressionLocalizationHelper.ConvertExpression uses an algorithm based on a worklist which is a Dictionary and requires unicity of Token's spans.
This fix is simply removing the duplicate spans (that always contain the exact same Token/string couple as this is the same input expression).

MikeStall:
>                 }

This was added here:
[https://github.com/microsoft/Power-Fx/pull/863/files#diff-3b97f603fa1daea008a12cd0138ee5a6bf8270bc878af497c2ff4a272d04a583](https://github.com/microsoft/Power-Fx/pull/863/files#diff-3b97f603fa1daea008a12cd0138ee5a6bf8270bc878af497c2ff4a272d04a583)



---
Refers to: src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs:4621 in 6ad8c58. [](commit_id = 6ad8c584b4498b48578de1a859b453e2dec84ecc, deletion_comment = False)

MikeStall:
>                         args[i].Accept(this);

For example, in SortByColumns(Accounts, 'Account Name', SortOrder.Ascending) 

Here's where we visit first time. 
And then visit a 2nd time below (4608) 


---
Refers to: src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs:4574 in 6ad8c58. [](commit_id = 6ad8c584b4498b48578de1a859b453e2dec84ecc, deletion_comment = False)

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4560,7 +4580,12 @@ public override bool PreVisit(CallNode node)
                         Contracts.Assert(argTypes[i].IsValid);
                     }
                     else
-                    {
+                    {                        
+                        if (args[i] is FirstNameNode firstNameNode)
+                        {
+                            GetLogicalNodeNameAndUpdateDisplayNames(argTypes[0], firstNameNode.Ident, out _);
+                        }
```
MikeStall, line 4587: this is key change

LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #2330, opened 2024-04-18
Jmstall/repro2325

MikeStall:
Hack to help repro #2325


src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -2702,10 +2704,12 @@ public DName GetLogicalNodeNameAndUpdateDisplayNames(DType type, Identifier iden
                     {
                         logicalNodeName = new DName(maybeLogicalName);
                         _txb.NodesToReplace.Add(new KeyValuePair<Token, string>(ident.Token, maybeDisplayName));
+                        _txb._worklist.Add(ident.Token.Span, "x");
```
MikeStall, line 2707: These Adds() will trigger the crash earlier.

Issue #2326, opened 2024-04-17
Should not suggest extra suggestions.


jas-valgotar:
InteelisenseTests.ConnectorIntellisenseTest2 in connector should not suggest User object, when User object is enabled while suggesting field name.

jas-valgotar:
for example `Collect([1,2],{|` would only suggest `Value`, similarly connector should follow the same pattern

PR #2331, opened 2024-04-19
Fix intellisense with Collect function and records

LucGenetier:
Resolves issue https://github.com/microsoft/Power-Fx/issues/2326


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

jas-valgotar (changes requested):
:clock1:

jas-valgotar:
Discussed offline on change needed.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

jas-valgotar:
>             }

curious, why don't we succeed in this if?
e.g. `SQL.ExecuteProcedureV2(""default"", ""default"", ""sp_1"", ",`
The arg type here would be `![p1:n, p2:s]` isn't it?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseData/IntellisenseData.cs:602 in ed4709d. [](commit_id = ed4709df2838ad36efd8607067cd296b298dbf51, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseData/IntellisenseData.cs
```diff
@@ -606,7 +605,7 @@ private bool IsSymbolSuggestableForFunctionArg(TexlFunction function, int argInd
                 // Case 2: Aggregate type with function manipulations or scope info.
                 isSuggestable = symbolType.CoercesTo(argType, false, false, features);
             }
-            else if (argType.IsAggregate && function.ParamTypes.Any(parm => parm.Kind == symbolType.Kind))
+            else if (argType.IsAggregate && function.ParamTypes.Any(parm => parm.Kind == symbolType.Kind) && curNode.Parent is not Syntax.RecordNode)
```
jas-valgotar, line 608: >curNode.Parent is not Syntax.RecordNode [](http://example.com/codeflow?start=109&length=39)

Should we rather check if ChildCount is 0 and argType is agrregate and Function is connector function then directly return false?

Because for connector, Dynamic intellisense would add this aggregate suggestions at IntellisenseHelper.AddTopLevelSuggestionsForCursorType()

LucGenetier: Discussed offline

LucGenetier:
>             }

the arg is ![] as the real type type is only known by dynamic intellisense and here we only get what's in the swagger file

---
In reply to: [2075032137](https://github.com/microsoft/Power-Fx/pull/2331#issuecomment-2075032137) [](http://example.com/codeflow?ancestors=2075032137)

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseData/IntellisenseData.cs:602 in ed4709d. [](commit_id = ed4709df2838ad36efd8607067cd296b298dbf51, deletion_comment = False)

jas-valgotar (approved):

LucGenetier:
✅ No public API change.

PR #2339, opened 2024-04-23
Fix #2241.

### LINKED ISSUE #2241: Conflict detection for default optionsets
We would like to protect against duplicate optionsets, and need to check for optionsets that already exist in PowerFx as defaults. SymbolTable.TryLookupSlot currently returns false for default optionsets, so we need another way to perform this logic. TIA

MikeStall:
Fix #2241 - adds a TryGetType to lookup a variable by name (very useful when it doesn't have a slot)  and return its type. 

This is particularly interesting for enums/optionsets  - which don't have slots (since they don't have per-eval storage). But 
 beware , #2342  - some of the FormulaTypes for enums/optionsets are not correct.


src/tests/Microsoft.PowerFx.Core.Tests/SymbolTableTests.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Security.Cryptography;
```
LucGenetier, line 7: >System.Security.Cryptography [](http://example.com/codeflow?start=6&length=28)

Why is this needed?

src/tests/Microsoft.PowerFx.Core.Tests/SymbolTableTests.cs
```diff
@@ -310,6 +311,49 @@ public void VoidIsNotAllowed()
...
+                DisplayNameUtility.MakeUnique(new Dictionary<string, string> { { "Yes", "Yes1" }, { "No", "No1" } }));
+
+            PowerFxConfig config = new PowerFxConfig();
+
+            bool fOk = config.SymbolTable.IsDefined("os1");
+            Assert.False(fOk);
+
+            config.AddOptionSet(os);
+
+            fOk = config.SymbolTable.IsDefined("os1");
```
LucGenetier, line 330: >fOk = config.SymbolTable.IsDefined("os1"); [](http://example.com/codeflow?start=12&length=42)

I would add a test with "OS1" to confirm that symbols are case sensitive

LucGenetier (dismissed):
:shipit:

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/SymbolTableTests.cs
MikeStall: VS  auto added it - will remove :(

src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
```diff
@@ -345,6 +345,25 @@ public IEnumerable<NamedFormulaType> SymbolNames
...
         }
 
+        /// <summary>
+        /// Returns true if the symbol is defined. 
+        /// This includes both variables (that would be found with <see cref="TryLookupSlot(string, out ISymbolSlot)"/>, as well as constants (like Enums, OptionSets, constants).
+        /// It does not include Functions since those are in a different namespace. 
+        /// </summary>
+        /// <param name="name">logical name of a symbol.</param>
+        /// <returns></returns>
+        public bool IsDefined(string name)
```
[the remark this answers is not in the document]
MikeStall: The problem I'm having is that they want to check things like Color; but that's an Enum and we don't have a public symbol to give back for it.  But we could probably give back FormulaType.

src/tests/Microsoft.PowerFx.Interpreter.Tests/RecalcEngineTests.cs
```diff
@@ -1533,6 +1533,17 @@ public double NextDouble()
...
         }
 
+        [Fact]
+        public void LookupBuiltinOptionSets()
+        {
+            var config = new PowerFxConfig();
+            var engine = new RecalcEngine(config);
+
+            // Builtin enums are on engine.SupportedFunctionm
+            var ok = engine.SupportedFunctions.IsDefined("Color");
```
MikeStall, line 1543: >IsDefined [](http://example.com/codeflow?start=47&length=9)

also give back type

src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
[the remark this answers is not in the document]
MikeStall: Turns out we hit another bug where FormulaType is wrong for both OptionSets (returns as Records) and Enums (returns as Invalid). 
#2342

So we can give type, but should fix this separately.

LucGenetier:
✅ No public API change.

LucGenetier (dismissed):
:shipit:

LucGenetier:
✅ No public API change.

Shpakh-MSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Config/ReadOnlySymbolTable.cs
```diff
@@ -345,6 +345,28 @@ public IEnumerable<NamedFormulaType> SymbolNames
...
 
+        /// <summary>
+        /// Returns true if the symbol is defined. 
+        /// This includes both variables (that would be found with <see cref="TryLookupSlot(string, out ISymbolSlot)"/>, as well as constants (like Enums, OptionSets, constants).
+        /// It does not include Functions since those are in a different namespace. 
+        /// </summary>
+        /// <param name="name">logical or display name of a symbol.</param>
+        /// <param name="type"></param>
+        /// <returns></returns>
+        public bool TryGetType(string name, out FormulaType type)
```
MikeStall, line 356: Due to another bug, this will not give right types for enums for option sets  - see  #2342
But you'd have the same problem if you enumerated the symbols and got the types that way.

LucGenetier (dismissed):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #2340, opened 2024-04-23
Refactor CreateUserDefinedFunctions

adithyaselv:
Minor refactor as a part of introducing user defined types effort.
* Moved private method CreateUserDefinedFunctions from UserDefinitions to UserDefinedFunction and exposing it to hosts. With this hosts can create IR UserDefinedFunction from parser UDF result without being tightly coupled with NamedFormulas or user defined types.
* This will help in introducing Userdefinedtypes . Userdefinedtypes should be processed and resolved first before creating UDF IR as argTypes should be known.

This will be followed by

1.  Retiring  `UserDefinitions.ProcessUserDefinitions` - this method simply parses UserDefinition script and creates UDF IR and NamedFormula ParseResult
2. Introduce NamedTypes, LookupType in INameResolver but support only Primitives. Remove PrimitiveTypeSymbolTable usage , Require INameResolver to create UDFs.
3. Introduce parsing and adding user defined NamedTypes. 
4. Optimize type graph to update incremental.


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -181,6 +182,87 @@ public UserDefinedFunction WithBinding(INameResolver nameResolver, IBinderGlue b
             return func;
         }
 
+        private static readonly ISet<string> _restrictedUDFNames = new HashSet<string> { "Type", "IsType", "AsType" };
+
+        // Helper method to create IR UserDefinedFunctions 
+        internal static IEnumerable<UserDefinedFunction> CreateFunctions(IEnumerable<UDF> uDFs, out List<TexlError> errors)
```
adithyaselv, line 188: Moving private method CreateUserDefinedFunctions here. With this hosts can create UDF from parser udf result without being tightly coupled with NamedFormulas

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -181,6 +182,87 @@ public UserDefinedFunction WithBinding(INameResolver nameResolver, IBinderGlue b
...
+            Contracts.AssertValue(uDFs);
+            Contracts.AssertAllValues(uDFs);
+
+            var userDefinedFunctions = new List<UserDefinedFunction>();
+            var texlFunctionSet = new TexlFunctionSet();
+            errors = new List<TexlError>();
+
+            foreach (var udf in uDFs)
+            {
+                Contracts.Assert(udf.IsParseValid);
```
adithyaselv, line 199: this function expects a Valid parsed UDF, so adding this here

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
```diff
@@ -76,12 +75,13 @@ private bool ProcessUserDefinitions(out UserDefinitionResult userDefinitionResul
...
             // Parser returns both complete & incomplete UDFs, and we are only interested in creating TexlFunctions for valid UDFs. 
-            var functions = CreateUserDefinedFunctions(parseResult.UDFs.Where(udf => udf.IsParseValid), out var errors);
+            var functions = UserDefinedFunction.CreateFunctions(parseResult.UDFs.Where(udf => udf.IsParseValid), out var errors);
 
             errors.AddRange(parseResult.Errors ?? Enumerable.Empty<TexlError>());
+
             userDefinitionResult = new UserDefinitionResult(
                 functions,
-                parseResult.Errors != null ? errors.Union(parseResult.Errors) : errors,
+                errors,
```
adithyaselv, line 84: copying/ union error twice , so removing that.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
MikeStall: This is still private :)
It's public method, but on an internal class.

MikeStall (approved):

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -181,6 +182,92 @@ public UserDefinedFunction WithBinding(INameResolver nameResolver, IBinderGlue b
...
+
+                var udfName = udf.Ident.Name;
+                if (_restrictedUDFNames.Contains(udfName) || texlFunctionSet.AnyWithName(udfName) || BuiltinFunctionsCore._library.AnyWithName(udfName) || BuiltinFunctionsCore.OtherKnownFunctions.Contains(udfName))
+                {
+                    errors.Add(new TexlError(udf.Ident, DocumentErrorSeverity.Severe, TexlStrings.ErrUDF_FunctionAlreadyDefined, udfName));
+                    continue;
+                }
```
CarlosFigueiraMSFT, line 211: Would this prevent us from introducing new functions in the future? For example, if someone has a UDF today called 'Summarize' in an app today, will it break when we introduce a function with that name?

adithyaselv: yes you are right .

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
Shpakh-MSFT: In general, the solution to this is that we always bind to UDF first before built-in functions.  I think there is a issue tracking this somewhere already.  But yes this check needs to be removed.

Shpakh-MSFT (approved):
:shipit:

PR #2341, opened 2024-04-23
Adds a way Hosts can add error without extending Engine.

jas-valgotar:
Hosts can now add custom host specific authoring error without extending the Engine class.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
@@ -94,7 +94,9 @@ namespace Microsoft.PowerFx
         /// <summary>
         /// List of error processor that will add host specific custom errors at the end of the check.
         /// </summary>
-        internal readonly IList<IPostCheckErrorHandler> PostCheckErrorHandlers = new List<IPostCheckErrorHandler>();
+        private readonly IList<IPostCheckErrorHandler> _postCheckErrorHandlers = new List<IPostCheckErrorHandler>();
+
+        public IList<IPostCheckErrorHandler> PostCheckErrorHandlers => _postCheckErrorHandlers;
         
         /// <summary>
         /// Get all functions from the config and symbol tables. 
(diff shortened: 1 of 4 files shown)


src/libraries/Microsoft.PowerFx.Core/Texl/PostCheckErrorHandler.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+using Microsoft.PowerFx.Syntax;
+
+namespace Microsoft.PowerFx
+{
+    public interface IPostCheckErrorHandler
+    {
+        public IEnumerable<ExpressionError> Process(TexlNode root);
```
MikeStall, line 11: >TexlNode [](http://example.com/codeflow?start=52&length=8)

CheckResult? so they can access binding

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -89,7 +89,12 @@ public Engine(PowerFxConfig powerFxConfig)
         /// <summary>
         /// List of transforms to apply to an IR. 
         /// </summary>
-        internal readonly List<Core.IR.IRTransform> IRTransformList = new List<Core.IR.IRTransform>();
+        internal readonly List<Core.IR.IRTransform> IRTransformList = new List<Core.IR.IRTransform>();
+        
+        /// <summary>
+        /// List of error processor that will add host specific custom errors at the end of the check.
+        /// </summary>
+        internal readonly IList<IPostCheckErrorHandler> PostCheckErrorHandlers = new List<IPostCheckErrorHandler>();
```
MikeStall, line 97: >internal [](http://example.com/codeflow?start=8&length=8)

public property?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
jas-valgotar: updated

src/libraries/Microsoft.PowerFx.Core/Texl/PostCheckErrorHandler.cs
jas-valgotar: updated

LucGenetier:
✅ No public API change.

PR #2345, opened 2024-04-24
Retire ProcessUserdefinitions

adithyaselv:
Step 2 of allowing user defined types effort .

Retires `UserDefinitions.ProcessUserdefinitions` as it creates a tight coupling between Parse and processing of all different kinds of Userdefinitions. 

We and our tests assume that client uses this method for processing, hence fixing all the tests as we will not use this moving forward.

Moving `AllowAttributes` transformation to `Parse` as it is a transformation of `ParseResult` and returns a `ParseResult` . Ideally this could be a transformation on NamedFormula `ParseResult` .

This will be followed with  

1. Introduce NamedTypes, LookupType in INameResolver but support only Primitives. Remove PrimitiveTypeSymbolTable usage , Require INameResolver to create UDFs.
2. Introduce parsing and adding user defined NamedTypes.
3. Optimize type graph to update incremental.

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests/ParseTests.cs
+++ b/src/tests/Microsoft.PowerFx.Core.Tests/ParseTests.cs
@@ -829,7 +829,7 @@ namespace Microsoft.PowerFx.Core.Tests
             Assert.False(parseResult.HasErrors);
         }
 
-        internal void TestFormulasParseError(string script)
+        private ParseUserDefinitionResult TestFormulasParseError(string script)
         {
             var parserOptions = new ParserOptions()
             {
@@ -837,6 +837,8 @@ namespace Microsoft.PowerFx.Core.Tests
             };
             var parseResult = UserDefinitions.Parse(script, parserOptions);
             Assert.True(parseResult.HasErrors);
+
+            return parseResult;
         }
 
         [Theory]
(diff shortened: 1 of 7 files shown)


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests/ParseTests.cs
```diff
@@ -862,11 +864,11 @@ public void TestFormulaParseRestart(string script, string key)
             {
                 AllowsSideEffects = false
             };
-            var userDefinitions = UserDefinitions.ProcessUserDefinitions(script, parserOptions, out var userDefinitionResult);
-            Assert.True(userDefinitionResult.HasErrors);
+            var parseResult = UserDefinitions.Parse(script, parserOptions);
+            Assert.True(parseResult.HasErrors);
```
CarlosFigueiraMSFT, line 869: This is exactly the same as TestFormulasParseError; consider calling that (and changing that method to return the parseResult value

src/tests/Microsoft.PowerFx.Core.Tests/ParseTests.cs
```diff
@@ -823,8 +825,8 @@ internal void TestFormulasParseRoundtrip(string script)
             {
                 AllowsSideEffects = false
             };
-            var userDefinitions = UserDefinitions.ProcessUserDefinitions(script, parserOptions, out var userDefinitionResult);
-            Assert.False(userDefinitionResult.HasErrors);
+            var parseResult = UserDefinitions.Parse(script, parserOptions);
+            Assert.False(parseResult.HasErrors);
         }
 
         internal void TestFormulasParseError(string script)
```
CarlosFigueiraMSFT, line 832: >internal [](http://example.com/codeflow?start=8&length=8)

Not part of your changes, but this is only called inside this class, consider changing to private

CarlosFigueiraMSFT:
>         internal void TestFormulasParseRoundtrip(string script)

Not part of your changes, but this is only called inside this class, consider changing to private

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests/ParseTests.cs:822 in 88915d3. [](commit_id = 88915d3663498f4a7e801f89442ef006943f402b, deletion_comment = False)

CarlosFigueiraMSFT (approved):
:shipit:

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #2346, opened 2024-04-24
Propagate display names with ShowColumns

CarlosFigueiraMSFT:
Now that ShowColumns can be used with display names, we should propagate the name mapping so that its output also has both logical and display names.


src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/ShowDropColumnsBase.cs
```diff
@@ -155,6 +164,11 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
             }
 
             returnType = colsToKeep;
+            if (newDisplayNameMapping != null)
```
lesaltzm, line 167: discussed, potential breaking change if an outerscope had the same set of display names, this could change which scope a name gets resolved from

lesaltzm (approved):

LucGenetier:
✅ No public API change.

PR #2347, opened 2024-04-24
Removed resource that should not be translated

CarlosFigueiraMSFT:
We have a string that is a fwlink redirector in our localized resources. Fwlinks should not be translated (it redirects to the user locale, based on the browser's language), so this change removes it.


toshio-msft (approved):

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #2352, opened 2024-04-26
Fix TryGetDsInfo for FirstNameNode, when using a variable

LucGenetier:
There are 3 fixes in this PR
1. in TryGetDsInfo, if FirstNameNode is a variable, we were not returning IExternalDataSource value. We now search in the symbol table for the variable corresponding type to get it. There is no direct test in this repo, this only impacts DV repo where another bug shows up (see below)
2. in PostVisit for AsNode. In CheckAndMarkAsPageable we could set stateful flag to true (line 1227) while reseting the flag at line 3933.
3. in IsBlankFunction, IsRowScopedServerDelegatable was not considering DottedNameNode


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -1224,7 +1224,7 @@ public void CheckAndMarkAsPageable(AsNode node)
                 FlagPathAsAsync(node);
 
                 // Pageable nodes are also stateful as data is always pulled from outside.
-                SetStateful(node, isStateful: true);
+                SetStateful(node.Left, isStateful: true);
```
MikeStall, line 1227: >Left [](http://example.com/codeflow?start=33&length=4)

This is a significant change. Can you elaborate on why this is needed?

MikeStall:
What's an expression that demonstrates the problem this is solving?

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
LucGenetier: I updated PR description and this fix

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/FunctionArgValidators/ConnectedDataSourceInfoArgValidator.cs
```diff
@@ -80,12 +84,19 @@ private bool TryGetDsInfo(FirstNameNode firstName, TexlBinding binding, out IExt
...
-                return formula.TryGetExternalDataSource(out dsInfo);
+                if (firstNameInfo.Data is IExternalNamedFormula formula)
+                {
+                    return formula.TryGetExternalDataSource(out dsInfo);
+                }
+                else if (firstNameInfo.Data is NameSymbol symbol && ((INameResolver)symbol.Owner).Lookup(new DName(symbol.Name), out NameLookupInfo lookupInfo))
+                {
+                    dsInfo = lookupInfo.Type.AssociatedDataSources.FirstOrDefault();
+                    return dsInfo != null;
+                }
```
LucGenetier, line 99: Key change

LucGenetier:
✅ No public API change.

lesaltzm (approved):

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -3923,11 +3926,11 @@ public override void PostVisit(AsNode node)
                 _txb.SetInfo(node, new AsInfo(node, node.Right.Name));
 
                 var left = node.Left;
-                _txb.CheckAndMarkAsPageable(node);
+                bool isNodeStateful = _txb.CheckAndMarkAsPageable(node);
                 _txb.CheckAndMarkAsDelegatable(node);
                 _txb.SetType(node, _txb.GetType(left));
                 _txb.SetSideEffects(node, _txb.HasSideEffects(left));
-                _txb.SetStateful(node, _txb.IsStateful(left));
+                _txb.SetStateful(node, _txb.IsStateful(left) || isNodeStateful);
```
lesaltzm, line 3933: When is an AsNode stateful but the left isn't?

lesaltzm (changes requested):
See comment

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
lesaltzm: If this is valid, worth adding a comment explaining it

LucGenetier: In these few lines, we can set stateful flag twice.
Once in CheckAndMarkAsPageable and once on this line, in an inconsistent manner (true first time, false the second).

An expression like `Sum(Filter(t1, ThisRecord.Price > 50) As X, X.Price)` is when we hit this problem, with t1 being a Dataverse table.
On this code, `node` is `Filter(...) As X` and the left part is CallNode `Filter(...)`.

We start in CheckAndMarkAsPageable, the left node is pageable (line 1219) and as a result we set stateful to true at line 1227 on AsNode.
Then, on line 3933 we want to set stateful state on AsNode, based on the value of left Node (Filter CallNode) which is false.
We fail on Contracts.Assert at line 572.

Thinking of it, a much simpler approach is to simply remove line 3933!
The stateful flag is set by CheckAndMarkAsPageable when we need to and we don't need to do anything more here.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

lesaltzm (approved):

LucGenetier:
✅ No public API change.

PR #2359, opened 2024-04-30
fix batch file

MikeStall:
Use NUGET_PACKAGES environment variable if set.


LucGenetier:
✅ No public API change.

LucGenetier (approved):
:shipit:

PR #2360, opened 2024-04-30
Jmstall/delegation fx1

MikeStall:
1. add delegation interfaces into fx.core so that other libraries can begin to delegate to TableValue.
2. make ConnectorTableValue implement these. 
3. switch to use IReadOnlyCollection . This is a) readonly, b) has a count . Count is important to emphasize that this is not delay execution - because we make a network call, we need to ensure the call happens upfront.


src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorTableValue.cs
```diff
@@ -34,5 +36,32 @@ internal ConnectorTableValue(IRContext irContext)
...
+            var op = parameters.ToOdataParameters();
+            var rows = await _tabularService.GetItemsAsync(services, op, cancel).ConfigureAwait(false);
+
+            return rows;
+        }
+    }   
+
+    internal static class ODataParametersExtensions
+    {
+        public static ODataParameters ToOdataParameters(this DelegationParameters parameters)
```
MikeStall, line 51: Could we eventually get rid of ODataParameters and just use DelegationParameters natively?

src/libraries/Microsoft.PowerFx.Core/Public/Values/DelegationParameters.cs
```diff
@@ -0,0 +1,71 @@
...
+using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Functions;
+
+namespace Microsoft.PowerFx.Types
+{
+    /// <summary>
+    /// For use with <see cref="TableValue"/> and delegation.
+    /// Describe delegation parameters such as filters, column selection, and sorting. 
+    /// </summary>
+    public abstract class DelegationParameters
```
MikeStall, line 23: Delegation analysis logic will convert queries `Filter(Accounts, Age > 30)` into DelegationParameter  and then invoke on IDelegatableTableValue.

LucGenetier:
✅ No public API change.

jas-valgotar (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Values/DelegationParameters.cs
MikeStall: Also tbd - we could consider defining in Interpreter (isntead of Core), since that's reasonably common.

LucGenetier:
✅ No public API change.

Issue #2313, opened 2024-04-10
Suggestions function to REPL


anderson-joyle:
Add a new `Suggestions` (or any better name) pseudo function to display engine suggestions, making use of the cursor pointer (`|`).
Examples can be found here:
https://github.com/microsoft/Power-Fx/blob/ed83c6709c8e633a999ac0b186951d2162be7e6e/src/tests/Microsoft.PowerFx.Core.Tests/IntellisenseTests/SuggestTest.cs#L193

LucGenetier:
https://github.com/microsoft/Power-Fx/pull/2023

PR #2365, opened 2024-05-02
Suggestions pseudo function

anderson-joyle:
Issue #2313.

image

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
@@ -76,6 +76,8 @@ namespace Microsoft.PowerFx
 
         internal Engine Engine => _engine;
 
+        public string Expression => _expression;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="CheckResult"/> class.
         /// Create a "failed" check around a set of errors. 
--- a/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
@@ -76,8 +76,6 @@ namespace Microsoft.PowerFx
 
         internal Engine Engine => _engine;
 
-        public string Expression => _expression;
-
         /// <summary>
         /// Initializes a new instance of the <see cref="CheckResult"/> class.
         /// Create a "failed" check around a set of errors. 
--- a/src/libraries/Microsoft.PowerFx.Repl/PsuedoFunctions.cs
+++ b/src/libraries/Microsoft.PowerFx.Repl/PsuedoFunctions.cs
@@ -54,7 +54,7 @@ namespace Microsoft.PowerFx.Repl
     {
         public async Task ExecuteAsync(CheckResult checkResult, PowerFxREPL repl, CancellationToken cancel)
         {
-            var innerExp = checkResult.Expression;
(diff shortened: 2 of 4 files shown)


LucGenetier:
✅ No public API change.

MikeStall (approved):

src/libraries/Microsoft.PowerFx.Repl/PsuedoFunctions.cs
```diff
@@ -45,4 +47,43 @@ await repl.Output.WriteLineAsync(irText, OutputKind.Repl, cancel)
...
+    /// </summary>
+    public class SuggestionsPseudoFunction : IPseudoFunction
+    {
+        public async Task ExecuteAsync(CheckResult checkResult, string innerExp, PowerFxREPL repl, CancellationToken cancel)
+        {
+            if (innerExp.StartsWith("\"", StringComparison.OrdinalIgnoreCase) && innerExp.EndsWith("\"", StringComparison.OrdinalIgnoreCase))
+            {
+                innerExp = innerExp.Substring(1, innerExp.Length - 2);
+            }
+
```
CarlosFigueiraMSFT, line 62: Why do we need to trim the `"` characters from the expression?

anderson-joyle: The function receives a string literal from REPL: `"Abs(|"`.
We then need to trim the quotes to compile it like a proper expression `Abs(|`.

CarlosFigueiraMSFT: Ah, I see, we need to pass the expression as a string, not as a "normal" expression... Could we do something like `Suggestions(Abs|)`? If not, then what you have is good then.

CarlosFigueiraMSFT: Nevermind, it wouldn't work - as it would be a compile-time error. Please ignore my suggestion :-)

CarlosFigueiraMSFT (approved):

src/libraries/Microsoft.PowerFx.Repl/PsuedoFunctions.cs
```diff
@@ -36,7 +38,7 @@ public interface IPseudoFunction
     /// </summary>
     public class IRPseudoFunction : IPseudoFunction
     {
-        public async Task ExecuteAsync(CheckResult checkResult, PowerFxREPL repl, CancellationToken cancel)
+        public async Task ExecuteAsync(CheckResult checkResult, string innerExp, PowerFxREPL repl, CancellationToken cancel)
```
MikeStall, line 41: nit - can we get innerExpr from checkResult.Text?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -76,6 +76,8 @@ public CheckResult(Engine source)
 
         internal Engine Engine => _engine;
 
+        public string Expression => _expression;
```
MikeStall, line 79: does this already exist on parse?

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #2368, opened 2024-05-03
Fix connector function when returning a blob

LucGenetier:
When a connector function was supposed to return a BlobValue, we were returning a string instead.


MikeStall (approved):

LucGenetier:
✅ No public API change.

Issue #2136, opened 2024-01-10
race in  TexlFunctionSet


MikeStall:
Non-deterministic crash:

```
Array.Copy(Array sourceArray, Int32 sourceIndex, Array destinationArray, Int32 destinationIndex, Int32 length, Boolean reliable)
List`1.CopyTo(T[] array, Int32 arrayIndex)
List`1.ctor(IEnumerable`1 collection)
<>c.<Add>b__29_5(KeyValuePair`2 kvp)
Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
TexlFunctionSet.Add(TexlFunctionSet functionSet)
TexlFunctionSet.ctor(IEnumerable`1 functionSets)
INameResolver.get_Functions()
ComposedReadOnlySymbolTable.LookupFunctions(DPath theNamespace, String name, Boolean localeInvariant)
Visitor.PreVisit(CallNode node)
CallNode.Accept(TexlVisitor visitor)
TexlBinding.Run(IBinderGlue glue, IExternalRuleScopeResolver scopeResolver, DataSourceToQueryOptionsMap queryOptionsMap, TexlNode node, INameResolver resolver, BindingConfig bindingConfig, Boolean updateDisplayNames, DType ruleScope, Boolean forceUpdateDisplayNames, IExternalRule rule, Features features, DelegationHintProvider delegationHintProvider)
Engine.ComputeBinding(CheckResult result)
CheckResult.ApplyBindingInternal()
Engine.Suggest(CheckResult checkResult, Int32 cursorPosition, IServiceProvider services)
IntellisenseTests.ConnectorIntellisenseTest(Int32 responseIndex, Int32 queryIndex, String expression, String expectedSuggestions) line 124
```

MikeStall:
fixed with a lock.

PR #2369, opened 2024-05-04
Add extra guard to protect symbol table 

MikeStall:
Caller is responsible Symbol tables shouldn't be mutated while they're being read.
See #1551 for details on rule about threading. 

This doesn't fix the problem, but gives us a better (and more deterministic) exception when it happens. 

Possibly related to tracking down #2136


jas-valgotar (approved):

LucGenetier:
✅ No public API change.

PR #2373, opened 2024-05-06
Deep mutation of records

gregli-msft:
Adds support for deep mutation through the Set function.  `Set( a, {b:1} )` doesn't change and will assign a reference to the record `{b:1}` to the variable `a`.  After `a` has been established, it now supports mutation through `Set( a.b, 2 )`.  This is an optional facility that is disabled by default and there are no semantic changes.  When creating variables the host has the option to enable this with `SymbolProperties.CanSetMutate`.

This also fixes two related bugs.  First, really deep mutations were not being properly copied on write, where `t` should be unaffected by the Patch https://github.com/microsoft/Power-Fx/issues/2450:
```
>> Set( deep, [[[[1,2,3],[4,5,6]]]] )
deep: [[[[1, 2, 3], [4, 5, 6]]]]

>> Set( t, deep )
t: [[[[1, 2, 3], [4, 5, 6]]]]

>> Patch( First(First(First(deep).Value).Value).Value, {Value:2}, {Value:99} )
{Value:99}

>> deep
[[[[1, 99, 3], [4, 5, 6]]]]

>> t
[[[[1, 99, 3], [4, 5, 6]]]]
```

And this should not return a casting error, instead returning Blank() https://github.com/microsoft/Power-Fx/issues/2451:
```
>> Patch(Blank(),Blank(),Blank())
Unable to cast object of type 'Microsoft.PowerFx.Types.BlankType' to type 'Microsoft.PowerFx.Types.RecordType'.
```


LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.TableValue.Index(System.Int32)<br />❌CP0002 M:Microsoft.PowerFx.Types.TableValue.TryGetIndex(System.Int32,Microsoft.PowerFx.Types.DValue{Microsoft.PowerFx.Types.RecordValue}@)CP0002 M:Microsoft.PowerFx.RecalcEngine.UpdateVariable(System.String,Microsoft.PowerFx.Types.FormulaValue)

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -126,8 +126,13 @@ internal abstract class TexlFunction : IFunction
...
         public virtual bool HasPreciseErrors => false;
 
-        // Returns true if the function will mutate the value of argument 0, as is the case with Patch, Collect, Remove, etc.
-        public virtual bool MutatesArg0 => false;
+        /// <summary>
+        /// Returns true if the function will mutate the value of the first argument, as is the case of Patch, Collect, Remove, etc.
+        /// Set can also mutate, but needs to make a decision based on the argument's node.
+        /// This function covers both CanMutate and CanSetMutate scenarios which is checked in CheckTypes/CheckSemantics.
+        /// </summary>
+        /// <param name="arg"></param>
```
CarlosFigueiraMSFT, line 134: Which arg is this? Is it the arg0?

gregli-msft: I've debated this one.  It is arg0, because this is only called for arg0.  But that is a little unclear, so I decided to add the index to the signature and make it clearer, which also adds slightly more expressiveness (any argument can mutate).

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.TableValue.Index(System.Int32)<br />❌CP0002 M:Microsoft.PowerFx.Types.TableValue.TryGetIndex(System.Int32,Microsoft.PowerFx.Types.DValue{Microsoft.PowerFx.Types.RecordValue}@)CP0002 M:Microsoft.PowerFx.RecalcEngine.UpdateVariable(System.String,Microsoft.PowerFx.Types.FormulaValue)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.TableValue.Index(System.Int32)<br />❌CP0002 M:Microsoft.PowerFx.Types.TableValue.TryGetIndex(System.Int32,Microsoft.PowerFx.Types.DValue{Microsoft.PowerFx.Types.RecordValue}@)CP0002 M:Microsoft.PowerFx.RecalcEngine.UpdateVariable(System.String,Microsoft.PowerFx.Types.FormulaValue)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.TableValue.Index(System.Int32)<br />❌CP0002 M:Microsoft.PowerFx.Types.TableValue.TryGetIndex(System.Int32,Microsoft.PowerFx.Types.DValue{Microsoft.PowerFx.Types.RecordValue}@)CP0002 M:Microsoft.PowerFx.RecalcEngine.UpdateVariable(System.String,Microsoft.PowerFx.Types.FormulaValue)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.TableValue.Index(System.Int32)<br />❌CP0002 M:Microsoft.PowerFx.Types.TableValue.TryGetIndex(System.Int32,Microsoft.PowerFx.Types.DValue{Microsoft.PowerFx.Types.RecordValue}@)CP0002 M:Microsoft.PowerFx.RecalcEngine.UpdateVariable(System.String,Microsoft.PowerFx.Types.FormulaValue)

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
```diff
@@ -99,30 +99,72 @@ public virtual int Count()
         /// <summary>
         /// Lookup the record at the given 1-based index, or return an error value if out of range.
         /// </summary>
-        /// <param name="index1">1-based index.</param>
+        /// <param name="index1">1-based index.</param>
+        /// <param name="mutationCopy">copies the element, and the table entry pointing to it, when in a mutation context.</param>
         /// <returns>The record or an errorValue.</returns>
-        public DValue<RecordValue> Index(int index1)
+        public DValue<RecordValue> Index(int index1, bool mutationCopy = false)
```
MikeStall, line 105: > Index(int index1, bool mutationCopy = false) [](http://example.com/codeflow?start=34&length=45)

Nit: Can you add this an overload? 

In C#, adding parameters (even with default values) is a *binary* breaking change.   That's what the Public API changes are alerting about.

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolProperties.cs
```diff
@@ -17,13 +17,29 @@ namespace Microsoft.PowerFx
...
         public bool CanSet { get; init; }
 
-        // Can this symbol be passed to Collect, Patch() for mutation 
+        // Can this symbol be passed to Collect, Patch, Remove, etc. for mutation. For example: Collect( t, {b:3} )
         // This only applies to symbols of a tabular type. 
         public bool CanMutate { get; init; }
 
-        // Is Copyable?
+        // Can this symbol be deep mutated with Set. For example: Set( a.b, 3 )
+        public bool CanSetMutate { get; init; }
```
MikeStall, line 43: This is good.  This is the right place to track this.

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
```diff
@@ -99,30 +99,72 @@ public virtual int Count()
...
             }
 
             return DValue<RecordValue>.Of(ArgumentOutOfRange(IRContext));
         }
 
-        // Index() does standard error messaging and then call TryGetIndex().
-        protected virtual bool TryGetIndex(int index1, out DValue<RecordValue> record)
-        {
+        // Index() does standard error messaging and then call TryGetIndex().
+        // By default, can't mutate through a table, must be overrideen for specific table types.
```
MikeStall, line 116: >overrideen [](http://example.com/codeflow?start=61&length=10)

nit: typo

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -165,8 +165,20 @@ private async Task<FormulaValue> TryHandleSet(CallNode node, EvalVisitorContext
...
-
+            var newValue = await arg1.Accept(this, context).ConfigureAwait(false);
+
+            if (arg0.IRContext.IsMutation && arg0 is RecordFieldAccessNode rfan)
+            {
+                var arg0value = await rfan.From.Accept(this, context).ConfigureAwait(false);
+                
+                if (arg0value is RecordValue rv)
+                {
+                    ((IMutationCopyField)rv).ShallowCopyFieldInPlace(rfan.Field);
```
MikeStall, line 176: >((IMutationCopyField)rv) [](http://example.com/codeflow?start=20&length=24)

How do we know this cast will always succeed?

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
```diff
@@ -99,30 +99,72 @@ public virtual int Count()
...
+        protected virtual bool TryGetIndex(int index1, out DValue<RecordValue> record, bool mutationCopy = false)
+        {
             var index0 = index1 - 1;
-            if (index0 < 0)
+            if (index0 < 0 || mutationCopy)
```
MikeStall, line 120: > || mutationCopy [](http://example.com/codeflow?start=26&length=16)

Why this?  

Is there any user expression that can hit this path? (does binder prevent this from happening). If the only way we can get here is is developer didn't implement the virtuals properly, should we throw a hard exception instead?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -1161,7 +1161,14 @@ private static FormulaValue Refresh(EvalVisitor runner, EvalVisitorContext conte
 
         public static FormulaValue PatchRecord(IRContext irContext, FormulaValue[] args)
         {
-            return CompileTimeTypeWrapperRecordValue.AdjustType((RecordType)FormulaType.Build(irContext.ResultType._type), (RecordValue)MutationUtils.MergeRecords(args).ToFormulaValue());
+            if (irContext.ResultType is BlankType)
```
MikeStall, line 1164: >if (irContext.ResultType is BlankType) [](http://example.com/codeflow?start=12&length=38)

I assume this is fix for #2451?

jas-valgotar:
>             }

Does this mean, If symbol was already preset, `newVarProps` will not be honored? 

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs:151 in 71e8530. [](commit_id = 71e8530f0fb0e7139232023476caaa62258ee75d, deletion_comment = False)

jas-valgotar:
>             }

Does this mean, If symbol was already preset, `newVarProps` will not be honored? 

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs:151 in 71e8530. [](commit_id = 71e8530f0fb0e7139232023476caaa62258ee75d, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
gregli-msft: Added as overload.

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolProperties.cs
gregli-msft: :)

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
gregli-msft: Fixed.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
gregli-msft: The check for arg0.IRContext.IsMutation above this wouldn't be true if this wasn't the case.  But it is fine to add another check to be sure.

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
gregli-msft: Good idea, I've changed it to an exception (in the mutationCopy argument overload version).

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
gregli-msft: Correct.

gregli-msft:
Jas, correct, these flags are only applied on create, just as the type of the variable is only applied on create.

gregli-msft:
Squashed changes into https://github.com/microsoft/Power-Fx/pull/2457, please review there instead.

PR #2377, opened 2024-05-07
Fix UDF with invalid body to be ParseInvalid

adithyaselv:
We considered UDF with parse failures in body to be parse valid UDFs which leads to issues when binding the body of UDFs with errors. 

In This PR we check if there are any new errors after parsing body and set parseValid appropriately.


rick-nguyen (approved):

LucGenetier:
✅ No public API change.

PR #2379, opened 2024-05-08
Moved test from Rand.txt that depended on number being decimal

CarlosFigueiraMSFT:
One test from Rand.txt assumes that numbers are decimal; this moves it to a separate file.


LucGenetier:
✅ No public API change.

anderson-joyle (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2381, opened 2024-05-09
Proper Target Relationship in AddRelatedColumns Public Method

jack-work:
The other methods in the query options classes use the datasource associated with the target relationship to resolve relationship metadata, where AddRelatedColumns uses the parent datasource, which refers to the source datasource (e.g. SourceDatasource.TargetDataSourceRelatedColumn.Name).  This PR fixes that.


sagar-bakliwal (approved):

src/libraries/Microsoft.PowerFx.Core/Entities/QueryOptions/ExpandQueryOptions.cs
```diff
@@ -90,16 +90,28 @@ public bool AddSelect(string selectColumnName)
...
-            if (!(ExpandInfo.ParentDataSource is IExternalCdsDataSource))
+            // This statement captures the properties of ExpandInfo and returns false iff all the captured values are
+            // non-null and of the correct type if specified
+            if (ExpandInfo is not {
+                    Identity: { } identity,
+                    ParentDataSource: IExternalCdsDataSource {
+                        Document: { GlobalScope: { } scope },
+                        DatasetName: { } datasetName } })
             {
+                Contracts.Assert(false, "The expand object lacks necessary dependencies.");
```
lesaltzm, line 101: Should this be a Check? Asserts in pfx only throw in pfx debug builds, not in host/canvas debug builds

lesaltzm (dismissed):

src/libraries/Microsoft.PowerFx.Core/Entities/QueryOptions/ExpandQueryOptions.cs
jack-work: My thinking was that we just wouldn't throw.  We weren't throwing before if the parent data source is not iexternalcds data source.  We know this manifests as a runtime error.  Perhaps it's preferable to a crash?  What's the design philosophy here?

sagar-bakliwal (approved):

lesaltzm (approved):

adithyaselv:
merged on behalf of @jack-work upon his request.

PR #2382, opened 2024-05-09
Fix strongly typed option sets issues in Power Apps

gregli-msft:
Option sets coming from Dataverse that are backed by a number are not handled correctly in Power Apps, a pre-V1 host.  An error is produced for comparisons where there was not one previously.  A few checks for pre-V1 were missing:
- Concatenate function (but not the operator)
- Numerical comparisons between option sets

Updating the test suite revealed that the interpreter's implementation of Concatenate didn't properly account for non-string option sets, which has also been fixed.

One reason we didn't detect these problems earlier is that we use EnumSymbol as a proxy for OptionSet, which is usually pretty good, but in the comparison case it led tests to fail but the product to fail.  There will be a separate task to expand OptionSet to support more data types and update the tests.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -843,7 +844,8 @@ private static BinderCheckTypeResult CheckComparisonArgTypesCore(IErrorContainer
                         }
                     };
                 }
-                else if (typeRight.Kind == DKind.OptionSetValue && typeRight.OptionSetInfo.CanCompareNumeric &&
+                else if (typeRight.Kind == DKind.OptionSetValue &&
+                        (!usePowerFxV1CompatibilityRules || typeRight.OptionSetInfo.CanCompareNumeric) &&
                         (!usePowerFxV1CompatibilityRules || typeRight.OptionSetInfo.CanCoerceFromBackingKind || typeRight.OptionSetInfo.CanCoerceToBackingKind) &&
                         (typeLeft.Kind == DKind.Number || typeLeft.Kind == DKind.Decimal))
```
Shpakh-MSFT, line 850: This can be a helper as it shares the same logic with above code.

Shpakh-MSFT:
>                     if (typeLeft.OptionSetInfo.CanCompareNumeric)

Should this be also checked with usePowerFxV1CompatibilityRules flag?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs:810 in 1cc00eb. [](commit_id = 1cc00eb187e577bdd228ebfc054e9602aae25369, deletion_comment = True)

Shpakh-MSFT:
>                     nonOptionSetArg = true;

This could be also now due to !PowerFxV1CompatibilityRules. Is this correct?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Concatenate.cs:85 in 1cc00eb. [](commit_id = 1cc00eb187e577bdd228ebfc054e9602aae25369, deletion_comment = False)

Shpakh-MSFT:
>             {

Can we make this logic closer to what it is today when PowerFxV1CompatibilityRules is false?


if(!context.Features.PowerFxV1CompatibilityRules)
{
    returnType = ReturnType;
    return fArgsValid;
}

Ideally we should just keep the two logics separately.
For example,

If(context.Features.PowerFxV1CompatibilityRules)
   return CheckTypesInternalWithV1(...
else
   return CheckTypesInternalLegacy(...


  

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Concatenate.cs:91 in 1cc00eb. [](commit_id = 1cc00eb187e577bdd228ebfc054e9602aae25369, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
gregli-msft: Done.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Concatenate.cs
```diff
@@ -62,48 +62,64 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
...
+            }
+            else
             {
-                concatAsString = true;
+                for (var i = 0; i < count; i++)
+                {
+                    fArgsValid &= CheckType(context, args[i], argTypes[i], DType.String, errors, ref nodeToCoercedTypeMap);
+                }
+
+                returnType = DType.String;
```
Shpakh-MSFT, line 115: > DType.String [](http://example.com/codeflow?start=28&length=13)

ReturnType

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Concatenate.cs
```diff
@@ -62,48 +62,64 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
...
+                    concatAsString = true;
                 }
-            }
 
-            if (nonOptionSetArg && argOptionSet != null && !argOptionSet.OptionSetInfo.CanCoerceFromBackingKind)
+                // All option sets can coerce to string and be concatenated as strings, which is the default returnType.
+                // If there was only one option set type seen, and it has CanConcatenatedStronglyTyped set,
+                // and either no string values were present or CanCoerceFromBackingKind is set, then Concatenate returns the enum type.
+                // This behavior should mach that of the & operator (BinderUtils, CheckBinaryOpCore).
+                returnType = argOptionSet != null && !concatAsString ? argOptionSet : DType.String;
```
Shpakh-MSFT, line 106: > DType.String [](http://example.com/codeflow?start=85&length=13)

ReturnType

Shpakh-MSFT (approved):
:shipit:

Shpakh-MSFT (approved):
:shipit:

PR #2385, opened 2024-05-09
Fix strongly typed option sets issues in Power Apps (#2382) t24052

gregli-msft:
Option sets coming from Dataverse that are backed by a number are not handled correctly in Power Apps, a pre-V1 host. An error is produced for comparisons where there was not one previously. A few checks for pre-V1 were missing:
- Concatenate function (but not the operator)
- Numerical comparisons between option sets

Updating the test suite revealed that the interpreter's implementation of Concatenate didn't properly account for non-string option sets, which has also been fixed.

One reason we didn't detect these problems earlier is that we use EnumSymbol as a proxy for OptionSet, which is usually pretty good, but in the comparison case it led tests to fail but the product to fail. There will be a separate task to expand OptionSet to support more data types and update the tests.


LucGenetier:
✅ No public API change.

PR #2390, opened 2024-05-10
More improvements to strongly typed option sets

gregli-msft:
Addresses customer reported bug in pre-V1 where option set value = Blank() resulted in an error.  General remedy is to return the pre-V1 code to its original state and have a large if around V1 vs. pre-V1 code.  And more tests.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
@@ -800,7 +800,7 @@ namespace Microsoft.PowerFx.Core.Binding
             var usePowerFxV1CompatibilityRules = features.PowerFxV1CompatibilityRules;
 
             // Special cases for comparing option set values
-            // Without StronglyTypedBuiltinEnums, option sets are convereted to their backing kind before getting here in Visitor.Visit(FirstNameNode)
+            // Without StronglyTypedBuiltinEnums, enums are convereted to their backing kind before getting here in Binder.cs, Visitor.Visit(FirstNameNode)
             if (typeLeft.Kind == DKind.OptionSetValue || typeRight.Kind == DKind.OptionSetValue)
             {
                 if (usePowerFxV1CompatibilityRules)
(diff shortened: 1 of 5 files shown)


LucGenetier:
✅ No public API change.

Shpakh-MSFT (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -800,14 +800,39 @@ private static BinderCheckTypeResult CheckComparisonArgTypesCore(IErrorContainer
...
-                if (typeLeft.Kind == DKind.OptionSetValue && typeRight.Kind == DKind.OptionSetValue
-                    && typeLeft.Accepts(typeRight, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: usePowerFxV1CompatibilityRules))
+                if (usePowerFxV1CompatibilityRules)
                 {
-                    if (!usePowerFxV1CompatibilityRules || typeLeft.OptionSetInfo.CanCompareNumeric)
+                    // Comparing values from two different option set values is not supported
+                    if (typeLeft.Kind == DKind.OptionSetValue && typeRight.Kind == DKind.OptionSetValue
+                        && typeLeft.Accepts(typeRight, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: usePowerFxV1CompatibilityRules))
+                    {
+                        if (!usePowerFxV1CompatibilityRules || typeLeft.OptionSetInfo.CanCompareNumeric)
```
CarlosFigueiraMSFT, line 812: >!usePowerFxV1CompatibilityRules ||  [](http://example.com/codeflow?start=28&length=35)

Can be removed, will always be false (because of `if` on 806)

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
gregli-msft: I'll an issue to resolve this later.

Issue #2370, opened 2024-05-04
Smmarize should have a compile time error for Identifier as Name


gregli-msft:
We should expliciltly check for and block all forms of "Identifier As Name", including if the identifier is in the row scope of the table or is a global variable outside.  We do this so we can support group by column renaming in the future. 

```
>> Summarize( [1,2,3], Value)
[1, 2, 3]

>> Summarize( [1,2,3], Value, Average( ThisGroup, Value ) As A )

  A   Value
 === =======
  1   1
  2   2
  3   3

>> Summarize( [1,2,3], Value, Value As A )  // this should have a compile time error, certainly not this runtime error
One or more errors occurred. (Unable to cast object of type 'Microsoft.PowerFx.Types.DecimalType' to type 'Microsoft.PowerFx.Types.TableType'.)

>> Set( Global, 1 )
Global: 1

>> Summarize( [1,2,3], Value, Global As A ) // this should have a compile time error too

  A   Value
 === =======
  1   1
  1   2
  1   1
```

PR #2395, opened 2024-05-10
Summarize should have a compile time error for Identifier as Name

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2370.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
@@ -126,7 +126,7 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
                 {
                     case AsNode asNode:
 
-                        if (asNode.Left is FirstNameNode || asNode.Left is DottedNameNode)
+                        if (asNode.Left is FirstNameNode)
                         {
                             isValid = false;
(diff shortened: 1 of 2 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
```diff
@@ -124,9 +124,16 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
                 // All args starting at index 1 need to be identifiers or aggregate functions (wrapped in AsNode node).
                 switch (arg)
                 {
-                    case AsNode nameNode:
+                    case AsNode asNode:
+
+                        if (asNode.Left is FirstNameNode || asNode.Left is DottedNameNode)
```
MikeStall, line 129: Double checking that we want to also fail on DottedName node?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
anderson-joyle: Turns out we shouldn't fail. Thanks.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

gregli-msft (approved):

PR #2396, opened 2024-05-10
Update Invalid Use of Dot error

lesaltzm:
Updates error to "The '.' operator cannot be used on <TypeName> values"

The error shows up in the case of
"the thing on the left of “.” is a number/string/other thing you can’t use “.” on ever, excluding polymorphic because it's weird"


LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
🚢

LucGenetier:
✅ No public API change.

PR #2398, opened 2024-05-13
Add tests specific for the Decimal and Float functions

CarlosFigueiraMSFT:
There are no tests specifically for those functions; this splits the Value tests and adds those.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

gregli-msft (approved):

PR #2400, opened 2024-05-13
Unifying clear function

anderson-joyle:
Part of https://github.com/microsoft/Power-Fx/issues/1550.


src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Clear.cs
```diff
@@ -0,0 +1,133 @@
...
+using Microsoft.PowerFx.Syntax;
+
+namespace Microsoft.PowerFx.Core.Texl.Builtins
+{
+    // !!!TODO 
+    //[RequiresErrorContext]
+    //[ComponentVisibilityRestriction(
+    //    DisableForDataComponent = true,
+    //    DisableForCommanding = true,
+    //    SuppressIntellisenseForComponent = true)]
```
anderson-joyle, line 21: Reminder to remove after PA work.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalDataSource.cs
```diff
@@ -21,6 +21,8 @@ internal interface IExternalDataSource : IExternalEntity, IExternalPageableSymbo
 
         bool IsWritable { get; }
 
+        bool IsClearable { get; }
```
MikeStall, line 24: >bool [](http://example.com/codeflow?start=8&length=4)

breaking internal

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
```diff
@@ -228,4 +228,33 @@ public async Task<FormulaValue> InvokeAsync(FormulaType irContext, FormulaValue[
...
+        public async Task<FormulaValue> InvokeAsync(FormulaType irContext, FormulaValue[] args, CancellationToken cancellationToken)
+        {
+            if (args[0] is ErrorValue errorValue)
+            {
+                return errorValue;
+            }
+
+            if (args[0] is BlankValue)
+            {
+                return irContext == FormulaType.Void ? FormulaValue.NewVoid() : FormulaValue.NewBlank(FormulaType.Boolean);
```
MikeStall, line 244: >FormulaType.Void  [](http://example.com/codeflow?start=36&length=17)

This raises an question  - how do voids propagate errors today?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Clear.cs
```diff
@@ -0,0 +1,126 @@
...
+            DType dataSourceType = argTypes[0];
+
+            if (binding.Features.PowerFxV1CompatibilityRules && dataSourceType.IsTable && dataSourceType.Kind != DKind.ObjNull)
+            {
+                base.ValidateArgumentIsMutable(binding, args[0], errors);
+            }
+
+            if (binding.EntityScope != null && 
+                binding.EntityScope.TryGetDataSource(args[0], out IExternalDataSource dataSourceInfo) &&
+                !dataSourceInfo.IsClearable)
```
MikeStall, line 84: >          !dataSourceInfo.IsClearable) [](http://example.com/codeflow?start=6&length=38)

This is good - 
Today, we totally block Clear() in certain scenarios to prevent use on wrong sources; but this means we could allow the function, and the enforcement now happens here. 

That would let makers use Clear() on safe targets (like in-memory collections) while blocking on unsafe targets (like DV tables)

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
anderson-joyle: For in memory collections, the only way for `Clear` function to return an error is if the argument is an `ErrorValue`. Other implementations (like PFx.Dataverse) will handle differently.

src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalDataSource.cs
MikeStall, line 24: @LucGenetier  - this is what caused your PR failure.  Just rebase on latest to pick this up. 
https://github.com/microsoft/Power-Fx/pull/2425/checks?check_run_id=25337498835

PR #2402, opened 2024-05-14
Validate type literal parse

adithyaselv:
Since we use generic expression parser for parsing type expression within type literal node, we consider an expression like `Type(5+5)` to be parsevalid.

This PR 
1. Adds a validator to allow only valid type expressions.
2. Simplifies `DTypeVisitor` to use `DefaultVisitor` 
3. overrides exception when visiting `Type(Type(..))`
4. Adds TexlPretty for `TypeLiteralNode` to enable ToString.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/TypeLiteralNode.cs
```diff
@@ -47,5 +51,148 @@ public override void Accept(TexlVisitor visitor)
 
         /// <inheritdoc />
         public override NodeKind Kind => NodeKind.TypeLiteral;
+
+        internal static bool IsValid(TypeLiteralNode typeNode, out List<TexlError> errors)
```
Shpakh-MSFT, line 55: >static [](http://example.com/codeflow?start=17&length=6)

Why static?

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/TypeLiteralNode.cs
```diff
@@ -47,5 +51,148 @@ public override void Accept(TexlVisitor visitor)
...
+            typeNode.TypeRoot.Accept(validator);
+            errors = validator.Errors;
+            return !errors.Any();
+        }
+
+        private class Validator : IdentityTexlVisitor
+        {
+            private readonly List<TexlError> _errors;
+
+            internal List<TexlError> Errors => _errors;
```
Shpakh-MSFT, line 67: >List [](http://example.com/codeflow?start=21&length=4)

IEnumerable<TexlError>

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/TypeLiteralNode.cs
```diff
@@ -47,5 +51,148 @@ public override void Accept(TexlVisitor visitor)
...
+
+            public Validator()
+            {
+                _errors = new List<TexlError>();
+            }
+
+            // Valid Nodes
+            public override void Visit(FirstNameNode node)
```
Shpakh-MSFT, line 75: >public override void Visit(FirstNameNode node) [](http://example.com/codeflow?start=12&length=46)

Why override it?

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/TypeLiteralNode.cs
```diff
@@ -47,5 +51,148 @@ public override void Accept(TexlVisitor visitor)
...
+            {
+            }
+
+            public override bool PreVisit(RecordNode node)
+            {
+                return true;
+            }
+
+            public override void PostVisit(RecordNode node)
```
Shpakh-MSFT, line 84: > public override void PostVisit(RecordNode node) [](http://example.com/codeflow?start=11&length=48)

It's no different than base.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4410,8 +4410,8 @@
     <comment>Warning for makers that a certain capability should be avoided.</comment>
   </data>
   <data name="ErrTypeLiteral_InvalidTypeDefinition" xml:space="preserve">
-    <value>Type literal declaration was invalid.</value>
-    <comment>Error message when type binding is invalid.</comment>
+    <value>Type literal declaration is invalid. The highlighted expression can not be used in Type definition.</value>
```
CarlosFigueiraMSFT, line 4413: >highlighted [](http://example.com/codeflow?start=52&length=11)

This is assuming that this error will be seen in a formula bar (where we can have highlighting), but that may not always be the case. It can be seen in the app checker, or in the CLI, where there are no highlighting capabilities.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/TypeLiteralNode.cs
adithyaselv: started without this but added override explicitly to ensure & show that we be doing anything in postvisit for these nodes

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -317,7 +317,16 @@ private ParseUserDefinitionResult ParseUDFsAndNamedFormulas(string script, Parse
                         var result = ParseExpr(Precedence.None);
                         if (result is TypeLiteralNode typeLiteralNode)
                         {
-                            definedTypes.Add(new DefinedType(thisIdentifier.As<IdentToken>(), typeLiteralNode));
+                            if (TypeLiteralNode.IsValid(typeLiteralNode, out var err))
```
CarlosFigueiraMSFT, line 320: >err [](http://example.com/codeflow?start=81&length=3)

Suggestion: since this can be multiple errors, call `errors` instead of `err` (which suggests singular)

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/TypeLiteralNode.cs
CarlosFigueiraMSFT, line 55: >List [](http://example.com/codeflow?start=67&length=4)

IEnumerable<TextError>

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/TypeLiteralNode.cs
Shpakh-MSFT: If you are going to implement all methods here then you can just extend from TexlVisitor.

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/TypeLiteralNode.cs
```diff
@@ -47,5 +51,148 @@ public override void Accept(TexlVisitor visitor)
...
+            {
+            }
+
+            public override bool PreVisit(TableNode node)
+            {
+                if (node.ChildNodes.Count > 1)
```
Shpakh-MSFT, line 90: >     if (node.ChildNodes.Count > 1) [](http://example.com/codeflow?start=11&length=35)

Is there a reason we don't need this check for record node?

src/libraries/Microsoft.PowerFx.Core/Syntax/Visitors/DefaultVisitor.cs
```diff
@@ -101,6 +101,16 @@ public override TResult Visit(TableNode node, TContext context)
...
+        {
+            return Default;
+        }
+
+        public override TResult Visit(StrInterpNode node, TContext context)
+        {
+            return Default;
+        }
+
+        public override TResult Visit(TypeLiteralNode node, TContext context)
```
Shpakh-MSFT, line 113: >TResult Visit(TypeLiteralNode node, [](http://example.com/codeflow?start=24&length=35)

Don't you need to update TexlFunctionalVisitor definition for it?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Syntax/Visitors/DefaultVisitor.cs
adithyaselv: resolved: clarified offline

LucGenetier:
✅ No public API change.

Shpakh-MSFT (dismissed):
:shipit:

Shpakh-MSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

Issue #2406, opened 2024-05-16
LSP crash:


MikeStall:
On .Net 4.6.2. 

System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at System.Linq.Enumerable.WhereListIterator`1.MoveNext()
   at System.Linq.Enumerable.<DefaultIfEmptyIterator>d__93`1.MoveNext()
   at Microsoft.PowerFx.Intellisense.IntellisenseResult..ctor(IIntellisenseData data, List`1 suggestions, IEnumerable`1 possibleOverloads)
   at Microsoft.PowerFx.Intellisense.Intellisense.Finalize(IIntellisenseContext context, IntellisenseData intellisenseData, CultureInfo culture)

PR #2407, opened 2024-05-16
stronger typing to help with #2406.

MikeStall:
See #2406. This doesn't fix issue, but makes types more strongly typed to help reduce the window. 

Avoid mutable fields.


LucGenetier:
✅ No public API change.

jas-valgotar (approved):

LucGenetier:
✅ No public API change.

PR #2412, opened 2024-05-16
Expose void return type and add reserved UDF names

rick-nguyen:
With imperative functions, add support for a void UDF type. Also some functions should not be allowed for UDF names like Set, Collect, etc.


src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -184,7 +185,11 @@ public UserDefinedFunction WithBinding(INameResolver nameResolver, IBinderGlue b
             return func;
         }
 
-        private static readonly ISet<string> _restrictedUDFNames = new HashSet<string> { "Type", "IsType", "AsType" };
+        private static readonly ISet<string> _restrictedUDFNames = new HashSet<string>
+        {
+            "Type", "IsType", "AsType", "Set", "Collect",
+            "ClearCollect", "UpdateContext", "Navigate", "Select", "Reset",
```
adithyaselv, line 191: https://github.com/microsoft/Power-Fx/blob/bd033f54ee476ea78467eb44ec8e1ff057ff299c/src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs#L25

dont we already block these names here ?

rick-nguyen: Changed so we would allow UDFs to be named after existing functions besides this restricted list

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -184,7 +185,11 @@ public UserDefinedFunction WithBinding(INameResolver nameResolver, IBinderGlue b
             return func;
         }
 
-        private static readonly ISet<string> _restrictedUDFNames = new HashSet<string> { "Type", "IsType", "AsType" };
+        private static readonly ISet<string> _restrictedUDFNames = new HashSet<string>
+        {
+            "Type", "IsType", "AsType", "Set", "Collect",
+            "ClearCollect", "UpdateContext", "Navigate", "Select", "Reset",
+        };
```
Shpakh-MSFT, line 193: Can we make this list attribute based on TexlFunction level? And then filter it using that property of that function instead of static list?

MikeStall: There is something nice about seeing the list outright here. 
Even if we do get rid of this list from the product, it would be good to have this explicit list in the tests so that we can verify. 

It would also be good if the analysis piece used this same list and enforced things.

src/tests/Microsoft.PowerFx.Core.Tests/UserDefinedFunctionTests.cs
```diff
@@ -519,6 +519,30 @@ public void TestUDFsWithRestrictedTypes(string script)
             errors.AddRange(parseResult.Errors ?? Enumerable.Empty<TexlError>());
 
             Assert.Contains(errors, x => x.MessageKey == "ErrUDF_InvalidReturnType" || x.MessageKey == "ErrUDF_InvalidParamType");
+        }
+
+        [Theory]
+        [InlineData("Set(x: Number):Number = x + 1;", true)]
+        [InlineData("Count():Number = 5;", false)]
+        public void TestUDFsReservedNames(string script, bool expectedError)
```
MikeStall, line 527: Since adding to this reserved list is a breaking change, can we have the list of reserverd names hardcoded in the test?  Ie, so if somebody adds a new reserved name, they need to update the test to allow it (this will help raise awareness that it's a breaking change. ) This is similar to what w edo with public surface tests.

rick-nguyen: Good call

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
Shpakh-MSFT: Adding it in tests make sense but in product this needs to be property based on TexlFunction to be more explicit.

CarlosFigueiraMSFT: ☝️ We should not be hardcoding the function names like this. This should be a property of the function itself, so it can be queried from other places such as in analysis.

adithyaselv: Wouldn't we still end up some list here? I see atleast one possibility which is `Type` and it is not a TexlFunction .

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -182,15 +183,20 @@ public UserDefinedFunction WithBinding(INameResolver nameResolver, IBinderGlue b
...
-        }
-
-        private static readonly ISet<string> _restrictedUDFNames = new HashSet<string> { "Type", "IsType", "AsType" };
+        }
+
+        // Adding a restricted UDF name is a breaking change, this test will need to be updated and a conversion will be needed for existing scenarios
+        private static readonly ISet<string> _restrictedUDFNames = new HashSet<string>
+        {
+            "Type", "IsType", "AsType", "Set", "Collect", "ClearCollect",
+            "UpdateContext", "Navigate", "Select", "Reset",
```
CarlosFigueiraMSFT, line 192: >"Select", "Reset" [](http://example.com/codeflow?start=41&length=17)

Curious: why are Select and Reset restricted? Does analysis use these functions for something special?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using Microsoft.CodeAnalysis;
```
CarlosFigueiraMSFT, line 8: Probably not needed

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -200,14 +206,15 @@ public static IEnumerable<UserDefinedFunction> CreateFunctions(IEnumerable<UDF>
...
             foreach (var udf in uDFs)
             {
                 Contracts.Assert(udf.IsParseValid);
 
                 var udfName = udf.Ident.Name;
-                if (_restrictedUDFNames.Contains(udfName) || texlFunctionSet.AnyWithName(udfName) || BuiltinFunctionsCore._library.AnyWithName(udfName) || BuiltinFunctionsCore.OtherKnownFunctions.Contains(udfName))
+                if (_restrictedUDFNames.Contains(udfName) || texlFunctionSet.AnyWithName(udfName) ||
+                    nameResolver.Functions.WithName(udfName).Any(func => func.IsRestrictedUDFName))
                 {
                     errors.Add(new TexlError(udf.Ident, DocumentErrorSeverity.Severe, TexlStrings.ErrUDF_FunctionAlreadyDefined, udfName));
```
CarlosFigueiraMSFT, line 219: >ErrUDF_FunctionAlreadyDefined [](http://example.com/codeflow?start=98&length=29)

In a host such as CLI or MCS which doesn't have the concept of UpdateContext or Navigate, will the error message "Function {0} is already defined." make sense to them?

CarlosFigueiraMSFT:
>         // Primitive types - Keeping it same as PrimitiveTypeSymbolTable

I know this is not part of your changes, but does this comment still make sense? I couldn't find any definition for "PrimitiveTypeSymbolTable"

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs:77 in ca62f6f. [](commit_id = ca62f6fb17e1306949d3b690797050c27a27925d, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
rick-nguyen: Added a separate error for restricted names

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -182,15 +182,20 @@ public UserDefinedFunction WithBinding(INameResolver nameResolver, IBinderGlue b
...
-
-        private static readonly ISet<string> _restrictedUDFNames = new HashSet<string> { "Type", "IsType", "AsType" };
+        }
+
+        // Adding a restricted UDF name is a breaking change, this test will need to be updated and a conversion will be needed for existing scenarios
+        private static readonly ISet<string> _restrictedUDFNames = new HashSet<string>
+        {
+            "Type", "IsType", "AsType", "Set", "Collect", "ClearCollect",
+            "UpdateContext", "Navigate", "Select", "Reset",
+        };
```
Shpakh-MSFT, line 192: You need to update all these functions to have override IsRestrictedUDFName.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
CarlosFigueiraMSFT: Good point on Type - it's not a function. Is it a keyword now that we have UDTs? If so, should we forbid all keywords as well?

Shpakh-MSFT (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -200,17 +205,23 @@ public static IEnumerable<UserDefinedFunction> CreateFunctions(IEnumerable<UDF>
...
                 Contracts.Assert(udf.IsParseValid);
 
                 var udfName = udf.Ident.Name;
-                if (_restrictedUDFNames.Contains(udfName) || texlFunctionSet.AnyWithName(udfName) || BuiltinFunctionsCore._library.AnyWithName(udfName) || BuiltinFunctionsCore.OtherKnownFunctions.Contains(udfName))
+                if (texlFunctionSet.AnyWithName(udfName))
                 {
                     errors.Add(new TexlError(udf.Ident, DocumentErrorSeverity.Severe, TexlStrings.ErrUDF_FunctionAlreadyDefined, udfName));
                     continue;
+                }
+                else if (_restrictedUDFNames.Contains(udfName) ||
```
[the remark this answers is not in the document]
Shpakh-MSFT: Discussed offline. Thanks!

LucGenetier:
✅ No public API change.

adithyaselv (approved):
:shipit:

PR #2427, opened 2024-05-24
Handle Entity types in UDFs

adithyaselv:
As we plan to support complex types from Datasource in UDFs, we want to handle recursively defined Entity types in UDFs.

We do not need to check for restricted types in DataEntity and metadata types that have expandinfo.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -142,7 +142,7 @@ public void CheckTypesOnDeclaration(CheckTypesContext context, DType actualBodyR
                 }
                 else
                 {
-                    binding.ErrorContainer.EnsureError(DocumentErrorSeverity.Severe, UdfBody, TexlStrings.ErrUDF_ReturnTypeDoesNotMatch);
+                    binding.ErrorContainer.EnsureError(DocumentErrorSeverity.Severe, UdfBody, TexlStrings.ErrUDF_ReturnTypeDoesNotMatch, ReturnType.GetKindString(), actualBodyReturnType.GetKindString());
```
adithyaselv, line 145: Making this consistent with Invalid argument error

LucGenetier:
✅ No public API change.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4075,7 +4075,7 @@
     <comment>This error message shows up when an invalid type is used</comment>
   </data>
   <data name="ErrUDF_ReturnTypeDoesNotMatch" xml:space="preserve">
-    <value>The stated function return type does not match the return type of the function body.</value>
+    <value>The stated function return type '{0}' does not match the return type of the function body '{1}'.</value>
     <comment>This error message shows up when expected return type does not match with actual return type</comment>
```
CarlosFigueiraMSFT, line 4079: Please add to the comment the description of the placeholders, and an example. Something like `The argument '{0}' and '{1}' will be replaced with data types. For example, "The stated function return type 'Number' does not match the return type of the function body 'Person'"`

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -310,7 +310,7 @@ internal static bool IsRestrictedType(FormulaType ft)
         {
             Contracts.AssertValue(ft);
 
-            if (ft is AggregateType aggType)
+            if (!ft._type.HasExpandInfo && ft is AggregateType aggType)
```
Shpakh-MSFT, line 313: >!ft._type.HasExpandInfo [](http://example.com/codeflow?start=16&length=23)

Can you add comment on why we added this check?

Shpakh-MSFT (approved):
:shipit:

Shpakh-MSFT (approved):
:shipit:

PR #2431, opened 2024-05-24
IntelliSense fix for UDFs

adithyaselv:
We need partial parse invalid UDF definitions for Intellisense to suggest options, This PR adds logic to return parse invalid UDF IRs which can be used by Intellisense .


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

rick-nguyen (approved):

LucGenetier:
✅ No public API change.

PR #2432, opened 2024-05-24
fix bug in summarize.

MikeStall:
This was failing:
`Summarize(incident, priority, CountRows(ThisGroup) As Count)`

because priority was binding as an expression, not as a column identifier.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
@@ -206,7 +206,17 @@ namespace Microsoft.PowerFx.Core.Texl.Builtins
 
         public override ParamIdentifierStatus GetIdentifierParamStatus(TexlNode node, Features features, int index)
         {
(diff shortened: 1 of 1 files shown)


CarlosFigueiraMSFT:
Can we add a test that would fail without this change?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Summarize.cs
```diff
@@ -205,8 +205,18 @@ public override bool CheckTypes(CheckTypesContext context, TexlNode[] args, DTyp
         }
 
         public override ParamIdentifierStatus GetIdentifierParamStatus(TexlNode node, Features features, int index)
-        {
-            return index > 0 && node is not AsNode ? ParamIdentifierStatus.PossiblyIdentifier : ParamIdentifierStatus.NeverIdentifier;
+        {
```
CarlosFigueiraMSFT, line 208: Can we add a test that would fail without this change?

MikeStall: yes, @anderson-joyle  is looking at that.

PR #2433, opened 2024-05-24
Jmstall/summarize fix2

MikeStall:


src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -1181,12 +1181,6 @@ public static async ValueTask<FormulaValue> Summarize(EvalVisitor runner, EvalVi
                 return CommonErrors.RuntimeTypeMismatch(irContext);
             }
 
-            // Nothing to do for empty tables
-            if (tableValue.Rows.Count() == 0)
```
MikeStall, line 1185: This call triggers an extra network call to query the table. 

But we shoudln't need it - algorithm should be correct it degenerative case of 0 rows.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

nguhoa (approved):

PR #2439, opened 2024-05-27
Make Texl builtin function statics not depend on order of execution

LucGenetier:
static initialization is deterministic per the C# spec but this code is confusing


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
```diff
@@ -23,243 +23,475 @@ internal class BuiltinFunctionsCore
             "RecordInfo", "Relate", "RemoveAll", "RemoveIf", "RequestHide", "Reset", "ResetForm", "Revert", "SaveData", "ScanBarcode", "Select", "SetFocus",
             "SetProperty", "ShowColumns", "State", "SubmitForm", "TraceValue", "Ungroup", "Unrelate", "Update", "UpdateContext", "UpdateIf", "User", "Validate", "ValidateRecord", "ViewForm",
             "Collect", "Clear", "Patch", "Remove", "ClearCollect", "Set"
+        };
+
+        internal static readonly TexlFunction[] __library = new TexlFunction[]
```
MikeStall, line 28: >__library [](http://example.com/codeflow?start=48&length=9)

nit, single underscore prefix (not double)

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
```diff
@@ -23,243 +23,475 @@ internal class BuiltinFunctionsCore
...
-        public static readonly TexlFunction VarP = _library.Add(new VarPFunction());
-        public static readonly TexlFunction VarPT = _library.Add(new VarPTableFunction());
-        public static readonly TexlFunction Weekday = _library.Add(new WeekdayFunction());
-        public static readonly TexlFunction WeekdaysLong = _library.Add(new WeekdaysLongFunction());
-        public static readonly TexlFunction WeekdaysShort = _library.Add(new WeekdaysShortFunction());
-        public static readonly TexlFunction WeekNum = _library.Add(new WeekNumFunction());
-        public static readonly TexlFunction With = _library.Add(new WithFunction());
-        public static readonly TexlFunction Year = _library.Add(new YearFunction());
+        public static readonly TexlFunction Abs = __library[0];
+        public static readonly TexlFunction AbsT = __library[1];
```
MikeStall, line 252: >__library[1] [](http://example.com/codeflow?start=51&length=12)

referencing by index seems extremely brittle. 
Especially if we expect to keep this alpha sorted.

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
LucGenetier: this was on purpose to avoid a name conflict
I'll rename to avoid the double underscore if that's a concern

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
LucGenetier: if you have an idea, it's always welcome
I'll use an enum to make things a little less brittle but that's unavoidable and a small cost compared to what this PR is adressing (random failures)

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
```diff
@@ -23,243 +23,709 @@ internal class BuiltinFunctionsCore
...
+            VarPT,
+            Weekday,
+            WeekdaysLong,
+            WeekdaysShort,
+            WeekNum,
+            With,
+            Year
+        }
+
+        public static readonly TexlFunction Abs = _functionArray[(int)Library.Abs];
```
CarlosFigueiraMSFT, line 471: >(int)Library.Abs [](http://example.com/codeflow?start=65&length=16)

This is dangerous, as if it assumes that the enum and the arrays will be kept in sync. If we want to reference those in this syntax, _functionArray should be a _functionDictionary<Library, TexlFunction> instead.

MikeStall, line 471: >_functionArray [](http://example.com/codeflow?start=50&length=14)

How is this PR solving the race?

MikeStall: although we could mitigate that concern with a unit test?

Dictionary is good too, but then Abs is a dictionary lookup, whereas just a direct reference to the texlFunction.

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
```diff
@@ -23,243 +23,709 @@ internal class BuiltinFunctionsCore
...
+            new BooleanLFunction(),
+            new BooleanLFunction_T(),
+            new SummarizeFunction(),
+        };
 
         // _featureGateFunctions functions, not present in all platforms
-        internal static readonly TexlFunctionSet _featureGateFunctions = new TexlFunctionSet();
+        internal static readonly TexlFunctionSet _featureGateFunctions = new TexlFunctionSet(_featureGateFunctionArray);
+
+        private enum FeaatureGateFunctions
```
anderson-joyle, line 705: >Feaature [](http://example.com/codeflow?start=21&length=8)

typo

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
```diff
@@ -1,8 +1,10 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
```
CarlosFigueiraMSFT, line 7: Needed?

CarlosFigueiraMSFT (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
```diff
@@ -23,250 +24,956 @@ internal class BuiltinFunctionsCore
...
-        public static readonly TexlFunction WeekNum = _library.Add(new WeekNumFunction());
-        public static readonly TexlFunction With = _library.Add(new WithFunction());
-        public static readonly TexlFunction Year = _library.Add(new YearFunction());
-
-        // Don't add new functions here, follow alpha order
+        };
+
+        static BuiltinFunctionsCore()
+        {
+            Dictionary<Library, TexlFunction> functionArray = new Dictionary<Library, TexlFunction>()
```
CarlosFigueiraMSFT, line 31: >functionArray [](http://example.com/codeflow?start=46&length=13)

[minor] This is not really an array, not sure we should call it so

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
LucGenetier: I'm not sure there is any real better solution
I've used a dictionary

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
LucGenetier: Nope, removed

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
LucGenetier: let's rename

CarlosFigueiraMSFT (approved):
:shipit:

MikeStall:
This 4 has duplicates - can we reduce?

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
```diff
@@ -23,250 +24,956 @@ internal class BuiltinFunctionsCore
...
-        public static readonly TexlFunction WeekdaysLong = _library.Add(new WeekdaysLongFunction());
-        public static readonly TexlFunction WeekdaysShort = _library.Add(new WeekdaysShortFunction());
-        public static readonly TexlFunction WeekNum = _library.Add(new WeekNumFunction());
-        public static readonly TexlFunction With = _library.Add(new WithFunction());
-        public static readonly TexlFunction Year = _library.Add(new YearFunction());
-
-        // Don't add new functions here, follow alpha order
+        };
+
+        static BuiltinFunctionsCore()
```
MikeStall, line 29: This is repeating each function 4 times.  Can we reduce?

MikeStall (changes requested):
:clock1:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #2445, opened 2024-05-30
Add ability to override output when calling the connector

jerry-santana:
This is useful in a situation when the output type is dynamic and is not available in the swagger


LucGenetier (approved):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier (approved):

PR #2446, opened 2024-05-30
Updated option set testing

gregli-msft:
Fixes a bug in Canvas apps coercing an option set value to number.  There were missing checks for pre-V1 semantics in DType.cs.

But this bug, along with the other recent fixes, is because we were not testing with actual option sets, instead relying on enums since those can be built for all the backing kinds.  In general, enums and option sets aren't that different in V1, but they are very different in pre-V1 which is what Canvas uses today.  This has been corrected in this PR.  We now run tests with the test enums registered as option sets instead, mimicing the semantics in Canvas for Dataverse number and Boolean option sets.  These new typed option sets are not exposed generally, as we should look at that implementation more closely, today it translates between string values more than it needs to.

By doing this testing, another bug was found in the interpeter, where we weren't back converting boolean values to the option set, for example in `Text( If( false, OptionSet.Value, true ) )`.  C# now has the same semantics as Canvas for Booleans.  Number and other backing kinds may also want this behavior long term, but since those aren't yet supported we have time to decide how to handle that scenario.  Without any changes, a `CalculatedOptionSetValue` will be returned for the label if coerced to a string, but the value itself will be correct.


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
```diff
@@ -1070,7 +1072,21 @@ public static OptionSetValue NumberToOptionSet(IRContext irContext, NumberValue[
...
         {
-            return new OptionSetValue(irContext, "CalculatedOptionSetValue", (OptionSetValueType)irContext.ResultType, (bool)args[0].Value);
+            var optionSetInfo = ((OptionSetValueType)irContext.ResultType)._type.OptionSetInfo;
+
+            if (optionSetInfo != null && optionSetInfo.BackingKind == Core.Types.DKind.Boolean && optionSetInfo.OptionNames.Count() == 2)
+            {
+                foreach (var option in optionSetInfo.OptionNames)
+                {
+                    optionSetInfo.TryGetValue(option, out var value);
+                    if ((bool)value.ExecutionValue == (bool)args[0].Value)
```
Shpakh-MSFT, line 1082: >value [](http://example.com/codeflow?start=30&length=5)

Any null check here?

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/StronglyTypedEnum_TestEnums.txt
```diff
@@ -273,6 +273,24 @@ Errors: Error 24-28: Invalid argument type (Enum (TestYeaNay)). Expecting a Enum
...
+>> Text( If( false, TestYeaNay.Yea, false ) )
+"Nay"
+
+>> If( false, TestNumberCompareNumericCoerceFrom.V, 10 )
+TestNumberCompareNumericCoerceFrom.CalculatedOptionSetValue
+
+>> Text( If( false, TestNumberCompareNumericCoerceFrom.V, 10 ) )
+"CalculatedOptionSetValue"
+
+>> If( false, TestNumberCompareNumericCoerceFrom.V, 11 )
```
Shpakh-MSFT, line 288: > If( false, TestNumberCompareNumericCoerceFrom.V, 11 ) [](http://example.com/codeflow?start=2&length=54)

Is this test different than one on line 282?

Shpakh-MSFT (dismissed):
:shipit:

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUnary.cs
gregli-msft: Good call, checked the result of TryGetValue first.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/StronglyTypedEnum_TestEnums.txt
gregli-msft: Added a comment.  10 is in the optionset, while 11 is not.  It is possible we could map the 10 to the option set value instead of using CalculatedOptionSetValue in the future.

LucGenetier:
✅ No public API change.

Shpakh-MSFT (approved):
:shipit:

PR #2453, opened 2024-06-03
Adds expression re-writing capabilities in LSP InitialFixup method.

jas-valgotar:


src/libraries/Microsoft.PowerFx.Core/Texl/IExpressionRewriter.cs
```diff
@@ -0,0 +1,16 @@
...
+using Microsoft.PowerFx.Syntax;
+
+namespace Microsoft.PowerFx
+{
+    /// <summary>
+    /// Expression re-writer interface. This will be leveraged by InitalFixup method to rewrite the expression in LSP.
+    /// </summary>
+    public interface IExpressionRewriter
+    {
+        public string Process(CheckResult check);
```
MikeStall, line 14: >string [](http://example.com/codeflow?start=15&length=6)

can you comment this further - 
Assuming this returns an expression for rewritten result?

MikeStall:
>         private readonly Dictionary<string, CodeFixHandler> _handlers = new Dictionary<string, CodeFixHandler>();

Can fixup be here (instead of engine)? 

EditorContextScope already has other state, like code fix handlers. 


---
Refers to: src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Public/EditorContextScope.cs:42 in 30406d6. [](commit_id = 30406d6a7951c5d9b955098b09d5d3bea2d5d3a9, deletion_comment = False)

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #2454, opened 2024-06-03
Adds ability to verify if FirstNameNode is valid or not

jas-valgotar:


src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -706,6 +706,21 @@ public CheckContextSummary ApplyGetContextSummary()
...
+        public bool IsNodeInvalid(FirstNameNode node)
+        {
+            if (node == null)
+            {
+                throw new ArgumentNullException(nameof(node));
+            }
+
+            var fnInfo = this.ApplyBindingInternal().GetInfo(node);
+
+            return fnInfo.Kind == BindKind.Min && fnInfo.Data == null;
```
Shpakh-MSFT, line 722: I didn't understand the logic here. What does valid firstnamenode mean here?

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -706,6 +706,21 @@ public CheckContextSummary ApplyGetContextSummary()
 
             return summary;
         }
+
+        /// <summary>
+        /// Used to check if given FirstNameNode(Variable) node is a valid variable or not.
+        /// </summary>
+        public bool IsNodeInvalid(FirstNameNode node)
```
anderson-joyle, line 713: >IsNodeInvalid [](http://example.com/codeflow?start=20&length=13)

nit: for code reading purposes, would be better to make this method resulting in a positive answer e.g. `IsNodeValid`?

anderson-joyle:
>         public async Task TestInitialFixupRewriter()

Not related to your PR, but it is possible to comment what this test is about?

---
Refers to: src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/LanguageServiceProtocol/RedesignedLanguageServerTests/InitialFixupTests.cs:66 in 673505b. [](commit_id = 673505bb4c23281a461edfb6ffdad0b6e127a618, deletion_comment = False)

Issue #2451, opened 2024-06-02
Patch(Blank(),Blank(),Blank()) results in cast error


gregli-msft:
Should return Blank(), but instead returns:

```
>> Patch(Blank(),Blank(),Blank())
Unable to cast object of type 'Microsoft.PowerFx.Types.BlankType' to type 'Microsoft.PowerFx.Types.RecordType'.
```

MikeStall:
fixed

Issue #2450, opened 2024-06-02
Deep mutations don't properly copy on write


gregli-msft:
Really deep mutations were not being properly copied on write, where t should be unaffected by the Patch:

```
>> Set( deep, [[[[1,2,3],[4,5,6]]]] )
deep: [[[[1, 2, 3], [4, 5, 6]]]]

>> Set( t, deep )
t: [[[[1, 2, 3], [4, 5, 6]]]]

>> Patch( First(First(First(deep).Value).Value).Value, {Value:2}, {Value:99} )
{Value:99}

>> deep
[[[[1, 99, 3], [4, 5, 6]]]]

>> t
[[[[1, 99, 3], [4, 5, 6]]]]
```

MikeStall:
fixed

PR #2457, opened 2024-06-05
Deep mutations of records, squashed

gregli-msft:
Adds support for deep mutation through the Set function.  `Set( a, {b:1} )` doesn't change and will assign a reference to the record `{b:1}` to the variable `a`.  After `a` has been established, it now supports mutation through `Set( a.b, 2 )`.  This is an optional facility that is disabled by default and there are no semantic changes.  When creating variables the host has the option to enable this with `SymbolProperties.CanSetMutate`.

This also fixes two related bugs.  First, really deep mutations were not being properly copied on write, where `t` should be unaffected by the Patch https://github.com/microsoft/Power-Fx/issues/2450:
```
>> Set( deep, [[[[1,2,3],[4,5,6]]]] )
deep: [[[[1, 2, 3], [4, 5, 6]]]]

>> Set( t, deep )
t: [[[[1, 2, 3], [4, 5, 6]]]]

>> Patch( First(First(First(deep).Value).Value).Value, {Value:2}, {Value:99} )
{Value:99}

>> deep
[[[[1, 99, 3], [4, 5, 6]]]]

>> t
[[[[1, 99, 3], [4, 5, 6]]]]
```

And this should not return a casting error, instead returning Blank() https://github.com/microsoft/Power-Fx/issues/2451:
```
>> Patch(Blank(),Blank(),Blank())
Unable to cast object of type 'Microsoft.PowerFx.Types.BlankType' to type 'Microsoft.PowerFx.Types.RecordType'.
```

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
@@ -126,8 +126,14 @@ namespace Microsoft.PowerFx.Core.Functions
         /// </summary>
         public virtual bool HasPreciseErrors => false;
 
-        // Returns true if the function will mutate the value of argument 0, as is the case with Patch, Collect, Remove, etc.
-        public virtual bool MutatesArg0 => false;
+        /// <summary>
+        /// Returns true if the function will mutate the argument, as is the case of Patch, Collect, Remove, etc.
+        /// Set can also mutate, but needs to make a decision based on the argument's node.
+        /// This function covers both CanMutate and CanSetMutate scenarios which is checked in CheckTypes/CheckSemantics.
+        /// </summary>
+        /// <param name="argIndex">Index of the argument.</param>
+        /// <param name="arg">Argument at that index.</param>
+        public virtual bool MutatesArg(int argIndex, TexlNode arg) => false;
 
         public virtual RequiredDataSourcePermissions FunctionPermission => RequiredDataSourcePermissions.None;
 
@@ -239,6 +245,23 @@ namespace Microsoft.PowerFx.Core.Functions
         protected void ValidateArgumentIsMutable(TexlBinding binding, TexlNode arg, IErrorContainer errors)
         {
             if (binding.Features.PowerFxV1CompatibilityRules && !binding.IsMutable(arg))
+            {
+                errors.EnsureError(
+                    arg,
+                    new ErrorResourceKey("ErrorResource_MutationFunctionCannotBeUsedWithImmutableValue"),
+                    this.Name);
+            }
+        }
+
+        /// <summary>
+        /// Adds an error to the container if the given argument is immutable.
+        /// </summary>
+        /// <param name="binding"></param>
+        /// <param name="arg"></param>
+        /// <param name="errors"></param>
+        protected void ValidateArgumentIsSetMutable(TexlBinding binding, TexlNode arg, IErrorContainer errors)
+        {
+            if (binding.Features.PowerFxV1CompatibilityRules && !binding.IsSetMutable(arg))
             {
                 errors.EnsureError(
                     arg,
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
@@ -129,6 +129,8 @@ namespace Microsoft.PowerFx.Core.Functions
         /// <summary>
         /// Returns true if the function will mutate the argument, as is the case of Patch, Collect, Remove, etc.
         /// Set can also mutate, but needs to make a decision based on the argument's node.
+        /// For example, Set(x,{a:1}) is not a mutate and has a single FirstName node for the first argument, 
+        /// while Set(x.a,1) is a mutate and has a more complex node for the first argument.
         /// This function covers both CanMutate and CanSetMutate scenarios which is checked in CheckTypes/CheckSemantics.
         /// </summary>
         /// <param name="argIndex">Index of the argument.</param>
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
@@ -108,7 +108,7 @@ namespace Microsoft.PowerFx.Types
                 return record;
             }
 
-            return DValue<RecordValue>.Of(ArgumentOutOfRange(IRContext));
+            return DValue<RecordValue>.Of(ArgumentOutOfRangeError(IRContext));
         }
 
         /// <summary>
@@ -124,7 +124,7 @@ namespace Microsoft.PowerFx.Types
                 return record;
             }
 
-            return DValue<RecordValue>.Of(ArgumentOutOfRange(IRContext));
+            return DValue<RecordValue>.Of(ArgumentOutOfRangeError(IRContext));
         }
 
         // Index() does standard error messaging and then call TryGetIndex().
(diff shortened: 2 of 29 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -128,11 +128,22 @@ public void UpdateVariable(string name, TimeSpan value)
...
+            UpdateVariable(name, value, new SymbolProperties { CanMutate = true, CanSet = true });
+        }
+
+        /// <summary>
+        /// Create or update a named variable to a value, with custom CanSet/Mutate attributes. 
+        /// </summary>
+        /// <param name="name">variable name. This can be used in other formulas.</param>
+        /// <param name="value">constant value.</param>
+        /// <param name="newVarProps">symbol properties.</param>
+        public void UpdateVariable(string name, FormulaValue value, SymbolProperties newVarProps)
```
jas-valgotar, line 146: >public void UpdateVariable(string name, FormulaValue value, SymbolProperties newVarProps) [](http://example.com/codeflow?start=8&length=89)

I see the only place to leverage CanSetMutate prop is here,
How would I leverage it at compile time? If I was just doing engine.check() 
(At that time formula value doesn't exist)

Should we also expose this flag at symbolTable.AddVariable() ?

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -151,7 +162,12 @@ public void UpdateVariable(string name, FormulaValue value)
             else
             {
                 // New
-                var slot = _symbolTable.AddVariable(name, value.Type, mutable: true);
+                if (newVarProps == null)
```
jas-valgotar, line 165: If variable already exists, does that mean we would not honor `newVarProps` props? We may want to document that.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
gregli-msft: AddVariable already has an overload that takes SymbolProperties as an argument in SymbolTable.cs line 119-ish:
```
public ISymbolSlot AddVariable(string name, FormulaType type, SymbolProperties props, string displayName = null)
```

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -126,8 +126,14 @@ internal abstract class TexlFunction : IFunction
         /// </summary>
         public virtual bool HasPreciseErrors => false;
 
-        // Returns true if the function will mutate the value of argument 0, as is the case with Patch, Collect, Remove, etc.
-        public virtual bool MutatesArg0 => false;
+        /// <summary>
+        /// Returns true if the function will mutate the argument, as is the case of Patch, Collect, Remove, etc.
+        /// Set can also mutate, but needs to make a decision based on the argument's node.
```
CarlosFigueiraMSFT, line 131: Can you give an example on the Set function where it "needs to make a decision based on the argument's node"?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/SetFunction.cs
```diff
@@ -70,52 +73,58 @@ public override void CheckSemantics(TexlBinding binding, TexlNode[] args, DType[
...
-                    var arg1 = argTypes[1];
+                errors.EnsureError(DocumentErrorSeverity.Critical, args[1], ErrBadType_ExpectedType_ProvidedType, arg0.GetKindString(), arg1.GetKindString());
+                return;
+            }
 
-                    if (!(arg0.Accepts(arg1, exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: binding.Features.PowerFxV1CompatibilityRules) ||
-                         (arg0.IsNumeric && arg1.IsNumeric)))
+            if (arg1.AggregateHasExpandedType())
+            {
+                if (binding.Features.SkipExpandableSetSemantics)
```
CarlosFigueiraMSFT, line 89: I know this code was already there, but can you add an example of a scenario where we would get this error?

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
```diff
@@ -109,20 +109,92 @@ public DValue<RecordValue> Index(int index1)
...
+        /// <summary>
+        /// Lookup the first record, or return blank if the table is empty.
+        /// </summary>
+        /// <param name="mutationCopy">copies the element, and the table entry pointing to it, when in a mutation context.</param>
+        /// <returns>The record or blank.</returns>
+        public virtual DValue<RecordValue> First(bool mutationCopy = false)
+        {
+            if (mutationCopy)
+            {
+                return DValue<RecordValue>.Of(ImmutableTable(IRContext));
```
CarlosFigueiraMSFT, line 169: Suggestion: rename `ImmutableTable` function to `ImmutableTableError`. It took me a while looking at this code to try to understand why First (and Last) would return a table, only to see down this file that the function returned an error.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -128,11 +128,22 @@ public void UpdateVariable(string name, TimeSpan value)
...
+        {
+            UpdateVariable(name, value, new SymbolProperties { CanMutate = true, CanSet = true });
+        }
+
+        /// <summary>
+        /// Create or update a named variable to a value, with custom CanSet/Mutate attributes. 
+        /// </summary>
+        /// <param name="name">variable name. This can be used in other formulas.</param>
+        /// <param name="value">constant value. The variable will take the type of this value on create.</param>
+        /// <param name="newVarProps">symbol properties. This is only used on the intiail create of the variable.</param>
```
CarlosFigueiraMSFT, line 145: [typo] `intiail` --> `initial`

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/MutationScripts/AutomatedLowCodePlugins.txt
```diff
@@ -0,0 +1,181 @@
...
+
+// USE IN SCOPE VARIABLES
+
+// Scope variables cannot be Set
+
+>> With( NewRecord, Field1 )
+20
+
+>> With( NewRecord, Set( Field1, 2 ) )
+Errors: Error 22-28: The first argument of 'Set' should be a valid variable name, and cannot conflict with any existing control, screen, collection, or data source names. Found type 'Field1'
```
CarlosFigueiraMSFT, line 90: The error message lists the possible conflicts (control / screen / collection / data source) - but this is not the case here. Should we update the message to be more explicit?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
gregli-msft: Good call, added.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/SetFunction.cs
gregli-msft: I'm not sure what this does.  :)  If you backtrack to the Feature definition, there is a comment that this was put in for Cards and is temporary and to be removed.  Anderson is following up and I opened https://github.com/microsoft/Power-Fx/issues/2461.

src/libraries/Microsoft.PowerFx.Core/Public/Values/TableValue.cs
gregli-msft: Good call, done.

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
gregli-msft: Fixed.

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/MutationScripts/AutomatedLowCodePlugins.txt
gregli-msft: Updated.

CarlosFigueiraMSFT (approved):
:ship:

PR #2468, opened 2024-06-07
Better SymbolValue Helpers

MikeStall:
Add helper for creating SymbolValues from CheckResults. 
This is tricky since it needs to include whatever symbolVAlues the checkresult already has. So Check.Symbols.CreateValues() may miss those and end up creating duplicate symbol values. 

Also add more tests

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Environment/ComposedReadOnlySymbolValues.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Environment/ComposedReadOnlySymbolValues.cs
@@ -34,7 +34,7 @@ namespace Microsoft.PowerFx
                 var st = kv.Key;
                 var sv = kv.Value;
 
-                tw.WriteLine($"{indent}{st.DebugName} --> {sv.DebugName}_{sv.GetHashCode()}");
+                tw.WriteLine($"{indent}{st.DebugName} --> {sv.DebugNameWithId}");
             }
         }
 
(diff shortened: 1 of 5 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/ParsedExpression.cs
```diff
@@ -91,6 +91,29 @@ internal static IExpressionEvaluator GetEvaluator(this CheckResult result, Stack
 
             return expr;
         }
+
+        /// <summary>
+        /// Create SymbolValues that match against <see cref="CheckResult.Symbols"/>.
+        /// </summary>
+        /// <param name="check"></param>
+        /// <returns></returns>
+        public static ReadOnlySymbolValues CreateValues(this CheckResult check)
```
MikeStall, line 100: This is useful method to add

src/libraries/Microsoft.PowerFx.Interpreter/Environment/ComposedReadOnlySymbolValues.cs
```diff
@@ -91,18 +104,24 @@ internal class ComposedReadOnlySymbolValues : ReadOnlySymbolValues
...
-            catch
+
+            if (map.TryGetValue(symTable, out var existing))
             {
-                // Give better error.
-                if (map.TryGetValue(symTable, out var existingSymValues))
+                if (object.ReferenceEquals(existing, symValues))
                 {
-                    throw new InvalidOperationException($"SymbolTable {symTable.DebugName()} already has SymbolValues '{existingSymValues.DebugName}' associated with it. Can't add '{symValues.DebugName}'");
+                    // Ok, same instance already added
```
MikeStall, line 112: This is fix. 
If it's literally same symbolvalue instance, it's ok

src/libraries/Microsoft.PowerFx.Interpreter/ParsedExpression.cs
jas-valgotar: Love it!

jas-valgotar (approved):
:shipit:

PR #2470, opened 2024-06-10
Revert comparison change in intellisense

CarlosFigueiraMSFT:
There was a recent change that updated the comparison for Intellisense suggestions. While the change looks good, it broke a test in Power Apps, so we're reverting it to unblock the integration, and later we can decide whether to reapply this change and update the test.


Shpakh-MSFT (approved):
:shipit:

CarlosFigueiraMSFT:
Will keep the change

Issue #2455, opened 2024-06-04
Get a list of all called functions in an expression 


anderson-joyle:
New API to get a list of all called functions in an expression. Maybe `ParseResults.GetListOfCalledFunctions()`.
Expression: `With({x:Abs(First(table).field)}, Abs(x-y))`
Result: ["With", "Abs", "First", "Abs"]

MikeStall:
This should hang off CheckResult, similalar to TopLevelIdentiifers. 

If we just want function names, we can get from parser:

IEnumerable<string> CheckResult.GetFunctionNames()
Very similar to what we already do with TopLevel identifiers. 
This can succeed after ApplyParse() and skip binding. 


If we want binding... we should return these: 
https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Core/Public/Reflection/FunctionInfo.cs

IEnumerable<FunctionInfo> CheckResult.GetFunctions()

PR #2474, opened 2024-06-11
Public API to get list of functions call

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2455.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
@@ -709,8 +709,7 @@ namespace Microsoft.PowerFx
 
         public IEnumerable<string> GetFunctionNames()
         {
-            var visitor = ListFunctionVisitor.Run(ApplyParse());
-            return visitor._functionNames.Keys;
+            return ListFunctionVisitor.Run(ApplyParse());
         }
     }
 
--- a/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
@@ -706,6 +706,12 @@ namespace Microsoft.PowerFx
 
             return summary;
         }
+
+        public IEnumerable<string> GetFunctionNames(bool fullNames = false)
+        {
+            var visitor = ListFunctionVisitor.Run(ApplyParse());
+            return fullNames ? visitor._functionNames.Keys : visitor._functionNames.Values;
+        }
     }
 
     // Internal interface to ensure that Result objects have a common contract
--- a/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
@@ -707,7 +707,7 @@ namespace Microsoft.PowerFx
             return summary;
         }
 
-        public IEnumerable<string> GetFunctionNames(bool fullNames = false)
+        public IEnumerable<string> GetFunctionNames(bool fullNames = true)
         {
             var visitor = ListFunctionVisitor.Run(ApplyParse());
             return fullNames ? visitor._functionNames.Keys : visitor._functionNames.Values;
--- a/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
@@ -707,10 +707,10 @@ namespace Microsoft.PowerFx
             return summary;
         }
 
-        public IEnumerable<string> GetFunctionNames(bool fullNames = true)
+        public IEnumerable<string> GetFunctionNames()
         {
             var visitor = ListFunctionVisitor.Run(ApplyParse());
-            return fullNames ? visitor._functionNames.Keys : visitor._functionNames.Values;
+            return visitor._functionNames.Keys;
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/libraries/Microsoft.PowerFx.Core/Syntax/Visitors/ListFunctionVisitor.cs
```diff
@@ -0,0 +1,55 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+
+namespace Microsoft.PowerFx.Syntax
+{
+    public class ListFunctionVisitor : IdentityTexlVisitor
```
MikeStall, line 8: >public [](http://example.com/codeflow?start=4&length=6)

internal

anderson-joyle: Sure. I kept it public so we do not need to change the code in other projects.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -706,6 +706,12 @@ public CheckContextSummary ApplyGetContextSummary()
 
             return summary;
         }
+
+        public IEnumerable<string> GetFunctionNames(bool fullNames = false)
```
MikeStall, line 710: can we default to true? Giving back with namespace is more useful.

MikeStall (dismissed):

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -706,6 +706,12 @@ public CheckContextSummary ApplyGetContextSummary()
 
             return summary;
         }
+
+        public IEnumerable<string> GetFunctionNames(bool fullNames = true)
```
CarlosFigueiraMSFT, line 710: >bool fullNames = true [](http://example.com/codeflow?start=52&length=21)

Is there a scenario where we **don't** want the full names? If so, can we avoid this argument?

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ListFunctionVisitorTests.cs
```diff
@@ -0,0 +1,59 @@
...
+        [InlineData("Abs(Abs(Abs(Abs(Abs(1)))))", "Abs", null)]
+        [InlineData("With({x:Today()}, x+1)", "With,Today", null)]
+        [InlineData("SomeNameSpace.Foo() + SomeNameSpace.Bar()", "Foo,Bar", "SomeNameSpace.Foo,SomeNameSpace.Bar")]
+        [InlineData("true And true", "", null)]
+        [InlineData("If(true, Blank(),Error())", "If,Blank,Error", null)]
+        public void ListFunctionNamesTest(string expression, string expectedNames, string expectedFullNames)
+        {
+            foreach (var textFirst in new bool[] { false, true })
+            {
+                if (textFirst)
```
CarlosFigueiraMSFT, line 22: >textFirst [](http://example.com/codeflow?start=20&length=9)

It would be good to have a test for testFirst without the leading `=` sign. For example: expression = `Hello ${Sum(1,Sqrt(2))} world`, and check that the visitor returns `Sum,Sqrt`.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ListFunctionVisitorTests.cs
```diff
@@ -0,0 +1,59 @@
...
+                var actualNames1 = string.Join(",", functionsNames1);
+                var actualNames2 = string.Join(",", functionsNames2);
+
+                Assert.Equal(expectedNames, actualNames1);
+                Assert.Equal(expectedNames, actualNames2);
+
+                if (expectedFullNames != null)
+                {
+                    var functionsFullNames = checkResult.GetFunctionNames();
+                    var actualFullNames2 = string.Join(",", functionsFullNames);
```
CarlosFigueiraMSFT, line 45: >actualFullNames2 [](http://example.com/codeflow?start=24&length=16)

[minor] why actualFullNames2? Why not just actualFullNames?

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
anderson-joyle: There are other hosts that can use the short name option.

src/libraries/Microsoft.PowerFx.Core/Syntax/Visitors/ListFunctionVisitor.cs
```diff
@@ -0,0 +1,35 @@
...
+
+using System.Collections.Generic;
+
+namespace Microsoft.PowerFx.Syntax
+{
+    internal class ListFunctionVisitor : IdentityTexlVisitor
+    {
+        // FullName --> Name 
+        // Use Fullname as key because it's unique. 
+        public Dictionary<string, string> _functionNames = new Dictionary<string, string>();
```
CarlosFigueiraMSFT, line 12: >Dictionary<string, string> [](http://example.com/codeflow?start=15&length=26)

Since the unqualified names are not used, consider changing it to a HashSet<string> instead.

CarlosFigueiraMSFT, line 12: >public [](http://example.com/codeflow?start=8&length=6)

Instead of having a public member, declare it as private, and expose it as an IEnumerable<string> instead

CarlosFigueiraMSFT (approved):
:shipit:

PR #2476, opened 2024-06-14
Improve TexlFunctionSet

LucGenetier:
config.EnableJsonFunctions was generating a full copy of all dictionaries storing our builtin functions
this fix is avoiding such copies to happen


src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -106,14 +130,13 @@ internal TexlFunctionSet(IEnumerable<TexlFunctionSet> functionSets)
...
             }
 
-            foreach (var functionSet in functionSets)
-            {
-                functionSet._guard.VerifyNoWriters();
+            List<TexlFunctionSet> tfsList = functionSets.SelectMany(fs => fs.Sets).Where(fs => fs._count > 0).ToList();
 
-                if (functionSet._count > 0)
-                {
-                    Add(functionSet);
```
LucGenetier, line 115: >Add(functionSet) [](http://example.com/codeflow?start=20&length=16)

Key memory and performance issue

src/tests/.Net4.6.2/Microsoft.PowerFx.Connectors.Tests/Microsoft.PowerFx.Connectors.Tests.csproj
```diff
@@ -14,8 +14,8 @@
     <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
     <Reference Include="System.Web" />
     
-    <PackageReference Include="Azure.Identity" Version="1.11.0" /> <!-- CVE-2024-29992 -->
-    <PackageReference Include="Microsoft.Identity.Client" Version="4.61.1" /> <!-- CVE-2024-27086 -->
+    <PackageReference Include="Azure.Identity" Version="1.11.4" />
```
MikeStall, line 17: >Identity [](http://example.com/codeflow?start=37&length=8)

is this a necessary  part of the hotfix? 

(especially considering this is a potential QFE candidate)

src/libraries/Microsoft.PowerFx.Core/Public/Config/ComposedReadOnlySymbolTable.cs
```diff
@@ -19,12 +19,12 @@ namespace Microsoft.PowerFx
     /// </summary>
     internal class ComposedReadOnlySymbolTable : ReadOnlySymbolTable, INameResolver, IGlobalSymbolNameResolver, IEnumStore
     {
-        private readonly IEnumerable<ReadOnlySymbolTable> _symbolTables;
+        private readonly List<ReadOnlySymbolTable> _symbolTables;
```
MikeStall, line 22: >List [](http://example.com/codeflow?start=25&length=4)

do we need this change?

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -328,20 +328,22 @@ internal void RemoveFunction(TexlFunction function)
...
 
-            if (functions._count == 0)
+            if (functions.Count() == 0)
             {
                 return;
             }
-
-            _functions.Add(functions);
+
+            TexlFunctionSet funcsToAdd = makeCopy ? functions.Clone() : functions;
```
MikeStall, line 341: >         TexlFunctionSet funcsToAdd = makeCopy ? functions.Clone() : functions; [](http://example.com/codeflow?start=3&length=79)

fwiw - I notice that if I switch this to always clone() (and ignore makeCopy), we still get same perf numbers in our benchmark.  

Is that expected?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -123,15 +150,23 @@ internal IEnumerable<TexlFunction> Functions
...
                     }
                 }
             }
-        }        
+        }
+
+        public void PreventRemoval()
+        {
+            _canremove = false;            
+        }
```
MikeStall, line 169: is there anything that calls this?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -28,45 +32,68 @@ internal class TexlFunctionSet
         // Count of functions
         internal int _count;
 
-        internal Dictionary<string, List<TexlFunction>>.KeyCollection FunctionNames => _functions.Keys;
+        internal bool _canremove;
```
MikeStall, line 35: I think this is a good idea. 
A concept of immutable sets is good and can help with optimizations.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -28,45 +32,68 @@ internal class TexlFunctionSet
         // Count of functions
         internal int _count;
```
MikeStall, line 33: with new Count() method - should this field now be private? 
Ie, should all callers go through Count() now?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -106,14 +134,13 @@ internal TexlFunctionSet(IEnumerable<TexlFunctionSet> functionSets)
...
 
-                if (functionSet._count > 0)
-                {
-                    Add(functionSet);
-                }
+            // keep local (empty) set for additional functions that could be added later
+            if (tfsList.Any())
+            {
+                _composed = true;
+                _texlFunctionSets = tfsList;
```
MikeStall, line 143: >        _texlFunctionSets = tfsList; [](http://example.com/codeflow?start=8&length=36)

Before - we took an (expensive) snapshot of the functionset.  So if the function set was modified later, we did not pick up those changes here. 

Now, we avoid the copy (win!) - but could that change the semantics?
Ie, do we need to verify that the incoming functionSets are not mutable?

src/libraries/Microsoft.PowerFx.Core/Public/Config/ComposedReadOnlySymbolTable.cs
LucGenetier: little perf improvement iirc but not stictly needed

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
LucGenetier: right could be removed, dead code

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
LucGenetier: very possible as this method isn't used a lot IIRC

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
MikeStall: Here's a test demonstrating the change in behavior: #2478

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
LucGenetier: yes we can make it private

src/tests/.Net4.6.2/Microsoft.PowerFx.Connectors.Tests/Microsoft.PowerFx.Connectors.Tests.csproj
MikeStall: Latest main has this updates. Clearly, this should not be part of the hotfix; but let's confirm it goes away when you rebase on latest.

PR #2484, opened 2024-06-17
Optimization for function sets 

MikeStall:
Optimize the pattern where:
- a host calls eval many times in a loop. 
- but using the same function set in all cases, as determined by just PowerFxConfig (with some enables).

This PR is an optimization and there shouldn't be any functional changes. However - we should review some of the behavior around TexlFunctionSet to enable us to pursue more comprehensive optimizations. 

Today, every call to Check() will compute the function set - which involves merging the function set from builtin functions plus the extra ones that were enabled.  This is expensive because TexlFunctionSet will "flatten" and copy all the functions into a single list. (Contrast to how we compose SymbolTables - which is entirely deferred). 

The optimization is to allow composing that once and caching it. But also to be sure we have correct invalidation strategy if anything changes. Added lots more tests as I discovered interesting corner cases. 

This makes the benchmark use 1/10 the working set and run 3x faster.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
@@ -52,8 +52,13 @@ namespace Microsoft.PowerFx.Core.Functions
 
         internal IEnumerable<string> Enums => _enums;
 
-        // Return an empty TexlFucntion set that can never be added to. 
-        internal static TexlFunctionSet Empty()
+        // Return an singleton for an empty TexlFunction set that can never be added to. 
+        internal static readonly TexlFunctionSet _empty = MakeEmpty();
+
+        internal static TexlFunctionSet Empty() => _empty;
+
+        // Create an empty function set that can never be added to.
+        private static TexlFunctionSet MakeEmpty()
         {
             var set = new TexlFunctionSet();
             set._guard.ForbidWriters();
@@ -97,6 +102,35 @@ namespace Microsoft.PowerFx.Core.Functions
                 Add(func);
             }
         }
+
+        // This will compose multiple function sets into a single one. 
+        // The assumption is that all incoming function sets are unchanged for the lifetime of the returned set.
+        // Hence it doesn't matter where this shares references or copies the values. 
+        internal static TexlFunctionSet Compose(IEnumerable<TexlFunctionSet> functionSets)
+        {
+            TexlFunctionSet nonEmpty = null;
+            int countNonEmpty = 0;
+
+            foreach (var set in functionSets)
+            {
+                if (set.Count() > 0)
+                {
+                    nonEmpty = set;
+                    countNonEmpty++;
+                }
+            }
+
+            if (countNonEmpty == 0)
+            {
+                return Empty();
+            }
+            else if (countNonEmpty == 1)
+            {
+                return nonEmpty;
+            }
+
+            return new TexlFunctionSet(functionSets);
+        }            
 
         internal TexlFunctionSet(IEnumerable<TexlFunctionSet> functionSets)
             : this()
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/ComposedSymbolTableCache.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/ComposedSymbolTableCache.cs
@@ -1,0 +1,72 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System;
+
+namespace Microsoft.PowerFx
+{
+    // This can cache the results of ReadOnlySymbolTable.Compose - and includes a safe 
+    // invalidation strategy. 
+    // Compose() is fast because it defers all work. But it will cache the results of that work,
+    // (notably compuing TexlFunctionSets), so caching Compose is important to reuse those
+    // other caches. 
+    internal class ComposedSymbolTableCache
+    {
+        // A cached value from ReadOnlySymbolTable.Compose
+        private ReadOnlySymbolTable _result;
+
+        // The symbol tables that were composed for _result.
+        private ReadOnlySymbolTable[] _list;
+
+        // Determine if the cache should be invalidated. 
+        // this will set _result to null to force a recompute.
+        private void MaybeInvalidateCache(ReadOnlySymbolTable[] tables)
+        {
+            // Check cache invalidation.
+            if (_list != null)
+            {
+                if (_list.Length != tables.Length)
+                {
+                    // Developer error - list that we cache on should be statically fixed. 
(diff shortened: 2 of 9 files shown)


src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -52,8 +52,13 @@ internal class TexlFunctionSet
 
         internal IEnumerable<string> Enums => _enums;
 
-        // Return an empty TexlFucntion set that can never be added to. 
-        internal static TexlFunctionSet Empty()
+        // Return an singleton for an empty TexlFunction set that can never be added to. 
+        internal static readonly TexlFunctionSet _empty = MakeEmpty();
```
MikeStall, line 56: this saved about 1kb per loop.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunctionSet.cs
```diff
@@ -98,6 +103,35 @@ internal TexlFunctionSet(IEnumerable<TexlFunction> functions)
             }
         }
 
+        // This will compose multiple function sets into a single one. 
+        // The assumption is that all incoming function sets are unchanged for the lifetime of the returned set.
+        // Hence it doesn't matter where this shares references or copies the values. 
+        internal static TexlFunctionSet Compose(IEnumerable<TexlFunctionSet> functionSets)
```
MikeStall, line 109: This is critical. 
Each symbol table has a function set, but most are empty. 
It's also key that the caller of this method knows the symbols won't change on us - this is true because this is only called during a binding pass (which operates on a fixed set of symbols - and has safety checks to throw if the symbols are changed during binding).

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -163,14 +163,21 @@ private protected INameResolver CreateResolverInternal(out ReadOnlySymbolTable s
...
         /// Get a combined engine symbol table, including builtins and config. 
         /// </summary>
         /// <returns></returns>
         public ReadOnlySymbolTable GetCombinedEngineSymbols()
         {
-            var symbols = ReadOnlySymbolTable.Compose(EngineSymbols, SupportedFunctions, Config.SymbolTable, PrimitiveTypes);
+            // Compose all the symbol tables most likely to have functions into a single 
+            // symbol table and then cache that. 
+            // That will cache unifying into a single TexlFunctionSet - which is the most expensive part. 
+            var functionList = _functionListCache.GetComposedCached(SupportedFunctions, Config.SymbolTable);
```
MikeStall, line 178: This is critical.  This combines the symbol tables that have non-empty function sets into a single table, which can then be cached.  And being a single table, then the Compose() operator above becomes trivial and fast.

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/ConfigTests.cs
```diff
@@ -788,6 +788,23 @@ public async Task DefaultFunctions()
...
+        {
+            var config = new PowerFxConfig();
+            config.AddFunction(new Func1Function());
+            
+            var engine = new Engine(config);
+           
+            var result = engine.Check("Func1(3)"); // 6
+            Assert.True(result.IsSuccess);
+
+            config.AddFunction(new MultiplyFunction());
```
MikeStall, line 803: This line is key - ensuring that even if the config is mutated, the invalidation is still working.

src/libraries/Microsoft.PowerFx.Core/Public/Config/ComposedSymbolTableCache.cs
```diff
@@ -0,0 +1,72 @@
...
+                }
+            }
+        }
+
+        public ReadOnlySymbolTable GetComposedCached(params ReadOnlySymbolTable[] tables)
+        {
+            // This is a conceptually read-only operation, so it could be called on multiple threads. 
+            // CA2002 - Lock(this) is safe since it's a private object that doesn't cross AD boundaries.
+#pragma warning disable CA2002 // Do not lock on objects with weak identity
+            lock (this)
```
jas-valgotar, line 55: >lock (this) [](http://example.com/codeflow?start=12&length=11)

why not just lock on private object to avoid this warning and keep it simple?

src/tests/Microsoft.PowerFx.Core.Tests.Shared/SymbolTableTests.cs
```diff
@@ -387,6 +387,76 @@ public void Threading()
...
+        // It's ok to pass nulls to GetComposedCached. 
+        [Fact]
+        public void CacheNullOk()
+        {
+            var table1 = new SymbolTable { DebugName = "Table1" };
+            table1.AddConstant("c1", FormulaValue.New("constant"));
+            ReadOnlySymbolTable nullTable = null;
+            
+            ComposedSymbolTableCache cache = new ComposedSymbolTableCache();
+            var c1 = cache.GetComposedCached(table1, nullTable);
```
jas-valgotar, line 453: >cache.GetComposedCached(table1, nullTable); [](http://example.com/codeflow?start=21&length=43)

Will we null ref if?
cache.GetComposedCached(nullTable);

src/tests/Microsoft.PowerFx.Core.Tests.Shared/SymbolTableTests.cs
```diff
@@ -387,6 +387,76 @@ public void Threading()
...
+            var c1 = cache.GetComposedCached(table1, table2);
+
+            // Same args, cache returns identical instance
+            {
+                var c1b = cache.GetComposedCached(table1, table2);
+
+                Assert.True(object.ReferenceEquals(c1, c1b));
+            }
+
+            // Mutating an existing talbe is ok - ComposedSymbolTable will catch that
```
anderson-joyle, line 406: >talbe [](http://example.com/codeflow?start=36&length=5)

nit: typo

jas-valgotar (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Config/ComposedSymbolTableCache.cs
MikeStall: will update

src/tests/Microsoft.PowerFx.Core.Tests.Shared/SymbolTableTests.cs
MikeStall: should be fine. null array will nullref, but that's ok - developer error.   null items in an array are fine.   I'll add this test.

jas-valgotar (approved):
:shipit:

anderson-joyle (approved):
:shipit:

Issue #2456, opened 2024-06-04
Get suggestions from the formula, throws StackOverFlow exception in TextFirst mode


gekarapa:
The following test case will throw a StackOverFlow exception:
```
namespace Microsoft.PowerFx.Tests
{
    public class GetSuggestionsTests : PowerFxTest
    {
        [Fact]
        public void GetSuggestionsWithRecursiveRecordTest()
        {

            var parserOptions = new ParserOptions()
            {
                Culture = System.Globalization.CultureInfo.InvariantCulture,
                AllowsSideEffects = true,
                TextFirst = true
            };

            var config = new PowerFxConfig();
            var engine = new Microsoft.PowerFx.Engine(config);

            var lazyTypeInstance = new LazyRecursiveRecordType();

            var res = engine.Check("=Filter(T,", lazyTypeInstance, parserOptions);
            var results = engine.Suggest(res, 10);
        }

        internal class LazyRecursiveRecordType : RecordType
        {
            public override IEnumerable<string> FieldNames => GetFieldNames();

            public bool EnumerableIterated = false;

            public LazyRecursiveRecordType()
                : base()
            {
            }

            public override bool TryGetFieldType(string name, out FormulaType type)
            {
                switch (name)
                {
                    case "SomeString":
                        type = FormulaType.String;
                        return true;
                    case "Loop":
                        type = this;
                        return true;

                    case "T":
                        type = this.ToTable();
                        return true;
                    default:
                        type = FormulaType.Blank;
                        return false;
                }
            }


            private IEnumerable<string> GetFieldNames()
            {
                EnumerableIterated = true;

                yield return "SomeString";
                yield return "Loop";
            }

            public override bool Equals(object other)
            {
                return other is LazyRecursiveRecordType; // All the same 
            }

            public override int GetHashCode()
            {
                return 1;
            }
        }
    }
}
```

MikeStall:
FYI, some existing related tests:
https://github.com/microsoft/Power-Fx/blob/5126fb4b11c627d83e8da2b38a16b153b4660c9f/src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/MarshalTests.cs#L150


https://github.com/microsoft/Power-Fx/blob/5126fb4b11c627d83e8da2b38a16b153b4660c9f/src/tests/Microsoft.PowerFx.Core.Tests.Shared/IntellisenseTests/SuggestTest.cs#L385

PR #2488, opened 2024-06-18
Avoid stackoverflow when getting suggestions

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2456.


src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseHelper.cs
```diff
@@ -898,5 +898,18 @@ internal static bool IsPunctuatorColonNextToCursor(int cursorPos, string script)
             var colonLen = TexlLexer.PunctuatorColon.Length;
             return script.Length >= cursorPos + colonLen && script.Substring(cursorPos, colonLen) == TexlLexer.PunctuatorColon;
         }
+
+        internal static bool ContainsDataEntityType(DType type, int maxDepth, int currentDepth = 1)
```
MikeStall, line 902: >int maxDepth, int currentDepth = 1 [](http://example.com/codeflow?start=64&length=34)

It's often simpler to have depth count down towards 0. (rather than count up and pass a maxdepth)

src/tests/Microsoft.PowerFx.Core.Tests.Shared/IntellisenseTests/SuggestTest.cs
```diff
@@ -539,6 +539,22 @@ public void SuggestAggregateReturnType(string expression, params string[] expect
...
+        {
+            var lazyInstance = new LazyRecursiveRecordType();
+            var config = PowerFxConfig.BuildWithEnumStore(
+                new EnumStoreBuilder(),
+                new TexlFunctionSet(new[] { BuiltinFunctionsCore.Filter }));
+
+            var suggestions = SuggestStrings(expression, config, null, lazyInstance);
+
+            // Just check that the execution didn't stack overflow.
+            Assert.True(suggestions.Any());
```
MikeStall, line 555: Can we merge with above?  TestSuggestLazyTypes

And test an expected result (not just Any())

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseHelper.cs
```diff
@@ -343,7 +343,7 @@ public static void AddTopLevelSuggestionsForCursorType(IntellisenseData.Intellis
             {
                 if (info.Function.IsLambdaParam(null, argPosition) && !info.Function.HasSuggestionsForParam(argPosition) && type.IsValid)
                 {
-                    if (type.ContainsDataEntityType(DPath.Root))
+                    if (ContainsDataEntityType(type, type.MaxDepth))
```
MikeStall, line 346: >type.MaxDepth [](http://example.com/codeflow?start=53&length=13)

what does this mean for a recursive type?

MikeStall, line 346: >ContainsDataEntityType [](http://example.com/codeflow?start=24&length=22)

This switches to a static instance call to a new method. 
But it doesn't fix the original method DType.ContainsDataEntityType... what about other callers to this method? They would still be subject to sthe stack overflow bug.

MikeStall (approved):
:shipit:

PR #2489, opened 2024-06-19
Move HttpClient to CdpTableValue and update ADS hack

LucGenetier:
HttpClient is now passed via InitAsync function and no more required in RuntimeConfig
Removed tabular IR rewriter and related logic
Updated ADS hack with a new approach - ADS will be update in another PR
Tested live with no side effect


MikeStall (approved):
:shipit:

PR #2490, opened 2024-06-19
Connectors: Don't log column names

LucGenetier:
When logging FormulaValue types, don't log column names in aggregated types, enums, option sets


src/libraries/Microsoft.PowerFx.Connectors/ConnectorHelperFunctions.cs
```diff
@@ -147,5 +151,111 @@ internal static string LogException(Exception ex)
         {
             return $"Exception {ex.GetType().FullName}, Message {ex.Message}, Callstack {ex.StackTrace}";
         }
+
+        // DType.ToString() equivalent which is safe for logging (column names are not displayed)
+        internal static string SafeDTypeToString(this DType dType)
```
anderson-joyle, line 156: >SafeDTypeToString [](http://example.com/codeflow?start=31&length=17)

This is a nice anonymizer. Should we move it to Core so other processes can call it when needed?

src/tests/Microsoft.PowerFx.Connectors.Tests.Shared/SafeDTypeToStringTests.cs
```diff
@@ -0,0 +1,517 @@
...
+            bool err = false;
+            var type1 = DType.CreateTable(
+                new TypedName(DType.Number, new DName("A")),
+                new TypedName(DType.Number, new DName("B")),
+                new TypedName(DType.Number, new DName("C")));
+
+            Assert.Equal("*[field1:n, field2:n, field3:n]", type1.SafeDTypeToString());
+            Assert.Equal("![field1:n, field2:n, field3:n]", type1.ToRecord().SafeDTypeToString());
+            Assert.Equal("![field1:n, field2:n, field3:n]", type1.ToRecord().ToRecord().SafeDTypeToString());
+            Assert.Equal("![field1:n, field2:n, field3:n]", type1.ToRecord().SafeDTypeToString());
```
CarlosFigueiraMSFT, line 109: >Assert.Equal("![field1:n, field2:n, field3:n]", type1.ToRecord().SafeDTypeToString()); [](http://example.com/codeflow?start=12&length=86)

Same assert as 2 lines above

src/tests/Microsoft.PowerFx.Connectors.Tests.Shared/SafeDTypeToStringTests.cs
```diff
@@ -0,0 +1,517 @@
...
+                });
+
+            Assert.Equal("*[field1:n, field2:n, field3:n]", type2.SafeDTypeToString());
+
+            type2 = DType.CreateRecord(type1.GetNames(DPath.Root).ToArray());
+            type2 = type2.Add(ref err, DPath.Root, new DName("B"), DType.String);
+            Assert.Equal("![field1:n, field2:s, field3:n]", type2.SafeDTypeToString());
+
+            type2 = DType.EmptyTable
+                .Add(ref err, DPath.Root, new DName("B"), DType.Number)
```
CarlosFigueiraMSFT, line 126: >.Add(ref err, DPath.Root, new DName("B"), DType.Number) [](http://example.com/codeflow?start=16&length=55)

If you're not using the `err` argument, and always passing DPath.Root to the second one, you can use the smaller overload that only takes DName/DType

src/tests/Microsoft.PowerFx.Connectors.Tests.Shared/SafeDTypeToStringTests.cs
```diff
@@ -0,0 +1,517 @@
...
+            DType type8 = type1.Add(ref err, DPath.Root, new DName("X"), type6);
+            Assert.Equal("*[field1:n, field2:n, field3:n, field4:*[field1:b, field2:*[field1:s, field2:s]]]", type8.SafeDTypeToString());
+        }
+
+        [Fact]
+        public void EnumDTypeTests()
+        {
+            Assert.True(DType.TryParse("%n[A:0, B:1, C:2, D:3]", out DType type) && type.IsEnum);
+            Assert.True(DType.TryParse("%n[A:0, B:1, C:2, D:3]", out DType type2) && type2.IsEnum);
+        }
```
CarlosFigueiraMSFT, line 187: Is it needed? There's no testing for the new function here.

src/libraries/Microsoft.PowerFx.Connectors/ConnectorHelperFunctions.cs
```diff
@@ -147,5 +151,111 @@ internal static string LogException(Exception ex)
...
+                    SafeAppendEnumType(sb, dType.ValueTree, dType.EnumSuperkind);
+                    break;
+            }
+        }
+
+        private static void SafeAppendAggregateType(StringBuilder sb, TypeTree tree)
+        {
+            sb.Append("[");
+
+            string strPre = string.Empty;
```
CarlosFigueiraMSFT, line 187: >strPre [](http://example.com/codeflow?start=19&length=6)

Suggestion: call it `separator` to better reflect what it does.

src/libraries/Microsoft.PowerFx.Connectors/ConnectorHelperFunctions.cs
MikeStall: I'm good starting in connectors, and moving it ot core when we need it.

MikeStall, line 156: >SafeDTypeToString [](http://example.com/codeflow?start=31&length=17)

nit- "safe" is ambiguous.  Something with Logging or Anonymous would be helpful here. 
"ToAnonymousString" or "ToLoggingString.


[https://github.com/microsoft/Power-Fx/blob/6bcbe7da6a468f0f9a55585fef7fc7ecf80fdb60/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs#L663](https://github.com/microsoft/Power-Fx/blob/6bcbe7da6a468f0f9a55585fef7fc7ecf80fdb60/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs#L663)

src/tests/Microsoft.PowerFx.Connectors.Tests.Shared/SafeDTypeToStringTests.cs
LucGenetier: Good catch! :)

src/tests/Microsoft.PowerFx.Connectors.Tests.Shared/SafeDTypeToStringTests.cs
LucGenetier: missing code :)

MikeStall (approved):

PR #2491, opened 2024-06-19
NullReferenceException fix

LucGenetier:
Fix NullReferenceException in connector's code
Remove MediaType check in HttpFunctionInvoker (no need)


MikeStall (changes requested):
:clock1:

Robulane (approved):

PR #2493, opened 2024-06-19
Revert/Change TexlFunction flag name restricting use inside UDF

rick-nguyen:
`CreatesImplicitScreenDependency` is a PowerApps specific behavior to InvokeControl functions and doesn't need to be in this class here. We can drive IsRestrictedInsideUdfBody from that flag in PowerApps.


CarlosFigueiraMSFT:
Can you add some description to this PR to explain why this change is being made?

Shpakh-MSFT (approved):

CarlosFigueiraMSFT (approved):
🚢

PR #2494, opened 2024-06-19
Fix imperative UDF throwing error without closing semicolon

rick-nguyen:
`Function():Void { Set(a, 1) }` should now be valid

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
@@ -384,6 +384,8 @@ namespace Microsoft.PowerFx.Core.Parser
                     ParseTrivia();
                     if (_curs.TidCur == TokKind.CurlyOpen)
                     {
+                        var isImperative = _curs.TidCur == TokKind.CurlyOpen && parserOptions.AllowsSideEffects;
+
                         _curs.TokMove();
                         _hasSemicolon = false;
                         ParseTrivia();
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -384,6 +384,8 @@ private ParseUserDefinitionResult ParseUDFsAndNamedFormulas(string script, Parse
                     ParseTrivia();
                     if (_curs.TidCur == TokKind.CurlyOpen)
                     {
+                        var isImperative = _curs.TidCur == TokKind.CurlyOpen && parserOptions.AllowsSideEffects;
```
anderson-joyle, line 387: >_curs.TidCur == TokKind.CurlyOpen [](http://example.com/codeflow?start=43&length=33)

This was already verified as true in line 387.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/UserDefinedFunctionTests.cs
```diff
@@ -575,6 +575,25 @@ public void TestUDFsReservedNamesTracking()
...
                 }
             }
+        }
+
+        [Theory]
+        [InlineData("func():Void { Set(x, 123) };")]
+        [InlineData("func():Void { Set(x, 123); Set(y, 123) };")]
+        [InlineData("func():Void = { Set(x, 123) };")]
+        [InlineData("func():Void = { Set(x, 123); Set(y, 123) };")]
+        public void TestImperativeUDFParseWithoutSemicolon(string script)
```
anderson-joyle, line 585: >TestImperativeUDFParseWithoutSemicolon [](http://example.com/codeflow?start=20&length=38)

Not sure if I understood. The test name state we are testing an UDF without semicolon but what differ between them is the equal sign.

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
anderson-joyle, line 387: >isImperative [](http://example.com/codeflow?start=28&length=12)

Can you add a comment explaining what this is used for?

src/tests/Microsoft.PowerFx.Core.Tests.Shared/UserDefinedFunctionTests.cs
rick-nguyen: Before this change, we would need a semicolon after the expression like this `Set(x, 123);`, if we didn't have an equal sign. This test is ensuring parity between equal sign or no.

anderson-joyle: I see. Thanks for explaining.

Shpakh-MSFT (dismissed):
:shipit:

anderson-joyle (approved):
:shipit:

PR #2497, opened 2024-06-20
Fix NullReferenceException on new IsOptionSetBackedByXXX properties

CarlosFigueiraMSFT:
Noticed from telemetry some NullReferenceException errors in those new properties. Adding guard to fix it.


Shpakh-MSFT:
>                 var typeSuffix = string.IsNullOrEmpty(OptionSetInfo?.EntityName) ? string.Empty : $" ({OptionSetInfo.EntityName})";

Here as well?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Types/DType.cs:1012 in 06637c1. [](commit_id = 06637c16d0371fa30a31d45db772d892a522fa82, deletion_comment = False)

Shpakh-MSFT (approved):
:shipit:

PR #2498, opened 2024-06-21
ReflectionFunction to handle BlankValue

anderson-joyle:
This change will support https://github.com/microsoft/Power-Fx-Dataverse/issues/507.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs
@@ -269,6 +269,10 @@ namespace Microsoft.PowerFx
                 {
                     arg = FormulaValue.New(string.Empty);
                 }
+                else if (arg is BlankValue && expectedType is RecordType)
+                {
+                    // Add arg as is.
+                }
                 else if (arg is BlankValue)
                 {
                     if (errors == null)
--- a/src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/ConfigTests.cs
+++ b/src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/ConfigTests.cs
@@ -118,6 +118,44 @@ namespace Microsoft.PowerFx.Interpreter.Tests
             Assert.Equal(6.0, result.ToObject());
         }
 
+        [Fact]
+        public async Task ExpectRecordValueFunctionTest()
+        {
+            var engine = new RecalcEngine();
+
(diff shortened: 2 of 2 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs
```diff
@@ -269,6 +269,10 @@ public async Task<FormulaValue> InvokeAsync(IServiceProvider serviceProvider, Fo
                 {
                     arg = FormulaValue.New(string.Empty);
                 }
+                else if (arg is BlankValue && expectedType is RecordType)
+                {
+                    // Add arg as is.
+                }
```
anderson-joyle, line 275: Adding this to avoid breaking changes to existing custom functions by other hosts.
Ideally a `BlankValue` should be able to hit runtime implementation.
I'll gather more info on how this might affect hosts.

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs
MikeStall: I think this is fine. Blank is a valid case and so we shouldn't fail, like the other cases here. 

Just add a test for it.

jas-valgotar, line 275: Should this be converting blankValue to empty record?

From the above logic, it looks like we are assigning default value in blank case to avoid invocation problem when Execute() method is strictly define.

Ie if you defined input as a record type, then you may as well have taken RecordValue as input, and that would fail if at some point you get BlankValue instead of the RecordValue

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/ConfigTests.cs
```diff
@@ -118,6 +118,44 @@ public async Task BasicBindAndEvalFunc()
...
+
+        private class ExpectRecordValueFunction : ReflectionFunction
+        {
+            public ExpectRecordValueFunction()
+                : base("ExpectRecordValue", FormulaType.String, RecordType.Empty())
+            {
+            }
+
+            // Must have "Execute" method. 
+            public FormulaValue Execute(FormulaValue fv)
```
jas-valgotar, line 148: >FormulaValue [](http://example.com/codeflow?start=40&length=12)

What if this is defined as RecordValue? would returning BlankValue cause problem?

PR #2499, opened 2024-06-24
Extend Script Minification to also allow removing comments

akshar2401:
This pr mainly focuses on extending existing script minification functionality (or remove formatting as it is called) with an ability to optionally removing comments. Removing comment tokens is not harmful as those are removed from the parse tree anyways. Along the way, i found few bugs in existing implementation of remove formatting that adds extra spaces instead of removing those like below
image

I also found a bug in TexlLexer that computes overlapping spans of tokens from LexComment() when it tries to consume preceding whitespaces of the first comment in the formula. This was supposed to be fix last year when this was addressed during semantic tokens work but seems like an edge case was missed. 

I am keeping all the bug fixes out of scope for now since these are very low priority and remove formatting is not that widely used.


src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
```diff
@@ -443,81 +425,37 @@ public static bool RequiresWhiteSpace(Token tk)
...
         /// </summary>
-        /// <param name="precedingToken">A token before the current token.</param>
-        /// <param name="currentToken">Current token that might be a string literal inside a larger interpolated string.</param>
-        /// <param name="followingToken">A token after the current token.</param>
-        /// <returns>True if current token represents a string literal inside a larger interpolated string or false otherwise.</returns>
-        private static bool IsStringPartOfALargerInterpolatedString(Token precedingToken, Token currentToken, Token followingToken)
+        /// <param name="text">Expression to minify.</param>
+        /// <param name="options">Minification options.</param>
+        /// <returns>Minified Expression.</returns>
+        public string GetMinifiedScript(string text, MinificationOptions options = null)
```
rick-nguyen, line 437: Do we need this if on 428 you are just calling the static method

rick-nguyen (approved):

src/libraries/Microsoft.PowerFx.Core/Lexer/TexlLexer.cs
akshar2401: I need this method in the doc server side.

PR #2500, opened 2024-06-24
Add Nullcheck to GetInfo result in TryGetEntityInfo

adithyaselv:
Getinfo may return null to certain callNodes and hence it is creating NullReferenceException consistently as a noise. this PR fixes that issue.


CarlosFigueiraMSFT (approved):
🚢

PR #2505, opened 2024-06-25
Add expression culture in NL2FxParameters

nguhoa:


src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Public/NLHandler.cs
```diff
@@ -64,7 +65,13 @@ public class NL2FxParameters
...
 
         // Engine can provide ambient details for NL, such as config and feature flags.
-        public Engine Engine { get; set; }
+        public Engine Engine { get; set; }
+
+        /// <summary>
+        /// Set the locale that expect the output expression to be.
+        /// For example: "vi-VN" or "fr-FR".
+        /// </summary>
+        public CultureInfo ExpressionCultureInfo { get; set; }
```
anderson-joyle, line 74: >ExpressionCultureInfo [](http://example.com/codeflow?start=27&length=21)

nit: most places we have named CultureInfo instance as `Locale`.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Nl2Fx/Nl2FxLanguageServerOperationHandler.cs
```diff
@@ -53,7 +53,8 @@ private void PreHandleNl2Fx(IPowerFxScope scope, LanguageServerOperationContext
             {
                 Sentence = _nl2FxRequestParams.Sentence,
                 SymbolSummary = summary,
-                Engine = check.Engine
+                Engine = check.Engine,
+                ExpressionCultureInfo = check.ParserCultureInfo
```
anderson-joyle, line 57: What if we pass `CheckResult` only?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Public/NLHandler.cs
nguhoa: update name

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Nl2Fx/Nl2FxLanguageServerOperationHandler.cs
nguhoa: Do you mean when ParserCultureInfo is null? then we just have locale null and we handle it in our NL library later.

akshar2401: No! we should avoid exposing check results out of this function. This function runs behind a host (document server for example) lock and check result holds references to host's not thread safe data structures. Just get what you need from check result and pass it through this object. don't throw out the whole check result from here

anderson-joyle: The idea was not exposing the whole object. We would pass `CheckResult` and then call something like `public Engine Engine => checkResult.Engine`. Just a suggestion. Whatever works best.

nguhoa: I think Akshar suggest to not pass checkresult here so i keep this as it is to safe.

MikeStall: yes, least privilege. 
Just culture is good here.

MikeStall (approved):
:shipit:

PR #2506, opened 2024-06-25
Add warning to UDFs that shadow existing built-ins, and remove Hyperlink from allowed return types

rick-nguyen:
This PR aims to improve UDF experience by providing a warning to users when they write a UDF that shares the same name as an existing built-in.


src/tests/Microsoft.PowerFx.Core.Tests.Shared/UserDefinedFunctionTests.cs
```diff
@@ -630,6 +630,23 @@ public void TestImperativeUDFParseWithoutSemicolon(string script)
...
+            {
+                AllowsSideEffects = true,
+            };
+            var nameResolver = ReadOnlySymbolTable.NewDefault(BuiltinFunctionsCore._library, FormulaType.PrimitiveTypes);
+
+            var parseResult = UserDefinitions.Parse(script, parserOptions);
+            var udfs = UserDefinedFunction.CreateFunctions(parseResult.UDFs.Where(udf => udf.IsParseValid), nameResolver, out var errors);
+            errors.AddRange(parseResult.Errors ?? Enumerable.Empty<TexlError>());
+
+            Assert.True(errors.Any(error => error.MessageKey == "WrnUDF_ShadowingBuiltInFunction" && error.Severity == DocumentErrorSeverity.Warning));
```
CarlosFigueiraMSFT, line 649: >True [](http://example.com/codeflow?start=19&length=4)

We should also assert that this is the only error (since that is a "valid" UDF definition)

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4076,6 +4076,10 @@
   <data name="ErrUDF_FunctionAlreadyDefined" xml:space="preserve">
     <value>Function {0} is already defined.</value>
     <comment>This error message shows up when user tries to define a function that already exists</comment>
+  </data>
+  <data name="WrnUDF_ShadowingBuiltInFunction" xml:space="preserve">
+    <value>Warning: User defined function {0} will shadow the existing built-in function {0}.</value>
```
CarlosFigueiraMSFT, line 4081: >{0} [](http://example.com/codeflow?start=89&length=3)

Please add to the comment the description of the parameter and an example, such as "The argument {0} will be replaced with a function name: e.g.: User defined function Count will shadow the existing built-in function Count.`

CarlosFigueiraMSFT, line 4081: >User defined [](http://example.com/codeflow?start=20&length=12)

I'm almost sure that we should have a hyphen here (`User-defined function`, instead of `User defined function`) - see [https://en.wikipedia.org/wiki/User-defined_function](https://en.wikipedia.org/wiki/User-defined_function), for example. Something to check with the content team and/or PM. Same with other resources that use this term.

CarlosFigueiraMSFT: Copilot also agrees with this :smile::

Both “user-defined function” and “user defined function” are grammatically correct, but they convey slightly different meanings. Let me explain:

1. “**User-Defined Function**”: This form emphasizes that the function is specifically created or defined by a user (or programmer). It suggests that the function is not part of the built-in language features or libraries.
2. “**User Defined Function**”: In this form, “user” serves as an adjective describing the type of function. It implies that the function is associated with or related to users, without necessarily emphasizing the act of defining it.

src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
```diff
@@ -87,7 +86,6 @@ internal FormulaType(DType type)
             { new DName("Number"), Number },
             { new DName("Decimal"), Decimal },
             { new DName("Text"), String },
-            { new DName("Hyperlink"), Hyperlink },
```
adithyaselv, line 90: >{ new DName("Hyperlink"), Hyperlink }, [](http://example.com/codeflow?start=12&length=38)

Please check with @jas-valgotar  if this is okay to do as this is also used in [https://github.com/microsoft/Power-Fx/blob/53da8f65899eaf8290eb09d6e9ca16f1efeedd56/src/libraries/Microsoft.PowerFx.Json/Public/Types/UserProvidedTypeDefinitions/FormulaTypeSchema.cs#L91](https://github.com/microsoft/Power-Fx/blob/53da8f65899eaf8290eb09d6e9ca16f1efeedd56/src/libraries/Microsoft.PowerFx.Json/Public/Types/UserProvidedTypeDefinitions/FormulaTypeSchema.cs#L91)

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -235,7 +235,7 @@ internal void AddUserDefinedFunction(string script, CultureInfo parseCulture = n
 
             errors.AddRange(parseResult.Errors ?? Enumerable.Empty<TexlError>());
 
-            if (errors.Any())
+            if (errors.Any(error => error.Severity > DocumentErrorSeverity.Warning))
```
adithyaselv, line 238: thanks for fixing this , can you also fix here - https://github.com/microsoft/Power-Fx/blob/f77304a685f385dc4dcb7fac58c40d4435271bee/src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs#L449

src/libraries/Microsoft.PowerFx.Core/Binding/BinderUtils.cs
```diff
@@ -146,7 +146,17 @@ internal static bool TryGetBestOverload(CheckTypesContext context, IErrorContain
...
+            Dictionary<TexlNode, DType> matchingFuncWithCoercionNodeToCoercedTypeMap = null;
+
+            // If a user-defined function exists as one of the possible overloads, always return that as it shadows existing built-in functions (provided it is not a restricted UDF name).
+            if (overloads.Any(overload => overload is UserDefinedFunction))
+            {
+                var udf = overloads.Where(overload => overload is UserDefinedFunction).First() as UserDefinedFunction;
+                var localWarnings = new LimitedSeverityErrorContainer(errors, DocumentErrorSeverity.Warning);
+                udf.CheckTypes(context, args, argTypes, localWarnings, out returnType, out nodeToCoercedTypeMap);
+                bestOverload = udf;
+                return true;
```
adithyaselv, line 158: >return true; [](http://example.com/codeflow?start=16&length=12)

should this always return true or should it be based on CheckTypes ?

also IIRC this path happens only when overloaded functions dont have lambda-arg/ ColumnNames-as-Id or something like that .

Can you also add some tests shadowing Filter , ShowColumns ?

rick-nguyen: UDF doesn't support lambda params so a Filter test wouldn't be too interesting I think

adithyaselv: clarified offline , comment was to verify if the logic works shadowing functions with LambdaArg and ColumnNames as Identifiers

src/libraries/Microsoft.PowerFx.Core/Syntax/UserDefinitions.cs
```diff
@@ -30,7 +30,7 @@ internal sealed class UserDefinitions
         private readonly Features _features;
 
         // Exposing it so hosts can filter out the intellisense suggestions
-        public static readonly ISet<DType> RestrictedTypes = new HashSet<DType> { DType.DateTimeNoTimeZone, DType.ObjNull,  DType.Decimal };
+        public static readonly ISet<DType> RestrictedTypes = new HashSet<DType> { DType.DateTimeNoTimeZone, DType.ObjNull, DType.Decimal, DType.Hyperlink };
```
CarlosFigueiraMSFT, line 33: >DType.Hyperlink [](http://example.com/codeflow?start=138&length=15)

I don't know if we already have, but we should have tests that assert that we would fail if we try to define a UDF using those types.

CarlosFigueiraMSFT (dismissed):
🚢

PR #2509, opened 2024-06-26
Add lexer method to strip comments in a script

adithyaselv:
Add a utility method to remove any comments from a given script.


akshar2401:
I added a similar method for my use case https://github.com/microsoft/Power-Fx/pull/2499 if you want to see if it can be reused for your purpose

PR #2510, opened 2024-06-27
CDP - Add TryGetFieldExternalTableName API

LucGenetier:
Add TryGetFieldExternalTableName API for CDP connectors
This API allows to know if an external table has a relationship with a given field
TryGetFieldType API gives the same information but the key difference is that it makes a network call to return the RecordType of the external table, while TryGetFieldExternalTableName doesn't make any network call


src/libraries/Microsoft.PowerFx.Connectors/Public/CdpRecordType.cs
```diff
@@ -26,19 +26,39 @@ internal CdpRecordType(ConnectorType connectorType, DType recordType, ICdpTableR
             ReferencedEntities = referencedEntities;
         }
 
-        public override bool TryGetFieldType(string name, out FormulaType type)
+        public bool TryGetFieldExternalTableName(string fieldName, out string tableName)
         {
-            if (!base.TryGetBackingDType(name, out _))
+            if (!base.TryGetBackingDType(fieldName, out _))
+            {
+                tableName = null;
```
nguhoa, line 33: can move line 33 to line 30. then can remove line 41

src/libraries/Microsoft.PowerFx.Connectors/Public/CdpRecordType.cs
```diff
@@ -26,19 +26,39 @@ internal CdpRecordType(ConnectorType connectorType, DType recordType, ICdpTableR
...
+                return false;
+            }
+
+            ConnectorType connectorType = ConnectorType.Fields.First(ct => ct.Name == fieldName);
```
nguhoa, line 37: can fields be null? should we use FirstOrDefault?

nguhoa: can we add test case where this fields is null as well?

nguhoa (dismissed):

src/libraries/Microsoft.PowerFx.Connectors/Public/CdpExtensions.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Connectors
+{
+    public static class CdpExtensions
```
jas-valgotar, line 8: >public [](http://example.com/codeflow?start=4&length=6)

nit: Can we keep it internal? or if we need to keep it public can we add comment on the below method.

src/libraries/Microsoft.PowerFx.Connectors/Public/CdpRecordType.cs
LucGenetier: I understsand your concern but we test the field existence at line 31 so we're sure this line will work fine

src/libraries/Microsoft.PowerFx.Connectors/Public/CdpRecordType.cs
LucGenetier: good idea

src/libraries/Microsoft.PowerFx.Connectors/Public/CdpExtensions.cs
LucGenetier: nope, I really want this API to be public and that's the reason for this extension as the CdpRecordType class is internal

src/libraries/Microsoft.PowerFx.Connectors/Public/CdpRecordType.cs
nguhoa: i see

nguhoa (approved):

nguhoa (approved):

PR #2517, opened 2024-07-01
Fix RecordAndTableDTypeTests inconsistency between Debug and DebugAll

LucGenetier:
Fix RecordAndTableDTypeTests inconsistency
Fixed "#if DEBUG" conditions which were false when Debug462/70/All is used
Fix DType.Add behavior which is different between Debug (exception) and Release (no exception) when we want to change the field type in a record/table type.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
@@ -1365,8 +1365,7 @@ namespace Microsoft.PowerFx.Core.Types
             {
                 fullType = LazyTypeProvider.GetExpandedType(IsTable);
             }
-
-            Contracts.Assert(!TypeTree.Contains(name));
+            
             var tree = TypeTree.SetItem(name, type);
             var newType = new DType(Kind, tree, AssociatedDataSources, DisplayNameProvider);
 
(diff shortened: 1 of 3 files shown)


src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1365,8 +1365,7 @@ public DType Add(DName name, DType type)
             {
                 fullType = LazyTypeProvider.GetExpandedType(IsTable);
             }
-
-            Contracts.Assert(!TypeTree.Contains(name));
```
MikeStall, line 1369: >Contracts [](http://example.com/codeflow?start=12&length=9)

When do we hit this?

MikeStall:
There's a broader problem here that Contact.Asserts are not enabled in PApps. We should investigate that. 
But safest thing is to just update the (recently added) test to avoid hitting this assert.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
CarlosFigueiraMSFT: When we try to add an existing field to an existing record / table type. Notice that DTypes are immutable, so we're not actually adding anything. Something like

    var type1 = Parse("![a:n,b:s,c:D]");
    var type2 = type1.Add(new DName("b"), DType.Number);

In release builds (i..e, what we've been using for years), this returns a type equivalent to "![a:n,b:n,c:D]"; in debug builds it asserts. Ideally it should return an error, but at this point this is the _de facto_ implementation.

MikeStall:
If we do change Dtype, the other thing to consider is adding a test (if we don't have it already) that tries this from the public surface area. Ie, ir we add 2 fields of same name to a RecordType

MikeStall (approved):

PR #2519, opened 2024-07-02
Fix Display Names

LucGenetier:
In Power Apps, action connections don't use Display Names

image


MikeStall:
Can this break existing customers using connectors?

PR #2521, opened 2024-07-02
Stackoverflow when calling DType.ContainsDataEntityType

anderson-joyle:
We are extending the range for https://github.com/microsoft/Power-Fx/issues/2456.

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests.Shared/RecursiveRecordTypeTest.cs
+++ b/src/tests/Microsoft.PowerFx.Core.Tests.Shared/RecursiveRecordTypeTest.cs
@@ -1,0 +1,27 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using Microsoft.CodeAnalysis;
+using Microsoft.PowerFx.Core.Texl.Builtins;
+using Microsoft.PowerFx.Syntax;
+using Xunit;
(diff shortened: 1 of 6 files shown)


src/tests/Microsoft.PowerFx.Core.Tests.Shared/RecursiveRecordTypeTest.cs
```diff
@@ -0,0 +1,27 @@
...
+using Microsoft.PowerFx.Syntax;
+using Xunit;
+
+namespace Microsoft.PowerFx.Core.Tests
+{
+    public class RecursiveTypeTest : PowerFxTest
+    {
+        [Theory]
+        [InlineData("Collect(TableLoop, First(TableLoop), First(TableLoop))")]
+        public void NoStackoverflowTest(string expression)
```
MikeStall, line 15: nit - is there an existing file this would below in?

MikeStall (approved):

src/tests/Microsoft.PowerFx.Core.Tests.Shared/RecursiveRecordTypeTest.cs
anderson-joyle: I didn't find a good fit. Let me know if you have a suggestion and I can move it later on.

PR #2522, opened 2024-07-02
Added another property on interface IExternalNamedFormula to detect reference to views in NF expressions.

Shpakh-MSFT:
Currently the logic to detect use of dataverse views in an expression doesn't work well when expression is abstracted under named formula.
For example,
T = Filter(Monthses, 'Monthses (Views)'.MyView);

Label1.Text = CountRows(T);

CountRows(T) is not detected currently as non-delegated expression as a result. This change fixes that. Unit test will be added on PAClient side for this where all of the delegation tests currently are.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Texl/ViewFinderVisitor.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Texl/ViewFinderVisitor.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using Microsoft.PowerFx.Core.Binding;
+using Microsoft.PowerFx.Core.Entities;
 using Microsoft.PowerFx.Core.Types;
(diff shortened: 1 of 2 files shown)


MikeStall (approved):

src/libraries/Microsoft.PowerFx.Core/Texl/ViewFinderVisitor.cs
```diff
@@ -34,7 +35,12 @@ public override void PostVisit(DottedNameNode node)
         }
 
         public override void Visit(FirstNameNode node)
-        {
+        {
```
MikeStall, line 38: curious - why are Views special cased (relative to tables or other things)?

Shpakh-MSFT: Answered offline.

PR #2524, opened 2024-07-03
Revert change to remove hyperlinks from allowed UDF types

rick-nguyen:
Dataverse columns support hyperlink types so this is needed for datasource types.


CarlosFigueiraMSFT (approved):
:shipit:

PR #2530, opened 2024-07-09
Add OrderBy

LucGenetier:
ODataParameters: Update OrderBy
DelegationParameters: Add GetOrderBy()

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/DelegationParameters.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/DelegationParameters.cs
@@ -34,6 +34,10 @@ namespace Microsoft.PowerFx.Types
             }
         }
 
+        /// <summary>
+        /// Returns the list of (column name, ascending/descending) where ascending=true
+        /// </summary>
+        /// <returns></returns>
         public abstract IList<(string, bool)> GetOrderBy();
 
         public abstract string GetOdataFilter();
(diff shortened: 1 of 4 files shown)


src/libraries/Microsoft.PowerFx.Core/Public/Values/DelegationParameters.cs
```diff
@@ -64,9 +54,43 @@ public virtual IReadOnlyCollection<string> GetColumns()
...
+        Levels = 1 << 9,
+
+        // $schemaversion
+        SchemaVersion = 1 << 10,
+
+        // $search
+        Search = 1 << 11,
+
+        // $skip
+        Skip = 1 << 12
```
MikeStall, line 94: these do not necessarily need to correspond to odata.
We don't need to to add these until we're actually consuming them.

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/Values/DelegationParameters.cs
```diff
@@ -44,6 +34,8 @@ public void EnsureOnlyFeatures(DelegationParameterFeatures allowedFeatures)
             }
         }
 
+        public abstract IList<(string, bool)> GetOrderBy();
```
MikeStall, line 37: >bool [](http://example.com/codeflow?start=39&length=4)

Assuming this is column names. 
can we comment what bool means?

MikeStall (approved):

MikeStall (approved):

Issue #2528, opened 2024-07-09
NullRef in IExternalStringResources.TryGetErrorResource


MikeStall:
NullRef.  Appears to be coming from here, value is null. 
```
		if (_errorResources.TryGetValue(locale, out var value) && value.TryGetValue(resourceKey.Key, out resourceValue))
		{
			return true;
		}
```

```
  Message: 
System.NullReferenceException : Object reference not set to an instance of an object.

  Stack Trace: 
Dictionary`2.FindEntry(TKey key)
Dictionary`2.TryGetValue(TKey key, TValue& value)
IExternalStringResources.TryGetErrorResource(ErrorResourceKey resourceKey, ErrorResource& resourceValue, String locale)
ErrorUtils.GetLocalizedErrorContent(ErrorResourceKey errKey, CultureInfo locale, ErrorResource& errorResource)
BaseError.ctor(IDocumentError innerError, Exception internalException, DocumentErrorKind kind, DocumentErrorSeverity severity, ErrorResourceKey errKey, Span textSpan, IEnumerable`1 sinkTypeErrors, Object[] args)
BaseError.ctor(IDocumentError innerError, Exception internalException, DocumentErrorKind kind, DocumentErrorSeverity severity, ErrorResourceKey errKey, Object[] args)
TexlError.ctor(TexlNode node, DocumentErrorSeverity severity, ErrorResourceKey errKey, Object[] args)
```

anderson-joyle:
Is there a way to repro this?

PR #2531, opened 2024-07-09
Fix nullref race in StringResources

MikeStall:
Fix for #2528 . See issue for description

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Localization/StringResources.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Localization/StringResources.cs
@@ -9,6 +9,7 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Resources;
+using System.Threading;
 using Microsoft.PowerFx.Core.Utils;
 
 namespace Microsoft.PowerFx.Core.Localization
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Core/Localization/StringResources.cs
```diff
@@ -230,18 +233,26 @@ bool IExternalStringResources.TryGetErrorResource(ErrorResourceKey resourceKey,
             // Error resources are a bit odd and need to be reassembled from separate keys.
             // Check to see if we've already retrieved one for this locale/key combo before 
             // rebuilding the full error resource.
-
-            if (_errorResources.TryGetValue(locale, out var localizedErrorResources))
+            try
             {
-                if (localizedErrorResources.TryGetValue(resourceKey.Key, out resourceValue))
+                _lock.EnterReadLock();
+                if (_errorResources.TryGetValue(locale, out var localizedErrorResources))
```
MikeStall, line 239: With the previous change, the lock around here was removed causing this to no longer be under the lock.  But we can't access a non-thread-safe collection in a multi-threaded way, even if it's a read. 

Use a reader-writer lock to minimize contention.

src/libraries/Microsoft.PowerFx.Core/Localization/StringResources.cs
```diff
@@ -108,9 +109,11 @@ internal class PowerFxStringResources : IExternalStringResources
 
         internal string ResourceLocation => _resourceManager.ResourceLocation;
 
-        [ThreadSafeProtectedByLockAttribute("_errorResources")]
+        [ThreadSafeProtectedByLock("_lock")]
         private static readonly Dictionary<string, Dictionary<string, ErrorResource>> _errorResources = new (StringComparer.OrdinalIgnoreCase);
```
MikeStall, line 113: Also concerned that this field is static.  But everybody holding onto the containing class here (PowerFxStringSources)  is static, so making it instance doesn't really buy us much. We should followup on this ...

jas-valgotar (approved):
:shipit:

lesaltzm (approved):

Shpakh-MSFT (approved):

Issue #2501, opened 2024-06-25
Support for Assigning Values to Untyped Record Field


gekarapa:
We need to extend the capabilities of the Untyped type to support field assignment. For example, implementing a function similar to `SetField(uo.field, "")`.

PR #2533, opened 2024-07-09
UO field set changes

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2501.

Allowing the following scenarios:
`Set(Index(uoTable, 1).Id, uoValue)`
`Set(Index(uoTable, 1).Id, 99)`

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/UntypedObjectValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/UntypedObjectValue.cs
@@ -30,6 +30,8 @@ namespace Microsoft.PowerFx.Types
 
         bool TryGetProperty(string value, out IUntypedObject result);
 
+        bool TrySetProperty(string name, IUntypedObject value);
+
         string GetString();
 
         double GetDouble();
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/UntypedObjectValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/UntypedObjectValue.cs
@@ -30,8 +30,6 @@ namespace Microsoft.PowerFx.Types
 
         bool TryGetProperty(string value, out IUntypedObject result);
 
-        bool TrySetProperty(string name, IUntypedObject value);
-
         string GetString();
 
         double GetDouble();
@@ -80,6 +78,34 @@ namespace Microsoft.PowerFx.Types
         {
             // Not supported for the time being.
             throw new NotImplementedException("UntypedObjectValue cannot be serialized.");
+        }
+    }
+
+    public abstract class UntypedObjectBase : IUntypedObject
+    {
+        public abstract IUntypedObject this[int index] { get; }
+
+        public abstract FormulaType Type { get; }
+
+        public abstract int GetArrayLength();
+
+        public abstract bool GetBoolean();
+
+        public abstract decimal GetDecimal();
+
+        public abstract double GetDouble();
+
+        public abstract string GetString();
+
+        public abstract string GetUntypedNumber();
+
+        public abstract bool TryGetProperty(string value, out IUntypedObject result);
+
+        public abstract bool TryGetPropertyNames(out IEnumerable<string> propertyNames);
+
+        public virtual bool TrySetProperty(string name, FormulaValue value)
+        {
+            throw new NotImplementedException();
         }
     }
 }
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/UntypedObjectValue.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/UntypedObjectValue.cs
@@ -79,31 +79,13 @@ namespace Microsoft.PowerFx.Types
             // Not supported for the time being.
             throw new NotImplementedException("UntypedObjectValue cannot be serialized.");
         }
-    }
 
-    public abstract class UntypedObjectBase : IUntypedObject
-    {
-        public abstract IUntypedObject this[int index] { get; }
+        internal virtual UntypedObjectValue CreateNew(IRContext irContext, IUntypedObject impl)
+        {
+            return new UntypedObjectValue(irContext, impl);
+        }
 
-        public abstract FormulaType Type { get; }
-
-        public abstract int GetArrayLength();
-
-        public abstract bool GetBoolean();
-
-        public abstract decimal GetDecimal();
-
-        public abstract double GetDouble();
-
-        public abstract string GetString();
-
-        public abstract string GetUntypedNumber();
-
-        public abstract bool TryGetProperty(string value, out IUntypedObject result);
-
-        public abstract bool TryGetPropertyNames(out IEnumerable<string> propertyNames);
-
-        public virtual bool TrySetProperty(string name, FormulaValue value)
+        public virtual bool TrySetProperty(string propertyName, FormulaValue value)
         {
             throw new NotImplementedException();
         }
--- a/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
@@ -167,15 +167,37 @@ namespace Microsoft.PowerFx
 
             var newValue = await arg1.Accept(this, context).ConfigureAwait(false);
 
-            if (arg0.IRContext.IsMutation && arg0 is RecordFieldAccessNode rfan)
+            if (arg0.IRContext.IsMutation)
             {
-                var arg0value = await rfan.From.Accept(this, context).ConfigureAwait(false);
-                
-                if (arg0value is RecordValue rv)
+                if (arg0 is RecordFieldAccessNode rfan)
                 {
-                    rv.ShallowCopyFieldInPlace(rfan.Field);
-                    rv.UpdateField(rfan.Field, newValue);
-                    return node.IRContext.ResultType._type.Kind == DKind.Boolean ? FormulaValue.New(true) : FormulaValue.NewVoid();
+                    var arg0value = await rfan.From.Accept(this, context).ConfigureAwait(false);
+
+                    if (arg0value is RecordValue rv)
+                    {
+                        rv.ShallowCopyFieldInPlace(rfan.Field);
+                        rv.UpdateField(rfan.Field, newValue);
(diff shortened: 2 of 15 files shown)


src/libraries/Microsoft.PowerFx.Core/Public/Values/UntypedObjectValue.cs
```diff
@@ -28,7 +28,9 @@ public interface IUntypedObject
         /// </summary>
         IUntypedObject this[int index] { get; }
 
-        bool TryGetProperty(string value, out IUntypedObject result);
+        bool TryGetProperty(string value, out IUntypedObject result);
+
+        bool TrySetProperty(string name, IUntypedObject value);
```
MikeStall, line 33: >IUntypedObject value [](http://example.com/codeflow?start=41&length=20)

This could be a FormulaValue

Ie, it's ok to Set(uo.field, 123) 
where 123 is a known strongly typed value.

MikeStall, line 33: >TrySetProperty [](http://example.com/codeflow?start=13&length=14)

Mecahnics issue: technically adding to interface is a breaking change. I'm wondering if we should add these as virtuals UntypedObjectValue instead. That's a class, so it's non-breaking. We can discuss offline

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -167,15 +167,37 @@ private async Task<FormulaValue> TryHandleSet(CallNode node, EvalVisitorContext
...
+                {
+                    var arg0value = await bon.Left.Accept(this, context).ConfigureAwait(false);
+
+                    if (arg0value is UntypedObjectValue uov)
+                    {
+                        TextLiteralNode textLiteralNode = (TextLiteralNode)bon.Right;
+                        UntypedObjectValue untypedObjectValue = (UntypedObjectValue)newValue;
+
+                        // !!!TODO What if property name does not exist?
+                        uov.Impl.TrySetProperty(textLiteralNode.LiteralValue, untypedObjectValue.Impl);
```
MikeStall, line 197: > untypedObjectValue.Impl [](http://example.com/codeflow?start=77&length=24)

just pass newValue here.  We can send FormulaValue straight through.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -167,15 +167,37 @@ private async Task<FormulaValue> TryHandleSet(CallNode node, EvalVisitorContext
...
+                else if (arg0 is BinaryOpNode bon && bon.Op == BinaryOpKind.DynamicGetField)
+                {
+                    var arg0value = await bon.Left.Accept(this, context).ConfigureAwait(false);
+
+                    if (arg0value is UntypedObjectValue uov)
+                    {
+                        TextLiteralNode textLiteralNode = (TextLiteralNode)bon.Right;
+                        UntypedObjectValue untypedObjectValue = (UntypedObjectValue)newValue;
+
+                        // !!!TODO What if property name does not exist?
```
MikeStall, line 196: >// !!!TODO What if property name does not exist? [](http://example.com/codeflow?start=24&length=48)

let the UO impl worry about this. 
May be we just allow expando. 

Key thing is that it can fail.  Let's just say host can throw a [https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Interpreter/Exceptions/CustomFunctionErrorException.cs](https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Interpreter/Exceptions/CustomFunctionErrorException.cs) and we catch that and convert to an errorValue.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/ArrayUntypedObject.cs
```diff
@@ -61,5 +61,10 @@ public bool TryGetPropertyNames(out IEnumerable<string> result)
             result = null;
             return false;
         }
+
+        public bool TrySetProperty(string name, IUntypedObject value)
+        {
+            throw new NotImplementedException();
+        }
```
MikeStall, line 68: All of these "Stub" impls are why it may be better to add as a virtual on the class, rather than the interface :)

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -167,15 +167,37 @@ private async Task<FormulaValue> TryHandleSet(CallNode node, EvalVisitorContext
...
+                        return CommonErrors.UnreachableCodeError(node.IRContext);
+                    }
+                }
+                else if (arg0 is BinaryOpNode bon && bon.Op == BinaryOpKind.DynamicGetField)
+                {
+                    var arg0value = await bon.Left.Accept(this, context).ConfigureAwait(false);
+
+                    if (arg0value is UntypedObjectValue uov)
+                    {
+                        TextLiteralNode textLiteralNode = (TextLiteralNode)bon.Right;
```
MikeStall, line 193: >(TextLiteralNode) [](http://example.com/codeflow?start=58&length=17)

How do we know this cast is safe? I guess it's implied by DynamicGetField check on line 187 

Nit - consider sharing  a static helper with 
OperatorDynamicGetField

That can also capture this safety check.

anderson-joyle: This is where Fx creates a `BinaryOpNode` and adds a `TextLiteralNode` to the right.
https://github.com/microsoft/Power-Fx/blob/b762be3ae8a45471ececad2d5b1b139361611b71/src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs#L787

MikeStall:
>                 if (info.Data is NameSymbol nameSymbol && nameSymbol.Props.CanSet)

even if arg0 is UO, we should still do this check. 

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/SetFunction.cs:121 in bf5df32. [](commit_id = bf5df32f42d7659f5cdfa6d1d6349f7d0506f3ff, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
MikeStall: bool IsDynamicSetProperty(arg0, out UntypedObjectValue uo, out string propName)

src/libraries/Microsoft.PowerFx.Core/Public/Values/UntypedObjectValue.cs
```diff
@@ -79,5 +79,33 @@ public override void ToExpression(StringBuilder sb, FormulaValueSerializerSettin
...
+        public abstract string GetUntypedNumber();
+
+        public abstract bool TryGetProperty(string value, out IUntypedObject result);
+
+        public abstract bool TryGetPropertyNames(out IEnumerable<string> propertyNames);
+
+        public virtual bool TrySetProperty(string name, FormulaValue value)
+        {
+            throw new NotImplementedException();
+        }
```
MikeStall, line 109: Let's just put this on UntypedObjectValue 

 and avoid the new class.

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/DateTimeUTCTests.cs
```diff
@@ -294,7 +294,7 @@ public FormulaValue Execute(StringValue val)
             }
         }
 
-        private class SimpleObject : IUntypedObject
+        private class SimpleObject : UntypedObjectBase
```
MikeStall, line 297: per other changes - we may not need any changes to this file.

src/libraries/Microsoft.PowerFx.Core/Public/Values/UntypedObjectValue.cs
```diff
@@ -78,6 +78,16 @@ public override void ToExpression(StringBuilder sb, FormulaValueSerializerSettin
         {
             // Not supported for the time being.
             throw new NotImplementedException("UntypedObjectValue cannot be serialized.");
+        }
+
+        internal virtual UntypedObjectValue CreateNew(IRContext irContext, IUntypedObject impl)
+        {
+            return new UntypedObjectValue(irContext, impl);
+        }
```
MikeStall, line 86: ohhh.... this is not ideal. I'll followup offline...

MikeStall:
>                     return new UntypedObjectValue(irContext, propertyValue);

Should we have called CreateNew() here? 

More broadly - the issue with CreateNew is that it's at odds with the UntypedObjectValue() ctor. 



---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUntypedObject.cs:458 in c92d9de. [](commit_id = c92d9de64e0f112b20803385bf9d902e8d0d5d85, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Public/Values/UntypedObjectValue.cs
MikeStall: after further exploration - that's too much churn. 
This approach is probably our best bet.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -167,15 +167,36 @@ private async Task<FormulaValue> TryHandleSet(CallNode node, EvalVisitorContext
...
+                }
+                else if (arg0 is BinaryOpNode bon && bon.Op == BinaryOpKind.DynamicGetField)
+                {
+                    var arg0value = await bon.Left.Accept(this, context).ConfigureAwait(false);
+
+                    if (arg0value is UntypedObjectValue uov && uov.Impl is UntypedObjectBase impl)
+                    {
+                        TextLiteralNode textLiteralNode = (TextLiteralNode)bon.Right;
+
+                        // !!!TODO What if property name does not exist?
```
MikeStall: > // !!!TODO What if property name does not exist? [](http://example.com/codeflow?start=23&length=49)


void SetProperty(string, formulaValue);
Throw on error. 

Encourage to throw: 
CustomFunctionErrorException

See later in this file how we catch that and convert to an ErrorValue.

MikeStall:
Superceded by #2547

Issue #2538, opened 2024-07-10
Regular expressions don't honor culture settings


gregli-msft:
All of our regular expression functions use the Invariant culture, ignoring the host's set culture, with no way to choose a culture otherwise (like the 2nd argument to Value).  This has implications for case insensitive matching.  We need to rationalize.

PR #2539, opened 2024-07-10
Culture support for in operator and Sort function

gregli-msft:
The `in` operator and `Sort` function were not culturally aware, resulting in incorrect results for Turkish and other I'm sure other languages.  

The `Lower`, `Upper`, `Proper`, and `SortByColumns` functions are already culturally aware.  

Our regular expression functions are not culturally aware, but that is an interesting design choice that is the subject of https://github.com/microsoft/Power-Fx/issues/2538.

For testing, this PR adds the ability to change the culture in this local REPL only, with a `Language` function overload that takes a language tag as the first argument.  Also added is a `#SETUP` handler to specify the culture for a .txt test file.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryOperators.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryOperators.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using System;
+using System.Globalization;
 using System.Linq;
 using System.Text;
 using Microsoft.PowerFx.Core.IR;
@@ -720,9 +721,9 @@ namespace Microsoft.PowerFx.Functions
         }
 
         // See in_SS in JScript membershipReplacementFunctions
-        public static Func<IRContext, FormulaValue[], FormulaValue> StringInOperator(bool exact)
+        public static Func<IServiceProvider, IRContext, FormulaValue[], FormulaValue> StringInOperator(bool exact)
         {
-            return (irContext, args) =>
+            return (services, irContext, args) =>
             {
                 var left = args[0];
                 var right = args[1];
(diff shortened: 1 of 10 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryOperators.cs
```diff
@@ -737,23 +738,25 @@ public static Func<IRContext, FormulaValue[], FormulaValue> StringInOperator(boo
 
                 var leftStr = (StringValue)left;
                 var rightStr = (StringValue)right;
-
-                return new BooleanValue(irContext, rightStr.Value.IndexOf(leftStr.Value, exact ? StringComparison.Ordinal : StringComparison.OrdinalIgnoreCase) >= 0);
```
CarlosFigueiraMSFT, line 741: Does `services` always have a CultureInfo value? In other places where I see CultureInfo being accessed (mostly in LibraryText.txt) it comes from the runner directly, not from the service provider.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -1297,8 +1297,16 @@ private static FormulaValue SortValueType<TPFxPrimitive, TDotNetPrimitive>(List<
                 }
 
                 var n1 = a.sortValue as TPFxPrimitive;
-                var n2 = b.sortValue as TPFxPrimitive;
-                return n1.Value.CompareTo(n2.Value) * compareToResultModifier;
+                var n2 = b.sortValue as TPFxPrimitive;
+                CultureInfo culture;
+                if (n1.Value is string n1s && n2.Value is string n2s && (culture = runner.GetService<CultureInfo>()) != null)
```
CarlosFigueiraMSFT, line 1302: >.GetService<CultureInfo>() [](http://example.com/codeflow?start=89&length=26)

You can access `runner.CultureInfo` directly

CarlosFigueiraMSFT (dismissed):
:shipit:

PR #2541, opened 2024-07-11
Adding optional range param to custom fx2nl param

NavneetThekkumpat:
Update the expected params for Fx2Nl (CustomFx2NLParams) with an optional range field.
This will be used and passed down for the explain this partial formula explanation.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/CustomFx2NLParams.cs
+++ b/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/CustomFx2NLParams.cs
@@ -21,6 +21,11 @@ namespace Microsoft.PowerFx.LanguageServerProtocol.Protocol
         /// Existing Power Fx Expression.
         /// </summary>
         public string Expression { get; set; }
+
+        /// <summary>
+        /// Optional range of the expression in the document.
+        /// </summary>
+        public Range Range { get; set; } = null;
     }
 
     /// <summary>
--- a/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/CustomFx2NLParams.cs
+++ b/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/CustomFx2NLParams.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using System;
+using System.Linq.Expressions;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -23,9 +24,10 @@ namespace Microsoft.PowerFx.LanguageServerProtocol.Protocol
         public string Expression { get; set; }
 
         /// <summary>
-        /// Optional range of the expression.
+        /// Optional range within <see cref="Expression"></see> to explain.
+        /// If missing, then explain the whole expression.
         /// </summary>
-        public Range Range { get; set; } = null;
+        public Range Range { get; set; }
     }
 
     /// <summary>
--- a/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/CustomFx2NLParams.cs
+++ b/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/CustomFx2NLParams.cs
@@ -23,7 +23,7 @@ namespace Microsoft.PowerFx.LanguageServerProtocol.Protocol
         public string Expression { get; set; }
 
         /// <summary>
-        /// Optional range of the expression in the document.
+        /// Optional range of the expression.
         /// </summary>
         public Range Range { get; set; } = null;
     }
--- a/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/CustomFx2NLParams.cs
+++ b/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/CustomFx2NLParams.cs
@@ -26,8 +26,9 @@ namespace Microsoft.PowerFx.LanguageServerProtocol.Protocol
         /// <summary>
         /// Optional range within <see cref="Expression"></see> to explain.
         /// If missing, then explain the whole expression.
+        /// Eg: [0,0,0,6] >> start line 0, start character 0, end line 0, end character 6.
         /// </summary>
-        public Range Range { get; set; }
+        public uint[] Range { get; set; }
     }
 
     /// <summary>
--- a/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/CustomFx2NLParams.cs
+++ b/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/CustomFx2NLParams.cs
@@ -26,7 +26,7 @@ namespace Microsoft.PowerFx.LanguageServerProtocol.Protocol
         /// <summary>
         /// Optional range within <see cref="Expression"></see> to explain.
         /// If missing, then explain the whole expression.
-        /// Eg: {1,6} >> 1d starting index is 1 and ending index is 6.
+        /// Eg: zero-based offsets {1,6} >> 1d starting index is 1 and ending index is 6.
         /// </summary>
         public ScalarRange Range { get; set; }
     }
--- a/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/CustomFx2NLParams.cs
+++ b/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/CustomFx2NLParams.cs
@@ -26,9 +26,9 @@ namespace Microsoft.PowerFx.LanguageServerProtocol.Protocol
         /// <summary>
         /// Optional range within <see cref="Expression"></see> to explain.
         /// If missing, then explain the whole expression.
-        /// Eg: [0,0,0,6] >> start line 0, start character 0, end line 0, end character 6.
+        /// Eg: {1,6} >> 1d starting index is 1 and ending index is 6.
         /// </summary>
-        public uint[] Range { get; set; }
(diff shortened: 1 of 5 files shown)


src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/CustomFx2NLParams.cs
```diff
@@ -20,7 +20,12 @@ public class CustomFx2NLParams : BaseNLParams, IHasTextDocument
         /// <summary>
         /// Existing Power Fx Expression.
         /// </summary>
-        public string Expression { get; set; }
+        public string Expression { get; set; }
+
+        /// <summary>
+        /// Optional range of the expression.
+        /// </summary>
+        public Range Range { get; set; } = null;
```
MikeStall, line 28: >= null; [](http://example.com/codeflow?start=41&length=7)

nit - this is the default behavior, so not required.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/CustomFx2NLParams.cs
```diff
@@ -20,7 +20,12 @@ public class CustomFx2NLParams : BaseNLParams, IHasTextDocument
         /// <summary>
         /// Existing Power Fx Expression.
         /// </summary>
-        public string Expression { get; set; }
+        public string Expression { get; set; }
+
+        /// <summary>
+        /// Optional range of the expression.
```
MikeStall, line 26: >range of the expression. [](http://example.com/codeflow?start=21&length=24)

doc comments can refer to other fields :) Something like <see>

Comment: 
range within <see="Expression">  to explain.
If missing, then explain the whole expression.

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/LanguageServiceProtocol/LegacyLanguageServerTests.cs
```diff
@@ -1869,6 +1871,45 @@ public void TestFx2NL()
             var response = JsonSerializer.Deserialize<JsonRpcFx2NLResponse>(_sendToClientData[0], _jsonSerializerOptions);
             Assert.Equal("123", response.Id);
 
+            // result has expected concat with symbols. 
+            Assert.Equal("Score > 3: True: sentence", response.Result.Explanation);
+        }
+
+        [Fact]
+        public void TestFx2NLWithRangeRequestParam()
```
akshar2401, line 1879: avoid adding any tests here. Look at the comment above the class name

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/LanguageServiceProtocol/RedesignedLanguageServerTests/Fx2NLTests.cs
```diff
@@ -202,8 +203,43 @@ public async Task TestFx2NLThrows()
             AssertErrorPayload(rawResponse, payload.id, JsonRpcHelper.ErrorCode.InternalError);
             Assert.NotEmpty(TestServer.UnhandledExceptions);
         }
+
+        [Fact]
+        public async Task TestFx2NLWithRequestParamRange()
+        {
```
akshar2401, line 209: What does this test accomplish? Do we really need to test passing of Range property? There's no functionality being added in this pr. Just adding another property. Imo, there's no point in testing the serialization or its opposite.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/CustomFx2NLParams.cs
akshar2401, line 28: Are you sure about using Ranges? PowerFx uses one dimensional indexing. It would be expensive to convert 2d ranges to 1d indices on the server side as you would need to go through whole expression. If you make this 1d, (start and end index both are 1d) then you can convert the 2d indices on the client side very cheaply with a better time complexity. This is not a native protocol so there's nothing that binds us with 2d ranges. Consider passing flatten 1d index into the expression to avoid any conversions (which you must do since Fx is 1d based) on the server side

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/LanguageServiceProtocol/RedesignedLanguageServerTests/Fx2NLTests.cs
NavneetThekkumpat: Yup, thinking more about it we do not necessarily need these tests . But left it to show that user can pass in optional param.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/CustomFx2NLParams.cs
NavneetThekkumpat: I thought about this one. Since the Range can take in line number and char (zero indexing), 
image
 I thought we could pass it in this format from the front end.

akshar2401: I am not pointing to structuring the range. I am taking about the a position in an range being 1d index or 2d index (row and column vs flatten index)

An 1d index 5 points to 6th character in expression (0 based)
A 2d representation of 1d index 5 is line (row): 1 and column 5

Go from Range(Start: Position(line, row), End: Position(line, row)) to Range(Start: Index, End: Index). you can easily and very conveniently calculate 1d indices on the client side as opposed to server

NavneetThekkumpat: I see what you mean, in that case we can follow the same indexing like the one used in publishcontroltokens for control highlighting work before. 
I can keep it as unit[]   // Eg: [0,0,0,6] >> start line 0, start character 0, end line 0, end character 6.

akshar2401: uint[] was done for response compression for semantic tokens. Here, you should introduce a class called 1DRange (better name please) and have two props inside it

class 1DRange {
  public int Start {get; set;}

  public int End {get; set;} 

}

and then use this type for your "range" param

on the UFB side of things, do model.getOffsetAt(2DRange) to get 1d start and 1d end

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/ScalarRange.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+namespace Microsoft.PowerFx.LanguageServerProtocol.Protocol
+{
+    public class ScalarRange
+    {
+        public ScalarRange(int start, int end)
+        {
```
akshar2401, line 9: nit: you don't need the constructor. Serializer would be creating this object during deserialization and you won't need to construct this manually. remove the constructor

NavneetThekkumpat: Sure, if needed for testing, we can add later.

akshar2401 (approved):

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/ScalarRange.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+namespace Microsoft.PowerFx.LanguageServerProtocol.Protocol
+{
+    public class ScalarRange
+    {
+        public bool IsValid()
```
MikeStall, line 8: nit - should this also check >= 0 ?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Protocol/ScalarRange.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        public bool IsValid()
+        {
+            return Start <= End;
+        }
+
+        /// <summary>
+        /// The range's start position.
+        /// </summary>
+        public int Start { get; set; }
```
MikeStall, line 16: >int [](http://example.com/codeflow?start=15&length=3)

nit - can we comment: is this 0-based or 1-based?

MikeStall:
> namespace Microsoft.PowerFx.Tests.LanguageServiceProtocol.Tests

is there a change in this file?

---
Refers to: src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/LanguageServiceProtocol/LegacyLanguageServerTests.cs:37 in 74bb146. [](commit_id = 74bb146d81dda163eb2492533ec524f06f5883b9, deletion_comment = False)

NavneetThekkumpat:
> > namespace Microsoft.PowerFx.Tests.LanguageServiceProtocol.Tests
> 
> is there a change in this file?
> 
> Refers to: src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/LanguageServiceProtocol/LegacyLanguageServerTests.cs:37 in [74bb146](https://github.com/microsoft/Power-Fx/commit/74bb146d81dda163eb2492533ec524f06f5883b9). [](commit_id = [74bb146](https://github.com/microsoft/Power-Fx/commit/74bb146d81dda163eb2492533ec524f06f5883b9), deletion_comment = False)

 Initially I had a test but removed the changes now.

PR #2547, opened 2024-07-15
UO set property

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2501.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
+++ b/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
@@ -19,6 +19,7 @@ using Microsoft.PowerFx.Interpreter;
 using Microsoft.PowerFx.Interpreter.Exceptions;
 using Microsoft.PowerFx.Types;
 using static Microsoft.PowerFx.Functions.Library;
+using static Microsoft.PowerFx.Syntax.PrettyPrintVisitor;
 
 namespace Microsoft.PowerFx
 {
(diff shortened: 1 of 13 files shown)


src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -154,28 +155,54 @@ public override async ValueTask<FormulaValue> Visit(LazyEvalNode node, EvalVisit
...
+                    if (arg0value is UntypedObjectValue uov && uov.Impl is UntypedObjectBase impl)
+                    {
+                        TextLiteralNode textLiteralNode = (TextLiteralNode)bon.Right;
+
+                        try
+                        {
+                            impl.SetProperty(textLiteralNode.LiteralValue, newValue);
+                            return node.IRContext.ResultType._type.Kind == DKind.Boolean ? FormulaValue.New(true) : FormulaValue.NewVoid();
+                        }
+                        catch (CustomFunctionErrorException ex)
```
MikeStall, line 201: >catch (CustomFunctionErrorException ex) [](http://example.com/codeflow?start=24&length=39)

since default impl is NotImplementedException, can we also catch that?

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/PadUntypedObjectTests.cs
```diff
@@ -45,9 +45,51 @@ public void PadUntypedObjectTest()
             Assert.Equal(FormulaType.UntypedObject, fv3.Type);
             Assert.True(fv3 is BlankValue);
         }
+
+        [Fact]
```
MikeStall, line 49: We should add test with ParseJson(). 

Ie, what does this do?
Set(ParseJson("{x:5}").x, 99)

anderson-joyle: `Error 22-24: The value passed to the 'Set' function cannot be changed.`

MikeStall (approved):
:shipit:

Issue #2502, opened 2024-06-25
Support for Assigning Values to to an untyped record field using an index


gekarapa:
We need to extend the capabilities of the Untyped type to support field assignment using field index. For example, implementing a function similar to `SetValue(uoRecord, index, value)`.

PR #2548, opened 2024-07-16
Set untyped object with Index function

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2502.

Supporting `Set(Index(uo,1),value)`


src/libraries/Microsoft.PowerFx.Core/Public/Values/UntypedObjectValue.cs
```diff
@@ -106,9 +106,9 @@ public abstract class UntypedObjectBase : IUntypedObject
         /// <summary>
         /// Set a property on the object.
         /// </summary>
-        /// <param name="propertyName">Property name.</param>
+        /// <param name="property">Property identifier. Usually a number or name.</param>
         /// <param name="value">FormulaValue to be set.</param>
-        public virtual void SetProperty(string propertyName, FormulaValue value)
+        public virtual void SetProperty(FormulaValue property, FormulaValue value)
```
MikeStall, line 111: >FormulaValue [](http://example.com/codeflow?start=40&length=12)

We can add a new virtual for this and stay strongly typed.

src/libraries/Microsoft.PowerFx.Core/Public/Values/UntypedObjectValue.cs
```diff
@@ -106,9 +106,9 @@ public abstract class UntypedObjectBase : IUntypedObject
         /// <summary>
         /// Set a property on the object.
         /// </summary>
-        /// <param name="propertyName">Property name.</param>
+        /// <param name="property">Property identifier. Usually a number or name.</param>
```
LucGenetier, line 109: >Property identifier. Usually a number or name. [](http://example.com/codeflow?start=35&length=46)

Why can't we keep string and use "1" when it's a number?

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -188,24 +188,21 @@ private async Task<FormulaValue> TryHandleSet(CallNode node, EvalVisitorContext
...
 
                     if (arg0value is UntypedObjectValue uov && uov.Impl is UntypedObjectBase impl)
                     {
-                        TextLiteralNode textLiteralNode = (TextLiteralNode)bon.Right;
+                        return Utility.SetUntypedObject(node.IRContext, impl, arg1value, newValue);
+                    }
+                }
+                else if (arg0 is CallNode callNode && callNode.Function == BuiltinFunctionsCore.Index_UO)
+                {
+                    var child0Value = await callNode.Args[0].Accept(this, context).ConfigureAwait(false);
```
LucGenetier, line 200: >Args[0] [](http://example.com/codeflow?start=53&length=7)

nit: could reuse arg0/arg1 variables

src/libraries/Microsoft.PowerFx.Core/Public/Values/UntypedObjectValue.cs
MikeStall: just add a new virtual. 
SetProperty(string ) 
SetIndex(int i) 

This simplifies responsibly on the host that has to implement these.  We don't want every host having to check for both cases.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Utility.cs
```diff
@@ -30,5 +35,34 @@ public static T GetService<T>(this IServiceProvider serviceProvider)
         {
             return (T)serviceProvider.GetService(typeof(T));
         }
+
+        public static FormulaValue SetUntypedObject(IRContext context, UntypedObjectBase untypedObject, FormulaValue property, FormulaValue value)
```
MikeStall, line 39: nit - this could be an extension method on untypedObject. That would make discoverability a bit easier

src/libraries/Microsoft.PowerFx.Core/Public/Values/UntypedObjectValue.cs
```diff
@@ -106,12 +106,23 @@ public abstract class UntypedObjectBase : IUntypedObject
...
             // In case of unwanted behavior, throw an CustomFunctionErrorException exception.
             throw new NotImplementedException();
         }
+
+        /// <summary>
+        /// Set a property on the object.
+        /// </summary>
+        /// <param name="index">Property identifier.</param>
+        /// <param name="value">FormulaValue to be set.</param>
+        public virtual void SetProperty(int index, FormulaValue value)
```
MikeStall, line 122: >index [](http://example.com/codeflow?start=44&length=5)

we should comment whether this is 0-based or 1-based.

anderson-joyle: Isn't this up to the host to decide?

src/libraries/Microsoft.PowerFx.Interpreter/Utility.cs
anderson-joyle: Good suggestion. Done.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Values/UntypedObjectValue.cs
MikeStall: Or simpler .. what does this return? 
First(ParseJSON("[10,20,30]"))

src/libraries/Microsoft.PowerFx.Interpreter/Utility.cs
```diff
@@ -30,5 +35,34 @@ public static T GetService<T>(this IServiceProvider serviceProvider)
...
+            try
+            {
+                if (property is StringValue stringValue)
+                {
+                    untypedObject.SetProperty(stringValue.Value, value);
+                }
+                else if (property is NumberValue numberValue)
+                {
+                    untypedObject.SetProperty((int)numberValue.Value, value);
+                }
```
MikeStall, line 50: We should avoid switching logic like this.  It makes it too tempting to allow wrong behavior. 

Safest to just have 2 separate methods. 

Or if we're trying to share wrapping, we can use a callback, 
SetUntypedAction(...,  Action worker) 

where worker can either be Set by nam, or set by index.

src/libraries/Microsoft.PowerFx.Core/Public/Values/UntypedObjectValue.cs
MikeStall, line 122: >SetProperty [](http://example.com/codeflow?start=28&length=11)

Just to avoid confusion, we should call this a different name.  like:
SetIndex() 

And have a test explicitly that 
Set(uo.'1') calls SetProperty and not gets smart and calls SetIndex.

src/libraries/Microsoft.PowerFx.Interpreter/Utility.cs
```diff
@@ -30,5 +35,34 @@ public static T GetService<T>(this IServiceProvider serviceProvider)
...
+        {
+            try
+            {
+                if (property is StringValue stringValue)
+                {
+                    untypedObject.SetProperty(stringValue.Value, value);
+                }
+                else if (property is NumberValue numberValue)
+                {
+                    untypedObject.SetProperty((int)numberValue.Value, value);
```
MikeStall, line 49: >(int)numberValue.Value [](http://example.com/codeflow?start=46&length=22)

Do we know this case is safe? Ie, it's not a float or decimal coming in.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #2551, opened 2024-07-17
Fix 2523 - Improve Json deserialization errors and be more permisive

LucGenetier:
Json deserialization can be "broken" with some exceptions which are impossible to understand for customers
This PR fixes/updates the error messages and wraps the Json deserialization issues into an ErrorValue instead of being inconsistent (previously, we would have an ErrorValue for invalid Json and a NotImplementedException if there was a type mismatch between what the connector code expects and what we got from the server)
We also accept some trivial mismatches
Add ErrorKind.InvalidJson
Add HttpExpressionError to return StatusCode in ErrorValue

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
@@ -459,11 +459,10 @@ namespace Microsoft.PowerFx.Connectors
             }
 
             return FormulaValue.NewError(
-                    new ExpressionError()
+                    new HttpExpressionError(statusCode)
                     {
                         Kind = ErrorKind.Network,
                         Severity = ErrorSeverity.Critical,
-                        Span = new Syntax.Span(statusCode, statusCode), // Return HTTP Status code in Span
                         Message = $"The server returned an HTTP error with code {statusCode}{reasonPhrase}. Response: {text}"
                     },
                     _function.ReturnType);
--- a/src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
@@ -463,6 +463,7 @@ namespace Microsoft.PowerFx.Connectors
                     {
                         Kind = ErrorKind.Network,
                         Severity = ErrorSeverity.Critical,
+                        Span = new Syntax.Span(statusCode, statusCode), // Return HTTP Status code in Span
                         Message = $"The server returned an HTTP error with code {statusCode}{reasonPhrase}. Response: {text}"
                     },
                     _function.ReturnType);
--- a/src/libraries/Microsoft.PowerFx.Core/Public/ErrorKind.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/ErrorKind.cs
@@ -39,6 +39,7 @@ namespace Microsoft.PowerFx
         NotApplicable = 27,
         Timeout = 28,
         ServiceUnavailable = 29,
+        InvalidJson = 30,
         Custom = 1000,
     }
 }
--- a/src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
+++ b/src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
@@ -22,14 +22,14 @@ namespace Microsoft.PowerFx.Types
     public class FormulaValueJSON
     {
         /// <summary>
-        /// Convenience method to create a value from a json representation. 
+        /// Convenience method to create a value from a json representation.
         /// </summary>
-        public static FormulaValue FromJson(string jsonString, FormulaType formulaType = null, bool numberIsFloat = false)
+        public static FormulaValue FromJson(string jsonString, FormulaType formulaType = null, bool numberIsFloat = false, string path = null)
         {
-            return FromJson(jsonString, new FormulaValueJsonSerializerSettings() { NumberIsFloat = numberIsFloat }, formulaType);
+            return FromJson(jsonString, new FormulaValueJsonSerializerSettings() { NumberIsFloat = numberIsFloat }, formulaType, path);
         }
-      
-        public static FormulaValue FromJson(string jsonString, FormulaValueJsonSerializerSettings settings, FormulaType formulaType = null)
+
+        public static FormulaValue FromJson(string jsonString, FormulaValueJsonSerializerSettings settings, FormulaType formulaType = null, string path = null)
         {
             try
             {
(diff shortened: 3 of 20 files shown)


src/libraries/Microsoft.PowerFx.Json/PowerFxJsonException.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+        }
+
+        public PowerFxJsonException(string message)
+            : base(message)
+        {
+        }
+
+        public PowerFxJsonException(string message, string name)
+            : base($"{message}{(string.IsNullOrEmpty(name) ? string.Empty : $", in {name}")}")
```
Robulane, line 23: rename name to path for better understanding

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -39,28 +39,38 @@ public static FormulaValue FromJson(string jsonString, FormulaValueJsonSerialize
...
                     Kind = ErrorKind.Network
                 });
             }
+            catch (PowerFxJsonException pfxje)
+            {
+                return new ErrorValue(IRContext.NotInSource(formulaType), new ExpressionError()
+                {
+                    Message = $"{pfxje.GetType().Name} {pfxje.Message}",
+                    Span = new Syntax.Span(0, 0),
+                    Kind = ErrorKind.Network
```
Robulane, line 56: ideally we would love to have a different kind but if it's too hard I think we can live without it as well

LucGenetier: Will add ErrorKind.InvalidJson here

Robulane (dismissed):

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -463,6 +463,7 @@ public async Task<FormulaValue> DecodeResponseAsync(HttpResponseMessage response
                     {
                         Kind = ErrorKind.Network,
                         Severity = ErrorSeverity.Critical,
+                        Span = new Syntax.Span(statusCode, statusCode), // Return HTTP Status code in Span
```
MikeStall, line 466: >// Return HTTP Status code in Span [](http://example.com/codeflow?start=72&length=34)

?
This seems like a mismatch.  Why is Message not enough?

MikeStall:
>     /// This must be kept in sync with the document enum "ErrorKind".

Is there something we need to do , per this comment?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/ErrorKind.cs:8 in b5b31e4. [](commit_id = b5b31e4c3812ea2a8b48742db20d340c3aa930ff, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -22,14 +22,14 @@ public class FormulaValueJsonSerializerSettings
     public class FormulaValueJSON
     {
         /// <summary>
-        /// Convenience method to create a value from a json representation. 
+        /// Convenience method to create a value from a json representation.
         /// </summary>
-        public static FormulaValue FromJson(string jsonString, FormulaType formulaType = null, bool numberIsFloat = false)
+        public static FormulaValue FromJson(string jsonString, FormulaType formulaType = null, bool numberIsFloat = false, string path = null)
```
MikeStall, line 27: > string path = null [](http://example.com/codeflow?start=122&length=19)

This is a breaking change. 
Why isn't the Pr pipeline flagging it.

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
LucGenetier: Callers have to parse the message which is not efficient

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
LucGenetier: Investigating

LucGenetier:
>     /// This must be kept in sync with the document enum "ErrorKind".

Done in BuiltInEnums.cs

---
In reply to: [2233509084](https://github.com/microsoft/Power-Fx/pull/2551#issuecomment-2233509084) [](http://example.com/codeflow?ancestors=2233509084)

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/ErrorKind.cs:8 in b5b31e4. [](commit_id = b5b31e4c3812ea2a8b48742db20d340c3aa930ff, deletion_comment = False)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.FormulaValueJSON.FromJson(System.String,Microsoft.PowerFx.Types.FormulaType,System.Boolean)<br />❌CP0002 M:Microsoft.PowerFx.Types.FormulaValueJSON.FromJson(System.String,Microsoft.PowerFx.Types.FormulaValueJsonSerializerSettings,Microsoft.PowerFx.Types.FormulaType)<br />❌CP0002 M:Microsoft.PowerFx.Types.FormulaValueJSON.FromJson(System.Text.Json.JsonElement,Microsoft.PowerFx.Types.FormulaType,System.Boolean)<br />❌CP0002 M:Microsoft.PowerFx.Types.FormulaValueJSON.FromJson(System.Text.Json.JsonElement,Microsoft.PowerFx.Types.FormulaValueJsonSerializerSettings,Microsoft.PowerFx.Types.FormulaType)

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
LucGenetier: Resolved

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
MikeStall: This seems very hacky to have Span sometimes be a range and sometimes be a http error code. 

Do callers need to handle http codes differently?

LucGenetier: Good idea

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/ErrorKind.cs
```diff
@@ -39,6 +39,7 @@ public enum ErrorKind
         NotApplicable = 27,
         Timeout = 28,
         ServiceUnavailable = 29,
+        InvalidJson = 30,
```
MikeStall, line 42: >InvalidJson = 30, [](http://example.com/codeflow?start=8&length=17)

Should ParseJSON() use this?

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/ErrorKind.cs
LucGenetier: The objectives are too different, I wouldn't try merging them

LucGenetier: I don't think so, this is really for connectors, when the result coming back from the connector (directly or via APIM) is invalid in a way or another (invalid Json like missing '}', or logically invalid like receiving a string instead of a double)

LucGenetier:
✅ No public API change.

PR #2552, opened 2024-07-17
Fix: Null Handling in NodeKind Intellisense

adithyaselv:
CurNode in IntellisenseData can be Null in different scenarios and we have a NullNodeIntellisense Handler for this specific reason which suggests global/ in scope Values  , However this may not be applicable to all the cases such as suggesting types for UDF.

Not handling null here forces us to have a very strict order in running intellisense handlers , otherwise raising an exception which leads to no Intellisense results.

It is safe for us to assume CurNode could be null and skip the current suggestion handler instead of throwing an exception.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

Shpakh-MSFT (dismissed):

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

rick-nguyen (approved):

LucGenetier:
✅ No public API change.

Issue #2503, opened 2024-06-25
Support for Reading Untyped Record's field Using Index


gekarapa:
We need to extend the capabilities of the Untyped type to support reading a field using its index. For example, implementing a function similar to `ReadProperty(uoRecord, index)`.

PR #2555, opened 2024-07-19
Read untyped value with Index function

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2503.
This scenario is already supported. Adding test cases.


LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #2559, opened 2024-07-23
Enable UDF on REPL

### LINKED ISSUE #2546: UDFs should have better compile time error handling
if I give it a bad type, it doesn't produce a standard compile time error like the above.  We should give a "type not found" error.  For a common mistake here, like using `String` or `string` we should probably suggest Text instead.

image

adithyaselv:
Adds ability to define User-defined functions in REPL.

Parsing UDFs requires a different parser from the Texl Expression parser and this is addressed by using the DefinitionsParser as a fallback to check when regular parsing fails. If definitions parsing is successful and we find UDF, we add it to the Engine.

image

Fixes #2546


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -214,7 +214,7 @@ public static IEnumerable<UserDefinedFunction> CreateFunctions(IEnumerable<UDF>
                 Contracts.Assert(udf.IsParseValid);
 
                 var udfName = udf.Ident.Name;
-                if (texlFunctionSet.AnyWithName(udfName))
+                if (texlFunctionSet.AnyWithName(udfName) || nameResolver.Functions.WithName(udfName).Any(func => func is UserDefinedFunction))
```
MikeStall, line 217: >.Any(func => func is UserDefinedFunction) [](http://example.com/codeflow?start=100&length=41)

Why this part?

Whether a function is a UDF or something else (builtin, reflectionFunction, etc) shouldn't impact our binding rules.

src/libraries/Microsoft.PowerFx.Core/Parser/ParserOptions.cs
```diff
@@ -52,6 +52,11 @@ public class ParserOptions
         /// </summary>
         internal bool AllowParseAsTypeLiteral { get; set; }
 
+        /// <summary>
+        /// Flag to allow single user-definitons to be parsed without semicolon in the end.
+        /// </summary>
+        public bool AllowSingleUserDefinition { get; set; }
```
MikeStall, line 58: >AllowSingleUserDefinition [](http://example.com/codeflow?start=20&length=25)

Is the intention this flag is always true?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
adithyaselv: this is to check for duplicates, If an UDF with same name is defined in the symbols , we error

src/libraries/Microsoft.PowerFx.Core/Parser/ParserOptions.cs
adithyaselv: I don't have a strong opinion on this but we need to have this today as PowerApps requires semicolon. I wish this was like AllowChainExpressions in which `;` is optional . For this to be eventually true / have  `AllowChainExpressions` like behavior here , we need to confirm that this is fine with PowerApps and bit more UserDefinition parser changes

MikeStall: But changing from error (semi is require) to non-error (it's not required) is not a breaking change...

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
MikeStall: by why specifically UDF, as opposed to any derived TexlFunction with the same name conflict.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -214,7 +214,7 @@ public static IEnumerable<UserDefinedFunction> CreateFunctions(IEnumerable<UDF>
...
                 {
                     errors.Add(new TexlError(udf.Ident, DocumentErrorSeverity.Severe, TexlStrings.ErrUDF_FunctionAlreadyDefined, udfName));
```
MikeStall, line 219: >  errors.Add(new TexlError(udf.Ident, DocumentErrorSeverity.Severe, TexlStrings.ErrUDF_FunctionAlreadyDefined, udfName)); [](http://example.com/codeflow?start=18&length=121)

for collision detection, could we push this to the caller?
And not create a local set. 
   var texlFunctionSet = new TexlFunctionSet();

MikeStall:
>                 texlFunctionSet.Add(func);

see above - we should remove this too. Only call texlFunctionSet.Add() once. 

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs:249 in 4169003. [](commit_id = 4169003ab9036c970afcfb4a8e196a1511a272da, deletion_comment = False)

MikeStall:
>                     nameResolver.Functions.WithName(udfName).Any(func => func.IsRestrictedUDFName))

suspicious?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs:223 in 4169003. [](commit_id = 4169003ab9036c970afcfb4a8e196a1511a272da, deletion_comment = False)

MikeStall:
>                     nameResolver.Functions.WithName(udfName).Any(func => func.IsRestrictedUDFName))

talked more. This is also used in other PA Client functions like IsType,Set,UpdateContext ... basically anything that can implicitly define variables. 


---
In reply to: [2253638869](https://github.com/microsoft/Power-Fx/pull/2559#issuecomment-2253638869) [](http://example.com/codeflow?ancestors=2253638869)

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs:223 in 4169003. [](commit_id = 4169003ab9036c970afcfb4a8e196a1511a272da, deletion_comment = False)

gregli-msft:
Is the parser assuming that the open curly braces is a function body?  The "a:" isn't valid as the start of a formula, and the ".a" on the end is a dead giveaway too.  We will need this to work (without the ".a" on the end) when we have UDFs that return typed records.
```
>> rr():Text = {a:"asdf"}.a
Error 4-5: Unexpected characters. Characters are used in the formula in an unexpected way.
Error 0-4: 'rr' is an unknown or unsupported function.

>> rr():Text = Proper( {a:"asdf"}.a )

>> rr()
"Asdf"
```

gregli-msft:
As discussed in chat, this is very odd, as a Text string should not be returnable from this function.  As we discussed, we may want to just block Void for now, both as an input and output type.
```
>> n(x:Number):Void = "asdf"

>> n(3)
"asdf"
```

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/DefinitionsCheckResult.cs
```diff
@@ -43,6 +51,7 @@ public class DefinitionsCheckResult : IOperationStatus
 
         public DefinitionsCheckResult()
         {
+            _localSymbolTable = new SymbolTable();
```
MikeStall, line 54: >new SymbolTable(); [](http://example.com/codeflow?start=32&length=18)

nit, we can set DebugName here to aid in keeping symbol tables straight. 

new SymbolTable { DebugName = "FunctionDefinitions" }

src/libraries/Microsoft.PowerFx.Core/Public/DefinitionsCheckResult.cs
```diff
@@ -125,16 +137,94 @@ internal IReadOnlyDictionary<DName, FormulaType> ApplyResolveTypes()
...
+
+            if (_userDefinedFunctions == null)
+            {
+                _userDefinedFunctions = new TexlFunctionSet();
+
+                var validUdfs = new List<UDF>();
+
+                foreach (var udf in _parse.UDFs.Where(udf => udf.IsParseValid))
+                {
+                    if (_symbols.Functions.WithName(udf.Ident.Name).Any(func => func is UserDefinedFunction))
```
MikeStall, line 165: >func is UserDefinedFunction) [](http://example.com/codeflow?start=80&length=28)

Can we avoid checking the type?

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
```diff
@@ -546,10 +546,10 @@ public string GetDisplayExpression(string expressionText, ReadOnlySymbolTable sy
             return ExpressionLocalizationHelper.ConvertExpression(expressionText, ruleScope, GetDefaultBindingConfig(), CreateResolverInternal(symbolTable), CreateBinderGlue(), culture, Config.Features, toDisplay: true);
         }
 
-        internal void AddUserDefinedFunction(string script, CultureInfo parseCulture = null, ReadOnlySymbolTable symbolTable = null, bool allowSideEffects = false)
+        public DefinitionsCheckResult AddUserDefinedFunction(string script, CultureInfo parseCulture = null, ReadOnlySymbolTable symbolTable = null, bool allowSideEffects = false)
```
MikeStall, line 549: What happens if the function is already defined?
(either as a builtin or a previous definition)

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -208,63 +208,39 @@ public void AddConstant(string name, FormulaValue data)
...
                 AllowsSideEffects = allowSideEffects, 
-                Culture = parseCulture ?? CultureInfo.InvariantCulture 
+                Culture = parseCulture ?? CultureInfo.InvariantCulture,
             };
-            var sb = new StringBuilder();
 
-            var parseResult = UserDefinitions.Parse(script, options);
-
-            // Compose will handle null symbols
             var composedSymbols = Compose(this, symbolTable, extraSymbolTable);
```
MikeStall, line 228: Can we pass in a single table here and let the caller do Compose if needed?

src/libraries/Microsoft.PowerFx.Interpreter/RecalcEngine.cs
```diff
@@ -460,15 +460,18 @@ private void AddUserDefinedFunctions(IEnumerable<UDF> parsedUdfs, ReadOnlySymbol
 
             foreach (var udf in udfs)
             {
-                Config.SymbolTable.AddFunction(udf);
                 var binding = udf.BindBody(nameResolver, new Glue2DocumentBinderGlue(), BindingConfig.Default, Config.Features);
 
                 List<TexlError> bindErrors = new List<TexlError>();
```
MikeStall, line 465: >bindErrors [](http://example.com/codeflow?start=32&length=10)

are these just binding errors? Or any errors (parse, etc). 

(I think this fixes a bug we had previously)

adithyaselv: just binding errors

src/libraries/Microsoft.PowerFx.Core/Public/Engine.cs
adithyaselv: would fail as a previous definition, and will shadow builtins except for few functions.

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #2561, opened 2024-07-25
Updating fx2nl parameters with range

NavneetThekkumpat:
Updating fx2nl parameters with optional range. This will be utilized by the explain this v2 feature, where range of user selection will be passed along with the full expression.


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Public/NLHandler.cs
```diff
@@ -1,44 +1,44 @@
...
+        public virtual bool SupportsNL2Fx { get; } = false;
+
+        public virtual bool SupportsFx2NL { get; } = false;
+
+        public virtual Task<CustomNL2FxResult> NL2FxAsync(NL2FxParameters request, CancellationToken cancel)
+        {
+            throw new NotImplementedException();
+        }
+
+        [Obsolete("Call overload with Fx2NLParameters")]
```
akshar2401, line 31: hard to understand changes in this file. please clean up

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Fx2Nl/Fx2NlLanguageServerOperationHandler.cs
```diff
@@ -52,17 +52,17 @@ private Task PreHandleFx2NlAsync(LanguageServerOperationContext operationContext
...
             {
                 _preHandleCheckResult = scope?.Check(_fx2NlRequestParams.Expression) ?? throw new NullReferenceException("Check result was not found for Fx2NL operation");
-                _fx2NlParameters = GetFx2NlHints(scope, operationContext);
+                _fx2NlParameters = GetFx2NlParameters(scope, operationContext);
             }, cancellationToken);
         }
 
         /// <summary>
-        /// Gets the Fx2Nl hints.
+        /// Gets the Fx2Nl hints and range.
```
akshar2401, line 60: range is part of hints/parameters. no need to mention it separately

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Public/EditorContextScope.cs
```diff
@@ -59,7 +59,11 @@ public void AddPostCheckExpressionRewriter(IExpressionRewriter expressionRewrite
 
         // Host can set optional hints about where this expression is used. 
         // This can feed into Fx2NL and other help services. 
-        public UsageHints UsageHints { get; init; }
+        public UsageHints UsageHints { get; init; }
+
+        // Host can set optional range of expression.
+        // This can feed into Fx2NL and other help services. 
+        public ScalarRange Range { get; init; }
```
MikeStall, line 66: can we avoid this? Nothing else in this class is specific to Fx2nl

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Public/NLHandler.cs
NavneetThekkumpat: The only change is adding
ScalarRange Range  to Fx2NLParameters. Not sure why it does this weirdness here in this file.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Public/EditorContextScope.cs
NavneetThekkumpat: Sure I can remove this from here if not needed

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Fx2Nl/Fx2NlLanguageServerOperationHandler.cs
```diff
@@ -52,17 +52,17 @@ private Task PreHandleFx2NlAsync(LanguageServerOperationContext operationContext
...
         /// </summary>
         /// <param name="scope"> PowerFx Scope.</param>
         /// <param name="operationContext">Language Server Operation Context.</param>
-        /// <returns>Fx2Nl hints.</returns>
-        private Fx2NLParameters GetFx2NlHints(IPowerFxScope scope,  LanguageServerOperationContext operationContext)
+        /// <returns>Fx2Nl parameters including hints and range.</returns>
+        private Fx2NLParameters GetFx2NlParameters(IPowerFxScope scope,  LanguageServerOperationContext operationContext)
```
akshar2401, line 65: wait why are you not automatically transferring the range to fx2nlparameters from customfx2nl params? that should be done here so not all hosts have to remember transferring range from custom fx2nl params to fx2nlparameters

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Fx2Nl/Fx2NlLanguageServerOperationHandler.cs
NavneetThekkumpat: Ah but since we separating usage hints and range separately made sense to call this GetFx2NlParameters instead.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Public/EditorContextScope.cs
akshar2401: yes you don't even need this range field here. The UsageHints in fx2nlparameters was supposed to be provided from host's server to lsp but the range comes directly host's client and not the server which means it is already a part of customfx2nl params which you can just copy over to fx2nlparameters to transfer to nl library. Sounds awkward but that's more convenient for hosts

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Fx2Nl/Fx2NlLanguageServerOperationHandler.cs
NavneetThekkumpat: Ah okay let me take a look at it.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Public/NLHandler.cs
NavneetThekkumpat, line 1: image
Only change

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Fx2Nl/Fx2NlLanguageServerOperationHandler.cs
```diff
@@ -57,14 +57,27 @@ private Task PreHandleFx2NlAsync(LanguageServerOperationContext operationContext
...
         /// </summary>
         /// <param name="scope"> PowerFx Scope.</param>
         /// <param name="operationContext">Language Server Operation Context.</param>
         /// <returns>Fx2Nl hints.</returns>
         private Fx2NLParameters GetFx2NlHints(IPowerFxScope scope,  LanguageServerOperationContext operationContext)
         {
-            return scope is IPowerFxScopeFx2NL fx2NLScope ? fx2NLScope.GetFx2NLParameters() : new Fx2NLParameters();
+            if (scope is IPowerFxScopeFx2NL fx2NLScope)
+            {
+                var parameters = fx2NLScope.GetFx2NLParameters();
```
akshar2401, line 69: you do not need to create the requirement that range should only be put in the fx2nl parameters if scope is of this type or that.

this can be simplified to

var parameters = scope is IPowerFxScopeFx2NL fx2NLScope ? fx2NlScope.GetFx2nlParameters() :  new Fx2NlParameters()

parameters.Range ??= _fx2nlRequestParams.Range

this covers an edge case you are missing here. not all hosts are expected to implement IPowerFxScopeFx2Nl but they still might want range/partial fx2nl and your logic doesn't account for that

NavneetThekkumpat: Ah gotcha, did not know about that edge case and thanks for the explanation. Makes sense. Let me address it soon.

akshar2401 (approved):

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

Issue #2560, opened 2024-07-24
NL sanitization name provider


anderson-joyle:
Provide NL with a way to sanitize known control and property names.

PR #2562, opened 2024-07-26
NL structural print

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2560.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
@@ -670,7 +670,7 @@ namespace Microsoft.PowerFx
             return _expressionAnonymous;
         }
 
-        public string ApplyGetLogging(TexlFunctionalVisitor<LazyList<string>, ISanitizedNameProvider> visitor, ISanitizedNameProvider nameProvider = null)
+        public string ApplyGetLogging(TexlFunctionalVisitor<IEnumerable<string>, ISanitizedNameProvider> visitor, ISanitizedNameProvider nameProvider = null)
         {
             if (_expressionAnonymous == null)
             {
--- a/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
@@ -670,7 +670,7 @@ namespace Microsoft.PowerFx
             return _expressionAnonymous;
         }
 
-        public string ApplyGetLogging(TexlFunctionalVisitor<IEnumerable<string>, ISanitizedNameProvider> visitor, ISanitizedNameProvider nameProvider = null)
+        internal string ApplyGetLogging(TexlFunctionalVisitor<LazyList<string>, ISanitizedNameProvider> visitor, ISanitizedNameProvider nameProvider = null)
         {
             if (_expressionAnonymous == null)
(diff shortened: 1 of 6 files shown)


src/libraries/Microsoft.PowerFx.Core/UtilityDataStructures/LazyList.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.PowerFx.Core.UtilityDataStructures
     /// the operation without the creation of many intermediate large
     /// memory blocks.
     /// </summary>
-    internal class LazyList<T> : IEnumerable<T>
+    public class LazyList<T> : IEnumerable<T>
```
MikeStall, line 17: >public [](http://example.com/codeflow?start=4&length=6)

I'm skeptical of the value of this class... wondering if we can avoid making it public.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -672,6 +670,18 @@ public string ApplyGetLogging()
             return _expressionAnonymous;
         }
 
+        internal string ApplyGetLogging(ISanitizedNameProvider nameProvider = null)
+        {
+            if (_expressionAnonymous == null)
```
MikeStall, line 675: >_expressionAnonymous [](http://example.com/codeflow?start=16&length=20)

We shouldn't reuse _expressionAnonymous cache here. 

Else calling ApplyGetLogging(sanitizer) will change the result of ApplyGetLogging(). 

or even calling: 
ApplyGetLogging(santizer1) 
ApplyGetLogging(santizer2)  // will fail and return sanitizer1 result. 

Probably fine to just not cache here.

MikeStall: Can you add a test that illustrates this problem? That will help verify the fix.  :)

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -672,6 +670,12 @@ public string ApplyGetLogging()
             return _expressionAnonymous;
         }
 
+        internal string ApplyGetLogging(ISanitizedNameProvider nameProvider = null)
```
MikeStall, line 673: Which overload does just calling `ApplyGetLogging()` bind against?  Line 661, or line 673 with nameProvider=null?

MikeStall: Having 2 separate methods is import from a breaking change perspective, but the trip up is the default value for the param.  Basically:
- should we remove the default value for the param?
- should the existing call in ApplyGetLogging() just become a call to ApplyGetLogging(null).

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #2563, opened 2024-07-29
Update to use Expression Locale string instead of CultureInfo

nguhoa:
Update to use Expression Locale string instead of CultureInfo to avoid json serialization failure in https://msazure.visualstudio.com/OneAgile/_workitems/edit/28863116

### DIFF
--- a/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Nl2Fx/Nl2FxLanguageServerOperationHandler.cs
+++ b/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Nl2Fx/Nl2FxLanguageServerOperationHandler.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using System;
+using System.Globalization;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.PowerFx.Intellisense;
@@ -54,7 +55,7 @@ namespace Microsoft.PowerFx.LanguageServerProtocol.Handlers
                 Sentence = _nl2FxRequestParams.Sentence,
                 SymbolSummary = summary,
                 Engine = check.Engine,
-                ExpressionLocale = check.ParserCultureInfo
+                ExpressionLocaleName = check.ParserCultureInfo?.Name
             };
          }
 
--- a/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Public/NLHandler.cs
+++ b/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Public/NLHandler.cs
@@ -68,10 +68,10 @@ namespace Microsoft.PowerFx.LanguageServerProtocol
         public Engine Engine { get; set; }
 
         /// <summary>
-        /// Set the locale that expect the output expression to be.
+        /// Set the locale name that expect the output expression to be.
         /// For example: "vi-VN" or "fr-FR".
         /// </summary>
-        public CultureInfo ExpressionLocale { get; set; }
+        public string ExpressionLocaleName { get; set; }
     }
 
     /// <summary>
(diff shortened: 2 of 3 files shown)


NavneetThekkumpat (approved):

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Nl2Fx/Nl2FxLanguageServerOperationHandler.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using System;
+using System.Globalization;
```
anderson-joyle, line 5: nit: This can be removed.

anderson-joyle (approved):
:shipit:

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Public/NLHandler.cs
```diff
@@ -68,10 +68,10 @@ public class NL2FxParameters
         public Engine Engine { get; set; }
 
         /// <summary>
-        /// Set the locale that expect the output expression to be.
+        /// Set the locale name that expect the output expression to be.
         /// For example: "vi-VN" or "fr-FR".
         /// </summary>
-        public CultureInfo ExpressionLocale { get; set; }
+        public string ExpressionLocaleName { get; set; }
```
akshar2401, line 74: this change isn't needed and not something we should anyways. This object Nl2FxParameters is never serialized. This object is converted to Nl2FxPayload and that is what we serialize. Can't we just replace the CultureInfo in Nl2FxPayload with the name and keep this untouched and avoiding compiler breaking change (type changes)? on the nl library side, just read the name from Nl2FxParameters.ExpressionLocale?.name and set it to the Nl2FxPayload.ExpressionLocaleName

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Public/NLHandler.cs
```diff
@@ -68,10 +68,10 @@ public class NL2FxParameters
         public Engine Engine { get; set; }
 
         /// <summary>
-        /// Set the locale that expect the output expression to be.
+        /// Set the locale name that expect the output expression to be.
         /// For example: "vi-VN" or "fr-FR".
         /// </summary>
-        public CultureInfo ExpressionLocale { get; set; }
```
MikeStall, line 74: >CultureInfo [](http://example.com/codeflow?start=15&length=11)

Why is CultureInfo bad, but Engine on line 68 is ok?

MikeStall:
Is there a test we can add that would catch this?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Public/NLHandler.cs
akshar2401: We never serialize the Nl2FxParmaters. This object is transferred to Nl2FxPayload so CultureInfo is transferred as it is. Nl2FxPayload's CultureInfo is serialized in the end. This is just the DTO. There's no need to change this. To answer your question, we never serialize Engine

nguhoa: ok so i only need to change in internal repo. I thought i need to change in both here and internal.

akshar2401: no you don't need this pr at all. also in general we should mostly avoid serializing built in types other than mostly collections, and primitive types or the ones which has explicit support for serialization. please add a comprehensive set of tests in Nl library testing the serialization of Nl2FxPayload to prevent this

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.LanguageServerProtocol.NL2FxParameters.get_ExpressionLocale<br />❌CP0002 M:Microsoft.PowerFx.LanguageServerProtocol.NL2FxParameters.set_ExpressionLocale(System.Globalization.CultureInfo)

PR #2566, opened 2024-07-30
Allow UDF to return record without special syntax

adithyaselv:
***Current State***
When Imperative UDF feature is enabled, UDF returning a record value needs to wrap it in parenthesis (eg: `F(): Point = ({x: 5, y:5})`) to parse it without errors.

***Fix***
* This PR adds logic to look ahead and decide between imperative definition and record value.
* UDFs with a body containing just `{}` is considered empty record and not imperative definition.


LucGenetier:
✅ No public API change.

lesaltzm (approved):
The code is right. I'm personally not a huge fan of adding this, but it's not actually harmful.

PR #2569, opened 2024-07-31
Introduce IsType, AsType and ParseJSON overloads with custom types

adithyaselv:
Support IsType, AsType and ParseJSON functions with custom type as input parameter

```
>> IsType(ParseJSON("42"), Number)
true

>> AsType(ParseJSON("42"), Number)
42

>> ParseJSON("42", Number)
42

>> AsType(ParseJSON("{ ""a"": 42 }"), Type({a: Number}))
{a : 42}

```


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4367,6 +4423,25 @@ public override bool PreVisit(CallNode node)
                         startArg++;
                     }
 
+                    if (overloadsWithTypeArgs.Any() && node.Args.Count > 1)
+                    {
+                        var nodeInp = node.Args.Children[0];
```
CarlosFigueiraMSFT, line 4428: Suggestion: `nodeInput` - no need to abbreviate here.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -5024,6 +5099,91 @@ private void PreVisitMetadataArg(CallNode node, TexlFunction func)
...
+                    ArityError(func.MinArity, func.MaxArity, node, argCount, _txb.ErrorContainer);
+                    _txb.SetInfo(node, new CallInfo(func, node));
+                    _txb.SetType(node, DType.Error);
+                    return;
+                }
+
+                Contracts.Assert(argCount > 1);
+                Contracts.AssertValue(args[1]);
+
+                if (args[1] is not FirstNameNode && args[1] is not TypeLiteralNode)
```
CarlosFigueiraMSFT, line 5128: Suggestion: move this after the ìf (args[1] is FirstNameNode typeName)` and `if (args[1] is TypeLiteralNode typeLiteral)` blocks - you can add it as a simple `else` block without any conditions.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -388,6 +388,8 @@ protected set
 
         public bool IsDeprecatedOrInternalFunction => this is IHasUnsupportedFunctions sdf && (sdf.IsDeprecated || sdf.IsInternal);
 
+        public virtual bool HasTypeArgs => false;
```
CarlosFigueiraMSFT, line 391: Can you add a comment with an invariant saying that a function HasTypeArgs == true if and only if there is an argIndex such that func.HasIsType(argIndex) == true (and add a test to validate that)?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AsType.cs
```diff
@@ -141,6 +141,53 @@ protected override bool IsValidAsyncOrImpureNode(TexlNode node, TexlBinding bind
...
+            Contracts.AssertValue(args);
+            Contracts.AssertAllValues(args);
+            Contracts.AssertValue(argTypes);
+            Contracts.Assert(args.Length == 2);
+            Contracts.Assert(argTypes.Length == 2);
+            Contracts.AssertValue(errors);
+
+            if (!base.CheckTypes(context, args, argTypes, errors, out returnType, out nodeToCoercedTypeMap))
+            {
+                return false;
```
CarlosFigueiraMSFT, line 185: Curious: would something like `AsType(ParseJSON("123"), 1)` succeed (and return the number 123)? I don't know if CheckTypes would handle a non-type argument properly. Whatever the answer, can we have a test for that? :-)

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/IsType.cs
```diff
@@ -0,0 +1,71 @@
...
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.PowerFx.Core.App.ErrorContainers;
+using Microsoft.PowerFx.Core.Binding;
+using Microsoft.PowerFx.Core.Entities;
+using Microsoft.PowerFx.Core.Errors;
+using Microsoft.PowerFx.Core.Functions;
+using Microsoft.PowerFx.Core.Functions.Delegation;
```
CarlosFigueiraMSFT, line 11: I'm pretty sure not all of those using statements are needed, can you clean them up?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/UntypedObject.cs
```diff
@@ -45,4 +48,49 @@ public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
             yield return new[] { TexlStrings.IndexArg1, TexlStrings.IndexArg2 };
         }
     }
+
+    // ParseJSON(JsonString:P, Type:T): ![]
+    internal class TypedParseJSONFunction : BuiltinFunction
+    {
+        public const string ParseJsonInvariantFunctionName = "ParseJSON";
```
CarlosFigueiraMSFT, line 55: Instead of repeating it here, can we reuse the name from ParseJSON?

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -17,6 +17,8 @@ public class FormulaValueJsonSerializerSettings
         public bool NumberIsFloat { get; init; } = false;
 
         public bool ReturnUnknownRecordFieldsAsUntypedObjects { get; init; } = false;
+
+        public bool AllowUnknownRecordFields { get; init; } = true;
```
CarlosFigueiraMSFT, line 21: What is this? Can you add a comment to explain what this property means?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/AsType.cs
adithyaselv: it would fail.  
because of this - https://github.com/microsoft/Power-Fx/pull/2569/files#diff-3b97f603fa1daea008a12cd0138ee5a6bf8270bc878af497c2ff4a272d04a583R5128

somewhat equivalent test is this https://github.com/microsoft/Power-Fx/pull/2569/files#diff-a2864451c36106fe453a61cef0229cc6b6dd73c0eb77ad88485dd954fbf5a804R123 .

will add the suggested test too

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
adithyaselv: runs in redeclaration issue issue as nodeInput is declared later in this method. thats why even the if block above has nodeInp

src/libraries/Microsoft.PowerFx.Json/Functions/IsTypeUOFunctionImpl.cs
```diff
@@ -0,0 +1,68 @@
...
+        public async Task<FormulaValue> InvokeAsync(FormulaValue[] args, CancellationToken cancellationToken)
+        {
+            var irContext = IRContext.NotInSource(FormulaType.UntypedObject);
+            var arg0 = args[0];
+
+            if (arg0 is BlankValue || arg0 is ErrorValue)
+            {
+                return arg0;
+            }
+            else if (arg0 is not UntypedObjectValue)
```
CarlosFigueiraMSFT, line 31: Can this ever happen? Wouldn't it be blocked by the compiler?

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4694,4 +4742,8 @@
     <value>The specified data source cannot be used with this function.</value>
     <comment>Error Message.</comment>
   </data>
+  <data name="ErrInvalidArgumentExpectedType" xml:space="preserve">
+    <value>Invalid argument '{0}'. Expected valid type name or typeliteral.</value>
```
CarlosFigueiraMSFT, line 4746: `typeliteral` --> `type literal` (space)

CarlosFigueiraMSFT:
Can we also add some .txt-based tests for these new functions? It should just work with untyped objects.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4286,6 +4312,33 @@ private void UntypedObjectScopeError(CallNode node, TexlFunction maybeFunc, Texl
 
                 _txb.SetInfo(node, new CallInfo(maybeFunc, node, null, default, false, _currentScope.Nest));
                 _txb.SetType(node, maybeFunc.ReturnType);
+            }
+
+            private bool IsFunctionWithTypeArg(CallNode node, TexlFunction maybeFunc)
```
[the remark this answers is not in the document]
adithyaselv: do you think this the current name is fine ?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
adithyaselv: done and added test

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

anderson-joyle:
Once this is merged can you also please update https://github.com/microsoft/Power-Fx/blob/main/releasenotes/releasenotes-1.3.0-rc.md?

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

Shpakh-MSFT (approved):
:shipit:

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/AsType_UO.txt
```diff
@@ -0,0 +1,94 @@
+#SETUP: AllowTypeLiteral
+
+// Primitives
+>> AsType(ParseJSON("987654321"), Number)
+987654321
```
CarlosFigueiraMSFT, line 5: It will be good to add some tests with decimal literals as well

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/AsType_UO.txt
```diff
@@ -0,0 +1,94 @@
...
+true
+
+>> AsType(ParseJSON("""AstypeFunction"""), Type(Text))
+"AstypeFunction"
+
+>> AsType(ParseJSON("""1984-01-01"""), Date)
+Date(1984,1,1)
+
+>> AsType(ParseJSON("""1900-12-31T23:59:59.999Z"""), DateTime)
+DateTime(1900,12,31,23,59,59,999)
```
CarlosFigueiraMSFT, line 17: Can we flip the types on those two? For example:

    >> AsType(ParseJSON("""1984-01-01"""), DateTime)
    Date(1984,1,1,0,0,0)

    >> AsType(ParseJSON("""1900-12-31T23:59:59.999Z"""), Date)
    DateTime(1900,12,31)

CarlosFigueiraMSFT, line 17: This doesn't look right. DateTime(1900,12,31,23,59,59,999) is 1900-12-31T23:59:59.999 _in the local time zone_, while the value that was parsed was from UTC

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/AsType_UO.txt
```diff
@@ -0,0 +1,94 @@
...
+// Table
+>> AsType(ParseJSON("[{""a"": ""Hello"", ""b"": ""2012-01-02""}, {""a"": ""Hi"", ""b"": ""2012-01-03""}]"), Type([{a: Text, b: Date}]))
+Table({a:"Hello",b:Date(2012,1,2)},{a:"Hi",b:Date(2012,1,3)})
+
+>> AsType(ParseJSON("[{""a"": [{""z"": true}, {""z"": false}]}, {""a"": [{""z"": false}, {""z"": true}]}]"), Type([{a: [{z: Boolean}]}]))
+Table({a:Table({z:true},{z:false})},{a:Table({z:false},{z:true})})
+
+// Negative tests
+>> AsType(ParseJSON("5"), Text)
+Error({Kind:ErrorKind.InvalidJSON})
```
CarlosFigueiraMSFT, line 49: >{Kind:ErrorKind.InvalidJSON [](http://example.com/codeflow?start=6&length=27)

This is not an invalid JSON - the JSON is valid. This should probably be InvalidArgument (or some other new error kind). Same with the next errors.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/AsType_UO.txt
```diff
@@ -0,0 +1,94 @@
...
+>> AsType(ParseJSON("1"), 1)
+Errors: Error 23-24: Invalid argument '1'. Expected valid type name or type literal.
+
+>> AsType(ParseJSON("5"), Type(5))
+Errors: Error 28-29: Type literal declaration is invalid. The expression '5' cannot be used in a type definition.|Error 27-28: Type literal declaration is invalid. The expression 'Type(5)' cannot be used in a type definition.
+
+>> AsType(ParseJSON("true"), UnKnown)
+Errors: Error 26-33: Name isn't valid. 'UnKnown' isn't recognized.|Error 0-34: Invalid argument 'UnKnown'. Expected valid type name or type literal.
+
+>> AsType(ParseJSON("fasle"), Boolean)
```
CarlosFigueiraMSFT, line 66: >fasle [](http://example.com/codeflow?start=21&length=5)

This, unlike the test on line 48, is invalid JSON, so the error kind should be InvalidJSON instead.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/AsType_UO.txt
```diff
@@ -0,0 +1,94 @@
...
+Error({Kind:ErrorKind.InvalidArgument})
+
+>> AsType(ParseJSON("true"), Void)
+Errors: Error 26-30: Invalid argument type (Void). Expecting a Error value instead.|Error 0-6: The function 'AsType' has some invalid arguments.
```
CarlosFigueiraMSFT, line 70: >Expecting a Error value instead [](http://example.com/codeflow?start=51&length=31)

This message is strange. We should not expect an error value.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/AsType_UO.txt
```diff
@@ -0,0 +1,94 @@
...
+
+>> AsType(ParseJSON("{""a"": 5, ""b"":true}"), Type({a: Number}))
+Error({Kind:ErrorKind.InvalidJSON})
+
+>> AsType(ParseJSON("{""a"": 5}"), Type({a: Number}), "Hello")
+Errors: Error 0-59: Invalid number of arguments: received 3, expected 2.
+
+>> AsType(ParseJSON("true"), None)
```
CarlosFigueiraMSFT, line 78: >AsType(ParseJSON("true"), None) [](http://example.com/codeflow?start=3&length=31)

Would `AsType(ParseJSON("null"), None)` work? If so, can we add a case for that?

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/AsType_UO.txt
```diff
@@ -0,0 +1,94 @@
...
+Errors: Error 26-30: Unsupported untyped/JSON conversion type 'ObjNull' in argument.
+
+>> AsType(ParseJSON("{}"), Type({a: Text, b: [Color]}))
+Errors: Error 28-29: Unsupported untyped/JSON conversion type 'Color' in argument.
```
CarlosFigueiraMSFT, line 82: >Unsupported untyped/JSON conversion type 'Color' in argument. [](http://example.com/codeflow?start=21&length=61)

Is this a problem with the Color type (i.e., if the second argument were `Type({a:Text,b:Number})`, would it work)? If so, then we should have a negative test with Color as a primitive too. If not, then why wouldn't this work? From line 33, we can have missing fields in the record.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/IsType_UO.txt
```diff
@@ -0,0 +1,101 @@
+#SETUP: AllowTypeLiteral
+
+// Primitives
```
CarlosFigueiraMSFT, line 3: >Primitives [](http://example.com/codeflow?start=3&length=10)

Many of the comments for AsType_UO.txt also apply here

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/TypedParseJSON.txt
```diff
@@ -0,0 +1,97 @@
+#SETUP: AllowTypeLiteral
+
+// Primitives
+>> ParseJSON("5", Number)
+5
+
+>> ParseJSON("98765.4321", Number)
+98765.4321
+
+>> ParseJSON("""HelloWorld""", Type(Text))
```
CarlosFigueiraMSFT, line 10: >Type(Text) [](http://example.com/codeflow?start=31&length=10)

Why do we need to say `Type(Text)` and not just `Text`? Would both work?

src/tests/Microsoft.PowerFx.Json.Tests.Shared/AsTypeIsTypeParseJSONTests.cs
```diff
@@ -0,0 +1,200 @@
...
+            else
+            {
+                var result = engine.Check(expression, options: ParseType);
+                Assert.False(result.IsSuccess);
+                Assert.Contains(result.Errors, e => e.MessageKey == expectedError);
+            }
+        }
+
+        [Fact]
+        public void TestFunctionsWithTypeArgs()
```
CarlosFigueiraMSFT, line 152: >TestFunctionsWithTypeArgs [](http://example.com/codeflow?start=20&length=25)

It would be also good to assert that for functions that do **not** have type args, then none of ArgIsType returns true. Something like

    var functionsWithoutTypeArgs = TestOnly_AllBuiltinFunctions.Where(f => !f.HasTypeArgs);
    Assert.All(functionsWithoutTypeArgs, f => !Enumerable.Range(0, Math.Min(f.MaxArity, 5)).Any(index => f.ArgIsType(index)))

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/AsType_UO.txt
adithyaselv: it will not work. None is not a allowed as valid type arg to be used here

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/TypedParseJSON.txt
adithyaselv: both would work , will add both

src/libraries/Microsoft.PowerFx.Json/Functions/TypedParseJSONImpl.cs
```diff
@@ -0,0 +1,53 @@
...
+            }
+
+            var typeString = (StringValue)args[1];
+            var json = ((StringValue)arg0).Value;
+
+            if (!DType.TryParse(typeString.Value, out DType dtype))
+            {
+                return new ErrorValue(irContext, new ExpressionError()
+                {
+                    Message = $"Internal error: Unable to parse type argument",
```
CarlosFigueiraMSFT, line 42: > = $"Internal error: Unable to parse type argument", [](http://example.com/codeflow?start=27&length=52)

Can you add a comment saying that this should never happen, as the typeString will be created by the IR (if that's the case) by serializing the type argument. Same comment for the other two new functions.

src/libraries/Microsoft.PowerFx.Json/Functions/IsTypeFunction_UOImpl.cs
```diff
@@ -0,0 +1,56 @@
...
+using Microsoft.PowerFx.Functions;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Core.Texl.Builtins
+{
+    internal class IsTypeFunction_UOImpl : IsTypeFunction_UO, IAsyncTexlFunction
+    {
+        public async Task<FormulaValue> InvokeAsync(FormulaValue[] args, CancellationToken cancellationToken)
+        {
+            var irContext = IRContext.NotInSource(FormulaType.UntypedObject);
```
CarlosFigueiraMSFT, line 19: >var [](http://example.com/codeflow?start=12&length=3)

There is a lot of repetition from this function and AsTypeFunction_UOImpl; can we refactor some of the code? For example, lines 19-43 are identical

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/UntypedObject.cs
```diff
@@ -45,4 +51,94 @@ public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
             yield return new[] { TexlStrings.IndexArg1, TexlStrings.IndexArg2 };
         }
     }
+
+    internal abstract class UntypedOrJSONConversionFunction : BuiltinFunction
```
CarlosFigueiraMSFT, line 55: >UntypedOrJSONConversionFunction [](http://example.com/codeflow?start=28&length=31)

Since this is going to be used by functions outside this file (AsType / IsType), can you move it to a separate file?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/UntypedObject.cs
```diff
@@ -45,4 +51,94 @@ public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
...
+        public override bool SupportsParamCoercion => false;
+
+        public override bool HasTypeArgs => true;
+
+        public override bool ArgIsType(int argIndex)
+        {
+            return argIndex == 1;
+        }
+
+        private static readonly ISet<DType> UnsupportedJSONTypes = new HashSet<DType> { DType.Color, DType.Void, DType.Time, DType.ObjNull };
```
CarlosFigueiraMSFT, line 68: >DType.Time [](http://example.com/codeflow?start=113&length=10)

Why is Time not supported? Something like `TimeValue(ParseJSON("""13:35:57"""))` works fine today

CarlosFigueiraMSFT, line 68: >UnsupportedJSONTypes [](http://example.com/codeflow?start=44&length=20)

Instead of having a list of unsupported types, it may be better to have a list of supported ones, otherwise we may add a type in the future, not add the proper support, and it this function will work incorrectly.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/UntypedObject.cs
```diff
@@ -45,4 +51,94 @@ public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
...
+                return;
+            }
+
+            if (UnsupportedJSONTypes.Contains(type))
+            {
+                errors.EnsureError(DocumentErrorSeverity.Severe, typeArg, TexlStrings.ErrUnsupportedTypeInTypeArgument, type.Kind);
+                return;
+            }
+
+            return;
```
CarlosFigueiraMSFT, line 127: Not needed

CarlosFigueiraMSFT:
>                         _flagsMode.Push(parserOptions.AllowsSideEffects ? Flags.EnableExpressionChaining : Flags.None);

Not part of your changes, but will this line remove the AllowTypeLiteral from the flags? Would something like `Fn() = { Set(a,1); Set(b, AsType(ParseJSON("2"), Number)) };` work?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs:392 in cb3397a. [](commit_id = cb3397ab50c478198c9fed0af38d3a1c9e619288, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
```diff
@@ -861,7 +861,7 @@ private async Task<FormulaValue> PostProcessResultAsync(FormulaValue result, Bas
         {
             ExpressionError er = null;
 
-            if (result is ErrorValue ev && (er = ev.Errors.FirstOrDefault(e => e.Kind == ErrorKind.Network || e.Kind == ErrorKind.InvalidJSON)) != null)
+            if (result is ErrorValue ev && (er = ev.Errors.FirstOrDefault(e => e.Kind == ErrorKind.Network || e.Kind == ErrorKind.InvalidJSON || e.Kind == ErrorKind.InvalidArgument)) != null)
```
CarlosFigueiraMSFT, line 864: >|| e.Kind == ErrorKind.InvalidArgument [](http://example.com/codeflow?start=142&length=38)

Can you check with Luc, who wrote this code, if this is ok?

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -126,18 +134,29 @@ internal static FormulaValue FromJson(JsonElement element, FormulaValueJsonSeria
...
                     }
                     else if (formulaType is DateTimeNoTimeZoneType)
                     {
                         DateTime dt3 = element.GetDateTime();
                         return DateTimeValue.New(TimeZoneInfo.ConvertTimeToUtc(dt3));
                     }
+                    else if (formulaType is TimeType)
+                    {
+                        var timeString = element.GetString();
+                        if (TimeSpan.TryParseExact(timeString, @"hh\:mm\:ss\.FFFFFFF", CultureInfo.InvariantCulture, TimeSpanStyles.None, out var res) ||
```
CarlosFigueiraMSFT, line 152: >FFFFFFF [](http://example.com/codeflow?start=77&length=7)

Does it need to be 0 or 7 decimal places? If we have something like 12:34:56.789, should it work?

src/libraries/Microsoft.PowerFx.Json/Functions/JSONFunctionUtils.cs
```diff
@@ -0,0 +1,76 @@
...
+using Microsoft.PowerFx.Functions;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Core.Utils
+{
+    internal class JSONFunctionUtils
+    {
+        private static readonly FormulaValueJsonSerializerSettings DefaultFormulaValueSerializerSettings = new FormulaValueJsonSerializerSettings { AllowUnknownRecordFields = false, ResultTimeZone = TimeZoneInfo.Local };
+
+        public static FormulaValue ConvertUnTypedObjectToFormulaValue(IRContext irContext, FormulaValue input, StringValue typeString)
```
CarlosFigueiraMSFT, line 18: >ConvertUnTyped [](http://example.com/codeflow?start=35&length=14)

[minor] Untyped (not UnTyped)

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
adithyaselv: using the same piece from TimeValue Function for it to be consistent

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/AsType_UO.txt
```diff
@@ -0,0 +1,151 @@
+#SETUP: AllowTypeLiteral
+
+// Primitives
+>> AsType(ParseJSON("987654321"), Number)
+987654321
+
+>> AsType(ParseJSON("98765.4321"), Decimal)
```
CarlosFigueiraMSFT, line 7: >98765.4321 [](http://example.com/codeflow?start=21&length=10)

For decimal, can you use a value that cannot be represented as a float? For example, `12345678901234.567890123456`

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/AsType_UO.txt
```diff
@@ -0,0 +1,151 @@
...
+>> AsType(ParseJSON("""1900-12-31T00:00:00.000Z"""), Date)
+Date(1900,12,31)
+
+>> AsType(ParseJSON("""11:59:59.999"""), Time)
+Time(11,59,59,999)
+
+>> AsType(ParseJSON("""00:00:00"""), Time)
+Time(0,0,0,0)
+
+>> AsType(ParseJSON("""1900-12-31T00:00:00.000Z"""), DateTimeTZInd)
```
CarlosFigueiraMSFT, line 52: >DateTimeTZInd [](http://example.com/codeflow?start=53&length=13)

What does this type mean? Is it the same as UTC? If so, then the next test should not add 8 hours (Pacific Time adjustment). Sorry, but date/times are hard :-D

CarlosFigueiraMSFT, line 52: >DateTimeTZInd [](http://example.com/codeflow?start=53&length=13)

What does this type mean? Is it the same as UTC? If so, then the next test should not add 8 hours (Pacific Time adjustment). Sorry, but date/times are hard :-D

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/AsType_UO.txt
```diff
@@ -0,0 +1,151 @@
...
+DateTime(1900,12,31,0,0,0,0)
+
+>> AsType(ParseJSON("""1900-12-31T00:00:00.000"""), DateTimeTZInd)
+DateTime(1900,12,31,8,0,0,0)
+
+>> AsType(ParseJSON("""1900-12-31T00:00:00.000-08:00"""), DateTimeTZInd)
+DateTime(1900,12,31,8,0,0,0)
+
+>> DateTimeValue(AsType(ParseJSON("""1900-12-31T00:00:00.000Z"""), UntypedObject))
```
CarlosFigueiraMSFT, line 61: >UntypedObject [](http://example.com/codeflow?start=67&length=13)

Didn't we decide to rename it to just `Untyped`?

adithyaselv: will do it in a different PR ?

CarlosFigueiraMSFT: Makes sense

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
LucGenetier: That should be ok

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
adithyaselv: Luc has confirmed that this should be good.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/AsType_UO.txt
adithyaselv: removed the test as per offline discussion. will work to greg to decide the semantics and send a fix as a follow up PR

Shpakh-MSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2570, opened 2024-07-31
Update order of checking for scoping variable during binder's visit of FirstName

CarlosFigueiraMSFT:
For an expression such as `With({a:1}, Set(a, 10))`, the first argument of the Set function should not refer to the scope variable `a` defined by the With function.


LucGenetier:
✅ No public API change.

anderson-joyle:
What's the expected result of `With({a:1}, Set(a, 10); a + 1)`?

CarlosFigueiraMSFT:
> What's the expected result of `With({a:1}, Set(a, 10); a + 1)`?

It will be 11; the `a` in the Set call will not be bound to the scope from With

LucGenetier:
✅ No public API change.

lesaltzm (approved):

PR #2572, opened 2024-08-02
Make RecordValue.TryGetFieldAsync public

LucGenetier:
This will allow to implement ColumnMapRecordValue in Dataverse repo


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall:
The liability is that we're enabling a new public entry point that circumvents checks. 

We have other patterns that use Composition over RecordValue. How do they accomplish it without this, and can we use that pattern? For example:
https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Core/Public/Values/CompileTimeTypeWrapperTableValue.cs

PR #2578, opened 2024-08-05
Make Nl2Fx/Fx2nl Handler creation Async to allow for setting up custom NlHandler with Nl2Fx context reduction filters

akshar2401:
This pr makes nl handler creation async so during creation time, PA can attach context reduction filters with Nl handler which would then later be used to trim down the context. This pr also refactors nl2fx, fx2nl handlers to avoid class fields

### DIFF
--- a/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Fx2Nl/Fx2NlLanguageServerOperationHandler.cs
+++ b/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Fx2Nl/Fx2NlLanguageServerOperationHandler.cs
@@ -88,8 +88,7 @@ namespace Microsoft.PowerFx.LanguageServerProtocol.Handlers
                 return;
             }
 
-            var nlHandler = await operationContext.GetNLHandler(fx2NlRequestParams.TextDocument.Uri, _nLHandlerFactory, fx2NlRequestParams, cancellationToken).ConfigureAwait(false);
-            _ = nlHandler ?? throw new ArgumentNullException(nameof(nlHandler));
+            var nlHandler = await operationContext.GetNLHandlerAsync(fx2NlRequestParams.TextDocument.Uri, _nLHandlerFactory, fx2NlRequestParams, cancellationToken).ConfigureAwait(false);
             if (!nlHandler.SupportsFx2NL)
             {
                 throw new NotSupportedException("Fx2Nl is not supported");
--- a/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Fx2Nl/Fx2NlLanguageServerOperationHandler.cs
+++ b/src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Fx2Nl/Fx2NlLanguageServerOperationHandler.cs
@@ -21,14 +21,6 @@ namespace Microsoft.PowerFx.LanguageServerProtocol.Handlers
 
         private readonly INLHandlerFactory _nLHandlerFactory;
 
-        private CheckResult _preHandleCheckResult;
-
-        private Fx2NLParameters _fx2NlParameters;
-
-        private CustomFx2NLParams _fx2NlRequestParams;
-
-        private NLHandler _nLHandler;
-
         public Fx2NlLanguageServerOperationHandler(INLHandlerFactory nLHandlerFactory)
         {
             _nLHandlerFactory = nLHandlerFactory;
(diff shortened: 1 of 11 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Fx2Nl/Fx2NlLanguageServerOperationHandler.cs
```diff
@@ -88,18 +83,25 @@ private async Task Fx2NlAsync(LanguageServerOperationContext operationContext, C
...
-            if (!operationContext.TryParseParamsAndAddErrorResponseIfNeeded(out _fx2NlRequestParams))
+            if (!operationContext.TryParseParamsAndAddErrorResponseIfNeeded(out CustomFx2NLParams fx2NlRequestParams))
             {
                 return;
             }
 
-            await PreHandleFx2NlAsync(operationContext, cancellationToken).ConfigureAwait(false);
-            if (_preHandleCheckResult == null)
+            var nlHandler = await operationContext.GetNLHandler(fx2NlRequestParams.TextDocument.Uri, _nLHandlerFactory, fx2NlRequestParams, cancellationToken).ConfigureAwait(false);
+            _ = nlHandler ?? throw new ArgumentNullException(nameof(nlHandler));
```
CarlosFigueiraMSFT, line 92: This exception (***Argument** NullException for the nlHandler argument) is not appropriate (there is not an argument called nlHandler, and the error isn't about an argument being null). Possibly InvalidOperationException would be better here.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/LanguageServerOperationContext.cs
```diff
@@ -77,10 +77,13 @@ private IPowerFxScope GetScope(string uri)
         /// <param name="uri">Uri to use for scope creation.</param>
         /// <param name="factory">Nl Handler Factory.</param>
         /// <param name="nLParams">Ml Params.</param>
+        /// <param name="cancellationToken"> Cancellation Token. </param>
         /// <returns>NlHandler Instance.</returns>
-        internal NLHandler GetNLHandler(string uri, INLHandlerFactory factory, BaseNLParams nLParams = null)
+        internal Task<NLHandler> GetNLHandler(string uri, INLHandlerFactory factory, BaseNLParams nLParams = null, CancellationToken cancellationToken = default)
```
CarlosFigueiraMSFT, line 82: Suggestion: rename to `GetNLHandlerAsync` since this is now an asynchronous method.

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/LanguageServiceProtocol/RedesignedLanguageServerTests/Nl2FxTests.cs
```diff
@@ -231,8 +232,35 @@ private static (string payload, string id) NL2FxMessageJson(string documentUri)
...
+                    await Task.Delay(DelayTime, cancellationToken).ConfigureAwait(false);
+                }
+
+                return base.GetNLHandler(scope, nlParams);
+            }
+
+            public static INLHandlerFactory Create(NLHandler handler)
+            {
+                var random = new Random();
+                var useAsyncFactory = random.Next(0, 2) == 1;
```
CarlosFigueiraMSFT, line 261: Is there a possibility (even small) that a test will pass with one case and not with the other? Having random values may lead to flaky tests. Can we pass the async value as a parameter, and in the tests we can be explicit about which one to use, possibly calling it with both values?

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/LanguageServerOperationContext.cs
MikeStall: +1 
This is actually a very important convention.

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/Fx2Nl/Fx2NlLanguageServerOperationHandler.cs
akshar2401: actually nl handler is not expected to be null. this is over defensiveness at best (to myself :))

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/LanguageServiceProtocol/RedesignedLanguageServerTests/Nl2FxTests.cs
akshar2401: I don't see a chance for flakiness here. while i agree that randomness should be avoided in tests, i tend to respect that as far as introducing randomness in the test cases or test inputs go (generating test cases randomly).  that said, i will remove it anyways

src/libraries/Microsoft.PowerFx.LanguageServerProtocol/Handlers/LanguageServerOperationContext.cs
akshar2401: agreed. just forgot about adding async there

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
🚢

LucGenetier:
✅ No public API change.

PR #2580, opened 2024-08-06
Fixes reuse of expandinfo from cache when expand is accessed via a different path

lesaltzm:
Unable to easily test in this repo. Tests added in PA Client. 

Scenario is, e.g. for this expression, the expand info to the second PrimaryContact access should have ManagingPartner as part of it's expand path. 

"ClearCollect(c3,  Filter(Accounts,'Primary Contact'.Status <> 'Status (Contacts)'.Active));" +
            "ClearCollect(c4,  Filter(Contacts, 'Managing Partner'.'Primary Contact'.Status <> 'Status (Contacts)'.Inactive))"


CarlosFigueiraMSFT (approved):
🚢

LucGenetier:
✅ No public API change.

PR #2584, opened 2024-08-12
CDP updates

LucGenetier:
Make GetDatasetsMetadataAsync a static method
Fix a NullReference exception when processing SQL relationships and no relationship is found


MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2588, opened 2024-08-13
add test to show ColumnNames usage with External.ArrayAndObject untyp…

spanopoulos-msft:
…ed object types


LucGenetier:
✅ No public API change.

PR #2595, opened 2024-08-16
Fix check to exclude UDFs from being marked as async if binding does not exist

rick-nguyen:
There was a mistake in following PR: https://github.com/microsoft/Power-Fx/commit/752efe46c476975a5433d9aab127f0b66808c187


anhtruong-MSFT (approved):

LucGenetier:
✅ No public API change.

PR #2597, opened 2024-08-19
Option to anonymize unknown public functions

anderson-joyle:
Previously at https://github.com/microsoft/Power-Fx/pull/2474 we introduced an API to get a list of called functions from an expression.
We are now adding an option to anonymize a custom function to mask any private information.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
@@ -715,9 +715,9 @@ namespace Microsoft.PowerFx
             return summary;
         }
 
-        public IEnumerable<string> GetFunctionNames()
+        public IEnumerable<string> GetFunctionNames(bool annonymizeUnknownPublicFunctions = false)
         {
-            return ListFunctionVisitor.Run(ApplyParse());
+            return ListFunctionVisitor.Run(ApplyParse(), annonymizeUnknownPublicFunctions);
         }
     }
 
--- a/src/libraries/Microsoft.PowerFx.Core/Syntax/Visitors/ListFunctionVisitor.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Syntax/Visitors/ListFunctionVisitor.cs
@@ -12,7 +12,7 @@ namespace Microsoft.PowerFx.Syntax
         // Use Fullname as key because it's unique. 
         private readonly HashSet<string> _functionNames = new HashSet<string>();
         private readonly Dictionary<string, string> _unknownFunctionNames = new Dictionary<string, string>();
-        private readonly bool _annonymizeUnknownPublicFunctions;
+        private readonly bool _anonymizedUnknownPublicFunctions;
 
         public static IEnumerable<string> Run(ParseResult parse, bool annonymizeUnknownPublicFunctions = false)
         {
@@ -23,29 +23,30 @@ namespace Microsoft.PowerFx.Syntax
 
         private ListFunctionVisitor(bool annonymizeUnknownPublicFunctions)
         {
-            _annonymizeUnknownPublicFunctions = annonymizeUnknownPublicFunctions;
+            _anonymizedUnknownPublicFunctions = annonymizeUnknownPublicFunctions;
         }
 
         public override bool PreVisit(CallNode node)
         {
             var hasNamespace = node.Head.Namespace.Length > 0;
+            var name = node.Head.Name;
 
-            if (_annonymizeUnknownPublicFunctions && !BuiltinFunctionsCore.IsKnownPublicFunction(node.Head.Name))
+            if (_anonymizedUnknownPublicFunctions && !BuiltinFunctionsCore.IsKnownPublicFunction(name))
             {
                 // An expression can have multiple unknown function names.
                 // Track them all to ensure they are uniquely anonymized.
-                if (!_unknownFunctionNames.ContainsKey(node.Head.Name))
+                if (!_unknownFunctionNames.ContainsKey(name))
                 {
-                    _unknownFunctionNames[node.Head.Name] = $"$#CustomFunction{_unknownFunctionNames.Count + 1}#$";
+                    _unknownFunctionNames[name] = $"$#CustomFunction{_unknownFunctionNames.Count + 1}#$";
                 }
 
-                _functionNames.Add(_unknownFunctionNames[node.Head.Name]);
+                _functionNames.Add(_unknownFunctionNames[name]);
             }
             else
             {
                 var fullName = hasNamespace ?
-                        node.Head.Namespace + "." + node.Head.Name :
-                        node.Head.Name;
+                        node.Head.Namespace + "." + name :
+                        name;
 
                 _functionNames.Add(fullName);
             }
--- a/src/libraries/Microsoft.PowerFx.Core/Syntax/Visitors/ListFunctionVisitor.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Syntax/Visitors/ListFunctionVisitor.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using System.Collections.Generic;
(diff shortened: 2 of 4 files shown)


src/libraries/Microsoft.PowerFx.Core/Syntax/Visitors/ListFunctionVisitor.cs
```diff
@@ -2,32 +2,53 @@
...
+        private ListFunctionVisitor(bool annonymizeUnknownPublicFunctions)
+        {
+            _annonymizeUnknownPublicFunctions = annonymizeUnknownPublicFunctions;
         }
 
         public override bool PreVisit(CallNode node)
         {
-            var hasNamespace = node.Head.Namespace.Length > 0;
-
-            var name = node.Head.Name;
```
jas-valgotar, line 25: > var name = node.Head.Name; [](http://example.com/codeflow?start=11&length=27)

Why did we revert this? hoisting that info at the start is more efficient.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -715,9 +715,9 @@ public CheckContextSummary ApplyGetContextSummary()
             return summary;
         }
 
-        public IEnumerable<string> GetFunctionNames()
+        public IEnumerable<string> GetFunctionNames(bool annonymizeUnknownPublicFunctions = false)
```
jas-valgotar, line 718: >(bool annonymizeUnknownPublicFunctions = false) [](http://example.com/codeflow?start=51&length=47)

Can we avoid a binary breaking change and chain it to a new function definition?

src/libraries/Microsoft.PowerFx.Core/Syntax/Visitors/ListFunctionVisitor.cs
```diff
@@ -2,32 +2,53 @@
...
-                    name;
-
-            _functionNames.Add(fullName);
+            var hasNamespace = node.Head.Namespace.Length > 0;
+
+            if (_annonymizeUnknownPublicFunctions && !BuiltinFunctionsCore.IsKnownPublicFunction(node.Head.Name))
+            {
+                // An expression can have multiple unknown function names.
+                // Track them all to ensure they are uniquely anonymized.
+                if (!_unknownFunctionNames.ContainsKey(node.Head.Name))
```
jas-valgotar, line 37: > if (!_unknownFunctionNames.ContainsKey(node.Head.Name)) [](http://example.com/codeflow?start=15&length=56)

In which scenario this will be false?

src/libraries/Microsoft.PowerFx.Core/Syntax/Visitors/ListFunctionVisitor.cs
```diff
@@ -2,32 +2,53 @@
...
-
-            var name = node.Head.Name;
-            var fullName = hasNamespace ?
-                    node.Head.Namespace + "." + name :
-                    name;
-
-            _functionNames.Add(fullName);
+            var hasNamespace = node.Head.Namespace.Length > 0;
+
+            if (_annonymizeUnknownPublicFunctions && !BuiltinFunctionsCore.IsKnownPublicFunction(node.Head.Name))
```
jas-valgotar, line 33: >BuiltinFunctionsCore.IsKnownPublicFunction(node.Head.Name) [](http://example.com/codeflow?start=54&length=58)

Is this check sufficient? Ie will Connector Function will also get anonymized as customFunction?

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
anderson-joyle: I dont think this is necessary in this case.

src/libraries/Microsoft.PowerFx.Core/Syntax/Visitors/ListFunctionVisitor.cs
anderson-joyle: `Foo() + Bar() + Foo()`
Whenever we have the same custom function called multiple times.

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.CheckResult.GetFunctionNames

src/libraries/Microsoft.PowerFx.Core/Syntax/Visitors/ListFunctionVisitor.cs
```diff
@@ -2,32 +2,53 @@
...
 namespace Microsoft.PowerFx.Syntax
 {
     internal class ListFunctionVisitor : IdentityTexlVisitor
     {
         // FullName --> Name 
         // Use Fullname as key because it's unique. 
-        private readonly HashSet<string> _functionNames = new HashSet<string>();
+        private readonly HashSet<string> _functionNames = new HashSet<string>();
+        private readonly Dictionary<string, string> _unknownFunctionNames = new Dictionary<string, string>();
+        private readonly bool _annonymizeUnknownPublicFunctions;
```
LucGenetier, line 15: >annonymize [](http://example.com/codeflow?start=31&length=10)

nit - typo: single 'n'

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ListFunctionVisitorTests.cs
```diff
@@ -12,10 +12,12 @@ public class ListFunctionVisitorTests : PowerFxTest
         [InlineData("Abs(1)", "Abs")]
         [InlineData("Abs(Abs(Abs(Abs(Abs(1)))))", "Abs")]
         [InlineData("With({x:Today()}, x+1)", "With,Today")]
-        [InlineData("SomeNameSpace.Foo() + SomeNameSpace.Bar()", "SomeNameSpace.Foo,SomeNameSpace.Bar")]
+        [InlineData("SomeNameSpace.Foo() + SomeNameSpace.Bar()", "SomeNameSpace.Foo,SomeNameSpace.Bar")]
+        [InlineData("SomeNameSpace.Foo() + SomeNameSpace.Bar()", "$#CustomFunction1#$,$#CustomFunction2#$", true)]
+        [InlineData("SomeNameSpace.Foo() + SomeNameSpace.Bar() + SomeNameSpace.Foo()", "$#CustomFunction1#$,$#CustomFunction2#$", true)]
```
LucGenetier, line 17: >"SomeNameSpace.Foo() + SomeNameSpace.Bar() + SomeNameSpace.Foo() [](http://example.com/codeflow?start=20&length=64)

Is it possible to test w/o a namespace?

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
jas-valgotar: It actually does help with the some deployments(e.g. Dataverse) where they can dynamically load new nugets without re-compiling and this kind of binary breaking changes can cause issues.

src/libraries/Microsoft.PowerFx.Core/Syntax/Visitors/ListFunctionVisitor.cs
jas-valgotar: `StructuralPrint` doesn't have this issue, because looks like it doesn't care about a difference between custom function vs connector function (because all of them are external function).

Here we are specifying that it is a CustomFunction. If instead CustomFunction, we were to say Function or ExternalFunction this would be fine.

LucGenetier (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
MikeStall: We should avoid the breaking change.

MikeStall, line 718: nit: misspelling : anonymize

PR #2601, opened 2024-08-19
ColumnName to work with array untyped object

anderson-joyle:
Reply to https://github.com/microsoft/Power-Fx/pull/2588.


LucGenetier:
✅ No public API change.

MikeStall (approved):

Issue #908, opened 2022-12-14
Ensure all messages are localized in all languages


LucGenetier:
Compliance task 16183202 - Ensure Product is Localizable

MikeStall:
See #872 for details on localization support.

MikeStall:
See #344 , #548 for localizing runtime errors, like CommonError.cs .

Compile time errors are already localized in https://github.com/microsoft/Power-Fx/tree/main/src/strings 

TBD - error strings from internal exceptions.

LucGenetier:
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/globalization-warnings

PR #2602, opened 2024-08-20
Localize runtime error messages

anderson-joyle:


MikeStall:
See #908, #548

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/RuntimeErrorTests.cs
```diff
@@ -0,0 +1,30 @@
...
+    public class RuntimeErrorTests
+    {
+        [Fact]
+        public void RuntimeErrorLocalizedTests()
+        {
+            var engine = new RecalcEngine();
+            var check = new CheckResult(engine)
+                .SetText("1/0")
+                .SetBindingInfo()
+                .SetDefaultErrorCulture(CultureInfo.CreateSpecificCulture("pt-BR"));
```
MikeStall, line 20: >  .SetDefaultErrorCulture [](http://example.com/codeflow?start=14&length=25)

ErrorCulture is a misnomer - this is really for design time errors via LSP, so would be better named "LSP culture". 

For runtime, we should be using the *runtime culture*.

[https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Interpreter/Environment/SymbolExtensions.cs#L33](https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Interpreter/Environment/SymbolExtensions.cs#L33)

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4694,4 +4694,24 @@
     <value>The specified data source cannot be used with this function.</value>
     <comment>Error Message.</comment>
   </data>
-</root>
\ No newline at end of file
+  <data name="ErrDivByZero" xml:space="preserve">
+    <value>Invalid operation: division by zero.</value>
```
MikeStall, line 4698: >Invalid [](http://example.com/codeflow?start=11&length=7)

The key challenge here will be sharing these with the JS resources in PowerApp's JScript runtime.

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/RuntimeErrorTests.cs
```diff
@@ -0,0 +1,30 @@
...
+// Licensed under the MIT license.
+
+using System.Globalization;
+using System.Linq;
+using Microsoft.PowerFx.Types;
+using Xunit;
+
+namespace Microsoft.PowerFx.Core.Tests
+{
+    public class RuntimeErrorTests
```
MikeStall, line 11: >public [](http://example.com/codeflow?start=4&length=6)

Note that error messages can also be retrieved programmatically:

IfError(""&1/0, FirstError.Message) // "Invalid operation: division by zero." // String


ToString/Parse expressions also depend on runtime culture
Value("123.456") // parses in current culture.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4694,4 +4694,24 @@
...
+  <data name="ErrRuntimeTypeMismatch" xml:space="preserve">
+    <value>Runtime type mismatch.</value>
+    <comment>Error Message.</comment>
+  </data>
+  <data name="ErrArgumentOutOfRange" xml:space="preserve">
+    <value>Argument out of range.</value>
+    <comment>Error Message.</comment>
+  </data>
+  <data name="ErrInvalidCharValue" xml:space="preserve">
+    <value>Invalid Char value, must be in 1...255 range.</value>
```
CarlosFigueiraMSFT, line 4710: The invalid value is not a char, it's a number (which will be converted to a char)

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4694,4 +4694,24 @@
     <value>The specified data source cannot be used with this function.</value>
     <comment>Error Message.</comment>
   </data>
-</root>
\ No newline at end of file
+  <data name="ErrDivByZero" xml:space="preserve">
```
CarlosFigueiraMSFT, line 4697: We should think about splitting the runtime resources from the "compile-time" ones. The latter are used in all platforms during "compilation", but the former are only used in those platforms that use the interpreter, so it will be a perf hit to add additional "runtime" resources that will not be used.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4756,4 +4756,3 @@
     <value>Unsupported untyped/JSON conversion type '{0}' in argument.</value>
     <comment>Error Message shown to user when a unsupported type is passed in type argument of AsType, IsType and ParseJSON functions.</comment>
   </data>
```
CarlosFigueiraMSFT, line 4758: >a> [](http://example.com/codeflow?start=7&length=2)

Removing the `</root>` will be a problem

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -98,7 +98,7 @@ internal async ValueTask<DValue<T>> EvalArgAsync<T>(FormulaValue arg, EvalVisito
                 T t => DValue<T>.Of(t),
                 BlankValue b => DValue<T>.Of(b),
                 ErrorValue e => DValue<T>.Of(e),
-                _ => DValue<T>.Of(CommonErrors.RuntimeTypeMismatch(irContext))
+                _ => DValue<T>.Of(CommonErrors.RuntimeTypeMismatch(irContext, null))
```
CarlosFigueiraMSFT, line 101: >null [](http://example.com/codeflow?start=78&length=4)

Should we pass the culture from the runner/context?

src/strings/PowerFxResources.en-US.resx
anderson-joyle: This has been discussed internally.

src/strings/PowerFxResources.en-US.resx
anderson-joyle: This is the original error message. This can be changed.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/HttpExpressionError.cs
```diff
@@ -14,7 +14,7 @@ public class HttpExpressionError : ExpressionError
         public int StatusCode { get; }
 
         public HttpExpressionError(int statusCode)
-            : base()
+            : base(null)
```
CarlosFigueiraMSFT, line 17: >null [](http://example.com/codeflow?start=19&length=4)

Not really needed - this is the default for the parameterless constructor.

src/libraries/Microsoft.PowerFx.Core/Public/Values/ErrorValue.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
```
CarlosFigueiraMSFT, line 6: Needed? This is the only change in this file

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
```diff
@@ -220,7 +219,8 @@ public override async Task<DValue<BooleanValue>> RemoveAsync(IEnumerable<Formula
 
                 if (!found)
                 {
-                    errors.Add(CommonErrors.RecordNotFound());
+                    // !!!TODO This validation should be moved to the interpreter.
```
CarlosFigueiraMSFT, line 222: >!!!TODO This validation should be moved to the interpreter. [](http://example.com/codeflow?start=23&length=59)

Should we move the entire folder (Values) to the interpreter? The values are a runtime concept, it should not be needed during compile-time.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -380,7 +380,7 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
                     }
                     else
                     {
-                        result = CommonErrors.NotYetImplementedError(node.IRContext, $"Missing func: {func.Name}");
+                        result = CommonErrors.NotYetImplementedError(node.IRContext, CultureInfo, $"{func.Name} function");
```
CarlosFigueiraMSFT, line 383: >{func.Name} function [](http://example.com/codeflow?start=100&length=20)

This needs to be localized

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -616,7 +616,7 @@ public override async ValueTask<FormulaValue> Visit(UnaryOpNode node, EvalVisito
                 return await unaryOp(this, context, node.IRContext, args).ConfigureAwait(false);
             }
 
-            return CommonErrors.NotYetImplementedError(node.IRContext, $"Unary op {node.Op}");
+            return CommonErrors.NotYetImplementedError(node.IRContext, CultureInfo, $"Unary op {node.Op}");
```
CarlosFigueiraMSFT, line 619: >$"Unary op {node.Op} [](http://example.com/codeflow?start=84&length=20)

This should be localized

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -753,7 +753,7 @@ public override async ValueTask<FormulaValue> Visit(RecordFieldAccessNode node,
 
         public override async ValueTask<FormulaValue> Visit(SingleColumnTableAccessNode node, EvalVisitorContext context)
         {
-            return CommonErrors.NotYetImplementedError(node.IRContext, "Single column table access");
+            return CommonErrors.NotYetImplementedError(node.IRContext, CultureInfo, "Single column table access");
```
CarlosFigueiraMSFT, line 756: >"Single column table access" [](http://example.com/codeflow?start=84&length=28)

This should be localized (or throw an exception instead of returning this error)

CarlosFigueiraMSFT:
>         private static string GetDefaultErrorMessage(ErrorKind errorKind)

All those default error messages will need to be localized

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/Library.cs:2473 in 1692e01. [](commit_id = 1692e013404a4112a78859fca7e158d760b79aba, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -153,7 +153,7 @@ public FormulaValue NoElementValue(IRContext context)
 
             public FormulaValue GetDefault(IRContext context)
             {
-                return CommonErrors.DivByZeroError(context);
+                return CommonErrors.DivByZeroError(context, null);
```
CarlosFigueiraMSFT, line 156: >null [](http://example.com/codeflow?start=60&length=4)

Do we need to pass the locale here (and in many other places of this file where there is a new `, null` argument added to the call?

CarlosFigueiraMSFT:
>                     Message = $"Lower value cannot be greater than Upper value",

This needs to be localized

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs:1288 in 1692e01. [](commit_id = 1692e013404a4112a78859fca7e158d760b79aba, deletion_comment = False)

CarlosFigueiraMSFT:
>                     Message = $"Lower value cannot be greater than Upper value",

This needs to be localized

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs:1330 in 1692e01. [](commit_id = 1692e013404a4112a78859fca7e158d760b79aba, deletion_comment = False)

CarlosFigueiraMSFT:
>                     Message = $"Places argument must be big enough to hold the result",

This needs to be localized

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs:1543 in 1692e01. [](commit_id = 1692e013404a4112a78859fca7e158d760b79aba, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryRegEx.cs
```diff
@@ -12,7 +12,9 @@
 using Microsoft.PowerFx.Core.Texl.Builtins;
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Interpreter.Localization;
 using Microsoft.PowerFx.Types;
+using static Microsoft.PowerFx.Syntax.PrettyPrintVisitor;
```
CarlosFigueiraMSFT, line 17: >Microsoft.PowerFx.Syntax.PrettyPrintVisitor [](http://example.com/codeflow?start=13&length=43)

Why do we need this?

CarlosFigueiraMSFT:
>                     Message = "Filter() only supports one predicate",

This needs to be localized

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs:499 in 1692e01. [](commit_id = 1692e013404a4112a78859fca7e158d760b79aba, deletion_comment = False)

CarlosFigueiraMSFT:
>             var invalidSortColumnTemplate = "The specified column '{0}' does not exist or is an invalid sort column type.";

This needs to be localized

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs:694 in 1692e01. [](commit_id = 1692e013404a4112a78859fca7e158d760b79aba, deletion_comment = False)

CarlosFigueiraMSFT:
>                         Message = "Order table can't have duplicate values",

This needs to be localized

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs:831 in 1692e01. [](commit_id = 1692e013404a4112a78859fca7e158d760b79aba, deletion_comment = False)

CarlosFigueiraMSFT:
>             return CommonErrors.CustomError(irContext, "Only managed connections can be refreshed.");

This needs to be localized

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs:1159 in 1692e01. [](commit_id = 1692e013404a4112a78859fca7e158d760b79aba, deletion_comment = False)

CarlosFigueiraMSFT:
>                         Message = "Untyped number is not a valid Decimal value, possible overflow",

This needs to be localized

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUntypedObject.cs:223 in 1692e01. [](commit_id = 1692e013404a4112a78859fca7e158d760b79aba, deletion_comment = False)

CarlosFigueiraMSFT:
>                         Message = "Untyped number is not a valid Decimal value, possible overflow",

This needs to be localized

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUntypedObject.cs:286 in 1692e01. [](commit_id = 1692e013404a4112a78859fca7e158d760b79aba, deletion_comment = False)

CarlosFigueiraMSFT:
>                 Message = $"The untyped object argument to the '{functionName}' function has an incorrect type. Expected: {expectedType}, Actual: {actualValue.Type}."

This needs to be localized

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryUntypedObject.cs:557 in 1692e01. [](commit_id = 1692e013404a4112a78859fca7e158d760b79aba, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
```diff
@@ -574,7 +577,7 @@ private static FormulaValue ExactValueType<T>(IRContext irContext, int index, Fo
             }
             else
             {
-                return CommonErrors.RuntimeTypeMismatch(irContext);
+                return CommonErrors.RuntimeTypeMismatch(irContext, null);
```
CarlosFigueiraMSFT, line 580: >null [](http://example.com/codeflow?start=67&length=4)

This needs to be localized

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
```diff
@@ -872,8 +877,9 @@ private static FormulaValue PositiveNumericNumberChecker(IRContext irContext, in
             {
                 var number = numberArg.Value;
                 if (number < 0)
-                {
-                    return CommonErrors.ArgumentOutOfRange(irContext);
+                {
+                    // The caller will localize this error message.
```
CarlosFigueiraMSFT, line 881: >/ The caller will localize this error message [](http://example.com/codeflow?start=21&length=45)

What do you mean that this will be localized by the caller?

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
```diff
@@ -19,6 +19,7 @@
 using Microsoft.PowerFx.Core.Types.Enums;
 using Microsoft.PowerFx.Functions;
 using Microsoft.PowerFx.Types;
+using static Microsoft.PowerFx.Syntax.PrettyPrintVisitor;
```
CarlosFigueiraMSFT, line 22: Curious: why do we need this?

src/strings/PowerFxRuntimeResources.en-US.resx
```diff
@@ -0,0 +1,220 @@
...
+  <data name="ErrNonFiniteArgument" xml:space="preserve">
+    <value>Arguments to the '{0}' function must be finite.</value>
+    <comment>Error Message.</comment>
+  </data>
+  <data name="ErrNonFiniteResult" xml:space="preserve">
+    <value>The function '{0}' returned a non-finite number.</value>
+    <comment>Error Message.</comment>
+  </data>
+  <data name="ErrInvalidDateTimeParsingError" xml:space="preserve">
+    <value>The Date/Time could not be parsed.</value>
```
CarlosFigueiraMSFT, line 149: >Date/Time [](http://example.com/codeflow?start=15&length=9)

Here we use `Date/Time`, while below we use `date/time` - we should be consistent. Suggestion: "The date or time value cannot be parsed." (and make it consistent throughout)

src/strings/PowerFxRuntimeResources.en-US.resx
```diff
@@ -0,0 +1,220 @@
...
+  <data name="ErrInvalidDateTimeError" xml:space="preserve">
+    <value>Invalid date/time value.</value>
+    <comment>Error Message.</comment>
+  </data>
+  <data name="ErrInvalidBooleanFormatError" xml:space="preserve">
+    <value>The value could not be interpreted as a Boolean.</value>
+    <comment>Error Message.</comment>
+  </data>
+  <data name="ErrInvalidColorFormatError" xml:space="preserve">
+    <value>The value could not be interpreted as a Color.</value>
```
CarlosFigueiraMSFT, line 161: >Color [](http://example.com/codeflow?start=51&length=5)

Color doesn't need to be capitalized

src/strings/PowerFxRuntimeResources.en-US.resx
```diff
@@ -0,0 +1,220 @@
...
+  <data name="ErrUnreachableCodeError" xml:space="preserve">
+    <value>Unknown error.</value>
+    <comment>Error Message.</comment>
+  </data>
+  <data name="ErrNotYetImplementedError" xml:space="preserve">
+    <value>Not implemented: {0}.</value>
+    <comment>Error Message.</comment>
+  </data>
+  <data name="ErrInvalidChain" xml:space="preserve">
+    <value>Invalid Chain: {0}.</value>
```
CarlosFigueiraMSFT, line 173: >Chain [](http://example.com/codeflow?start=19&length=5)

`Invalid chain` - we use sentence casing throughout the product.

CarlosFigueiraMSFT, line 173: >Invalid Chain: {0} [](http://example.com/codeflow?start=11&length=18)

This is a very technical message, we should replace it with something more user-friendly.

src/strings/PowerFxRuntimeResources.en-US.resx
```diff
@@ -0,0 +1,220 @@
...
+  <data name="ErrOnlyPrimitiveValuesAllowed" xml:space="preserve">
+    <value>Only primitive values are allowed for the operation.</value>
+    <comment>Error Message.</comment>
+  </data>
+  <data name="ErrCanNotConvertToNumber" xml:space="preserve">
+    <value>The value '{0}' cannot be converted to a number.</value>
+    <comment>Error Message.</comment>
+  </data>
+  <data name="ErrStartOfWeekInvalid" xml:space="preserve">
+    <value>Expected a value from the StartOfWeek enumeration to indicate how to number the weekdays.</value>
```
CarlosFigueiraMSFT, line 193: >StartOfWeek [](http://example.com/codeflow?start=37&length=11)

Add `{Locked=StartOfWeek}` to the comment to prevent this from being translated.

src/strings/PowerFxRuntimeResources.en-US.resx
```diff
@@ -0,0 +1,220 @@
...
+  <data name="ErrRuntimeExceptionError" xml:space="preserve">
+    <value>An exception has been thrown: {0}</value>
+    <comment>Error Message.</comment>
+  </data>
+  <data name="ErrNameIsNotValid" xml:space="preserve">
+    <value>Name isn't valid. '{0}' isn't recognized.</value>
+    <comment>Error Message.</comment>
+  </data>
+  <data name="ErrMondayZeroValueNotSupported" xml:space="preserve">
+    <value>The MondayZero value, from the StartOfWeek enumeration, is not supported in the WeekNum function.</value>
```
CarlosFigueiraMSFT, line 205: >MondayZero [](http://example.com/codeflow?start=15&length=10)

Add {Locked=MondayZero}{Locked=StartOfWeek}{Locked=WeekNum} to the comment to prevent these from being translated.

src/strings/PowerFxRuntimeResources.pt-BR.resx
```diff
@@ -0,0 +1,128 @@
...
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="ErrDivByZero" xml:space="preserve">
+    <value>Operação inválida: divisão por zero.</value>
+    <comment>Error Message.</comment>
+  </data>
+  <data name="ErrNameIsNotValid" xml:space="preserve">
+    <value>Nome inválidop. '{0}' não é reconhecido.</value>
```
CarlosFigueiraMSFT, line 125: >inválidop [](http://example.com/codeflow?start=16&length=9)

[typo] - inválido

CarlosFigueiraMSFT:
> <?xml version="1.0" encoding="UTF-8" standalone="yes"?>

We typically don't check in translated (satellite) resources, those are managed by the localization team.

---
Refers to: src/strings/PowerFxRuntimeResources.pt-BR.resx:1 in 1692e01. [](commit_id = 1692e013404a4112a78859fca7e158d760b79aba, deletion_comment = False)

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/RuntimeErrorTests.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        [Theory]
+
+        //[InlineData("1/0", "Error 1-2: Invalid operation: division by zero.", "Error 1-2: Operação inválida: divisão por zero.")]
+        //[InlineData("IfError(1/0, FirstError.Message, \"no error\")", "Invalid operation: division by zero.", "Operação inválida: divisão por zero.")]
+
```
CarlosFigueiraMSFT, line 17: Remove code that is commented out

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
anderson-joyle: I ran a quick test to see how big would be the damage moving the value to interpreter. Not as big as I expected. This might be doable but in a separate PR.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryRegEx.cs
anderson-joyle: We dont. Must had been auto added while coding.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
anderson-joyle: `StandardErrorHandlingAsync` (the caller) will localize the errors.

CarlosFigueiraMSFT: That sounds inefficient. We're creating an error here, and later we'll create new errors while replacing the message. Why not do the right thing here in the first place?

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
anderson-joyle: We dont. Auto added by VS. Thanks for catching this.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
anderson-joyle: This has been removed due to shifting the approach.

src/libraries/Microsoft.PowerFx.Interpreter/Localization/RuntimeStringResources.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using Microsoft.PowerFx.Core.Localization;
+
+namespace Microsoft.PowerFx.Interpreter.Localization
+{
+    public static class RuntimeStringResources
```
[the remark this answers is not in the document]
anderson-joyle: Moving back to internal.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/RuntimeErrorTests.cs
```diff
@@ -0,0 +1,40 @@
...
+                var resultLocale = (StringValue)evaluator.Eval(new RuntimeConfig(null, culture));
+
+                Assert.Equal(expectedInvariant, stringValue.Value);
+                Assert.Equal(expectedLocale, resultLocale.Value);
+            }
+            else if (result is ErrorValue errorValue)
+            {
+                // Host can call GetMessageInLocale to get a localized message.
+                Assert.Equal(expectedInvariant, errorValue.Errors.First().ToString());
+                Assert.Equal(expectedLocale, errorValue.Errors.First().GetMessageInLocale(culture, true));
```
anderson-joyle, line 36: >errorValue.Errors.First().GetMessageInLocale(culture, true) [](http://example.com/codeflow?start=45&length=59)

@MikeStall @CarlosFigueiraMSFT

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/RuntimeErrorTests.cs
```diff
@@ -0,0 +1,40 @@
...
+using Microsoft.PowerFx.Types;
+using Xunit;
+
+namespace Microsoft.PowerFx.Core.Tests
+{
+    public class RuntimeErrorTests
+    {
+        [Theory]
+        [InlineData("1/0", "Error 1-2: Invalid operation: division by zero.", "Error 1-2: Operação inválida: divisão por zero.")]
+        [InlineData("IfError(1/0, FirstError.Message, \"no error\")", "Invalid operation: division by zero.", "Operação inválida: divisão por zero.")]
```
anderson-joyle, line 15: >FirstError.Message [](http://example.com/codeflow?start=34&length=18)

@MikeStall @CarlosFigueiraMSFT

src/libraries/Microsoft.PowerFx.Core/Public/Values/CollectionTableValue.cs
MikeStall: definitely separate PR. 

It would also be a massive breaking change.  (moving classes between assemblies is breaking).

src/libraries/Microsoft.PowerFx.Core/Public/Values/PrimitiveValueConversions.cs
```diff
@@ -167,6 +167,7 @@ private static bool TryConvertToDouble(object value, out double result)
 
         internal static ErrorValue OverflowError(IRContext irContext)
         {
+            // !!!TODO move this to interpreter
```
MikeStall, line 170: >!!! [](http://example.com/codeflow?start=15&length=3)

note we shouldn't check these markers in. 

At least, file a github issue and include the link. 
Then we havea  structured way to search for the github issue in code.

src/libraries/Microsoft.PowerFx.Interpreter/Microsoft.PowerFx.Interpreter.csproj
```diff
@@ -29,6 +30,12 @@
     <ProjectReference Include="..\Microsoft.PowerFx.Core\Microsoft.PowerFx.Core.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <EmbeddedResource Include="..\..\strings\PowerFxRuntimeResources.*.resx">
+      <Link>strings\%(Filename)%(Extension)</Link>
+    </EmbeddedResource>
+  </ItemGroup>
```
MikeStall, line 37: Does this impact deployment?
IE, if the resx is missing, do we still run ok and just return english?

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -192,7 +193,7 @@ private async Task<FormulaValue> TryHandleSet(CallNode node, EvalVisitorContext
 
                     if (arg0value is UntypedObjectValue uov && uov.Impl is UntypedObjectBase impl)
                     {
-                        return impl.SetUntypedObject(node.IRContext, (StringValue)arg1value, newValue);
+                        return impl.SetUntypedObject(node.IRContext, (StringValue)arg1value, newValue, CultureInfo);
```
MikeStall, line 196: >CultureInfo [](http://example.com/codeflow?start=103&length=11)

can we avoid passing this? It seems suspicious

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -753,7 +754,7 @@ public override async ValueTask<FormulaValue> Visit(RecordFieldAccessNode node,
 
         public override async ValueTask<FormulaValue> Visit(SingleColumnTableAccessNode node, EvalVisitorContext context)
         {
-            return CommonErrors.NotYetImplementedError(node.IRContext, "Single column table access");
+            throw new NotImplementedException();
```
MikeStall, line 757: >     throw new NotImplementedException(); [](http://example.com/codeflow?start=7&length=41)

switching from a return error value  (soft error) to an exception (hard error) might be significant. Double checking if this is ok?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Values/PrimitiveValueConversions.cs
anderson-joyle: Replaced by github issue links.

src/libraries/Microsoft.PowerFx.Interpreter/Microsoft.PowerFx.Interpreter.csproj
anderson-joyle: Return English.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -543,14 +544,14 @@ private ValueTask<FormulaValue> VisitBinaryOpNode(BinaryOpNode node, EvalVisitor
...
-                    return new ValueTask<FormulaValue>(OperatorDynamicGetField(node, args));
+                    return new ValueTask<FormulaValue>(OperatorDynamicGetField(node, args, CultureInfo));
 
                 default:
                     return new ValueTask<FormulaValue>(CommonErrors.UnreachableCodeError(node.IRContext));
             }
         }
 
-        private static FormulaValue OperatorDynamicGetField(BinaryOpNode node, FormulaValue[] args)
+        private static FormulaValue OperatorDynamicGetField(BinaryOpNode node, FormulaValue[] args, CultureInfo locale)
```
CarlosFigueiraMSFT, line 554: >, CultureInfo locale [](http://example.com/codeflow?start=98&length=20)

With the new resource localization model, there's no need to pass this argument anymore

src/libraries/Microsoft.PowerFx.Interpreter/MaxCallDepthException.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Text;
```
CarlosFigueiraMSFT, line 7: Not needed

src/libraries/Microsoft.PowerFx.Interpreter/Utility.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using System;
+using System.Globalization;
 using Microsoft.PowerFx.Core.IR;
```
CarlosFigueiraMSFT, line 6: Not needed

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs
```diff
@@ -0,0 +1,220 @@
...
+                irContext,
+                NewExpressionError(RuntimeStringResources.ErrInvalidColorFormatError, ErrorKind.InvalidArgument, irContext.SourceContext));
+        }
+
+        // Unknown error.
+        public static ErrorValue UnreachableCodeError(IRContext irContext)
+        {
+            return new ErrorValue(
+                irContext,
+                NewExpressionError(RuntimeStringResources.ErrInvalidColorFormatError, ErrorKind.Validation, irContext.SourceContext));
```
CarlosFigueiraMSFT, line 104: >Validation [](http://example.com/codeflow?start=96&length=10)

If 'UnreachableCode' is something that should never happen, this should be ErrorKind.Internal

CarlosFigueiraMSFT:
>                     return CommonErrors.CustomError(formulaResult.IRContext, string.Format(CultureInfo.InvariantCulture, "Return type should have been {0}, found {1}", retType, formulaResultType));

This should be localized

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/ReflectionFunction.cs:375 in 8b9a944. [](commit_id = 8b9a944b04fa5124f71cd9a9dec927208f6228b8, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryDate.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.PowerFx.Core.IR;
 using Microsoft.PowerFx.Core.Utils;
 using Microsoft.PowerFx.Interpreter.Exceptions;
+using Microsoft.PowerFx.Interpreter.Localization;
 using Microsoft.PowerFx.Types;
```
CarlosFigueiraMSFT, line 13: Needed?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMath.cs
```diff
@@ -7,7 +7,7 @@
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.PowerFx.Core.IR;
-using Microsoft.PowerFx.Interpreter;
+using Microsoft.PowerFx.Interpreter.Localization;
```
CarlosFigueiraMSFT, line 10: >Localization [](http://example.com/codeflow?start=36&length=12)

Needed?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/StandardErrorHandling.cs
```diff
@@ -3,10 +3,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
```
CarlosFigueiraMSFT, line 6: >Globalization [](http://example.com/codeflow?start=13&length=13)

Needed?

CarlosFigueiraMSFT:
> <?xml version="1.0" encoding="UTF-8" standalone="yes"?>

That's a question for the localization team - if they're ok with having this file here for now, then it's ok. If that will mess their pipeline, then we should remove it.

---
In reply to: [2315639609](https://github.com/microsoft/Power-Fx/pull/2602#issuecomment-2315639609) [](http://example.com/codeflow?ancestors=2315639609,2313486363)

---
Refers to: src/strings/PowerFxRuntimeResources.pt-BR.resx:1 in 1692e01. [](commit_id = 1692e013404a4112a78859fca7e158d760b79aba, deletion_comment = False)

src/tests/Microsoft.PowerFx.Connectors.Tests.Shared/PowerPlatformConnectorTests.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.Globalization;
```
CarlosFigueiraMSFT, line 7: >Globalization [](http://example.com/codeflow?start=13&length=13)

Needed?

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/Error.txt
```diff
@@ -163,103 +163,103 @@ Error({Kind:ErrorKind.EditPermissions})
 // ************* Default error messages *************
 
 >> IfError(Error({Kind:ErrorKind.None}), $"{FirstError.Kind}: {FirstError.Message}")
-"0: System no error"
+"0: System no error."
```
CarlosFigueiraMSFT, line 166: >r. [](http://example.com/codeflow?start=18&length=2)

Those changes will need to be made in Power Apps, otherwise those tests will fail there. Or we leave the default error messages for those error kinds without a `.` - your call.

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/BlobTestFunctions.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
```
CarlosFigueiraMSFT, line 6: >Globalization [](http://example.com/codeflow?start=13&length=13)

Needed?

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/BlobTests.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
```
CarlosFigueiraMSFT, line 6: >Globalization [](http://example.com/codeflow?start=13&length=13)

Needed?

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/RecalcEngineTests.cs
```diff
@@ -22,6 +22,7 @@
 using Microsoft.PowerFx.Core.Utils;
 using Microsoft.PowerFx.Functions;
 using Microsoft.PowerFx.Interpreter;
+using Microsoft.PowerFx.Interpreter.Localization;
```
CarlosFigueiraMSFT, line 25: >Localization [](http://example.com/codeflow?start=36&length=12)

Needed?

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/Error.txt
anderson-joyle: Easy call. Ill roll them back.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/CommonErrors.cs
```diff
@@ -0,0 +1,220 @@
...
+                irContext,
+                NewExpressionError(RuntimeStringResources.ErrInvalidChain, ErrorKind.NotSupported, irContext.SourceContext, message));
+        }
+
+        // Max call depth exceeded.
+        internal static ErrorValue MaxCallDepth(IRContext irContext)
+        {
+            return new ErrorValue(
+                irContext,
+                NewExpressionError(RuntimeStringResources.ErrMaxCallDepth, ErrorKind.Validation, irContext.SourceContext));
```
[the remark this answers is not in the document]
anderson-joyle: Isnt `Internal` a better choice since this error comes from an exception?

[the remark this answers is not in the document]
CarlosFigueiraMSFT: In which case does this happen? Because of some input from the user? If so, then InvalidArgument. If it should not happen, then Internal.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

PR #2608, opened 2024-08-23
Add JsonSettings MaxDepth

LucGenetier:


CarlosFigueiraMSFT:
Do we need to add that setting (i.e., is there any scenario where the current value is not enough - or it is too big)?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Json.cs
```diff
@@ -181,11 +188,11 @@ public override void CheckSemantics(TexlBinding binding, TexlNode[] args, DType[
 
             if (!ignoreUnsupportedTypes)
             {
-                if (HasUnsupportedType(dataArgType, supportsLazyTypes, out DType unsupportedNestedType, out var unsupportedColumnName))
+                if (HasUnsupportedType(dataArgType, supportsLazyTypes, _settings.MaxDepth, out DType unsupportedNestedType, out var unsupportedColumnName))
```
CarlosFigueiraMSFT, line 191: >_settings.MaxDepth [](http://example.com/codeflow?start=71&length=18)

This may break Power Apps - which uses this function _defintion_ directly, and the value of `_settings` will be null here.

LucGenetier:
The goal is to allow JSON(First(ServiceNow)) to work when using MaxDepth=1

---
In reply to: [2307093981](https://github.com/microsoft/Power-Fx/pull/2608#issuecomment-2307093981) [](http://example.com/codeflow?ancestors=2307093981)

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Json.cs
LucGenetier: Good point, let me fix

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Json/Environment/PowerFxConfigExtensions.cs
```diff
@@ -21,5 +21,14 @@ public static void EnableJsonFunctions(this PowerFxConfig config)
             config.AddFunction(new IsTypeFunction_UOImpl());
             config.AddFunction(new TypedParseJSONFunctionImpl());
```
MikeStall, line 22: Can this just chain? 
EnableJsonFunctions(this PowerFxConfig config, null);

LucGenetier:
✅ No public API change.

PR #2613, opened 2024-08-26
Split WeekNum tests to better reflect setup

CarlosFigueiraMSFT:
Some tests for WeekNum should be split properly between those with support for strongly-typed enums and those without.


MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #2614, opened 2024-08-27
Dont anonymize known custom functions

anderson-joyle:
We are adding a way for the hosts to declare a list of non-core public functions that should be kept plain when getting function names from an expression.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
@@ -724,6 +724,11 @@ namespace Microsoft.PowerFx
         {
             return ListFunctionVisitor.Run(ApplyParse(), anonymizeUnknownPublicFunctions);
         }
+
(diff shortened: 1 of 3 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -724,6 +724,11 @@ public IEnumerable<string> GetFunctionNames(bool anonymizeUnknownPublicFunctions
         {
             return ListFunctionVisitor.Run(ApplyParse(), anonymizeUnknownPublicFunctions);
         }
+
+        public IEnumerable<string> GetFunctionNames(bool anonymizeUnknownPublicFunctions, ICollection<string> customKnownFunctions)
```
MikeStall, line 728: Can you add a xml doc comment explaining clearly what these parameters do?

LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #2615, opened 2024-08-28
Fix UDFs not properly handled in TryGetDataSource recursion

rick-nguyen:
`FilterUDF():DataSource = Filter(DataSource, Value > 10); FilterUDF2():DataSource = Filter(FilterUDF(), Value > 11);`

`TryGetDsInfo(Callnode)` would not get a datasource for `FilterUDF()`, leading to Binder not marking this node async. This led to an issue of delegation not being preserved on `FilterUDF2()`.


src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/RecalcEngineTests.cs
```diff
@@ -679,33 +681,56 @@ public void ImperativeUserDefinedFunctionTest(string udfExpression, string expre
...
+                "MyDataSource",
+                schema,
+                new TestDelegationMetadata(
+                        new DelegationCapability(DelegationCapability.Filter | DelegationCapability.Count | DelegationCapability.Sort),
+                        schema,
+                        new FilterOpMetadata(
+                            schema,
+                            new Dictionary<DPath, DelegationCapability>(),
+                            new Dictionary<DPath, DelegationCapability>(),
+                            new DelegationCapability(DelegationCapability.Equal | DelegationCapability.GreaterThanOrEqual),
```
CarlosFigueiraMSFT, line 699: >DelegationCapability.Equal | DelegationCapability.GreaterThanOrEqual [](http://example.com/codeflow?start=53&length=68)

The delegation capabilities for this data source is for the `=` and the `>=`  operators, but in the expressions below you are using `>`. Should the UDF be marked as delegatable? If not, why is the test passing - is it a bug?

rick-nguyen: Yes, it appears to be a bug on this check in `IsSupportedOpNode`.

`        public virtual bool IsSupportedOpNode(TexlNode node, OperationCapabilityMetadata metadata, TexlBinding binding)
        {
            Contracts.AssertValue(node);
            Contracts.AssertValue(metadata);

            var binaryOpNode = node.AsBinaryOp();
            if (binaryOpNode == null)
            {
                return false;
            }

            var opDelStrategy = _function.GetOpDelegationStrategy(binaryOpNode.Op, binaryOpNode);
            var binaryOpDelStrategy = (opDelStrategy as BinaryOpDelegationStrategy).VerifyValue();
            Contracts.Assert(binaryOpNode.Op == binaryOpDelStrategy.Op);

            // Check if binaryOp is supported by datasource in the context of filter operation.
            // If this is not allowed then there is no point in checking lhs and rhs
            // It's only safe to do so if lhs and rhs is first/dotted name node as columns (FirstName/DottedName node) can have additional capabilities defined.
            if (!(binaryOpNode.Left is FirstNameNode || binaryOpNode.Left is DottedNameNode) &&
                !(binaryOpNode.Right is FirstNameNode || binaryOpNode.Right is DottedNameNode) &&
                !opDelStrategy.IsOpSupportedByTable(metadata, node, binding))`

rick-nguyen: for binaryOp `Value > 10` it shortcircuits at `!(binaryOpNode.Left is FirstNameNode || binaryOpNode.Left is DottedNameNode)`, and I am not sure what the point of that check is

rick-nguyen: Will file bug

CarlosFigueiraMSFT (approved):
:shipit:

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/RecalcEngineTests.cs
```diff
@@ -679,33 +681,56 @@ public void ImperativeUserDefinedFunctionTest(string udfExpression, string expre
...
+            config.EnableSetFunction();
 
             var recalcEngine = new RecalcEngine(config);
 
-            recalcEngine.AddUserDefinedFunction("A():MyDataSourceTableType = Filter(Sort(MyDataSource,Value), Value > 10);C():MyDataSourceTableType = A();", CultureInfo.InvariantCulture, symbolTable: recalcEngine.EngineSymbols, allowSideEffects: true);
+            recalcEngine.AddUserDefinedFunction("A():MyDataSourceTableType = Filter(MyDataSource, Value > 10);C():MyDataSourceTableType = A(); B():MyDataSourceTableType = Filter(C(), Value > 11);", CultureInfo.InvariantCulture, symbolTable: recalcEngine.EngineSymbols, allowSideEffects: true);
             var func = recalcEngine.Functions.WithName("A");
 
-            if (func is UserDefinedFunction udf)
+            if (func.First() is UserDefinedFunction udf)
```
CarlosFigueiraMSFT, line 714: >func.First() is UserDefinedFunction udf [](http://example.com/codeflow?start=16&length=39)

Instead of checking whether it is a UDF, consider casting it directly to UserDefinedFunction - so if it is not a UDF (which would be a bug) then the test would fail.

src/libraries/Microsoft.PowerFx.Core/Functions/FunctionArgValidators/ConnectedDataSourceInfoArgValidator.cs
```diff
@@ -56,6 +56,11 @@ private bool TryGetDsInfo(CallNode callNode, TexlBinding binding, out IExternalD
                 return false;
             }
 
+            if (function is UserDefinedFunction udf)
+            {
+                return udf.TryGetExternalDataSource(out dsInfo);
+            }
+
             var success = function.TryGetDataSource(callNode, binding, out var external);
```
adithyaselv, line 64: >TryGetDataSource [](http://example.com/codeflow?start=35&length=16)

is there a reason why we cant override this method in UDF to return correct result ?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -70,6 +70,11 @@ public bool TryGetExternalDataSource(out IExternalDataSource dataSource)
             return ArgValidators.DelegatableDataSourceInfoValidator.TryGetValidValue(_binding.Top, _binding, out dataSource);
```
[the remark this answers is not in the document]
adithyaselv: clarified with shantanu that vardiac node makes it non delegable. we can just have a test

adithyaselv (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2620, opened 2024-08-29
Mark UDFs and NFs Async If there's a reference to cross screen control

akshar2401:
Marks UDF and Nf async if they refer to cross screen control and a new binding config to do that is enabled. There's no definite way to say where binder is binding nf or udf rules so config would be used to communicate the marking of top parse node as async upon countering those references. We don't need to mark those async if all they refer to is component definition but there's no definite way to also check that. ExternalTemplate.IsComponent returns true for component defs and instances but we need to check only for the former. However that shouldn't matter as code gen to lazy load component defs would be no op since component defs are loaded upfront.


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/Helpers/SymbolTableHelperMocks/MockGlue.cs
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    bool IBinderGlue.IsPrimaryCommandComponentProperty(IExternalControlProperty externalControlProperty)
+    {
+        return false;
+    }
+
+    bool IBinderGlue.TryGetCdsDataSourceByBind(object lhsInfoData, out IExternalControl o)
+    {
+        o = new DummyExternalControl();
```
CarlosFigueiraMSFT, line 58: >DummyExternalControl [](http://example.com/codeflow?start=16&length=20)

In a TryXXX method, if it returns false then the `out` argument should be set to null (or default, if it's a struct type)

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/Helpers/SymbolTableHelperMocks/MockGlue.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        return false;
+    }
+
+    bool IBinderGlue.IsPrimaryCommandComponentProperty(IExternalControlProperty externalControlProperty)
+    {
+        return false;
+    }
+
+    bool IBinderGlue.TryGetCdsDataSourceByBind(object lhsInfoData, out IExternalControl o)
```
CarlosFigueiraMSFT, line 56: >IExternalControl [](http://example.com/codeflow?start=71&length=16)

Not related to your changes, but what does it mean for a method to return a "CdsDataSource" to return a control?

akshar2401: I am not really sure. i would need to look into it. 3

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

PR #2621, opened 2024-08-29
Checking primary keys when patching

anderson-joyle:
This check applies only to
`Patch(DS, table_of_rows_with_updates)`
`Patch(DS, record_with_keys_and_updates)`


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #2628, opened 2024-09-05
CDP connectors - manage enums per OpenApi spec

LucGenetier:
Add CDP compatibility flag
When enabled, we don't require format=enum

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
@@ -1378,7 +1378,8 @@ namespace Microsoft.PowerFx.Connectors
                         // Ex: Api-Version
                         hiddenRequired = true;
                     }
-                    else if (ConnectorSettings.Compatibility == ConnectorCompatibility.SwaggerCompatibility)
+                    else if (ConnectorSettings.Compatibility == ConnectorCompatibility.SwaggerCompatibility || 
+                             ConnectorSettings.Compatibility == ConnectorCompatibility.CdpCompatibility)
                     {
                         continue;
                     }
@@ -1448,7 +1449,8 @@ namespace Microsoft.PowerFx.Connectors
 
                                             bodyPropertyHiddenRequired = ConnectorSettings.Compatibility != ConnectorCompatibility.PowerAppsCompatibility || !requestBody.Required;
                                         }
-                                        else if (ConnectorSettings.Compatibility == ConnectorCompatibility.SwaggerCompatibility)
+                                        else if (ConnectorSettings.Compatibility == ConnectorCompatibility.SwaggerCompatibility || 
+                                                 ConnectorSettings.Compatibility == ConnectorCompatibility.CdpCompatibility)
                                         {
                                             continue;
                                         }
--- a/src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
@@ -1378,8 +1378,7 @@ namespace Microsoft.PowerFx.Connectors
                         // Ex: Api-Version
                         hiddenRequired = true;
                     }
-                    else if (ConnectorSettings.Compatibility == ConnectorCompatibility.SwaggerCompatibility || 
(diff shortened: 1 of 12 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorSettings.cs
```diff
@@ -64,6 +64,11 @@ public enum ConnectorCompatibility
         // Swagger File Conformity
         // - parameters appear in the order specified in the swagger file
         // - internal visible parameters are completely hidden (required/optional, with or without default value)
-        SwaggerCompatibility = 2
+        SwaggerCompatibility = 2,
+
+        // Swagger File Conformity for CDP connectors
+        // - same as Swagger File Conformity
+        // - does not require format="enum" for identifying enumerations
+        CdpCompatibility = 3
```
MikeStall, line 72: >CdpCompatibility = 3 [](http://example.com/codeflow?start=8&length=20)

can you add test? 

1 swagger, applying each flag

src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
```diff
@@ -1448,7 +1449,8 @@ private ConnectorParameterInternals Initialize()
 
                                             bodyPropertyHiddenRequired = ConnectorSettings.Compatibility != ConnectorCompatibility.PowerAppsCompatibility || !requestBody.Required;
                                         }
-                                        else if (ConnectorSettings.Compatibility == ConnectorCompatibility.SwaggerCompatibility)
+                                        else if (ConnectorSettings.Compatibility == ConnectorCompatibility.SwaggerCompatibility || 
+                                                 ConnectorSettings.Compatibility == ConnectorCompatibility.CdpCompatibility)
```
MikeStall, line 1453: Can we use an extension method instead? 

So we don't need each caller to make this decision. 



ConnectorSettings.Compatibility.IsXXXX()

and 

public static IsXXXX(this ConnectorCompatibility settings)
{
return settings == ConnectorCompatibility.SwaggerCompatibility ||      
                                            settings == ConnectorCompatibility.CdpCompatibility);

}

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -431,26 +431,27 @@ internal static ConnectorType GetConnectorType(this ISwaggerParameter openApiPar
...
-                            else
-                            {
-                                return new ConnectorType(error: $"Unsupported enum type {schema.Enum.GetType().Name}");
-                            }
+                            return new ConnectorType(schema, openApiParameter, FormulaType.Blob);
+                    }
 
-                        default:
-                            return new ConnectorType(schema, openApiParameter, FormulaType.String);
+                    if (schema.Enum != null && (settings.Compatibility == ConnectorCompatibility.CdpCompatibility || schema.Format == "enum"))
```
MikeStall, line 437: > (settings.Compatibility == ConnectorCompatibility.CdpCompatibility || schema.Format == "enum") [](http://example.com/codeflow?start=46&length=95)

similar, can you move this to extension method? 

The idea is that all the policy around what the settings mean is then captured in a central place.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorSettings.cs
MikeStall, line 72: >CdpCompatibility [](http://example.com/codeflow?start=8&length=16)

This really is a misnomer.  It's what we should have originally done for SwaggerCompat,  (but can't change swagger compat due to breaking changes). 
But not sure what a better name would be.

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorSettings.cs
LucGenetier: Not sure what better name to use

PR #2629, opened 2024-09-05
SAP connector fix, avoid crash when getting Items

LucGenetier:
Fixes a crash when using SAP connector


LucGenetier:
✅ No public API change.

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #2631, opened 2024-09-09
Add OnlyOneLevel option to JSON function

LucGenetier:
Allows to serialize only one level of the object 
Needed for complex database/tabular connectors where following relationships is very expensive like Dataverse or ServiceNow

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/JSON_OnlyTopLevel.txt
+++ b/src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/JSON_OnlyTopLevel.txt
@@ -1,0 +1,142 @@
+#SETUP: EnableJsonFunctions
+
+>> JSON(Blank(), JSONFormat.OnlyOneLevel)
+"null"
+
+>> JSON(If(1<0,"true",{a:1}), JSONFormat.OnlyOneLevel)
+Errors: Error 5-25: The JSON function cannot serialize objects of type 'Void'.
+
+>> JSON("hello", JSONFormat.OnlyOneLevel)
+"""hello"""
+
+>> JSON(1.23, JSONFormat.OnlyOneLevel)
+"1.23"
+
+>> JSON(0, JSONFormat.OnlyOneLevel)
+"0"
+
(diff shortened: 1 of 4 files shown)


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/JSON_OnlyTopLevel.txt
```diff
@@ -0,0 +1,142 @@
...
+"{""a"":1,""b"":""hello"",""c"":true}"
+
+>> JSON({a:1,b:"hello",c:true}, JSONFormat.Compact & JSONFormat.OnlyOneLevel)
+"{""a"":1,""b"":""hello"",""c"":true}"
+
+>> Substitute(Substitute(Substitute(JSON({a:1,b:"hello",c:true}, JSONFormat.IndentFour & JSONFormat.OnlyOneLevel), Char(10), "*"), Char(13), ""), "    ", "<t>")
+"{*<t>""a"": 1,*<t>""b"": ""hello"",*<t>""c"": true*}"
+
+>> JSON(Table({a:1,b:"hello",c:true,e:Blank()},{a:-3,b:"world",c:false,d:GUID("01234567-89AB-CDEF-0123-456789ABCDEF")}), JSONFormat.OnlyOneLevel)
+"[{},{}]"
```
CarlosFigueiraMSFT, line 73: >[{},{}] [](http://example.com/codeflow?start=1&length=7)

Going into a table also counts as one level? That doesn't sound very useful... That means that we would only be able to serialize unwrapped arrays of primitives - which we don't support in Power Fx

LucGenetier: Agreed, tables shouldn't count as one level
This will allow us to be consistent between
JSON(First(table))
JSON(FirstN(table, n))

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
No need for this functionality
Relationships shouldn't be followed/returned

PR #2633, opened 2024-09-10
Add new ConnectorSettings "IncludeWebhookFunctions" to allow using more Teams operation

yeze322:
In the Teams connector, the operation "PostCardAndWaitForResponse" contains extension "x-ms-notification-content" and is excluded from the parse result of `GetFunctions`.

In MCS, we want to make this operation available to our internal users to keep parity with Power Automate behavior.

This PR fixes the issue by
1. Creating a new setting "IncludeWebhookFunctions" to conditionally include those "webhook" functions on demand.
2. Consume that new setting when filtering on the webhook functions, it set to true, it won't filter away those functions.

As a result, 4 more operations inside Teams connector will be supported by merging this change.
On MCS side, PFX `GetFunctions` will be invoked with an additional setting parameter `IncludeWebhookFunctions = true`.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/OpenApiParser.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/OpenApiParser.cs
@@ -155,7 +155,7 @@ namespace Microsoft.PowerFx.Connectors
                 string notSupportedReasonForPath = string.Empty;
 
                 // Skip Webhooks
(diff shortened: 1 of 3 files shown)


yeze322:
test pass result:
image

LucGenetier (approved):
:shipit:

PR #2634, opened 2024-09-10
WIP Remove function

anderson-joyle:
Unifying Remove function between PFx/PA.


LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Types.CollectionTableValue`1.FindAsync(Microsoft.PowerFx.Types.RecordValue,System.Threading.CancellationToken,System.Boolean)

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Remove.cs
```diff
@@ -0,0 +1,626 @@
...
+using Microsoft.PowerFx.Core.Utils;
+using Microsoft.PowerFx.Syntax;
+
+namespace Microsoft.PowerFx.Core.Texl.Builtins
+{
+    // Remove(collection:*[], item1:![], item2:![], ..., ["All"])
+
+    // !!!TODO MOVE THIS TO src/Cloud/DocumentServer.Core/Document/Texl/Functions/Libraries/BuiltinFunctions.cs
+    //[RequiresErrorContext]
+    internal class RemoveFunction : BuiltinFunction, ISuggestionAwareFunction
```
anderson-joyle, line 27: >RemoveFunction [](http://example.com/codeflow?start=19&length=14)

RequiresErrorContext

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Remove.cs
```diff
@@ -0,0 +1,626 @@
...
+            // For the first argument, we need only metadata. No actual data from datasource is required.
+            return paramIndex > 0;
+        }
+    }
+
+    // Remove(collection:*[], source:*[], ["All"])
+    // !!!TODO MOVE THIS TO src/Cloud/DocumentServer.Core/Document/Texl/Functions/Libraries/BuiltinFunctions.cs
+    //[RequiresErrorContext]
+    //[TexlRuntimeNameOverride(Suffix = "All")]
+    internal class RemoveAllFunction : BuiltinFunction
```
anderson-joyle, line 364: >RemoveAllFunction [](http://example.com/codeflow?start=19&length=17)

[RequiresErrorContext]
[TexlRuntimeNameOverride(Suffix = "All")]

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Config/Features.cs
```diff
@@ -91,6 +91,11 @@ public sealed class Features
         /// </summary>
         internal bool IsLookUpReductionDelegationEnabled { get; set; }
 
+        /// <summary>
+        /// Temporary flag.
```
CarlosFigueiraMSFT, line 95: >Temporary flag. [](http://example.com/codeflow?start=12&length=15)

Can you add more details about what this flag does?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Remove.cs
```diff
@@ -0,0 +1,579 @@
...
+            int argCount = argTypes.Length;
+            for (int i = 1; i < argCount; i++)
+            {
+                DType argType = argTypes[i];
+
+                if (!argType.IsRecord)
+                {
+                    if (argCount >= 3 && i == argCount - 1)
+                    {
+                        if (context.AnalysisMode)
```
CarlosFigueiraMSFT, line 125: >context.AnalysisMode [](http://example.com/codeflow?start=28&length=20)

Why is this check for the stringly-typed enum done based on context.AnalysisMode? Shouldn't it be done based on Features.StronglyTypedBuiltinEnums?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Remove.cs
```diff
@@ -0,0 +1,579 @@
...
+                    }
+                }
+                
+                // Only warn about no-op record inputs if there are no data sources that would use reference identity for comparison.
+                else if (!collectionType.AssociatedDataSources.Any() && !recordAcceptsCollection)
+                {
+                    errors.EnsureError(DocumentErrorSeverity.Warning, args[i], TexlStrings.ErrCollectionDoesNotAcceptThisType);
+                }
+
+                if (!context.AnalysisMode)
```
CarlosFigueiraMSFT, line 172: >AnalysisMode [](http://example.com/codeflow?start=29&length=12)

Like the previous comment, should this be based on AnalysisMode? Or PFxV1 would be more appropriate?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Remove.cs
```diff
@@ -0,0 +1,579 @@
...
+                        fValid = false;
+                        if (!SetErrorForMismatchedColumns(collectionType, argType, args[i], errors, context.Features))
+                        {
+                            errors.EnsureError(DocumentErrorSeverity.Severe, args[i], TexlStrings.ErrTableDoesNotAcceptThisType);
+                        }
+                    }
+                }
+            }
+
+            if (context.AnalysisMode)
```
CarlosFigueiraMSFT, line 189: >AnalysisMode [](http://example.com/codeflow?start=24&length=12)

This definitely should not be context.AnalysisMode, but `!context.Features.PowerFxV1CompatibilityRules` instead

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Remove.cs
```diff
@@ -0,0 +1,579 @@
...
+
+            var identifiers = new List<Identifier>
+            {
+                firstNameNode.Ident
+            };
+            return identifiers;
+        }
+
+        /// <summary>
+        /// As Remove uses the source record in it's entirity to find the entry in table, uses deepcompare at runtime, we need all fields from source.
```
CarlosFigueiraMSFT, line 242: >entirity [](http://example.com/codeflow?start=53&length=8)

[typo] entirety

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Remove.cs
```diff
@@ -0,0 +1,579 @@
...
+                    // The item schema should be compatible with the collection schema.
+                    if (!checkAggregateNames)
+                    {
+                        fValid = false;
+                        if (!SetErrorForMismatchedColumns(collectionType, argType, args[i], errors, context.Features))
+                        {
+                            errors.EnsureError(DocumentErrorSeverity.Severe, args[i], TexlStrings.ErrTableDoesNotAcceptThisType);
+                        }
+                    }
+                }
```
CarlosFigueiraMSFT, line 186: This check wasn't in the original Remove implementation; what is the scenario that it is covering? In the existing implementation, there is already a check for the types of the arguments compared to the data source / collection.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Remove.cs
```diff
@@ -0,0 +1,579 @@
...
+            Contracts.AssertValue(binding);
+            Contracts.Assert(binding.IsPageable(Contracts.VerifyValue(args[paramIndex])));
+
+            // For the first argument, we need only metadata. No actual data from datasource is required.
+            return paramIndex > 0;
+        }
+    }
+
+    // Remove(collection:*[], source:*[], ["All"])
+    internal class RemoveAllFunction : BuiltinFunction
```
CarlosFigueiraMSFT, line 342: >RemoveAllFunction [](http://example.com/codeflow?start=19&length=17)

This function looks very similar to the previous one; can we have a single one?

src/libraries/Microsoft.PowerFx.Core/Types/Enums/BuiltInEnums.cs
```diff
@@ -199,5 +199,15 @@ private static Dictionary<string, object> TraceSeverityDictionary()
...
+
+        public static readonly EnumSymbol RemoveFlagsEnum = new EnumSymbol(
+            new DName(LanguageConstants.RemoveFlagsEnumString),
+            DType.String,
+            new Dictionary<string, object>()
+            {
+                { "First", "first" },
+                { "All", "all" },
+            },
+            canCoerceFromBackingKind: true);
```
CarlosFigueiraMSFT, line 211: >canCoerceFromBackingKind: true [](http://example.com/codeflow?start=12&length=30)

This should not be coercible - if we use strongly-typed built-in enums, we want makers to write Remove.All, and not "all" .

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Remove.cs
CarlosFigueiraMSFT: Ah, I noticed the difference (record arguments vs. table arguments). It can probably be refactored, though, to a common base type, so that most common code can be written once.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/Mutation/MutationUtils.cs
```diff
@@ -47,5 +45,89 @@ public static DValue<RecordValue> MergeRecords(IEnumerable<FormulaValue> records
...
+            if (args[1] is TableValue sourceTable)
+            {
+                recordsToRemove = sourceTable.Rows.Select(row => row.Value).ToList();
+            }
+            else
+            {
+               recordsToRemove = args
+                    .Skip(1)
+                    .Where(arg => arg is RecordValue)
+                    .Select(row => (RecordValue)row)
```
CarlosFigueiraMSFT, line 108: `.OfType<RecordValue>()`

CarlosFigueiraMSFT:
> #SETUP: PowerFxV1CompatibilityRules

The "base" remove test (Remove.txt) already uses the PowerFxV1CompatibilityRules setup; do we need this file, or can we merge it into the base one?

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/Remove_V1Compact.txt:1 in f4ede39. [](commit_id = f4ede391db4c8cca337219954bf8e2b01942ac30, deletion_comment = False)

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/PADIntegrationTests.cs
```diff
@@ -164,7 +164,7 @@ public void DataTableEvalTest2()
             var result7 = engine.Eval("Patch(robintable, First(robintable),{Names:\"new-name\"});robintable", options: opt);
             Assert.Equal("Table({Names:\"new-name\",Scores:10},{Names:\"name3\",Scores:30},{Names:\"name100\",Scores:10})", ((DataTableValue)result7).Dump());
 
-            var result8 = engine.Eval("Remove(robintable, {Scores:10}, \"All\");robintable", options: opt);
+            var result8 = engine.Eval("Remove(robintable, {Scores:10}, RemoveFlags.All);robintable", options: opt);
```
CarlosFigueiraMSFT, line 167: >RemoveFlags.All [](http://example.com/codeflow?start=71&length=15)

Will this be considered a breaking change for PAD?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Remove.cs
anderson-joyle: My understanding is that the returning types are different between PA/PFx.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Remove.cs
anderson-joyle: The original code wasn't checking if `context.Features.PowerFxV1CompatibilityRules` was active or not. The idea here is, whatever PA was returning, keep returning it. Since PFxV1 in Core is active, return `Void`.

src/libraries/Microsoft.PowerFx.Core/Types/Enums/BuiltInEnums.cs
anderson-joyle: What about other enums that are also coercible (DateTimeFormatEnum)? I unmderstand we want to force makers to to write `Remove.All` but shouldn't we also take in consideration that this is still supported in PA?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Remove.cs
CarlosFigueiraMSFT: When Power Apps moves to PFxV1, it should have the same behavior as the C# interpreter. Remove is one of the functions that we will update the return type on V1 to be Void.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Remove.cs
CarlosFigueiraMSFT: If I understand correctly, this is not about the return type of the function, but for input validation. And as I mentioned in another comment, whenever we expose Power Fx V1 compatibility in Power Apps, it should have the same behavior as the C# interpreter.

src/libraries/Microsoft.PowerFx.Core/Types/Enums/BuiltInEnums.cs
CarlosFigueiraMSFT: DateTimeFormat is an enum that is passed to the Text function - and it can also take a string value.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #2635, opened 2024-09-11
Try fallback to "value" and "displayName" when building connector suggestions API response

yeze322:
Address the schema and actual response misalignment issue especially for Teams Connector.

**Context**:
In Teams connector, the suggestions API "GetMessageLocations" is declared as below:

```ts
// dynamic schema
"x-ms-dynamic-values": {
    "operationId": "GetMessageLocations",
    "value-path": "id",
    "value-title": "displayName",
    "value-collection": "value",
    "parameters": {
        "messageType": "ParentMessage",
        "poster": {
            "parameter": "poster"
        }
    }
}

// response type
{
  "type": "object",
  "properties": {
      "locations": {
          "type": "array",
          "items": {
              "type": "object"
          },
          "description": "valid locations to post a message or reply, make verbose"
      }
  }
}
```

It's a historical issue (PAuto supports weak type and consumes this format) that the actual response format ("value" and "displayName") is misaligned with the expected format ("id" and "displayName"). Then it caused the MCS suggestions API returning empty result.

**Proposed solution**:
Fallback to accept "value" and "displayName" when building the suggestion response.
This approached was proved to work in MCS but since some version update there was a regression.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
@@ -1177,7 +1177,7 @@ namespace Microsoft.PowerFx.Connectors
                     JsonElement value = ExtractFromJson(jElement, cdv.ValuePath);
 
                     // Note: Some connectors have misalignment between the Swagger definition and the actual response, caused suggestion API
-                    // returning empty result.For example, Teams declares the GetMessageLocations response as List<{id, displayName}> but
+                    // returning empty result. For example, Teams declares the GetMessageLocations response as List<{id, displayName}> but
                     // in fact returns List<{value, displayName}>.
                     // Fallback to "displayName" and "value" which are the most commonly used property names in suggestion response.
                     if (title.ValueKind == JsonValueKind.Undefined)
--- a/src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
@@ -1176,6 +1176,19 @@ namespace Microsoft.PowerFx.Connectors
                     JsonElement title = ExtractFromJson(jElement, cdv.ValueTitle);
                     JsonElement value = ExtractFromJson(jElement, cdv.ValuePath);
 
+                    // Note: Some connectors have misalignment between the Swagger definition and the actual response, caused suggestion API returns empty result.
+                    // For example, Teams declares the GetMessageLocations response as List<{id, displayName}> but actually returns List<{value, displayName}>.
+                    // Fallback to "displayName" and "value" which are the most commonly used property names in suggestion response.
+                    if (title.ValueKind == JsonValueKind.Undefined)
+                    {
+                        title = ExtractFromJson(jElement, "displayName");
+                    }
+
+                    if (value.ValueKind == JsonValueKind.Undefined)
+                    { 
+                        value = ExtractFromJson(jElement, "value");
+                    }
+
                     if (title.ValueKind == JsonValueKind.Undefined || value.ValueKind == JsonValueKind.Undefined)
                     {
                         continue;
@@ -1214,6 +1227,16 @@ namespace Microsoft.PowerFx.Connectors
             {
                 JsonElement title = ExtractFromJson(jElement, cdl.ItemTitlePath);
                 JsonElement value = ExtractFromJson(jElement, cdl.ItemValuePath);
+
+                if (title.ValueKind == JsonValueKind.Undefined)
+                {
+                    title = ExtractFromJson(jElement, "displayName");
+                }
+
+                if (value.ValueKind == JsonValueKind.Undefined)
+                {
+                    value = ExtractFromJson(jElement, "value");
+                }
 
                 if (title.ValueKind == JsonValueKind.Undefined || value.ValueKind == JsonValueKind.Undefined)
                 {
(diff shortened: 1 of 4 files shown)


src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
```diff
@@ -334,14 +334,15 @@ internal void SetUnsupported(string notSupportedReason)
         /// <param name="connectorParameter">Parameter for which we need a set of suggestions.</param>
         /// <param name="runtimeContext">Runtime connector context.</param>
         /// <param name="cancellationToken">Cancellation token.</param>
+        /// <param name="connectorSettings">Connector settings.</param>
         /// <returns>ConnectorParameters class with suggestions.</returns>
-        public async Task<ConnectorParameters> GetParameterSuggestionsAsync(NamedValue[] knownParameters, ConnectorParameter connectorParameter, BaseRuntimeConnectorContext runtimeContext, CancellationToken cancellationToken)
+        public async Task<ConnectorParameters> GetParameterSuggestionsAsync(NamedValue[] knownParameters, ConnectorParameter connectorParameter, BaseRuntimeConnectorContext runtimeContext, CancellationToken cancellationToken, ConnectorSettings connectorSettings = null)
```
LucGenetier, line 339: >ConnectorSettings connectorSettings = null [](http://example.com/codeflow?start=226&length=42)

Why do we need a parameter here as it's already present in the class (line 51)

src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
```diff
@@ -1174,8 +1176,25 @@ private async Task<ConnectorEnhancedSuggestions> GetConnectorSuggestionsFromDyna
...
                     JsonElement title = ExtractFromJson(jElement, cdv.ValueTitle);
-                    JsonElement value = ExtractFromJson(jElement, cdv.ValuePath);
-
+                    JsonElement value = ExtractFromJson(jElement, cdv.ValuePath);
+
+                    // Note: Some connectors have misalignment between the Swagger definition and the actual response, caused suggestion API
+                    // returning empty result. For example, Teams declares the GetMessageLocations response as List<{id, displayName}> but
+                    // in fact returns List<{value, displayName}>.
+                    // Fallback to "displayName" and "value" which are the most commonly used property names in suggestion response.
+                    if (connectorSettings != null && connectorSettings.AllowSuggestionMappingFallback)
```
LucGenetier, line 1185: >connectorSettings != null && connectorSettings.AllowSuggestionMappingFallback [](http://example.com/codeflow?start=24&length=77)

nit
could be written `connectorSetting?s.AllowSuggestionMappingFallback == true`

src/tests/Microsoft.PowerFx.Connectors.Tests.Shared/OpenApiParserTests.cs
```diff
@@ -1374,6 +1374,49 @@ public async Task Teams_PostCardAndWaitForResponse()
...
+        [Fact]
+        public async Task Teams_GetMessageLocation_WithSuggestions()
+        {
+            using var testConnector = new LoggingTestServer(@"Swagger\Teams.json", _output);
+            using var httpClient = new HttpClient(testConnector);
+            using PowerPlatformConnectorClient client = new PowerPlatformConnectorClient("https://tip1002-002.azure-apihub.net", "7592282b-e371-e3f6-8e04-e8f23e64227c" /* environment Id */, "shared-cardsforpower-eafc4fa0-c560-4eba-a5b2-3e1ebc63193a" /* connectionId */, () => "eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiIsIng1dC...", httpClient) { SessionId = "a41bd03b-6c3c-4509-a844-e8c51b61f878" };
+
+            BaseRuntimeConnectorContext runtimeContext = new TestConnectorRuntimeContext("DV", client, console: _output);
+
+            ConnectorFunction[] functions = OpenApiParser.GetFunctions(new ConnectorSettings("DV") { Compatibility = ConnectorCompatibility.SwaggerCompatibility }, testConnector._apiDocument).ToArray();
```
LucGenetier, line 1388: >ConnectorSettings [](http://example.com/codeflow?start=75&length=17)

This is where I'd set the new flag
yes, that will be connector wide but i think that's acceptable as long as it doesn't impact other connectors (other than Teams)

LucGenetier (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
yeze322: updated as we discussed

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorSettings.cs
```diff
@@ -36,6 +37,22 @@ public ConnectorSettings(string @namespace)
...
+        /// <summary>
+        /// Enables the suggestion mapping logic to use "value" and "displayName" as fallback property names.
+        /// Serves as a safeguard when the actual response from the suggestion API doesn't align with the Swagger specification.
+        /// </summary>
+        public bool AllowSuggestionMappingFallback { get; init; } = false;
+
+        /// <summary>
+        /// Suggestion item's value fallback key. Defaults to "value". 
+        /// </summary>
+        public string SuggestionMappingFallbackValueKey { get; init; } = "value";
```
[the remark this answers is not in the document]
yeze322: not really, pushed a change to keep it simple as discussed

src/tests/Microsoft.PowerFx.Connectors.Tests.Shared/OpenApiParserTests.cs
yeze322: yes, updated

PR #2639, opened 2024-09-16
JSON function: serialize UO

LucGenetier:


src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
```diff
@@ -319,7 +319,63 @@ public void Visit(TimeValue timeValue)
...
+                    else if (type is ExternalType externalType)
+                    {
+                        if (externalType.Kind == ExternalTypeKind.Array)
+                        {
+                            _writer.WriteStartArray();
+
+                            for (var i = 0; i < untypedObject.GetArrayLength(); i++)
+                            {
+                                IUntypedObject row = untypedObject[i];
+                                Visit(row);
```
CarlosFigueiraMSFT, line 350: > Visit(row);  [](http://example.com/codeflow?start=31&length=13)

Since a host may provide its own implementation of IUntypedObject, we should add some limit to prevent a stack overflow in this call.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/JSON.txt
```diff
@@ -133,3 +133,60 @@ Error({Kind:ErrorKind.Div0})
...
+>> JSON(ParseJSON("""John"""), JSONFormat.IgnoreUnsupportedTypes)
+"""John"""
+
+>> JSON(ParseJSON("true"), JSONFormat.IgnoreUnsupportedTypes)
+"true"
+
+>> JSON(ParseJSON("false"), JSONFormat.IgnoreUnsupportedTypes)
+"false"
+
+>> JSON(ParseJSON("{""a"": 1}"), JSONFormat.IgnoreUnsupportedTypes)
```
CarlosFigueiraMSFT, line 167: >ParseJSON("{""a"": 1}") [](http://example.com/codeflow?start=8&length=23)

This test already shows it briefly, but we should have one case where it is more explicit that whitespaces are not round-tripped, something like `JSON(ParseJSON("   {\n  ""a"" : \t1\r}\r\n"))` ==> `"{""a"":1}"`

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/JSON.txt
```diff
@@ -133,3 +133,60 @@ Error({Kind:ErrorKind.Div0})
 // Flattening nested tables
 >> JSON([[1,2,3],[4,5],[6]], JSONFormat.FlattenValueTables)
 "[[1,2,3],[4,5],[6]]"
+
+>> JSON(ParseJSON("{}"))
```
MikeStall, line 137: >>> [](http://example.com/codeflow?start=0&length=2)

Can we call out an example that does not round-trip? 
And mention in comments // Does not roundtrip  because ... 

The test cases help us document the feature.

LucGenetier: C# tests give an example

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
```diff
@@ -319,7 +319,68 @@ public void Visit(TimeValue timeValue)
...
+                    Visit(untypedObjectValue.Impl);
+                }
+
+                private void Visit(IUntypedObject untypedObject, int index = 0)
+                {
+                    FormulaType type = untypedObject.Type;
+
+                    if (index > 40)
+                    {
+                        return;
```
CarlosFigueiraMSFT, line 331: >return [](http://example.com/codeflow?start=24&length=6)

We should return an error instead of silently not serializing what the user has

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
```diff
@@ -319,7 +319,68 @@ public void Visit(TimeValue timeValue)
...
+                    }
+
+                    if (type is StringType)
+                    {
+                        _writer.WriteStringValue(untypedObject.GetString());
+                    }
+                    else if (type is NumberType || (type is ExternalType et && et.Kind == ExternalTypeKind.UntypedNumber))
+                    {
+                        _writer.WriteNumberValue(untypedObject.GetDouble());
```
CarlosFigueiraMSFT, line 340: >.GetDouble() [](http://example.com/codeflow?start=62&length=12)

Will it lose precision if we have a decimal number? For example, `JSON(Decimal(ParseJSON("123456789012345.6789012345678")))` does return `"123456789012345.6789012345678"`. If we were doing JSON on the result of ParseJSON itself, would it truncate it?

LucGenetier: IUntypedObject doesn't support decimal, by design

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
MikeStall: Is that true? I see:
IUntypedObject.GetDecimal()

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
```diff
@@ -319,7 +319,76 @@ public void Visit(TimeValue timeValue)
...
                 {
-                    throw new ArgumentException($"Unable to serialize type {untypedObjectValue.GetType().FullName} to Json format.");
+                    Visit(untypedObjectValue.Impl);
+                }
+
+                private void Visit(IUntypedObject untypedObject, int index = 0)
+                {
+                    FormulaType type = untypedObject.Type;
+
+                    if (index > 40)
```
MikeStall, line 329: >40 [](http://example.com/codeflow?start=32&length=2)

can you hoist to a const field?  We shouldn't have "magic" numerical literals in the code like this.

I'm sure at some future point, we'll want this to come from some config object, but fixed is fine now.  

FWIW - 40 is probably very high.

LucGenetier: I've used 20 instead

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
LucGenetier: Done

MikeStall (approved):
:shipit:

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/JSON.txt
```diff
@@ -133,3 +133,75 @@ Error({Kind:ErrorKind.Div0})
...
+Error({Kind:ErrorKind.Custom})
+
+>> JSON(ParseJSON("[[1]]"))
+"[[1]]"
+
+// Depth 41
+>> JSON(ParseJSON("[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[1]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]"))
+Error({Kind:ErrorKind.Custom})
+
+>> JSON(Decimal(ParseJSON("123456789012345.6789012345678")))
```
CarlosFigueiraMSFT, line 205: >JSON(Decimal(ParseJSON("123456789012345.6789012345678"))) [](http://example.com/codeflow?start=3&length=57)

You shouldn't need the Decimal call there - otherwise you won't be testing the new logic (serializing untyped values)

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/JSON.txt
```diff
@@ -133,3 +133,75 @@ Error({Kind:ErrorKind.Div0})
...
+>> JSON(ParseJSON("[""True"", ""False""]"))
+"[""True"",""False""]"
+
+>> JSON(ParseJSON("   {  ""a"" :  1 }  "))
+"{""a"":1}"
+
+>> JSON(ParseJSON("{""a"": {""a"": 1}}"))
+"{""a"":{""a"":1}}"
+
+// Depth 41
```
CarlosFigueiraMSFT, line 194: Since we lowered the maxdepth to 20, the test should have only a depth of 21 (same for test on line 202)

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/JSON.txt
LucGenetier: I'll add an extra test

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
```diff
@@ -319,7 +321,80 @@ public void Visit(TimeValue timeValue)
...
+                private void Visit(IUntypedObject untypedObject, int index = 0)
+                {
+                    FormulaType type = untypedObject.Type;
+
+                    if (index > _maxDepth)
+                    {
+                        IRContext irContext = IRContext.NotInSource(FormulaType.UntypedObject);
+                        ErrorValues.Add(new ErrorValue(irContext, new ExpressionError()
+                        {
+                            Message = "Max depth reached while traversing JSON payload",
```
CarlosFigueiraMSFT, line 336: >"Max depth reached while traversing JSON payload" [](http://example.com/codeflow?start=38&length=49)

This should be localized; Anderson has done some work to localize runtime errors already.

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
```diff
@@ -319,7 +321,80 @@ public void Visit(TimeValue timeValue)
...
+                            Span = irContext.SourceContext,
+                            Kind = ErrorKind.Custom
```
CarlosFigueiraMSFT, line 338: >Custom [](http://example.com/codeflow?start=45&length=6)

InvalidArgument - custom errors are those used for customer-created errors.

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
CarlosFigueiraMSFT, line 336: >Max [](http://example.com/codeflow?start=39&length=3)

`Maximum` - in a customer-facing message we should not abbreviate

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/JsonSerializeUOTests.cs
```diff
@@ -0,0 +1,173 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.PowerFx.Core.Tests;
+using Microsoft.PowerFx.Types;
+using Xunit;
+
+#pragma warning disable CA1065
```
CarlosFigueiraMSFT, line 14: Do we need to re-enable it at the end of the file, or is the scope of the #pragma directive the file itself?

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/JsonSerializeUOTests.cs
LucGenetier: Seems to be per file
[https://stackoverflow.com/questions/40044067/what-is-the-scope-of-a-c-sharp-pragma-directive](https://stackoverflow.com/questions/40044067/what-is-the-scope-of-a-c-sharp-pragma-directive)

CarlosFigueiraMSFT:
> #SETUP: EnableJsonFunctions, PowerFxV1CompatibilityRules

Can you split the tests that need this flag into a separate file? We want to continue running those existing tests in Power Apps even if PFxV1 is disabled.

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/JSON.txt:1 in cfc1021. [](commit_id = cfc1021e2c50afa29709af5dbd2451cfb49f78b7, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/JSON.txt
```diff
@@ -133,3 +133,80 @@ Error({Kind:ErrorKind.Div0})
...
+>> JSON(ParseJSON("[[[[[[[[[[[[[[[[[[[[[1]]]]]]]]]]]]]]]]]]]]]"))
+Error({Kind:ErrorKind.InvalidArgument})
+
+>> JSON(Decimal(ParseJSON("123456789012345.6789012345678")))
+"123456789012345.6789012345678"
+
+>> JSON(ParseJSON("123456789012345.6789012345678"))
+"123456789012345.6789012345678"
+
+>> JSON(ParseJSON("""\u0048\u0065\u006c\u006c\u006f"""))
```
CarlosFigueiraMSFT, line 211: >> JSON(ParseJSON("""\u0048\u0065\u006c\u006c\u006f""")) [](http://example.com/codeflow?start=1&length=55)

Can you add a comment saying something like "Round-trip is not guaranteed - escaped characters that don't need escaping will not be re-escaped when JSON-ified".

Similar comment on the test on line 188 (that whitespaces are not round-tripped)

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
```diff
@@ -319,7 +324,80 @@ public void Visit(TimeValue timeValue)
...
+                        }));
+                        _writer.WriteStringValue(string.Empty);
+                        return;
+                    }
+
+                    if (type is StringType)
+                    {
+                        _writer.WriteStringValue(untypedObject.GetString());
+                    }
+                    else if (type is DecimalType || (type is ExternalType et && et.Kind == ExternalTypeKind.UntypedNumber))
```
gregli-msft, line 351: UntypedNumber needs to be serialized unmodified.  These tests are currently failing:

```
>> JSON(ParseJSON("1e300"))
One of the identified items was in an invalid format.

>> JSON(ParseJSON("1111111111111111111111111111111.2222222222222222222222222222222222"))
One of the identified items was in an invalid format.
```
Even with the super long decimal, it should be preserved.  We don't know if it is float or decimal until it is used.

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
```diff
@@ -7,6 +7,7 @@
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Security.Cryptography;
 using System.Text;
```
CarlosFigueiraMSFT, line 11: Needed?

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/JSON_PowerFxV1CompatibilityRules.txt
```diff
@@ -0,0 +1,87 @@
+#SETUP: PowerFxV1CompatibilityRules
+#SETUP: EnableJsonFunctions
+
+>> JSON(ParseJSON("{}"))
```
CarlosFigueiraMSFT, line 4: Most of those tests will work in Power Apps both with and without PFx1.0 enabled, so they can go in the "base" JSON.txt test file. The only ones which require V1 are the ones with decimal, and the ones that rely on the "untyped number" logic, where numbers are stored as raw values (strings) until needed.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4760,4 +4760,8 @@
     <value>Unsupported untyped/JSON conversion type '{0}' in argument.</value>
     <comment>Error Message shown to user when a unsupported type is passed in type argument of AsType, IsType and ParseJSON functions.</comment>
   </data>
+  <data name="ErrReachedMaxJsonDepth" xml:space="preserve">
```
CarlosFigueiraMSFT, line 4763: >ErrReachedMaxJsonDepth [](http://example.com/codeflow?start=14&length=22)

This is a runtime error message, so it should go to the resources for runtime, not the one for core.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
LucGenetier: Not really ;)

LucGenetier:
>                     foreach (NamedValue namedValue in recordValue.Fields)

order by culture invariant

---
Refers to: src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs:265 in 1a33203. [](commit_id = 1a33203fd86a94965ab4b87e2a8def6165a0f42f, deletion_comment = False)

src/strings/PowerFxResources.en-US.resx
LucGenetier: We'd have to put in Json assembly as we can't access Interpreter resources
If that's not a huge deal for now, I'd keep it in Core

LucGenetier: [https://github.com/microsoft/Power-Fx/issues/2646](https://github.com/microsoft/Power-Fx/issues/2646)

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
MikeStall: if not needed, can you remove it?

MikeStall:
>                     foreach (NamedValue namedValue in recordValue.Fields)

StringComparison.Ordinal ? 

That's implest, linguistic agnostic. 
Invariant till uses linguistic rules, but just en-Us based. 


---
In reply to: [2368783168](https://github.com/microsoft/Power-Fx/pull/2639#issuecomment-2368783168) [](http://example.com/codeflow?ancestors=2368783168)

---
Refers to: src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs:265 in 1a33203. [](commit_id = 1a33203fd86a94965ab4b87e2a8def6165a0f42f, deletion_comment = False)

MikeStall:
>                     foreach (DValue<RecordValue> row in tableValue.Rows)

we should accept a cancellation token (passed in from eval runner) and check it at each loop and Visit method. 




---
Refers to: src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs:294 in 980ef8b. [](commit_id = 980ef8b8e48a353d4df233104dba4f08624843fe, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
LucGenetier: VS seems to be in love for it...

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/JSON_DisablePowerFxV1CompatibilityRules.txt
```diff
@@ -0,0 +1,84 @@
...
+>> JSON(ParseJSON("1"))
+"1"
+
+>> JSON(ParseJSON("1.77"))
+"1.77"
+
+>> JSON(ParseJSON("-871"))
+"-871"
+
+>> JSON(ParseJSON("""John"""))
```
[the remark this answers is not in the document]
LucGenetier: As there is no change with or without PowerFxV1CompatibilityRules, I'll remove this file

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/IAsyncTexlFunction4.cs
```diff
@@ -11,6 +11,6 @@ namespace Microsoft.PowerFx.Core.Functions
     // A Texl function capable of async invokes, using TimeZoneInfo and IRContext. 
     internal interface IAsyncTexlFunction4
     {
-        Task<FormulaValue> InvokeAsync(TimeZoneInfo timezoneInfo, FormulaType irContext, FormulaValue[] args, CancellationToken cancellationToken);
+        Task<FormulaValue> InvokeAsync(TimeZoneInfo timezoneInfo, FormulaType irContext, FormulaValue[] args, Action checkCancellation);
```
CarlosFigueiraMSFT, line 14: >checkCancellation [](http://example.com/codeflow?start=117&length=17)

Instead of updating this interface, can we move the functions to IAsyncTexlFunction5, so that at least we'll be closer to merging all of those into a single interface to be able to finally remove them?

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
```diff
@@ -319,7 +373,96 @@ public void Visit(TimeValue timeValue)
...
+                    {
+                        _writer.WriteStringValue(string.Empty);
+                        return;
+                    }
+
+                    //if (index > _maxDepth)
+                    //{
+                    //    IRContext irContext = IRContext.NotInSource(FormulaType.UntypedObject);
+                    //    ErrorValues.Add(new ErrorValue(irContext, new ExpressionError()
+                    //    {
```
CarlosFigueiraMSFT, line 393: Remove commented-out code

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
```diff
@@ -256,8 +298,14 @@ public void Visit(RecordValue recordValue)
                 {
                     _writer.WriteStartObject();
 
-                    foreach (NamedValue namedValue in recordValue.Fields)
+                    foreach (NamedValue namedValue in recordValue.Fields.OrderBy(f => f.Name, StringComparer.Ordinal))
                     {
+                        if (_checkCancellation(0))
+                        {
+                            _writer.WriteEndObject();
```
CarlosFigueiraMSFT, line 305: >_writer.WriteEndObject() [](http://example.com/codeflow?start=28&length=24)

Do we need to write the end object, since it will be ignored (as there is an error)?

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
```diff
@@ -287,6 +335,12 @@ public void Visit(TableValue tableValue)
 
                     foreach (DValue<RecordValue> row in tableValue.Rows)
                     {
+                        if (_checkCancellation(0))
+                        {
+                            _writer.WriteEndArray();
```
CarlosFigueiraMSFT, line 340: >_writer.WriteEndArray(); [](http://example.com/codeflow?start=28&length=24)

Do we need to write the end array, since it will be ignored (as there is an error)?

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4760,4 +4760,10 @@
...
+    <value>Maximum depth reached while traversing JSON payload.</value>    
+  </data>
+  <data name="ErrReachedMaxJsonLength" xml:space="preserve">
+    <value>Maximum length reached in JSON function.</value>
```
CarlosFigueiraMSFT, line 4767: >JSON [](http://example.com/codeflow?start=37&length=4)

Please add a comment with a `{Locked=JSON}` directive -since this is a function name we don't want it translated in case some language has a different acronym for that.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4760,4 +4760,10 @@
     <value>Unsupported untyped/JSON conversion type '{0}' in argument.</value>
     <comment>Error Message shown to user when a unsupported type is passed in type argument of AsType, IsType and ParseJSON functions.</comment>
   </data>
+  <data name="ErrReachedMaxJsonDepth" xml:space="preserve">
+    <value>Maximum depth reached while traversing JSON payload.</value>
```
CarlosFigueiraMSFT, line 4764: >Maximum depth reached while traversing JSON payload. [](http://example.com/codeflow?start=11&length=52)

Please add a comment with more details about this, to help translators avoid bugs that have happened in the past (it also helps us understand where it is used). Something like "Error message returned by the JSON function when a document that is too deeply nested is passed to it. The term JSON refers to the data format described in www.json.org".

Same comment for the next string.

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
```diff
@@ -129,17 +132,56 @@ private JsonFlags GetFlags()
 
             private class Utf8JsonWriterVisitor : IValueVisitor
             {
+                private const int _maxDepth = 20;
+                private const int _maxLength = 1048576; // 1 MB
```
MikeStall, line 136: >1 MB [](http://example.com/codeflow?start=59&length=4)

nit - you can use an expression here for more readability,  like 1024 * 1024.

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
```diff
@@ -129,17 +132,56 @@ private JsonFlags GetFlags()
...
                     _writer = writer;
                     _timeZoneInfo = timeZoneInfo;
                     _flattenValueTables = flattenValueTables;
+
+                    _checkCancellation = (int index) =>
+                    {
+                        checkCancellation();
+
+                        bool error = false;
+
```
MikeStall, line 156: We don't need to return bool. 

Just throw OperationCanceledException; 

Particularly because checkCancellation already throws , so there's no need to introduce a 2nd return convention (where cancel could either throw or return true)

MikeStall: it might be simpler to just


1. save the _checkCancellation directly as a field, rather than the field be a newly created action that closes over the parameter. 

2. have an instance method:

private void CheckCancel(int depth) 
{
  this._checkCancellation(); // this throws 
  // other checks, which also throw

}

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
MikeStall: nit - is this bytes? or characters?
(a .net char is 2 bytes)

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/JsonSerializeUOTests.cs
```diff
@@ -0,0 +1,181 @@
...
+            yield return (14, new TestUO(new string[0]), @"""[]""");
+            yield return (15, new TestUO(new[] { new[] { 1, 2 }, new[] { 3, 4 } }), @"""[[1,2],[3,4]]""");
+            yield return (16, new TestUO(new[] { new object[] { 1, 2 }, new object[] { true, "a", 7 } }), @"""[[1,2],[true,""""a"""",7]]""");
+            yield return (17, new TestUO(new { a = 10, b = -20m, c = "abc" }), @"""{""""a"""":10,""""b"""":-20,""""c"""":""""abc""""}""");
+            yield return (18, new TestUO(new { x = new { y = true } }), @"""{""""x"""":{""""y"""":true}}""");
+            yield return (19, new TestUO(new { x = new { y = new[] { 1 }, z = "a", t = new { } }, a = false }), @"""{""""a"""":false,""""x"""":{""""t"""":{},""""y"""":[1],""""z"""":""""a""""}}""");
+            yield return (20, new TestUO(123456789012345.6789012345678m), @"""123456789012345.6789012345678""");
+        }
+
+        public class TestUO : IUntypedObject
```
MikeStall, line 69: >IUntypedObject [](http://example.com/codeflow?start=30&length=14)

nit - there's now a base class that may simplify this UntypedObjectBase

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
MikeStall: agreed. Check should just throw, and so write end is a non-issue

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
```diff
@@ -319,7 +373,96 @@ public void Visit(TimeValue timeValue)
...
+                        _writer.WriteBooleanValue(untypedObject.GetBoolean());
+                    }
+                    else if (type is ExternalType externalType)
+                    {
+                        if (externalType.Kind == ExternalTypeKind.Array || externalType.Kind == ExternalTypeKind.ArrayAndObject)
+                        {
+                            _writer.WriteStartArray();
+
+                            for (var i = 0; i < untypedObject.GetArrayLength(); i++)
+                            {
```
MikeStall, line 425: >{ [](http://example.com/codeflow?start=28&length=1)

CheckCancel()?

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
LucGenetier: When reaching the max depth, we were returning a valid string.
I'll make it throw

LucGenetier:
✅ No public API change.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4760,4 +4760,12 @@
     <value>Unsupported untyped/JSON conversion type '{0}' in argument.</value>
     <comment>Error Message shown to user when a unsupported type is passed in type argument of AsType, IsType and ParseJSON functions.</comment>
   </data>
+  <data name="ErrReachedMaxJsonDepth" xml:space="preserve">
+    <value>Maximum depth reached while traversing {Locked=JSON} payload.</value>
```
CarlosFigueiraMSFT, line 4764: >{Locked=JSON} [](http://example.com/codeflow?start=50&length=13)

This doesn't need the locked directive (we aren't talking about the JSON function, but the JSON format, which may be translated in other languages).

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
```diff
@@ -16,24 +16,30 @@
...
     {
-        public Task<FormulaValue> InvokeAsync(TimeZoneInfo timezoneInfo, FormulaType type, FormulaValue[] args, CancellationToken cancellationToken)
+        public Task<FormulaValue> InvokeAsync(IServiceProvider runtimeServiceProvider, FormulaType type, FormulaValue[] args, CancellationToken cancellationToken)
         {
-            cancellationToken.ThrowIfCancellationRequested();
-            return Task.FromResult(new JsonProcessing(timezoneInfo, type, args, supportsLazyTypes).Process());
+            TimeZoneInfo timeZoneInfo = runtimeServiceProvider.GetService(typeof(TimeZoneInfo)) as TimeZoneInfo ?? throw new InvalidOperationException("TimeZoneInfo is required");
+            Canceller canceller = runtimeServiceProvider.GetService(typeof(Canceller)) as Canceller ?? new Canceller(() => cancellationToken.ThrowIfCancellationRequested());
+
+            return Task.FromResult(new JsonProcessing(timeZoneInfo, type, args, supportsLazyTypes).Process(canceller));
```
CarlosFigueiraMSFT, line 32: >canceller [](http://example.com/codeflow?start=107&length=9)

Why do we need this new indirection? Can we just pass the cancellation token instead of this wrapper?

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
```diff
@@ -61,10 +69,16 @@ internal FormulaValue Process()
...
+                Utf8JsonWriterVisitor jsonWriterVisitor = new Utf8JsonWriterVisitor(writer, _timeZoneInfo, flattenValueTables: flags.FlattenValueTables, canceller);
 
-                _arguments[0].Visit(jsonWriterVisitor);
-                writer.Flush();
+                try
+                {
+                    _arguments[0].Visit(jsonWriterVisitor);
+                    writer.Flush();
+                }
+                catch (InvalidOperationException)
```
CarlosFigueiraMSFT, line 79: >InvalidOperationException [](http://example.com/codeflow?start=23&length=25)

Can you add a comment saying why this exception is being ignored? Or we can move the error logic (lines 83-91) into the catch block, something like

    catch (InvalidOperationException) {
        if (writer.ErrorValues.Any()) {
            // logic to return error
        }

        // Unexpected error, rethrow
        throw;
    }

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
```diff
@@ -129,17 +143,58 @@ private JsonFlags GetFlags()
...
+                internal Utf8JsonWriterVisitor(Utf8JsonWriter writer, TimeZoneInfo timeZoneInfo, bool flattenValueTables, Canceller canceller)
                 {
                     _writer = writer;
                     _timeZoneInfo = timeZoneInfo;
                     _flattenValueTables = flattenValueTables;
+
+                    _canceller = canceller;
+                }
+
+                private void ThrowIfCancellationRequested(int index)
```
CarlosFigueiraMSFT, line 169: >ThrowIfCancellationRequested [](http://example.com/codeflow?start=29&length=28)

Suggestion: change the method name to something like "CheckLimitsAndCancellation" as this doesn't only check for cancellation

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
```diff
@@ -129,17 +143,58 @@ private JsonFlags GetFlags()
...
+                    {
+                        IRContext irContext = IRContext.NotInSource(FormulaType.UntypedObject);
+                        ErrorValues.Add(new ErrorValue(irContext, new ExpressionError()
+                        {
+                            ResourceKey = TexlStrings.ErrReachedMaxJsonDepth,
+                            Span = irContext.SourceContext,
+                            Kind = ErrorKind.InvalidArgument
+                        }));
+
+                        throw new InvalidOperationException("Maximum depth reached while traversing JSON payload.");
```
MikeStall, line 183: >Maximum depth [](http://example.com/codeflow?start=61&length=13)

nit- it's good to include the actual max depth/length in the error message. 
This will be more interesting once those numbers are adjustable.

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
```diff
@@ -319,7 +378,77 @@ public void Visit(TimeValue timeValue)
 
                 public void Visit(UntypedObjectValue untypedObjectValue)
                 {
-                    throw new ArgumentException($"Unable to serialize type {untypedObjectValue.GetType().FullName} to Json format.");
+                    Visit(untypedObjectValue.Impl);
+                }
+
+                private void Visit(IUntypedObject untypedObject, int index = 0)
```
MikeStall, line 384: >index [](http://example.com/codeflow?start=69&length=5)

depth?

MikeStall (dismissed):
:shipit:

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4760,4 +4760,12 @@
...
+    <comment>Error message returned by the JSON function when a document that is too deeply nested is passed to it. The term JSON refers to the data format described in www.json.org.</comment>
+  </data>
+  <data name="ErrReachedMaxJsonLength" xml:space="preserve">
+    <value>Maximum length reached in {Locked=JSON} function.</value>
```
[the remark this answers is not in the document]
LucGenetier: I don't understand what it could be used for if that's in comments

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
LucGenetier: Due to Governor class which is not accessible here

src/libraries/Microsoft.PowerFx.Json/Functions/JsonFunctionImpl.cs
LucGenetier: good idea!

src/strings/PowerFxResources.en-US.resx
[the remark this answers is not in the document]
CarlosFigueiraMSFT: The localization pipeline takes all the strings and comments and send them to the localization team. They send those to translators for multiple languages, and the only context they have is the comment we have. If the term JSON is translated in some language, then if we didn't have that Locked directive, then they may translate that - and we don't want it, as the JSON function name is not translated

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4760,4 +4760,12 @@
     <value>Unsupported untyped/JSON conversion type '{0}' in argument.</value>
     <comment>Error Message shown to user when a unsupported type is passed in type argument of AsType, IsType and ParseJSON functions.</comment>
   </data>
+  <data name="ErrReachedMaxJsonDepth" xml:space="preserve">
+    <value>Maximum depth reached while traversing JSON payload.</value>
+    <comment>Error message returned by the {Locked=JSON} function when a document that is too deeply nested is passed to it. The term JSON refers to the data format described in www.json.org.</comment>
```
CarlosFigueiraMSFT, line 4765: >{Locked=JSON} [](http://example.com/codeflow?start=43&length=13)

In this case we _don't_ want to lock the term JSON. It doesn't refer to the function, but to the format itself. If the term is translated in some language, we want the translators to be able to do that.

CarlosFigueiraMSFT:
> 

Those tests (except the one with the escaped "Hello") require PowerFxV1; all others don't. So you don't need to add the #setup for PFxV1 to this file - we still want all of those tests to run in Power Apps even with PFx1.0 disabled. Those 4 tests (and only those) should be moved to a new test file JSON_PowerFxV1.txt

---
Refers to: src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/JSON.txt:232 in 43b9343. [](commit_id = 43b9343069b20d63c6a37be47c7fa43f81b826f9, deletion_comment = False)

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2640, opened 2024-09-17
Add a warning to the CountRows function if a data source that caches its count is passed to it

CarlosFigueiraMSFT:
Some data sources, like Dataverse, will cache its count and not return the accurate result if a CountRows call is made to it (see note at https://learn.microsoft.com/power-platform/power-fx/reference/function-table-counts#description). This adds a warning to the node if this happens, to make it more explicit to the makers that this is the case.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalTabularDataSource.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalTabularDataSource.cs
@@ -11,6 +11,13 @@ namespace Microsoft.PowerFx.Core.Entities
     {
         TabularDataQueryOptions QueryOptions { get; }
 
+        /// <summary>
+        /// Some data sources (like Dataverse) may return a cached value for
+        /// the number of rows (calls to CountRows) instead of always retrieving
+        /// the latest count.
+        /// </summary>
+        bool HasCachedCountRows { get; }
+
         IReadOnlyList<string> GetKeyColumns();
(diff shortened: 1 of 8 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalTabularDataSource.cs
```diff
@@ -11,6 +11,13 @@ internal interface IExternalTabularDataSource : IExternalDataSource, IDisplayMap
     {
         TabularDataQueryOptions QueryOptions { get; }
 
+        /// <summary>
+        /// Some data sources (like Dataverse) may return a cached value for
+        /// the number of rows (calls to CountRows) instead of always retrieving
+        /// the latest count.
+        /// </summary>
+        bool HasCachedCountRows { get; }
```
anderson-joyle, line 19: >bool HasCachedCountRows { get; } [](http://example.com/codeflow?start=8&length=32)

Why not at `IExternalDataSource`?

CarlosFigueiraMSFT: CountRows can only be applied to tabular data sources. An API data source such as Office365Users, which has lots of operations, doesn't have the concept of a count.

anderson-joyle (approved):
:shipit:

PR #2642, opened 2024-09-18
Adds async way to get value from NamedValue

jas-valgotar:


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Values/NamedValue.cs
```diff
@@ -17,7 +18,17 @@ public class NamedValue
     {
         public string Name { get; }
 
-        public FormulaValue Value => _value ?? _getFormulaValue().ConfigureAwait(false).GetAwaiter().GetResult();
+        public FormulaValue Value => ValueAsync().GetAwaiter().GetResult();
+
+        public async Task<FormulaValue> ValueAsync()
```
MikeStall, line 23: Should this be a property? (instead of a method)?

MikeStall (approved):

src/libraries/Microsoft.PowerFx.Core/Public/Values/NamedValue.cs
jas-valgotar: C# don't support async props.
https://stackoverflow.com/questions/6602244/how-to-call-an-async-method-from-a-getter-or-setter#:~:text=There%20is%20no%20technical%20reason,be%20kicking%20off%20background%20operations.

LucGenetier:
✅ No public API change.

PR #2644, opened 2024-09-19
Prevent NullReferenceException found in telemetry

CarlosFigueiraMSFT:
We found a NRE in Power Apps telemetry that was happening in the Intellisense code on Power Fx. We don't know exactly the cause, but this change is a defense-in-depth that will help preventing that.


anderson-joyle (approved):
:shipit:

LucGenetier (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2650, opened 2024-09-24
Update the Trace function to return a Void type

CarlosFigueiraMSFT:
The Trace function doesn't return anything. For legacy reasons, it currently returns DType.Boolean (when it was created, we didn't have DType.Void). This PR updates it to return Void instead.

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests.Shared/TexlTests.cs
+++ b/src/tests/Microsoft.PowerFx.Core.Tests.Shared/TexlTests.cs
@@ -2703,7 +2703,7 @@ namespace Microsoft.PowerFx.Core.Tests
             {
                 var features = powerFxV1 ? Features.PowerFxV1 : Features.None;
                 var engine = new Engine(new PowerFxConfig(features));
-                var options = new ParserOptions() { NumberIsFloat = true, AllowsSideEffects = true };
+                var options = new ParserOptions() { AllowsSideEffects = true };
                 var result = engine.Check(expression, options);
 
                 var expectedType = powerFxV1 ? DType.Void : DType.Boolean;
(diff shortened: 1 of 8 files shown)


src/tests/Microsoft.PowerFx.Core.Tests.Shared/TexlTests.cs
```diff
@@ -2693,6 +2693,25 @@ public void TexlFunctionTypeSemanticsTable_Negative(string script, string expect
...
+        [InlineData("Trace(\"hello\")")]
+        [InlineData("Trace(\"hello\", TraceSeverity.Warning)")]
+        [InlineData("Trace(\"hello\", TraceSeverity.Warning, { a: 1 })")]
+        public void TexlFunctionTypeSemanticsTrace(string expression)
+        {
+            foreach (var powerFxV1 in new[] { false, true })
+            {
+                var features = powerFxV1 ? Features.PowerFxV1 : Features.None;
+                var engine = new Engine(new PowerFxConfig(features));
+                var options = new ParserOptions() { NumberIsFloat = true, AllowsSideEffects = true };
```
anderson-joyle, line 2706: >NumberIsFloat = true [](http://example.com/codeflow?start=52&length=20)

Do we need this parse option to run the test?

anderson-joyle (dismissed):
:shipit:

src/tests/Microsoft.PowerFx.Core.Tests.Shared/TexlTests.cs
CarlosFigueiraMSFT: Good point, it's not necessary. Removed.

LucGenetier:
✅ No public API change.

anderson-joyle (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2653, opened 2024-09-25
Clarify error and instructions when user uses behavior function in non-behavior UDF

rick-nguyen:
Add an informative error for behavior functions used in non-imperative UDF


src/strings/PowerFxResources.en-US.resx
```diff
@@ -3621,6 +3621,17 @@
     <value>https://go.microsoft.com/fwlink/?linkid=2132570</value>
     <comment>{Locked}</comment>
   </data>
+  <data name="ErrorResource_ErrBehaviorFunctionInDataUDF_ShortMessage" xml:space="preserve">
+    <value>Behavior function in a non-behavior user-defined function. Please wrap user-defined function with {} to declare a behavior UDF.</value>
+    <comment>Error Message.</comment>
+  </data>
+  <data name="ErrorResource_ErrBehaviorFunctionInDataUDF_LongMessage" xml:space="preserve">
+    <value>Behavior functions change the state of the app by changing values elsewhere in the app. 'Navigate', 'Patch', 'UpdateContext', and 'Collect' are common behavior functions. 'OnSelect', 'OnVisible', and other 'On …' properties are common behavior-based properties.</value>
```
CarlosFigueiraMSFT, line 3629: >state of the app by changing values elsewhere in the app [](http://example.com/codeflow?start=41&length=56)

This error message is very app-specific, but this message would appear for any hosts that uses UDFs. We should try to have a more neutral message here.

CarlosFigueiraMSFT, line 3629: >'Navigate', 'Patch', 'UpdateContext', and 'Collect' [](http://example.com/codeflow?start=99&length=51)

Similar to the previous comment, some of those functions are very Power Apps-specific. We should use examples of functions that are available in more hosts.

CarlosFigueiraMSFT, line 3629: >Navigate [](http://example.com/codeflow?start=100&length=8)

Please add a <comment> node to this resource, and add the directives {Locked=Navigate}{Locked=Patch}... to it. Those function names are not localized, and we don't want them translated by mistake. The comment should also provide additional context to the translation team about how this message is shown to users.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -3621,6 +3621,17 @@
     <value>https://go.microsoft.com/fwlink/?linkid=2132570</value>
     <comment>{Locked}</comment>
   </data>
+  <data name="ErrorResource_ErrBehaviorFunctionInDataUDF_ShortMessage" xml:space="preserve">
+    <value>Behavior function in a non-behavior user-defined function. Please wrap user-defined function with {} to declare a behavior UDF.</value>
```
CarlosFigueiraMSFT, line 3625: >UDF [](http://example.com/codeflow?start=134&length=3)

We're using an acronym that has not been defined. Consider updating this string to define it (e.g.: ...in a non-behavior user-defined function (UDF). Please wrap...)

CarlosFigueiraMSFT, line 3625: >wrap user-defined function with {} [](http://example.com/codeflow?start=77&length=34)

Suggestion: "Please wrap **the** user-defined function **body with curly braces ({...})** to declare..."

src/strings/PowerFxResources.en-US.resx
```diff
@@ -3621,6 +3621,18 @@
     <value>https://go.microsoft.com/fwlink/?linkid=2132570</value>
     <comment>{Locked}</comment>
   </data>
+  <data name="ErrorResource_ErrBehaviorFunctionInDataUDF_ShortMessage" xml:space="preserve">
+    <value>Behavior function in a non-behavior user-defined function (UDF). Please wrap the user-defined function body with curly braces ({...}) to declare a behavior UDF.</value>
+    <comment>Error Message.</comment>
+  </data>
+  <data name="ErrorResource_ErrBehaviorFunctionInDataUDF_LongMessage" xml:space="preserve">
+    <value>Behavior functions change the current session state. 'Clear', 'Collect', 'Patch', and 'Refresh' are common behavior functions. 'OnSelect', 'OnVisible', and other 'On …' properties are common behavior-based properties.</value>
```
CarlosFigueiraMSFT, line 3629: >'OnSelect', 'OnVisible', and other 'On …' properties are common behavior-based properties. [](http://example.com/codeflow?start=138&length=90)

There is no need to talk about behavior properties in this error message (and those are Power Apps specific, not generic to be in a Power Fx error).

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #2654, opened 2024-09-25
Fix mismatch return type error span for chained expression UDFs

rick-nguyen:
In chained expressions, the error for the bad return type would only highlight the semicolon instead of the last child node of the variadic op node.


src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/RecalcEngineTests.cs
```diff
@@ -678,6 +678,32 @@ public void ImperativeUserDefinedFunctionTest(string udfExpression, string expre
...
         }
 
+        // Behavior function in imperative udf
+        [InlineData(
+            "MismatchType():Number = { 1+3; Set(x, 123); };",
+            true,
+            "does not match the return type of the function body",
+            true)]
+
+        public void TestMismatchReturnType(string udfExpression, bool expectedError, string expectedMessage, bool allowSideEffects)
```
CarlosFigueiraMSFT, line 688: [minor] remove empty line

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/RecalcEngineTests.cs
```diff
@@ -678,6 +678,32 @@ public void ImperativeUserDefinedFunctionTest(string udfExpression, string expre
             }
         }
 
+        // Behavior function in imperative udf
+        [InlineData(
```
CarlosFigueiraMSFT, line 682: >InlineData [](http://example.com/codeflow?start=9&length=10)

Without the `[Theory]` attribute, this test won't run

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/RecalcEngineTests.cs
```diff
@@ -711,6 +711,34 @@ public void ImperativeUserDefinedFunctionTest(string udfExpression, string expre
             }
         }
 
+        [Theory]
+
+        // Behavior function in imperative udf
+        [InlineData(
```
CarlosFigueiraMSFT, line 717: >InlineData [](http://example.com/codeflow?start=9&length=10)

If there is only one case, then we can make it a [Fact] and inline the arguments.

We can also add some additional cases for things such as examples where the types are mismatched, but there is an implicit coercion that can make it work (e.g., `MyFunc(): Text = { 1; 2 };`)

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/RecalcEngineTests.cs
```diff
@@ -711,6 +711,39 @@ public void ImperativeUserDefinedFunctionTest(string udfExpression, string expre
...
+            true,
+            36,
+            41)]
+        [InlineData(
+            "MatchType():Text = { 4; 3 };",
+            false,
+            true,
+            0,
+            0)]
+        public void TestMismatchReturnType(string udfExpression, bool expectedError, bool allowSideEffects, int min, int lim)
```
CarlosFigueiraMSFT, line 728: >allowSideEffects [](http://example.com/codeflow?start=90&length=16)

For behavior UDFs we always need the allowSideEffects to be true, so there's no need to pass this along

PR #2656, opened 2024-09-26
[SQL CDP] Use logical names for tables instead of display names

LucGenetier:
We were returning display name instead of logical name for SQL relationships
Other CDP connectors ruse logical names and we need to be consistent


LucGenetier:
✅ No public API change.

PR #2657, opened 2024-09-26
Add new test to validate the DType always orders its fields

CarlosFigueiraMSFT:
Currently the tree that stores fields of a complex DType (records or arrays) stores those in an ordered way (ordinal comparison). There is some code that relies on that, so codifying it as a test to make it a "specification"


LucGenetier (approved):

Shpakh-MSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2658, opened 2024-09-26
Add test for ZD CDP connector, using enums

LucGenetier:
Validate that we can process option sets properly with CDP connectors


MikeStall (approved):

LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #2663, opened 2024-09-30
Make syntax localization util internal instead of private

lesaltzm:
Working on some cleanup of localization code, and this utility is duplicated in PA Client.


CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

Issue #2671, opened 2024-10-03
IsEmpty Function (and Possibly Other Table Functions) Does Not Support Untyped Objects


gekarapa:
The **IsEmpty** function, along with other table functions, does not support untyped objects, resulting in an error when such objects are passed as arguments.

PR #2672, opened 2024-10-03
IsEmpty supports UO

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2671.
Still need to verify other functions.


MikeStall (dismissed):

src/libraries/Microsoft.PowerFx.Core/Texl/BuiltinFunctionsCore.cs
```diff
@@ -254,6 +254,7 @@ internal class BuiltinFunctionsCore
         public static readonly TexlFunction Decimal_UO = _featureGateFunctions.Add(new DecimalFunction_UO());
         public static readonly TexlFunction Float = _featureGateFunctions.Add(new FloatFunction());
         public static readonly TexlFunction Float_UO = _featureGateFunctions.Add(new FloatFunction_UO());
+        public static readonly TexlFunction IsEmpty_UO = _featureGateFunctions.Add(new IsEmptyFunction_UO());
         public static readonly TexlFunction IsUTCToday = _featureGateFunctions.Add(new IsUTCTodayFunction());
```
CarlosFigueiraMSFT, line 258: This change is simple enough that we should just make it in Power Apps as well instead of diverging the platforms. Can you move it to the core _library instead?

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2676, opened 2024-10-07
Making TryGetFieldType virtual

anderson-joyle:


LucGenetier:
✅ No public API change.

MikeStall:
>             if (!TryGetFieldType(logical, out type))

This is already virtual. 

The derived class can control the display/logic pairs via the DNP is passes in. 


---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/Types/AggregateType.cs:112 in 2fa8d31. [](commit_id = 2fa8d31b57ce43d426718c046dc28bec3c54916d, deletion_comment = False)

PR #2691, opened 2024-10-10
Fix PlainText's handling of open tags

### LINKED ISSUE #2679: Incorrect handling of open tags on the PlainText function
Some examples:
- `PlainText("<div>One paragraph, <diva>not a paragraph</diva>, and the end</div>")`
    - It adds a newline on the `<diva>` tag, it should not
- `PlainText("This should <bra/> not be a newline")`
    - It adds a newline on the `<bra/>` tag, it should not

CarlosFigueiraMSFT:
Fixes #2679 . Currently the PlainText doesn't handle some open tags correctly - the function is supposed to convert things like &lt;br/&gt; to a new line or &lt;p&gt; to two new lines, but it is matching tags such as &lt;bra&gt; and &lt;para&gt; which should not be matched.


MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #2692, opened 2024-10-11
NF/UDF with Delegation warning is not passed on when NF/UDF is used

rick-nguyen:
1. UDF():DS = Filter(DS, Sqrt(Value) > 5));
2. use UDF() elsewhere, no delegation warning is shown

This has always been the case for named formulas as well, change for NF delegation warning flag is in other repo so test cannot be written to for that scenario.


LucGenetier:
✅ No public API change.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4026,6 +4026,60 @@
...
+  </data>
+    <data name="ErrorResource_SuggestRemoteExecutionHint_NF_ShortMessage" xml:space="preserve">
+    <value>Delegation warning. The named formula {0} is not delegatable so its usage in this formula might not work correctly on large data sets. </value>
+    <comment>Error Message.</comment>
+  </data>
+  <data name="ErrorResource_SuggestRemoteExecutionHint_NF_LongMessage" xml:space="preserve">
+    <value>The data source might not be able to process the formula and might return an incomplete data set. Your application might not return correct results or behave correctly if the data set is incomplete.</value>
+  </data>
+  <data name="ErrorResource_SuggestRemoteExecutionHint_NF_HowToFix_1" xml:space="preserve">
+    <value>Try simplifying the definition for the user-defined function.</value>
```
CarlosFigueiraMSFT, line 4038: >user-defined function [](http://example.com/codeflow?start=50&length=21)

Should it be 'named formula' here?

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4026,6 +4026,60 @@
   <data name="ErrorResource_SuggestRemoteExecutionHint_InOpInvalidColumn_Link_2_URL" xml:space="preserve">
     <value>https://go.microsoft.com/fwlink/?linkid=2132702</value>
     <comment>{Locked}</comment>
+  </data>
+    <data name="ErrorResource_SuggestRemoteExecutionHint_NF_ShortMessage" xml:space="preserve">
+    <value>Delegation warning. The named formula {0} is not delegatable so its usage in this formula might not work correctly on large data sets. </value>
```
CarlosFigueiraMSFT, line 4031: >delegatable [](http://example.com/codeflow?start=60&length=11)

We use the term 'delegable' instead of 'delegatable' elsewhere, we should be consistent. Personally I find 'delegatable' better, but I'm guessing the decision was made before to go with 'delegable' :-)

src/libraries/Microsoft.PowerFx.Core/Entities/External/IExternalNamedFormula.cs
```diff
@@ -13,6 +13,8 @@ internal interface IExternalNamedFormula : IExternalPageableSymbol, IExternalDel
 
         bool IsConstant { get; }
 
-        bool ContainsReferenceToView { get; }
+        bool ContainsReferenceToView { get; }
+
+        bool HasDelegationWarning { get; }
```
CarlosFigueiraMSFT, line 18: >HasDelegationWarning [](http://example.com/codeflow?start=13&length=20)

Tagged this PR as 'Breaking Internals'

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/RecalcEngineTests.cs
```diff
@@ -798,6 +798,43 @@ public void DelegableUDFTest()
 
             // Binding fails for recursive definitions and hence function is not added.
             Assert.False(recalcEngine.AddUserDefinedFunction("E():Void = { E(); };", CultureInfo.InvariantCulture, symbolTable: recalcEngine.EngineSymbols, allowSideEffects: true).IsSuccess);
+        }
+
+        [Fact]
+        public void TestInheritanceOfDelegationWarningsInUDFs()
```
CarlosFigueiraMSFT, line 804: >TestInheritanceOfDelegationWarningsInUDFs [](http://example.com/codeflow?start=20&length=41)

Can we have a test for the logic on Named Formulas?

rick-nguyen: Will add integrated test in PA client for NF scenario

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

PR #2695, opened 2024-10-14
Cleanup for some tests

CarlosFigueiraMSFT:
Removing duplication between pre-V1 and V1 tests.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/inScalar.txt
```diff
@@ -105,3 +105,27 @@ false
...
+>> Blank() in Blank()
+true
+
+>> Blank() exactin Blank()
+true
+
+>> Blank() in "hello"
+true
+
+>> Blank() exactin "hello"
```
anderson-joyle, line 130: >Blank() exactin "hello" [](http://example.com/codeflow?start=3&length=23)

Should `Blank()` always return true against `arg1` of any type? For instance: 
`Blank() in ParseJSON( ... ) // untyped`
`Blank() in Remove(...) // void type`

anderson-joyle (approved):
:shipit:

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/inScalar.txt
CarlosFigueiraMSFT: Not in all cases. We don't have the `in` operator working where the second argument is a record (e.g., `1 in {a:1}`), so even if we implemented `in` for untyped objects, there would be cases where it would fail (e.g., `1 in ParseJSON("{""a"":1}")`). Same with void values - the `in` operator (and all other operators) do not work with them.

PR #2699, opened 2024-10-16
Draft: Improve UDF handling in delegation strategy

rick-nguyen:
This PR adds a flag that allows us to maintain UDF context when visiting UDF call nodes in delegation strategy. This is needed to provide accurate warnings when using UDF inside filter function args.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
@@ -75,6 +75,8 @@ namespace Microsoft.PowerFx.Core.Functions
             return TryGetExternalDataSource(out dsInfo);
         }
 
+        public bool HasDelegationWarning => _binding?.ErrorContainer.GetErrors().Any(error => error.MessageKey.Contains("SuggestRemoteExecutionHint")) ?? false;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="UserDefinedFunction"/> class.
         /// </summary>
--- a/src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/RecalcEngineTests.cs
+++ b/src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/RecalcEngineTests.cs
@@ -800,6 +800,43 @@ namespace Microsoft.PowerFx.Tests
             Assert.False(recalcEngine.AddUserDefinedFunction("E():Void = { E(); };", CultureInfo.InvariantCulture, symbolTable: recalcEngine.EngineSymbols, allowSideEffects: true).IsSuccess);
         }
 
+        [Fact]
+        public void TestInheritanceOfDelegationWarningsInUDFs()
+        {
+            var symbolTable = new DelegatableSymbolTable();
+            var schema = DType.CreateTable(
+                new TypedName(DType.Number, new DName("Value")));
+            symbolTable.AddEntity(new TestDelegableDataSource(
+                "MyDataSource",
+                schema,
+                new TestDelegationMetadata(
+                        new DelegationCapability(DelegationCapability.Filter),
+                        schema,
+                        new FilterOpMetadata(
+                            schema,
+                            new Dictionary<DPath, DelegationCapability>(),
+                            new Dictionary<DPath, DelegationCapability>(),
+                            new DelegationCapability(DelegationCapability.GreaterThan),
+                            null)),
+                true));
+            symbolTable.AddType(new DName("MyDataSourceTableType"), FormulaType.Build(schema));
+            var config = new PowerFxConfig()
+            {
+                SymbolTable = symbolTable
+            };
+            var engine = new RecalcEngine(config);
+
+            var result = engine.AddUserDefinedFunction("NonDelegatableUDF():MyDataSourceTableType = Filter(MyDataSource, Sqrt(Value) > 5);", CultureInfo.InvariantCulture, symbolTable: engine.EngineSymbols, allowSideEffects: true);
+            Assert.True(result.IsSuccess);
+            var func = engine.Functions.WithName("NonDelegatableUDF").First() as UserDefinedFunction;
+            Assert.True(func.HasDelegationWarning);
+
+            result = engine.AddUserDefinedFunction("NonDelegatableUDF2():MyDataSourceTableType = NonDelegatableUDF();", CultureInfo.InvariantCulture, symbolTable: engine.EngineSymbols, allowSideEffects: true);
+            Assert.True(result.IsSuccess);
+            func = engine.Functions.WithName("NonDelegatableUDF2").First() as UserDefinedFunction;
+            Assert.True(func.HasDelegationWarning);
+        }
+
         // Binding to inner functions does not impact outer functions. 
         [Fact]
         public async Task FunctionInner()
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -2941,6 +2941,12 @@ namespace Microsoft.PowerFx.Core.Binding
                     if (lookupInfo.Data is IExternalNamedFormula formula)
                     {
                         isConstantNamedFormula = formula.IsConstant;
+
+                        // If the definition of the named formula has a delegation warning, every use should also inherit this warning
+                        if (formula.HasDelegationWarning)
+                        {
+                            _txb.ErrorContainer.EnsureError(DocumentErrorSeverity.Warning, node, TexlStrings.SuggestRemoteExecutionHint_NF, node.Ident.Name);
+                        }
                     }
                 }
                 else if (lookupInfo.Kind == BindKind.Data)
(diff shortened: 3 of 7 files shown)


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -945,12 +945,12 @@ public virtual bool SupportsPaging(CallNode callNode, TexlBinding binding)
 
         // Returns true if function is row scoped and supports delegation.
         // Needs to be overriden by functions (For example, IsBlank) which are not server delegatable themselves but can become one when scoped inside a delegatable function.
-        public virtual bool IsRowScopedServerDelegatable(CallNode callNode, TexlBinding binding, OperationCapabilityMetadata metadata)
+        public virtual bool IsRowScopedServerDelegatable(CallNode callNode, TexlBinding binding, OperationCapabilityMetadata metadata, bool nodeInheritsRowScope = false)
```
CarlosFigueiraMSFT, line 948: >= false [](http://example.com/codeflow?start=161&length=7)

Can we make this parameter required, to ensure that we pass the proper value everywhere?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -44,7 +45,9 @@ internal class UserDefinedFunction : TexlFunction, IExternalPageableSymbol, IExt
 
         public bool IsPageable => _binding.IsPageable(_binding.Top);
 
-        public bool IsDelegatable => _binding.IsDelegatable(_binding.Top);
+        public bool IsDelegatable => _binding.IsDelegatable(_binding.Top);
+
+        public bool SupportsRowScopedServerDelegation;
```
CarlosFigueiraMSFT, line 50: >public [](http://example.com/codeflow?start=8&length=6)

Can we make this private? We shouldn't have public fields in general

src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/DelegationValidationStrategy.cs
```diff
@@ -405,9 +407,16 @@ protected bool IsValidCallNodeInternal(CallNode node, TexlBinding binding, Opera
...
             }
 
-            if (callInfo?.Function != null && ((TexlFunction)callInfo.Function).IsRowScopedServerDelegatable(node, binding, metadata))
+            if (callInfo?.Function != null && ((TexlFunction)callInfo.Function).IsRowScopedServerDelegatable(node, binding, metadata, nodeInheritsRowScope: nodeInheritsRowScope))
             {
                 return true;
+            }
+
+            if (callInfo?.Function is UserDefinedFunction udf &&
+                ((FilterFunction)trackingFunction).IsValidDelegatableFilterPredicateNode(udf.Binding.Top, udf.Binding, metadata as FilterOpMetadata, nodeInheritsRowScope: isRowScoped))
```
CarlosFigueiraMSFT: >FilterFunction [](http://example.com/codeflow?start=18&length=14)

Can we guarantee that this is going to be a FilterFunction?

rick-nguyen: No, sorry that was an outdated commit

src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/DelegationValidationStrategy.cs
```diff
@@ -18,17 +20,17 @@ namespace Microsoft.PowerFx.Core.Functions.Delegation.DelegationStrategies
...
     internal interface ICallNodeDelegatableNodeValidationStrategy
     {
-        bool IsValidCallNode(CallNode node, TexlBinding binding, OperationCapabilityMetadata metadata, TexlFunction trackingFunction = null);
+        bool IsValidCallNode(CallNode node, TexlBinding binding, OperationCapabilityMetadata metadata, TexlFunction trackingFunction = null, bool nodeInheritsRowScope = false);
     }
 
     internal interface IDottedNameNodeDelegatableNodeValidationStrategy
     {
-        bool IsValidDottedNameNode(DottedNameNode node, TexlBinding binding, OperationCapabilityMetadata metadata, IOpDelegationStrategy opDelStrategy);
+        bool IsValidDottedNameNode(DottedNameNode node, TexlBinding binding, OperationCapabilityMetadata metadata, IOpDelegationStrategy opDelStrategy, bool nodeInheritsRowScope = false);
```
CarlosFigueiraMSFT, line 28: >bool nodeInheritsRowScope = false [](http://example.com/codeflow?start=152&length=33)

I would avoid declaring default values on interfaces, as it can lead to surprising behavior (discussed offline)

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/DateTime.cs
```diff
@@ -53,13 +53,13 @@ public ExtractDateTimeFunctionBase(string name, TexlStrings.StringGetter descrip
         {
         }
 
-        public override bool IsRowScopedServerDelegatable(CallNode callNode, TexlBinding binding, OperationCapabilityMetadata metadata)
+        public override bool IsRowScopedServerDelegatable(CallNode callNode, TexlBinding binding, OperationCapabilityMetadata metadata, bool nodeInheritsRowScope = false)
```
CarlosFigueiraMSFT, line 56: >IsRowScopedServerDelegatable [](http://example.com/codeflow?start=29&length=28)

This was already there, but do we need this override at all? If it only calls the base implementation, we can remove that.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Len.cs
```diff
@@ -31,9 +31,9 @@ public LenFunction()
 
         public override DelegationCapability FunctionDelegationCapability => DelegationCapability.Length;
 
-        public override bool IsRowScopedServerDelegatable(CallNode callNode, TexlBinding binding, OperationCapabilityMetadata metadata)
+        public override bool IsRowScopedServerDelegatable(CallNode callNode, TexlBinding binding, OperationCapabilityMetadata metadata, bool nodeInheritsRowScope = false)
```
CarlosFigueiraMSFT, line 34: >IsRowScopedServerDelegatable [](http://example.com/codeflow?start=29&length=28)

This was already there, but do we need this override at all? If it only calls the base implementation, we can remove that.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Trim.cs
```diff
@@ -38,13 +38,13 @@ public TrimEndsFunction()
 
         public override DelegationCapability FunctionDelegationCapability => DelegationCapability.Trim;
 
-        public override bool IsRowScopedServerDelegatable(CallNode callNode, TexlBinding binding, OperationCapabilityMetadata metadata)
+        public override bool IsRowScopedServerDelegatable(CallNode callNode, TexlBinding binding, OperationCapabilityMetadata metadata, bool nodeInheritsRowScope = false)
```
CarlosFigueiraMSFT, line 41: >IsRowScopedServerDelegatable [](http://example.com/codeflow?start=29&length=28)

This was already there, but do we need this override at all? If it only calls the base implementation, we can remove that.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/RecalcEngineTests.cs
```diff
@@ -835,6 +835,50 @@ public void TestInheritanceOfDelegationWarningsInUDFs()
             Assert.True(result.IsSuccess);
             func = engine.Functions.WithName("NonDelegatableUDF2").First() as UserDefinedFunction;
             Assert.True(func.HasDelegationWarning);
+        }
+
+        [Theory]
+        [InlineData(
+            "DelegatableUDF(Value: Number):Boolean = Value > 5;",
+            "DelegatableUDF2():MyDataSourceTableType = Filter(MyDataSource, DelegatableUDF(Value));",
```
CarlosFigueiraMSFT, line 843: >DelegatableUDF [](http://example.com/codeflow?start=76&length=14)

The fact that we can delegate this is surprising to me... do we expand the UDFs when translating it?

Shpakh-MSFT: We need to support it at least for non-row scope scenarios. For row-scope scenarios, it gets complicated in translation which we need to design it well.

Using UDF which is non-row scope shouldn't make the parent function call non-delegable. This works today at least which is good.

Shpakh-MSFT: For example,
UDF2(Name:Text) = Name;

This UDF2 could be row-scoped or non-row scoped depending in the context it is used.

Filter(Accounts, Name = UDF2(Name)) --> Row scoped
Filter(Accounts, Name = UDF2("")) --> Non-Row scoped

PR #2702, opened 2024-10-21
As rule binding could be null, we need to make sure to check that before checking for errors/warning

Shpakh-MSFT:
…fore checking for any error/warnings.


lesaltzm (approved):

LucGenetier:
✅ No public API change.

PR #2703, opened 2024-10-21
Change Lookup order in BinderUtils.TryGetConstantValue to check entitystore first

lesaltzm:
NameResolver.Lookup is a weirdly expensive operation when done from the middle of a running analysis pass. 
A few functions call TryGetConstantValue from CheckTypes, which is called during analysis to get functions result types.
In JSON it was possible to move that to CheckSemantics, but in Match, we want to avoid the expensive call to Lookup if we can, so we first look in the EntityStore to resolve enums. 

This doesn't meaningfully change the scenario for non PA hosts, as the entity store is a PA-only concept.


LucGenetier:
✅ No public API change.

Shpakh-MSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2705, opened 2024-10-22
Fix OptionSets in symbol tables

LucGenetier:
Make sure composed tables return all option sets


LucGenetier:
✅ No public API change.

PR #2709, opened 2024-10-22
Use Uri.EscapeDataString instead of HttpUtility.UrlEncode in order to…

LucGenetier:
… generate %20 instead of +


anderson-joyle (approved):
:shipit:

jas-valgotar (approved):

LucGenetier:
✅ No public API change.

PR #2712, opened 2024-10-23
QFEs on connectors

LucGenetier:
QFE1 Use Uri.EscapeDataString instead of HttpUtility.UrlEncode in order to generate %20 instead of +
QFE2 Add ExposeInternalParamsWithoutDefaultValue in ConnectorSettings to expose internal parameters without default value
QFE3 Make Json serializer less strict on JSON encoding


LucGenetier:
✅ No public API change.

jas-valgotar (approved):

PR #2715, opened 2024-10-23
Added a new property on IExternalRule interface.

Shpakh-MSFT:
As PA is moving away from rules with null binding, this PR introduces a new property called HasValidBinding which needs to be used to check the binding validity in future rather than Binding property itself. Eventually we should remove the Binding property from rule to abstract that. This change is first step towards that direction.


lesaltzm (approved):

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
🚢

PR #2719, opened 2024-10-24
Fix formatting and structural printing for type literals

adithyaselv:
Formatting , structural printing and cloning were broken for type literals. This PR fixes these issues and adds tests to ensure the changes work as expected.

Formatting of user definitions is still not perfect and has a bug where it would fail if we had definitions of same kind with same name.  That is not in the scope of this PR and shall be fixed as a follow up

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests.Shared/FormatterTests.cs
+++ b/src/tests/Microsoft.PowerFx.Core.Tests.Shared/FormatterTests.cs
@@ -27,16 +27,20 @@ namespace Microsoft.PowerFx.Tests
         [InlineData(
             "$\"Hello {5}\"",
             "$\"#$string$#{#$decimal$#}\"")]
+        [InlineData(
+            "ParseJSON(\"[{ \"\"Age\"\": 5}]\", Type([{Age: Number}]))",
+            "ParseJSON(#$string$#, Type([ { #$fieldname$#:#$firstname$# } ]))")]
         public void TestStucturalPrint(string script, string expected)
         {
             var result = ParseScript(
                 script,
-                flags: Flags.EnableExpressionChaining);
+                flags: Flags.EnableExpressionChaining,
+                features: Features.PowerFxV1);
 
             Assert.Equal(expected, StructuralPrint.Print(result.Root));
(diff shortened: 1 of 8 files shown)


src/tests/Microsoft.PowerFx.Core.Tests.Shared/FormatterTests.cs
```diff
@@ -256,6 +260,8 @@ public void TestMinificationWithCommentRemovalInTextFirst(string script, string
         [InlineData("Set(str, $\"{{}}\")", "Set(\n    str,\n    $\"{{}}\"\n)")]
         [InlineData("Set(additionText, $\"The sum of 1 and 3 is {{{1 + 3}}})\")", "Set(\n    additionText,\n    $\"The sum of 1 and 3 is {{{1 + 3}}})\"\n)")]
         [InlineData("$\"This is {{\"Another\"}} interpolated {{string}}\"", "$\"This is {{\"Another\"}} interpolated {{string}}\"")]
+        [InlineData("ParseJSON(\"[]\", Type([{Age: Number}]))", "ParseJSON(\n    \"[]\",\n    Type([{Age: Number}])\n)")]
```
CarlosFigueiraMSFT, line 263: >Type([{Age: Number}]) [](http://example.com/codeflow?start=94&length=21)

Should the type declaration be pretty-printed as well? Something like `Type([{ Name: Text, Age: Number, Address: { Street: Text, City: Text, State: Text, Zip: Text }, PhoneNumber: [{ Value: Number }]}])` can get really long, so it may be good to break it up

adithyaselv: it pretty prints if its long (more than one field) it was getting too long . thats why i added the next test

LucGenetier:
Public API changes: <br />❌CP0012 M:Microsoft.PowerFx.Syntax.TexlFunctionalVisitor`2.Visit(Microsoft.PowerFx.Syntax.TypeLiteralNode,`1)

CarlosFigueiraMSFT (approved):
:shipit:

PR #2720, opened 2024-10-28
Fixes in ComposedReadOnlySymbolTable

LucGenetier:
- OptionSets should only return first OS by name (first element wins)
- TryGetVariable not working as expected


MikeStall (approved):
:shipit:

PR #2721, opened 2024-10-28
Error when '=' syntax is used to define UDT

adithyaselv:
Today we allow something like `T = Type(Number)` which is incorrect and should not be allowed . this PR adds error to that scenario

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
@@ -385,6 +385,12 @@ namespace Microsoft.PowerFx.Core.Parser
                         definitionBeforeTrivia.Add(ParseTrivia());
                         var result = ParseExpr(Precedence.None);
 
+                        if (result is TypeLiteralNode typeLiteralNode)
+                        {
+                            CreateError(result.Token, TexlStrings.ErrUserDefinedTypeIncorrectSyntax);
+                            continue;
+                        }
+
                         namedFormulas.Add(new NamedFormula(thisIdentifier.As<IdentToken>(), new Formula(result.GetCompleteSpan().GetFragment(script), result), _startingIndex, attribute));
                         userDefinitionSourceInfos.Add(new UserDefinitionSourceInfo(index++, UserDefinitionType.NamedFormula, thisIdentifier.As<IdentToken>(), declaration, new SourceList(definitionBeforeTrivia), GetExtraTriviaSourceList()));
                         definitionBeforeTrivia = new List<ITexlSource>();
--- a/src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
@@ -362,6 +362,8 @@ namespace Microsoft.PowerFx.Core.Parser
                 }
                 else if (_curs.TidCur == TokKind.Equ)
                 {
+                    var equalToken = _curs.TokCur;
+                    
                     var declaration = script.Substring(declarationStart, _curs.TokCur.Span.Min - declarationStart);
                     _curs.TokMove();
                     definitionBeforeTrivia.Add(ParseTrivia());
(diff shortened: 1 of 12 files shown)


src/strings/PowerFxResources.en-US.resx
```diff
@@ -4846,4 +4846,11 @@
     <value>The 'User-defined types' experimental feature is not enabled.</value>
     <comment>Error message returned when user uses User-defined types with the User-defined types feature flag turned off.</comment>
   </data>
+  <data name="ErrUserDefinedTypeIncorrectSyntax" xml:space="preserve">
+    <value>Incorrect syntax `=` used to define User-defined type.</value>
```
CarlosFigueiraMSFT: >`=` [](http://example.com/codeflow?start=28&length=3)

Suggestion: use normal single quotes, not back ticks (consistent with other places in this file)

CarlosFigueiraMSFT: >User-defined type [](http://example.com/codeflow?start=47&length=17)

Doesn't need to be capitalized

CarlosFigueiraMSFT: >value [](http://example.com/codeflow?start=5&length=5)

If we know the problem and how it can be fixed, we should  tell the customer that, something like "Incorrect syntax: '=' cannot be used to declare user-defined types. Use ':=' instead."

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ParseTests.cs
```diff
@@ -1014,5 +1014,17 @@ public void TestUDFReturnsRecord(string script, bool expectErrors, bool isImpera
...
+        [Theory]
+        [InlineData("A = Type(Number);", 0)]
+        [InlineData("A = \"hello\";B = Type([Boolean]); C = 5;", 2)]
+        public void TestTypeLiteralInNamedFormula(string script, int namedFormulaCount)
+        {
+            var parserOptions = new ParserOptions();
+            var parseResult = UserDefinitions.Parse(script, parserOptions, Features.PowerFxV1);
+            Assert.True(parseResult.HasErrors);
+            Assert.Equal(namedFormulaCount, parseResult.NamedFormulas.Count());
+            Assert.Contains(parseResult.Errors, e => e.ShortMessage == StringResources.Get(TexlStrings.ErrUserDefinedTypeIncorrectSyntax));
```
anderson-joyle, line 1027: >Assert.Contains(parseResult.Errors, e => e.ShortMessage == StringResources.Get(TexlStrings.ErrUserDefinedTypeIncorrectSyntax)); [](http://example.com/codeflow?start=12&length=127)

Can compare by `ResourceKey` only?

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -385,6 +387,12 @@ private ParseUserDefinitionResult ParseUDFsAndNamedFormulas(string script, Parse
                         definitionBeforeTrivia.Add(ParseTrivia());
                         var result = ParseExpr(Precedence.None);
 
+                        if (result is TypeLiteralNode _)
+                        {
+                            CreateError(equalToken, TexlStrings.ErrUserDefinedTypeIncorrectSyntax);
```
anderson-joyle, line 392: >equalToken [](http://example.com/codeflow?start=40&length=10)

Can we simply call `_curs.TokCur` and remove the variable declaration at ln 365?

adithyaselv: we can't, if we want the error highlighting at equal token. `_curs.TokCur` will be different at this point as we do more parsing

anderson-joyle (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
anderson-joyle: Got it.

PR #2723, opened 2024-10-29
Fixes enum store when SymbolTable is replaced while setting up config.

jas-valgotar:
This PR fixes the issue where inbuilt enums are no longer part of the config if the config's symbol table is replaced after initialization.

After initializing the config if SymbolTable was replaced, we were loosing the enums attached to the previous default symbol table. This PR will again attach the enums if SymbolTable is replaced.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
@@ -23,13 +23,31 @@ namespace Microsoft.PowerFx
         internal static readonly int DefaultMaxCallDepth = 20;
         internal static readonly int DefaultMaximumExpressionLength = 1000;
 
-        /// <summary>
-        /// Global symbols. Additional symbols beyond default function set and primitive types. 
-        /// </summary>
-        public SymbolTable SymbolTable { get; set; } = new SymbolTable
+        private SymbolTable _symbolTable = new SymbolTable
         {
             DebugName = "DefaultConfig"
         };
+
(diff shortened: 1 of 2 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -23,13 +23,31 @@ public sealed class PowerFxConfig
...
-        };
+            get => _symbolTable;
+            set
+            {
+                if (value == null)
+                {
+                    value = new SymbolTable();
+                }
+
+                value.EnumStoreBuilder = BuiltInEnumStoreBuilder;
```
MikeStall, line 47: >value.EnumStoreBuilder  [](http://example.com/codeflow?start=16&length=23)

mutating the incoming symbol table is concerning

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/RecalcEngineTests.cs
```diff
@@ -400,13 +400,21 @@ public void UserDefinitionOnUpdateTest()
         }
 
         [Fact]
-        public void BasicEval()
-        {
-            var engine = new RecalcEngine();
-            engine.UpdateVariable("M", 10.0);
-            engine.UpdateVariable("M2", -4);
-            var result = engine.Eval("M + Abs(M2)");
```
anderson-joyle, line 408: >"M + Abs(M2)" [](http://example.com/codeflow?start=37&length=13)

Has this test become obsolete?

PR #2725, opened 2024-10-31
Added new method on IExternalRule to help with invariantexpression ch…

Shpakh-MSFT:
On an effort to remove null rule binding, we need the ability to have an option to apply null binding before doing the check if binding is null. This change adds a new method on IExternalRule to help with that.


lesaltzm (approved):
.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

lesaltzm (approved):

LucGenetier:
✅ No public API change.

PR #2728, opened 2024-11-04
Join function

anderson-joyle:
`Join(t1, t2, LeftRecord.id = RightRecord.refid) //JoinType.Inner by default`  
`Join(t1, t2, LeftRecord.id = RightRecord.refid, JoinType.Inner) // JoinType.Inner / JoinType.Left / JoinType.Right / JoinType.Full`  
`Join(t1, t2, LeftRecord.id = RightRecord.refid, JoinType.Inner, LeftRecord.id As NewId, ...) // Renaming columns to avoid name collision`


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Types/Enums/BuiltInEnums.cs
```diff
@@ -199,5 +199,16 @@ private static Dictionary<string, object> TraceSeverityDictionary()
                 { "IgnoreUnsupportedTypes", TraceFunction.IgnoreUnsupportedTypesEnumValue },
             },
             canConcatenateStronglyTyped: true);
+
+        public static readonly EnumSymbol JoinTypeEnum = new EnumSymbol(
+            new DName(LanguageConstants.JoinTypeName),
+            DType.String,
+            new Dictionary<string, object>()
+            {
+                { "Inner", "inner" },
```
MikeStall, line 208: nit - What is the object (value) here? Can we build this around a strongly typed C# enum?

LucGenetier:
✅ No public API change.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4822,6 +4822,42 @@
   <data name="AboutClear_source" xml:space="preserve">
     <value>The source to clear.</value>
   </data>
+  <data name="AboutJoin" xml:space="preserve">
+    <value>Retrieves data that spans across multiple tables with relationships.</value>
+    <comment>Description of 'Clear' function.</comment>
```
CarlosFigueiraMSFT, line 4827: Stale comment

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4853,4 +4889,7 @@
       Some examples for valid named type declarations - "Point := Type({x: Number, y: Number});" , "T1 := Type(Number);" , "T2 := Type([Boolean]);".
     </comment>
   </data>
+  <data name="ErrJoinCantUnion" xml:space="preserve">
+    <value>Data sources can't union. Possibly due to duplicated columns.</value>
```
CarlosFigueiraMSFT, line 4893: This error message is too "dry". We should try to give some possible solution to the customer.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/Join.txt
```diff
@@ -0,0 +1,81 @@
+﻿#SETUP: PowerFxV1CompatibilityRules,StronglyTypedBuiltinEnums
```
CarlosFigueiraMSFT, line 1: Do we need the 'StronglyTypedBuiltinEnums' setup for this test? If not, we should remove it.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -168,6 +168,11 @@ internal abstract class TexlFunction : IFunction
         /// </summary>
         public virtual bool CanSuggestInputColumns => false;
 
+        /// <summary>
+        /// Identifies which functions args to use to compose the scope type for subsequent lambda args.
```
CarlosFigueiraMSFT, line 172: "which functions args" - does it mean that the 1st argument is a scope-defining argument? Or that the second argument (using a 0-based index) is a scope-defining argument? Or something else?

src/libraries/Microsoft.PowerFx.Core/Functions/FunctionScopeInfo.cs
```diff
@@ -85,6 +85,20 @@ public FunctionScopeInfo(
             _function = function;
             AppliesToArgument = appliesToArgument ?? (i => i > 0);
             CanBeCreatedByRecord = canBeCreatedByRecord;
+        }
+
+        /// <summary>
+        /// Allows to type check multiple scopes.
+        /// </summary>
+        /// <param name="features"></param>
```
CarlosFigueiraMSFT, line 93: Comments?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Join.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        public JoinFunction()
+            : base("Join", TexlStrings.AboutJoin, FunctionCategories.Table, DType.EmptyTable, 0x8, 3, 4, DType.EmptyTable, DType.EmptyTable)
+        {
+            ScopeInfo = new FunctionJoinScopeInfo(this);
+        }
+
+        public override IEnumerable<TexlStrings.StringGetter[]> GetSignatures()
+        {
+            yield return new[] { TexlStrings.JoinArg1, TexlStrings.JoinArg2 };
```
CarlosFigueiraMSFT, line 30: If the minimum arity is 3 (from line 23), we should not suggest a signature with 2 elements only

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/Join.txt
```diff
@@ -0,0 +1,81 @@
...
+  Table({Id:"AAA111",Name:"Contoso",Country:"USA"},{Id:"BBB222",Name:"Fabrikam",Country:"CAN"},{Id:"EEE555",Name:"ACME INC.",Country:"MEX"},{Id:"FFF666",Name:"Dunder Mifflin",Country:"USA"}),
+  Table({SupplierId:"AAA111", Fruit:"Grapes", Price:220, Purchase:Date(2015,10,1), RowId:1},
+  {SupplierId:"BBB222", Fruit:"Lemons", Price:31, Purchase:Date(2015,10,1), RowId:2},
+  {SupplierId:"AAA111", Fruit:"Lemons", Price:29, Purchase:Date(2015,10,2), RowId:3},
+  {SupplierId:"AAA111", Fruit:"Grapes", Price:210, Purchase:Date(2015,10,2), RowId:4},
+  {SupplierId:"BBB222", Fruit:"Lemons", Price:30, Purchase:Date(2015,10,3), RowId:5},
+  {SupplierId:"AAA111", Fruit:"Bananas", Price:12, Purchase:Date(2015,10,3), RowId:6},
+  {SupplierId:"CCC333", Fruit:"Apples", Price:15, Purchase:Date(2015,12,3), RowId:7},
+  {SupplierId:"DDD444", Fruit:"Mangos", Price:30, Purchase:Date(2016,12,3), RowId:8}),
+  LeftRecord.Id = RightRecord.SupplierId, JoinType.Inner)
```
MikeStall, line 13: >JoinType.Inner [](http://example.com/codeflow?start=42&length=14)

Does this need to be a constant? 
(it would be ideal to restrict it to that - we can always lift the restriction later) 

Ie, fail if
If(true, JoinType.Inner, JoinType.Left)

gregli-msft: Starting out as a compile time constant is probably a good idea - it may help with delegation.

Related, JoinType should be a strongly typed enum that can't be combined.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Join.cs
```diff
@@ -0,0 +1,76 @@
...
+using Microsoft.PowerFx.Core.Errors;
+using Microsoft.PowerFx.Core.Functions;
+using Microsoft.PowerFx.Core.Localization;
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.Types.Enums;
+using Microsoft.PowerFx.Syntax;
+
+namespace Microsoft.PowerFx.Core.Texl.Builtins
+{
+    internal class JoinFunction : FilterFunctionBase
```
gregli-msft, line 16: Please don't merge this PR until we have the XR meeting.  It is fine to take review feedback now.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/Join.txt
anderson-joyle: We need it.

CarlosFigueiraMSFT: For all tests, or only for a few of them? If the latter, we should split it out into different test files.

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
anderson-joyle: This property defines how many argument, starting from 0, will be use to build the scope type.
`Join` function has `ScopeArgs => 2`, which means args `arg0` and `arg1` will be used to compose the scope type.

src/strings/PowerFxResources.en-US.resx
anderson-joyle: I changed the error message to include the duplicated columns.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/Join.txt
```diff
@@ -0,0 +1,189 @@
+﻿#SETUP: PowerFxV1CompatibilityRules,StronglyTypedBuiltinEnums, SupportColumnNamesAsIdentifiers
+
+// Inner
```
MikeStall, line 3: can we move to MutationScripts? Might be cleaner since it lets us Set.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -519,6 +519,204 @@ public static async ValueTask<FormulaValue> FilterTable(EvalVisitor runner, Eval
...
+
+        // Join(t1, t2, LeftRecord.Id = RightRecord.RefId, JoinType.Inner)
+        public static async ValueTask<FormulaValue> JoinTables(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
+        {
+            var leftTable = (TableValue)args[0];
+            var rightTable = (TableValue)args[1];
+            var predicate = (LambdaFormulaValue)args[2];
+            var joinType = (OptionSetValue)args[3];
+            var leftRenaming = (RecordValue)args[4];
+            var rightRenaming = (RecordValue)args[5];
```
MikeStall, line 532: comment above only shows 4 args; what do these last 2 mean?

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/Join.txt
```diff
@@ -0,0 +1,189 @@
...
+>> Join(
```
MikeStall, line 4: >>> [](http://example.com/codeflow?start=0&length=2)

we should have some tests passing blank, error arguments. To verify those are being handled.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Join.cs
```diff
@@ -0,0 +1,244 @@
...
+            var sourceNameProvider = new Dictionary<DName, DName>();
+
+            for (int i = 0; i < carg; i++)
+            {
+                var arg = node.Args.Children[i];
+
+                // Left
+                if (i == 0)
+                {
+                    sourceNameProvider[arg is AsNode asNode ? asNode.Left.AsDottedName().Left.AsFirstName().Ident.Name : FunctionJoinScopeInfo.LeftRecord] = FunctionJoinScopeInfo.LeftRecord;
```
MikeStall, line 174: >arg is AsNode asNode ? asNode.Left.AsDottedName().Left.AsFirstName().Ident.Name : FunctionJoinScopeInfo.LeftRecord [](http://example.com/codeflow?start=39&length=114)

nit - for readability, can we hoist this to a separate expression so that it's easier to read? 

ie, prefer a few shorter lines (with intermiedate assignments) over 1 larger line.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Join.cs
```diff
@@ -0,0 +1,244 @@
...
+
+                // Right
+                if (i == 1)
+                {
+                    sourceNameProvider[arg is AsNode asNode ? asNode.Left.AsDottedName().Left.AsFirstName().Ident.Name : FunctionJoinScopeInfo.RightRecord] = FunctionJoinScopeInfo.RightRecord;
+                }
+
+                if (i > 3 && IsLambdaParam(arg, i))
+                {
+                    var asNode = (AsNode)arg;
```
MikeStall, line 185: >(AsNode) [](http://example.com/codeflow?start=33&length=8)

How do we know this cast will succeed?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Join.cs
```diff
@@ -0,0 +1,244 @@
...
+                    var dName = sourceNameProvider[asNode.Left.AsDottedName().Left.AsFirstName().Ident.Name];
+
+                    if (!recordTypesMap.TryGetValue(dName, out _))
+                    {
+                        recordTypesMap[dName] = RecordType.Empty();
```
MikeStall, line 190: >[dName] [](http://example.com/codeflow?start=38&length=7)

super nit - [] will overwrite. If there should not collide, use Add()

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
MikeStall: Is this right:
Source may have arbitrary number of args. 
And then IR transform will effectively transform it and "group" them into 2 args (leftRenaming, RightRenaming) ?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -519,6 +519,204 @@ public static async ValueTask<FormulaValue> FilterTable(EvalVisitor runner, Eval
...
+
+        private static async Task<DValue<RecordValue>> LazyCheckPredicateAsync(
+           EvalVisitor runner,
+           EvalVisitorContext context,
+           DValue<RecordValue> leftRow,
+           DValue<RecordValue> rightRow,
+           LambdaFormulaValue predicate)
+        {
+            var scopeValue = FormulaValue.NewRecordFromFields(
+                new NamedValue("LeftRecord", leftRow.Value),
```
MikeStall, line 662: >LeftRecord [](http://example.com/codeflow?start=32&length=10)

nit - can we share this identifer with Core so that's the literal is only used once?

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/Join.txt
MikeStall: also - a case creating conflicts via 'As' 

..., F1 As R1, F2 As R1) // force conflict 

..., F1 As F2, F2) // implicit conflict.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Join.cs
anderson-joyle: The combination of `i > 3` and `IsLambdaParam(arg, i)` check.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Environment/PowerFxConfigExtensions.cs
```diff
@@ -80,6 +80,12 @@ public static void EnableRegExFunctions(this PowerFxConfig config, TimeSpan regE
             }
         }
 
+        [Obsolete("Join is still in preview.")]
```
MikeStall, line 83: > [Obsolete("Join is still in preview.")] [](http://example.com/codeflow?start=7&length=40)

Greg & I chatted and are good with this. 
It will let us test in playground and DV.

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
```diff
@@ -66,6 +66,11 @@ private DName ValidateName(string name)
             return new DName(name);
         }
 
+        public SymbolTable()
+        {
+            EnumStoreBuilder = new EnumStoreBuilder();
+        }
```
MikeStall, line 72: @jas-valgotar - fyi

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/Join.txt
```diff
@@ -0,0 +1,33 @@
+﻿#SETUP: PowerFxV1CompatibilityRules,StronglyTypedBuiltinEnums, SupportColumnNamesAsIdentifiers
+
+// Blank args
+>> Join(If(false,Table({a:1})),If(false,Table({b:1})),true)
+Blank()
+
+// Invalid arguments
+>> Join(1/0, Table({SupplierId:"AAA111"}), LeftRecord.Id = RightRecord.SupplierId, JoinType.Full)
```
MikeStall, line 8: >/ [](http://example.com/codeflow?start=9&length=1)

This is a compiler-time error. 

We also want to test runtime errors.  Ie:
If(1/0, Table({a:1})

src/libraries/Microsoft.PowerFx.Interpreter/Functions/IAsyncTexlFunctionJoin.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Functions;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Core.Functions
+{
+    [Obsolete("This interface is temporary and its meant only to support Join function. It will be soon removed.")]
+    internal interface IAsyncTexlFunctionJoin
```
MikeStall, line 14: >IAsyncTexlFunctionJoin [](http://example.com/codeflow?start=23&length=22)

?

anderson-joyle: Internal discussion.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -1342,5 +1550,20 @@ public static NamedLambda[] Parse(FormulaValue[] args)
                 return l.ToArray();
             }
         }
+    }
+
+#pragma warning disable CS0618 // Type or member is obsolete
+    internal class JoinImpl : JoinFunction, IAsyncTexlFunctionJoin
+#pragma warning restore CS0618 // Type or member is obsolete
+    {
+        public async Task<FormulaValue> InvokeAsync(FormulaType irContext, FormulaValue[] args, CancellationToken cancellationToken)
```
MikeStall, line 1559: >InvokeAsync [](http://example.com/codeflow?start=40&length=11)

Why does this have 2 Invoke impls?

anderson-joyle: This is wrong. Thanks for catching it.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT:
>                 for (var i = 1; i < carg; i++)

Suggestion: start from ScopeArgs instead of 1; we don't need to do any of the logic inside this loop for any of the scope arguments, since they have already been visited

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs:4623 in a438704. [](commit_id = a438704adee7a6e3f87e47859c717dbe3fa5f0fa, deletion_comment = False)

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/MutationScripts/Join.txt
```diff
@@ -0,0 +1,55 @@
...
+  {SupplierId:"AAA111", Fruit:"Grapes", Price:210, Purchase:Date(2015,10,2), RowId:4},
+  {SupplierId:"BBB222", Fruit:"Lemons", Price:30, Purchase:Date(2015,10,3), RowId:5},
+  {SupplierId:"AAA111", Fruit:"Bananas", Price:12, Purchase:Date(2015,10,3), RowId:6},
+  {SupplierId:"CCC333", Fruit:"Apples", Price:15, Purchase:Date(2015,12,3), RowId:7},
+  {SupplierId:"DDD444", Fruit:"Mangos", Price:30, Purchase:Date(2016,12,3), RowId:8}))
+Table({Fruit:"Grapes",Price:220,Purchase:Date(2015,10,1),RowId:1,SupplierId:"AAA111"},{Fruit:"Lemons",Price:31,Purchase:Date(2015,10,1),RowId:2,SupplierId:"BBB222"},{Fruit:"Lemons",Price:29,Purchase:Date(2015,10,2),RowId:3,SupplierId:"AAA111"},{Fruit:"Grapes",Price:210,Purchase:Date(2015,10,2),RowId:4,SupplierId:"AAA111"},{Fruit:"Lemons",Price:30,Purchase:Date(2015,10,3),RowId:5,SupplierId:"BBB222"},{Fruit:"Bananas",Price:12,Purchase:Date(2015,10,3),RowId:6,SupplierId:"AAA111"},{Fruit:"Apples",Price:15,Purchase:Date(2015,12,3),RowId:7,SupplierId:"CCC333"},{Fruit:"Mangos",Price:30,Purchase:Date(2016,12,3),RowId:8,SupplierId:"DDD444"})
+
+// Inner
+>> Join(t1, t2, LeftRecord.Id = RightRecord.SupplierId, JoinType.Inner)
```
CarlosFigueiraMSFT, line 22: >t1, t2 [](http://example.com/codeflow?start=8&length=6)

We should have tests for renaming the left/right scopes:

    Join(
        t1 As MyLeft,
        t2 As MyRight,
        MyLeft.Id = MyRight.SupplierId,
        JoinType.Inner,
        MyRight.Fruit As Fruit)

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Join.cs
```diff
@@ -0,0 +1,237 @@
...
+            // JoinType argument is present?
+            if (argTypes.Count() > 3 &&
+                (!BuiltInEnums.JoinTypeEnum.FormulaType._type.Accepts(argTypes[3], exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: context.Features.PowerFxV1CompatibilityRules) ||
+                args[3] is not DottedNameNode))
+            {
+                errors.EnsureError(DocumentErrorSeverity.Severe, args[3], TexlStrings.ErrJoinNotPlainJoinTypeEnum);
+                valid = false;
+            }
+
+            ScopeInfo.GetScopeIdent(args, out var scopeIdent);
```
CarlosFigueiraMSFT, line 72: >ScopeInfo.GetScopeIdent(args, out var scopeIdent); [](http://example.com/codeflow?start=12&length=50)

Do we need to check the result?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Join.cs
```diff
@@ -0,0 +1,237 @@
...
+                    {
+                        // !!TODO shouldn't the binder have caught this?
+                        if (!rightTable.TryGetType(dottedNameNode.Right.Name, out var rightType))
+                        {
+                            errors.EnsureError(DocumentErrorSeverity.Severe, args[0], TexlStrings.ErrColDNE_Name, dottedNameNode);
+                            valid = false;
+                            break;
+                        }
+
+                        returnType = returnType.Add(ref fError, DPath.Root, asNode.Right.Name, rightType);
```
CarlosFigueiraMSFT, line 117: >returnType = returnType.Add(ref fError, DPath.Root, asNode.Right.Name, rightType); [](http://example.com/codeflow?start=24&length=82)

We may need to check for duplicates here:

    Join(
      [{a:1,b:2},{a:3,b:4}],
      [{a:1,c:5},{a:3,c:6}],
      LeftRecord.a == RightRecord.a,
      JoinType.Inner,
      RightRecord.c As b
    )

src/libraries/Microsoft.PowerFx.Core/Texl/Intellisense/IntellisenseHelper.cs
```diff
@@ -357,7 +357,10 @@ public static void AddTopLevelSuggestionsForCursorType(IntellisenseData.Intellis
                     {
                         if (info.ScopeIdentifier != default)
                         {
-                            AddSuggestion(intellisenseData, info.ScopeIdentifier, SuggestionKind.Global, SuggestionIconKind.Other, type, requiresSuggestionEscaping: false);
+                            foreach (var ident in info.ScopeIdentifier)
```
CarlosFigueiraMSFT, line 360: >var ident in info.ScopeIdentifier [](http://example.com/codeflow?start=37&length=33)

Do we have tests for this change (suggestions on left/right records)?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Join.cs
```diff
@@ -0,0 +1,237 @@
...
+namespace Microsoft.PowerFx.Core.Texl.Builtins
+{
+    internal class JoinFunction : FilterFunctionBase
+    {
+        public override bool IsSelfContained => true;
+
+        public override int ScopeArgs => 2;
+
+        public JoinFunction()
+            : base("Join", TexlStrings.AboutJoin, FunctionCategories.Table, DType.EmptyTable, 0x8, 3, int.MaxValue, DType.EmptyTable, DType.EmptyTable)
```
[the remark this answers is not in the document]
anderson-joyle: This begs the question: should we swap the `JoinType` placement? Since it is an optional arg but `RightRecord.column As NewColumn` is mandatory.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Join.cs
```diff
@@ -0,0 +1,237 @@
...
+                    {
+                        if (!leftTable.TryGetType(dottedNameNode.Right.Name, out var leftType))
+                        {
+                            errors.EnsureError(DocumentErrorSeverity.Severe, args[0], TexlStrings.ErrColDNE_Name, dottedNameNode);
+                            valid = false;
+                            break;
+                        }
+
+                        returnType = returnType.Drop(ref fError, DPath.Root, dottedNameNode.Right.Name);
+                        returnType = returnType.Add(ref fError, DPath.Root, asNode.Right.Name, leftType);
```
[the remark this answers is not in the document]
anderson-joyle: `fError` will tell if there are any errors (including duplicates).

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Join.cs
anderson-joyle: `fError` will tell if there are any errors (including duplicates).
This test cases has been added.

src/libraries/Microsoft.PowerFx.Core/Functions/FunctionScopeInfo.cs
```diff
@@ -200,6 +219,18 @@ public void CheckLiteralPredicates(TexlNode[] args, IErrorContainer errors)
                     }
                 }
             }
+        }
+
+        public virtual bool GetScopeIdent(TexlNode[] nodes, out DName[] scopeIdents)
```
MikeStall, line 224: >GetScopeIdent [](http://example.com/codeflow?start=28&length=13)

Can you comment what the intention is here? And what does the return value mean? (I notice we return false with a non-null out param, which is unusual)

Maybe with an example?


The signature takes arrays, but we always return 1 item. 
Why is the output an array? Is it a parallel array to the input? 
Or could output jut be an IEnumerable? if it's just a list.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -134,11 +134,6 @@ internal sealed partial class TexlBinding
         /// </summary>
         public const int MaxSelectsToInclude = 100;
 
-        /// <summary>
-        /// Default name used to access a Lambda scope.
-        /// </summary>
-        internal static DName ThisRecordDefaultName => new DName("ThisRecord");
```
MikeStall, line 140: >ThisRecordDefaultName [](http://example.com/codeflow?start=30&length=21)

Would be interesting to know if this breaks PAClient builds? 
That's ok  the build can update. But it may mean extra churn. 

Rather than move this to a new definition on FunctionScopeInfo,  consider avoiding new one and just using existing one.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -2809,7 +2804,21 @@ public override void Visit(FirstNameNode node)
...
+                    DType nodeType = null;
+
+                    if (scope.ScopeIdentifiers == null || scope.ScopeIdentifiers.Length == 1)
+                    {
+                        nodeType = scope.Type;
+                    }
+                    else
+                    {
+                        // If scope.ScopeIdentifier.Length > 1, it meant the function creates more than 1 scope and the scope types are contained within a record.
+                        // Example: Join(t1, t2, LeftRecord.a = RigthRecord.a, ...)
```
MikeStall, line 2816: >RigthRecord [](http://example.com/codeflow?start=64&length=11)

typo

MikeStall:
>             EnumStoreBuilder?.WithRequiredEnums(functions);

not this PR ... but this seems suspicious. 
If we don't have EnumStoreBuilder allocated, does that mean we silently drop the enums here? 


---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs:325 in b5c4f5b. [](commit_id = b5c4f5b5ea383cc2db9536cebcacf6e80462317d, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Core/Public/Config/SymbolTable.cs
MikeStall: discussed - this is to prevent silently ignoring enums on line 325. 
That's probably a deeper issue to fix.

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -2705,7 +2731,7 @@ internal static bool TryGetConvertedDisplayNameAndLogicalNameForColumn(DType typ
             return false;
         }
 
-        public static DType Union(ref bool fError, DType leftType, DType rightType, bool useLegacyDateTimeAccepts, Features features, bool allowCoerce = false, bool unionToLeftTypeOnly = false)
+        public static DType Union(ref bool fError, DType leftType, DType rightType, bool useLegacyDateTimeAccepts, Features features, bool allowCoerce = false, bool unionToLeftTypeOnly = false, bool forceUniqueColumns = false)
```
MikeStall, line 2734: double check

src/libraries/Microsoft.PowerFx.Interpreter/Functions/IAsyncTexlFunctionJoin.cs
MikeStall: See #2751

src/tests/Microsoft.PowerFx.Core.Tests.Shared/EnumSyncTests.cs
```diff
@@ -7,6 +7,7 @@
 using System.Linq;
 using System.Net.Http;
 using System.Numerics;
+using System.Runtime.InteropServices;
```
MikeStall, line 10: ?

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/Join.txt
```diff
@@ -0,0 +1,62 @@
...
+>> Join(Table({a:1}),Table({b:1}),LeftRecord.a = RightRecord.b, JoinType.Inner, LeftRecord.Void As Abc)
+Errors: Error 87-92: Name isn't valid. 'Void' isn't recognized.|Error 93-95: Invalid argument type.|Error 5-17: The specified column 'LeftRecord.Void' does not exist.|Error 0-4: The function 'Join' has some invalid arguments.
+
+>> Join(Table({a:1}),Table({b:1}),LeftRecord.a = RightRecord.b, JoinType.Inner, Test.Void As Abc)
+Errors: Error 77-81: Name isn't valid. 'Test' isn't recognized.|Error 81-86: The '.' operator cannot be used on Error values.|Error 87-89: Invalid argument type.|Error 0-4: The function 'Join' has some invalid arguments.
+
+>> Join(Table({a:1}),Table({b:1}),LeftRecord.a = RightRecord.b, JoinType.Inner, LeftRecord.a As A1, RightRecord.b As A1)
+Errors: Error 5-17: 'RightRecord.b As A1' can not be added/renamed due to colission with another column with same name.|Error 0-4: The function 'Join' has some invalid arguments.
+
+>> Join(If(1/0, Table({a:1})), Table({b:1}), LeftRecord.a = RightRecord.b,JoinType.Inner,RightRecord.b As b)
```
MikeStall, line 40: >(If(1/0, Table({a:1})) [](http://example.com/codeflow?start=7&length=22)

error is good. 

Should also check Blank:
If(false, Table({}))

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/Join.txt
```diff
@@ -0,0 +1,62 @@
...
+
+>> Join(1/0, Table({SupplierId:"AAA111"}), LeftRecord.Id = RightRecord.SupplierId, JoinType.Full)
+Errors: Error 0-94: Invalid number of arguments: received 4, expected 5 or more.|Error 40-50: Name isn't valid. 'LeftRecord' isn't recognized.|Error 50-53: The '.' operator cannot be used on Error values.|Error 56-67: Name isn't valid. 'RightRecord' isn't recognized.|Error 67-78: The '.' operator cannot be used on Error values.|Error 54-55: Incompatible types for comparison. These types can't be compared: Error, Error.
+
+// 'JoinType' arg must be a plain enum value. Passing it through another function is not allowed.
+>> Join(Table({Id:"AAA111"}),  Table({SupplierId:"AAA111"}), LeftRecord.Id = RightRecord.SupplierId, If(true,JoinType.Inner), RightRecord.SupplierId As SupplierId)
+Errors: Error 98-121: Expected plain 'JoinType' enum value. Different enums or passing it throught another function is not supported i.e. Join(..., If(true, JoinType.Inner, JoinType.Full)).|Error 0-4: The function 'Join' has some invalid arguments.
+
+// Testing error when evaluating the predicate
+>> Join(Table({a:1},{a:1}),Table({b:1},{b:2}),1/0,JoinType.Inner,RightRecord.b As b)
```
MikeStall, line 25: >1/0 [](http://example.com/codeflow?start=46&length=3)

What if some were errors but some were correct?

12 / LeftRecord.Value ?

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4544,26 +4564,40 @@ public override bool PreVisit(CallNode node)
...
+                {
+                    // Starting from 1 since 0 was visited above.
+                    for (int i = 1; i < maybeFunc.ScopeArgs; i++)
+                    {
+                        _txb.AddVolatileVariables(node, _txb.GetVolatileVariables(args[i]));
+                        args[i].Accept(this);
+                        typeInputs[args[i].Id] = _txb.GetType(args[i]);
+                    }
+
+                    fArgsValid = scopeInfo.CheckInput(_txb.Features, node, args, out typeScope, typeInputs.Values.ToArray());
```
MikeStall, line 4597: >typeInputs.Values.ToArray() [](http://example.com/codeflow?start=96&length=27)

This maybe wrong.   Dictionaries are unordered.

src/libraries/Microsoft.PowerFx.Core/Functions/FunctionScopeInfo.cs
```diff
@@ -200,6 +215,26 @@ public void CheckLiteralPredicates(TexlNode[] args, IErrorContainer errors)
...
+        {
+            scopeIdents = new[] { TexlBinding.ThisRecordDefaultName };
+            if (nodes[0] is AsNode asNode)
+            {
+                scopeIdents = new[] { asNode.Right.Name };
+                return true;
+            }
+
+            return false;
         }
```
jas-valgotar, line 238: When does this return true/false, how is that used? should this always return scopeIdents?

anderson-joyle: It always return scopeIdents. The return will dictate if binder will work with "standard" scope names (ThisRecord, LeftRecord, RightRecord) or a alias (`t1 As NewAlias`).

PR #2729, opened 2024-11-04
Fixes enum store when SymbolTable is replaced while setting up config

jas-valgotar:
This PR fixes the issue where inbuilt enums are no longer part of the config if the config's symbol table is replaced after initialization.

After initializing the config if SymbolTable was replaced, we were loosing the enums attached to the previous default symbol table. This PR will adds a default symbol table(seprate from existing table) to handle that.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/ComposedReadOnlySymbolTable.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/ComposedReadOnlySymbolTable.cs
@@ -287,20 +287,5 @@ namespace Microsoft.PowerFx
                 return default;
             }
         }
-
-        internal override bool TryGetVariable(DName name, out NameLookupInfo symbol, out DName displayName)
-        {
-            foreach (var symbolTable in _symbolTables)
-            {
-                if (symbolTable.TryGetVariable(name, out symbol, out displayName))
-                {
-                    return true;
-                }
-            }
-
-            symbol = default;
-            displayName = default;
-            return false;
-        }
     }
 }
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/ComposedReadOnlySymbolTable.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/ComposedReadOnlySymbolTable.cs
@@ -287,5 +287,20 @@ namespace Microsoft.PowerFx
                 return default;
             }
         }
+
+        internal override bool TryGetVariable(DName name, out NameLookupInfo symbol, out DName displayName)
+        {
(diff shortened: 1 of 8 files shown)


src/libraries/Microsoft.PowerFx.Core/Public/Config/ComposedReadOnlySymbolTable.cs
```diff
@@ -287,5 +287,20 @@ internal override IExternalEntityScope InternalEntityScope
                 return default;
             }
         }
+
+        internal override bool TryGetVariable(DName name, out NameLookupInfo symbol, out DName displayName)
```
MikeStall, line 291: >override [](http://example.com/codeflow?start=17&length=8)

not a PR blocker, but I wonder if we need a minor refactoring to catch these. IE, make this abstract; and more more things out of ReadOnlySymbolTable.

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -4,6 +4,8 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Xml.Linq;
+using Microsoft.CodeAnalysis;
```
MikeStall, line 8: ?

src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
```diff
@@ -23,20 +25,41 @@ public sealed class PowerFxConfig
...
+        /// <summary>
+        /// Global symbols. Additional symbols beyond default function set and primitive types defined by host.
+        /// </summary>
+        /// 
+        public SymbolTable SymbolTable
         {
-            DebugName = "DefaultConfig"
-        };
+            get => _symbolTable;
+            set => _symbolTable = value ?? new SymbolTable();
```
MikeStall, line 54: this is significant?
Why allow setting to null?

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

MikeStall (approved):

MikeStall:
It turns out this is a breaking change,. 

        [Fact]
        public async Task NewFunctionShowsInEnumeration()
        {
            config.AddFunction(new Func1Function());            
        }
        
Before,         config.SymbolTable.FunctionNames.ToArray() has the new function "func1". Now, it does not, and the function is at: config.InternalConfigSymbols.FunctionNames.ToArray()

This is breaking some tests in Dataverse repro trying to uptake it (see CompileBasic).

PR #2731, opened 2024-11-06
Fix TimeValue function, add test

### LINKED ISSUE #2674: TimeValue function has many issues
Some values should not be accepted, but they currently are. For example:
image
is rejected on Excel, but "works" on the REPL:
image

Some values are incorrect:
- `TimeValue("27:00:00")` should return the time equivalent to 3am (0.125); it returns 27 (00:00:00 27 days after date 0)
- `TimeValue("6:00 pm")` returns an error (it should return Time(18,0,0,0))
- `TimeValue("1234")` return 00:00:00 1234 days after date 0 (should be an error, see above)

CarlosFigueiraMSFT:
Fixes #2674 .

The TimeValue function has a few issues, which are fixed by this change:
- Better validation (so that calls such as `TimeValue("1")` fail, as it does in Excel
- Support for AM/PM designators (`TimeValue("6:00 PM")`), as it is supported in Excel
- Support for wrapping hours (`Value(TimeValue("27:00:00")) = Value(TimeValue("3:00:00"))`), as it is done in Excel

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/TimeValue.txt
+++ b/src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/TimeValue.txt
@@ -1,0 +1,52 @@
+﻿>> TimeValue("12:00:00")
+Time(12,0,0,0)
+
+>> TimeValue("27:00:00")
+Time(3,0,0,0)
+
+// TimeValue only returns values between [00:00:00.000, 23:59:59.999)
+>> Value(TimeValue("27:00:00"))
+0.125
+
+>> TimeValue("6:00")
+Time(6,0,0,0)
+
+>> TimeValue("6:00 PM")
+Time(18,0,0,0)
+
+// Date portion is ignored
+>> Value(TimeValue("10/12/2024 6:00:00 AM"))
+0.25
+
+>> TimeValue("29 Feb 2008 9:21:33 AM")
+Time(9,21,33,0)
+
+>> TimeValue("12:34:56.7")
+Time(12,34,56,700)
+
+>> TimeValue("12:34:56.78")
+Time(12,34,56,780)
+
+>> TimeValue("12:34:56.789")
+Time(12,34,56,789)
+
+>> TimeValue("6:01:02")
+Time(6,1,2,0)
+
+>> TimeValue("14:18")
+Time(14,18,0,0)
+
+>> TimeValue("24:11:11")
+Time(0,11,11,0)
+
+>> TimeValue("Not a time")
+Error({Kind:ErrorKind.InvalidArgument})
+
+>> TimeValue("One PM")
+Error({Kind:ErrorKind.InvalidArgument})
+
+>> TimeValue("1234")
+Error({Kind:ErrorKind.InvalidArgument})
+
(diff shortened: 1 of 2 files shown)


LucGenetier:
✅ No public API change.

adithyaselv:
Probably not something for this PR but should consider having a similar support in JSON parsing - https://github.com/microsoft/Power-Fx/blob/c5a3aeadb5701acf023c8a01ef687186272555fd/src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs#L153

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/TimeValue.txt
```diff
@@ -0,0 +1,52 @@
...
+Time(6,0,0,0)
+
+>> TimeValue("6:00 PM")
+Time(18,0,0,0)
+
+// Date portion is ignored
+>> Value(TimeValue("10/12/2024 6:00:00 AM"))
+0.25
+
+>> TimeValue("29 Feb 2008 9:21:33 AM")
```
adithyaselv, line 21: does this also work with time zone specified ? a test would help

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/TimeValue.txt
```diff
@@ -0,0 +1,52 @@
...
+Time(9,21,33,0)
+
+>> TimeValue("12:34:56.7")
+Time(12,34,56,700)
+
+>> TimeValue("12:34:56.78")
+Time(12,34,56,780)
+
+>> TimeValue("12:34:56.789")
```
adithyaselv, line 30: may be another test with millisecond more than 1000 ?

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/TimeValue.txt
CarlosFigueiraMSFT: It does, but it can lead to some really strange expressions (`TimeValue("09:00:00-08:00"), which doesn't look like what it does. I'd rather leave this as an "unspecified" feature, instead of using the test to spec what it should do.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/TimeValue.txt
CarlosFigueiraMSFT: Good idea, added.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT:
JSON parsing of date/time values is restricted to the values that can be generated by the JSON function - it doesn't support arbitrary formats like the typed versions do. We can decide to relax those restrictions later, but that's a historical reason why they exist...

---
In reply to: [2462887581](https://github.com/microsoft/Power-Fx/pull/2731#issuecomment-2462887581) [](http://example.com/codeflow?ancestors=2462887581)

adithyaselv (approved):
:shipit:

Issue #2735, opened 2024-11-11
Evaluate PowerFX String Expressions for Static Analysis


gekarapa:
For specific use cases, for static analysis, we need to retrieve the evaluated result of a PowerFX string expressions. To enable this, we need a method to get the PowerFX output from both string and interpolated string expressions, provided that the interpolated strings do not contain any island expressions.

MikeStall:
Can you elaborate on the scenario?

gekarapa:
In the DLP endpoint filtering feature, we need to restrict specific values within action arguments. For example, a user might want to block requests to certain URLs, which will be done by comparing the runtime expression value against the endpoint filtering rule. Note that this restriction will only apply on design time when the expression is a static string, not in other cases.

PR #2736, opened 2024-11-11
Helper to get static string from parser

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2735.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Parser/ParseResult.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Parser/ParseResult.cs
@@ -118,5 +118,34 @@ namespace Microsoft.PowerFx
         {
             return StructuralPrint.Print(Root);
         }
+
+        /// <summary>
+        /// Checks if the expression does not depend on extra evaluations.
+        /// Examples: 
+        /// "Lorem ipsum" // is a static expression since there is no extra evaluation needed.
+        /// "Today is " & Today() // is not a static expression since the function call 'Today()' needs to be evaluated.
+        /// $"Today is {""some text""}" // is not a static expression since the interpolation needs to be evaluated.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Core/Parser/ParseResult.cs
```diff
@@ -118,5 +118,34 @@ public string GetAnonymizedFormula()
...
+        /// Checks if the expression does not depend on extra evaluations.
+        /// Examples: 
+        /// "Lorem ipsum" // is a static expression since there is no extra evaluation needed.
+        /// "Today is " & Today() // is not a static expression since the function call 'Today()' needs to be evaluated.
+        /// $"Today is {""some text""}" // is not a static expression since the interpolation needs to be evaluated.
+        /// </summary>
+        /// <returns>True if no extra evaluation is needed. False otherwise.</returns>
+        public bool TryGetStaticTextExpression(out string staticExpression)
+        { 
+            if (Options.TextFirst &&
```
MikeStall, line 132: >Options.TextFirst && [](http://example.com/codeflow?start=16&length=20)

can we avoid this?

src/libraries/Microsoft.PowerFx.Core/Parser/ParseResult.cs
```diff
@@ -118,5 +118,34 @@ public string GetAnonymizedFormula()
...
+                return true;
+            }
+
+            if (Root is StrLitNode strLitNode1)
+            {
+                staticExpression = strLitNode1.Value;
+                return true;
+            }
+
+            staticExpression = null;
```
MikeStall, line 147: what about numbers? (this gets to the purpose of the feature)

gekarapa: Today, this requirement is limited to string type arguments, but in the future, we might need to extend it.

PR #2738, opened 2024-11-12
Add x-ms-enum-values management in CDP connectors

LucGenetier:
Add x-ms-enum-values in CDP connectors
Accumulate option sets to surface them in CdpTable

Optionset names will be 'propertyName (tableName)'
If existing, propertyName will be replaced with 'x-ms-enum' value


src/libraries/Microsoft.PowerFx.Connectors/Internal/OptionSetList.cs
```diff
@@ -0,0 +1,115 @@
...
+        public OptionSetList()
+        {
+            _optionSets = new Dictionary<string, OptionSet>();            
+        }
+
+        public IEnumerable<OptionSet> OptionSets => _optionSets.Values;
+
+        // if the option set doesn't exist, it will be added to the list and we return optionSet
+        // if the option set exists with same name and options, it is not added, we return the existing optionSet
+        // if the option set exists with same name but different options, we rename it and return that new optionSet
```
MikeStall, line 24: we should discuss

LucGenetier: we will not allow conflicts

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
```diff
@@ -292,15 +294,23 @@ public ConnectorType ReturnParameterType
         // Those properties are only used by HttpFunctionInvoker
         internal ConnectorParameterInternals _internals = null;
 
-        private readonly ConnectorLogger _configurationLogger = null;
+        private readonly ConnectorLogger _configurationLogger = null;
+
+        internal readonly OptionSetList OptionSets;
+
+        // CDP only
+        internal readonly string TableName;
```
MikeStall, line 302: This seems strange - 
Why would CDP use ConnectorFunction at all? CDP can directly call the endpoint.

LucGenetier: This is used in GetCdpTableType even though we don't use the connection function
It is there as we share existing code (used for dynamic intellisense)
As explained during our call, the CDP schema is a swagger that defines types (tables) and no API and we share the same code/extensions

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Connectors.Tests.Shared/Responses/ZD Tickets GetSchema.json
```diff
@@ -18,254 +18,308 @@
...
         "priority": {
-          "title": "priority",
-          "x-ms-capabilities": { "filterFunctions": [ "lt", "le", "eq", "ne", "gt", "ge", "min", "max", "countdistinct", "contains", "startswith", "endswith", "length", "indexof", "replace", "substring", "substringof", "tolower", "toupper", "trim", "concat" ] },
           "type": "string",
-          "enum": [
-            "low",
-            "normal",
-            "high",
-            "urgent"
+          "x-ms-enum-values": [
```
MikeStall, line 96: >"x-ms-enum-values": [](http://example.com/codeflow?start=10&length=19)

Do we need a test case for x-ms-enum? 

Are we losing coveraging dropping "enum" on line 95?

LucGenetier: already tested with action connectors (this code is shared)
dropping "enum" is not an issue as this extension has precedence (that's how i define it - I can't prove it)

src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
LucGenetier: Can be removed

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorParameter.cs
```diff
@@ -22,30 +22,31 @@ public class ConnectorParameter : ConnectorSchema
 
         internal bool IsBodyParameter = false;
 
-        internal ConnectorParameter(OpenApiParameter openApiParameter, ConnectorCompatibility compatibility)
-            : this(openApiParameter, null, false, compatibility)
+        internal ConnectorParameter(OpenApiParameter openApiParameter, SymbolTable optionSets, ConnectorCompatibility compatibility)
```
MikeStall, line 25: >SymbolTable [](http://example.com/codeflow?start=71&length=11)

Can this be ReadOnlySymbolTable?
ie, are we mutating this?

LucGenetier: I've removed it

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2740, opened 2024-11-13
Only create one instance of Features.None/PFxV1

lesaltzm:


CarlosFigueiraMSFT (dismissed):

src/libraries/Microsoft.PowerFx.Core/Public/Config/Features.cs
```diff
@@ -96,9 +96,9 @@ public sealed class Features
         /// </summary>
         internal bool IsUserDefinedTypesEnabled { get; set; } = false;
 
-        internal static Features None => new Features();
+        internal static readonly Features None = new Features();
 
-        public static Features PowerFxV1 => new Features
+        public static readonly Features PowerFxV1 = new Features
```
CarlosFigueiraMSFT, line 101: The instances are readonly, but now it's possible that someone inadvertently change the meaning of those properties (i.e., by doing something like `Features.None.IsUserDefinedTypesEnabled = true`)

CarlosFigueiraMSFT: If we are to do this change, we should change this class to be immutable, and add a constructor where we can toggle those values.

lesaltzm: Yup, changing set to init on all these

MikeStall (approved):

CarlosFigueiraMSFT:
>         internal Features(Features other)

I know this is not related to your changes, but since you're touching this file, can you update this copy constructor to add the missing properties?

AllowImpureNodeDelegation, AllowImpureNodeDelegation, AsTypeLegacyCheck, IsLegacyAnalysis, JsonFunctionAcceptsLazyTypes, IsLookUpReductionDelegationEnabled

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/Config/Features.cs:121 in 909a83b. [](commit_id = 909a83b6b2c5837e9258c16d9f09861f51ce1973, deletion_comment = False)

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Features.get_PowerFxV1

CarlosFigueiraMSFT (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
```diff
@@ -707,7 +707,7 @@ public CheckContextSummary ApplyGetContextSummary()
             var summary = new CheckContextSummary
             {
                 AllowsSideEffects = allowSideEffects,
-                IsPreV1Semantics = isV1,
+                IsPreV1Semantics = !isV1,
```
Shpakh-MSFT, line 710: >!isV1 [](http://example.com/codeflow?start=35&length=5)

Was this a bug?

Shpakh-MSFT (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
lesaltzm: Yes, discussed with Mike and Hoa. The equality check above was always false before

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Features.get_PowerFxV1

MikeStall:
Somehow, this change is causing a regression and this is now failing:

            PowerFxConfig config = new PowerFxConfig(); // Pass in per engine
            Assert.Equal(config.Features.SupportColumnNamesAsIdentifiers, true);

It changes a static property to a static field - probably some .Net race condition around static field initialization.

PR #2741, opened 2024-11-14
Block Type literals from being used in invalid contexts

adithyaselv:
Today we allow expressions like the following in PowerApps

```
If(Type(Boolean), 1, 2)
``` 

this PR blocks such unintended usage of type literals at compile time.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -2557,6 +2557,9 @@ namespace Microsoft.PowerFx.Core.Binding
                 _txb.SetType(node, DType.ObjNull);
             }
 
+            // Binding TypeLiteralNode from anywhere other than valid type context should be an error.
+            // This ensures that binding of unintended use of TypeLiteralNode eg: "If(Type(Boolean), 1, 2)" will result in an error.
+            // VisitType method is used to resolve the type of TypeLiteralNode from valid context. 
             public override void Visit(TypeLiteralNode node)
             {
                 AssertValid();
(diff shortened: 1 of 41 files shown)


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -2558,6 +2558,15 @@ public override void Visit(BlankNode node)
...
             public override void Visit(TypeLiteralNode node)
+            {
+                AssertValid();
+                Contracts.AssertValue(node);
+                
+                _txb.SetType(node, DType.Error);
+                _txb.ErrorContainer.Error(node, TexlStrings.ErrTypeLiteral_UnsupportedUsage);
+            }
+
+            private void VisitType(TypeLiteralNode node)
```
lesaltzm, line 2569: comment to explain the difference/why the normal visitor pattern isn't right.

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4560,6 +4560,10 @@
     <value>Type literal declaration is invalid. The expression '{0}' cannot be used in a type definition.</value>
     <comment>Error message when validation of type literal fails.</comment>
   </data>
+  <data name="ErrTypeLiteral_UnsupportedUsage" xml:space="preserve">
+    <value>Unsupported usage: Type literals can only be used in type arguments and type definitions.</value>
```
CarlosFigueiraMSFT, line 4564: casing: "Unsupported usage: type literals can only be used..."

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2743, opened 2024-11-15
JSON - don't generate an error for Lazy types when they are allowed

LucGenetier:


anderson-joyle (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2744, opened 2024-11-16
Undo breaking change in Features

MikeStall:
Followup to #2740
This changes Features.PowerFxV1 from property to field - which is binary breaking change. 
Undo the break  - but still keep as singleton. 

Add test.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/Features.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/Features.cs
@@ -94,7 +94,12 @@ namespace Microsoft.PowerFx
 
         internal static readonly Features None = new Features();
 
-        public static readonly Features PowerFxV1 = new Features
+        /// <summary>
+        /// The default V1 Power Fx feature set. 
+        /// </summary>
+        public static Features PowerFxV1 => _powerFxV1;
+
+        private static readonly Features _powerFxV1 = new Features
         {
             TableSyntaxDoesntWrapRecords = true,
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Core/Public/Config/Features.cs
```diff
@@ -94,7 +94,12 @@ public sealed class Features
 
         internal static readonly Features None = new Features();
 
-        public static readonly Features PowerFxV1 = new Features
+        /// <summary>
+        /// The default V1 Power Fx feature set. 
+        /// </summary>
+        public static Features PowerFxV1 => _powerFxV1;
```
MikeStall, line 100: Keep this as a property,  #2740 changed it to a field, which is a binary breaking change.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/FeaturesTest.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Microsoft.PowerFx.Tests
+{
+    public class FeaturesTest
+    {
+        [Fact]
+        public void Singleton()
+        {
+            // Ensure that we have a singleton - this is for performance reasons.
+            Assert.True(object.ReferenceEquals(Features.PowerFxV1, Features.PowerFxV1));
```
MikeStall, line 14: This test verifies the key benefit of #2740

lesaltzm (approved):

LucGenetier:
✅ No public API change.

PR #2745, opened 2024-11-19
Fix flaky tests - Features is now a singleton.

MikeStall:
Tests would use reflection to mutate the Features object. 

But #2740 made `Features` a singleton.  So the object was getting randomly mutated. We'd see this as dataverse tests failing.


Ananya-Gupta24 (approved):

nguhoa (approved):

LucGenetier:
✅ No public API change.

PR #2746, opened 2024-11-19
Remove reflection on test setup

CarlosFigueiraMSFT:


LucGenetier:
✅ No public API change.

MikeStall:
Followup to #2740. Better fix than what we had in #2745.

MikeStall (approved):

PR #2747, opened 2024-11-20
Allow multiple relationships in ConnectorType

LucGenetier:
Allow multiple relationships in ConnectorType
Fix CompileTimeTypeWrapperRecordValue.TryGetField
Add Equals to ConnectorType

This update doesn't update CdpRecordType yet
DV and NL repos are not impacted by these changes

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
@@ -18,6 +18,8 @@ using Microsoft.PowerFx.Core.Utils;
 using Microsoft.PowerFx.Types;
 using static Microsoft.PowerFx.Connectors.Constants;
 
+#pragma warning disable SA1000 // The keyword 'new' should be followeed by a space
+
 namespace Microsoft.PowerFx.Connectors
 {
     // See definitions for x-ms extensions:
@@ -325,7 +327,7 @@ namespace Microsoft.PowerFx.Connectors
             }
             else if (openApiAny is IDictionary<string, IOpenApiAny> o)
             {
-                Dictionary<string, FormulaValue> dvParams = new ();
+                Dictionary<string, FormulaValue> dvParams = new();
 
                 foreach (KeyValuePair<string, IOpenApiAny> kvp in o)
                 {
@@ -675,17 +677,14 @@ namespace Microsoft.PowerFx.Connectors
                                 // Here, we have a circular reference and default to a string
                                 return new ConnectorType(schema, openApiParameter, FormulaType.String, hiddenfields.ToRecordType());
                             }
-                            
+
                             ConnectorType propertyType = new SwaggerParameter(propLogicalName, schema.Required.Contains(propLogicalName), kv.Value, kv.Value.Extensions).GetConnectorType(settings.Stack(schemaIdentifier));
 
-                            if (settings.SqlRelationships != null)
-                            {
-                                IEnumerable<SqlRelationship> relationships = settings.SqlRelationships.Where(sr => sr.ColumnName == propLogicalName);
+                            IEnumerable<SqlRelationship> relationships = settings.SqlRelationships?.Where(sr => sr.ColumnName == propLogicalName);
 
-                                if (relationships.Any())
-                                {
-                                    propertyType.SetRelationships(relationships);
-                                }
+                            if (relationships?.Any() == true)
+                            {
+                                propertyType.SetRelationships(relationships);
                             }
 
                             settings.UnStack();
@@ -817,7 +816,7 @@ namespace Microsoft.PowerFx.Connectors
         }
 
         public static FormulaType GetReturnType(this OpenApiOperation openApiOperation, ConnectorCompatibility compatibility)
-        {            
+        {
             ConnectorType connectorType = openApiOperation.GetConnectorReturnType(compatibility);
             FormulaType ft = connectorType.HasErrors ? ConnectorType.DefaultType : connectorType?.FormulaType ?? new BlankType();
             return ft;
@@ -889,7 +888,7 @@ namespace Microsoft.PowerFx.Connectors
         /// <exception cref="NotImplementedException">When we cannot determine the content type to use.</exception>
         public static (string ContentType, OpenApiMediaType MediaType) GetContentTypeAndSchema(this IDictionary<string, OpenApiMediaType> content)
         {
-            Dictionary<string, OpenApiMediaType> list = new ();
+            Dictionary<string, OpenApiMediaType> list = new();
 
             foreach (var ct in _knownContentTypes)
             {
@@ -1044,7 +1043,7 @@ namespace Microsoft.PowerFx.Connectors
             {
                 // Parameters is required in the spec but there are examples where it's not specified and we'll support this condition with an empty list
                 IDictionary<string, IOpenApiAny> op_prms = apiObj.TryGetValue("parameters", out IOpenApiAny openApiAny) && openApiAny is IDictionary<string, IOpenApiAny> apiString ? apiString : null;
-                ConnectorDynamicValue cdv = new (op_prms);
+                ConnectorDynamicValue cdv = new(op_prms);
 
                 // Mandatory operationId for connectors, except when capibility or builtInOperation are defined
                 apiObj.WhenPresent("operationId", (opId) => cdv.OperationId = OpenApiHelperFunctions.NormalizeOperationId(opId));
@@ -1073,7 +1072,7 @@ namespace Microsoft.PowerFx.Connectors
                 {
                     // Parameters is required in the spec but there are examples where it's not specified and we'll support this condition with an empty list
                     IDictionary<string, IOpenApiAny> op_prms = apiObj.TryGetValue("parameters", out IOpenApiAny openApiAny) && openApiAny is IDictionary<string, IOpenApiAny> apiString ? apiString : null;
-                    ConnectorDynamicList cdl = new (op_prms)
+                    ConnectorDynamicList cdl = new(op_prms)
                     {
                         OperationId = OpenApiHelperFunctions.NormalizeOperationId(opId.Value),
                     };
@@ -1106,7 +1105,7 @@ namespace Microsoft.PowerFx.Connectors
 
         internal static Dictionary<string, IConnectorExtensionValue> GetParameterMap(this IDictionary<string, IOpenApiAny> opPrms, SupportsConnectorErrors errors, bool forceString = false)
         {
-            Dictionary<string, IConnectorExtensionValue> dvParams = new ();
+            Dictionary<string, IConnectorExtensionValue> dvParams = new();
 
             if (opPrms == null)
             {
@@ -1203,7 +1202,7 @@ namespace Microsoft.PowerFx.Connectors
                     // Parameters is required in the spec but there are examples where it's not specified and we'll support this condition with an empty list
                     IDictionary<string, IOpenApiAny> op_prms = apiObj.TryGetValue("parameters", out IOpenApiAny openApiAny) && openApiAny is IDictionary<string, IOpenApiAny> apiString ? apiString : null;
 
-                    ConnectorDynamicSchema cds = new (op_prms)
+                    ConnectorDynamicSchema cds = new(op_prms)
                     {
                         OperationId = OpenApiHelperFunctions.NormalizeOperationId(opId.Value),
                     };
@@ -1235,7 +1234,7 @@ namespace Microsoft.PowerFx.Connectors
                     // Parameters is required in the spec but there are examples where it's not specified and we'll support this condition with an empty list
                     IDictionary<string, IOpenApiAny> op_prms = apiObj.TryGetValue("parameters", out IOpenApiAny openApiAny) && openApiAny is IDictionary<string, IOpenApiAny> apiString ? apiString : null;
 
-                    ConnectorDynamicProperty cdp = new (op_prms)
+                    ConnectorDynamicProperty cdp = new(op_prms)
                     {
                         OperationId = OpenApiHelperFunctions.NormalizeOperationId(opId.Value),
                     };
--- a/src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
@@ -675,17 +675,16 @@ namespace Microsoft.PowerFx.Connectors
                                 // Here, we have a circular reference and default to a string
                                 return new ConnectorType(schema, openApiParameter, FormulaType.String, hiddenfields.ToRecordType());
                             }
-
-                            //ConnectorType propertyType = new OpenApiParameter() { Name = propLogicalName, Required = schema.Required.Contains(propLogicalName), Schema = kv.Value, Extensions = kv.Value.Extensions }.GetConnectorType(settings.Stack(schemaIdentifier));
+                            
                             ConnectorType propertyType = new SwaggerParameter(propLogicalName, schema.Required.Contains(propLogicalName), kv.Value, kv.Value.Extensions).GetConnectorType(settings.Stack(schemaIdentifier));
 
                             if (settings.SqlRelationships != null)
                             {
-                                SqlRelationship relationship = settings.SqlRelationships.FirstOrDefault(sr => sr.ColumnName == propLogicalName);
+                                IEnumerable<SqlRelationship> relationships = settings.SqlRelationships.Where(sr => sr.ColumnName == propLogicalName);
 
-                                if (relationship != null)
+                                if (relationships.Any())
                                 {
-                                    propertyType.SetRelationship(relationship);
+                                    propertyType.SetRelationships(relationships);
                                 }
                             }
 
--- a/src/libraries/Microsoft.PowerFx.Connectors/Public/CdpExtensions.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/Public/CdpExtensions.cs
@@ -1,22 +1,23 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+using System.Collections.Generic;
+using System.Linq;
 using Microsoft.PowerFx.Types;
 
 namespace Microsoft.PowerFx.Connectors
 {
     public static class CdpExtensions
     {
-        public static bool TryGetFieldExternalTableName(this RecordType recordType, string fieldName, out string tableName, out string foreignKey)
+        public static bool TryGetFieldRelationships(this RecordType recordType, string fieldName, out IEnumerable<ConnectorRelationship> relationships)
         {
-            if (recordType is not CdpRecordType cdpRecordType)
+            if (recordType is not CdpRecordType cdpRecordType || (!cdpRecordType.TryGetFieldRelationships(fieldName, out relationships) && relationships.Any()))
             {
-                tableName = null;
-                foreignKey = null;
+                relationships = null;
                 return false;
-            }
+            }           
 
-            return cdpRecordType.TryGetFieldExternalTableName(fieldName, out tableName, out foreignKey);
+            return true;
         }
     }
 }
--- a/src/libraries/Microsoft.PowerFx.Connectors/Public/CdpExtensions.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/Public/CdpExtensions.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.PowerFx.Types;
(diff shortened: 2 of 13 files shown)


src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorType.cs
```diff
@@ -112,11 +112,7 @@ public class ConnectorType : SupportsConnectorErrors
         internal ISwaggerSchema Schema { get; private set; } = null;
 
         // Relationships to external tables
-        internal List<string> ExternalTables { get; set; }
-
-        internal string RelationshipName { get; set; }
```
LucGenetier, line 117: >RelationshipName [](http://example.com/codeflow?start=24&length=16)

Removing as this is never used anywhere

src/libraries/Microsoft.PowerFx.Connectors/Tabular/Capabilities/ComplexColumnCapabilities.cs
```diff
@@ -25,5 +25,10 @@ public void AddColumnCapability(string name, ColumnCapabilitiesBase capability)
 
             _childColumnsCapabilities.Add(name, capability);
         }
+
+        public override int GetHashCode()
+        {
+            throw new System.NotImplementedException();
```
LucGenetier, line 31: >NotImplementedException [](http://example.com/codeflow?start=29&length=23)

This class is coming from PowerApps and currently not used

src/libraries/Microsoft.PowerFx.Core/Public/Values/CompileTimeTypeWrapperRecordValue.cs
```diff
@@ -44,10 +44,10 @@ public override bool TryShallowCopy(out FormulaValue copy)
         protected override bool TryGetField(FormulaType fieldType, string fieldName, out FormulaValue result)
         {
             if (Type.TryGetFieldType(fieldName, out var compileTimeType))
-            {
+            {                                
                 // Only return field which were specified via the expectedType (IE RecordType),
                 // because inner record value may have more fields than the expected type.
-                if (compileTimeType == fieldType && _fields.TryGetValue(fieldName, out result))
+                if (compileTimeType.Equals(fieldType) && _fields.TryGetValue(fieldName, out result))
```
LucGenetier, line 50: >Equals [](http://example.com/codeflow?start=36&length=6)

Key change

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -675,17 +675,16 @@ internal static ConnectorType GetConnectorType(this ISwaggerParameter openApiPar
...
+                            
                             ConnectorType propertyType = new SwaggerParameter(propLogicalName, schema.Required.Contains(propLogicalName), kv.Value, kv.Value.Extensions).GetConnectorType(settings.Stack(schemaIdentifier));
 
                             if (settings.SqlRelationships != null)
                             {
-                                SqlRelationship relationship = settings.SqlRelationships.FirstOrDefault(sr => sr.ColumnName == propLogicalName);
+                                IEnumerable<SqlRelationship> relationships = settings.SqlRelationships.Where(sr => sr.ColumnName == propLogicalName);
 
-                                if (relationship != null)
+                                if (relationships.Any())
```
jas-valgotar, line 685: >relationships.Any() [](http://example.com/codeflow?start=36&length=19)

Should this be? relationships.?Any() == true

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorType.cs
```diff
@@ -148,9 +144,7 @@ internal ConnectorType(ISwaggerSchema schema, ISwaggerParameter openApiParameter
                 // SalesForce only
                 if (schema.ReferenceTo != null && schema.ReferenceTo.Count == 1)
                 {
-                    ExternalTables = new List<string>(schema.ReferenceTo);
-                    RelationshipName = schema.RelationshipName;
-                    ForeignKey = null; // SalesForce doesn't provide it, defaults to "Id"
+                    ExternalTables = new List<(string, string)>(schema.ReferenceTo.Select(r => (r, null as string /* Seems to always be Id */)));
```
jas-valgotar, line 147: >null as string /* Seems to always be Id */ [](http://example.com/codeflow?start=99&length=42)

If its always Id, should this change from null -> "Id"?

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorType.cs
```diff
@@ -336,5 +332,88 @@ private Dictionary<string, FormulaValue> GetEnum()
 
             return enumDisplayNames.Zip(enumValues, (dn, ev) => new KeyValuePair<string, FormulaValue>(dn, ev)).ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
         }
+
+        public bool Equals(ConnectorType other)
+        {
+            return this.Name == other.Name &&
```
jas-valgotar, line 338: >this.Name == other.Name && [](http://example.com/codeflow?start=19&length=26)

We should check for null first.

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorType.cs
```diff
@@ -20,7 +20,7 @@ namespace Microsoft.PowerFx.Connectors
     // FormulaType is used to represent the type of the parameter in the Power Fx expression as used in Power Apps
     // ConnectorType contains more details information coming from the swagger file and extensions
     [DebuggerDisplay("{FormulaType._type}")]
-    public class ConnectorType : SupportsConnectorErrors
+    public class ConnectorType : SupportsConnectorErrors, IEquatable<ConnectorType>
```
jas-valgotar, line 23: >IEquatable<ConnectorType> [](http://example.com/codeflow?start=58&length=25)

Why do we need this?

src/libraries/Microsoft.PowerFx.Connectors/Tabular/Capabilities/ColumnCapabilitiesDefinition.cs
```diff
@@ -33,5 +35,32 @@ public ColumnCapabilitiesDefinition(IEnumerable<string> filterFunction)
             // ex: lt, le, eq, ne, gt, ge, and, or, not, contains, startswith, endswith, countdistinct, day, month, year, time
             FilterFunctions = filterFunction;           
         }
+
+        public bool Equals(ColumnCapabilitiesDefinition other)
+        {
+            return Enumerable.SequenceEqual(this.FilterFunctions ?? Array.Empty<string>(), other.FilterFunctions ?? Array.Empty<string>()) &&
+                   this.QueryAlias == other.QueryAlias &&
```
jas-valgotar, line 42: nit: check for null first

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorType.cs
```diff
@@ -112,11 +112,7 @@ public class ConnectorType : SupportsConnectorErrors
         internal ISwaggerSchema Schema { get; private set; } = null;
 
         // Relationships to external tables
-        internal List<string> ExternalTables { get; set; }
-
-        internal string RelationshipName { get; set; }
-
-        internal string ForeignKey { get; set; }
+        internal List<(string foreignTable, string foreignKey)> ExternalTables { get; set; }
```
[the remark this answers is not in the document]
LucGenetier: on internal objects I don't think IEnumrable<> is necessary but let's change that

On the POCO, I agree it's a very good idea as this new class might evolve over time

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
LucGenetier: Good optimization

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorType.cs
LucGenetier: The problem is that nobody has been able to confirm if it's "Id" or the primary key (and if the primary key is always "Id" in SalesForce) or if it could be something else.
I have also searched the SalesForce dev docs w/o finding anything.

As a result I prefer keeping "null" to say "I don't know" rather than asserting, maybe wrongly", "it"s Id"

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorType.cs
LucGenetier: IEquatable<T> is the interface implementing Equals(T)
It's a good practice to use this interface when you implement Equals

See [https://learn.microsoft.com/en-us/dotnet/api/system.iequatable-1?view=net-8.0](https://learn.microsoft.com/en-us/dotnet/api/system.iequatable-1?view=net-8.0)

The IEquatable<T> interface is used by generic collection objects such as Dictionary<TKey,TValue>, List<T>, and LinkedList<T> when testing for equality in such methods as Contains, IndexOf, LastIndexOf, and Remove. It should be implemented for any object that might be stored in a generic collection.

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Connectors.CdpExtensions.TryGetFieldExternalTableName(Microsoft.PowerFx.Types.RecordType,System.String,System.String@,System.String@)

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorRelationship.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using Microsoft.PowerFx.Core.Utils;
+
+namespace Microsoft.PowerFx.Connectors
+{
+    public class ConnectorRelationship
```
MikeStall, line 8: >ConnectorRelationship [](http://example.com/codeflow?start=17&length=21)

Is this CDP specific? 

If so, how do non-CDP shims (ServiceNow) leverage it?

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorType.cs
MikeStall: but why are we putting these in a collection? My concern is the complexity of .equals below. Is there some simplification we can apply here?   Ie, is it sufficient to just compare logical names; or do we need a full comparison?

LucGenetier: We can probably simplify it a bit (logical names aren't enough) and i'll check
We need it due to the change in CompileTimeTypeWrapperRecordValue
IEquatable<T> is a best practice anyway, regardless of using collections

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorType.cs
```diff
@@ -336,5 +342,92 @@ private Dictionary<string, FormulaValue> GetEnum()
...
+            if (other == null)
+            {
+                return false;
+            }
+
+            return this.Name == other.Name &&
+                   this.DisplayName == other.DisplayName &&
+                   this.Description == other.Description &&
+                   this.IsRequired == other.IsRequired &&
+                   Enumerable.SequenceEqual((IList<ConnectorType>)this.Fields ?? Array.Empty<ConnectorType>(), (IList<ConnectorType>)other.Fields ?? Array.Empty<ConnectorType>()) &&
```
MikeStall, line 357: >(IList<ConnectorType>)this.Fields [](http://example.com/codeflow?start=44&length=33)

This is recursive - how do we verify this is not a stack overflow?

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorRelationship.cs
LucGenetier: for now, yes
as discussed this is a temporary thing
I'll add [Obsolete] to it

src/libraries/Microsoft.PowerFx.Connectors/Public/CdpExtensions.cs
```diff
@@ -1,22 +1,23 @@
...
+using System.Collections.Generic;
+using System.Linq;
 using Microsoft.PowerFx.Types;
 
 namespace Microsoft.PowerFx.Connectors
 {
     public static class CdpExtensions
     {
-        public static bool TryGetFieldExternalTableName(this RecordType recordType, string fieldName, out string tableName, out string foreignKey)
+        public static bool TryGetFieldRelationships(this RecordType recordType, string fieldName, out IEnumerable<ConnectorRelationship> relationships)
```
LucGenetier, line 12: >TryGetFieldRelationships [](http://example.com/codeflow?start=27&length=24)

I'll add [Obsolete]

src/libraries/Microsoft.PowerFx.Connectors/Tabular/Capabilities/ColumnCapabilities.cs
```diff
@@ -72,5 +73,32 @@ public static ColumnCapabilities ParseColumnCapabilities(IDictionary<string, IOp
                 IsChoice = isChoice
             });
         }
+
+        public bool Equals(ColumnCapabilities other)
```
MikeStall, line 77: >public bool Equals(ColumnCapabilities other) [](http://example.com/codeflow?start=8&length=44)

Intrinsically,  why would we ever need to compare ColumnCapabilities?

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorType.cs
LucGenetier: A ConnectorType is a sorta converted version of a Swagger schema and here we are comparing all elements we can identify in the schema ... and there are simply a lot
We could imagine converting the schema to a string and make a comparison of that string but the equality won't be valid in all cases as you could define types locally or as a reference, both identical, but get an equality difference (this is just one example)

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorRelationship.cs
LucGenetier: I added [Obsolete]

src/libraries/Microsoft.PowerFx.Connectors/Public/CdpExtensions.cs
LucGenetier: This API should not exist but Tahir's team is currently using it here [https://powerbi.visualstudio.com/Insights%20Apps%20Platform/_git/Platform.Service?path=/src/service/FederatedKnowledge/TabularDatasource/Implementation/DatasetTableProviders/TabularDatasetTableProvider.cs&version=GBmain&_a=contents](https://powerbi.visualstudio.com/Insights%20Apps%20Platform/_git/Platform.Service?path=/src/service/FederatedKnowledge/TabularDatasource/Implementation/DatasetTableProviders/TabularDatasetTableProvider.cs&version=GBmain&_a=contents)
We now know they shouldn't be using it (like TryGetUnderlyingFieldType) and this is due to the recent discovery that we cannot reuse Dataverse "syntax" (a.b.c.d ...) for fields having the relationships in the general CDP case (or "shims")

When Greg will have told us which direction to follow, we'll be able to remove all this code (or replace it with an improved and working version for shim-based tabular connectors)

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -17,6 +17,8 @@
 using Microsoft.PowerFx.Core.Utils;
 using Microsoft.PowerFx.Types;
 using static Microsoft.PowerFx.Connectors.Constants;
+
+#pragma warning disable SA1000 // The keyword 'new' should be followeed by a space
```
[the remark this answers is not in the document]
LucGenetier: SA1000 ([https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1000.md](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1000.md)) forces spacing for C# keywords but doesn't consider `Dictionary<string, FormulaValue> dvParams = new();` as valid due to the "missing" space between new and the constructor parenthesis
As I think it's adding some consistency and value I prefer keeping it enabled at the project level and turn it off only in this file
Very debatable, I know...

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Connectors.CdpExtensions.TryGetFieldExternalTableName(Microsoft.PowerFx.Types.RecordType,System.String,System.String@,System.String@)

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorType.cs
```diff
@@ -114,12 +114,12 @@ public class ConnectorType : SupportsConnectorErrors
...
-        // Relationships to external tables
-        internal List<string> ExternalTables { get; set; }
+        // Relationships to external tables        
+#pragma warning disable CS0618 // Type or member is obsolete
+        internal IEnumerable<ConnectorRelationship> ExternalTables => _externalTables;
 
-        internal string RelationshipName { get; set; }
-
-        internal string ForeignKey { get; set; }
+        private List<ConnectorRelationship> _externalTables;
```
jas-valgotar, line 121: >List<ConnectorRelationship> [](http://example.com/codeflow?start=16&length=27)

nit: IEnumerable<>
I know its annoying but immutable collections are much easier to maintain/debug.

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorType.cs
```diff
@@ -340,5 +352,132 @@ private Dictionary<string, FormulaValue> GetEnum()
...
+                   left.NotificationUrl == right.NotificationUrl &&
+                   ((left.HiddenRecordType == null && right.HiddenRecordType == null) || left.HiddenRecordType.Equals(right.HiddenRecordType)) &&
+                   left.Binary == right.Binary &&
+                   left.MediaKind == right.MediaKind &&
+                   Enumerable.SequenceEqual((IList<string>)left.ExternalTables ?? Array.Empty<string>(), (IList<string>)right.ExternalTables ?? Array.Empty<string>());
+        }
+
+        internal static bool ArrayEquals(ConnectorType[] leftArray, ConnectorType[] rightArray, int depth)
+        {
+            if (depth > 30)
```
jas-valgotar, line 397: >30 [](http://example.com/codeflow?start=24&length=2)

Can we hoist this out in a constant, Is 30 too big? Ie can we add a test to verify we don't stack overflow with 30 depth.

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorType.cs
```diff
@@ -340,5 +352,132 @@ private Dictionary<string, FormulaValue> GetEnum()
...
+                   left.Visibility == right.Visibility &&
+                   ((left.Capabilities == null && right.Capabilities == null) || left.Capabilities.Equals(right.Capabilities)) &&
+                   left.KeyType == right.KeyType &&
+                   left.KeyOrder == right.KeyOrder &&
+                   left.Permission == right.Permission &&
+                   left.NotificationUrl == right.NotificationUrl &&
+                   ((left.HiddenRecordType == null && right.HiddenRecordType == null) || left.HiddenRecordType.Equals(right.HiddenRecordType)) &&
+                   left.Binary == right.Binary &&
+                   left.MediaKind == right.MediaKind &&
+                   Enumerable.SequenceEqual((IList<string>)left.ExternalTables ?? Array.Empty<string>(), (IList<string>)right.ExternalTables ?? Array.Empty<string>());
```
jas-valgotar, line 392: We should also consider minimalizing it, maybe we should think of this comparison in terms of just coercion? Ie just compare field names... the problem is .Equals doesn't have good documentation on Aggregate type... but from usage it looks like its primarily used for coercion?

jas-valgotar: Do we have some test, testing equality?

MikeStall (changes requested):
:clock1:

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorType.cs
MikeStall: We'd hit millions of nodes and blow up by the time we got depth 30.  But we should identify a solution that is fundamentally correct and not approximated by depth level.

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorType.cs
```diff
@@ -340,5 +352,132 @@ private Dictionary<string, FormulaValue> GetEnum()
 
             return enumDisplayNames.Zip(enumValues, (dn, ev) => new KeyValuePair<string, FormulaValue>(dn, ev)).ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
         }
+
+        public bool Equals(ConnectorType other)
+        {
+            return Equals(this, other, 0);                  
+        }
+
+        internal static bool Equals(ConnectorType left, ConnectorType right, int depth)
```
MikeStall, line 361: >Equals [](http://example.com/codeflow?start=29&length=6)

This is a "structural type equivalence". 

Is there a stronger sense of identify for the types? (ie, why is name + container not sufficient?)

src/libraries/Microsoft.PowerFx.Connectors/Tabular/Capabilities/ColumnCapabilitiesBase.cs
```diff
@@ -8,5 +8,6 @@ namespace Microsoft.PowerFx.Connectors
     [JsonConverter(typeof(AbstractTypeConverter<ColumnCapabilitiesBase>))]
     internal abstract class ColumnCapabilitiesBase
     {
+        public abstract override int GetHashCode();
```
MikeStall, line 11: >public abstract override int GetHashCode();    [](http://example.com/codeflow?start=8&length=46)

we should not need this. 
This is just a poco - the comparison check could be done at a higher semantic level.

Issue #2752, opened 2024-11-25
Connectors - remove SQL queries to discern relationships


MikeStall:
SQL allows joins on any fields and does not tell us the join relationships.  

This code was intended to get the relationships from SQL (similar to how we can get relationships from Dataverse): 

https://github.com/microsoft/Power-Fx/blob/07a817b14a7e7df2a3524fc4a748ec60089254ba/src/libraries/Microsoft.PowerFx.Connectors/Tabular/CdpTableResolver.cs#L75

(This was added relatively recently in #2550)

But it is not correct and should be removed. (Instead, it's telling us foreign key constraints. )

This means expressions would be written like:
`LookUp(Table2, ThisRecord.Id2 = Id1).Field2`

rather than 
`_tabld1Record.Id1.Field2`

The LookUp is consistent with what PowerApps does today.

PR #2753, opened 2024-11-26
CDP - Remove SQL relationships

LucGenetier:
#2752

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/Tabular/CdpTableResolver.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/Tabular/CdpTableResolver.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net.Http;
-using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.PowerFx.Core.Entities;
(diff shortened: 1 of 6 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/Tabular/CdpTableResolver.cs
```diff
@@ -104,47 +107,51 @@ public async Task<ConnectorType> ResolveTableAsync(string tableName, Cancellatio
...
-                    ParentTable = fk.Parent_Table,
-                    ColumnName = fk.Parent_Column,
-                    ReferencedTable = fk.Referenced_Table,
-                    ReferencedColumnName = fk.Referenced_Column
-                });
-            }
-
-            return sqlRelationShips;
-        }
+        // Only keeping code for reference
```
MikeStall, line 110: nit - consider `#if false`?

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Connectors/Tabular/CdpTableResolver.cs
LucGenetier: Just a different syntax, yes

LucGenetier:
✅ No public API change.

PR #2754, opened 2024-11-27
Fix TryAddOptionSet

LucGenetier:
Was throwing a NullRefException for some connectors having option sets

### DIFF
--- a/src/tests/Microsoft.PowerFx.TexlFunctionExporter.Shared/SwaggerFileIdentification.cs
+++ b/src/tests/Microsoft.PowerFx.TexlFunctionExporter.Shared/SwaggerFileIdentification.cs
@@ -144,11 +144,13 @@ namespace Microsoft.PowerFx.TexlFunctionExporter
             return list2;
         }
 
+        // searches '<space>[v or V]<number>' pattern
+        private static readonly Regex _versionRegex = new Regex(@"\b[vV](?<v>[0-9]+)\\", RegexOptions.Compiled);
+
         private static int GetVersion(string str)
-        {
-            // searches '<space>[v or V]<number>' pattern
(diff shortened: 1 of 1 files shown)


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.TexlFunctionExporter.Shared/SwaggerFileIdentification.cs
```diff
@@ -128,6 +144,13 @@ public static Dictionary<string, string> LocateSwaggerFiles(string[] folders, st
             return list2;
         }
 
+        private static int GetVersion(string str)
+        {
+            // searches '<space>[v or V]<number>' pattern
+            Match m = new Regex(@"\b[vV](?<v>[0-9]+)\\").Match(str);
+            return m.Success ? int.Parse(m.Groups["v"].Value, CultureInfo.InvariantCulture.NumberFormat) : 0;
```
MikeStall, line 151: nit- int could throw on illegal input.  Just confirming that we want to take the exception here. 
else, Do you want TryParse and return 0 in that case?

MikeStall (dismissed):

src/tests/Microsoft.PowerFx.TexlFunctionExporter.Shared/SwaggerFileIdentification.cs
LucGenetier: "v" group can only be a set of digits
if there could be an overflow, i'll use TryParse

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #2762, opened 2024-12-02
Fixing join result when using source alias

anderson-joyle:
When joining along with alias (`Join(t1 As x1, ...)`), the binder would try to access the fields from the wrong scope to evaluate the predicate, resulting in a wrong output.


LucGenetier (approved):

LucGenetier:
✅ No public API change.

PR #2767, opened 2024-12-03
[DRAFT] UDT in REPL

adithyaselv:
draft to play around with UDT in REPL

image


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #2769, opened 2024-12-06
Add Join capability

LucGenetier:


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationCapability.cs
```diff
@@ -87,7 +87,8 @@ internal struct DelegationCapability
                 { DelegationMetadataOperatorConstants.Top, new DelegationCapability(Top) },
                 { DelegationMetadataOperatorConstants.AsType, new DelegationCapability(AsType) },
                 { DelegationMetadataOperatorConstants.ArrayLookup, new DelegationCapability(ArrayLookup) },
-                { DelegationMetadataOperatorConstants.Distinct, new DelegationCapability(Distinct) }
+                { DelegationMetadataOperatorConstants.Distinct, new DelegationCapability(Distinct) },
+                { DelegationMetadataOperatorConstants.Join, new DelegationCapability(Join) }
```
MikeStall, line 91: Is Join a single bit? Or do we need to distinct between:

LeftRecord.Id1 = RightRecord.Id2 //  [1] simple
LeftRecord.Id1 = RightRecord.Key1 & RightRecord.Key2 //  [2]  composite 

IE, most systems might let you do 1 but not 2. SQL would let you do 2.

LucGenetier: For now, I made it simple and depending on the need, we'll add more capabilities

lesaltzm (approved):

PR #2775, opened 2024-12-13
Fix Aggregate types in UDF

adithyaselv:


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -77,6 +80,27 @@ public override bool TryGetDataSource(CallNode callNode, TexlBinding binding, ou
...
+        {
+            if (!base.CheckTypes(context, args, argTypes, errors, out returnType, out nodeToCoercedTypeMap))
+            {
+                return false;
+            }
+
+            for (int i = 0; i < argTypes.Length; i++)
+            {
+                if ((argTypes[i].IsTableNonObjNull || argTypes[i].IsRecordNonObjNull) &&
+                    !ParamTypes[i].Accepts(argTypes[i], exact: true, useLegacyDateTimeAccepts: false, usePowerFxV1CompatibilityRules: context.Features.PowerFxV1CompatibilityRules, true) &&
```
CarlosFigueiraMSFT, line 93: >true [](http://example.com/codeflow?start=180&length=4)

Can you use a named argument to make it easier to understand what this parameter means?

src/libraries/Microsoft.PowerFx.Core/Types/DType.cs
```diff
@@ -1851,12 +1851,13 @@ private bool AcceptsEntityType(DType type, bool usePowerFxV1CompatibilityRules)
         /// <param name="useLegacyDateTimeAccepts">Legacy rules for accepting date/time types.</param>
         /// <param name="usePowerFxV1CompatibilityRules">Use PFx v1 compatibility rules if enabled (less
         /// permissive Accepts relationships).</param>
+        /// <param name="restrictiveAggregateTypes">Flag to restrict using aggregate types with more fields than expected.</param>
         /// <returns>
         /// True if <see cref="DType"/> accepts <paramref name="type"/>, false otherwise.
         /// </returns>
-        public bool Accepts(DType type, bool exact, bool useLegacyDateTimeAccepts, bool usePowerFxV1CompatibilityRules)
+        public bool Accepts(DType type, bool exact, bool useLegacyDateTimeAccepts, bool usePowerFxV1CompatibilityRules, bool restrictiveAggregateTypes = false)
```
CarlosFigueiraMSFT, line 1858: >restrictiveAggregateTypes [](http://example.com/codeflow?start=125&length=25)

Can you add tests to DTypeTests.cs? It is good to have unit tests closer to where the definition is

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/RecalcEngineTests.cs
```diff
@@ -1949,7 +1937,58 @@ protected override bool TryGetField(FormulaType fieldType, string fieldName, out
...
             true,
-            1.0)]
+            1.0)]
+
+        // Aggregate types with more than expected fields are not allowed in UDF args and return types
+        // Records
+        [InlineData(
+            "f():T = {x: 5, y: 5}; T := Type({x: Number});",
+            "",
+            false)]
```
CarlosFigueiraMSFT, line 1947: >false [](http://example.com/codeflow?start=12&length=5)

Can we have the tests check whether the error comes from the UDF definition or from its usage? Currently we just check that there is an error in any of those.

CarlosFigueiraMSFT:
>         public void CheckTypesOnDeclaration(CheckTypesContext context, DType actualBodyReturnType, TexlBinding binding)

Not related to your changes, but since you are touching this file: this function is only used in this class (8 lines above). Can you make it private?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs:188 in 00c51ac. [](commit_id = 00c51ac12962a3bdb0744c51548ec95600a9a5d2, deletion_comment = False)

jas-valgotar:
✅ No public API change.

CarlosFigueiraMSFT:
>         public bool AreFieldsOptional { get; set; } = false;

I know this was already there, but there is an assumption that a DType is immutable. And this breaks it - and in your test code you're using it... Not a blocker for this PR, but consider filing an issue to clean it up.

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Types/DType.cs:217 in 944c1a4. [](commit_id = 944c1a42cb9977b869103875bc22aa46f737ca9d, deletion_comment = False)

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4241,6 +4245,14 @@
...
     <comment>This error message shows up when expected return type does not match with actual return type. The arguments '{0}' and '{1}' will be replaced with data types. For example, "The stated function return type 'Number' does not match the return type of the function body 'Table'"</comment>
   </data>
+  <data name="ErrUDF_ReturnTypeSchemaIncompatible" xml:space="preserve">
+    <value>Incompatible function body. The '{0}' column in the function return type expects a '{1}' type and you’re returning a '{2}' type in function body.</value>
+    <comment>This error message shows up when expected return type schema is incompatible with actual return type. The argument '{0}' will be replaced with column name with incompatible type, '{1}' and '{2}' are replaced with expected and actual type. For example, "Incompatible function body. The 'A' column in the function return type expects a 'Text' type and you’re returning a 'GUID' type in function body."</comment>
+  </data>
+  <data name="ErrUDF_ReturnTypeSchemaAdditionalFields" xml:space="preserve">
+    <value>Invalid function body. Returned {0} value, contains an unexpected additional field '{1}'.</value>
```
CarlosFigueiraMSFT, line 4253: >nvalid function body. Returned {0} value, contains an unexpected additional field '{1}' [](http://example.com/codeflow?start=12&length=87)

Would this error message be confusing if the error was in a _nested_ type? For example: declared type `*[a:![b:n]]`, actual type `*[a:![b:n,c:n]]` - would it say that the error is on a table, or on a record?

adithyaselv: it would say Table (which is the Top level type) according to current implementation. not sure of a good way to describe as it could be deep nestings

adithyaselv: have requested greg to help with better error message and can be solved in a follow up PR as discussed offline

PR #2777, opened 2024-12-15
Fix NRE on IntellisenseHelper.TryGetInnerMostFunction

CarlosFigueiraMSFT:
Telemetry from Power Apps showed a large number of NullReferenceException instances coming from this method. A TryXXX method should return false instead of throwing a NRE.


LucGenetier:
✅ No public API change.

LucGenetier (approved):

PR #2782, opened 2024-12-17
Fix error message for mismatched types on untyped values

### LINKED ISSUE #2781: Bad error message for invalid argument to ColumnNames
See below:

Image

It should not say that the actual type is "ColumnNames"

CarlosFigueiraMSFT:
Fixes #2781 

Examples of old error:
Image

Example of new error:
image


anderson-joyle (approved):
:shipit:

PR #2783, opened 2024-12-17
Fix UDF arg parse null reference

adithyaselv:
Arg Identifier is null for certain UDF declarations eg: `SomeFunc(:` and `F(a: Boolean,:):Number = 1;`. these are intermediate states when writing UDFs and these cause Null ref exception. this makes changes to prevent NRE.


carlosfigueira (approved):
🚢

CarlosFigueiraMSFT (approved):
🚢

LucGenetier:
✅ No public API change.

Issue #2780, opened 2024-12-17
Connectors shouldn't need both HttpClient and uriPrefix


MikeStall:
HttpClient already provides the base address. So we shouldn't also need uriPrefix. 

An additional benefit is that this lets the host fully control the endpoint.  
- They can pick v1 vs. v2, 
- They can call CDP hosted outside of APIM.   (this includes localhost)

MikeStall:
This will also flush out other concerns. We extract a connectorName from the uri: 
https://github.com/microsoft/Power-Fx/blob/8d2a19061d747f0baaefc35b6ee7c3bbf42c7691/src/libraries/Microsoft.PowerFx.Connectors/Tabular/CdpTableResolver.cs#L96

we should avoid this because:
- Fx.Connectors (client code) shouldn't be having per-connector logic. CDP is supposed to be a standard interface.  
- we can't always assume the uri implies the connector type. 

Worst case - we could still get this information from httpClient.baseAddress.

PR #2785, opened 2024-12-18
Issue 2780

LucGenetier:


src/libraries/Microsoft.PowerFx.Connectors/Tabular/Services/CdpDataSource.cs
```diff
@@ -32,6 +39,21 @@ public static async Task<DatasetMetadata> GetDatasetsMetadataAsync(HttpClient ht
...
         }
 
+        public virtual async Task<IEnumerable<CdpTable>> GetTablesAsync(HttpClient httpClient, CancellationToken cancellationToken, ConnectorLogger logger = null)
+        {
+            if (DatasetMetadata == null)
+            {
+                DatasetMetadata = await GetDatasetsMetadataAsync(httpClient, cancellationToken, logger).ConfigureAwait(false);
+            }
+
+            string queryName = httpClient is PowerPlatformConnectorClient ppcc && ppcc.RequestUrlPrefix.Contains("/sharepointonline/") ? "/alltables" : "/tables";
```
MikeStall, line 49: >httpClient is PowerPlatformConnectorClient ppcc && ppcc.RequestUrlPrefix.Contains [](http://example.com/codeflow?start=31&length=81)

The issue description in #2780 suggested dealing with this  case by looking at HttpClient.BaseAddress - would that work? It would let us avoid the type cast. 

nit-  either way - can you encapsulate this in a tiny helper method IsSharepoint(httpClient)?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/PowerPlatformConnectorClient.cs
```diff
@@ -43,6 +43,8 @@ public class PowerPlatformConnectorClient : HttpClient
 
         public string EnvironmentId { get; set; }
 
+        public string RequestUrlPrefix { get; }
```
MikeStall, line 46: >RequestUrlPrefix [](http://example.com/codeflow?start=22&length=16)

consider making this an Init property, and then we don't need the extra ctors.

MikeStall:
>         protected internal static async Task<T> GetObject<T>(HttpClient httpClient, string message, string uri, string content, CancellationToken cancellationToken, ConnectorLogger logger = null, [CallerMemberName] string callingMethod = "")

can we remove (coming from httpClient now)

---
Refers to: src/libraries/Microsoft.PowerFx.Connectors/Tabular/Services/CdpServiceBase.cs:21 in 7a35813. [](commit_id = 7a3581307e7ed59286a19acafd632491bcf74399, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Connectors/Tabular/CdpTableResolver.cs
```diff
@@ -24,10 +24,21 @@ internal class CdpTableResolver : ICdpTableResolver
 
         private readonly HttpClient _httpClient;
 
-        private readonly string _uriPrefix;
+        [Obsolete]
+        private readonly string _uriPrefix = null;
```
MikeStall, line 28: >_uriPrefix [](http://example.com/codeflow?start=32&length=10)

can we remove and use httpClient?

src/libraries/Microsoft.PowerFx.Connectors/Tabular/CdpTableResolver.cs
```diff
@@ -103,8 +122,7 @@ public async Task<ConnectorType> ResolveTableAsync(string tableName, Cancellatio
             return connectorType;
         }
 
-        internal static bool IsSql(string uriPrefix) => uriPrefix.Contains("/sql/");
-
+        [Obsolete]
         internal static bool UseV2(string uriPrefix) => uriPrefix.Contains("/sql/") ||
                                                         uriPrefix.Contains("/zendesk/");
```
MikeStall, line 127: Since this is ultimately used to determine the endpoint, can we remove from here and push higher?

src/libraries/Microsoft.PowerFx.Connectors/Tabular/Services/CdpDataSource.cs
LucGenetier: It's not in the HttpClient base address unfortunately
I will encapsulate in IsSharepoint()

src/libraries/Microsoft.PowerFx.Connectors/Tabular/CdpTableResolver.cs
LucGenetier: I don't think so
This one is only used for backward compatibility

src/libraries/Microsoft.PowerFx.Connectors/Tabular/CdpTableResolver.cs
LucGenetier: I'm not sure where you'd put it
This is used twice in this class, twice in CdpDataSource and once in CdpTable, always in [Obsolete] methods

If we want to absorb the breaking change fully we can jsut delete all these [Obsolete] things
Just let know if is you agree/want that?

LucGenetier:
>         protected internal static async Task<T> GetObject<T>(HttpClient httpClient, string message, string uri, string content, CancellationToken cancellationToken, ConnectorLogger logger = null, [CallerMemberName] string callingMethod = "")

This method is used in CdpDataSource to convert Http responses to POCOs 
I don't understand why we'd remove it

---
In reply to: [2551745584](https://github.com/microsoft/Power-Fx/pull/2785#issuecomment-2551745584) [](http://example.com/codeflow?ancestors=2551745584)

---
Refers to: src/libraries/Microsoft.PowerFx.Connectors/Tabular/Services/CdpServiceBase.cs:21 in 7a35813. [](commit_id = 7a3581307e7ed59286a19acafd632491bcf74399, deletion_comment = False)

LucGenetier:
✅ No public API change.

MikeStall:
My concern about current PR is that it _duplicates_ the code. I'm we can solve this in a way that avoids duplicating.

PR #2786, opened 2024-12-18
Add "supportsJoin" capability

LucGenetier:


src/libraries/Microsoft.PowerFx.Core/Entities/External/TableDelegationInfo.cs
```diff
@@ -37,6 +38,9 @@ public abstract class TableDelegationInfo
         // Supports per record permission
         internal bool SupportsRecordPermission { get; init; }
 
+        [Obsolete("preview")]
+        internal bool SupportsJoinFunction { get; init; }
```
MikeStall, line 42: >internal [](http://example.com/codeflow?start=8&length=8)

This is going to need to be public so that other shims can implement it.

src/libraries/Microsoft.PowerFx.Connectors/Tabular/Capabilities/ServiceCapabilities.cs
```diff
@@ -94,10 +94,15 @@ internal sealed class ServiceCapabilities : IColumnsCapabilities
         [JsonPropertyName(CapabilityConstants.SupportsRecordPermission)]
         public readonly bool SupportsRecordPermission;
 
+        [JsonInclude]
+        [JsonPropertyName(CapabilityConstants.SupportsJoin)]
+        public readonly bool SupportsJoinFunction;
```
MikeStall, line 99: >SupportsJoinFunction [](http://example.com/codeflow?start=29&length=20)

does this imply we're getting through CDP?

MikeStall (dismissed):
:shipit:

src/libraries/Microsoft.PowerFx.Core/Entities/External/TableDelegationInfo.cs
LucGenetier: good point
I was too conservative

src/libraries/Microsoft.PowerFx.Connectors/Tabular/Capabilities/ServiceCapabilities.cs
LucGenetier: that's how a CDP connector could enable this capability
today we have none but I imagine CDP via shims could support it (not SN but another)

MikeStall: So this is a proposed update to CDP contract?

MikeStall (dismissed):
:shipit:

MikeStall:
>         internal IReadOnlyDictionary<string, ColumnCapabilitiesBase> ColumnsCapabilities { get; init; }

Curious - how does this relate to virtual GetColumnCapability below?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Entities/External/TableDelegationInfo.cs:36 in d9bec3c. [](commit_id = d9bec3c6b3529f59f27decc6f58e77d98354169a, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Connectors/Tabular/Capabilities/ServiceCapabilities.cs
LucGenetier: Very likely yes

LucGenetier:
>         internal IReadOnlyDictionary<string, ColumnCapabilitiesBase> ColumnsCapabilities { get; init; }

It's orthogonal or complementary at least
To make a JOIN work with delegation you'd need this SupportsJoinFunctions to be true and Equal to be supported on the columns in the predicate (which is usually always the case)

---
In reply to: [2551793759](https://github.com/microsoft/Power-Fx/pull/2786#issuecomment-2551793759) [](http://example.com/codeflow?ancestors=2551793759)

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Entities/External/TableDelegationInfo.cs:36 in d9bec3c. [](commit_id = d9bec3c6b3529f59f27decc6f58e77d98354169a, deletion_comment = False)

LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #2805, opened 2025-01-06
Update capabilities to support different Join types

LucGenetier:
This is to support datasources that don't necessarily support all Join types like DV which doesn't support right joins

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationCapability.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationCapability.cs
@@ -88,7 +88,10 @@ namespace Microsoft.PowerFx.Core.Functions.Delegation
                 { DelegationMetadataOperatorConstants.AsType, new DelegationCapability(AsType) },
                 { DelegationMetadataOperatorConstants.ArrayLookup, new DelegationCapability(ArrayLookup) },
                 { DelegationMetadataOperatorConstants.Distinct, new DelegationCapability(Distinct) },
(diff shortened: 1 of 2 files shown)


LucGenetier:
Public API changes: <br />❌CP0002 F:Microsoft.PowerFx.Core.Functions.Delegation.DelegationOperator.Join

src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationCapability.cs
```diff
@@ -140,10 +143,13 @@ internal struct DelegationCapability
...
+        public static readonly BigInteger JoinLeft = BigInteger.Pow(2, 48);                     //   0x1000000000000
+        public static readonly BigInteger JoinRight = BigInteger.Pow(2, 49);                    //   0x2000000000000
+        public static readonly BigInteger JoinFull = BigInteger.Pow(2, 50);                     //   0x4000000000000
 
         // Please update it as max value changes.
-        private static BigInteger maxSingleCapabilityValue = Join;
+        private static BigInteger maxSingleCapabilityValue = JoinFull;
 
         // Indicates support all functionality.
         public static BigInteger SupportsAll
```
MikeStall, line 155: This function seems highly suspicious. 
Does adding new capabilities cause this behavior to change?

LucGenetier: this constant is only used for testing the validity of the capability or in tests
It's not used in DV or PA

MikeStall (approved):

Issue #2751, opened 2024-11-25
Unify IAsyncTexlFunction


MikeStall:
We shouldn't have 5+ variations of IAysncTexlFunction plus AsyncFunctionPtr delegate.

We should be able to remove this branch in EvalVisitor and have it be a single dispatch. 
https://github.com/microsoft/Power-Fx/blob/07a817b14a7e7df2a3524fc4a748ec60089254ba/src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs#L311

Ideally, unify across:
1. builtin invocations
2. UDFs
3. Custom Functions  (ReflectionFunction) 

Broader design things to consider:
- can we unify with StandardErrorHandlingAsync? 
- can we still follow Least-privilege? At least with many interfaces, the functions are describing statically which subset of functionality they need. 

This is an engineering hygiene issue. Need to come up with a good design.

PR #2812, opened 2025-01-12
Unify IAsyncTexlFunction

MikeStall:
Address #2751.  
This unifies most of them. To scope this PR, there are subissues on 2751 for remaining issues. 

We have a plethora of IAsyncTexlFunction* interfaces. Want to consolidate on 1, and make it public.  

This 
Ttake the parameters on the various IAsyncTexlFunction* overloads and move them as properties on a new public `FunctionInvokerInfo` class.  And then have a new IAsyncTexlFunction* that takes the invoker.  And remove all the others, migrating them onto this - and fixing the various bugs/hacks along the way that would impede migration. 

Most fundamentally, this FunctionInvokerInfo must have access to interpreter state and so must live in the interpreter. (Whereas some of the IAsyncFunction* interfaces live in core). 

Successfully removes several here so we get a net reduction, and shows a path to remove the rest. 

This problem quickly touches other problems:
- **How to map from the TexlFunction to the invoker**.  This is made more complicated because we register on PowerFxConfig, but that lives in Fx.Core and can't refer to interpreter implementations. 
- **dll layering around Fx.Json**:  Some function implementations live in Fx.Json, but that specifically _does not_ depend on interpreter (because it is included by PowerApps). 
- **Standard Pipeline**: how to handle default args, common error handling scenarios, etc. Today, that's still in interpreter. Whereas we really want those checks to be in the IR and shared across front-ends (and available to designer). 
- **Split up Library.cs and class** - we have one giant class, partial over many files, containing 100s of function impls. Just give each impl its own file, similar to how we did for TexlFunction and the static declarations. 
- **lack of unit testing**: Today, most of our interpreter coverage comes from .txt files. But we should have C# unit tests on interpreter classes that provide coverage on core classes, like EvalVisitor. 

Some related prereq fixes that will directly help here:
- Can we remove runner/context from Join? and leverage LambdaValue  - which already closes over these. 
- fix Json layering problem. 
- IsType/AsType _UO shouldn't live in Json.  They should work on any UO. 
- Remove _additionalFunctions from the serviceProvider.


src/libraries/Microsoft.PowerFx.Connectors/Texl/ConnectorTexlFunction.cs
```diff
@@ -85,8 +86,11 @@ public override bool TryGetParamDescription(string paramName, out string paramDe
 
         public override bool HasSuggestionsForParam(int argumentIndex) => argumentIndex <= MaxArity;
 
-        public async Task<FormulaValue> InvokeAsync(FormulaValue[] args, IServiceProvider serviceProvider, CancellationToken cancellationToken)
+        public async Task<FormulaValue> InvokeAsync(FunctionInvokeInfo invokeInfo, CancellationToken cancellationToken)
         {
+            FormulaValue[] args = invokeInfo.Args.ToArray(); // $$$ remove ToArray
```
MikeStall, line 91: >// $$$ remove ToArray [](http://example.com/codeflow?start=61&length=21)

IReadOnlyList vs. Array. 
IRL ensures immutablility. 

We just need to plumb IRL through a bunch of files.

src/libraries/Microsoft.PowerFx.Interpreter/CustomFunction/CustomTexlFunction.cs
```diff
@@ -22,7 +23,7 @@ namespace Microsoft.PowerFx
     /// <summary>
     /// Internal adapter for adding custom functions. 
     /// </summary>
-    internal class CustomTexlFunction : TexlFunction
+    internal class CustomTexlFunction : TexlFunction, IAsyncTexlFunction999
```
MikeStall, line 26: >IAsyncTexlFunction999 [](http://example.com/codeflow?start=54&length=21)

Adding this here lets EvalVisitor avoid having to specifically check for CustomTexlFunction

src/libraries/Microsoft.PowerFx.Interpreter/Functions/FileInfoImpl.cs
```diff
@@ -6,15 +6,18 @@
 using System.Threading.Tasks;
 using Microsoft.PowerFx.Core.Functions;
 using Microsoft.PowerFx.Core.Texl.Builtins;
+using Microsoft.PowerFx.Functions;
 using Microsoft.PowerFx.Interpreter;
 using Microsoft.PowerFx.Types;
 
 namespace Microsoft.PowerFx
 {
-    internal class FileInfoFunctionImpl : FileInfoFunction, IAsyncTexlFunction3
```
MikeStall, line 14: >IAsyncTexlFunction3 [](http://example.com/codeflow?start=60&length=19)

This is just demonstrating an example of converting IAsyncTexlFunction3.  The other uses look similarly easy.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/IAsyncTexlFunction999.cs
```diff
@@ -0,0 +1,62 @@
...
+using Microsoft.PowerFx.Core.IR;
+using Microsoft.PowerFx.Interpreter;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.PowerFx.Functions
+{
+    /// <summary>
+    /// Invoker to execute a function via the interpreter. 
+    /// </summary>
+    public interface IAsyncTexlFunction999
```
MikeStall, line 18: >IAsyncTexlFunction999 [](http://example.com/codeflow?start=21&length=21)

The new interface. 
We will find a better name.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/IAsyncTexlFunction999.cs
```diff
@@ -0,0 +1,62 @@
...
+#if true
+        // Can we find a way to get rid of these ones? Keep internal to limit usage.
+
+        // IrContext has a mutability flag. 
+        internal IRContext IRContext { get; set; }
+
+        // Get rid of these... Capture in them in the closure of a lambdaValue.
+        internal EvalVisitor Runner { get; init; }
+
+        internal EvalVisitorContext Context { get; init; }
```
MikeStall, line 59: it would be really nice to get rid of these- they ought to already be captured in the Lambda. 

Need to fix join.

src/libraries/Microsoft.PowerFx.Interpreter/Values/LambdaFormulaValue.cs
```diff
@@ -34,6 +34,19 @@ public async ValueTask<FormulaValue> EvalAsync()
             return await EvalInRowScopeAsync(_context).ConfigureAwait(false);
         }
 
+        public async ValueTask<FormulaValue> EvalInRowScopeAsync(RecordValue rowScope)
```
MikeStall, line 37: >EvalInRowScopeAsync [](http://example.com/codeflow?start=45&length=19)

Can we do something like this?  To remove EvalVisitor/Context from the  FunctionInvokeInfo.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -346,63 +460,19 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
...
-                        throw new Exception("Something went wrong. UDF stack values didn't match.");
-                    }
+                    result = CommonErrors.NotYetImplementedFunctionError(node.IRContext, func.Name);
                 }
 
-#pragma warning disable CS0618 // Type or member is obsolete
-
-                // This is a temporary solution to release the Join function for host that want to use it.
-                else if (func is IAsyncTexlFunctionJoin asyncJoin)
+                if (invoker is Adapter1) // $$$ we should remove this line, so we apply this check to all impls.
```
MikeStall: >if (invoker is Adapter1)  [](http://example.com/codeflow?start=16&length=25)

This is interesting - 
before, we only did this check on certain functions. 

But we ought to do it on all functions - and are in fact failing it on toher cases (hence the need to preserver the check on 468) 

A common invoke pipeline enforces this kind of consistency,.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -300,38 +391,61 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
...
             IReadOnlyDictionary<TexlFunction, IAsyncTexlFunction> extraFunctions = _services.GetService<IReadOnlyDictionary<TexlFunction, IAsyncTexlFunction>>();
 
             try
             {
-                if (func is IAsyncTexlFunction asyncFunc || extraFunctions?.TryGetValue(func, out asyncFunc) == true)
+                var invoker = GetInvoker(func);
+
+                // $$$ New standard invoke path. Make everything go through here. 
+                // Eventually collapse all cases to this. 
+                if (invoker != null)
```
MikeStall, line 414: This is the key change. 
Trying to take this complicated dispatch switch-case  and reduce it down to a single case. 

This iteration succesfully removes:
- IAsyncTexlFunction2
- IAsyncConnectorTexlFunction
- CustomTexlFunction
- IAsyncTexlFunctionJoin

And has adapters for
- UserDefinedFunction
- AsyncFunctionPtr

And a clear path to remove remaining ones.

LucGenetier:
✅ No public API change.

MikeStall:
>                         services2.AddService(new Canceller(CheckCancel));

this is hackery - we should find a way to remove it. 
Thii s the kind of thing that standardizing will help us avoid in the future. 

---
Refers to: src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs:458 in 7b1c96d. [](commit_id = 7b1c96db90b416d37ba7d2be09f6997126d19e77, deletion_comment = False)

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -265,6 +265,100 @@ private async Task<FormulaValue> TryHandleSetProperty(CallNode node, EvalVisitor
...
+            if (FunctionImplementations.TryGetValue(func, out AsyncFunctionPtr ptr))
+            {
+                return new AsyncFunctionPtrAdapter(ptr);
+            }
+
+            return null;
+        }
+
+        // Adapter for AsyncFunctionPtr to common invoker interface.
+        private class AsyncFunctionPtrAdapter : IFunctionInvoker
```
anderson-joyle, line 291: >AsyncFunctionPtrAdapter [](http://example.com/codeflow?start=22&length=23)

nit: can we separate these new classes into a new file?

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -300,40 +394,59 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
...
                     args[i] = new LambdaFormulaValue(node.IRContext, child, this, context);
                 }
             }
 
             var childContext = context.SymbolContext.WithScope(node.Scope);
 
             FormulaValue result;
+
+            // Remove this: https://github.com/microsoft/Power-Fx/issues/2821
             IReadOnlyDictionary<TexlFunction, IAsyncTexlFunction> extraFunctions = _services.GetService<IReadOnlyDictionary<TexlFunction, IAsyncTexlFunction>>();
```
anderson-joyle, line 408: >extraFunctions = _services.GetService<IReadOnlyDictionary<TexlFunction, IAsyncTexlFunction>>(); [](http://example.com/codeflow?start=66&length=95)

Were you trying to address this problem in this PR?

src/libraries/Microsoft.PowerFx.Interpreter/Functions/IAsyncTexlFunctionJoin.cs
```diff
@@ -1,18 +0,0 @@
...
-using System.Threading;
-using System.Threading.Tasks;
-using Microsoft.PowerFx.Core.IR;
-using Microsoft.PowerFx.Functions;
-using Microsoft.PowerFx.Types;
-
-namespace Microsoft.PowerFx.Core.Functions
-{
-    [Obsolete("This interface is temporary and its meant only to support Join function. It will be soon removed.")]
-    internal interface IAsyncTexlFunctionJoin
```
anderson-joyle, line 14: >IAsyncTexlFunctionJoin [](http://example.com/codeflow?start=23&length=22)

You wont be missed :joy:

anderson-joyle (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -300,40 +394,59 @@ public override async ValueTask<FormulaValue> Visit(CallNode node, EvalVisitorCo
...
+                // Eventually collapse all cases to this. 
+                if (invoker != null)
                 {
-                    result = await asyncFunc.InvokeAsync(args, _cancellationToken).ConfigureAwait(false);
+                    var invokeInfo = new FunctionInvokeInfo
+                    {
+                        Args = args,
+                        FunctionServices = _services,
+                        Runner = this,
+                        Context = context.IncrementStackDepthCounter(childContext),
```
adithyaselv, line 423: >IncrementStackDepthCounter [](http://example.com/codeflow?start=42&length=26)

wondering how we got away not increasing the stack depth here. i see we increase it when we visit the arg above which makes sense ,

this is just one step in -> invoke and out right ?

do we have test around max stack depth ?

MikeStall: Calling IncrementStackDepthCounter will test max depth and throw at the limit. 

This is another benefit of unifying the invoke paths - we ensure consistent handling here across all paths.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
MikeStall: No - that was out of scope, but I filed #2821 to track it.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
MikeStall: these are nested private classes - which need that to accessing private state on this class. And only used within this file.  
We could consider making a partial class and split to other files.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -265,6 +265,100 @@ private async Task<FormulaValue> TryHandleSetProperty(CallNode node, EvalVisitor
...
+            {
+                return new UserDefinedFunctionAdapter(userDefinedFunc);
+            }
+
+            if (FunctionImplementations.TryGetValue(func, out AsyncFunctionPtr ptr))
+            {
+                return new AsyncFunctionPtrAdapter(ptr);
+            }
+
+            return null;
```
LucGenetier, line 287: >return null; [](http://example.com/codeflow?start=12&length=12)

when is this happening?

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
```diff
@@ -265,6 +265,100 @@ private async Task<FormulaValue> TryHandleSetProperty(CallNode node, EvalVisitor
...
+                    result = await irnode.Accept(evalVisitor, context).ConfigureAwait(false);
+                }
+                finally
+                {
+                    framePop = udfStack.Pop();
+                }
+
+                if (frame != framePop)
+                {
+                    throw new Exception("Something went wrong. UDF stack values didn't match.");
```
LucGenetier, line 356: >Exception [](http://example.com/codeflow?start=30&length=9)

Is there any better Exception we could throw here?
We could add in the message some more info

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
```diff
@@ -15,10 +15,12 @@
 namespace Microsoft.PowerFx.Core.Texl.Builtins
 {
     // Patch(dataSource:*[], Record, Updates1, Updates2,…)
-    internal class PatchImpl : PatchFunction, IAsyncTexlFunction3
+    internal class PatchImpl : PatchFunction, IFunctionInvoker
     {
-        public async Task<FormulaValue> InvokeAsync(FormulaType irContext, FormulaValue[] args, CancellationToken cancellationToken)
+        public async Task<FormulaValue> InvokeAsync(FunctionInvokeInfo invokeInfo, CancellationToken cancellationToken)
```
LucGenetier, line 20: >cancellationToken [](http://example.com/codeflow?start=101&length=17)

unrelated to this PR but we should test if we shall throw before calling PatchAsync at line 48 and AppendAsync at line 53

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
```diff
@@ -57,10 +59,12 @@ public async Task<FormulaValue> InvokeAsync(FormulaType irContext, FormulaValue[
 
     // If arg1 is pure PFx record, it will return a runtime not supported error.
     // Patch(DS, record_with_keys_and_updates)
-    internal class PatchSingleRecordImpl : PatchSingleRecordFunction, IAsyncTexlFunction3
+    internal class PatchSingleRecordImpl : PatchSingleRecordFunction, IFunctionInvoker
     {
-        public async Task<FormulaValue> InvokeAsync(FormulaType irContext, FormulaValue[] args, CancellationToken cancellationToken)
+        public async Task<FormulaValue> InvokeAsync(FunctionInvokeInfo invokeInfo, CancellationToken cancellationToken)
```
LucGenetier, line 64: >cancellationToken [](http://example.com/codeflow?start=101&length=17)

Same here

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
```diff
@@ -1575,15 +1581,13 @@ public static NamedLambda[] Parse(FormulaValue[] args)
...
 
-#pragma warning disable CS0618 // Type or member is obsolete
-    internal class JoinImpl : JoinFunction, IAsyncTexlFunctionJoin
-#pragma warning restore CS0618 // Type or member is obsolete
+    internal class JoinImpl : JoinFunction, IFunctionInvoker
     {
         public override Type DeclarationType => typeof(JoinFunction);
 
-        public async Task<FormulaValue> InvokeAsync(EvalVisitor runner, EvalVisitorContext context, IRContext irContext, FormulaValue[] args)
+        public async Task<FormulaValue> InvokeAsync(FunctionInvokeInfo invokeInfo, CancellationToken cancellationToken)
```
LucGenetier, line 1588: >cancellationToken [](http://example.com/codeflow?start=101&length=17)

this is unused - should be tested & passed along

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
MikeStall: All the cases we haven't yet converted. See below where it is called and we handle the null.

src/libraries/Microsoft.PowerFx.Interpreter/EvalVisitor.cs
MikeStall: This is just moving existing code so I didn't change the logic.  cc @adithyaselv 

Note that this condition should only happen with a bug in the interpreter (and not triggered by user code).

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryMutation.cs
MikeStall: no harm, but low value. The caller has already checked this, and there's no looping or network calls here.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs
MikeStall: good catch. 
Existing code. JoinTables is checking cancellation via EvalVisitor.CheckCancel.

src/libraries/Microsoft.PowerFx.Interpreter/Functions/FunctionInvokeInfo.cs
```diff
@@ -0,0 +1,79 @@
...
+        // Since every method impl would get this, consider add other useful operators on here, like:
+        //  - CheckCancel?
+        //  - error checks, blank checks. 
+
+        // Can we find a way to get rid of these ones? Keep internal to limit usage.
+        // Keep internal until we kind a way to remove. 
+        #region Remove these 
+
+        // IrContext has a mutability flag. 
+        internal IRContext IRContext { get; set; }
```
CarlosFigueiraMSFT, line 47: >set [](http://example.com/codeflow?start=44&length=3)

What is the case where IRContext needs to be set in this class?

adithyaselv (dismissed):
:shipit:

jas-valgotar (dismissed):
:shipit:

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/EvalVisitorTests.cs
```diff
@@ -0,0 +1,50 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Text;
+using Xunit;
+
+namespace Microsoft.PowerFx.Interpreter.Tests
+{
+    // This file should get 100% code-coverage on EvalVistor.
```
CarlosFigueiraMSFT, line 9: >should [](http://example.com/codeflow?start=17&length=6)

"will eventually"

src/libraries/Microsoft.PowerFx.Interpreter/Functions/FunctionInvokeInfo.cs
MikeStall: ah, you mean 'make this init, not set'.

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2814, opened 2025-01-13
Avoid NullRef in CdpDelegationInfo

LucGenetier:


src/libraries/Microsoft.PowerFx.Connectors/Public/CdpDelegationInfo.cs
```diff
@@ -10,7 +10,7 @@ internal class CdpDelegationInfo : TableDelegationInfo
     {
         public override ColumnCapabilitiesDefinition GetColumnCapability(string fieldName)
         {
-            if (ColumnsCapabilities.TryGetValue(fieldName, out ColumnCapabilitiesBase columnCapabilitiesBase))
+            if (ColumnsCapabilities.TryGetValue(fieldName, out ColumnCapabilitiesBase columnCapabilitiesBase) && columnCapabilitiesBase != null)
```
CarlosFigueiraMSFT, line 13: Is adding a `null` value in the ColumnsCapabilities dictionary a valid scenario? It doesn't feel like it should be. If that is not the case, then we should block this in the addition, not in the retrieval. If that's a valid scenario, then this fix looks good.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

MikeStall:
>             Dictionary<string, Core.Entities.ColumnCapabilitiesBase> columnCapabilities = serviceCapabilities?._columnsCapabilities?.ToDictionary(

It doesn't look like this could ever produce a dict with null values.  (although to be safer, the 'as' operators ought to be replaced with static casts) 

So is the fix on 179 sufficient to catch the null ref?




---
Refers to: src/libraries/Microsoft.PowerFx.Connectors/Tabular/Capabilities/ServiceCapabilities.cs:167 in 489dfdf. [](commit_id = 489dfdf618f9c8e7c4e118e9a94d5e8634c14788, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Connectors/Public/CdpDelegationInfo.cs
LucGenetier: I think having missing capabilities is allowed and we shouldn't throw
This is clearly suboptimal when you don't have capabilities (as this will prevent us from delegating) but I don't think we're hiding a bug

LucGenetier: Let me remove that part

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2815, opened 2025-01-13
Add UseDefaultBodyNameForSinglePropertyObject

LucGenetier:
In Power Apps, when a body parameter is used it's flattened and we create one paramaeter for each
body object property. With that logic each parameter name will be the object property name.

When UseDefaultBodyNameForSinglePropertyObject is set, we will use the real body name specified 
in the swagger instead of the property name of the object, provided there is only one property.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorSettings.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorSettings.cs
@@ -94,6 +94,14 @@ namespace Microsoft.PowerFx.Connectors
         /// </summary>
         public bool ReturnEnumsAsPrimitive { get; init; } = false;
 
+        /// <summary>
+        /// In Power Apps, when a body parameter is used it's flattened and we create one paramaeter for each
+        /// body object property. With that logic each parameter name will be the object property name.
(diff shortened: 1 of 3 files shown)


src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorSettings.cs
```diff
@@ -92,7 +92,15 @@ public bool ExposeInternalParamsWithoutDefaultValue
         /// This flag will force all enums to be returns as FormulaType.String or FormulaType.Decimal regardless of x-ms-enum-*.
         /// This flag is only in effect when SupportXMsEnumValues is true.
         /// </summary>
-        public bool ReturnEnumsAsPrimitive { get; init; } = false;
+        public bool ReturnEnumsAsPrimitive { get; init; } = false;
+
+        /// <summary>
+        /// In Power Apps, when a body parameter is used it's flattened and we create one paramaeter for each
```
MikeStall, line 98: >paramaeter [](http://example.com/codeflow?start=90&length=10)

nit - typo

MikeStall (approved):
:shipit:

MikeStall (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2816, opened 2025-01-16
Fix delegation check on UDF for datasource type

yuchenglin:
## Probelm
UDF with direct assigned datasource type such as `UDF1():Accounts = Accounts` is always calculated as non-delegable.
There is no issue for `UDF2():Accounts = Filter(Account, ...)` since it would go through override validation in Filter class, where check data source metadata directly. UDF gets IsDelegatable from Top which is Filter().

## Option 1 (current)
Handle in overridden `IsServerDelegatable` of UserDefinedFunction. It will go check data source metadata directly if it's FirstNameNode and haven't been processed.

## Option 2 
The delegable check on FirstNameNode during binding validates the data is type of `IExternalDelegatableSymbol `, but from this scenario, `IExternalCdsDataSource` is not implanted this interface. So always returns IsServerDelegatable false.
I decided to add the implantation to align with `IsPagable`.

https://github.com/microsoft/Power-Fx/blob/0fa19687742caa771c4a8f0716f07005497bec49/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs#L688C13-L688C75

https://dev.azure.com/msazure/OneAgile/_workitems/edit/29998102

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
@@ -52,17 +52,7 @@ namespace Microsoft.PowerFx.Core.Functions
             Contracts.AssertValue(binding);
             Contracts.Assert(binding.GetInfo(callNode).Function is UserDefinedFunction udf && udf.Binding != null);
 
-            if (binding.DelegationHintProvider?.TryGetWarning(callNode, this, out var warning) ?? false)
-            {
-                SuggestDelegationHint(callNode, binding);
-            }
-
-            if (TryGetDataSource(callNode, binding, out var dataSource) && dataSource.IsDelegatable)
-            {
-                return true;
-            }
-
-            return IsDelegatable;
+            return base.IsServerDelegatable(callNode, binding) || IsDelegatable;
         }
 
         public override bool SupportsParamCoercion => true;
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
@@ -52,7 +52,17 @@ namespace Microsoft.PowerFx.Core.Functions
             Contracts.AssertValue(binding);
             Contracts.Assert(binding.GetInfo(callNode).Function is UserDefinedFunction udf && udf.Binding != null);
 
-            return base.IsServerDelegatable(callNode, binding) || IsDelegatable;
+            if (base.IsServerDelegatable(callNode, binding) || IsDelegatable)
+            {
+                return true;
+            }
+
+            if (_binding.Top.Kind == NodeKind.FirstName && ArgValidators.DelegatableDataSourceInfoValidator.TryGetValidValue(_binding.Top, _binding, out _))
+            {
+                return true;
+            }
+
+            return false;
         }
 
         public override bool SupportsParamCoercion => true;
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
@@ -34,7 +34,7 @@ namespace Microsoft.PowerFx.Core.Functions
     /// This includings the binding (and hence IR for evaluation) - 
     /// This is conceptually immutable after initialization - if the body or signature changes, you need to create a new instance.
     /// </summary>
-    internal class UserDefinedFunction : TexlFunction, IExternalPageableSymbol, IExternalDelegatableSymbol
+    internal class UserDefinedFunction : TexlFunction, IExternalPageableSymbol
     {
         private readonly bool _isImperative;
         private readonly IEnumerable<UDFArg> _args;
@@ -44,15 +44,13 @@ namespace Microsoft.PowerFx.Core.Functions
 
         public bool IsPageable => _binding.IsPageable(_binding.Top);
 
-        public bool IsDelegatable => _binding.IsDelegatable(_binding.Top);
-
         public override bool IsServerDelegatable(CallNode callNode, TexlBinding binding)
         {
             Contracts.AssertValue(callNode);
             Contracts.AssertValue(binding);
             Contracts.Assert(binding.GetInfo(callNode).Function is UserDefinedFunction udf && udf.Binding != null);
 
-            if (base.IsServerDelegatable(callNode, binding) || IsDelegatable)
+            if (base.IsServerDelegatable(callNode, binding) || _binding.IsDelegatable(_binding.Top))
             {
                 return true;
             }
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
@@ -52,7 +52,17 @@ namespace Microsoft.PowerFx.Core.Functions
             Contracts.AssertValue(binding);
             Contracts.Assert(binding.GetInfo(callNode).Function is UserDefinedFunction udf && udf.Binding != null);
 
-            return base.IsServerDelegatable(callNode, binding) || IsDelegatable;
+            if (binding.DelegationHintProvider?.TryGetWarning(callNode, this, out var warning) ?? false)
+            {
+                SuggestDelegationHint(callNode, binding);
+            }
+
+            if (TryGetDataSource(callNode, binding, out var dataSource) && dataSource.IsDelegatable)
+            {
+                return true;
+            }
+
+            return IsDelegatable;
         }
 
         public override bool SupportsParamCoercion => true;
--- a/src/tests/Microsoft.PowerFx.Core.Tests.Shared/AssociatedDataSourcesTests/TestDelegationValidation.cs
+++ b/src/tests/Microsoft.PowerFx.Core.Tests.Shared/AssociatedDataSourcesTests/TestDelegationValidation.cs
@@ -3,12 +3,16 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Text;
 using Microsoft.PowerFx.Core.Entities.QueryOptions;
 using Microsoft.PowerFx.Core.Errors;
+using Microsoft.PowerFx.Core.Functions.Delegation;
 using Microsoft.PowerFx.Core.Tests.Helpers;
 using Microsoft.PowerFx.Core.Texl;
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Core.Utils;
 using Microsoft.PowerFx.Types;
 using Xunit;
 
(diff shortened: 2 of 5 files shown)


yuchenglin:
> @yuchenglin please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree company="Microsoft"

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/RecalcEngineTests.cs
```diff
@@ -793,30 +793,32 @@ public void DelegableUDFTest()
...
             // Imperative function is not delegable
             Assert.True(func.IsAsync);
             Assert.True(!func.IsDelegatable);
 
             // Binding fails for recursive definitions and hence function is not added.
             Assert.False(recalcEngine.AddUserDefinedFunction("E():Void = { E(); };", CultureInfo.InvariantCulture, symbolTable: recalcEngine.EngineSymbols, allowSideEffects: true).IsSuccess);
+            
+            func = recalcEngine.Functions.WithName("F").First() as UserDefinedFunction;
+            Assert.True(func.IsAsync);
+            Assert.True(func.IsDelegatable);
```
Shpakh-MSFT, line 821: Can you add a testcase where IsDelegatable is false?

Shpakh-MSFT: For example,
 Filter(MyDataSource, Value < 10)

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -50,9 +50,19 @@ public override bool IsServerDelegatable(CallNode callNode, TexlBinding binding)
...
+
+            if (binding.DelegationHintProvider?.TryGetWarning(callNode, this, out var warning) ?? false)
+            {
+                SuggestDelegationHint(callNode, binding);
+            }
+
+            if (TryGetDataSource(callNode, binding, out var dataSource) && dataSource.IsDelegatable)
+            {
+                return true;
+            }
```
Shpakh-MSFT, line 63: Why do we need this? The binder doesn't mark the Top node as delegable?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -50,9 +50,19 @@ public override bool IsServerDelegatable(CallNode callNode, TexlBinding binding)
...
             Contracts.AssertValue(callNode);
             Contracts.AssertValue(binding);
-            Contracts.Assert(binding.GetInfo(callNode).Function is UserDefinedFunction udf && udf.Binding != null);
-
-            return base.IsServerDelegatable(callNode, binding) || IsDelegatable;
+            Contracts.Assert(binding.GetInfo(callNode).Function is UserDefinedFunction udf && udf.Binding != null);
+
+            if (binding.DelegationHintProvider?.TryGetWarning(callNode, this, out var warning) ?? false)
+            {
+                SuggestDelegationHint(callNode, binding);
```
Shpakh-MSFT, line 57: >SuggestDelegationHint(callNode, binding); [](http://example.com/codeflow?start=16&length=41)

Should return false here. Delegation warning means that you can't delegate the call.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
yuchenglin: Binder doesn't mark Top as delegable. The setter `CheckAndMarkAsDelegatable(CallNode node)` in Binder ends up checking this method to see if marking as delegable, and it is false since setter and getter depend on the same field (the binary bit array).

adithyaselv: that's strange, we dont mark FirstNameNodes as delegable if they can be delegated? likely is the case , is it still possible to fix at UDF bind ?

yuchenglin: the problem is interface `IExternalDelegatableSymbol` doesn't implement on some data source types. Trying to fix and test in this way

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/RecalcEngineTests.cs
yuchenglin: This seems to be another bug. It returns IsDelegatable = true

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
yuchenglin: removed change here

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/RecalcEngineTests.cs
yuchenglin: https://github.com/microsoft/Power-Fx/blame/0fa19687742caa771c4a8f0716f07005497bec49/src/libraries/Microsoft.PowerFx.Core/Functions/Delegation/DelegationStrategies/OpDelegationStrategy.cs#L281-L287C14

This is by design not checking on FirstNameNode.

yuchenglin: Added a mock non delegatable data source for unit tests

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
yuchenglin: Please take a look at the new solution

adithyaselv:
can you please add a test like this . 

```
N = DelegableDataSource;

UDF(): DelegableDataSource = N;

Filter(UDF(), "ColumnName" <> "")

```
this doesn't work today for the same reasons discussed offline

yuchenglin:
> can you please add a test like this .
> 
> ```
> N = DelegableDataSource;
> 
> UDF(): DelegableDataSource = N;
> 
> N2 = Filter(UDF(), "ColumnName" <> "")
> ```
> 
> this doesn't work today for the same reasons discussed offline

Synced offline. This can be better testing at Client

yuchenglin:
With option 2, 
1. Client tests (code gen) are failing since it will add `appendQueryOptionsAsync` to original query. From the query itself, the behavior looks isn't changed. 
https://msazure.visualstudio.com/OneAgile/_build/results?buildId=112823118&view=ms.vss-test-web.build-test-results-tab

2.  We got a report from Playground. There are some class out there depends on this interface
```
System.MissingMethodException: Method not found: 'Boolean Microsoft.PowerFx.Core.Entities.IExternalDataSource.get_IsDelegatable()'.
at Microsoft.PowerFx.Dataverse.DelegationIRVisitor.Visit(ResolvedObjectNode node, Context context)
```

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -52,7 +52,17 @@ public override bool IsServerDelegatable(CallNode callNode, TexlBinding binding)
             Contracts.AssertValue(binding);
             Contracts.Assert(binding.GetInfo(callNode).Function is UserDefinedFunction udf && udf.Binding != null);
 
-            return base.IsServerDelegatable(callNode, binding) || IsDelegatable;
+            if (base.IsServerDelegatable(callNode, binding) || IsDelegatable)
```
anhtruong-MSFT, line 55: nit: can we just have 1 return like 
```suggestion
   return a || b || c
```

adithyaselv:
>         public bool IsDelegatable => _binding.IsDelegatable(_binding.Top);

is it possible to just get rid of this property and directly call _binding.IsDelegatable(_binding.Top) below .

this pattern makes sense for `FirstNameNodes` - refer binder .  but for UDF binder can just use IsServerDelegatable method to determine delegability . i dont think this is used anywhere , can you check ?

probably the same for pageable too.

this pattern may not be needed here at all . can you investigate ?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs:47 in 6d3adac. [](commit_id = 6d3adac85ad8cc3f44ddcd0d25a7054a3a2596b6, deletion_comment = False)

yuchenglin:
> > ```
> >     public bool IsDelegatable => _binding.IsDelegatable(_binding.Top);
> > ```
> 
> is it possible to just get rid of this property and directly call _binding.IsDelegatable(_binding.Top) below .
> 
> this pattern makes sense for `FirstNameNodes` - refer binder . but for UDF binder can just use IsServerDelegatable method to determine delegability . i dont think this is used anywhere , can you check ?
> 
> this pattern may not be needed here at all . can you investigate ?
> 
> Refers to: src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs:47 in [6d3adac](https://github.com/microsoft/Power-Fx/commit/6d3adac85ad8cc3f44ddcd0d25a7054a3a2596b6). [](commit_id = [6d3adac](https://github.com/microsoft/Power-Fx/commit/6d3adac85ad8cc3f44ddcd0d25a7054a3a2596b6), deletion_comment = False)

Looks like `IsServerDelegatable` is the common entry point for functions. I've updated the code.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/AssociatedDataSourcesTests/TestDelegationValidation.cs
```diff
@@ -73,20 +77,33 @@ public void TestDelegableExpressions_ColumnNamesAsLiteralStrings(string expressi
...
+        [Theory]
+        [InlineData("UDF1()", "UDF1():Accounts = Accounts;", true)]
+        [InlineData("Filter(UDF1(), \"name\" <> \"\")", "UDF1():Accounts = Accounts;", true)]
+        public void TestDelegableExpressions_UserDfeinedFunction(string expression, string script, bool isDelegable)
+        {
+            TestDelegableExpressions(Features.PowerFxV1, expression, isDelegable, script);
         }
 
-        private void TestDelegableExpressions(Features features, string expression, bool isDelegable)
+        private void TestDelegableExpressions(Features features, string expression, bool isDelegable, string script = null)
```
adithyaselv, line 90: nit : udf script

src/tests/Microsoft.PowerFx.Core.Tests.Shared/AssociatedDataSourcesTests/TestDelegationValidation.cs
```diff
@@ -73,20 +77,33 @@ public void TestDelegableExpressions_ColumnNamesAsLiteralStrings(string expressi
...
         {
             var symbolTable = new DelegatableSymbolTable();
-            symbolTable.AddEntity(new AccountsEntity());
-            symbolTable.AddVariable("varString", FormulaType.String);
```
adithyaselv, line 82: >symbolTable.AddVariable("varString", FormulaType.String); [](http://example.com/codeflow?start=12&length=57)

guess this is accidentally removed . i see varString being used above

LucGenetier:
✅ No public API change.

adithyaselv (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2823, opened 2025-01-20
Use item dynamic properties special handling for body parameter

LucGenetier:
Add new connector setting _UseItemDynamicPropertiesSpecialHandling_
Remove _UseDefaultBodyNameForSinglePropertyObject_ as it was not fixing the issue as expected
Special case handling when
- body name is 'item'
- body inner object is 'dynamicProperties'
- there is only one property in inner object
In that base the body will be fully flattened and we will retain the 'body' name for the parameter.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorSettings.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorSettings.cs
@@ -96,7 +96,7 @@ namespace Microsoft.PowerFx.Connectors
 
         /// <summary>
         /// This flag enables some special handling for the body parameter, when
-        /// - body nams is 'item'
+        /// - body name is 'item'
         /// - body inner object is 'dynamicProperties'
         /// - there is only one property in inner object
         /// In that base the body will be fully flattened and we will retain the 'body' name for the parameter.


src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
```diff
@@ -1458,14 +1459,21 @@ private ConnectorParameterInternals Initialize()
...
-                                    string bodyPropertyName = bodyProperty.Key;
-                                    bool bodyPropertyRequired = bodySchema.Required.Contains(bodyPropertyName);
-                                    bool bodyPropertyHiddenRequired = false;
-
-                                    if (ConnectorSettings.UseDefaultBodyNameForSinglePropertyObject && bodySchema.Properties.Count == 1)
+                                    string bodyPropertyName = bodyProperty.Key;                              
+                                    bool bodyPropertyHiddenRequired = false;                                    
+                                    
+                                    // Power Apps has a special handling for the body in this case
+                                    // where it doesn't follow the swagger file
```
MikeStall, line 1466: This is key point.

MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
LucGenetier: Yes

LucGenetier:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Connectors.ConnectorSettings.get_UseDefaultBodyNameForSinglePropertyObject<br />❌CP0002 M:Microsoft.PowerFx.Connectors.ConnectorSettings.set_UseDefaultBodyNameForSinglePropertyObject(System.Boolean)

src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
```diff
@@ -1458,14 +1459,21 @@ private ConnectorParameterInternals Initialize()
...
+                                        bodyName == "item" &&
+                                        bodyPropertyName == "dynamicProperties" &&
+                                        bodySchema.Properties.Count == 1)
                                     {
                                         bodyPropertyName = bodyName;
-                                    }
+                                        specialBodyHandling = true;
+                                    }
+                                    
+                                    bool bodyPropertyRequired = bodySchema.Required.Contains(bodyPropertyName) || (ConnectorSettings.UseItemDynamicPropertiesSpecialHandling && requestBody.Required);
```
MikeStall, line 1476: This has bug, fixed in #2834 . Should use _specialBodyHandling   computed here.

PR #2824, opened 2025-01-22
Update REPL multiline logic and improve identifier parse error messages

gregli-msft:
The REPL attempts to determine if a line continuation is needed to create a valid formula.  Today, it does this by using the Power Fx parser and attempting to interpret the error messages, which turns out doesn't work so well.  It also isn't very smart about operators and doesn't handle named formulas or UDFs.

For example, these would not continue:
- `IsMatch( "asdf", "(?x)`       long regex best started on the next line
- `NamedFormula =`     long named formula best started on the next line
- `Func( x: Number ): Text =`   long udf, best started on the next line
- `1 + ` long operand, best started on the next line

The new logic is written by hand specifically to answer this continuation question and does a much better job.  It is always a best guess and only determines when to send the formula to the parser/eval loop - annoyance either being too conservative or liberal is all that this controls.  Even if a continuation is detected where there should not be one, the user can always enter a blank line to conclude the continuation. Not detecting a continuation will be annoying and should result an issue being raised, but the user can always remove the extra newline.  

Yes, we'll need to add logic in two places if the parser changes significantly, although that is not anticipated anytime soon. If this becomes a burden long term, we can look at enhancing the mainline parser to detect continuation situations, but at this point that feels like the REPL tail wagging the Power Fx dog.

In the process of learning how identifiers were being parsed, I found the error messages were very poor for an identifier parser error.  I didn't change the logic, just updated the messages and added some more tests.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs
```diff
@@ -1429,13 +1429,13 @@ private Identifier ParseIdentifier(Token at = null)
             if (_curs.TidCur == TokKind.Ident)
```
MikeStall, line 1429: This is the most important part to scrutinize..

MikeStall (approved):
:shipit:

PR #2831, opened 2025-01-27
Fix issue when using Unicode characters in CDP datasets

LucGenetier:
When using Unicode characters, we were seeing HttpRequestException: Request headers must contain only ASCII characters.


MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #2833, opened 2025-01-28
Add support for x-ms-property-entity-type

LucGenetier:
Add support for x-ms-property-entity-type extension
Expose it on all parameters, when available
Different from display name as it's a non-localized extension


riosn-n (approved):

LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #2834, opened 2025-01-28
Fix special body param required condition

AndreasVolkmann:
The condition for the body param required bool was wrong, which resulted in different properties being required when `UseItemDynamicPropertiesSpecialHandling` was on vs off.

See test case which was failing before the fix.


MikeStall (approved):
:shipit:

src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
```diff
@@ -1480,7 +1480,7 @@ private ConnectorParameterInternals Initialize()
                                         _specialBodyHandling = true;
```
MikeStall, line 1480: >_specialBodyHandling [](http://example.com/codeflow?start=40&length=20)

Change is to pick up this value

PR #2835, opened 2025-01-29
Adding support to Unicode characters for CDP table

skg-marimuthu:
Tested locally to Preview and Search a CDP table
Fix to resolve Request headers must contain only ASCII characters

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/Tabular/Services/CdpTable.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/Tabular/Services/CdpTable.cs
@@ -110,12 +110,11 @@ namespace Microsoft.PowerFx.Connectors
 
             string queryParams = (odataParameters != null) ? "&" + odataParameters.ToQueryString() : string.Empty;
 
-            Uri uri = new Uri(
-                   (_uriPrefix ?? string.Empty) +
+            var uri = (_uriPrefix ?? string.Empty) +
                    (CdpTableResolver.UseV2(_uriPrefix) ? "/v2" : string.Empty) +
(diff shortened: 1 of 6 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/Tabular/Services/CdpTable.cs
```diff
@@ -110,12 +110,11 @@ protected override async Task<IReadOnlyCollection<DValue<RecordValue>>> GetItems
 
             string queryParams = (odataParameters != null) ? "&" + odataParameters.ToQueryString() : string.Empty;
 
-            Uri uri = new Uri(
-                   (_uriPrefix ?? string.Empty) +
+            var uri = (_uriPrefix ?? string.Empty) +
                    (CdpTableResolver.UseV2(_uriPrefix) ? "/v2" : string.Empty) +
-                   $"/datasets/{(DatasetMetadata.IsDoubleEncoding ? DoubleEncode(DatasetName) : SingleEncode(DatasetName))}/tables/{Uri.EscapeDataString(TableName)}/items?api-version=2015-09-01" + queryParams, UriKind.Relative);
```
MikeStall, line 116: >. [](http://example.com/codeflow?start=217&length=1)

why this change?

MikeStall:
This is related to #2831

src/libraries/Microsoft.PowerFx.Connectors/Tabular/Services/CdpTable.cs
skg-marimuthu: encoded characters for query parameters which has Unicode chars like Biométricos then the encoded Table name/query params value lost on calling uri.ToString() function and hence fixing the same issue which will construct the Url as below with the above fix

items?api-version=2015-09-01&$filter=%28Name%20eq%20%27Biom%C3%A9tricos%27%29&$select=Name&$top=1000

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #2836, opened 2025-01-30
Reenable [ThreadSafeImmutable] tests.

MikeStall:
These tests had been disabled and were not actually asserting on failures. 
Remove some #ifdefs !NET7_0_OR_GREATER

No real product change other than adding some more annotations. 

See #1519.

Add IRNode to list.


src/tests/Microsoft.PowerFx.Core.Tests.Shared/AnalyzeThreadSafety.cs
```diff
@@ -192,22 +189,65 @@ public static void CheckStatics(Assembly asm, HashSet<Type> bugsFieldType, HashS
...
-        public static void CheckImmutableTypes(Assembly[] assemblies, bool enableAssert = false)
+        // Does the property have 'init' keyword.
+        private static bool HasInitKeyword(PropertyInfo prop)
+        {
+            var attrs = prop.SetMethod.ReturnParameter.GetRequiredCustomModifiers();
+
+            // Can't use typeof() because it may not be available in .Net 5.0+
+            // So check type name instead. 
+            // typeof(System.Runtime.CompilerServices.IsExternalInit)
+            bool hasInit = attrs.Any(type => type.Name == "IsExternalInit");
```
MikeStall, line 200: This was key fix. Before, we used typeof() , which mandated ifdefs and didn't work on all platforms. 
Just compare the type name instead.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

PR #2844, opened 2025-02-12
Fix parsing of OpenApiDate(Time) default values

LucGenetier:
Make sure ListEnvelopes function in DocuSign swagger can parse DateTime default value for *from_date* param


LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #2845, opened 2025-02-12
Fix issue when body is an array

LucGenetier:
Issue: when the body was an array, the parameter was fully ignored


MikeStall (dismissed):

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #2850, opened 2025-02-17
Add serialization of "items" less arrays

LucGenetier:
Avoid Microsoft.PowerFx.Connectors.PowerFxConnectorException: Incompatible Table for supporting array, RecordValue has more than one column - propertyName msdyn__conversation, number of fields 3

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
@@ -153,13 +153,13 @@ namespace Microsoft.PowerFx.Connectors.Execution
                 return;
             }
 
-            if (propertySchema == null)
+            if (propertySchema == null && !_schemaLessBody)
             {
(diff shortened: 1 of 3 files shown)


LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
```diff
@@ -193,6 +199,64 @@ private async Task WritePropertyAsync(string propertyName, ISwaggerSchema proper
...
+                        await WritePropertyAsync(field.Name, null, field.Value).ConfigureAwait(false);
+                    }
+
+                    EndObject();
+                }
+                else
+                {
+                    throw new NotImplementedException($"Unsupported type {fv.GetType().FullName} for schemaless body");
+                }
+
```
MikeStall, line 256: Can we share this code with somewhere else? 
It seems especially easy to miss a case.

Issue #2852, opened 2025-02-18
Allow more than 20 And operatrors.


MikeStall:
This should succeed. 

```
true && true && true && true && true &&
true && true && true && true && true && 
true && true && true && true && true &&
true && true && true && true && true &&
true && true && true && true && true &&
true && true && true && true && true &&
true && true && true && true && true &&
true && true && true && true && true
```

It's hitting protections for max nesting depth: Message:"Max call depth exceeded.

Note that arranging in flat form works: 

```
And(
    true,true,true,true,true,
    true,true,true,true,true,
    true,true,true,true,true,
    true,true,true,true,true,
    true,true,true,true,true,
    true,true,true,true,true,
    true,true,true,true,true,
    true,true,true,true,true)
```

anderson-joyle:
Can't hosts simply
```
new PowerFxConfig(...)
{
    MaxCallDepth = NEW_MAX_NUMBER
}
```

MikeStall:
The problem is that fundamentally, `a && b && c` shouldn't threaten a stack overflow.

jas-valgotar:
This may help: https://github.com/microsoft/Power-Fx/blob/5132996c77c101f816a91e5a5e6287db6052e0ca/src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs#L600

I think similar optimization may help for all the operator that we convert to call node.

MikeStall:
This is doable, but will require a breaking change to update TexlNode.  The current node is a binary op with just left and right - which mandates a deep tree structure which makes us suspectable to stack overflows. 
We need to change to a new node that is flat. 

So:
`a+b+c+d`

becomes this tree:

`BinOp(BinOp(BinOp(a,b),c),d)`

which is deep and recursive. 
We need a flatter structure:
BinOp(a,b,c,d)

MikeStall:
Fixing this on Fx side is a too large of breaking change.  Even after the breaking change to AST, every visitor needs to be updated 

We have made a fix on the MCS side to use And()  instead of &&

PR #2855, opened 2025-02-20
Rewrite binary operations IR

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2852.

This pull request includes several changes to the `IRTranslator` class in the `Microsoft.PowerFx.Core` library to improve handling of binary operations and simplify the code. Additionally, new tests have been added to ensure the correctness of these changes.

### Improvements to binary operations:

* [`src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs`](diffhunk://#diff-48f6ad402d967eb1a8162c26bca88be5eb6dde544097b7b59d8fb9705e95cb59L228-R231): Refactored the handling of binary operations by replacing the `ConcatenateArgs` method with a more general `UnpackArgs` method. This change allows for more flexible unpacking of arguments for binary operations such as `&&`, `||`, and `&`. [[1]](diffhunk://#diff-48f6ad402d967eb1a8162c26bca88be5eb6dde544097b7b59d8fb9705e95cb59L228-R231) [[2]](diffhunk://#diff-48f6ad402d967eb1a8162c26bca88be5eb6dde544097b7b59d8fb9705e95cb59R595-R626)

### Codebase simplification:

* [`src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs`](diffhunk://#diff-48f6ad402d967eb1a8162c26bca88be5eb6dde544097b7b59d8fb9705e95cb59L12): Removed the unused import statement for `Microsoft.PowerFx.Core.Texl.Builtins`.

### New tests:

* [`src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/IRTests/RewritingIRVisitorTests.cs`](diffhunk://#diff-8b9487be80b794f12a5795a6babc23aa1f3fc7f76e45272d901eabc22a1ac235R77-R107): Added new tests to ensure that binary operation chains are correctly rewritten and do not get blocked by depth limits. These tests cover various binary operations such as `&&`, `||`, `And`, and `Or`.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
@@ -9,7 +9,6 @@ using Microsoft.PowerFx.Core.Functions;
 using Microsoft.PowerFx.Core.IR.Nodes;
 using Microsoft.PowerFx.Core.IR.Symbols;
 using Microsoft.PowerFx.Core.Texl;
-using Microsoft.PowerFx.Core.Texl.Builtins;
 using Microsoft.PowerFx.Core.Types;
 using Microsoft.PowerFx.Core.Utils;
 using Microsoft.PowerFx.Syntax;
@@ -225,11 +224,11 @@ namespace Microsoft.PowerFx.Core.IR
                         binaryOpResult = new CallNode(context.GetIRContext(node), BuiltinFunctionsCore.Power, args);
                         break;
                     case BinaryOpKind.Concatenate:
-                        binaryOpResult = ConcatenateArgs(left, right, context.GetIRContext(node));
+                        binaryOpResult = UnpackArgs(BuiltinFunctionsCore.Concatenate, left, right, context.GetIRContext(node));
                         break;
                     case BinaryOpKind.Or:
                     case BinaryOpKind.And:
-                        binaryOpResult = new CallNode(context.GetIRContext(node), node.Op == BinaryOp.And ? BuiltinFunctionsCore.And : BuiltinFunctionsCore.Or, left, new LazyEvalNode(context.GetIRContext(node), right));
+                        binaryOpResult = UnpackArgs(node.Op == BinaryOp.And ? BuiltinFunctionsCore.And : BuiltinFunctionsCore.Or, left, new LazyEvalNode(context.GetIRContext(node), right), context.GetIRContext(node));
                         break;
 
                     // Reversed Args:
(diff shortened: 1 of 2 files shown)


src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/IRTests/RewritingIRVisitorTests.cs
```diff
@@ -74,6 +74,37 @@ public void Rewrite100x(string expr, string expectedIR, object expected)
             }
         }
 
+        // This test is to ensure a binary op chain is rewritten and does not get blocked by the depth limit.
+        [Theory]
+        [InlineData(
+            "true && true && If(true && true && true, true) && true && true && true && true && true && true && true && true && true && true && true && true && true && true && true && true && true && true && true",
```
CarlosFigueiraMSFT, line 80: Can you add a test with a really long && or || chain (e.g., 1000 items)? It can be written in C# as writing a 1000+ item expression in the inline data would not be very readable. Something like this (the expected IR is wrong, this is just to give an idea):

        [Theory]
        [InlineData("&&", "And")]
        [InlineData("And", "And")]
        [InlineData("||", "Or")]
        [InlineData("Or", "Or")]
        public void Test(string op, string func)
        {
            var sbExpr = new StringBuilder();
            var sbExpected = new StringBuilder();
            sbExpr.Append("true");
            sbExpected.Append($"{func}(true");
            for (var i = 0; i < 1000; i++)
            {
                sbExpr.Append($" {op} true");
                sbExpected.Append(", true");
            }

            sbExpected.Append(')');

            var engine = new RecalcEngine();
            var check = new CheckResult(engine)
                .SetText(sbExpr.ToString())
                .SetBindingInfo();

            var actualIR = check.GetCompactIRString();
            Assert.Equal(sbExpected.ToString(), actualIR);
        }

anderson-joyle: Thanks for the suggestion. I've added it.

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -591,33 +590,40 @@ private static IntermediateNode BlankToEmptyString(IntermediateNode arg)
...
-            }
-
-            /// <summary>
-            /// This is not a generic arg concatenate function, but a special case for the Concatenate function,
-            /// used for Binary Concatenate operator.
-            /// </summary>
-            private static IntermediateNode ConcatenateArgs(IntermediateNode arg1, IntermediateNode arg2, IRContext irContext)
+            }
+
+#pragma warning disable CS1570 // XML comment has badly formed XML
```
CarlosFigueiraMSFT, line 595: >// XML comment has badly formed XML [](http://example.com/codeflow?start=31&length=35)

Can we make it well-formed, instead of disabling this warning? That will help with intellisense. I'm guessing that the `&` characters will have to be escaped as `&amp;`

src/libraries/Microsoft.PowerFx.Core/IR/IRTranslator.cs
```diff
@@ -591,33 +590,40 @@ private static IntermediateNode BlankToEmptyString(IntermediateNode arg)
...
+#pragma warning restore CS1570 // XML comment has badly formed XML
             {
-                var concatenateArgs = new List<IntermediateNode>();
+                var rangeArgs = new List<IntermediateNode>();
                 foreach (var arg in new[] { arg1, arg2 })
                 {
-                    // if arg is call node to Concatenate unpack it, and pass it as arg to outer Concatenate
-                    if (arg is CallNode maybeConcatenate && maybeConcatenate.Function is ConcatenateFunction concatenateFunction)
+                    // If arg is a call node to the same function, unpack the args.
+                    if (arg is CallNode callNode && callNode.Function.Name == func.Name)
```
CarlosFigueiraMSFT, line 612: >callNode.Function.Name == func.Name [](http://example.com/codeflow?start=52&length=35)

Two functions can have the same name if they are in different namespaces. We should do a stricter comparison in this case.

anderson-joyle: Yes, but we only call this for And/Or/Concatenate.

src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/IRTests/RewritingIRVisitorTests.cs
```diff
@@ -74,6 +74,43 @@ public void Rewrite100x(string expr, string expectedIR, object expected)
...
+        [InlineData("&&", "And")]
+        [InlineData("And", "And")]
+        [InlineData("||", "Or")]
+        [InlineData("Or", "Or")]
+        public void BinaryOpIRRewriteTest(string op, string func)
+        {
+            var sbExpr = new StringBuilder($"true {op} If(true {op} true {op} true, true)");
+            var sbExpected = new StringBuilder($"{func}(True, (If({func}(True, (True), (True)), (True)))");
+
+            for (var i = 0; i < 1000; i++)
```
MikeStall, line 88: >1000 [](http://example.com/codeflow?start=32&length=4)

Jus for kicks, can we set this limit to 100k and verify that doesn't cause a stack overflow?

LucGenetier:
✅ No public API change.

PR #2857, opened 2025-02-25
Block at compile time invalid arguments to Distinct

CarlosFigueiraMSFT:
The Distinct function only works for primitive values, but currently this error only happens at runtime. This PR adds a compile-time check to ensure that only primitive values are passed to the Distinct function. The check is done under the PFxV1 flag, since this currently works in Power Apps (although there are many bugs on the implementation).


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

Ananya-Gupta24 (approved):

PR #2859, opened 2025-03-01
Example test demonstrating failure for 2858

MikeStall:
Demonstrates failure for #2858


src/tests/Microsoft.PowerFx.Core.Tests.Shared/TypeSystemTests/LazyTypeTests.cs
```diff
@@ -117,6 +117,35 @@ public void KindString()
...
+            int c2 = _getter1CalledCount;
+            Assert.Equal(1, c1 + 1); // only does 1 lookup based on the expression.
+
+            Assert.True(check.IsSuccess);
+
+            // Logging shouldn't make any more calls. 
+            var log = check.ApplyGetLogging();
+
+            int c3 = _getter1CalledCount;
+            Assert.Equal(c2, c3);
```
MikeStall, line 146: this assert fails. Bug is to fix this.

PR #2864, opened 2025-03-10
EvalVisitor.Culture is always non-null.

MikeStall:
Add enforcement.

This is true because the class is private and it's always set here: 
https://github.com/microsoft/Power-Fx/blob/27928265741af482f8c650ccca7902231ead44c8/src/libraries/Microsoft.PowerFx.Interpreter/ParsedExpression.cs#L138

### DIFF
--- a/src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/LanguageTest.cs
+++ b/src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/LanguageTest.cs
@@ -49,13 +49,6 @@ namespace Microsoft.PowerFx.Tests
         [Fact]
         public void GetLanguageForNullCulture()
         {
-            var runtimeConfig = new RuntimeConfig();
-            var runner = new EvalVisitor(runtimeConfig, CancellationToken.None);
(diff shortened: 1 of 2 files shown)


src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/LanguageTest.cs
```diff
@@ -49,13 +49,6 @@ public void GetDefaultLanguageTest()
         [Fact]
         public void GetLanguageForNullCulture()
         {
-            var runtimeConfig = new RuntimeConfig();
-            var runner = new EvalVisitor(runtimeConfig, CancellationToken.None);
```
MikeStall, line 53: This test is failing because it hits private surface area.  EvalVisitor is private, and we have to go through ParsedExpression, which will set the culture.

CarlosFigueiraMSFT (approved):
Can you also update ParsedExpression.cs, line 147, to remove the check that is not necessary?

LucGenetier:
✅ No public API change.

PR #2866, opened 2025-03-11
ConnectorType fix

LucGenetier:
Avoids a NullRef in DisplayNameProvider creation
Update ConnectorType error message
Expose ConnectorType in CdpTable

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
@@ -470,7 +470,7 @@ namespace Microsoft.PowerFx.Connectors
 
             if (settings.Level == 30)
             {
-                return new ConnectorType(error: "GetConnectorType() excessive recursion");
+                return new ConnectorType(error: "GetConnectorType() excessive recursion") { Name = openApiParameter.Name };
             }
 
             // schema.Format is optional and potentially any string
@@ -546,7 +546,7 @@ namespace Microsoft.PowerFx.Connectors
                             return TryGetOptionSet(openApiParameter, settings) ?? new ConnectorType(schema, openApiParameter, FormulaType.Decimal);
 
                         default:
-                            return new ConnectorType(error: $"Unsupported type of integer: {schema.Format}");
+                            return new ConnectorType(error: $"Unsupported type of integer: {schema.Format}") { Name = openApiParameter.Name };
                     }
 
                 case "array":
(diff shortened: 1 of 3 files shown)


src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -691,7 +691,7 @@ internal static ConnectorType GetConnectorType(this ISwaggerParameter openApiPar
                     return new ConnectorType(schema, openApiParameter, FormulaType.Blob);
 
                 default:
-                    return new ConnectorType(error: $"Unsupported schema type {schema.Type}");
+                    return new ConnectorType(error: $"Unsupported schema type {schema.Type} for field {openApiParameter.Name}");
             }
```
MikeStall, line 695: should we set name / displayname here?

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/Public/ConnectorType.cs
```diff
@@ -316,7 +316,7 @@ internal DisplayNameProvider DisplayNameProvider
         {
             get
             {
-                _displayNameProvider ??= DisplayNameUtility.MakeUnique(Fields.Select(field => new KeyValuePair<string, string>(field.Name, field.DisplayName ?? field.Name)));                    
+                _displayNameProvider ??= DisplayNameUtility.MakeUnique(Fields.Where(f => !string.IsNullOrEmpty(f.Name)).Select(field => new KeyValuePair<string, string>(field.Name, field.DisplayName ?? field.Name)));
```
MikeStall, line 319: >!string.IsNullOrEmpty(f.Name)). [](http://example.com/codeflow?start=89&length=31)

why this change?

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
MikeStall: but you're not failing ... you're passing a partially initialized object through the rest of the system and letting it fail at some future point in totally unpredictable ways.  

So the current test shows us this will hit a null ref later ... and you added an extra null check in ConnectorType to protect against that one failure ...

LucGenetier: Not here, but later with a null ref

LucGenetier:
✅ No public API change.

MikeStall (approved):
:shipit:

PR #2871, opened 2025-03-14
Add test for localization

MikeStall:
Add test demonstrating how to extract text literals.


src/tests/Microsoft.PowerFx.Interpreter.Tests.Shared/Scenarios/TextLiteralFinderVisitor.cs
```diff
@@ -0,0 +1,469 @@
...
+    // Demonstrate mutation example using IUntypedObject
+    public class TextLiteralFinderTests
+    {
+        [Fact]
+        public void Test()
+        {
+            string formula = @"
+                {
+                    type: ""AdaptiveCard"",
+                    body: [ //comment
```
anderson-joyle, line 27: Where this whole string comes from?

PR #2877, opened 2025-03-25
Regular expression update for Unicode characters

gregli-msft:
Updates for Unicode behavior:
- Supports surrogate pairs in the regular expression when on .NET by wrapping with a non-capture group
- Supports surrogate pair matching for dot on .NET by including surrogate pair range character classes
- Supports `\p{Lt}`, `\p{Lc}`, `\p{Ll}` under option ignore case by matching with `[\p{Lt}\p{Lc}\p{Ll}]` when on JavaScript (matching .NET behavior)
- Addes `\u{...}` syntax support

Also simplified capture group tracking and now support backref to a capture group if they are in the same containing group , for example: `(\?<one>a)\k<one>|b`


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT:
>                                      \\u[dD][c-fC-F][0-9a-fA-F]{2}) |    # \u{...} syntax is not supported for surrogate pairs (same as JavaScript and PCRE2)

Won't it match everything that \\u[dD][0-9a-fA-F]{3}?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Match.cs:580 in 5694a34. [](commit_id = 5694a340068196ea73ab7d7588ed177046ec640d, deletion_comment = False)

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/IsMatch_V1Compat.txt
```diff
@@ -28,33 +28,30 @@ false
...
 false
 
-// \u{...} notation not supported by .net
-// will be blocked in V1 by the regular expression checker (for now) with a compile time error
-// when V1 is disabled, runtime error on .NET because it doesn't support it in the interpreter
 >> IsMatch( "a", "\u{0061}" )
-Errors: Error 14-24: Invalid regular expression: Invalid escape code, found "\u".|Error 0-7: The function 'IsMatch' has some invalid arguments.
+true
 
 >> IsMatch( "a", "\u{1F47B}" )
```
CarlosFigueiraMSFT, line 34: Do we have a test where IsMatch for a unicode escape returns true?

CarlosFigueiraMSFT (approved):
:shipit:

LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/IsMatch_V1Compat.txt
gregli-msft: We had tests for Match in the Match_Unicode file.  Since the interpreter now supports `\u{...}`, with or without V1, it is easier to write tests that work both with and without V1.  So, I've pulled all the Unicode content, for all three functions, into the Match_Unicode* files.  This makes it easier to deal with V1/pre-V1, 462/70, all in one place.

LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

PR #2885, opened 2025-04-08
[PowerApps] Localization hand-back [CTAS - RunID=20250408-213525-vxbs…

toshio-msft:
…ksvka2]


LucGenetier:
✅ No public API change.

PR #2888, opened 2025-04-11
Update IsMatch constant check for pre-v1

gregli-msft:
Restore IsMatch behavior where we fall back to IsConstant if TryGetConstantValue fails.  Takes us back to the old behavior in Power Apps' component properties when the compiler may believe that IsConstant is true when in reality it is not.


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

PR #2889, opened 2025-04-14
Update IsMatch constant check for pre-v1 (#2888)

gregli-msft:


LucGenetier:
✅ No public API change.

gregli-msft:
Pull request not needed for QFE to proceed.  Branch is still legitimate and used for that train.

PR #2890, opened 2025-04-15
Must Compute Return Type For Patch For Dataflow Analysis

akshar2401:
It is possible to have an error type as one of the arguments to Patch during initial passes of dataflow analysis in Canvas Apps. The type corrects itself later which is when Patch.CheckTypes would not find any errors to report to. During dataflow analysis, we utilize CheckTypes just to compute the return type that we can propagate forward in the Dataflow graph. The current Patch.CheckTypes would set the return type to DType.Error which is then propagated forward and never corrected leading to misleading type errors in the app. Considering that we only care about return type during dataflow analysis and not the actual errors which actually become meaningful during later binding stage, this pr ensures that return type is computed even when there are errors.


LucGenetier:
✅ No public API change.

lesaltzm (approved):

PR #2892, opened 2025-04-16
[Perf] Allow skipping QueryOptions operations while binding if they're only needed for canvas legacy analysis

lesaltzm:
a few things done while binding are only relevant to canvas apps running legacy analysis, and they've got a perf cost. We should be able to skip them for other hosts or for canvas when running new analysis.


anderson-joyle (approved):
:shipit:

LucGenetier:
✅ No public API change.

PR #2895, opened 2025-04-22
Update FormulaValueSerializer to handle null schemas

madevoge:
_Today,_ if we have an unknown schema, it requires the record to have a single .Value property (as if it is a primitive).
If it is not, we throw.

_With this change,_ when we are encountering complex (unknown) schemas that arent all .Value, we gracefully write the full record instead of throwing.

Change required updates to the UriFormEncoder to better handle nested objects within arrays.


LucGenetier:
✅ No public API change.

MikeStall (approved):

PR #2896, opened 2025-04-23
PAIO bug 32386341

LucGenetier:
Improve CDP GetResult API
Remove SF referencedEntities useless parsing


LucGenetier:
✅ No public API change.

LucGenetier:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/Tabular/Services/CdpTable.cs
```diff
@@ -146,9 +146,13 @@ private async Task<string> Query(IServiceProvider serviceProvider, DelegationPar
         }
 
         private IReadOnlyCollection<DValue<RecordValue>> GetResult(string text)
-        {
-            // $$$ Is this always this type?
-            RecordValue rv = FormulaValueJSON.FromJson(text, RecordType.Empty().Add("value", TableType)) as RecordValue;
+        {            
+            if (FormulaValueJSON.FromJson(text, RecordType.Empty().Add("value", TableType)) is not RecordValue rv)
+            {
+                // FromJson can return an ErrorValue and in that case let's return an empty result
```
MikeStall, line 152: > empty result [](http://example.com/codeflow?start=85&length=13)

My concern is dropping the error. 

Should we use DValue.Of(ErrorValue) and propagate the error through?

LucGenetier: Let's propagate the error properly

src/libraries/Microsoft.PowerFx.Connectors/Execution/FormulaValueSerializer.cs
```diff
@@ -168,37 +168,40 @@ private async Task WritePropertyAsync(string propertyName, ISwaggerSchema proper
...
                 else if (tableValue.Rows.All(r => r.Value.Fields.Count() == 1 && r.Value.Fields.First().Name == "Value"))
                 {
                     // Working with an array of simply types
                     foreach (DValue<RecordValue> item in tableValue.Rows)
                     {
-                        StartArrayElement(null);
-                        RecordValue rva = item.Value;
-                        WriteValue(rva.Fields.First().Value);
+                        if (!item.IsError)
+                        {
```
jas-valgotar, line 186: How does FVSerializer treat ErrorValue? Can we add a test for this?

LucGenetier: This code is just a safety net that was probably missing for a long time. When an ErrorValue would be present, item.Value would be null and we'd fail with a NullRef exception
For now we ignore errors and I'm not sure what sort of test we could add / what the right behavior should really be

src/libraries/Microsoft.PowerFx.Connectors/Tabular/Services/CdpTable.cs
```diff
@@ -147,8 +147,26 @@ private async Task<string> Query(IServiceProvider serviceProvider, DelegationPar
...
+                ErrorValue err = new ErrorValue(IRContext.NotInSource(rt), new ExpressionError()
+                {
+                    Message = $"FormulaValueJSON.FromJson doesn't return a RecordValue - Received {fv.GetType().Name}",
+                    Span = new Syntax.Span(0, 0),
+                    Kind = ErrorKind.InvalidJSON
+                });
+
+                return new List<DValue<RecordValue>>() { DValue<RecordValue>.Of(err) };
+            }
+
```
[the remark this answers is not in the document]
LucGenetier: The PAIO bug I'm looking at is mentioning a NullRef exception in that function (GetResult)
One possibililty is 'as RecordValue' returning null.

FormulaValueJSON.FromJson can potentially fail and will return an ErrorValue which we can propagate - lines 153-155.

Now, for lines 160-167, I agree it's debatable but we can also return an ErrorValue to be consistent with previous lines

LucGenetier:
✅ No public API change.

MikeStall (approved):

LucGenetier:
✅ No public API change.

PR #2901, opened 2025-05-05
dont force raw output when override is set

madevoge:
When the caller explicitly provides a type, we should honor it as such type, no matter the flags on the response type
Today, the override is completely skipped if there was no known output type.


LucGenetier:
✅ No public API change.

PR #2902, opened 2025-05-06
Block Void type use in UDF argument type

adithyaselv:
Today we allow Void type to be used as a valid type in UDF argument while it serves no real purpose. We do not want to allow this pattern and this PR blocks such usage.

image


LucGenetier:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/UserDefinedFunctionTests.cs
```diff
@@ -660,14 +660,37 @@ public void TestUDFRestrictedTypes()
                 AllowsSideEffects = true,
```
CarlosFigueiraMSFT, line 660: >AllowsSideEffects = true [](http://example.com/codeflow?start=16&length=24)

Needed? The script doesn't use imperative constructs.

CarlosFigueiraMSFT (dismissed):
:shipit:

CarlosFigueiraMSFT (approved):
:shipit:

PR #2905, opened 2025-05-12
[Draft] Helper function for UDF optimization

adithyaselv:


LucGenetier:
✅ No public API change.

jas-valgotar:
✅ No public API change.

jas-valgotar:
✅ No public API change.

jas-valgotar:
✅ No public API change.

jas-valgotar:
✅ No public API change.

jas-valgotar:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/UDFHasSameDefinitionTests.cs
```diff
@@ -0,0 +1,135 @@
...
+using Xunit;
+
+namespace Microsoft.PowerFx.Core.Tests
+{
+    public class UDFHasSameDefinitionTests : PowerFxTest
+    {
+        private static readonly ReadOnlySymbolTable _primitiveTypes = ReadOnlySymbolTable.PrimitiveTypesTableInstance;
+
+        [Theory]
+        [InlineData("Foo(x: Number): Number = Abs(x);", "Foo(x: Number): Number = Abs(x);", true)]
```
CarlosFigueiraMSFT, line 21: >[InlineData("Foo(x: Number): Number = Abs(x);", "Foo(x: Number): Number = Abs(x);", true)] [](http://example.com/codeflow?start=8&length=90)

Can you add a trivial equal case for imperative UDFs?

src/tests/Microsoft.PowerFx.Core.Tests.Shared/UDFHasSameDefinitionTests.cs
```diff
@@ -0,0 +1,135 @@
...
+        
+        // test with different udf body
+        [InlineData("Foo(x: Number): Number = Abs(x);", "Foo(x: Number): Number = Abs( /*Hello*/ x);", false)]
```
CarlosFigueiraMSFT, line 25: >[InlineData("Foo(x: Number): Number = Abs(x);", "Foo(x: Number): Number = Abs( /*Hello*/ x);", false)] [](http://example.com/codeflow?start=8&length=102)

Shouldn't they be considered equal? Comments on the test on line 29 are disregarded when checking for equality

src/tests/Microsoft.PowerFx.Core.Tests.Shared/UDFHasSameDefinitionTests.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        // test with different whitespace and trivials
+        [InlineData("Foo(x: Number):   Number =   Abs(x)  ;", "Foo(x:Number):Number=Abs(x);", true)]
+        [InlineData("Foo(x: Number): Number = Abs(x);", "Foo(x: /*comment*/ Number): Number = /*comment*/ Abs(x);", true)]
+
+        // test with different function names
+        [InlineData("Foo(x: Number): Number = Abs(x);", "FooBar(x: Number): Number = Abs(x);", false)]
+
+        // test with different parameter names
+        [InlineData("Foo(x: Number): Number = Abs(x);", "Foo(y: Number): Number = Abs(x);", false)]
```
CarlosFigueiraMSFT, line 36: >Foo(y: Number): Number = Abs(x); [](http://example.com/codeflow?start=57&length=32)

This is an error, right? Should it be Abs(y)?

src/tests/Microsoft.PowerFx.Core.Tests.Shared/UDFHasSameDefinitionTests.cs
```diff
@@ -0,0 +1,135 @@
...
+        [InlineData("Foo(a: Number, b: Number, c: Number): Number = a+b+c;", "Foo(a: Number, b: Number, d: Number): Number = a+b+c;", false)]
```
CarlosFigueiraMSFT, line 37: >c [](http://example.com/codeflow?start=129&length=1)

`d`?

src/tests/Microsoft.PowerFx.Core.Tests.Shared/UDFHasSameDefinitionTests.cs
```diff
@@ -0,0 +1,135 @@
...
+        [InlineData("Foo(x: Number): Number = Abs(x);", "Foo(x: Number): Number = Abs(5);", false)]
+
+        // test with different whitespace and trivials
+        [InlineData("Foo(x: Number):   Number =   Abs(x)  ;", "Foo(x:Number):Number=Abs(x);", true)]
+        [InlineData("Foo(x: Number): Number = Abs(x);", "Foo(x: /*comment*/ Number): Number = /*comment*/ Abs(x);", true)]
+
+        // test with different function names
+        [InlineData("Foo(x: Number): Number = Abs(x);", "FooBar(x: Number): Number = Abs(x);", false)]
+
+        // test with different parameter names
```
CarlosFigueiraMSFT, line 35: >parameter names [](http://example.com/codeflow?start=31&length=15)

Those functions are technically the same, right?

- Foo(x:Number): Number = Abs(x);
- Foo(y:Number): Number = Abs(y);

Are we considering them different because it's hard to "unify" the arguments?

src/tests/Microsoft.PowerFx.Core.Tests.Shared/UDFHasSameDefinitionTests.cs
adithyaselv: correct I am not doing ignoring trivials in this iteration . doing a check with TexlNodes ignoring trivials would be the best thing to do but equality check with texlNode is not good today so plan to improve this part probably in a future PR

src/tests/Microsoft.PowerFx.Core.Tests.Shared/UDFHasSameDefinitionTests.cs
adithyaselv: not a parse error . this would be a bind error , i check sameness before we bind

src/tests/Microsoft.PowerFx.Core.Tests.Shared/UDFHasSameDefinitionTests.cs
adithyaselv: yes they have the same eta equivalence, but our current design would require reprocessing as we map the parameter names to type in our internal UDF bind dictionary and rebinding would fail

src/tests/Microsoft.PowerFx.Core.Tests.Shared/UDFHasSameDefinitionTests.cs
adithyaselv: clarified offline these are just trivial test with no body change but with param name change and to see if they are detected in the sameness check

src/tests/Microsoft.PowerFx.Core.Tests.Shared/UDFHasSameDefinitionTests.cs
adithyaselv: same as the other comment

CarlosFigueiraMSFT (dismissed):
:shipit:

src/tests/Microsoft.PowerFx.Core.Tests.Shared/UDFHasSameDefinitionTests.cs
adithyaselv: done

CarlosFigueiraMSFT (approved):
:shipit:

jas-valgotar:
✅ No public API change.

PR #2906, opened 2025-05-12
{Draft] Rename UntypedObject to Dynamic at user level

adithyaselv:
Renames UntypedObject to Dynamic at end user perspective. Internal name UntypedObject stays the same


LucGenetier:
✅ No public API change.

adithyaselv:
fixed by #2912

PR #2909, opened 2025-05-15
Introduce PowerPlatformConnectorClient2

shgogna:
This PR introduces a new version of `PowerPlatformConnectorClient` called `PowerPlatformConnectorClient2`. Naming for the new client is open for debate 😄

The differences include:
* An interface for testing/mocking: `IPowerPlatformConnectorClient2`
* A delegate for providing the bearer token: `PowerPlatformConnectorClient2BearerTokenProvider`
* POCO for diagnostic values: `PowerPlatformConnectorClient2DiagnosticOptions` which specifies values that are typically sent by clients
* Connector client takes specific parameters for `method`, `operationPathAndQuery`, `headers`, and `content` instead of reusing `HttpRequestMessage`. This was done to avoid client teams creating a dummy object that would not even get used.
* Connector client takes the full base URL (also known internally as the "primaryRuntimeUrl")
* Added `GetFirstServerUri` extension to `OpenApiExtensions`

Misc:
* Remove trailing whitespace from `OpenApiExtensions`

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/IPowerPlatformConnectorClient2.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/IPowerPlatformConnectorClient2.cs
@@ -1,0 +1,39 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.PowerFx.Connectors
+{
+    /// <summary>
+    /// Delegate for providing a bearer token for authentication.
+    /// </summary>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    /// <returns>Token without "Bearer" scheme as the prefix.</returns>
+    public delegate Task<string> PowerPlatformConnectorClient2BearerTokenProvider(
+        CancellationToken cancellationToken);
+
+    public interface IPowerPlatformConnectorClient2
+    {
+        /// <summary>
+        /// Sends an HTTP request to the Power Platform connector.
+        /// </summary>
+        /// <param name="method">HTTP method.</param>
+        /// <param name="operationPathAndQuery">Operation path and query string.</param>
+        /// <param name="headers">Headers.</param>
+        /// <param name="content">Content.</param>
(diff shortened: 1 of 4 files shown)


src/libraries/Microsoft.PowerFx.Connectors/PowerPlatformConnectorClient2.cs
```diff
@@ -0,0 +1,174 @@
...
+                // without worrying about the "/".
+                str = str.TrimEnd('/');
+                return str;
+            }
+        }
+
+        // <inheritdoc />
+        public async Task<HttpResponseMessage> SendAsync(
+            HttpMethod method,
+            string operationPathAndQuery,
```
shgogna, line 75: Note to self: potential vuln with `..` injected into the path and query. Might need to enforce `Uri` which handles and resolves `..` during construction.

shgogna: Fixed. Path traversal detected by matching the prefix. Exception will be thrown.

src/libraries/Microsoft.PowerFx.Connectors/IPowerPlatformConnectorClient2.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// Sends an HTTP request to the Power Platform connector.
+        /// </summary>
+        /// <param name="requestMessage">HTTP request message. URI path will be used as a relative path for the connector request.</param>
+        /// <param name="diagnosticOptions">Diagnostic options.</param>
+        /// <param name="cancellationToken">Cancellation token.</param>
+        /// <returns><see cref="HttpResponseMessage"/>.</returns>
+        Task<HttpResponseMessage> SendAsync(
+            HttpRequestMessage requestMessage,
+            PowerPlatformConnectorClient2DiagnosticOptions diagnosticOptions,
+            CancellationToken cancellationToken);
```
MikeStall, line 31: Since this is a public SDK, the interface should be minimal. 
Can this method be an extension method that calls into the other "true" one?

shgogna: Moved this to an extension method

PR #2912, opened 2025-05-21
Rename UntypedObject to Dynamic (UI side)

Ananya-Gupta24:
After discussion, it was decided to rename UntypedObject to Dynamic (what customers see). This PR makes that change.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs
@@ -89,7 +89,7 @@ namespace Microsoft.PowerFx.Types
             { new DName("Text"), String },
             { new DName("Hyperlink"), Hyperlink },
             { new DName("None"), Blank },
-            { new DName("UntypedObject"), UntypedObject },
+            { new DName("Dynamic"), UntypedObject },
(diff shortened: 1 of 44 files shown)


CarlosFigueiraMSFT:
>         // Primitive types - Keeping it same as PrimitiveTypeSymbolTable

This comment doesn't make sense anymore - there's no PrimitiveTypeSymbolTable. Can you remove it?

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Public/Types/FormulaType.cs:77 in 285e07f. [](commit_id = 285e07f558b091fd50b54aa697cc8e334e829538, deletion_comment = False)

CarlosFigueiraMSFT:
There are many tests that still reference "UntypedObject" - although they seem to be negative tests. If the tests are passing, do the error message for those scenarios need to be updated? A few examples:
- tests\Microsoft.PowerFx.Interpreter.Tests.Shared\RecalcEngineTests.cs
- tests\Microsoft.PowerFx.Json.Tests.Shared\AsTypeIsTypeParseJSONTests.cs

Ananya-Gupta24:
Draft PR to check for test failures

CarlosFigueiraMSFT (approved):
:shipit:

Issue #2221, opened 2024-02-21
Unexpected void type in the formula


Robulane:
This test fails

        [Fact]
        public void CheckParseSuccess()
        {
            var config = new PowerFxConfig();
            var engine = new Engine(config);
            var parse = engine.Parse("If(Len(x) = 0, y, x)");

            var r = RecordType.Empty()
                .Add(new NamedFormulaType("x", FormulaType.String))
                .Add(new NamedFormulaType("y", FormulaType.UntypedObject));

            var check = engine.Check(parse, r);
            Assert.True(check.IsSuccess);

            Assert.IsType<UntypedObjectType>(check.ReturnType);
        }

Robulane:
I disagree that this is expected. The Void should never be returned for the expression that expects the value. You should either return the error or the type of the first argument. If the first argument is Untyped, I think you can return Untyped (because it's untyped). Expressions like this should return an error though: If(true, 2, "")

Robulane:
We should not wait the Runtime to throw an error, the check function should return the error.

MikeStall:
For imperative, void is useful. 
For expression-only, void should be an error. 

Fortunately, we do know whether an expression allows side-effects. This is determined via the parser options, and it impacts things like whether we have have chaining operating  or call behavior functions. 

Proposal: if the expression does not allow side-effects, then any usage of void becomes an error from power fx.

PR #2917, opened 2025-06-03
Converts void to error for non sideeffecting expressions.

jas-valgotar:
Fixes https://github.com/microsoft/Power-Fx/issues/2221
This pull request introduces functionality to detect and handle void expressions in non-side-effecting contexts by converting them into error expressions. It includes updates to the core binding logic, feature configurations, and associated tests.

### Core Binding Enhancements:
* Added a new `VoidToErrorTexlVisitor` class in `src/libraries/Microsoft.PowerFx.Core/Binding/VoidToErrorTexlVisitor.cs` to traverse expressions and ensure void expressions are flagged as errors.
* Updated the `Run` method in `TexlBinding` within `src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs` to invoke the `VoidToErrorTexlVisitor` when side effects are disallowed.

### Test Enhancements:
* Added test cases in `src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/VoidToError.txt` to validate the behavior of void-to-error conversion across various expressions and scenarios.
* Added test cases in `src/tests/Microsoft.PowerFx.Interepter.Tests.Shared/MutationScripts/Void.txt` to validate the Void behavior when side effects are enabled, largely not leading to an error.

### Breaking Language Change:
The behavior of `IsError` and `IfError` are being modified with respect to Void, _only when side effects are disabled_. Since use of Void in these functions with side effects disabled would not have been useful, it would always treated it as a non-error, this change is not expected to impact existing customers.

The reason we turn a blind eye to Void results in side effect enabled scenarios, such as Canvas app Button.OnSelect, is so that makers can write control flow with the If function. For example, in this case where Orders {Value: Number} and BadSessions {Value: GUID} have incompatible schemas: 
```
If( allGood, 
    Collect( Orders, {Value:Int(TextInput1.Text)} ), 
    Collect( BadSessions, {Value:SessionGUID} ) 
)
```

Makers should be able to wrap that whole formula with `IfError` or `IsError` to detect runtime problems and take appropriate actions. For example, the value of TextInput1.Text in this example may not be a parseable number.
 
But without side effects, the `If` function isn't about control flow, the formula is expected to produce a single, non-Void return value.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
+++ b/src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
@@ -88,6 +88,22 @@ namespace Microsoft.PowerFx.Types
             return FromJson(element, settings, new FormulaValueJsonSerializerWorkingData(), formulaType);  
         }
 
+        // // caller verified element is non-null and is of type string 
+        internal static FormulaValue ParseDate(JsonElement element, FormulaType targetType, Func<DateTime, FormulaValue> funcParse)
+        {
+            var strValue = element.GetString(); 
+            if (string.IsNullOrWhiteSpace(strValue))
+            {
+                return FormulaValue.NewBlank(targetType);
+            }
+
+            // Any exceptions will be caught at higher level. 
+            var dateTime = element.GetDateTime();
+
+            var value = funcParse(dateTime);
+            return value;
+        }
+
         internal static FormulaValue FromJson(JsonElement element, FormulaValueJsonSerializerSettings settings, FormulaValueJsonSerializerWorkingData data, FormulaType formulaType = null)
         { 
             if (formulaType is UntypedObjectType uot)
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Core/Public/Config/Features.cs
```diff
@@ -102,7 +107,15 @@ public sealed class Features
             PrimaryOutputPropertyCoercionDeprecated = true,
             AsTypeLegacyCheck = false,
             JsonFunctionAcceptsLazyTypes = true,
-            IsUserDefinedTypesEnabled = true
+            IsUserDefinedTypesEnabled = true,
+            IsVoidError = false,
+        };
+
+        public static Features PowerFxV2 => _powerFxV2;
```
MikeStall: let's chat more about ways to manage this.

jas-valgotar:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/ExpressionTestCases/VoidToError.txt
```diff
@@ -0,0 +1,29 @@
+﻿#SETUP: PowerFxV2CompatibilityRules
+
+>> If(false, 3, TimeUnit.Years)
+Errors: Error 0-28: A void expression cannot be used in this context.
```
CarlosFigueiraMSFT: Do we need a more helpful error to the customers? "A void expression cannot be used in this context" doesn't tell exactly what they need to do to fix it. #Resolved

jas-valgotar:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Config/Features.cs
```diff
@@ -82,6 +82,11 @@ public sealed class Features
         /// </summary>
         internal bool IsUserDefinedTypesEnabled { get; init; } = false;
 
+        /// <summary>
+        /// Enables treating void type expression as error in non-side-effecting expressions <see cref="ParserOptions.AllowsSideEffects"/>.
+        /// </summary>
+        internal bool IsVoidError { get; init; } = false;
```
CarlosFigueiraMSFT: Do we need this at all? Given that any void expressions in a context where bindingContext.AllowsSideEffects is false today may be a runtime error (it would now become a compile-time error). While this is technically a breaking change, it looks like one that we should make.

CC: @MikeStall, @gregli-msft

src/libraries/Microsoft.PowerFx.Core/Binding/VoidToErrorTexlVisitor.cs
```diff
@@ -0,0 +1,138 @@
...
+        public override void Visit(ParentNode node)
+        {
+            EnsureNotVoid(node);
+        }
+
+        public override void Visit(SelfNode node)
+        {
+            EnsureNotVoid(node);
+        }
+
```
CarlosFigueiraMSFT, line 96: Those nodes can never have void values; do we need those overrides? #Resolved

jas-valgotar: yes I believe most of them can never be void, keeping it just for future proofing...

jas-valgotar:
✅ No public API change.

jas-valgotar:
✅ No public API change.

jas-valgotar:
✅ No public API change.

Shpakh-MSFT (approved):
:shipit:

jas-valgotar:
✅ No public API change.

PR #2921, opened 2025-06-11
Adding Copilot generated /// comments for connector namespace …

JimDaly:
I spent about an hour trying to see which descriptions could be updated using VSCode Copilot Agent mode with this prompt:

```
Please scan all .cs files in the current workspace. For each file:

1. Only process public interfaces, enumerations, classes, methods, properties, and fields.
2. Do not touch private, internal, or protected members.
3. If a public member already has a /// XML comment, leave it unchanged.
4. If a public member is missing a /// comment, add a basic XML summary comment.
5. Do not modify formatting or indentation.
This operation must be idempotent and safe to run multiple times.
```

And a little trial & error.

If these descriptions can be updated, then perhaps this reference documentation will be more valuable:
https://learn.microsoft.com/en-us/dotnet/api/microsoft.powerfx.connectors?view=powerfx-sdk-latest

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/ConnectorConstants.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/ConnectorConstants.cs
@@ -2,33 +2,114 @@
 // Licensed under the MIT license.
 namespace Microsoft.PowerFx.Connectors
 {
+    /// <summary>
+    /// Contains constant values for x-ms-* OpenAPI extensions used in connectors.
+    /// </summary>
     public static class Constants
     {
+        /// <summary>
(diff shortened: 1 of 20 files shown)


jas-valgotar:
✅ No public API change.

JimDaly:
@jas-valgotar can you approve this PR?
@MikeStall  approved this earlier version: https://github.com/microsoft/Power-Fx/pull/2919
But I the test were failing. I thought this was because I was working on a fork of the repo. But now I expect it was actually due to the very strict style requirements, that I finally overcame.

JimDaly:
@MikeStall Will you approve this?

jas-valgotar:
✅ No public API change.

PR #2923, opened 2025-06-16
Control AllowSideEffects in Repl

gregli-msft:
Adds control over ParserOptions.AllowSideEffects in the repl. On by default, `Option( Options.AllowSideEffects, false )` will disable side effects. Useful for testing new Void semantics.

Also fixed a small bug to disallow initial `Set` when side effects are disallowed.


jas-valgotar:
✅ No public API change.

jas-valgotar:
✅ No public API change.

jas-valgotar:
✅ No public API change.

PR #2927, opened 2025-06-22
Adding Copilot generated /// comments for Core namespace

JimDaly:
Using Copilot I've generated /// comments for the PowerFx.Core Namespacw with a goal to make the generated reference content more useful. This reference content:  https://learn.microsoft.com/dotnet/api/microsoft.powerfx.core

Mostly, I used this prompt, but there was a lot of manual changes too.

> For the active document:
> 1. Only process public interfaces, enumerations, classes, methods, constructors, properties, and fields.
>    •  Do not process, modify, remove, or comment internal, protected, or private members. These must remain in the file exactly as they are.
> 2. Add, modify, or remove comments ONLY for public members, including public constructors. Do not include private, internal, or any type of constructor that isn't public.
> 3. If a public member already has a /// XML comment, leave it unchanged unless any <summary>, <param>, or <returns> tags are empty or missing.
> 4. If a public member is missing a /// comment, add a basic XML summary comment.
> 5. For all <param> and <returns> tags, ensure they contain meaningful, non-empty descriptions relevant to the parameter or return value. If a tag is empty or missing, generate appropriate content.
> 6. Do not modify formatting or indentation.
> 7. This operation must be idempotent and safe to run multiple times.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Entities/External/TableDelegationInfo.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Entities/External/TableDelegationInfo.cs
@@ -30,22 +30,22 @@ namespace Microsoft.PowerFx.Core.Entities
         /// </summary>
         public SelectionRestrictions SelectionRestriction { get; init; }
 
-        [Obsolete("preview")]
         /// <summary>
         /// Gets the summarize capabilities for the table.
         /// </summary>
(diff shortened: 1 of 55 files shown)


jas-valgotar:
✅ No public API change.

jas-valgotar:
✅ No public API change.

PR #2931, opened 2025-06-24
Fix 2926

MikeStall:
Better error handling on parsing dates

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
+++ b/src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
@@ -88,6 +88,31 @@ namespace Microsoft.PowerFx.Types
             return FromJson(element, settings, new FormulaValueJsonSerializerWorkingData(), formulaType);  
         }
 
+        internal static FormulaValue ParseDate(JsonElement element, Func<DateTime, FormulaValue> funcParse) 
+        {
+            var strValue = element.GetString(); // caller verified this is a string 
+            if (string.IsNullOrWhiteSpace(strValue))
+            {
+                return FormulaValue.NewBlank();
+            }
+
+            try
+            {
+                var dateTime = element.GetDateTime();
+
+                var value = funcParse(dateTime);
+                return value;
+            }
+            catch (Exception ex)
+            {
+                return FormulaValue.NewError(new ExpressionError
+                {
+                    Kind = ErrorKind.Validation,
+                    Message = $"Can't parse date: '{strValue}'"
+                });
+            }
+        }
+
         internal static FormulaValue FromJson(JsonElement element, FormulaValueJsonSerializerSettings settings, FormulaValueJsonSerializerWorkingData data, FormulaType formulaType = null)
         { 
             if (formulaType is UntypedObjectType uot)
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -88,6 +88,31 @@ public static FormulaValue FromJson(JsonElement element, FormulaValueJsonSeriali
             return FromJson(element, settings, new FormulaValueJsonSerializerWorkingData(), formulaType);  
         }
 
+        internal static FormulaValue ParseDate(JsonElement element, Func<DateTime, FormulaValue> funcParse) 
+        {
+            var strValue = element.GetString(); // caller verified this is a string 
+            if (string.IsNullOrWhiteSpace(strValue))
+            {
+                return FormulaValue.NewBlank();
```
jas-valgotar: > FormulaValue.NewBlank(); [](http://example.com/codeflow?start=22&length=25)

nit: we should add the type to the Blank.

jas-valgotar:
>                         return DecimalValue.New(element.GetDecimal());

Why do these other place not need Blank Handling? 

---
Refers to: src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs:137 in 0ed79ee. [](commit_id = 0ed79ee78a0dcf325b442be7e13410401d46e21f, deletion_comment = False)

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -88,6 +88,31 @@ public static FormulaValue FromJson(JsonElement element, FormulaValueJsonSeriali
             return FromJson(element, settings, new FormulaValueJsonSerializerWorkingData(), formulaType);  
         }
 
+        internal static FormulaValue ParseDate(JsonElement element, Func<DateTime, FormulaValue> funcParse) 
+        {
+            var strValue = element.GetString(); // caller verified this is a string
```
MohamedAbdekader: Can element be null? Could it throw an exception or is that handled? #Pending

src/tests/Microsoft.PowerFx.Json.Tests.Shared/ParseJSONTests.cs
```diff
@@ -581,5 +582,40 @@ public async Task FromArrayWithIncompatibleType2(string prototypeJson, string va
             FormulaValue prototypeValue = FormulaValueJSON.FromJson(prototypeJson);
             Assert.Throws<InvalidOperationException>(() => FormulaValueJSON.FromJson(valueJson, prototypeValue.Type));
         }
+
+        [Fact]
+        public void ParseDates_Blank()
```
MohamedAbdekader, line 587: could we add one more add test to check for null? is it a valid scenario?

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
MikeStall: Non-null. 
Caller has verified that and called `element.ValueKind` , and also verified it's a string

I can add comment. 
(unfortunately, fx codebase doesn't have nullable support)

jas-valgotar:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -88,6 +88,22 @@ public static FormulaValue FromJson(JsonElement element, FormulaValueJsonSeriali
             return FromJson(element, settings, new FormulaValueJsonSerializerWorkingData(), formulaType);  
         }
 
+        // // caller verified element is non-null and is of type string
```
MikeStall, line 91: **Missing XML Documentation**

This new internal method lacks XML documentation. Consider adding:

```csharp
/// <summary>
/// Parses a JSON string element as a date/datetime value.
/// </summary>
/// <param name="element">JSON element containing a date string. Must be non-null and of type string.</param>
/// <param name="targetType">Target FormulaType for blank values.</param>
/// <param name="funcParse">Function to convert parsed DateTime to specific FormulaValue type.</param>
/// <returns>Parsed date value, or BlankValue if string is empty/whitespace.</returns>
/// <exception cref="FormatException">Thrown if string is not a valid date format.</exception>
```

This helps future maintainers understand:
- Caller preconditions
- Exception behavior
- Blank handling logic

MikeStall, line 91: **Comment Typo**

Double slashes in comment: `// // caller verified`

Should be:
```csharp
// Caller verified element is non-null and is of type string
```

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -88,6 +88,22 @@ public static FormulaValue FromJson(JsonElement element, FormulaValueJsonSeriali
...
+        // // caller verified element is non-null and is of type string 
+        internal static FormulaValue ParseDate(JsonElement element, FormulaType targetType, Func<DateTime, FormulaValue> funcParse)
+        {
+            var strValue = element.GetString(); 
+            if (string.IsNullOrWhiteSpace(strValue))
+            {
+                return FormulaValue.NewBlank(targetType);
+            }
+
+            // Any exceptions will be caught at higher level.
```
MikeStall, line 100: **Vague Exception Comment**

The comment "Any exceptions will be caught at higher level" could be more specific about which exceptions and where they're caught.

Consider:
```csharp
// GetDateTime() may throw FormatException if string is not a valid date format.
// Caller (FromJson) wraps all exceptions appropriately.
```

This makes the exception handling contract clearer for maintainers.

PR #2933, opened 2025-06-30
Include children when generating string interpolation call node

rick-nguyen:
PA codegen throws an exception when calling TexlFunction.RequiresPagedDataForParam() because string interpolation nodes were created with an empty list of args. This led to an issue with UDFs that were dependent on a pageable datasource throwing an assert failure when called inside string interpolation.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
@@ -833,9 +833,7 @@ namespace Microsoft.PowerFx.Core.Functions
         public bool RequiresPagedDataForParam(CallNode callNode, int paramIndex, TexlBinding binding)
         {
             Contracts.AssertValue(callNode);
-
-            // ConcatenateFunction is a special case where we may not have any args on call node.
-            Contracts.Assert(this is ConcatenateFunction || callNode.Args.Count > 0);
+            Contracts.Assert(callNode.Args.Count > 0);
             Contracts.Assert(paramIndex >= 0 && paramIndex < callNode.Args.Children.Count());
             Contracts.AssertValue(binding);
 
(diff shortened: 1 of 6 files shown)


jas-valgotar:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/VariadicBase.cs
```diff
@@ -34,7 +34,11 @@ private protected VariadicBase(ref int idNext, Token primaryToken, SourceList so
             foreach (var child in children)
             {
                 Contracts.AssertValue(child);
-                child.Parent = this;
+                if (child.Parent == null)
+                {
```
adithyaselv, line 40: can you comment the reason for this change ?

src/libraries/Microsoft.PowerFx.Core/Functions/TexlFunction.cs
```diff
@@ -832,7 +833,9 @@ protected virtual bool RequiresPagedDataForParamCore(TexlNode[] args, int paramI
         public bool RequiresPagedDataForParam(CallNode callNode, int paramIndex, TexlBinding binding)
         {
             Contracts.AssertValue(callNode);
-            Contracts.AssertValue(callNode.Args);
+
+            // ConcatenateFunction is a special case where we may not have any args on call node.
+            Contracts.Assert(this is ConcatenateFunction || callNode.Args.Count > 0);
```
adithyaselv: if Concatenate can have no Args, 
wont this `callNode.Args.Children[paramIndex]` still fail ?

rick-nguyen: Oops, it should have args because I am including them now from my change in Binder

jas-valgotar:
✅ No public API change.

jas-valgotar:
✅ No public API change.

Shpakh-MSFT:
>             // We generate a transient CallNode (with no arguments) to the Concatenate function

Do we know why we did this way earlier? Also this comment needs to be updated if we are changing it.

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs:2224 in f99d035. [](commit_id = f99d03511faa51a5b3c4d26dd90eb3144a4223f5, deletion_comment = False)

PR #2934, opened 2025-07-08
Restrict void return type in non-behavior UDF

adithyaselv:
Restrict void return type from being used in non-behavior UDF. Also blocks Void to be used in aggregate types in UDTs. 

image

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Data;
 using System.Linq;
-using System.Xml.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.PowerFx.Core.App;
 using Microsoft.PowerFx.Core.App.Controls;
--- a/src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
@@ -3,8 +3,10 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Data;
 using System.Linq;
+using System.Xml.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.PowerFx.Core.App;
 using Microsoft.PowerFx.Core.App.Controls;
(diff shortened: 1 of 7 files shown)


src/tests/Microsoft.PowerFx.Core.Tests.Shared/UserDefinedFunctionTests.cs
```diff
@@ -686,6 +686,20 @@ public void TestUDFRestrictedParameterTypes()
                     Assert.Contains(errors, x => x.MessageKey == "ErrUDF_InvalidParamType");
                 }
             }
+        }
+
+        [Theory]
+        [InlineData("F():Void = 5;")]
+        [InlineData("F():Void = Set(x,5);")]
```
Ananya-Gupta24: Could you add a case here for a function with a param as well?

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -403,15 +404,22 @@ private static bool CheckParameters(ISet<UDFArg> args, List<TexlError> errors, I
...
             if (!nameResolver.LookupType(returnTypeToken.Name, out var returnTypeFormulaType))
             {
                 errors.Add(new TexlError(returnTypeToken, DocumentErrorSeverity.Severe, TexlStrings.ErrUDF_UnknownType, returnTypeToken.Name));
                 returnType = DType.Invalid;
                 return false;
-            }
-            
+            }
+
+            if (!isImperative && returnTypeFormulaType._type.IsVoid)
```
Ananya-Gupta24, line 415: Nit: would this logic be useful as a helper?

adithyaselv: i can't imagine of another place where this would be relevant, i think its fine to be inlined.

Ananya-Gupta24 (dismissed):

jas-valgotar:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Data;
```
CarlosFigueiraMSFT: Needed?

CarlosFigueiraMSFT (approved):
:shipit:

jas-valgotar:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

PR #2935, opened 2025-07-09
UDFs were not properly marked for Pagination based on binding

rick-nguyen:
This led to issue in PA codegen where UDFs were marked pageable when the binding of UDF was not.

Bug:
```UDF():Number = CountRows(DS);```
Binding of body should be unpagable because CountRows function class is not pageable

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/PowerPlatformConnectorClient.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/PowerPlatformConnectorClient.cs
@@ -194,6 +194,7 @@ namespace Microsoft.PowerFx.Connectors
             req.Headers.Add("x-ms-client-environment-id", "/providers/Microsoft.PowerApps/environments/" + EnvironmentId);
             req.Headers.Add("x-ms-user-agent", UserAgent);
             req.Headers.Add("x-ms-request-url", url);
+            req.Headers.Add("x-ms-enable-selects", "true");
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by jas-valgotar** [comment]
✅ No public API change.

File: src/libraries/Microsoft.PowerFx.Core/Functions/UserDefinedFunction.cs [hunk 1]
```diff
@@ -62,6 +62,14 @@ public override bool IsServerDelegatable(CallNode callNode, TexlBinding binding)
...
             return false;
+        }
+
+        public override bool SupportsPaging(CallNode callNode, TexlBinding binding)
+        {
+            Contracts.AssertValue(callNode);
+            Contracts.AssertValue(binding);
+
+            return _binding.IsPageable(_binding.Top);
         }
```
Shpakh-MSFT, line 74: Do we know why base implementation doesn't return the correct value?

Generally IsDelegatable == IsPageable

Shpakh-MSFT: Please add a comment on why we are overriding pageable implementation here. CountRows is probably unique example in this case.

Shpakh-MSFT (approved):
:shipit:

jas-valgotar:
✅ No public API change.

PR #2937, opened 2025-07-17
Allow additional fields in Dynamic to Strongly typed conversion functions

adithyaselv:
Allows additional fields to be part of the input string in Untyped Object conversion functions.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Json/Functions/JSONFunctionUtils.cs
+++ b/src/libraries/Microsoft.PowerFx.Json/Functions/JSONFunctionUtils.cs
@@ -35,7 +35,7 @@ namespace Microsoft.PowerFx.Core.Utils
             var uo = untypedObjectValue.Impl;
             var jsElement = ((JsonUntypedObject)uo)._element;
 
-            var settings = new FormulaValueJsonSerializerSettings { AllowUnknownRecordFields = false, ResultTimeZone = timeZoneInfo };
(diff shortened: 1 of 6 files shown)


jas-valgotar:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Json/Functions/JSONFunctionUtils.cs
```diff
@@ -35,7 +35,7 @@ public static FormulaValue ConvertUntypedObjectToFormulaValue(IRContext irContex
             var uo = untypedObjectValue.Impl;
             var jsElement = ((JsonUntypedObject)uo)._element;
 
-            var settings = new FormulaValueJsonSerializerSettings { AllowUnknownRecordFields = false, ResultTimeZone = timeZoneInfo };
+            var settings = new FormulaValueJsonSerializerSettings { AllowUnknownRecordFields = true, ResultTimeZone = timeZoneInfo };
```
CarlosFigueiraMSFT, line 38: >true [](http://example.com/codeflow?start=95&length=4)

With this change, is there any other place where AllowUnknownRecordFields is false? If not, can we remove that property from FormulaValueJSON?

CarlosFigueiraMSFT: Actually, nevermind, it's a public property on a public class, we can't remove it.

adithyaselv: Also AFAIK we may introduce overloads to these functions that have stricter check. this would help there

PR #2938, opened 2025-07-18
Fixes NRE

jas-valgotar:


CarlosFigueiraMSFT (approved):
🚢

MikeStall (approved):

src/libraries/Microsoft.PowerFx.Connectors/Public/CdpRecordType.cs
```diff
@@ -102,14 +102,19 @@ private bool TryGetFieldType(string fieldName, bool ignorelationship, out Formul
...
+
                 ConnectorType connectorType = TableResolver.ResolveTableAsync(tableName, CancellationToken.None).ConfigureAwait(false).GetAwaiter().GetResult();
 
                 type = connectorType.FormulaType;
                 return true;
             }
             catch (Exception ex)
             {
-                TableResolver?.Logger.LogException(ex, $"Cannot resolve external table {tableName}");
+                TableResolver?.Logger?.LogException(ex, $"Cannot resolve external table {tableName}");
```
MikeStall, line 117: Note it TableResolver is null, then we still won't log anything.

src/libraries/Microsoft.PowerFx.Connectors/Public/CdpRecordType.cs
```diff
@@ -102,14 +102,19 @@ private bool TryGetFieldType(string fieldName, bool ignorelationship, out Formul
 
             try
             {
+                if (TableResolver == null)
```
MikeStall, line 105: Should this be outside the try/Catch?

src/libraries/Microsoft.PowerFx.Connectors/Public/CdpRecordType.cs
```diff
@@ -102,14 +102,19 @@ private bool TryGetFieldType(string fieldName, bool ignorelationship, out Formul
...
+                {
+                    throw new InvalidOperationException("TableResolver is not set.");
+                }
+
                 ConnectorType connectorType = TableResolver.ResolveTableAsync(tableName, CancellationToken.None).ConfigureAwait(false).GetAwaiter().GetResult();
 
                 type = connectorType.FormulaType;
                 return true;
             }
             catch (Exception ex)
```
MikeStall, line 115: fwiw - the usage of Catch(Exception ) is a huge anti-pattern.

jas-valgotar:
✅ No public API change.

PR #2944, opened 2025-07-31
Adds aliasing capabilities.

jas-valgotar:


jas-valgotar:
✅ No public API change.

PR #2945, opened 2025-07-31
QFE DO NOT MERGE/REBASE

jas-valgotar:


jas-valgotar:
✅ No public API change.

PR #2947, opened 2025-08-11
Change trace level from error to warning

Robulane:
Having the http status code higher than 300 or even 500 doesn't necessary mean a service error.


src/libraries/Microsoft.PowerFx.Connectors/Execution/HttpFunctionInvoker.cs
```diff
@@ -487,11 +487,11 @@ private async Task<FormulaValue> ExecuteHttpRequest(string cacheScope, bool thro
 
             if ((int)response.StatusCode >= 300)
             {
-                _logger?.LogError($"In {nameof(HttpFunctionInvoker)}.{nameof(ExecuteHttpRequest)}, response status code: {(int)response.StatusCode} {response.StatusCode}");
+                _logger?.LogWarning($"In PowerFX {nameof(HttpFunctionInvoker)}.{nameof(ExecuteHttpRequest)}, response status code: {(int)response.StatusCode} {response.StatusCode}");
```
MikeStall: super nit: `Power Fx`
space, lowercase x.

MikeStall (approved):

PR #2949, opened 2025-08-12
Remove logging of error message, move responsibility to caller

madevoge:
Since we don't know what is inside this error message, move the responsibility of logging this to the caller where there is more context on the call itself.


jas-valgotar (approved):

MikeStall (approved):

jas-valgotar:
✅ No public API change.

PR #2951, opened 2025-08-18
Fix possible NullReferenceException in DType.TryGetTypeCore

CarlosFigueiraMSFT:
We see NREs in telemetry for that function, and there's only one place on which a null value can be dereferenced. We haven't found a way to reproduce it locally (at least not with debug build, only in release), so submitting this change as a tentative fix.


jas-valgotar:
✅ No public API change.

CarlosFigueiraMSFT:
Abandoning it

PR #2952, opened 2025-08-21
When expected return type is a blob, download HTTP response as bytes instead of string

EricBuist:
The current PowerFx connector code always downloads the HTTP response from connector as a string. If the response is binary, the string is then converted to binary. This binary->string->binary causes data corruption on binary files while it works correctly on text file, especially ASCII-only text. We instead need to detect that the output type is a byte stream and download the bytes, storing them in a blob formula value instead of a string formula value.


EricBuist:
@microsoft-github-policy-service agree [company="Microsoft"]

EricBuist:
@microsoft-github-policy-service agree company="Microsoft"

PR #2959, opened 2025-09-07
Prevent non-imperative UDFs from calling imperative UDFs

gregli-msft:
Currently, a single binding context is used for all UDFs to be added. The `AllowsSideEffects` for each UDF needs to be different since some UDFs are defined to be imperative (with curly braces) and some are not. As a result we currently allow `F():Number = {5}; G():Number = F();` which we should not.  

Canvas currently blocks this, presumably somewhere in PA Client. But Power Fx core should block it as a compile time error.

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests.Shared/UserDefinedFunctionTests.cs
+++ b/src/tests/Microsoft.PowerFx.Core.Tests.Shared/UserDefinedFunctionTests.cs
@@ -701,5 +701,21 @@ namespace Microsoft.PowerFx.Core.Tests
             var errors = checkResult.ApplyErrors();
             Assert.Contains(errors, e => e.MessageKey.Contains("ErrUDF_NonImperativeVoidType"));
         }
+
+        [Theory]
+        [InlineData("F():Number = 5;", true)]
+        [InlineData("F():Number = {5};", true)]
+        [InlineData("F():Number = {5}; G():Number = F();", false)]
+        [InlineData("F():Number = 5;   G():Number = F();", true)]
+        [InlineData("F():Number = {5}; G():Number = {F()};", true)]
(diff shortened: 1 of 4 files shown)


jas-valgotar:
✅ No public API change.

jas-valgotar:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/UserDefinedFunctionTests.cs
```diff
@@ -700,6 +700,22 @@ public void TestNonBehaviorUDFReturnsVoid(string expression)
...
             var errors = checkResult.ApplyErrors();
             Assert.Contains(errors, e => e.MessageKey.Contains("ErrUDF_NonImperativeVoidType"));
+        }
+
+        [Theory]
+        [InlineData("F():Number = 5;", true)]
+        [InlineData("F():Number = {5};", true)]
+        [InlineData("F():Number = {5}; G():Number = F();", false)]
+        [InlineData("F():Number = 5;   G():Number = F();", true)]
+        [InlineData("F():Number = {5}; G():Number = {F()};", true)]
```
CarlosFigueiraMSFT, line 712: >[InlineData("F():Number = {5}; G():Number = {F()};" [](http://example.com/codeflow?start=8&length=51)

Please add a test where an imperative UDF calls a declarative one, to complete the matrix:

    [InlineData("F():Number = 5; G():Number = {F()};", true)]

gregli-msft: Done! I also added a non-UDF functions to the test.

jas-valgotar:
✅ No public API change.

jas-valgotar:
✅ No public API change.

PR #2966, opened 2025-09-10
Support := syntax for named formulas, with compatibility option for no colon

gregli-msft:
We decided last year that `:=` should be used instead of `=` for defining UDTs. We did this because we wanted to rationalize the syntax of Named Formulas:
```
x = 3; // no colon, and no way to express return type
```
with UDFs:
```
x(): Text = 3; // colon, and return type is required
```
To rationalize, we should make the syntax consistently `:=`, with or without a type specification.

Since Canvas already has `=` syntax in GA and has been in use for years, we have not changed it there. But we have not GA'd the REPL or other products that may support UDFs, and we should fix this before then.

This PR introduces a new ParserOption to support the `=` syntax in Canvas, _in addition_ to the new `:=` syntax. This option is off by default for other Power Fx hosts. But we would turn this option on for Canvas.


jas-valgotar:
✅ No public API change.

jas-valgotar:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/NamedFormulasTests.cs
```diff
@@ -73,6 +75,39 @@ public void FailParsingTest(string script)
...
+        [InlineData("a := 3;", false)]
+        [InlineData("a = \"hello\"; b = \"world\"; c := \"colon is optional\";", true)]
+        [InlineData("a := \"hello\"; b = \"world\"; c := \"colon is optional\";", true)]
+        [InlineData("a := \"hello\"; b := \"world\"; c := \"colon is required\";", false)]
+        public void OptionalColonTests(string script, bool requiresEqualOnly)
+        {
+            var parserOptions = new ParserOptions()
+            {
+                AllowsSideEffects = false,
+                AllowEqualOnlyNamedFormulas = requiresEqualOnly
```
jas-valgotar: >requiresEqualOnly [](http://example.com/codeflow?start=46&length=17)

super nit: would be good to have a test which does not set `false` explicitly so we have a test for default behavior.
(Please ignore if there is already such test in some other place)

gregli-msft: Done! Added testing with default.

jas-valgotar:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Core.Parser.ParseFormulasResult.ParseFormulasScript(System.String,System.Globalization.CultureInfo)

jas-valgotar:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Core.Parser.ParseFormulasResult.ParseFormulasScript(System.String,System.Globalization.CultureInfo)

jas-valgotar:
✅ No public API change.

jas-valgotar:
✅ No public API change.

jas-valgotar:
✅ No public API change.

PR #2967, opened 2025-09-10
Fix reassembly of link URLs on error resources

CarlosFigueiraMSFT:
The assembly of the error resources is incorrectly placing the link description in the URL field. This fixes it.


anderson-joyle (approved):
:shipit:

jas-valgotar:
✅ No public API change.

ian-legler (approved):

PR #2969, opened 2025-09-11
Engine configuration for UDT supported data types

gregli-msft:
Currently, UDFs and UDTs have a restricted type list in UserDefinitions.cs that is not customizable per host. It includes Decimal which makes sense for Canvas but doesn't work for C# hosts. We need to generalize this.

- Canvas doesn't want Decimal.
- C# interpreter doesn't want TimeZoneTZInd. Crashes if used.
- Dataverse formula columns want both of these.

This PR uses the Engine.PrimitiveTypes list, to be set by a host that cares about UDFs, UDTs, or JSON serialization. It is null by default, and if a host doesn't use these features, it doesn't matter. RecalcEngine will provide PrimitiveTypes that the interpreter can support, so any hosts using it, which is most hosts, don't need to make a change.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/Features.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/Features.cs
@@ -73,11 +73,6 @@ namespace Microsoft.PowerFx
         internal bool JsonFunctionAcceptsLazyTypes { get; init; }
 
         /// <summary>
-        /// Enables more robust lookup reduction delegation.
-        /// </summary>
-        internal bool IsLookUpReductionDelegationEnabled { get; init; }
-
-        /// <summary>
         /// Enables User-defined types functionality.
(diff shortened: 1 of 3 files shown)


jas-valgotar:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Engine.set_PrimitiveTypes(Microsoft.PowerFx.ReadOnlySymbolTable)

jas-valgotar:
Public API changes: <br />❌CP0002 M:Microsoft.PowerFx.Engine.set_PrimitiveTypes(Microsoft.PowerFx.ReadOnlySymbolTable)

jas-valgotar:
✅ No public API change.

jas-valgotar:
✅ No public API change.

jas-valgotar:
✅ No public API change.

PR #2977, opened 2025-09-23
Support integer enums for OpenAPI

anderson-joyle:
This pull request introduces support for integer-based enums in Swagger/OpenAPI connector parameter parsing, allowing these enums to be handled either as primitive decimals or as option sets, based on connector settings. It also adds new tests and test resources to validate this behavior.

**Support for integer-based enums in connector parsing:**

* Updated `TryGetOptionSet` in `OpenApiExtensions.cs` to support parameters with integer enums, mapping them to either decimals or option sets depending on the `ReturnEnumsAsPrimitive` setting. Previously, integer enums were marked as unsupported.
* Improved handling for string-based enums by ensuring the dictionary creation uses the correct enum names in `TryGetOptionSet`.

**Testing enhancements:**

* Added a new test case `EnumTypeInteger` in `OpenApiParserTests.cs` to verify that integer enums are correctly parsed as decimals when `ReturnEnumsAsPrimitive` is true, and as option sets otherwise.
* Added the corresponding Swagger file `enum-integer.json` as an embedded test resource. [[1]](diffhunk://#diff-86e84f2180955760f4b41d508d285d539a4268ef0d3a78ab0dcadf57288ed25bR1-R53) [[2]](diffhunk://#diff-bdf174ba6be83bc868dc2168d02bbcfe53b306f8d766f3187399a819e1be2531R327-L332)

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
@@ -766,9 +766,7 @@ namespace Microsoft.PowerFx.Connectors
                     else if (schema.Enum.All(e => e is OpenApiInteger))
                     {
                         var enumName = openApiParameter.Name;
-#pragma warning disable CA1305 // Specify IFormatProvider
-                        Dictionary<DName, DName> dic = schema.Enum.Select(e => new DName((e as OpenApiInteger).Value.ToString())).ToDictionary(k => k, e => e);
-#pragma warning restore CA1305 // Specify IFormatProvider
+                        Dictionary<DName, DName> dic = schema.Enum.Select(e => new DName((e as OpenApiInteger).Value.ToString(CultureInfo.InvariantCulture))).ToDictionary(k => k, e => e);
 
                         if (settings.Settings.ReturnEnumsAsPrimitive)
                         {


jas-valgotar:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Connectors/OpenApiExtensions.cs
```diff
@@ -766,8 +765,21 @@ private static ConnectorType TryGetOptionSet(ISwaggerParameter openApiParameter,
                     }
                     else if (schema.Enum.All(e => e is OpenApiInteger))
                     {
-                        // $$$ Not supported yet
-                        return new ConnectorType("Unsupported enum type (int)", openApiParameter.Name, FormulaType.BindingError);
+                        var enumName = openApiParameter.Name;
+#pragma warning disable CA1305 // Specify IFormatProvider
```
MikeStall: Can we eliminate this warning and speciy the formatter? Is this CultureInfo.InvariantCulture?

MikeStall: My fear in not addressing this would be if ToString() takes the thread's culture, and for some reason, and so we break in other locales

jas-valgotar:
✅ No public API change.

PR #2980, opened 2025-10-03
Add Simple Expression constraint

lesaltzm:
Some properties that depend on ThisItem scope will have significant performance costs to async operations. We've discussed and landed on restricting the set of things that can be accessed from those properties to just sync functions, the thisitem scope, enums, and operators. 

This PR is a first pass at this, but it won't be immediately exposed to customers, so we'll iterate on it.


src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/If.cs
```diff
@@ -27,6 +27,8 @@ internal sealed class IfFunction : BuiltinFunction
 
         public override bool IsSelfContained => true;
 
+        public override bool IsAllowedInSimpleExpressions => true;
```
lesaltzm, line 30: also add switch

CarlosFigueiraMSFT:
>             return new BindingConfig(

Need to pass the enforceSimpleExpressions parameter to the .ctor when cloning.

---
Refers to: src/libraries/Microsoft.PowerFx.Core/Binding/BindingConfig.cs:57 in 8f2a273. [](commit_id = 8f2a273305ce99dd23d45eccdb1af496d61fd0e2, deletion_comment = False)

src/strings/PowerFxResources.en-US.resx
```diff
@@ -4771,6 +4771,22 @@
     <value>The value passed to the '{0}' function cannot be changed.</value>
     <comment>Error message shown to the user if they try to use a mutation function (such as Patch / Collect) with a value that cannot be mutated.</comment>
   </data>
+  <data name="ErrorResource_ErrViolatedSimpleConstraintAccess_ShortMessage" xml:space="preserve">
+    <value>{0} cannot be accessed in expressions on this property.</value>
+    <comment>Error Message. This property restricts the set of values that can be referenced to just  </comment>
+  </data>
+  <data name="ErrorResource_ErrViolatedSimpleConstraintAccess_LongMessage" xml:space="preserve">
+    <value>This property restricts access to values outside of ThisItem to ensure good performance. {0} is not a value that can be accessed here.</value>
```
CarlosFigueiraMSFT, line 4779: >ThisItem [](http://example.com/codeflow?start=63&length=8)

I think you mean ThisRecord here, instead of ThisItem, right?

lesaltzm: No, I meant ThisItem. The point of the SimpleExpression constraint is for use in controls like Data Grid and Combobox where we want properties to have ThisItem access but not be async

jas-valgotar:
✅ No public API change.

CarlosFigueiraMSFT (approved):
:shipit:

jas-valgotar:
✅ No public API change.

PR #2982, opened 2025-10-07
Support for more functions that take type arguments

CarlosFigueiraMSFT:
Currently the support for user-defined type arguments is hardcoded to specific functions. This makes the support more generic, allowing new functions to be created that receive UDTs as arguments.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -4421,23 +4421,15 @@ namespace Microsoft.PowerFx.Core.Binding
 
                     if (overloadsWithTypeArgs.Any() && node.Args.Count > 1)
                     {
-                        var nodeInp = node.Args.Children[0];
-                        nodeInp.Accept(this);
-
                         Contracts.Assert(overloadsWithTypeArgs.Count() == 1);
 
                         var functionWithTypeArg = overloadsWithTypeArgs.First();
 
-                        if (MatchOverloadWithUntypedOrJSONConversionFunctions(node, functionWithTypeArg))
+                        // Either one of the untyped JSON conversion functions,
+                        // or other functions where the last argument is a type
+                        if (MatchOverloadWithUntypedOrJSONConversionFunctions(node, functionWithTypeArg) ||
+                            functionWithTypeArg.ArgIsType(node.Args.Count - 1))
                         {
-                            PreVisitTypeArgAndProccesCallNode(node, functionWithTypeArg);
-                            FinalizeCall(node);
-                            return false;
-                        }
-
-                        if (functionWithTypeArg.ArgIsType(node.Args.Count - 1))
-                        {
-                            // Other functions where the type argument is the last one
                             PreVisitTypeArgAndProccesCallNode(node, functionWithTypeArg);
                             FinalizeCall(node);
                             return false;
@@ -5215,7 +5207,7 @@ namespace Microsoft.PowerFx.Core.Binding
                 }
 
                 // Process other non-type arguments
-                for (var i = 1; i < args.Length - 1; i++)
+                for (var i = 0; i < args.Length - 1; i++)
                 {
                     args[i].Accept(this);
                 }
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -4435,6 +4435,14 @@ namespace Microsoft.PowerFx.Core.Binding
                             return false;
                         }
 
+                        if (functionWithTypeArg.ArgIsType(node.Args.Count - 1))
+                        {
+                            // Other functions where the type argument is the last one
+                            PreVisitTypeArgAndProccesCallNode(node, functionWithTypeArg);
+                            FinalizeCall(node);
+                            return false;
+                        }
+
                         startArg++;
                     }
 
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4434,6 +4434,14 @@ public override bool PreVisit(CallNode node)
                             FinalizeCall(node);
                             return false;
                         }
+
+                        if (functionWithTypeArg.ArgIsType(node.Args.Count - 1))
```
lesaltzm: should this be an Or on the condition above on 4431? Seems like the body is the same

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -5185,23 +5195,29 @@ private void PreVisitTypeArgAndProccesCallNode(CallNode node, TexlFunction func)
...
                 else
                 {
-                    _txb.ErrorContainer.Error(args[1], TexlStrings.ErrInvalidArgumentExpectedType, args[1]);
-                    _txb.SetType(args[1], DType.Error);
+                    _txb.ErrorContainer.Error(typeArg, TexlStrings.ErrInvalidArgumentExpectedType, typeArg);
+                    _txb.SetType(typeArg, DType.Error);
+                }
+
+                // Process other non-type arguments
+                for (var i = 1; i < args.Length - 1; i++)
```
lesaltzm, line 5229:
```                    if (overloadsWithTypeArgs.Any() && node.Args.Count > 1)
                    {
                        var nodeInp = node.Args.Children[0];
                        nodeInp.Accept(this);
                        Contracts.Assert(overloadsWithTypeArgs.Count() == 1);
                        var functionWithTypeArg = overloadsWithTypeArgs.First();
```

This bit seems odd (lines 4424-4425 earlier). Why do we visit the 1st arg first for these functions? Should we move that into the loop here as well?

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
CarlosFigueiraMSFT: Good point, will do

lesaltzm:
Any test to add? Maybe a mock function that defines argistype? 

also not really necessary, the canvas side will cover it, just checking.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
CarlosFigueiraMSFT: Sure, I don't see why it was made differently. Will move it

jas-valgotar:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
CarlosFigueiraMSFT: Actually the first argument visit is a requirement for the call to MatchOverloadWithUntypedOrJSONConversionFunctions. We probably shouldn't have that dependency, but I don't think I'll touch that code that is working right now

jas-valgotar:
✅ No public API change.

CarlosFigueiraMSFT:
> Any test to add? Maybe a mock function that defines argistype?
> 
> also not really necessary, the canvas side will cover it, just checking.

Added a test to cover this scenario.

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Copilot.cs
```diff
@@ -0,0 +1,94 @@
...
+using Microsoft.PowerFx.Core.Functions;
+using Microsoft.PowerFx.Core.Localization;
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Syntax;
+
+namespace Microsoft.AppMagic.Authoring.Texl
+{
+    // Copilot(Prompt:Text): Text
+    // Copilot(Prompt:Text, Input:Any): Text
+    // Copilot(Prompt:Text, Input:Any, ReturnType:AnyType): AnyType
```
gregli-msft: >AnyType [](http://example.com/codeflow?start=60&length=7)

This should be "Any (of type ReturnType)", not "AnyType"?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Copilot.cs
```diff
@@ -0,0 +1,94 @@
...
+
+using System.Collections.Generic;
+using Microsoft.PowerFx.Core.App.ErrorContainers;
+using Microsoft.PowerFx.Core.Errors;
+using Microsoft.PowerFx.Core.Functions;
+using Microsoft.PowerFx.Core.Localization;
+using Microsoft.PowerFx.Core.Types;
+using Microsoft.PowerFx.Syntax;
+
+namespace Microsoft.AppMagic.Authoring.Texl
```
jas-valgotar: >Microsoft.AppMagic.Authoring.Texl [](http://example.com/codeflow?start=10&length=33)

nit: should we update the namespace?

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Copilot.cs
CarlosFigueiraMSFT: Updated

src/libraries/Microsoft.PowerFx.Core/Texl/Builtins/Copilot.cs
CarlosFigueiraMSFT: Great catch, updating

jas-valgotar:
✅ No public API change.

PR #2983, opened 2025-10-07
QueryMarshallerSettings - c#8 init support

amiraf:
Add constructor support for QueryMarshallerSettings to enable initialization in C# versions prior to 9.0

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Values/DelegationParameters.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Values/DelegationParameters.cs
@@ -120,6 +120,12 @@ namespace Microsoft.PowerFx.Types
         public QueryMarshallerSettings()
         {
         }
+
+        public QueryMarshallerSettings(bool encodeDateAsString = false, bool encodeBooleanAsInteger = false)
+        {
+            EncodeDateAsString = encodeDateAsString;
+            EncodeBooleanAsInteger = encodeBooleanAsInteger;
+        }
     }
 
     /// <summary>


src/libraries/Microsoft.PowerFx.Core/Public/Values/DelegationParameters.cs
```diff
@@ -120,6 +120,12 @@ public class QueryMarshallerSettings
         public QueryMarshallerSettings()
         {
         }
+
+        public QueryMarshallerSettings(bool encodeDateAsString = false, bool encodeBooleanAsInteger = false)
```
anderson-joyle, line 124: >bool encodeDateAsString = false, bool encodeBooleanAsInteger = false [](http://example.com/codeflow?start=39&length=68)

What's the difference of adding this new constructor and calling something like `new QueryMarshallerSettings() { EncodeBooleanAsInteger = true, EncodeDateAsString = true }`?

anderson-joyle: Oh, I see this is about 8.0.
Isn't better to make the arguments mandatory to avoid conflicts with the original constructor?

jas-valgotar:
✅ No public API change.

src/libraries/Microsoft.PowerFx.Core/Public/Values/DelegationParameters.cs
amiraf: The reason for using optional parameters is to avoid the need to pass all argument values and allow overriding only what's necessary—even though there are currently only two parameters.

PR #2984, opened 2025-10-13
Fix binding for functions with type args that have non-type overloads

CarlosFigueiraMSFT:
The change #2982 introduced a change that causes a bug with functions that have typed and non-typed overloads. This fixes it. We still need to have a generic way to deal with functions with type arguments, will address this later.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -4314,8 +4314,9 @@ namespace Microsoft.PowerFx.Core.Binding
                 _txb.SetType(node, maybeFunc.ReturnType);
             }
 
-            // Checks if the call node best matches function overloads with UntypedObject/JSON
-            private bool MatchOverloadWithUntypedOrJSONConversionFunctions(CallNode node, TexlFunction maybeFunc)
+            // Checks if the call node best matches function overloads with UntypedObject/JSON,
+            // or the typed overload for the Copilot function
+            private bool MatchOverloadWithUntypedOrJSONConversionOrCopilotFunctions(CallNode node, TexlFunction maybeFunc)
             {
                 Contracts.AssertValue(node);
                 Contracts.AssertValue(maybeFunc);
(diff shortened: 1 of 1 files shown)


src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4342,6 +4342,21 @@ private bool MatchOverloadWithUntypedOrJSONConversionFunctions(CallNode node, Te
                 return false;
             }
 
+            // checks if the call node best matches function overloads with UntypedObject/JSON
```
lesaltzm: extreme nit, only if you push another: start comments with a capital letter

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -4342,6 +4342,21 @@ private bool MatchOverloadWithUntypedOrJSONConversionFunctions(CallNode node, Te
...
+            private bool MatchTypedCopilotOverload(CallNode node, TexlFunction maybeFunc)
+            {
+                Contracts.AssertValue(node);
+                Contracts.AssertValue(maybeFunc);
+                Contracts.Assert(maybeFunc.HasTypeArgs);
+
+                if (maybeFunc is CopilotFunction && node.Args.Count > 2)
```
lesaltzm, line 4343: Honestly, not sure this needs to be another function, it's basically the same as what we're doing in MatchOverloadWithUntypedOrJSONConversionFunctions, you could add it as another condition there and keep all this weirdness together for the eventual refactor

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
CarlosFigueiraMSFT: Simple enough, can fix it.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
CarlosFigueiraMSFT: Sure, will update

jas-valgotar:
✅ No public API change.

Issue #2582, opened 2024-08-07
Public PrettyPrint (similar to TexlParser.Format)


modery:
I'm looking at integrating PowerFx.Core into my application so that I can format code snippets to a more user-friendly format.

I can see that https://github.com/microsoft/Power-Fx/blob/main/src/tests/Microsoft.PowerFx.Core.Tests.Shared/FormatterTests.cs#L259 has got a test that covers what I would need, however the TexlParser.Format (https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Core/Parser/TexlParser.cs#L1964) function is internal only.

I also saw that TexlNode.ToString() prettifies the output, but not in the same way (https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Core/Syntax/Nodes/TexlNode.cs#L278)

For example, using the following code snippet
`If(CountRows(Filter('Time Off Requests','Created On'>=Today()&&Owner=LookUp(Users_1,'User Name'=User().Email,User)))>0,Navigate([@'Attendance Already Submitted'],ScreenTransition.Cover),NewForm(AttendanceForm))`

Running the above through TexlNode.ToString() (via _Engine.Parse("...").Root.ToString()_) gives me 
`If(CountRows(Filter('Time Off Requests', 'Created On' >= Today() && Owner = LookUp(Users_1, 'User Name' = User().Email, User))) > 0, Navigate([@'Attendance Already Submitted'], ScreenTransition.Cover), NewForm(AttendanceForm))`
(so some spacing has been added)

However, TexlParser.Format would return
```
If(
    CountRows(
        Filter(
            'Time Off Requests',
            'Created On' >= Today() && Owner = LookUp(
                Users_1,
                'User Name' = User().Email,
                User
            )
        )
    ) > 0,
    Navigate(
        [@'Attendance Already Submitted'],
        ScreenTransition.Cover
    ),
    NewForm(AttendanceForm)
)
```

Is there a possibility to get a public PrettyPrint function somewhere (TexlNode.PrettyPrint(), for example)?

MikeStall:
Totally possible! Would you like to submit a Pull request to enable this?

Possible places to expose this are either:
- a new ToString(...) overload on TexlNode.
- something on CheckResult. That already has helpers to anonymize it. https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs

PR #3030, opened 2026-03-11
Publicly accessible PrettyPrint in CheckResult #2582

modery:
Currently there is no publicly accessible PrettyPrint function. The TexlParser.Format function does the right job, but it is only accessible internally. By providing a publicly accessible function, other tools can make use of it to format PowerFx code in a nicer way.
This contribution adds a publicly accessible PrettyPrint function, CheckResult.ApplyFormat(), as requested in #2582

### Fix
‎src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
CheckResult.cs has been updated with a new function called ApplyFormat(), which returns the pretty-printed form of the expression. This way, we can use an Engine() object with CheckResult, set the text, and apply formatting:
```
var check = new CheckResult(new Engine(new PowerFxConfig()));
check.SetText(script);
var formatted = check.ApplyFormat();
```

### Tests
‎src/tests/Microsoft.PowerFx.Core.Tests.Shared/FormatterTests.cs
Test cases have been added similar to the existing TestPrettyPrint test case

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/CheckResult.cs
@@ -11,6 +11,7 @@ using Microsoft.PowerFx.Core.Binding;
 using Microsoft.PowerFx.Core.IR;
 using Microsoft.PowerFx.Core.Localization;
 using Microsoft.PowerFx.Core.Logging;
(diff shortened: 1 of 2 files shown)


modery:
Hi, anything else needed from my side?

PR #3046, opened 2026-03-31
Fix NRE on type resolution

CarlosFigueiraMSFT:
If we have a type definition that references an invalid type declaration, that is currently throwing a NullReferenceException during the type resolution. This fixes it.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Types/DefinedTypeResolver.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Types/DefinedTypeResolver.cs
@@ -137,7 +137,11 @@ namespace Microsoft.PowerFx.Core.Types
                     // Ensure no edges from non-existent node.
                     if (_typesDict.ContainsKey(processFirst))
                     {
-                        _edges.Add(new TopologicalSortEdge<string>(processFirst, definedType.Ident.Name.Value));
+                        // Ensure no edges to non-existent node.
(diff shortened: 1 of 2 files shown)


jas-valgotar:
✅ No public API change.

tehcrashxor (approved):

lesaltzm (approved):

src/libraries/Microsoft.PowerFx.Core/Types/DefinedTypeResolver.cs
```diff
@@ -166,7 +170,8 @@ private bool CheckTypeName(DefinedType dt)
 
         private IReadOnlyDictionary<DName, FormulaType> ResolveTypes()
         {
-            var containsCycles = !TopologicalSort.TrySort(_nodes, _edges, out var resolveOrder, out var cycles);
+            TopologicalSort.TrySort(_nodes, _edges, out var resolveOrder, out var cycles);
```
lesaltzm, line 173: do we not care about this return value/do we assume it always succeeds?

CarlosFigueiraMSFT: The assumption was that it would always succeed. It was wrong before, but I'm assuming it is correct now that we're creating the graph correctly

jas-valgotar:
✅ No public API change.

PR #3047, opened 2026-04-01
Allow Left/Right in simple expressions

CarlosFigueiraMSFT:
Those are simple functions that should be allowed. Mid is currently allowed today, there's no reason for Left/Right to be forbidden.


lesaltzm (approved):

jas-valgotar:
✅ No public API change.

PR #3048, opened 2026-04-02
Fix ParseDate to handle non-strict ISO 8601 date formats from connectors

EricBuist:
ParseDate used element.GetDateTime() which only accepts strict ISO 8601. Connector APIs like Jira return valid dates with timezone offsets without colons (e.g. "+0000" instead of "+00:00") that GetDateTime() rejects. Now falls back to DateTimeOffset.TryParse on FormatException, and returns an ErrorValue instead of throwing if both parsers fail.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
+++ b/src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
@@ -87,17 +87,41 @@ namespace Microsoft.PowerFx.Types
             return FromJson(element, settings, new FormulaValueJsonSerializerWorkingData(), formulaType);  
         }
 
-        // // caller verified element is non-null and is of type string 
+        // // caller verified element is non-null and is of type string
         internal static FormulaValue ParseDate(JsonElement element, FormulaType targetType, Func<DateTime, FormulaValue> funcParse)
         {
-            var strValue = element.GetString(); 
+            var strValue = element.GetString();
             if (string.IsNullOrWhiteSpace(strValue))
             {
                 return FormulaValue.NewBlank(targetType);
             }
 
-            // Any exceptions will be caught at higher level. 
-            var dateTime = element.GetDateTime();
+            DateTime dateTime;
+
+            try
+            {
+                dateTime = element.GetDateTime();
+            }
+            catch (FormatException)
+            {
+                // element.GetDateTime() uses a strict ISO 8601 parser that rejects valid
+                // date formats commonly returned by connectors (e.g. timezone offsets
+                // without a colon like "+0000" instead of "+00:00").
+                // Fall back to the more lenient DateTimeOffset.Parse.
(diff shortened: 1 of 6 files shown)


src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -87,20 +87,31 @@ public static FormulaValue FromJson(JsonElement element, FormulaValueJsonSeriali
...
             if (string.IsNullOrWhiteSpace(strValue))
             {
                 return FormulaValue.NewBlank(targetType);
             }
 
-            // Any exceptions will be caught at higher level. 
-            var dateTime = element.GetDateTime();
+            // Use DateTimeOffset.TryParse for consistent, machine-independent parsing that matches
+            // Power Automate / Logic Apps behavior. Treats offset-less values as UTC and normalizes
+            // offset-bearing values to UTC, so the resulting instant does not depend on the host's
```
davidrea-MS: typo: instant -> instance

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
```diff
@@ -87,20 +87,31 @@ public static FormulaValue FromJson(JsonElement element, FormulaValueJsonSeriali
...
+            // local timezone. Also accepts non-strict ISO 8601 forms emitted by connectors (e.g.
+            // "+0000" instead of "+00:00") that element.GetDateTime() would reject.
+            if (!DateTimeOffset.TryParse(strValue, CultureInfo.InvariantCulture, DateTimeStyles.AssumeUniversal | DateTimeStyles.AllowWhiteSpaces, out var dto))
```
davidrea-MS: I'm not opposed to including 'AllowWhiteSpaces' in principle but its not consistent with the non-PowerFx parsers.
I'd rather just be as consistent as possible for now unless we know that there's a blocked scenario.

MikeStall (dismissed):

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
EricBuist: Fixed

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
EricBuist: Fixed

src/libraries/Microsoft.PowerFx.Json/FormulaValueJSON.cs
davidrea-MS: I think this got missed?

davidrea-MS (approved):
Typo is still there, otherwise looks good.

davidrea-MS (approved):

anderson-joyle (approved):

PR #3050, opened 2026-04-20
Return null from GetConnectorType on empty/missing schema payload

EricBuist:
When a connector's x-ms-dynamic-schema or x-ms-dynamic-properties sub-call returns a 200 OK response whose body is empty or whose valuePath doesn't resolve, ExtractFromJson yields default(JsonElement) (ValueKind == Undefined). GetConnectorType then forwarded that default value to GetConnectorTypeInternal, which called je.ToString() (producing "") and passed it to OpenApiStringReader.ReadFragment<OpenApiSchema>. Deep inside Microsoft.OpenApi.Readers.OpenApiTextReaderReader.LoadYamlDocument, YamlStream.Documents.First() was invoked on an empty document stream and threw InvalidOperationException("Sequence contains no elements"). The exception surfaced out of GetConnectorReturnTypeAsync / GetConnectorTypeAsync as an opaque LINQ failure with no hint that the real cause was a dynamic-schema payload that didn't contain the expected property.

Real-world repro: Azure DevOps GetQueryResultsV2 declares its items as x-ms-dynamic-properties with operationId=GetQueryResultWorkItemSchema and itemValuePath=Fields. For some valid query shapes (e.g. queries whose saved columnOptions don't map to populated Fields), the sub-call returns 200 OK with a body that has no usable value at Fields. The static 5-layer recursion through GetConnectorTypeInternalAsync then hit LoadYamlDocument and failed with "Sequence contains no elements" instead of simply falling back to the static return type.

# Fix

src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs

GetConnectorType now checks JsonElement.ValueKind against Undefined / Null before delegating to GetConnectorTypeInternal and returns null when the extracted element is not usable. The two async callers (GetConnectorSuggestionsFromDynamicSchemaAsync and GetConnectorSuggestionsFromDynamicPropertyAsync) recognize the new null return, log a warning identifying the offending valuePath / itemValuePath, and return null themselves. The rest of the recursion already treats null as "no dynamic extension resolved" (see GetConnectorTypeInternalAsync line 683), so null propagates cleanly all the way back to GetConnectorReturnTypeAsync callers that can then fall back to the static ReturnParameterType.

# Test

src/tests/Microsoft.PowerFx.Connectors.Tests.Shared/PowerPlatformConnectorTests.cs

Added four direct unit tests for GetConnectorType:
- GetConnectorType_MissingValuePath_ReturnsNull — exact AzDO repro shape ({"otherProperty":"value"} with valuePath="Fields").
- GetConnectorType_EmptyStringValue_ReturnsNull — empty StringValue.
- GetConnectorType_ValuePathResolvesToNull_ReturnsNull — {"Fields":null}.
- GetConnectorType_ValidSchemaAtValuePath_ReturnsConnectorType — happy path, guards against the early-return swallowing valid schemas.

All four pass. Pre-existing GetConnectorType_* and dynamic-schema tests (DVDynamicReturnType, ServiceNowOutputType_GetRecord, ExcelOnlineAddRowV2_ItemParamHasDynamicSchema) continue to pass unchanged.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/ConnectorFunction.cs
@@ -1007,6 +1007,12 @@ namespace Microsoft.PowerFx.Connectors
 
             ConnectorType connectorType = GetConnectorType(cds.ValuePath, sv, ConnectorSettings);
 
+            if (connectorType == null)
+            {
+                runtimeContext.ExecutionLogger?.LogWarning($"Exiting {this.LogFunction(nameof(GetConnectorSuggestionsFromDynamicSchemaAsync))}, no schema at valuePath '{cds.ValuePath}'");
+                return null;
+            }
+
             if (connectorType.HasErrors)
             {
                 runtimeContext.ExecutionLogger?.LogError($"Exiting {this.LogFunction(nameof(GetConnectorSuggestionsFromDynamicSchemaAsync))}, with {LogConnectorType(connectorType)}");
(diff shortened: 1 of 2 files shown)


EricBuist:
Fix implemented as part of resolution of https://portal.microsofticm.com/imp/v5/incidents/details/780317635/summary.

anderson-joyle (approved):

PR #3051, opened 2026-04-20
Fix #2997: PrettyPrint preserves trailing comment

### LINKED ISSUE #2997: PrettyPrint: last comment is removed for user definitions
If we try to pretty print those user definitions:
```
MyNF=Pi()/2; // Half PI
```
The last comment will be deleted.

Test file: FormatterTests.cs; test method: TestUserDefinitionsPrettyPrint
```
[InlineData("MyNF=Pi()/2; // Half PI", "MyNF = Pi() / 2; // Half PI")]
```

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2997

The formatter dropped any comment that appeared after the terminating ";" of the last user definition (e.g. "MyNF=Pi()/2; // Half PI" lost "// Half PI"). Comments between definitions survived because FormatUserDefinitions picks them up via the declaration substring of the following definition, but the final comment has no follower to claim it.

Capture trailing trivia in ParseUserDefinitionResult.TrailingTrivia, set it from the post-semicolon ParseTrivia() in ParseUDFsAndNamedFormulas (the overwriting-each-iteration local naturally ends up holding only the trivia between the last ";" and EOF), and emit any contained comment tokens with a single leading space after the final ";".

Also clear _extraTrivia inside GetExtraTriviaSourceList once its contents are transferred to UserDefinitionSourceInfo.After. Previously the next ParseTrivia() call drained and discarded _extraTrivia; with trailing-trivia capture that same drain would re-emit pre-semicolon comments (e.g. "y = 3/* test */;" rendered as "y = 3/* test */; /* test */"). Clearing on transfer keeps the prior effective behavior while preventing the double-emit.

Adds five InlineData rows to TestUserDefinitionsPrettyPrint covering the issue repro (line comment), a block-comment variant, a multi-definition case that proves only the truly-trailing comment is appended, and UDF / DefinedType variants. The existing test body already re-formats the output to assert idempotence.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Parser/ParseUserDefinitionResult.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Parser/ParseUserDefinitionResult.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.PowerFx.Core.Errors;
 using Microsoft.PowerFx.Syntax;
+using Microsoft.PowerFx.Syntax.SourceInformation;
 
 namespace Microsoft.PowerFx.Core.Parser
 {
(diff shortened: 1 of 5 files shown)


jas-valgotar:
✅ No public API change.

CarlosFigueiraMSFT (approved):
🚢

PR #3052, opened 2026-04-21
Fix PrettyPrint corruption when user-definition header has a trailing…

anderson-joyle:
… // comment

PrettyPrintVisitor.FormatUserDefinitions joined the trimmed declaration with the assignment operator using a single-space separator. When the original source placed a // line comment on the header line and the '=' body on the next line, Trim() stripped the newline that terminated the comment, and the subsequent " = " pushed the operator inside the // comment, corrupting the output so it could no longer be parsed back.

Detect when the trimmed declaration ends inside an unterminated // line comment (tracking /*..*/ nesting so a // inside a block comment is ignored) and, in that case, insert a newline + 4-space indent before the operator. Applied uniformly to NamedFormula, UDF, and DefinedType branches.

Adds two xUnit InlineData cases in TestUserDefinitionsPrettyPrint: the exact repro from the issue and a mixed /*..*/ + // UDF header that covers the generalization named in the issue title.

Issue: https://github.com/microsoft/Power-Fx/issues/2998

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Syntax/TexlPretty.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Syntax/TexlPretty.cs
@@ -518,6 +518,9 @@ namespace Microsoft.PowerFx.Syntax
             foreach (var info in result.UserDefinitionSourceInfos)
             {
                 var declaration = info.Declaration.Trim();
+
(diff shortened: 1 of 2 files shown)


jas-valgotar:
✅ No public API change.

jas-valgotar:
✅ No public API change.

CarlosFigueiraMSFT (approved):

Issue #2893, opened 2025-04-17
throw better exception on unrecognized fields.


MikeStall:
https://github.com/microsoft/Power-Fx/blob/0ba6c707c4402e00d2bb7328d805b661dc74502a/src/libraries/Microsoft.PowerFx.Connectors/Tabular/Services/CdpTable.cs#L152

If we encounter a field we don't recognize - don't just null-ref. Throw a better exception.

PR #3053, opened 2026-04-21
Throw descriptive exception for missing/mistyped 'value' field

anderson-joyle:
CdpTable.GetResult previously extracted the 'value' TableValue via FirstOrDefault(...).Value as TableValue. When the parsed CDP response did not contain a 'value' field, or that field was not a TableValue, this silently produced a null local, which then surfaced as a NullReferenceException on the next line's tv.Rows access - far from the root cause.

Replace the cast with two explicit guards that throw PowerFxConnectorException (the idiomatic exception type already used in this file) with a descriptive message naming the table, the expected 'value' field, and either the observed field names (for the missing case) or the observed runtime type name (for the wrong-type case).

Add two regression tests exercising both failure modes via the existing LoggingTestServer + CdpDataSource pattern.

Tracking: microsoft/Power-Fx#2893

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Connectors/Tabular/Services/CdpTable.cs
+++ b/src/libraries/Microsoft.PowerFx.Connectors/Tabular/Services/CdpTable.cs
@@ -184,7 +184,21 @@ namespace Microsoft.PowerFx.Connectors
                 return new List<DValue<RecordValue>>() { DValue<RecordValue>.Of(err) };
             }
 
-            TableValue tv = rv.Fields.FirstOrDefault(field => field.Name == "value").Value as TableValue;
+            NamedValue valueField = rv.Fields.FirstOrDefault(field => field.Name == "value");
+            if (valueField == null)
+            {
+                throw new PowerFxConnectorException(
+                    $"CDP response for table '{TableName}' is missing the expected 'value' field. " +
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Connectors/Tabular/Services/CdpTable.cs
```diff
@@ -184,7 +184,21 @@ private IReadOnlyCollection<DValue<RecordValue>> GetResult(string text)
...
+            if (valueField.Value is not TableValue tv)
+            {
+                throw new PowerFxConnectorException(
+                    $"CDP response for table '{TableName}' has an unexpected shape: " +
+                    $"the 'value' field was expected to be a table, " +
+                    $"but was {valueField.Value?.GetType().Name ?? "null"}.");
+            }
 
             // The call we make contains more fields and we want to remove them here ('@odata.etag')
             return new InMemoryTableValue(IRContext.NotInSource(TableType), tv.Rows).Rows.ToArray();
```
MikeStall, line 204: It looks like the key difference is before, tv could be null and we'd get a nullref at `tv.Rows` here ; and now we get a better exception

MikeStall (approved):

jas-valgotar:
✅ No public API change.

PR #3054, opened 2026-04-21
Logging skips lazy aggregate field enumeration

anderson-joyle:
Issue: https://github.com/microsoft/Power-Fx/issues/2858

CheckResult.ApplyGetLogging funnels through StructuralPrint.IsExpandedType, which called DType.AggregateHasExpandedType for every FirstName reference. For lazy aggregate types (DKind.LazyRecord / LazyTable used by CDP-style data sources), GetAllNames forces per-field resolution via LazyTypeProvider.TryGetFieldType, potentially materializing relationship metadata that a logging call should never need.

Short-circuit lazy types in StructuralPrint.IsExpandedType so no TryGetFieldType calls happen on the logging path. Non-lazy aggregate behavior is unchanged: the '#$fne$#' marker continues to fire exactly as before for the dominant (non-CDP) population of expressions, and all 223 existing CheckResult + Formatter tests pass unchanged.

Note for consumers that pattern-match on '#$fne$#': for FirstName references whose bound type is a lazy aggregate whose lazily-resolved fields would have contained an expand entity, the marker is no longer emitted. A follow-up can restore marker parity for lazy aggregates by adding a non-materializing predicate to LazyTypeProvider.

Adds LazyLoggingSpyRecordType test helper and
ApplyGetLoggingDoesNotEnumerateLazyFields regression test. Spy asserts TryGetFieldTypeCallCount and FieldNamesIterationCount stay at zero after ApplyGetLogging; without the fix the counter reaches 2 (one per field).

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Logging/StructuralPrint.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Logging/StructuralPrint.cs
@@ -483,7 +483,13 @@ namespace Microsoft.PowerFx.Core.Logging
         {
             var type = _binding?.GetTypeAllowInvalid(node);
 
-            return type != null && type.AggregateHasExpandedType();
+            if (type == null)
+            {
+                return false;
+            }
+
+            // Skip lazy aggregates: AggregateHasExpandedType forces per-field materialization.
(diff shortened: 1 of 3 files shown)


jas-valgotar:
✅ No public API change.

Issue #2876, opened 2025-03-24
Summarize and direct groupby field access result in runtime error or hang


gregli-msft:
This trivial case is reduced from a much bigger example.
```
Summarize( [{Value:1,Sixteen:1}], Sixteen, Dec2Hex( Sixteen, 3 ) As LevelOne )
```
results in runtime errors in the column LevelOne, while this does not:
```
Summarize( [{Value:1,Sixteen:1}], Sixteen, Dec2Hex( First(ThisGroup).Sixteen, 3 ) As LevelOne )
```

![Image](https://github.com/user-attachments/assets/f48e85b4-2285-404e-a8c6-2be965f2f435)

anderson-joyle:
https://github.com/microsoft/Power-Fx/blob/19e7538df82c40b0603a2b4e957ce47e216fb1b5/src/libraries/Microsoft.PowerFx.Interpreter/Functions/LibraryTable.cs#L1436

PR #3055, opened 2026-04-21
Summarize aggregate scope includes scalar groupby keys

anderson-joyle:
Issue: https://github.com/microsoft/Power-Fx/issues/2876

The Summarize runtime was pushing only the group's subset TableValue into the aggregate lambda's scope. The binder (FunctionThisGroupScopeInfo) builds the scope DType as { <key>: scalar, ..., ThisGroup: table }, so FormulaValueScope.Resolve for any bare name fell through to returning the whole table, making Dec2Hex(<key>, ...) error and Text(<key>) hang.

Build a RecordValue matching the binder's shape (scalar key fields plus ThisGroup -> group table) and pass that to WithScopeValues. Adds regression tests covering the Dec2Hex and Text repros from the issue, equivalence with First(ThisGroup).<field>, and a multi-key groupby.


jas-valgotar:
✅ No public API change.

Issue #2821, opened 2025-01-18
Remove PowerFxConfig.AdditionalFunctions


MikeStall:
Remove PowerFxConfig.AdditionalFunctions  
https://github.com/microsoft/Power-Fx/blob/0fa19687742caa771c4a8f0716f07005497bec49/src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs#L55

This is a layering violation. It's a map in Config that maps TexlFunction to Interpreter implementations. 
But Config lives in Fx.Core, so it can't point to interpreter implementations. 

Having "EnableXXX" on config that adds function implementations forces a problem here. 

One workaround we used elsewhere was like for Json - add a derived TexlFunction that implements the interpreter interface.

PR #3056, opened 2026-04-22
Remove PowerFxConfig.AdditionalFunctions

anderson-joyle:
Issue https://github.com/microsoft/Power-Fx/issues/2821

The AdditionalFunctions dictionary on PowerFxConfig (in Fx.Core) mapped TexlFunction keys to IAsyncTexlFunction implementations from the Interpreter, a layering violation since Core should not reference Interpreter implementation types.

Migrate the single producer (EnableRegExFunctions) to the same pattern already used by Join, Patch, and Set: a class that derives from the Core TexlFunction base and also implements the interpreter interface. Three thin wrapper classes (IsMatchImpl, MatchImpl, MatchAllImpl) in LibraryRegEx.cs inherit IsMatchFunction/MatchFunction/MatchAllFunction and delegate InvokeAsync to the existing Implementation classes.

With that, registration is a plain symbolTable.AddFunction call and the EvalVisitor `func is IAsyncTexlFunction` branch handles dispatch. The field, its plumbing through ParsedExpression and EvalVisitor, and the matching test helpers for NodeJS/PCRE2/Compare are all removed.

No public API change. All 64,946 interpreter tests pass.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Public/Config/PowerFxConfig.cs
@@ -52,9 +52,6 @@ namespace Microsoft.PowerFx
             set => _symbolTable = value;
         }
 
-        // Remove this: https://github.com/microsoft/Power-Fx/issues/2821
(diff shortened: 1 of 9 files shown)


jas-valgotar:
✅ No public API change.

jas-valgotar:
✅ No public API change.

PR #3064, opened 2026-05-19
Implement Update mutation function

### LINKED ISSUE #1000: Update function

GHX5T-SOL:
Fixes #1000

Summary:
- adds the interpreter `Update` mutation function and registers it with mutation functions
- builds a full replacement record so omitted fields are blanked during replacement
- adds indexed collection update support for `Update(..., ..., ..., "ALL")` so each matching row is replaced once
- adds V1/V1-disabled mutation script coverage and a primary-key all-match regression

Validation:
- `DOTNET_CLI_TELEMETRY_OPTOUT=1 DOTNET_ROOT=/Users/brucewayne/Money-MacMini/.dotnet PATH=/Users/brucewayne/Money-MacMini/.dotnet:$PATH dotnet test src/tests/.Net7.0/Microsoft.PowerFx.Interpreter.Tests/Microsoft.PowerFx.Interpreter.Tests.csproj --filter "FullyQualifiedName~RunMutationTests|FullyQualifiedName~UpdateAllUpdatesEachPrimaryKeyMatchOnce" /p:EnableSourceControlManagerQueries=false --no-restore`
  - Passed: 55, Failed: 0, Skipped: 0
  - SourceLink warnings only, because local SCM queries were disabled to avoid `Microsoft.Build.Tasks.Git` failing on this worktree with `Unsupported repository version 1`.
- `git diff --check`
- `git diff --binary | gitleaks stdin --redact --no-banner`
  - no leaks found

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Localization/Strings.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Localization/Strings.cs
@@ -530,6 +530,11 @@ namespace Microsoft.PowerFx.Core.Localization
         public static StringGetter AboutRemove = (b) => StringResources.Get("AboutRemove", b);
         public static StringGetter RemoveDataSourceArg = (b) => StringResources.Get("RemoveDataSourceArg", b);
         public static StringGetter RemoveRecordsArg = (b) => StringResources.Get("RemoveRecordsArg", b);
+        public static StringGetter AboutUpdate = (b) => StringResources.Get("AboutUpdate", b);
+        public static StringGetter UpdateDataSourceArg = (b) => StringResources.Get("UpdateDataSourceArg", b);
+        public static StringGetter UpdateOldRecordArg = (b) => StringResources.Get("UpdateOldRecordArg", b);
+        public static StringGetter UpdateNewRecordArg = (b) => StringResources.Get("UpdateNewRecordArg", b);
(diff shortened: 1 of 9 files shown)


GHX5T-SOL:
@microsoft-github-policy-service agree [company="{GHX5T-SOL}"]

GHX5T-SOL:

PR #3067, opened 2026-06-02
Update conversion from untyped (dynamic) values to date to assume local time without time zone information

CarlosFigueiraMSFT:
We had a recent change that broke the assumption that date / time values are local by default. This fixes that assumption.


MikeStall (dismissed):

jas-valgotar:
✅ No public API change.

MikeStall (approved):

PR #3071, opened 2026-06-08
Fix IsType function to not return error for valid untyped input

CarlosFigueiraMSFT:
A recent change made an update on the untyped version of the IsType function where it can return an error for a valid untyped (dynamic) input. If given a non-blank, non-error dynamic input and a valid type, IsType should only return true or false. This fixes it.


MikeStall (approved):

jas-valgotar:
✅ No public API change.

PR #3072, opened 2026-06-08
Preserve incomplete attributes for IntelliSense

VamsiModem:
## Summary
- Preserve parsed attributes that are not yet attached to a UDF or named formula so hosts can provide IntelliSense while makers are mid-typing attribute arguments, e.g. `[RecordLink(`.
- Capture attribute delimiter tokens (`OpenParen`, `CloseParen`, `CloseBracket`) so consumers can compute parser-backed replacement spans for `[RecordLink(` and `[RecordLink()]` without raw script scanning.
- Preserve `IncompleteAttributes` through Partial-attribute post-processing.

## Validation
- `dotnet test tests\.Net7.0\Microsoft.PowerFx.Core.Tests\Microsoft.PowerFx.Core.Tests.csproj --filter "FullyQualifiedName~AttributeParserTests" --nologo -v minimal`

### DIFF
--- a/src/tests/Microsoft.PowerFx.Core.Tests.Shared/AttributeParserTests.cs
+++ b/src/tests/Microsoft.PowerFx.Core.Tests.Shared/AttributeParserTests.cs
@@ -124,6 +124,129 @@ namespace Microsoft.PowerFx.Core.Tests
         }
 
         [Fact]
+        public void TestAnnotationDelimiterTokensCaptured()
+        {
+            var result = UserDefinitions.Parse(
+            @"
+                [MyAttr(someIdent)]
+                MyFunc(): Number = 1;
+            ", _parseOptions);
+
+            Assert.False(result.HasErrors);
+
+            var attribute = result.UDFs.First().Attributes[0];
+            Assert.NotNull(attribute.OpenParen);
+            Assert.Equal(TokKind.ParenOpen, attribute.OpenParen.Kind);
+            Assert.NotNull(attribute.CloseParen);
+            Assert.Equal(TokKind.ParenClose, attribute.CloseParen.Kind);
+            Assert.NotNull(attribute.CloseBracket);
+            Assert.Equal(TokKind.BracketClose, attribute.CloseBracket.Kind);
+        }
+
+        [Fact]
+        public void TestAnnotationWithoutArgsHasNoParenTokens()
+        {
+            var result = UserDefinitions.Parse(
+            @"
+                [SomeName]
+                Foo = 123;
+            ", _parseOptions);
+
+            Assert.False(result.HasErrors);
+            var attribute = result.NamedFormulas.First().Attributes[0];
+            Assert.Null(attribute.OpenParen);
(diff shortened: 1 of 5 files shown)


jas-valgotar:
✅ No public API change.

jas-valgotar:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/AttributeParserTests.cs
```diff
@@ -123,6 +123,129 @@ public void TestAnnotationWithIdentArgOnUDF()
             Assert.Equal("someIdent", udf.Attributes[0].Arguments[0]);
         }
 
+        [Fact]
+        public void TestAnnotationDelimiterTokensCaptured()
```
lesaltzm, line 127: for incomplete definitions, when do we resume the parsing? e.g., what do we get for:

```
Foo = 123;

[Record
Bar(): Void = { Notify("Abc") };

Baz = "Valid?";
```

lesaltzm: What about if there's a start/close bracket in the body of Bar()?

VamsiModem: Yeah, these weren't handled properly, I have fixed it and added tests!

jas-valgotar:
✅ No public API change.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/AttributeParserTests.cs
```diff
@@ -234,6 +234,210 @@ public void TestIncompleteAttribute_MultipleAttributes()
...
+        [Fact]
+        public void TestIncompleteAttribute_ResumesAtNextDefinition()
+        {
+            var parseOptions = new ParserOptions() { AllowAttributes = true, AllowEqualOnlyNamedFormulas = true, AllowsSideEffects = true };
+            var result = UserDefinitions.Parse(
+            @"
+                Foo = 123;
+
+                [Record
+                Bar(): Void = { Notify(""Abc"") };
```
lesaltzm, line 246: could we resume on a newline, as well as waiting for a semicolon (so we still have the Bar() definition?)

VamsiModem: So the parser already resumes mid-stream, the incomplete [Record doesn't swallow Bar(), The semicolon isn't required for recovery; the parser breaks out of MaybeParseAttributes as soon as it sees a missing ] and immediately falls through to TokEat(TokKind.Ident) on the next line, which picks up Bar.

lesaltzm: whoops, yes, nice. misread the test

lesaltzm (approved):

PR #3073, opened 2026-06-08
Mark data-backed record variables async

jaemin-capslock:
Assume a case where
App.OnStart= Set(varAccount, First(Accounts)
Toggle1.Default=Coalesce(varAccount.'Option set', false)

The emitted `Rule.IsAsync` was false because pfx interpreted `varAccount.'Option set'` as a synchronous variable lookup, which depends on the datasource `Accounts`. As line 44 in NameLookupInfo.cs states, they should have been accessed async, hence Rule.IsAsync should have been true. 

This caused issues in Powerapps. Emitted rule had `IsAsync: False`, so it translated that as synchronous js code. But the argument was actually async, which meant we were passing a `Promise` to a sync js function.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/BindInfo/NameLookupInfo.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/BindInfo/NameLookupInfo.cs
@@ -42,7 +42,7 @@ namespace Microsoft.PowerFx.Core.Binding.BindInfo
             DisplayName = displayName;
 
             // Any connectedDataSourceInfo or option set or view needs to be accessed asynchronously to allow data to be loaded.
-            IsAsync = Data is IExternalTabularDataSource || Kind == BindKind.OptionSet || Kind == BindKind.View || isAsync || TypeRequiresAsync(kind, type, data);
+            IsAsync = Data is IExternalTabularDataSource || Kind == BindKind.OptionSet || Kind == BindKind.View || isAsync;
         }
 
         public bool TryToSymbolEntry(out SymbolEntry x)
@@ -52,7 +52,7 @@ namespace Microsoft.PowerFx.Core.Binding.BindInfo
                 x = new SymbolEntry
                 {
                     Name = ns.Name,
-                    DisplayName = this.DisplayName,                
+                    DisplayName = this.DisplayName,
                     Properties = ns.Props,
                     Type = FormulaType.Build(this.Type),
                     Slot = ns
@@ -62,46 +62,6 @@ namespace Microsoft.PowerFx.Core.Binding.BindInfo
 
             x = null;
             return false;
-        }
-
-        private static bool TypeRequiresAsync(BindKind kind, DType type, object data)
-        {
-            Contracts.AssertValid(type);
-
-            if (!IsVariableBinding(kind, data))
-            {
-                return false;
-            }
-
-            if (!type.IsRecord)
-            {
-                return false;
-            }
-
-            foreach (var dataSource in type.AssociatedDataSources)
-            {
-                if (dataSource.RequiresAsync)
-                {
-                    return true;
-                }
-            }
-
-            return type.HasExpandInfo && type.ExpandInfo?.ParentDataSource?.RequiresAsync == true;
-        }
-
-        private static bool IsVariableBinding(BindKind kind, object data)
-        {
-            if (kind == BindKind.ScopeVariable)
-            {
-                return true;
-            }
-
-            if (kind != BindKind.PowerFxResolvedObject || data is not NameSymbol nameSymbol)
-            {
-                return false;
-            }
-
-            return nameSymbol.Owner is not Microsoft.PowerFx.SymbolTableOverRecordType;
         }
     }
 }
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/BindInfo/NameLookupInfo.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/BindInfo/NameLookupInfo.cs
@@ -42,7 +42,7 @@ namespace Microsoft.PowerFx.Core.Binding.BindInfo
             DisplayName = displayName;
 
             // Any connectedDataSourceInfo or option set or view needs to be accessed asynchronously to allow data to be loaded.
-            IsAsync = Data is IExternalTabularDataSource || Kind == BindKind.OptionSet || Kind == BindKind.View || isAsync;
+            IsAsync = Data is IExternalTabularDataSource || Kind == BindKind.OptionSet || Kind == BindKind.View || isAsync || TypeRequiresAsync(kind, type, data);
         }
 
         public bool TryToSymbolEntry(out SymbolEntry x)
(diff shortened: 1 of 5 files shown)


src/libraries/Microsoft.PowerFx.Core/Binding/BindInfo/NameLookupInfo.cs
```diff
@@ -62,6 +62,46 @@ public bool TryToSymbolEntry(out SymbolEntry x)
...
+            if (!type.IsRecord)
+            {
+                return false;
+            }
+
+            foreach (var dataSource in type.AssociatedDataSources)
+            {
+                if (dataSource.RequiresAsync)
+                {
+                    return true;
```
lesaltzm: that's definitely not right. Almost every type from a data source will have an associated data source marked as requires async.

src/libraries/Microsoft.PowerFx.Core/Binding/BindInfo/NameLookupInfo.cs
```diff
@@ -42,7 +42,7 @@ public NameLookupInfo(BindKind kind, DType type, DPath path, int upCount, object
             DisplayName = displayName;
 
             // Any connectedDataSourceInfo or option set or view needs to be accessed asynchronously to allow data to be loaded.
-            IsAsync = Data is IExternalTabularDataSource || Kind == BindKind.OptionSet || Kind == BindKind.View || isAsync;
+            IsAsync = Data is IExternalTabularDataSource || Kind == BindKind.OptionSet || Kind == BindKind.View || isAsync || TypeRequiresAsync(kind, type, data);
```
lesaltzm: are you sure it's not just that we're missing OptionSetValue in this list?

lesaltzm: or that specifically coercion from an optionset value to a string should be marked as async?

lesaltzm (changes requested):

src/libraries/Microsoft.PowerFx.Core/Binding/BindInfo/NameLookupInfo.cs
jaemin-capslock: I did some testing. `First(Accounts).optionsetname` is async, but:
`varAccount.'Account Name'` is sync which is correct
`varAccount.optionsetname` is sync.. ???
which leads to
Coalesce(varAccount.optionsetname, false) is sync.

Fix should be in `Binder.cs`, not here

jaemin-capslock: Coercion seems to be working. I had to specifically set `canCoerceFromBackingKind` and `canCoerceToBackingKind` as true when I defined optionset

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -3783,6 +3788,55 @@ public override void PostVisit(DottedNameNode node)
...
+
+                return false;
+            }
+
+            private static bool IsVariableBackedRecord(FirstNameInfo firstNameInfo)
+            {
+                return firstNameInfo?.Kind == BindKind.ScopeVariable ||
+                    (firstNameInfo?.Kind == BindKind.PowerFxResolvedObject && firstNameInfo.Data is NameSymbol);
+            }
+
```
jaemin-capslock, line 3839: I had to add this, so we don't get things like `ThisRecord` or constants.

src/tests/Microsoft.PowerFx.Core.Tests.Shared/AssociatedDataSourcesTests/TestDelegationValidation.cs
```diff
@@ -154,13 +155,64 @@ public void TestCountRowsWarningForCachedData(bool isCachedData)
...
+                .Add(new TypedName(boolOptionSet.FormulaType._type, new DName("donotallowemails")));
+
+            var dataSource = new TestDataSource("Accounts", accountsType, requiresAsync: true);
+            var dataBackedAccountsType = dataSource.Type;
+
             var symbolTable = new DelegatableSymbolTable();
-            var accountRecordType = AccountsTypeHelper.GetDType().ToRecord();
-            symbolTable.AddVariable("varAccount", FormulaType.Build(accountRecordType));
+            symbolTable.AddVariable("plainAccount", FormulaType.Build(accountsType.ToRecord()));
+            symbolTable.AddVariable("varAccount", FormulaType.Build(dataBackedAccountsType.ToRecord()));
```
CarlosFigueiraMSFT, line 227: What is the difference between a "plain accounts type" and a "data-backed accounts type"? How can we create each of those types in a real application?

jaemin-capslock: plain account is effectively just a local record of a optionset:
Set(plainAccount, { 'Account Name': "xyz", donotallowemails: DoNotAllowEmailsOptions.No})
This should be sync.
while the data backed accounts is assigned to `varAccount`, so when the option set field is in a dataverse table etc. This should be async.

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -3783,6 +3788,55 @@ public override void PostVisit(DottedNameNode node)
                 _txb.SetIsUnliftable(node, _txb.IsUnliftable(node.Left));
             }
 
+            private bool ShouldFlagDottedOptionSetAccessAsAsync(DottedNameNode node, DType typeRhs)
+            {
+                // Data-backed option-set fields can require async loading when accessed from a record producer that can cross the data-source boundary.
```
lesaltzm, line 3793: > "that can cross the data-source boundary" what does that mean?

lesaltzm: Basically, why shouldn't we just treat optionsetvalue access as async? Why this whole runaround, when would a dottednamenode with an rhstype  that's an optionsetvalue not be async?

lesaltzm (changes requested):

src/tests/Microsoft.PowerFx.Core.Tests.Shared/AssociatedDataSourcesTests/TestDelegationValidation.cs
lesaltzm: That doesn't sound right. Our codegen shouldn't differ for `accountVar.donotallowemails` between 

`Set(accountVar, {donotallowemails: DoNotAllowEmailsOptions.No, .... (rest of accounts schema with dummy values)`
and 

`Set(accountVar, First(Accounts)`

(or if we have both, which is also totally valid)

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
jaemin-capslock: Consider `Filter(varAccounts as account, IsBlank(account.address1_addresstypecode))` or same formula using `ThisRecord`. Accessing `varAccounts` can make this rule async, but `account.address1_addresstypecode` shouldn't

PR #3074, opened 2026-06-16
Mark async option-set coercions

jaemin-capslock:
Mark option-set-to-primitive coercions as async when the binder records them, including top-level coercions used by host properties. Keep raw option-set field access sync and avoid row-scope coercion over-marking.

### DIFF
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -1607,6 +1607,11 @@ namespace Microsoft.PowerFx.Core.Binding
             Contracts.AssertValid(fromType);
             Contracts.AssertValid(toType);
 
+            if (toType.Kind == DKind.OptionSetValue)
+            {
+                return fromType.Kind is DKind.Boolean or DKind.Number or DKind.Decimal or DKind.String;
+            }
+
             if (fromType.Kind != DKind.OptionSetValue)
             {
                 return false;
--- a/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
+++ b/src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
@@ -1596,7 +1596,7 @@ namespace Microsoft.PowerFx.Core.Binding
 
             _coerceMap[node.Id] = type;
 
-            if (!IsRowScope(node) && IsAsyncCoercion(GetType(node), type))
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -1595,6 +1595,36 @@ public void SetCoercedType(TexlNode node, DType type)
             Contracts.Assert(!_coerceMap[node.Id].IsValid);
 
             _coerceMap[node.Id] = type;
+
+            if (!IsRowScope(node) && IsAsyncCoercion(GetType(node), type))
```
lesaltzm: why the !IsRowScope check?

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
```diff
@@ -1595,6 +1595,36 @@ public void SetCoercedType(TexlNode node, DType type)
...
+            {
+                return toType.Kind is DKind.String or DKind.Boolean or DKind.Number or DKind.Decimal or DKind.Color;
+            }
+
+            return toType.Kind switch
+            {
+                DKind.String => true,
+                DKind.Boolean => fromType.OptionSetInfo.BackingKind == DKind.Boolean,
+                DKind.Number or DKind.Decimal => fromType.OptionSetInfo.BackingKind == DKind.Number,
+                DKind.Color => fromType.OptionSetInfo.BackingKind == DKind.Color,
```
lesaltzm: is color-> option set an async coercion in JS?

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
jaemin-capslock: For example, `Filter(optionSetRows, ThisRecord.donotallowemails = true)` or `Filter(optionSetRows, ColorFade(ThisRecord.coloroption, 0.5) = RGBA(0, 0, 255, 1))`

src/libraries/Microsoft.PowerFx.Core/Binding/Binder.cs
jaemin-capslock: Currently, we only have `OptionSetToColor` but not the other way around.

lesaltzm: so is this check valid?

lesaltzm (approved):

